PR #1, opened 2023-08-21
Storage Abstractions & Postgres implementation

kayibal:
Some ground work on designing storage and gateways in Tycho.

Had to sketch out some parts of the extractor traits as well although these are not very final.

### DIFF
--- a/src/storage/postgres.rs
+++ b/src/storage/postgres.rs
@@ -36,6 +36,8 @@
 //! throughout the application lifetime, even if the process panics during
 //! database operations.
 use std::collections::HashMap;
+use std::fmt::Display;
+use std::hash::Hash;
 use std::marker::PhantomData;
 use std::sync::Arc;
 
@@ -43,7 +45,6 @@ use async_trait::async_trait;
 use diesel::prelude::*;
 use diesel_async::{AsyncConnection, AsyncPgConnection, RunQueryDsl};
 use ethers::types::{H160, H256};
-use tokio::sync::RwLock;
 
 use super::{BlockIdentifier, ChainGateway, StorableBlock, StorableTransaction};
 use crate::extractor::evm;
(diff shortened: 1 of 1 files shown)


src/storage/mod.rs
```diff
@@ -0,0 +1,558 @@
...
+    }
+}
+
+/// `StorableBlock` abstracts (de)serialization behavior of a block entity
+/// specific to a storage backend.
+///
+/// It defines methods for converting from a storage-specific type to a block,
+/// converting from a block to a storage-specific type, and getting the block's
+/// chain.
+pub trait StorableBlock<S, N> {
```
michcio1234, line 83: Why are S and N different?

src/storage/mod.rs
```diff
@@ -0,0 +1,558 @@
...
+
+/// The `StorableTransaction` is a public trait which lays out the necessary
+/// interface needed to store and retrieve transactions from storage. This trait
+/// is generic over four types - `S`, `N`, `BH`, and `DbId`.
+///
+/// `S` represents the storage specific transaction type e.g. orm struct. `N`
+/// represents the storage specific type for a new transaction,. `BH`
+/// corresponds to the block hash type that uniquely identifies a block. `DbId`
+/// is the ID used to refer to each transaction in the database.
+pub trait StorableTransaction<S, N, BH, DbId>
```
michcio1234, line 123: Shouldn't `BH` be replaced with something like `BId`? IIUC, block can be identified by hash or by number.
Unless we intentionally want hash here, because number might be ambiguous.

src/storage/mod.rs
```diff
@@ -0,0 +1,558 @@
...
+    /// - An Ok result containing the transaction. Might fail if the transaction
+    ///   does not exist yet.
+    async fn get_tx(
+        &self,
+        hash: &[u8],
+        db: &mut Self::DB,
+    ) -> Result<Self::Transaction, StorageError>;
+}
+
+/// Store and retrieve state for Extractors.
```
michcio1234, line 228: Here I got confused. I thought that `state` refers to blockchain state. I suggested small rewording in my PR.

src/storage/postgres.rs
```diff
@@ -0,0 +1,533 @@
...
+//! ProtocolSystem enums.
+//!
+//! With this representation, it's important to synchronize them whenever the
+//! enums members changed. This can be done automatically once at system
+//! startup.
+//!
+//!
+//! Note: A removed enum can be ignored safely even though it might instigate a
+//! panic if an associated entity still exists in the database and retrieved
+//! with a codebase which no longer presents the enum value.
```
michcio1234, line 24: I didn't get that. Why can it be ignored safely then?

src/extractor/evm/ambient.rs
```diff
@@ -0,0 +1,98 @@
...
+    >,
+    inner: Arc<Mutex<Inner>>,
+}
+
+impl<DB> AmbientContractExtractor<DB> {
+    async fn update_cursor(&self, cursor: String) {
+        let cursor_bytes: Vec<u8> = cursor.into();
+        let mut state = self.inner.lock().await;
+        state.cursor = cursor_bytes;
+    }
```
michcio1234, line 41: Is the lock released automatically when exiting the scope?

src/storage/mod.rs
kayibal: Because types that you retrieve (S) and the ones that you use to insert (N) are usually different with diesel.

src/storage/mod.rs
kayibal: Number is ambiguous so I indeed want the hash here which should be unique for all blockchain protocols.

src/storage/mod.rs
kayibal: I merged you PR

src/storage/postgres.rs
kayibal: I mean removed from the DB. If you remove an enum from the DB and rust still know how to deal with it, it should basically never load/use it. That is safe.

src/extractor/evm/ambient.rs
kayibal: Yes, because the type of state is a lock object, that dereferences to the inner type. So once we exit the scope the lock is automatically released.

src/storage/postgres.rs
michcio1234: Huh, sorry, but I don't follow. You remove an enum from a DB, but something in the DB still refers to this enum - is this the situation we're talking about? I guess it won't happen in relational DBs, and IIUC we're limiting ourselves to relational DBs by using Diesel, no? So we should be safe.

src/storage/mod.rs
michcio1234: Huh, kinda weird, no? Intuitively I would expect that if I save an object of some class to the database, I will retrieve an object of the same class.

kayibal: Not in strictly typed languages see the example for Block and NewBlock. E.g NewBlock as no primary key id member because when inserting you don't know it yet. And if you would set it to None it is ambiguous if you want the column Val to be Null or if you want to ignore it.

src/storage/postgres.rs
kayibal: The enum is an entry in the table. You can only remove it if all data associated with it is gone. If you remove it from db and not from rust we are safe. Diesel means only Postgres mostly there is some SQLite and MySQL support but it was built for Postgres.

src/storage/mod.rs
michcio1234: Hmm, okay. I am curious how it will be in practice - don't want to get lost in all the types. :)

src/storage/postgres.rs
michcio1234: Okay, then Postgres should not allow us to do anything wrong. The only situation where things can go wrong is removing something from the code but keeping it in the DB - then the program won't understand the data it receives from DB. But that's rather obvious.
All good.

PR #3, opened 2023-08-23
ExtractionState gateway

carloszanella:
WIP

### DIFF
--- a/src/storage/postgres/extraction_state_gateway.rs
+++ b/src/storage/postgres/extraction_state_gateway.rs
@@ -120,7 +120,6 @@ mod test {
             .unwrap();
         let extractor_name = "setup_extractor";
 
-        let gateway = get_dgw(&mut conn).await;
         let cursor = Some("10".as_bytes());
         let attributes = serde_json::json!({"test": "test"});
         let orm_state = orm::NewExtractionState {
@@ -138,7 +137,6 @@ mod test {
             .await
             .unwrap();
 
-        // gateway.save_state(&state, &mut conn).await.unwrap();
         conn
     }
 


src/storage/orm.rs
```diff
@@ -21,16 +20,93 @@ pub struct Chain {
...
+    pub version: &'a str,
+    pub chain_id: i64,
+    pub cursor: Option<&'a [u8]>,
+    pub attributes: Option<&'a serde_json::Value>,
+    pub modified_ts: NaiveDateTime,
+}
+
+#[derive(AsChangeset)]
+#[diesel(table_name = extraction_state)]
+pub struct ExtractionStateForm<'a> {
```
kayibal, line 107: Form?

src/storage/postgres/extraction_state_gateway.rs
```diff
@@ -0,0 +1,234 @@
...
+            Ok(Some(orm_state)) => {
+                let state = ExtractionState {
+                    name: orm_state.name,
+                    chain,
+                    attributes: orm_state.attributes.into(),
+                    cursor: orm_state.cursor.unwrap_or_default(),
+                };
+                Ok(Some(state))
+            }
+            Ok(None) => Ok(None), // No matching entry in the DB
```
kayibal: Hmm do you prefer to handle these via None? I handled this case via a special error type when implementing the ChainGateway. 

I slightly prefer returning `Err(StorageError::NotFound)` because it is more self explaining.

src/storage/orm.rs
```diff
@@ -21,16 +20,93 @@ pub struct Chain {
...
+    /// # Parameters
+    /// - `extractor`: The name of the extractor to filter by.
+    /// - `chain_id`: The ID of the chain to filter by.
+    /// - `conn`: A mutable reference to an asynchronous PostgreSQL connection.
+    ///
+    /// # Returns
+    /// - `Ok(Some(ExtractionState))` if a matching `ExtractionState` is found.
+    /// - `Ok(None)` if no matching entry is found in the database.
+    /// - `Err(DieselError)` if a Diesel error occurs during the query.
+    pub async fn get_for_extractor(
```
kayibal: I'd prefer if we can agree on a naming scheme here, I think for the `orm::Block` I used something like `by_number`, `by_hash`. Just to stay consistent I would suggest to name this one `by_name`.

src/storage/postgres/extraction_state_gateway.rs
```diff
@@ -0,0 +1,234 @@
...
+        }
+    }
+
+    async fn save_state(
+        &self,
+        state: &ExtractionState,
+        conn: &mut Self::DB,
+    ) -> Result<(), StorageError> {
+        let block_chain_id = self.get_chain_id(state.chain);
+        match orm::ExtractionState::get_for_extractor(&state.name, block_chain_id, conn).await {
```
kayibal: Nice, I wonder though if we can achieve the upsert with a single query... given that this will run for every block, it might be worth the optimisation. I think these kind of queries are possible with postgres:

https://docs.diesel.rs/master/diesel/upsert/struct.IncompleteOnConflict.html#set-specific-value-on-conflict

I probably should do the same for chain gateway this way we simply don't care if the blockwas already inserted, if it already was, we can simply ignore and proceed.

src/storage/postgres/extraction_state_gateway.rs
```diff
@@ -0,0 +1,234 @@
...
+
+        let gateway = get_dgw(&mut conn).await;
+        let state = ExtractionState {
+            name: extractor_name.to_owned(),
+            chain: Chain::Ethereum,
+            attributes: serde_json::json!({"test": "test"}),
+            cursor: Vec::new(),
+        };
+
+        gateway.save_state(&state, &mut conn).await.unwrap();
```
kayibal: Not a big fan of using the thing you are testing (gateway) itself during setup but ok.

src/storage/postgres/extraction_state_gateway.rs
```diff
@@ -0,0 +1,234 @@
...
+        conn
+    }
+
+    async fn get_dgw(
+        conn: &mut AsyncPgConnection,
+    ) -> PostgresGateway<evm::Block, evm::Transaction> {
+        PostgresGateway::<evm::Block, evm::Transaction>::from_connection(conn).await
```
kayibal, line 146: nit: Probably not worth wrapping this in a function.

src/storage/postgres/extraction_state_gateway.rs
```diff
@@ -0,0 +1,234 @@
...
+    async fn test_save_new_state() {
+        let mut conn = setup_db().await;
+        let extractor_name = "test_extractor";
+
+        let gateway = get_dgw(&mut conn).await;
+        let state = ExtractionState {
+            name: extractor_name.to_owned(),
+            chain: Chain::Ethereum,
+            attributes: serde_json::json!({"test": "test"}),
+            cursor: "10".as_bytes().to_vec(),
```
kayibal: nit: `"10".to_owned().into()` Sounds nicer to me (if it works)

src/storage/postgres/extraction_state_gateway.rs
```diff
@@ -0,0 +1,234 @@
...
+        };
+
+        // Save the state using the gateway
+        gateway.save_state(&state, &mut conn).await.unwrap();
+
+        let query_res: orm::ExtractionState = schema::extraction_state::table
+            .filter(schema::extraction_state::name.eq(extractor_name))
+            .limit(1)
+            .select(orm::ExtractionState::as_select())
+            .load(&mut conn)
```
kayibal: Did you mean to use `first(&mut conn)`?

kayibal: I see... you could also use the query dsl to insert in setup_db instead and then reuse the gateway here to retrieve your state. 
I'd prefer that style in database tests - because it is easier if e.g. during a refactoring your fixtures keep working.

src/storage/postgres/extraction_state_gateway.rs
```diff
@@ -0,0 +1,234 @@
...
+        gateway.save_state(&state, &mut conn).await.unwrap();
+        assert_eq!(
+            gateway
+                .get_state(extractor_name, Chain::Ethereum, &mut conn)
+                .await
+                .unwrap()
+                .unwrap()
+                .cursor,
+            "10".as_bytes().to_vec()
+        );
```
kayibal: You already have a state in the db from the setup, can't you update that one? WOuld keep the test a bit shorter.

src/storage/orm.rs
carloszanella: Diesel example. Its a form you fill to update a field on the DB

src/storage/postgres/extraction_state_gateway.rs
carloszanella: I find it quite verbose... its just a test util right

src/storage/postgres/extraction_state_gateway.rs
carloszanella: I used first as you suggested. Idk what you mean with the second message

src/storage/orm.rs
kayibal: Yes but it relates to a frontend facing application where a user fills out a form at least that is how I interpret it. Should we use a clearer name maybe?

src/storage/orm.rs
```diff
@@ -102,12 +102,12 @@ pub struct NewExtractionState<'a> {
...
 }
 
-#[derive(AsChangeset)]
+#[derive(AsChangeset, Debug)]
 #[diesel(table_name = extraction_state)]
 pub struct ExtractionStateForm<'a> {
     pub cursor: Option<&'a [u8]>,
     pub attributes: Option<&'a serde_json::Value>,
-    pub modified_ts: NaiveDateTime,
+    pub modified_ts: Option<NaiveDateTime>,
```
kayibal, line 110: This is set automatically by the DB, let's not support setting this from the application layer. If you need that in a test, maybe you can somehow scope it to be part of test compile only?

PR #6, opened 2023-08-28
WIP ContractState gateway

carloszanella:


src/storage/schema.rs
kayibal, line 1: I think this file should be gone, please adjust the location under `/diesel.toml`, else this will be autogenerated into the wrong directory..

PR #7, opened 2023-08-28
Implement getting storage slots deltas.

kayibal:
We'll use this when we get a revert.

### DIFF
--- a/src/storage/postgres/mod.rs
+++ b/src/storage/postgres/mod.rs
@@ -23,6 +23,56 @@
 //! panic if an associated entity still exists in the database and retrieved
 //! with a codebase which no longer presents the enum value.
 //!
+//!
+//! ### Versioning
+//!
+//! This implementation utilizes temporal tables for recording the changes in
+//! entities over time. In this model, `valid_from` and `valid_to` determine the
+//! timeframe during which the facts provided by the record are regarded as
+//! accurate (validity period). Typically, in temporal tables, a valid version
+//! for a specific timestamp is found using the following predicate:
+//!
+//! ```sql
+//! valid_from < version_ts AND (version_ts <= valid_to OR valid_to is NULL)
+//! ```
+//!
+//! The `valid_to` can be set to null, signifying that the version remains
+//! valid. However, as all alterations within a block happen simultaneously,
+//! this predicate might yield multiple valid versions for a single entity.
+//!
+//! To further assign a temporal sequence to these entities, the transaction
+//! index within the block is recorded, usually through a `modify_tx` foreign
+//! key.
+//!
+//! ```sql
+//! SELECT * FROM table
+//! JOIN transaction
+//! WHERE valid_from < version_ts
+//!     AND (version_ts <= valid_to OR valid_to is NULL)
+//! ORDER BY entity_id, transaction.index DESC
+//! DISTINCT ON entity_id
+//! ```
+//!
+//! Here we select a set of versions by timestamp, then arrange rows by their
+//! transaction index (descending) and choose the first row, thus obtaining the
+//! latest version within the block (aka version at end of block).
+//!
+//! #### Contract Storage Table
+//!
+//! Special attention must be given to the contract_storage table, which also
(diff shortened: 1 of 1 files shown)


migrations/2023-08-15-154244_tycho_initial_setup/up.sql
```diff
@@ -387,15 +387,23 @@ create table if not exists contract_storage (
...
 	-- the contract this entry refers to.
 	"account_id" bigint references account(id) not null,
 	
 	-- the transaction that modified the slot to this entry.
 	"modify_tx" bigint references "transaction"(id) not null,
+
+	-- this is redundant with transaction.index, but included 
+	--	for performance reasons. Orders entries that are valid 
+	--	within the same version.
+	"ordinal" bigint not null,
```
carloszanella, line 406: isn't this missing on ORM?

src/extractor/evm/mod.rs
```diff
@@ -84,4 +84,4 @@ impl NormalisedMessage for AccountUpdate {
 }
 
 pub type EVMStateGateway<DB> =
-    StateGatewayType<DB, Block, Transaction, ERC20Token, SwapPool, Account, U256, U256>;
+    StateGatewayType<DB, Block, Transaction, ERC20Token, SwapPool, Account, H160, U256, U256>;
```
carloszanella, line 87: why was this needed?

src/storage/mod.rs
```diff
@@ -538,10 +548,11 @@ pub trait ContractStateGateway {
     ///     can't be located in storage.
     async fn get_slots_delta(
         &self,
-        id: ContractId,
+        id: Chain,
```
carloszanella, line 551: why did you change this to be per chain instead of address?

src/storage/postgres/contract_state.rs
```diff
@@ -70,14 +74,153 @@ where
 
     async fn get_slots_delta(
```
carloszanella, line 75: should we already start adding docs to the public methods? or do you get documentation from the trait?

src/storage/postgres/contract_state.rs
```diff
@@ -70,14 +74,153 @@ where
...
-            "id".to_owned(),
-        ))
+        target_version: Option<BlockOrTimestamp>,
+        conn: &mut AsyncPgConnection,
+    ) -> Result<HashMap<H160, HashMap<U256, U256>>, StorageError> {
+        let chain_id = self.get_chain_id(chain);
+        // To support blocks as versions, we need to ingest all blocks, else the
+        // below method can error for any blocks that are not present.
+        let start_version_ts = version_to_ts(&start_version, conn).await?;
+        let target_version_ts = version_to_ts(&target_version, conn).await?;
```
carloszanella, line 86: This versioning is separate from the one of the extractors? Asking because there I think you defined server?

src/extractor/evm/mod.rs
kayibal: Because slots_deltas are returned for the whole chain now, the return type still needs to group them by account and I need something that serves as an account id here.

src/storage/mod.rs
kayibal: Because it is more sensible. There is no point of getting the delta changes of a single contract. Usually the chain reverts, not a single contract.

src/storage/postgres/contract_state.rs
kayibal: I think ideally things should be documented on the trait.

src/storage/postgres/contract_state.rs
kayibal: I think I wrote a big documentation about versioning here. The extractors `version` attribute does not achieve table versioning btw, it just records which version of the extractor substreams package we want to run.

migrations/2023-08-15-154244_tycho_initial_setup/up.sql
kayibal: I'll add this in my next PR.

PR #8, opened 2023-08-29
Implement reverting a chain to prev. block.

kayibal:
Reverts a complete chain to a previous block. 

We do this leveraging CASCADE DELETE rule on foreign keys.

### DIFF
--- a/migrations/2023-08-15-154244_tycho_initial_setup/up.sql
+++ b/migrations/2023-08-15-154244_tycho_initial_setup/up.sql
@@ -1,476 +1,386 @@
 -- Your SQL goes here
 CREATE EXTENSION IF NOT EXISTS hstore;
+
 CREATE EXTENSION IF NOT EXISTS pg_trgm;
-CREATE TYPE financial_protocol_type AS ENUM ('swap', 'psm', 'debt', 'leverage');
-CREATE TYPE protocol_implementation_type AS ENUM ('custom', 'vm');
+
+CREATE TYPE financial_protocol_type AS ENUM(
+    'swap',
+    'psm',
+    'debt',
+    'leverage'
+);
+
+CREATE TYPE protocol_implementation_type AS ENUM(
+    'custom',
+    'vm'
+);
 
 -- Enumeration of supported blockchains
-create table if not exists "chain" (
-	"id" BIGSERIAL PRIMARY KEY,
-	
-	-- The name of the blockchain
-	"name" varchar(255) unique not null,
-	
-	-- Timestamp this entry was inserted into this table.
-	"inserted_ts" timestamptz not null default current_timestamp,
-	
-	-- Timestamp this entry was inserted into this table.
-	"modified_ts" timestamptz not null default current_timestamp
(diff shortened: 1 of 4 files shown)


migrations/2023-08-15-154244_tycho_initial_setup/up.sql
```diff
@@ -176,6 +176,11 @@ create table if not exists protocol_component (
     --	it could be inferred from related contracts but it might not be clear
     --	in case the protocol relates or more than one contract.
     "created_at" timestamptz not null,
+
+	-- This is reference makes it easy to revert to a previous state by 
+	--	simply deleting the corresponding blocks. Once again somewhat 
+	--	redundant but makes bookkeeping much easier.
+	"creation_tx" bigint references "transaction"(id) on delete cascade not null,
```
carloszanella: indentation looks weird

src/storage/postgres/contract_state.rs
```diff
@@ -225,9 +225,65 @@ where
...
+        .await?;
+
+        // Any versioned table's rows, which have `valid_to` set to "> block.ts"
+        // need, to be updated to be valid again (thus, valid_to = NULL).
+        diesel::update(
+            schema::contract_storage::table.filter(schema::contract_storage::valid_to.gt(block.ts)),
+        )
+        .set(schema::contract_storage::valid_to.eq(Option::<NaiveDateTime>::None))
+        .execute(conn)
+        .await?;
```
carloszanella, line 256: don't we need to delete records that have `valid_from > block.ts`? or does this get handled automatically with the cascade somehow? is this case covered in tests?

src/storage/postgres/contract_state.rs
```diff
@@ -457,4 +489,130 @@ mod test {
...
+            conn,
+            c1,
+            txn[3],
+            "2020-01-01T01:00:00",
+            &[(0, 128), (1, 256)],
+        )
+        .await;
+    }
+
+    #[tokio::test]
```
carloszanella, line 568: this test module is so big with these setup functions.. is there no way around? my latest PR also adds more to it

src/storage/postgres/contract_state.rs
carloszanella: also don't we need to invalidate extraction state on reverts? e.g. its cursor

kayibal: 1. Yes this should be handled by the deletion of blocks, because usually this should hold: `valid_from == tx.block.ts`. I assert that certain contract_storage entries are deleted in the test.
2. No, the stream will tell us to revert, and also include an updated cursor that says "this client already executed that revert and it's db is therefore reverted". So in case we restart at that moment and then reconnect the stream will not push that revert again to us.

src/storage/postgres/contract_state.rs
kayibal: We could have one module per method - or separate into files. But I usually collapse things so I haven't been bothered too much by it atm.

PR #10, opened 2023-08-29
Gateway's Get Contract method

carloszanella:
I will add some more unit tests. I am unsure about creation/modify TXs for the Account struct. Where do I get the last modified TX from?


src/storage/postgres/contract_state.rs
```diff
@@ -24,20 +24,79 @@ where
         + 'static,
 {
     type DB = AsyncPgConnection;
-    type ContractState = AccountUpdate;
+    type ContractState = Account;
```
kayibal, line 27: Here you are basically saying that all gateway instances will return `evm::Account` objects. I think this type is going to be yet another generic argument. But we can do that later let's keep things simple for now.

We have the same issue below with Slot, Value and Address..

src/storage/postgres/contract_state.rs
```diff
@@ -24,20 +24,79 @@ where
...
-        Err(StorageError::NotFound(
-            "ContractState".to_owned(),
-            "id".to_owned(),
-        ))
+        let h160_address = H160::from_slice(&id.1);
+        let account_orm = orm::Account::by_id(id, db).await;
+        let version_ts = version_to_ts(&version, db).await?;
+
+        match account_orm {
+            Ok(account_orm) => {
```
kayibal, line 43: I would slightly prefer this, to avoid the indentation and immediately make clear what happen in an error case:
```
account_orm: orm::Account = orm::Account::by_id(id, db)
    .await
    .map_err(|err| StorageError::from_diesel(
                err,
                "Account",
                &h160_address.to_string(),
                None,
            ))?;
let balance_query = schema::account_balance::table
    ...
```

carloszanella: I will implement this change on my new PR just so I can merge this one

PR #11, opened 2023-08-30
Insert accounts

carloszanella:
Some notes:
- I did some interface changes that felt required. you can see them on storage/mod.rs
- I will add some more assertions and tests in the next commits
- This is just an insert, I can try doing an upsert next.. not sure it will be needed?

### DIFF
--- a/src/storage/postgres/contract_state.rs
+++ b/src/storage/postgres/contract_state.rs
@@ -113,9 +113,9 @@ where
                     .map_err(|err| {
                         StorageError::from_diesel(err, "Transaction", &tx.to_string(), None)
                     })?;
-                (Some(tx_id), Some(_created_ts))
+                (Some(tx_id), _created_ts)
             }
-            None => (None, Some(now)),
+            None => (None, now),
         };
 
         let query = diesel::insert_into(schema::account::table).values((
@@ -140,7 +140,7 @@ where
         let orm_balance = orm::NewAccountBalance {
             account_id: acc_id,
             balance: balance_bytes.to_vec(),
-            valid_from: created_ts.unwrap(),
+            valid_from: created_ts,
             modify_tx: creation_tx_id,
             valid_to: None,
         };
@@ -162,7 +162,7 @@ where
                 schema::contract_code::hash.eq(code_hash),
                 schema::contract_code::account_id.eq(acc_id),
                 schema::contract_code::modify_tx.eq(tx_id),
-                schema::contract_code::valid_from.eq(created_ts.unwrap()),
+                schema::contract_code::valid_from.eq(created_ts),
             );
 
             diesel::insert_into(schema::contract_code::table)
@@ -578,14 +578,15 @@ mod test {
         assert_eq!(expected, actual);
 
         let orm_account = orm::Account::by_id(&contract_id, &mut conn).await.unwrap();
-        let block_ts = schema::transaction::table
+        let (block_ts, tx_ts) = schema::transaction::table
             .inner_join(schema::block::table)
             .filter(schema::transaction::id.eq(txn[1]))
-            .select(schema::block::ts)
-            .first::<NaiveDateTime>(&mut conn)
+            .select((schema::block::ts, schema::transaction::inserted_ts))
+            .first::<(NaiveDateTime, NaiveDateTime)>(&mut conn)
             .await
             .unwrap();
         assert_eq!(block_ts, orm_account.created_at.unwrap());
+        println!("tx_ts: {:?}, block_ts: {:?}", tx_ts, block_ts);
     }
 
     #[tokio::test]
--- a/src/storage/postgres/mod.rs
+++ b/src/storage/postgres/mod.rs
@@ -332,30 +332,22 @@ mod fixtures {
     pub async fn insert_txns(conn: &mut AsyncPgConnection, txns: &[(i64, i64, &str)]) -> Vec<i64> {
         let from_val = H160::from_str("0x4648451b5F87FF8F0F7D622bD40574bb97E25980").unwrap();
         let to_val = H160::from_str("0x6B175474E89094C44Da98b954EedeAC495271d0F").unwrap();
-        let mut data = Vec::new();
-
-        for &(b, i, h) in txns {
-            use schema::transaction::dsl::*;
-
-            let block = schema::block::table
-                .find(b)
-                .select(orm::Block::as_select())
-                .first(conn)
-                .await
-                .unwrap();
-
-            data.push((
-                block_id.eq(b),
-                index.eq(i),
-                hash.eq(H256::from_str(h)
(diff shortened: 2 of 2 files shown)


src/storage/postgres/contract_state.rs
```diff
@@ -91,8 +91,97 @@ where
...
+                    .first::<(i64, NaiveDateTime)>(db)
+                    .await
+                    .map_err(|err| {
+                        StorageError::from_diesel(err, "Transaction", &tx_hash.to_string(), None)
+                    })?;
+                created_ts = Some(_created_ts);
+                Some(tx_id)
+            }
+            None => None,
+        };
```
carloszanella: this feels a bit hacky but Idk how to like get `tx_hash` and `created_ts` only if there is a creation transaction

src/storage/postgres/contract_state.rs
```diff
@@ -91,8 +91,97 @@ where
...
+                .map_err(|err| {
+                    StorageError::from_diesel(err, "ContractCode", &new.address.to_string(), None)
+                })?;
+
+            Self::upsert_slots(
+                self,
+                ContractId(new.chain, new.address.as_bytes().to_vec()),
+                tx_hash.as_bytes(),
+                &new.slots,
+            )
```
carloszanella, line 181: this method right now is not implemented

src/storage/mod.rs
```diff
@@ -480,7 +480,11 @@ pub trait ContractStateGateway {
     ///
     /// # Parameters
     /// - `new` the new contract state to be saved.
-    async fn add_contract(&mut self, new: Self::ContractState) -> Result<(), StorageError>;
+    async fn add_contract(
+        &self,
+        new: &Self::ContractState,
+        db: &mut Self::DB,
+    ) -> Result<i64, StorageError>;
```
carloszanella, line 487: Idk why self was `mut` here and I made `&new` a reference. Not sure if you had a special reason for it to be moved

src/storage/postgres/contract_state.rs
```diff
@@ -30,77 +30,158 @@ where
...
+        let now = chrono::Utc::now().naive_utc();
+        let chain_id = self.get_chain_id(new.chain);
+        let mut tx_hash = H256::zero();
+        let mut created_ts = None;
+        let creation_tx_id = match new.creation_tx {
+            Some(tx) => {
+                tx_hash = tx;
+                let (tx_id, _created_ts) = schema::transaction::table
+                    .filter(schema::transaction::hash.eq(tx_hash.as_bytes()))
+                    .select((schema::transaction::id, schema::transaction::inserted_ts))
```
kayibal: This should be the `block.ts`, not the time at which we inserted the transaction. We have a few options:
1. We retrieve the block through txhash like you tried here
2. We save the block or the timestamp as part of the model
3. We pass created_at: Block through the interface, but for upserts that would be strange

src/storage/postgres/contract_state.rs
```diff
@@ -30,77 +30,158 @@ where
...
+                    .first::<(i64, NaiveDateTime)>(db)
+                    .await
+                    .map_err(|err| {
+                        StorageError::from_diesel(err, "Transaction", &tx_hash.to_string(), None)
+                    })?;
+                created_ts = Some(_created_ts);
+                Some(tx_id)
```
kayibal: This is a bit weird, probably better to return a tuple. But indeed a weird case..

src/storage/postgres/contract_state.rs
```diff
@@ -30,77 +30,158 @@ where
...
+            })?;
+
+        // Insert initial account balances to the respective table.
+        let mut balance_bytes = [0; 32];
+        new.balance.to_big_endian(&mut balance_bytes);
+
+        let orm_balance = orm::NewAccountBalance {
+            account_id: acc_id,
+            balance: balance_bytes.to_vec(),
+            valid_from: now,
```
kayibal: This needs to correspond to the `block.ts` of the insertion transaction. Remember we have to insert historical data so it can't be just `now` here - we might be inserting a contract that was created years ago.

src/storage/postgres/contract_state.rs
kayibal: Maybe this?
```rust
let (creation_tx_id, created_ts): (Option<i64>, Option<NaiveDateTime>) = 
    match new.creation_tx {
        Some(tx) => {
            let (tx_id, _created_ts) = schema::transaction::table
                    .filter(schema::transaction::hash.eq(tx.as_bytes()))
                    .select((schema::transaction::id, schema::transaction::inserted_ts))
                    .first::<(i64, NaiveDateTime)>(db)
                    .await
                    .map_err(|err| {
                        StorageError::from_diesel(err, "Transaction", &tx.to_string(), None)
                    })?;
            (Some(tx_id), Some(_created_ts))
        }
        None => (None, None),
    };
```

src/storage/postgres/contract_state.rs
kayibal: We will get all data block & tx scoped from streamingfast - so it will look sth like this:
```
{
    "block": {"number": ..., "hash": ...},
    "transactions": [{
         "hash": "0x...",
         "created_contracts": [{...}, ..],
         "code_changes": [..],
         "balance_changes": [..],
         "storage_changes": [..],
    }],
}
```

src/storage/postgres/mod.rs
```diff
@@ -332,22 +332,30 @@ mod fixtures {
...
+            data.push((
+                block_id.eq(b),
+                index.eq(i),
+                hash.eq(H256::from_str(h)
+                    .expect("valid txhash")
+                    .as_bytes()
+                    .to_owned()),
+                from.eq(from_val.as_bytes()),
+                to.eq(to_val.as_bytes()),
+                inserted_ts.eq(block.ts),
```
kayibal: Is there any logic on this field? Or why did you add it? The transaction timestamp is the same as the block assigned to it. Both `inserted_ts` and `modified_ts` are maintained by Postgres triggers, so no need to set it them.

src/storage/postgres/contract_state.rs
```diff
@@ -98,23 +98,24 @@ where
...
                     .map_err(|err| {
-                        StorageError::from_diesel(err, "Transaction", &tx_hash.to_string(), None)
+                        StorageError::from_diesel(err, "Transaction", &tx.to_string(), None)
                     })?;
-                created_ts = Some(_created_ts);
-                Some(tx_id)
+                (Some(tx_id), Some(_created_ts))
             }
-            None => None,
+            None => (None, Some(now)),
```
kayibal: Does it make sense to wrap created_ts with Some() it seems you unwrap it most of the times below, doing it the other way around would probably be nicer. Retrieve without Option and wrap in Some if needed.

PR #12, opened 2023-08-31
Upsert slots

kayibal:
Implements upsert slots for Postgres.

Unfortunately I decided to change the interface again to avoid multiple round trips to/from the db. The current interface takes a more complex type, but therefore is more flexible. E.g. we could even batch all storage changes into a single call.

I also simplified the StorableTransaction trait, by simply assuming that we can convert simple types into bytes and that this is sensible for a storable representation.

### DIFF
--- a/src/storage/postgres/contract_state.rs
+++ b/src/storage/postgres/contract_state.rs
@@ -97,7 +97,6 @@ where
     ) -> Result<i64, StorageError> {
         let now = chrono::Utc::now().naive_utc();
         let chain_id = self.get_chain_id(new.chain);
-        let tx_hash = H256::zero();
         let (creation_tx_id, created_ts) = match new.creation_tx {
             // If there is a transaction hash assigned to the Account, we load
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,65 @@
+name: Continuous Integration
```
zizou0x, line 1: Nice

src/extractor/evm/mod.rs
```diff
@@ -22,7 +22,7 @@ pub struct Block {
     pub ts: NaiveDateTime,
 }
 
-#[derive(Debug, PartialEq, Copy, Clone)]
+#[derive(Debug, PartialEq, Copy, Clone, Default)]
```
zizou0x, line 25: FMI why do we need it here?

kayibal: Because I am a lazy man.

PR #13, opened 2023-09-01
ENG-1745: Implement get state request parsing

zizou0x:
[Task](https://datarevenue.atlassian.net/jira/software/projects/ENG/boards/24?selectedIssue=ENG-1745)

### DIFF
--- a/src/rpc/mod.rs
+++ b/src/rpc/mod.rs
@@ -1,7 +1,18 @@
 //! This module contains Tycho RPC implementation
 
+use serde::de::{self, Deserializer};
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
+
+// This will convert a hex string (with or without 0x) to a Vec<u8>
+fn hex_to_bytes<'de, D>(deserializer: D) -> Result<Vec<u8>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let s: String = Deserialize::deserialize(deserializer)?;
+    let hex_str = s.strip_prefix("0x").unwrap_or(&s);
+    hex::decode(hex_str).map_err(de::Error::custom)
+}
 
 #[derive(Error, Debug)]
 pub enum RpcError {
@@ -24,7 +35,8 @@ struct StateRequestBody {
 
 #[derive(Debug, Serialize, Deserialize, PartialEq)]
 struct ContractId {
-    address: String,
+    #[serde(deserialize_with = "hex_to_bytes")]
+    address: Vec<u8>,
     chain: String,
 }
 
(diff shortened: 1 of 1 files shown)


src/rpc/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+                    "address": "0xb4eccE46b8D4e4abFd03C9B806276A6735C9c092",
+                    "chain": "ethereum"
+                }
+            ],
+            "version": {
+                "timestamp": "2069-01-01T04:20:00",
+                "block": {
+                    "hash": "0xd76628379905b342fe3f40a4aa2ef60747fb61e3f10e1c0052313aafc0a73566",
+                    "number": 213,
+                    "chain": "ethereum"
```
kayibal, line 85: I think it would be nice to pass the parent hash of the block as well. The client can use that to understand what is going on, It can also be used to detect gaps, e.g. if we had a reconnect we might have missed some data or not...

src/rpc/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_parse_state_request() {
+        let json_str = r#"
+        {
+            "contract_ids": [
```
kayibal: Although I am not a big fan of camel case, I think we should stick to it when parsing or emitting json payloads. So it should be:  `"contractIds"` as it would be expected by most api users.

src/rpc/mod.rs
```diff
@@ -0,0 +1,94 @@
+//! This module contains Tycho RPC implementation
+
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum RpcError {
+    #[error("Failed to parse JSON: {0}")]
+    SerdeJsonError(serde_json::Error),
```
kayibal: I would call this simply ParseError

src/rpc/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+                assert_eq!(body.contract_ids[0].chain, "ethereum");
+                assert_eq!(
+                    body.version.block.hash,
+                    "0xd76628379905b342fe3f40a4aa2ef60747fb61e3f10e1c0052313aafc0a73566"
+                );
+                assert_eq!(body.version.block.chain, "ethereum");
+                assert_eq!(body.version.block.number, 213);
+            }
+            Err(err) => panic!("Parsing failed: {}", err),
+        }
```
kayibal: Nit: I find it generally more readable to specify the expected structs here. In this case it would look something like:
```
let c0 = hex::decode("b4eccE46b8D4e4abFd03C9B806276A6735C9c092").expect("address ok");
let blk_hash = ...;
let tx_hash = ....;
let exp = StateRequestBody{
    contract_ids: vec![ContractId(Chain::Ethereum, c0)]
    version: Version(...)
}
```

kayibal: Also no need to manually panic on error with such a specific message you could simply do:

```
parse_state_request(json_str).unwrap();
```

This will panic and will be clearer cause the traceback will point to the excact line where the panic happened.

zizou0x: I didn't use `hex::decode("b4eccE46b8D4e4abFd03C9B806276A6735C9c092")` because it makes it a bit more complex to parse and use. I can do it if we decide that we want to use `Vec<u8>` for hex.

zizou0x: I changed this to use `Vec<u8>` for memory optimisation.

PR #14, opened 2023-09-03
Get slots

kayibal:


src/storage/postgres/contract_state.rs
```diff
@@ -204,18 +247,18 @@ where
...
         // To support blocks as versions, we need to ingest all blocks, else the
         // below method can error for any blocks that are not present.
         let start_version_ts = version_to_ts(&start_version, conn).await?;
-        let target_version_ts = version_to_ts(&target_version, conn).await?;
+        let target_version_ts = version_to_ts(&Some(target_version), conn).await?;
 
         let changed_values = if start_version_ts <= target_version_ts {
             // Going forward
-            //                  ]     relevant changes     ]
+            //                  ]     changes to revert    ]
```
zizou0x, line 334: I don't understand why "to revert" in this case

src/storage/postgres/contract_state.rs
```diff
@@ -105,20 +105,63 @@ where
...
+                    "Unsupported version kind: {:?}",
+                    kind
+                )));
+            }
+            version_to_ts(&Some(version), conn).await?
+        } else {
+            Utc::now().naive_utc()
+        };
+
+        dbg!(&version_ts);
```
zizou0x, line 206: This line needs to be removed no?

src/storage/postgres/contract_state.rs
```diff
@@ -416,6 +415,85 @@ where
     }
 }
 
+/// Parses an evm address hash from the db
+///
+/// The db id is required to provide additional error context in case the
+/// parsing fails.
+fn parse_id_h160(db_id: &i64, v: &[u8]) -> Result<(i64, H160), StorageError> {
```
zizou0x, line 495: Do you think we should make a special type for addresses or it's fine to keep it generic? Can we enforce the length in the type hint like `v: &[u8; 20]` or it makes things too complicated?

src/storage/postgres/contract_state.rs
```diff
@@ -416,6 +415,85 @@ where
...
+/// Given an iterator of slot data construct an evm specific hash map
+///
+/// Will contruct a hashmap representing contract storage of several contracts.
+/// It does so by combining a query result and a mapping between database id and
+/// account addresses.
+fn construct_contract_storage(
+    slot_values: impl Iterator<Item = (i64, Vec<u8>, Option<Vec<u8>>)>,
+    addresses: &HashMap<i64, H160>,
+) -> Result<HashMap<H160, HashMap<U256, U256>>, StorageError> {
+    let mut result: HashMap<H160, HashMap<U256, U256>> = HashMap::with_capacity(addresses.len());
```
zizou0x, line 515: This is a bit hard to read 😅. Can we use a `where` and/or type aliases to make this more readable?

src/storage/postgres/contract_state.rs
```diff
@@ -416,6 +415,85 @@ where
...
+    Ok((*db_id, H160::from_slice(v)))
+}
+
+/// Given an iterator of slot data construct an evm specific hash map
+///
+/// Will contruct a hashmap representing contract storage of several contracts.
+/// It does so by combining a query result and a mapping between database id and
+/// account addresses.
+fn construct_contract_storage(
+    slot_values: impl Iterator<Item = (i64, Vec<u8>, Option<Vec<u8>>)>,
```
zizou0x, line 512: What does `(i64, Vec<u8>, Option<Vec<u8>>)` represent?

src/storage/postgres/contract_state.rs
```diff
@@ -416,6 +415,85 @@ where
...
+    addresses: &HashMap<i64, H160>,
+) -> Result<HashMap<H160, HashMap<U256, U256>>, StorageError> {
+    let mut result: HashMap<H160, HashMap<U256, U256>> = HashMap::with_capacity(addresses.len());
+    for (cid, raw_key, raw_val) in slot_values.into_iter() {
```
zizou0x, line 516: Ah ok I understand now, is there a way to make it clearer directly from the interface?

src/storage/postgres/contract_state.rs
kayibal: Well revert is probably not the right term, better call it forward to the correct state. I'll rephrase this once again in another PR.

src/storage/postgres/contract_state.rs
kayibal: This is actually a method specific for ethereum, currently the types are "hardcoded" to ethereum, but I do not want to keep it like that for long. E.g. Starknet has 32 bytes addresses no?

src/storage/postgres/contract_state.rs
kayibal: Yes I will try to completely encapsulate that into Account type and if not I will add something like account storage.

src/storage/postgres/contract_state.rs
kayibal: Yes I still hope to make this a bit clearer by reworking the types.

src/storage/mod.rs
```diff
@@ -282,22 +284,34 @@ pub trait ExtractionStateGateway {
     ) -> Result<(), StorageError>;
 }
 
+/// Point in time as either block or timestamp. If a block is chosen it
```
tvinagre, line 287: I don't understand this

PR #15, opened 2023-09-04
Implement delete_contract

kayibal:


src/storage/mod.rs
```diff
@@ -520,7 +520,7 @@ pub trait ContractStateGateway {
     async fn delete_contract(
         &self,
         id: ContractId,
-        at_tx: Option<&Self::Transaction>,
+        at_tx: &Self::Transaction,
```
zizou0x, line 527: Please update the doc above : `None retrieves the latest state`

src/storage/postgres/orm.rs
```diff
@@ -268,18 +278,19 @@ pub struct Account {
...
     pub deleted_at: Option<NaiveDateTime>,
+    pub deletion_tx: Option<i64>,
     pub inserted_ts: NaiveDateTime,
     pub modified_ts: NaiveDateTime,
 }
 
 impl Account {
     pub async fn by_id(
-        account_id: ContractId,
+        account_id: &ContractId,
```
pistomat: FMI What is the reasoning behind this change? When do you use the value and when a reference? In contrast, the `ContractStateGateway::delete_contract(self, id: ContractId,...)` has a value and not a reference.

kayibal: I think unfortunately that is some inconsistency from working on this in parallel. I think unless we need to consume the value we should probably use a reference if possible.

PR #16, opened 2023-09-05
Improve error handling for Chain and use it in RPC

zizou0x:


src/models.rs
```diff
@@ -5,16 +5,15 @@ pub enum Chain {
...
-        if value == "ethereum" {
-            Chain::Ethereum
-        } else if value == "starknet" {
-            Chain::Starknet
-        } else if value == "zksync" {
-            Chain::ZkSync
-        } else {
-            panic!("Can't interpret {} as chain!", value);
+impl TryFrom<String> for Chain {
+    type Error = String;
```
kayibal: Should we really use String as Error?

src/storage/postgres/mod.rs
```diff
@@ -173,7 +174,7 @@ where
             map_enum: HashMap::new(),
         };
         for (id_, name_) in entries {
-            let val = E::from(name_.to_owned());
+            let val = E::try_from(name_.to_owned()).unwrap();
```
kayibal: Can we use expect here instead pls?

src/models.rs
zizou0x: I've created a custom error type for this in my last commit

PR #17, opened 2023-09-06
ENG 1869: Code formatting and general improvements

pistomat:


.github/workflows/ci.yaml
```diff
@@ -37,16 +37,14 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
-      - name: Setup Rust
-        uses: actions-rs/toolchain@v1
+      - name: Setup toolchain
+        uses: dtolnay/rust-toolchain@master
```
tvinagre: Can you explain what's the difference?

tvinagre: Also, can we explicit a version? `master` can change and alter the behaviour of the ci

src/main.rs
```diff
@@ -21,37 +25,44 @@ mod rpc;
...
-        println!("The environment variable SUBSTREAMS_API_TOKEN must be set also");
-        println!("and should contain a valid Substream API token.");
-        exit(1);
-    }
+    let args = Args::parse();
 
-    let endpoint_url = env::args().nth(1).unwrap();
-    let package_file = env::args().nth(2).unwrap();
-    let module_name = env::args().nth(3).unwrap();
+    let api_token = env::var("SUBSTREAMS_API_TOKEN").ok();
```
tvinagre: If we enable `env` from `clap` you can parse envvars as well I think

` clap = { version = "...", features = ["env"] } `

.github/workflows/ci.yaml
pistomat: `actions-rs/toolchain` is [no longer maintained](https://github.com/actions-rs/toolchain/issues/216), so switched to actively maintained (and slimmer) action that seems to be the standard now.

src/substreams/stream.rs
```diff
@@ -113,36 +117,31 @@ fn stream_blocks(
...
+                                if let Some(duration) = backoff.next() {
+                                    sleep(duration).await
+                                } else {
+                                    return Err(anyhow!("Backoff requested to stop retrying, quitting"))?;
                                 }
 
-                                println!("Received tonic error {:#}", status);
-                                encountered_error = true;
-                                break;
+                                continue 'retry_loop;
```
pistomat, line 132: I tried to refactor this function using loop labels to make it more legible and remove the `encountered_error` flag, I am not 100% sure I didn't break it.

src/substreams/stream.rs
```diff
@@ -113,36 +117,31 @@ fn stream_blocks(
...
+                                    return Err(anyhow::Error::new(status.clone()))?;
+                                }
+
+                                error!("Received tonic error {:#}", status);
+
+                                // If we reach this point, we must wait a bit before retrying
+                                if let Some(duration) = backoff.next() {
+                                    sleep(duration).await
+                                } else {
+                                    return Err(anyhow!("Backoff requested to stop retrying, quitting"))?;
```
pistomat, line 129: The `Err` was not returned before.

src/substreams/stream.rs
```diff
@@ -113,36 +117,31 @@ fn stream_blocks(
                                 // Unauthenticated errors are not retried, we forward the error back to the
                                 // stream consumer which handles it
                                 if status.code() == tonic::Code::Unauthenticated {
-                                    Err(anyhow::Error::new(status.clone()))?;
+                                    return Err(anyhow::Error::new(status.clone()))?;
```
pistomat, line 120: The `Err` was not returned before.

.github/workflows/ci.yaml
tvinagre: Awesome, thanks!

src/substreams/stream.rs
kayibal: Haven't reviewed yet, but this code is from a template, so we have no tests for it. A bit careful to change things in there. Would be ideal to run this against the substreams-ambient spkg. To make sure things still work.

src/storage/mod.rs
```diff
@@ -484,8 +478,7 @@ pub trait ContractStateGateway {
     ///
     /// # Parameters
     /// - `id` The identifier for the contract.
-    /// - `version` Version at which to retrieve state for. None retrieves the latest
-    ///   state.
+    /// - `version` Version at which to retrieve state for. None retrieves the latest state.
```
kayibal, line 481: Hm any idea why this diff?

rustfmt.toml
```diff
@@ -0,0 +1,9 @@
+reorder_imports = true
+imports_granularity = "Crate"
+use_small_heuristics = "Max"
+comment_width = 100
```
kayibal, line 4: Any string preference for 100? We usually use 88 or 89.

src/storage/postgres/contract_state.rs
```diff
@@ -123,11 +123,8 @@ where
             schema::account::created_at.eq(created_ts),
             schema::account::address.eq(new.address.as_bytes()),
         ));
-        let acc_id = query
-            .returning(schema::account::id)
-            .get_result::<i64>(db)
-            .await
-            .map_err(|err| {
+        let acc_id =
+            query.returning(schema::account::id).get_result::<i64>(db).await.map_err(|err| {
```
kayibal: To be honest, I find this way less readable compared to the old version. What do you think? Can we keep the old way of formatting for chains?

src/substreams/stream.rs
kayibal: Hm interesting buggy template repo, they probably appreciate the fix here as well: https://github.com/streamingfast/substreams-sink-rust

rustfmt.toml
pistomat: This is a copy paste from Reth, we can ofc use 88 or 89 if it is the consensus, but I generally use widescreen and prefer 100 or 120.

src/storage/mod.rs
pistomat: I guess because `comment_width = 100` so it unwraps the line, because it still fits.

src/storage/postgres/contract_state.rs
tvinagre: I agree with @kayibal I prefer the new lines

rustfmt.toml
pistomat: Also this is just the comment width, the default code width is also 100, so this makes code length match comment length.

src/storage/postgres/contract_state.rs
pistomat: We can set it to `chain_width=40` which would make chains with width over 40 unroll like this.

src/models.rs
```diff
@@ -16,7 +17,7 @@ impl TryFrom<String> for Chain {
     type Error = ChainError;
 
     fn try_from(value: String) -> Result<Self, Self::Error> {
-        match value.as_str() {
+        match value.to_lowercase().as_str() {
```
carloszanella, line 20: Logic changes, refactoring, and formatting changes usually better in separate commits, so it's easier for the reviewer

src/extractor/runner.rs
```diff
@@ -187,9 +186,6 @@ where
             control_rx: ctrl_rx,
         };
 
-        Ok(ExtractorHandle {
-            handle: runner.run(),
-            control_tx: ctrl_tx,
-        })
+        Ok(ExtractorHandle { handle: runner.run(), control_tx: ctrl_tx })
```
carloszanella, line 195: Are we using a new formatter now? Or different settings?

src/storage/postgres/contract_state.rs
carloszanella: I also prefer the previous format

src/models.rs
pistomat: Agree, I had too many commits though, squashed some and then tried to separate the changes into logical commits, but missed some.

src/extractor/runner.rs
pistomat: I added the `rustfmt.toml` settings and we are using nightly formatting now to increase the number of lints.

src/substreams/stream.rs
pistomat: I tried to run both the previous and the new version, found a bug, fixed it, now it works as expected.

src/substreams/stream.rs
pistomat: I created a PR there

PR #18, opened 2023-09-06
ENG-1746: Get contract state in the RPC

zizou0x:
This PR implements the logic for [this task](https://datarevenue.atlassian.net/jira/software/projects/ENG/boards/24?selectedIssue=ENG-1746) and some other improvements (cleaning, optional args handling and some interface refactoring)

### DIFF
--- a/src/rpc/mod.rs
+++ b/src/rpc/mod.rs
@@ -1,13 +1,15 @@
 //! This module contains Tycho RPC implementation
 
-use crate::extractor::evm::{self, Account};
-use crate::models::Chain;
-use crate::rpc::deserialization_helpers::{chain_from_str, hex_to_bytes};
-use crate::storage::postgres::PostgresGateway;
-use crate::storage::{BlockIdentifier, ContractStateGateway};
-use crate::storage::{BlockOrTimestamp, ContractId};
-use chrono::NaiveDateTime;
-use chrono::Utc;
+use crate::{
+    extractor::evm::{self, Account},
+    models::Chain,
+    rpc::deserialization_helpers::{chain_from_str, hex_to_bytes},
+    storage::{
+        postgres::PostgresGateway, BlockIdentifier, BlockOrTimestamp, ContractId,
+        ContractStateGateway,
+    },
+};
+use chrono::{NaiveDateTime, Utc};
 use diesel_async::AsyncPgConnection;
 use serde::Deserialize;
 use thiserror::Error;
@@ -80,10 +82,7 @@ struct Version {
 
 impl Default for Version {
     fn default() -> Self {
-        Version {
-            timestamp: Utc::now().naive_utc(),
-            block: None,
-        }
+        Version { timestamp: Utc::now().naive_utc(), block: None }
     }
 }
 
@@ -109,8 +108,7 @@ mod tests {
     use crate::storage::postgres::db_fixtures;
     use diesel_async::AsyncConnection;
     use ethers::types::{H160, H256, U256};
-    use std::collections::HashMap;
-    use std::str::FromStr;
+    use std::{collections::HashMap, str::FromStr};
 
     use super::*;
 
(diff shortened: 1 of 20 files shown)


src/rpc/mod.rs
```diff
@@ -1,27 +1,56 @@
...
 
-fn chain_from_str<'de, D>(deserializer: D) -> Result<Chain, D::Error>
-where
-    D: Deserializer<'de>,
-{
-    let s: String = Deserialize::deserialize(deserializer)?;
+impl RequestHandler {
+    async fn get_state(&mut self, request: &StateRequestBody) -> StateRequestResponse {
+        //TODO: handle when no contract is specified with filters
+        let at = match &request.version.block {
```
pistomat, line 27: The `at` attribute should not be an `Option` if you return `Some` in both branches. Rather wrap it in `Some(..)` later at line 37.

src/rpc/mod.rs
```diff
@@ -1,27 +1,56 @@
...
+            Some(b) => Some(BlockOrTimestamp::Block(BlockIdentifier::Hash(
+                b.hash.clone(),
+            ))),
+            None => Some(BlockOrTimestamp::Timestamp(request.version.timestamp)),
+        };
+
+        let mut accounts: Vec<Account> = Vec::new();
```
pistomat, line 32: FYI: I don't like declaring the `mut accounts` and then populating it in a for cycle very much, but given the `get_contract()` is async, you would have to convert it to a stream first, which would make it actually more complex, so it is not worth it.

If it wasn't async inside, you would just filter_map() and then collect to just get the accounts in one pass.

src/rpc/mod.rs
```diff
@@ -1,27 +1,56 @@
...
+        if let Some(contract_ids) = &request.contract_ids {
+            for contract_id in contract_ids {
+                match self
+                    .db_gw
+                    .get_contract(contract_id, &at, &mut self.db_connection)
+                    .await
+                {
+                    Ok(contract_state) => accounts.push(contract_state),
+                    Err(e) => {
+                        print!("Error while getting contract state {}", e)
```
pistomat: Replace with `tracing::error!()`.

src/rpc/mod.rs
kayibal: I think the best and probably simplest would be to add a gateway method that simply returns all the contracts requested.

src/rpc/mod.rs
```diff
@@ -111,22 +148,180 @@ mod tests {
...
+            db_fixtures::insert_account(conn, acc_address, "account0", chain_id, None).await;
+
+        db_fixtures::insert_account_balances(conn, tid[0], acc_id).await;
+        let contract_code = hex::decode("1234").unwrap();
+        db_fixtures::insert_contract_code(conn, acc_id, tid[0], contract_code).await;
+        acc_address.to_string()
+    }
+
+    #[tokio::test]
+    async fn test_get_state() {
```
kayibal, line 285: This is a integration test right? That is why you setup the database above? I wonder if we should do them or opt for a mocked gateway impl. For now I think we can keep this one.

src/rpc/mod.rs
```diff
@@ -1,27 +1,56 @@
...
+            Some(b) => Some(BlockOrTimestamp::Block(BlockIdentifier::Hash(
+                b.hash.clone(),
+            ))),
+            None => Some(BlockOrTimestamp::Timestamp(request.version.timestamp)),
+        };
+
+        let mut accounts: Vec<Account> = Vec::new();
 
-    Chain::try_from(s).map_err(de::Error::custom)
+        if let Some(contract_ids) = &request.contract_ids {
```
kayibal, line 34: Problem I see here is what if the user does not specify contract ids? In that case he will get an empty response but the behaviour should be to get all... I will add a method to retrieve multiple contracts at once soon then we can come back to this.

zizou0x: Yes this case is still to be handled (it's in my "TODO" above). We need to make a function to retrieve multiple contract at once with the filters

src/rpc/mod.rs
zizou0x: Yes this tests the whole logic from the request to the response

PR #19, opened 2023-09-07
Cleanup interfaces

kayibal:
Changes:
- Decouple PostgresGateway from evm models
- StorableContract trait was introduced for that purpose
- Interfaces require References to any types that are not copy: Version, BlockOrTimestamp, etc.
- Avoid some unnecessary clones when doing insertions

### DIFF
--- a/src/storage/postgres/contract_state.rs
+++ b/src/storage/postgres/contract_state.rs
@@ -40,7 +40,7 @@ impl StorableContract<orm::Contract, orm::NewContract, i64> for evm::Account {
             H256::from_slice(&val.code.hash),
             H256::from_slice(balance_modify_tx),
             H256::from_slice(code_modify_tx),
-            creation_tx.map(|b| H256::from_slice(b)),
+            creation_tx.map(H256::from_slice),
         )
     }
 
@@ -244,7 +244,7 @@ where
         at_tx: &Self::Transaction,
         conn: &mut AsyncPgConnection,
     ) -> Result<(), StorageError> {
-        let account = orm::Account::by_id(&id, conn)
+        let account = orm::Account::by_id(id, conn)
             .await
             .map_err(|err| StorageError::from_diesel(err, "Account", &id.to_string(), None))?;
         let tx = orm::Transaction::by_hash(at_tx.hash(), conn)
@@ -268,10 +268,10 @@ where
                     "Account {} was already deleted at {:?}!",
                     hex::encode(account.address),
                     account.deleted_at,
-                )));
+                )))
             }
             // Noop if called twice on deleted contract
-            return Ok(());
+            return Ok(())
         };
         diesel::update(schema::account::table.filter(schema::account::id.eq(account.id)))
             .set((schema::account::deletion_tx.eq(tx.id), schema::account::deleted_at.eq(block_ts)))
@@ -314,9 +314,9 @@ where
                 return Err(StorageError::Unsupported(format!(
                     "Unsupported version kind: {:?}",
                     kind
-                )));
+                )))
             }
-            version_to_ts(&Some(&version), conn).await?
+            version_to_ts(&Some(version), conn).await?
         } else {
             Utc::now().naive_utc()
         };
@@ -435,7 +435,7 @@ where
         // To support blocks as versions, we need to ingest all blocks, else the
         // below method can error for any blocks that are not present.
         let start_version_ts = version_to_ts(&start_version, conn).await?;
-        let target_version_ts = version_to_ts(&Some(&target_version), conn).await?;
+        let target_version_ts = version_to_ts(&Some(target_version), conn).await?;
 
         let changed_values = if start_version_ts <= target_version_ts {
             // Going forward
@@ -614,7 +614,7 @@ fn parse_id_h160(v: &[u8]) -> Result<H160, StorageError> {
         return Err(StorageError::DecodeError(format!(
             "Invalid contract address found: {}",
             hex::encode(v)
-        )));
+        )))
     }
     Ok(H160::from_slice(v))
 }
@@ -658,14 +658,14 @@ fn parse_u256_slot_entry(
         return Err(StorageError::DecodeError(format!(
             "Invalid byte length for U256 in slot key! Found: 0x{}",
             hex::encode(raw_key)
-        )));
+        )))
     }
     let v = if let Some(val) = raw_val {
         if val.len() != 32 {
             return Err(StorageError::DecodeError(format!(
                 "Invalid byte length for U256 in slot value! Found: 0x{}",
                 hex::encode(val)
-            )));
+            )))
         }
         U256::from_big_endian(val)
     } else {
--- a/src/services/rpc.rs
+++ b/src/services/rpc.rs
@@ -232,8 +232,8 @@ mod tests {
         let time_difference = expected
             .version
             .timestamp
-            .timestamp_millis()
-            - result
+            .timestamp_millis() -
+            result
                 .version
                 .timestamp
                 .timestamp_millis();
--- a/src/storage/postgres/mod.rs
+++ b/src/storage/postgres/mod.rs
@@ -226,7 +226,7 @@ impl StorageError {
             ) => {
                 if let Some(col) = details.column_name() {
                     if col == "id" {
-                        return StorageError::DuplicateEntry(entity.to_owned(), id.to_owned());
+                        return StorageError::DuplicateEntry(entity.to_owned(), id.to_owned())
                     }
                 }
                 StorageError::Unexpected(err_string)
@@ -237,7 +237,7 @@ impl StorageError {
                         entity.to_owned(),
                         id.to_owned(),
                         related_entitiy,
-                    );
+                    )
                 }
                 StorageError::NotFound(entity.to_owned(), id.to_owned())
             }


src/storage/mod.rs
```diff
@@ -545,10 +539,10 @@ pub trait ContractStateGateway {
     async fn get_contract_slots(
         &self,
         chain: Chain,
-        contracts: Option<&[Self::Address]>,
-        at: Option<Version>,
+        contracts: Option<&[&[u8]]>,
```
pistomat, line 546: Why is the Address type removed?

src/storage/mod.rs
```diff
@@ -455,15 +427,40 @@ pub trait ProtocolGateway {
     /// # Return
     /// Ok if all tokens could be inserted, Err if at least one token failed to
     /// insert.
-    async fn add_tokens(&self, chain: Chain, token: &[Self::Token]) -> Result<(), StorageError>;
+    async fn add_tokens(&self, chain: Chain, token: &[&Self::Token]) -> Result<(), StorageError>;
 }
 
-pub trait StorableContract<S, N, DbId> {
-    fn from_storage(val: S) -> Self;
+/// A binary key value store for an account
```
pistomat, line 440: Could you please add type hints to the docs? I am not sure what the `ContractStore` key and values are supposed to be. Ideally there would be a descriptive type alias for them, so one does not have to search documentation.

src/storage/postgres/chain.rs
```diff
@@ -1,5 +1,6 @@
 use async_trait::async_trait;
 use diesel::prelude::*;
+#[allow(unused_imports)]
```
pistomat: Move the `diesel_async::AsyncConnection` import to `mod test` and remove this macro.

src/storage/postgres/extraction_state.rs
```diff
@@ -4,15 +4,18 @@ use super::orm;
 use super::schema;
 use super::{Chain, PostgresGateway, StorageError};
 use async_trait::async_trait;
+#[allow(unused_imports)]
```
pistomat: Move the import to `mod test` and remove this macro.

src/storage/postgres/extraction_state.rs
```diff
@@ -4,15 +4,18 @@ use super::orm;
...
 use diesel::prelude::*;
 use diesel::ExpressionMethods;
+#[allow(unused_imports)]
```
pistomat: Move the `AsyncConnection` import to `mod test` and remove this macro.

src/storage/postgres/contract_state.rs
```diff
@@ -94,90 +176,71 @@ where
...
+        let chain_id = self.get_chain_id(new.chain());
+        let mut txns = HashSet::new();
+        txns.insert(new.code_modify_tx());
+        txns.insert(new.balance_modify_tx());
+        if let Some(h) = new.creation_tx() {
+            txns.insert(h);
+        }
+        let tx_data: HashMap<Vec<u8>, (i64, NaiveDateTime)> = schema::transaction::table
+            .inner_join(schema::block::table)
+            .filter(schema::transaction::hash.eq_any(&txns))
```
pistomat, line 189: Add this `filter` before `inner_join`

src/storage/postgres/contract_state.rs
```diff
@@ -6,86 +6,168 @@ use diesel_async::RunQueryDsl;
...
-            StorageError::from_diesel(err, "Account", &h160_address.to_string(), None)
-        })?;
-        let version_ts = version_to_ts(&version, db).await?;
+        let account_orm: orm::Account = orm::Account::by_id(id, db)
+            .await
+            .map_err(|err| StorageError::from_diesel(err, "Account", &hex::encode(&id.1), None))?;
+        let version_ts = version_to_ts(version, db).await?;
 
-        let balance_query = schema::account_balance::table
+        let (balance_tx, balance_orm) = schema::account_balance::table
```
pistomat, line 128: The join with `transaction` and the corresponding select can be moved to the end to have smaller joins.

```
let (balance_tx, balance_orm) = schema::account_balance::table
            .filter(schema::account_balance::account_id.eq(account_orm.id))
            .filter(schema::account_balance::valid_from.le(version_ts))
            .filter(
                schema::account_balance::valid_to
                    .gt(Some(version_ts))
                    .or(schema::account_balance::valid_to.is_null()),
            )
            .inner_join(schema::transaction::table)
            .select((schema::transaction::hash, orm::AccountBalance::as_select()))
            .first::<(Vec<u8>, orm::AccountBalance)>(db)
            .await?;
```

src/storage/postgres/contract_state.rs
```diff
@@ -6,86 +6,168 @@ use diesel_async::RunQueryDsl;
...
                     .gt(Some(version_ts))
                     .or(schema::account_balance::valid_to.is_null()),
-            );
+            )
+            .first::<(Vec<u8>, orm::AccountBalance)>(db)
+            .await?;
 
-        let balance = balance_query.first::<Vec<u8>>(db).await?;
-        let (code, code_hash) = schema::contract_code::table
+        let (code_tx, code_orm) = schema::contract_code::table
```
pistomat, line 141: Again the join with transaction and the corresponding select can be moved to the end to have smaller joins.
```
let (code_tx, code_orm) = schema::contract_code::table
            .filter(schema::contract_code::account_id.eq(account_orm.id))
            .filter(schema::contract_code::valid_from.le(version_ts))
            .filter(
                schema::contract_code::valid_to
                    .gt(Some(version_ts))
                    .or(schema::contract_code::valid_to.is_null()),
            )
            .inner_join(schema::transaction::table)
            .select((schema::transaction::hash, orm::ContractCode::as_select()))
            .first::<(Vec<u8>, orm::ContractCode)>(db)
            .await?;
```

src/storage/postgres/contract_state.rs
```diff
@@ -94,90 +176,71 @@ where
...
-                    })?;
-                (Some(tx_id), _created_ts)
-            }
-            None => (None, now),
+    ) -> Result<(), StorageError> {
+        let chain_id = self.get_chain_id(new.chain());
+        let mut txns = HashSet::new();
+        txns.insert(new.code_modify_tx());
+        txns.insert(new.balance_modify_tx());
+        if let Some(h) = new.creation_tx() {
```
pistomat, line 184: There is the expression `if let Some(h) = new.creation_tx()` twice - here and line 199, can you merge the logic?

src/storage/postgres/contract_state.rs
```diff
@@ -94,90 +176,71 @@ where
...
-                    .await
-                    .map_err(|err| {
-                        StorageError::from_diesel(err, "Transaction", &tx.to_string(), None)
-                    })?;
-                (Some(tx_id), _created_ts)
-            }
-            None => (None, now),
+    ) -> Result<(), StorageError> {
+        let chain_id = self.get_chain_id(new.chain());
+        let mut txns = HashSet::new();
```
pistomat, line 181: You can remove the `mut` initiation with functional
```
let txns: HashSet<&[u8]> = HashSet::from_iter(
    vec![
        Some(new.code_modify_tx()),
        Some(new.balance_modify_tx()),
        new.creation_tx(),
    ]
    .into_iter()
    .flatten()
);
```

src/storage/postgres/contract_state.rs
kayibal: I wonder if this makes a difference or if this is not latest optimised later on anyway by Postgres..

src/storage/mod.rs
kayibal: I will probably add sth like Bytes, so most of the `Vec<u8>` are simple variable length byte strings and they should then be interpreted by the corresponding chain specific strict e.g. `evm::Account` this struct would then interpret a `tx_hash` as H256 and an address as H160.

src/storage/mod.rs
kayibal: I wanted to reduce the number of generic type variables that we need on the Gateway. 

The downside of this is that the gateway now has to expose the raw type bytes instead in some of its interfaces. I think though that it is ok for a Gateway to expose a general purpose low level type like Bytes. Although not very ergonomical I think in this case it warrants the benefit of having simpler types.

I'll try to improve the readability a bit though!

src/storage/postgres/contract_state.rs
kayibal: IMO all of these should be handled by the optimiser internally. We worked previously on a versioned system and we ended up hitting the limits of the optimiser. Had some queries where I needed to fix the position of a certain join to optimise it but those were a bit complexer and usually involved some subqueries.

So given that the optimiser here can probably figure out the best way to order things (which might well depend on the content of our columns - not just the logic). I think we should order by best readability. Not sure if join at the top is the best.

src/storage/postgres/contract_state.rs
kayibal: I saw it too when I wrote this but I don't think it is easily possible because I need HashSet in the query below before I need to check again is `creation_tx` was set.

src/storage/postgres/contract_state.rs
pistomat: It probably will be optimised somehow, but Postgres runs in mysterious ways and the execution plans are not reliable always.

PR #23, opened 2023-09-08
Implement getting multiple contracts

kayibal:
- Add a method to retrieve multiple contracts at once.
- Add and use more descriptive type aliases in gateway package.

### DIFF
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -54,7 +54,7 @@ use crate::models::{Chain, ExtractionState, ProtocolComponent, ProtocolSystem};
 // We use variable length bytes type for literals of the contract store for
 // flexibility and because postgres does not support a fixed size byte array
 // column type. On blockchains these are usually fixed, e.g. 32 bytes on
-// ethereuE but might have different lengths (e.g. addresses in Ethereum has 20
+// ethereum but might have different lengths (e.g. addresses in Ethereum has 20
 // bytes length vs Starknet uses 32 byte addresses).
 
 /// Address hash literal type to uniquely identify contracts/accounts on a


src/storage/postgres/contract_state.rs
```diff
@@ -101,6 +104,21 @@ impl StorableContract<orm::Contract, orm::NewContract, i64> for evm::Account {
...
+    tx: Option<Vec<u8>>,
+}
+
+impl<T> Deref for WithTxHash<T> {
+    type Target = T;
+
+    fn deref(&self) -> &Self::Target {
+        &self.entity
+    }
+}
```
carloszanella, line 120: this Deref is just to unwrap?

src/storage/postgres/contract_state.rs
```diff
@@ -187,23 +205,31 @@ where
...
                 .filter(
                     deleted_at
                         .is_null()
-                        .or(deleted_at.lt(version_ts)),
+                        .or(deleted_at.gt(version_ts)),
                 )
                 .order_by(id)
-                .select(orm::Account::as_select())
+                .select((orm::Account::as_select(), schema::transaction::hash.nullable()))
                 .into_boxed();
```
carloszanella, line 228: what does "into_boxed" do?

src/storage/mod.rs
```diff
@@ -51,6 +51,23 @@ use thiserror::Error;
 
 use crate::models::{Chain, ExtractionState, ProtocolComponent, ProtocolSystem};
 
+// We use variable length bytes type for literals of the contract store for
+// flexibility and because postgres does not support a fixed size byte array
+// column type. On blockchains these are usually fixed, e.g. 32 bytes on
+// ethereuE but might have different lengths (e.g. addresses in Ethereum has 20
```
carloszanella: typo

src/storage/mod.rs
```diff
@@ -51,6 +51,23 @@ use thiserror::Error;
...
+/// Address hash literal type to uniquely identify contracts/accounts on a
+/// blockchain.
+type Address = Vec<u8>;
+/// Block hash literal type to uniquely identify a block in the chain and
+/// likely across chains.
+type BlockHash = Vec<u8>;
+
+type AddressRef<'a> = &'a [u8];
+type BlockHashRef<'a> = &'a [u8];
+type TxHashRef<'a> = &'a [u8];
```
carloszanella, line 69: 👍

src/storage/mod.rs
```diff
@@ -113,33 +140,37 @@ pub trait StorableBlock<S, N> {
...
-/// `S` represents the storage specific transaction type e.g. orm struct. `N`
-/// represents the storage specific type for a new transaction,. `BH`
-/// corresponds to the block hash type that uniquely identifies a block. `DbId`
-/// is the ID used to refer to each transaction in the database.
-pub trait StorableTransaction<S, N, DbId> {
+/// Generics:
+/// * `S`: This represents the storage-specific data type used when converting from storage to the
+///   transaction.
+/// * `N`: This represents the storage-specific data type used when converting from the transaction
+///   to storage.
```
carloszanella, line 150: why is S and N different? in the implementation they are the ORM objects no?

src/storage/postgres/contract_state.rs
```diff
@@ -1056,7 +1062,7 @@ mod test {
                     .parse()
                     .unwrap(),
                 title: "account0".to_owned(),
-                slots: make_evm_slots(&vec![(1, 5), (2, 1), (0, 1)]),
+                slots: make_evm_slots(&[(1, 5), (2, 1), (0, 1)]),
```
carloszanella, line 1065: "Comfort clippy" what does this mean?

pistomat: This is the lint he is referring to https://rust-lang.github.io/rust-clippy/v0.0.212/index.html#useless_vec
The `make_evm_slots()` takes an array as the argument, so it is useless to convert the tuples to vector and then back to array.

src/storage/postgres/contract_state.rs
kayibal: It basically lets me access attributes/methods of `self.entity` without having to redeclare them.

src/storage/postgres/contract_state.rs
kayibal: It returns a boxed version of the query. It basically let's me do conditional filtering on it:

> This is useful for cases where you want to conditionally modify a query, but need the type to remain the same. The backend must be specified as part of this. It is not possible to box a query and have it be useable on multiple backends.

https://docs.rs/diesel/latest/diesel/prelude/trait.QueryDsl.html#method.into_boxed

src/storage/mod.rs
kayibal: Well S is how we get it from the DB, including it's primary key id etc and N is a struct that can be inserted. Do it does not contain its own id yet, as it has not been inserted.

PR #25, opened 2023-09-08
Upsert contract

kayibal:


src/storage/postgres/contract_state.rs
```diff
@@ -347,7 +364,30 @@ where
...
+    /// `account_id` and `modify_tx`. This approach allows only one change per
+    /// account and transaction.
+    ///
+    /// To ensure this behavior and support upserts on versioned tables, we have
+    /// implemented an *AFTER INSERT* trigger. This trigger updates the `valid_to`
+    /// column but only after a successful insert. The need for this strategy comes
+    /// from the fact that a *BEFORE INSERT* trigger would update the row even if
+    /// the subsequent insertion were to fail.
+    ///
+    /// Please refer to the interface documentation for more detailed information.
```
pistomat: Maybe add a link to the interface as per https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html ?

PR #26, opened 2023-09-11
Protobuf Setup

kayibal:
## Protobuf setup: 

We want to share some protobuf definition between substreams packages and tycho. Therefore Tycho defines a protobuf module, which the substream packages can reference. They can still define their own internal messages.

Big bulk of the changed lines are pulled protobuf definitions as well as autogenerated protobuf message structs 

**please skip changes under `./proto` and under `./src/pb/`**

### DIFF
--- a/check.sh
+++ b/check.sh
@@ -1,0 +1,5 @@
+set -e 
+
+cargo +nightly fmt -- --check
+cargo +nightly clippy --all --all-features --all-targets -- -D warnings
+cargo test --all --all-features


check.sh
```diff
@@ -0,0 +1,5 @@
+set -e 
+
+cargo +nightly fmt -- --check
+cargo +nightly clippy --all --all-features --all-targets -- -D warnings
```
pistomat, line 4: Nitpick: replace `--all` with `--workspace` as it is a deprecated alias.

kayibal: Will include in the next PR :)

PR #27, opened 2023-09-11
[WIP] ENG-1750: Parse incoming (un)subscription requests

pistomat:
There are a bunch of things I am unclear on, I put them in comments, I would appreciate an input from you.

### DIFF
--- a/src/services/ws.rs
+++ b/src/services/ws.rs
@@ -5,7 +5,7 @@ use crate::{
     models::{ExtractorIdentity, NormalisedMessage},
 };
 use actix::{Actor, ActorContext, StreamHandler};
-use actix_web::{web, App, Error, HttpRequest, HttpResponse, HttpServer};
+use actix_web::{web, Error, HttpRequest, HttpResponse};
 use actix_web_actors::ws;
 use serde::{Deserialize, Serialize};
 use std::{
@@ -15,7 +15,6 @@ use std::{
 use thiserror::Error;
 use tokio::sync::mpsc::Receiver;
 use tracing::{error, info};
-use tracing_actix_web::TracingLogger;
 use uuid::Uuid;
 
 #[derive(Error, Debug)]


src/services/ws.rs
```diff
@@ -1,31 +1,150 @@
...
+            .app_state
+            .extractors
+            .lock()
+            .unwrap();
+
+        if let Some(extractor_handle) = extractors_guard.get(extractor_id) {
+            // Generate a unique ID for this subscription
+            let _subscription_id = Uuid::new_v4();
+
+            // FIXME: Subscribe to the events and spawn an Actor forwarding the messages to the
```
pistomat, line 110: The `extractor_handle.subscribe()` is actually async and I am not sure how to handle this properly.

According to [stackoverflow](https://stackoverflow.com/questions/64434912/how-to-correctly-call-async-functions-in-a-websocket-handler-in-actix-web) I could spawn a new task from the context, but I need to persist the returned `Receiver<Arc<M>>` to handle unsubscribing.

src/services/ws.rs
```diff
@@ -1,31 +1,150 @@
...
+    }
+
+    fn unsubscribe(&mut self, ctx: &mut ws::WebsocketContext<Self>, subscription_id: Uuid) {
+        info!("Unsubscribing from subscription: {:?}", subscription_id);
+
+        if let Some(mut receiver) = self
+            .subscriptions
+            .remove(&subscription_id)
+        {
+            // FIXME: Should I also call extractor_handle.stop()?
```
pistomat, line 128: Probably not, I assume the extractor_handle.stop() would stop the whole extractor, even though there can be other connections consuming the extractor.

Cargo.toml
```diff
@@ -9,10 +9,6 @@ edition = "2021"
 name = "tycho-indexer"
 path = "src/main.rs"
 
-[[bin]]
```
pistomat, line 12: Declaring the `src/services/ws.rs` as a new executable messes up with imports. So the whole App should be instantiated in `main.rs` after the whole Websocket service is properly implemented.

src/services/ws.rs
```diff
@@ -1,31 +1,149 @@
...
+            WebsocketError::ParseError(e) => {
+                serializer.serialize_str(&format!("Parse error: {:?}", e))
+            }
+        }
+    }
+}
+
+/// Shared application data between all connections
+/// Parameters are hidden behind a Mutex to allow for sharing between threads
+pub struct AppState<M> {
```
kayibal, line 59: Hm what do you think about M here. It basically means we will only be able to send a single message type. I guess it is ok for the first iteration but this might turn out to be problematic later on.

src/services/ws.rs
kayibal: Yeah no we only stop the extractor if we are shutting down the whole application, the extractors should run even if they have no subscribers atm.

src/services/ws.rs
pistomat: I know, I will try to think of a way how to remove the M from ExtractorHandle.

We can perhaps remove it with something like `Box<dyn NormalisedMessage + Sync + Send>`, but this would incur runtime costs.

PR #28, opened 2023-09-12
ENG-1751: Spawn a task forwarding extractor

pistomat:
Implementation of the websocket logic. Handles:
- New connections
- New subscriptions
- Closing subscriptions
- Forwarding messages from the Extractors

It uses a single Actor per connection, new subscriptions are added as a new stream to the Actor's context, subscriptions are closed with the SpawnHandle.

TODO:
- Remove the M generic from AppState or at least make it dyn

### DIFF
--- a/src/extractor/runner.rs
+++ b/src/extractor/runner.rs
@@ -1,5 +1,5 @@
 use anyhow::{format_err, Context};
-use futures03::executor::block_on;
+use async_trait::async_trait;
 use prost::Message;
 use std::{collections::HashMap, env, error::Error, sync::Arc};
 use tokio::{
@@ -24,11 +24,15 @@ pub enum ControlMessage<M> {
     Subscribe(Sender<Arc<M>>),
 }
 
-pub trait ExtractorSubscriber<M>: Send + Sync
+/// A trait for a message sender that can be used to subscribe to messages
+///
+/// Extracted out of the [ExtractorHandle] to allow for easier testing
+#[async_trait]
+pub trait MessageSender<M>: Send + Sync
 where
     M: NormalisedMessage + Sync + Send + 'static,
 {
-    fn subscribe(&self) -> Result<Receiver<Arc<M>>, SendError<ControlMessage<M>>>;
+    async fn subscribe(&self) -> Result<Receiver<Arc<M>>, SendError<ControlMessage<M>>>;
 }
 
 pub struct ExtractorHandle<M> {
@@ -53,16 +57,17 @@ where
     }
 }
 
-impl<M> ExtractorSubscriber<M> for ExtractorHandle<M>
+#[async_trait]
+impl<M> MessageSender<M> for ExtractorHandle<M>
 where
     M: NormalisedMessage + Sync + Send + 'static,
 {
-    fn subscribe(&self) -> Result<Receiver<Arc<M>>, SendError<ControlMessage<M>>> {
+    async fn subscribe(&self) -> Result<Receiver<Arc<M>>, SendError<ControlMessage<M>>> {
         let (tx, rx) = mpsc::channel(1);
-        block_on(
-            self.control_tx
-                .send(ControlMessage::Subscribe(tx)),
-        )?;
+        self.control_tx
+            .send(ControlMessage::Subscribe(tx))
+            .await?;
+
         Ok(rx)
     }
 }
--- a/src/services/ws.rs
+++ b/src/services/ws.rs
@@ -1,13 +1,14 @@
 //! This module contains Tycho Websocket implementation
 
 use crate::{
-    extractor::runner::ExtractorSubscriber,
+    extractor::runner::MessageSender,
     models::{ExtractorIdentity, NormalisedMessage},
 };
 use actix::{Actor, ActorContext, AsyncContext, SpawnHandle, StreamHandler};
 use actix_web::{web, Error, HttpRequest, HttpResponse};
 use actix_web_actors::ws;
-use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use futures03::executor::block_on;
+use serde::{Deserialize, Serialize};
 use std::{
     collections::HashMap,
     fmt::Debug,
@@ -15,7 +16,7 @@ use std::{
     time::{Duration, Instant},
 };
 use thiserror::Error;
-use tracing::error;
+use tracing::{error, info, warn};
 use uuid::Uuid;
 
 /// How often heartbeat pings are sent
@@ -32,13 +33,10 @@ pub enum WebsocketError {
     SubscriptionNotFound(Uuid),
 
     #[error("Failed to parse JSON: {0}")]
-    ParseError(serde_json::Error),
-}
+    ParseError(#[from] serde_json::Error),
 
-impl From<serde_json::Error> for WebsocketError {
-    fn from(e: serde_json::Error) -> Self {
-        WebsocketError::ParseError(e)
-    }
+    #[error("Failed to subscribe to extractor: {0}")]
+    SubscribeError(ExtractorIdentity),
 }
 
 impl Serialize for WebsocketError {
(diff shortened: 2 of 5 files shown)


src/services/ws.rs
```diff
@@ -1,140 +1,420 @@
...
+use actix_web::{web, Error, HttpRequest, HttpResponse};
 use actix_web_actors::ws;
-use std::sync::Mutex;
-use tracing::info;
-use tracing_actix_web::TracingLogger;
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use std::{
+    collections::HashMap,
+    fmt::Debug,
+    sync::{Arc, Mutex},
```
kayibal, line 15: Is there a specific reason to use the Mutex from std. I'd prefer to use tokios Mutex as we are in an async context and waiting for a lock from `std::sync::Mutex` will block the entire thread.

src/services/ws.rs
```diff
@@ -1,140 +1,420 @@
...
+#[derive(Error, Debug)]
+pub enum WebsocketError {
+    #[error("Extractor not found: {0}")]
+    ExtractorNotFound(ExtractorIdentity),
+
+    #[error("Subscription not found: {0}")]
+    SubscriptionNotFound(Uuid),
+
+    #[error("Failed to parse JSON: {0}")]
+    ParseError(serde_json::Error),
```
kayibal:
```suggestion
    ParseError(#[from] serde_json::Error),
```

src/services/ws.rs
```diff
@@ -1,140 +1,420 @@
...
+}
+
+impl From<serde_json::Error> for WebsocketError {
+    fn from(e: serde_json::Error) -> Self {
+        WebsocketError::ParseError(e)
+    }
```
kayibal: If you use the macro like suggested above I think you don't need this.

src/services/ws.rs
```diff
@@ -1,140 +1,420 @@
...
+    ) -> Result<HttpResponse, Error> {
+        ws::start(WsActor::new(data), &req, stream)
+    }
+
+    fn heartbeat(&mut self, ctx: &mut <Self as Actor>::Context) {
+        // Send a first heartbeat ping
+        ctx.run_interval(HEARTBEAT_INTERVAL, |act, ctx| {
+            if Instant::now().duration_since(act.heartbeat) > CLIENT_TIMEOUT {
+                dbg!("Websocket Client heartbeat failed, disconnecting!");
+                ctx.stop();
```
kayibal, line 120: This basically stops the actor, right?

src/services/ws.rs
```diff
@@ -1,140 +1,420 @@
...
+    pub async fn ws_index(
+        req: HttpRequest,
+        stream: web::Payload,
+        data: web::Data<AppState<M>>,
+    ) -> Result<HttpResponse, Error> {
+        ws::start(WsActor::new(data), &req, stream)
+    }
+
+    fn heartbeat(&mut self, ctx: &mut <Self as Actor>::Context) {
+        // Send a first heartbeat ping
```
kayibal: Code below looks more like it sends pings repeatedly.

src/services/ws.rs
```diff
@@ -1,140 +1,420 @@
...
+
+                        let error = WebsocketError::ExtractorNotFound(extractor_id.clone());
+                        ctx.text(serde_json::to_string(&error).unwrap());
+                    }
+                };
+            } else {
+                error!("Extractor not found: {:?}", extractor_id);
+
+                let error = WebsocketError::ExtractorNotFound(extractor_id.clone());
+                ctx.text(serde_json::to_string(&error).unwrap());
```
kayibal, line 174: Will this send just a string or a nice error json?

src/services/ws.rs
```diff
@@ -1,140 +1,420 @@
...
+                            .insert(subscription_id, handle);
+                        dbg!("Added subscription to hashmap: {:?}", subscription_id);
+
+                        let message = Response::NewSubscription { subscription_id };
+                        ctx.text(serde_json::to_string(&message).unwrap());
+                    }
+                    Err(e) => {
+                        error!("Failed to subscribe to extractor: {:?}", e);
+
+                        let error = WebsocketError::ExtractorNotFound(extractor_id.clone());
```
kayibal: Hm error type here is not great. I guess if errors here are very unexpected, the extractor would need to have died or sth..

src/services/ws.rs
```diff
@@ -1,140 +1,420 @@
...
+                }
             }
             Ok(ws::Message::Binary(bin)) => {
-                info!("Websocket binary message received: {:?}", bin);
+                dbg!("Websocket binary message received: {:?}", bin.clone());
                 ctx.binary(bin)
             }
             Ok(ws::Message::Close(reason)) => {
-                info!("Websocket close message received: {:?}", reason);
+                dbg!("Websocket close message received: {:?}", reason.clone());
```
kayibal: I like the `dbg!` in this module, but I think they should be made `debug!`, in some places a bit more context could help debugging things in the future. I think the clone here is also avoidable with a reference.

src/services/ws.rs
```diff
@@ -1,140 +1,420 @@
...
+            let error = WebsocketError::SubscriptionNotFound(subscription_id);
+            ctx.text(serde_json::to_string(&error).unwrap());
+        }
     }
 }
 
-impl Actor for WsActor {
+impl<M> Actor for WsActor<M>
+where
+    M: NormalisedMessage + Serialize + DeserializeOwned + Sync + Send + 'static,
```
kayibal: Maybe we can move all these additional trait bounds to the NormalisedMessage trait as super traits?

src/extractor/runner.rs
```diff
@@ -23,6 +24,13 @@ pub enum ControlMessage<M> {
     Subscribe(Sender<Arc<M>>),
 }
 
+pub trait ExtractorSubscriber<M>: Send + Sync
+where
+    M: NormalisedMessage + Sync + Send + 'static,
+{
+    fn subscribe(&self) -> Result<Receiver<Arc<M>>, SendError<ControlMessage<M>>>;
+}
+
```
kayibal: I guess you pulled that out for testing purposes? I find the name a bit weird but the only other things that comes to mind is MessagePublisher, or MessageSender.

src/services/ws.rs
pistomat: Yes, I am not very knowledgable on websockets, but my understanding is that it is a best practice to expect a heartbeat ping to close unused connections i.e. WsActors. The timeout can be of course setup to however long we want.

src/services/ws.rs
pistomat: It does both, we want to check received heartbeat pings and also send them to the client.

[Here](https://github.com/propeller-heads/tycho-indexer/blob/3a017be7fbfc59e85a2cdf67b5ea0ee6274eda3b/src/services/ws.rs#L259) we set the time of last receiver heartbeat and in this function we periodically check that the client didn't timeout. If the check is successful, we send a heartbeat back.

src/extractor/runner.rs
pistomat: Yes, it is pulled out so I can test it. I am open to any naming suggestion.

src/services/ws.rs
pistomat: Because I am not handing the context to any other threads, I am using it in sync `fn subscribe(...)` where the standard Mutex is preferred [according to the docs](https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use) afaik.

src/services/ws.rs
pistomat: `ctx.text` sends text, the formatting of the `WebsocketError` is done manually now with a simple text explanation, because `serde_json::Error` does not implement the `Serialize` trait. I am open to suggestions on better formatting.

src/services/ws.rs
```diff
@@ -105,37 +110,41 @@ where
...
 
-            if let Some(extractor_subscriber) = extractors_guard.get(extractor_id) {
+            if let Some(message_sender) = extractors_guard.get(extractor_id) {
                 // Generate a unique ID for this subscription
                 let subscription_id = Uuid::new_v4();
-                dbg!("Generated subscription ID: {:?}", subscription_id);
+                info!("Generated subscription ID: {:?}", subscription_id);
 
-                match extractor_subscriber.subscribe() {
+                match block_on(message_sender.subscribe()) {
```
kayibal, line 147: Nice

PR #29, opened 2023-09-13
Ambient Extractor - Part I

kayibal:
In this PR I fully implement Ambient Extractor and hook it up to main.rs, so it runs fine now and inserts all the data we expect into the database.

_The commits contain a lot of restructuring work. So unless you are interested how I arrived at the current solution this is best reviewed as a single diff._

## Gateway Interface Changes

To make the extractor DB interactions a bit nicer and make future extractors easier to add, the changed the following:

- Added a new generic type which represents delta changes to an account
- Removed any method operating solely on slots
- Added support to retrieve delta changes for accounts
- Added separate methods to insert contracts and update contracts using deltas. This way, extractors are not required to keep a local state of all the accounts they extract. It also reduces the amount of database queries required to update a contract.

### DIFF
--- a/src/extractor/evm/mod.rs
+++ b/src/extractor/evm/mod.rs
@@ -19,12 +19,12 @@ use serde::{Deserialize, Serialize};
 
 use super::ExtractionError;
 
-#[derive(Debug)]
+#[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct SwapPool {}
 
 pub struct ERC20Token {}
 
-#[derive(Debug, PartialEq, Copy, Clone)]
+#[derive(Debug, PartialEq, Copy, Clone, Deserialize, Serialize)]
 pub struct Block {
     pub number: u64,
     pub hash: H256,
@@ -150,7 +150,7 @@ impl AccountUpdate {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct BlockAccountChanges {
     extractor: String,
     chain: Chain,


src/extractor/evm/ambient.rs
```diff
@@ -31,28 +47,201 @@ impl<DB> AmbientContractExtractor<DB> {
...
+        debug!("Upserting block: {:?}", &changes.block);
+        self.state_gateway
+            .upsert_block(&changes.block, conn)
+            .await?;
+        for update in changes.tx_updates.iter() {
+            debug!("Processing tx: 0x{:x}", &update.tx.hash);
+            self.state_gateway
+                .upsert_tx(&update.tx, conn)
+                .await?;
+            if update.tx.hash == H256(DEPLOY_TX) {
```
zizou0x, line 115: How does this work?

src/extractor/evm/ambient.rs
```diff
@@ -31,28 +47,201 @@ impl<DB> AmbientContractExtractor<DB> {
...
+        let mut conn = self.pool.get().await.unwrap();
+        conn.transaction(|conn| {
+            async move {
+                self.forward(changes, new_cursor, conn)
+                    .await
+            }
+            .scope_boxed()
+        })
+        .await?;
+        Ok(())
```
zizou0x, line 167: I think we can just do `.await` here

migrations/2023-08-15-154244_tycho_initial_setup/up.sql
```diff
@@ -429,7 +429,8 @@ BEGIN
     WHERE
         valid_to IS NULL
         AND account_id = NEW.account_id
-        -- we run this as an after insert trigger so we can use upserts
+        -- running this after inserts allows us to use upserts, 
+        -- currently the application does not use that though
```
carloszanella, line 433: currently, or after this PR?

carloszanella: this applies only to contract code and account balance?

src/extractor/evm/ambient.rs
```diff
@@ -31,28 +47,201 @@ impl<DB> AmbientContractExtractor<DB> {
...
 #[async_trait]
-trait AmbientGateway: Send + Sync {
+pub trait AmbientGateway: Send + Sync {
     async fn get_cursor(&self, name: &str, chain: Chain) -> Result<Vec<u8>, StorageError>;
     async fn upsert_contract(
         &self,
-        changes: evm::BlockStateChanges,
-        new_cursor: &[u8],
+        changes: &evm::BlockStateChanges,
+        new_cursor: &str,
```
zizou0x, line 63: What is this cursor used for?

src/extractor/evm/ambient.rs
```diff
@@ -20,6 +34,8 @@ pub struct AmbientContractExtractor<G> {
     gateway: G,
     name: String,
     chain: Chain,
+    // TODO: There is not reason this needs to be shared
+    // try removing the Mutex
     inner: Arc<Mutex<Inner>>,
```
carloszanella, line 39: what is inner for? maybe some short docs but also happy with just an answer to the question here

src/storage/postgres/mod.rs
```diff
@@ -286,6 +316,92 @@ impl<B, TX, A> PostgresGateway<B, TX, A> {
...
+    TX: StorableTransaction<orm::Transaction, orm::NewTransaction, i64>,
+    D: ContractDelta,
+    A: StorableContract<orm::Contract, orm::NewContract, i64>,
+{
+    // No methods in here - this just ties everything together
+}
+
+/// Establishes a connection to the database and creates a connection pool.
+///
+/// This function takes in the URL of the database as an argument and returns a pool
```
zizou0x, line 341: Reading this, it's not clear to me what this "pool of connections" is used for. Is it documented somewhere? 
Intuitively I would expect a single connection to the database.

src/extractor/evm/ambient.rs
kayibal: This is a bit hacky but it basically will create the ambient account if it sees the deployment transaction (which is hardcoded). We'll later have to add a flag that will let us know if an account was updated, created or destroyed.

src/extractor/evm/ambient.rs
kayibal: https://substreams.streamingfast.io/developers-guide/sink-targets/substreams-sink-files#cursors

src/storage/postgres/mod.rs
kayibal: https://github.com/djc/bb8

> Opening a new database connection every time one is needed is both inefficient and can lead to resource exhaustion under high traffic conditions. A connection pool maintains a set of open connections to a database, handing them out for repeated use.

src/storage/mod.rs
```diff
@@ -575,25 +567,85 @@ pub trait StorableContract<S, N, I>: Send + Sync + 'static {
...
+    /// # Returns
+    /// - Result containing the instance of the `ContractDelta` implementation if successful, and a
+    ///   `StorageError` if there was an issue reading from storage.
+    fn from_storage(
+        chain: Chain,
+        address: AddressRef,
+        slots: Option<&ContractStore>,
+        balance: Option<&[u8]>,
+        code: Option<&[u8]>,
+    ) -> Result<Self, StorageError>;
```
carloszanella, line 598: don't you need a period? I might be confusing what you mean with delta here. First thought was this is a difference in state between two points

src/storage/mod.rs
```diff
@@ -732,18 +781,17 @@ pub trait ContractStateGateway {
...
     ///
     /// # Returns
-    /// A map containing the necessary changes to update a state from
-    /// start_version to end_version.
+    /// A map containing the necessary changes to update a state from start_version to end_version.
     /// Errors if:
     ///     - The versions can't be located in storage.
     ///     - There was an error with the database
-    async fn get_slots_delta(
+    async fn get_account_delta(
```
carloszanella, line 788: > Removed any method operating solely on slots

Why does this make it nicer to add new extractors? Would like to understand the reasoning a bit more

src/storage/mod.rs
```diff
@@ -767,12 +815,13 @@ pub trait ContractStateGateway {
 pub trait StateGateway<DB>:
     ExtractionStateGateway<DB = DB>
     + ChainGateway<DB = DB>
-    + ProtocolGateway<DB = DB>
+    // + ProtocolGateway<DB = DB>
```
carloszanella, line 818: leaving the comment?

src/storage/mod.rs
```diff
@@ -767,12 +815,13 @@ pub trait ContractStateGateway {
...
     + ContractStateGateway<DB = DB>
     + Send
     + Sync
 {
 }
 
-pub type StateGatewayType<DB, B, TX, T, C> =
-    Arc<dyn StateGateway<DB, Transaction = TX, Block = B, Token = T, ContractState = C>>;
+pub type StateGatewayType<DB, B, TX, C, D> =
+    Arc<dyn StateGateway<DB, Transaction = TX, Block = B, ContractState = C, Delta = D>>;
```
carloszanella, line 827: why is token no longer needed?

src/storage/postgres/chain.rs
```diff
@@ -144,21 +148,23 @@ impl StorableBlock<orm::Block, orm::NewBlock, i64> for evm::Block {
 
 impl StorableTransaction<orm::Transaction, orm::NewTransaction, i64> for evm::Transaction {
     fn from_storage(val: orm::Transaction, block_hash: &[u8]) -> Self {
+        let to = if !val.to.is_empty() { Some(H160::from_slice(&val.to)) } else { None };
         Self {
             hash: H256::from_slice(&val.hash),
             block_hash: H256::from_slice(block_hash),
             from: H160::from_slice(&val.from),
-            to: H160::from_slice(&val.to),
+            to,
```
carloszanella, line 156: this being empty is a contract creation transaction?

src/storage/postgres/mod.rs
```diff
@@ -158,6 +163,23 @@ where
...
+            use schema::chain::dsl::*;
+            chain
+                .select((id, name))
+                .load(&mut conn)
+                .await
+                .expect("Failed to load chain ids!")
+        }
+        .await;
+        Ok(Self::from_tuples(&results))
+    }
```
carloszanella, line 182: hmm I didn't look in detail at this before but this enum class is a very generic util it seems. I find it a bit weird to have a method referring to `Pool` here

carloszanella: oh damn just saw that Pool comes from dsl.. ok makes more sense now

src/storage/postgres/orm.rs
```diff
@@ -415,7 +415,7 @@ impl ContractCode {
 #[diesel(check_for_backend(diesel::pg::Pg))]
 pub struct NewContractCode<'a> {
     pub code: &'a [u8],
-    pub hash: &'a [u8],
+    pub hash: Vec<u8>,
```
carloszanella, line 418: can you explain this change? 

Didn't we have a type alias for hash? maybe we could have type aliases for ORM fields too because I had some struggle onboarding the DB with these types..

src/storage/postgres/orm.rs
```diff
@@ -456,7 +456,7 @@ impl NewContract {
         modify_ts: NaiveDateTime,
     ) -> NewAccountBalance {
         NewAccountBalance {
```
carloszanella, line 458: I didn't see the PR where this struct was added. I see some explanation in comments but I still didn't get why we need it for, therefore I can't assess the change that adds `self.balance.clone()`.

Can you explain what we need this struct for?

src/main.rs
```diff
@@ -61,143 +68,38 @@ async fn main() -> Result<(), Error> {
...
+    let ambient_gw = AmbientPgGateway::new("vm:ambient", Chain::Ethereum, pool, evm_gw);
+    let extractor =
+        AmbientContractExtractor::new("vm:ambient", Chain::Ethereum, ambient_gw).await?;
+
+    let builder = ExtractorRunnerBuilder::new(spkg, Arc::new(extractor))
+        .start_block(17361664)
+        // for testing only
+        .end_block(17362000);
+    let handle = builder.run().await?;
+    Ok(handle)
```
carloszanella, line 102: do we want to have all extractors code on main.rs in the future?

src/extractor/evm/mod.rs
```diff
@@ -79,46 +89,287 @@ impl Account {
...
+            return Err(ExtractionError::Unknown(format!(
+                "Can't merge AccountUpdates from the same transaction: {:x}",
+                self.tx.hash
+            )))
+        }
+        if self.tx.index > other.tx.index {
+            return Err(ExtractionError::Unknown(format!(
+                "Can't merge AccountUpdates with lower transaction index: {} > {}",
+                self.tx.index, other.tx.index
+            )))
```
carloszanella, line 204: couldn't we do `other.merge` on this case?

carloszanella: maybe nice to document assumption that other comes _after_ self?

src/storage/postgres/contract_state.rs
```diff
@@ -120,20 +161,412 @@ impl<T> Deref for WithTxHash<T> {
...
+where
+    B: StorableBlock<orm::Block, orm::NewBlock, i64>,
+    TX: StorableTransaction<orm::Transaction, orm::NewTransaction, i64>,
+    D: ContractDelta,
+    A: StorableContract<orm::Contract, orm::NewContract, i64>,
+{
+    /// Retrieves deltas for any accounts balance changes.
+    ///
+    /// See [`ContractStateGateway::get_account_deltas`] for more information.
+    async fn get_balance_deltas(
```
carloszanella, line 174: why does "get_balance_deltas" points me to "get_account_deltas"? Same for the others below

src/storage/postgres/contract_state.rs
```diff
@@ -551,187 +1065,28 @@ where
...
-        }
-        diesel::insert_into(schema::contract_storage::table)
-            .values(&new_entries)
-            .execute(conn)
-            .await?;
-        Ok(())
-    }
-
-    async fn get_slots_delta(
+    async fn get_account_delta(
```
carloszanella, line 1068: I think this one could use a reference to the trait, and maybe some comments on its implementation?

src/storage/postgres/contract_state.rs
```diff
@@ -105,6 +100,52 @@ impl StorableContract<orm::Contract, orm::NewContract, i64> for evm::Account {
...
+
+        let update = evm::AccountUpdate::new(
+            parse_id_h160(address)?,
+            chain,
+            slots,
+            balance.map(U256::from_big_endian),
+            code.map(|v| v.to_vec()),
+        );
+        Ok(update)
+    }
```
carloszanella, line 146: I expected some diesel query on the `from_storage` method

src/storage/postgres/contract_state.rs
```diff
@@ -364,126 +805,199 @@ where
...
-            // We do not want to make a noop update as this might cause db side
-            // triggers to fire unnecessarily.
-            schema::account::table
-                .filter(schema::account::address.eq(new.address()))
-                .filter(schema::account::chain_id.eq(chain_id))
-                .select(schema::account::id)
-                .first::<i64>(db)
-                .await?
-        };
+            let slots = delta.dirty_slots();
```
carloszanella, line 963: do you prefer "dirty" over other options? like updated, modified, changed

migrations/2023-08-15-154244_tycho_initial_setup/up.sql
kayibal: > Added separate methods to insert contracts and update contracts using deltas.

Yes, I got rid of upserts - they require the "upserter" to have the complete state present - which is kind of cumbersome. I think I'll keep the triggers as they are though in case we ever need the upserts back.

src/extractor/evm/ambient.rs
kayibal: I only store the cursor there. I initially thought that this state might need to be shared but I don't think it needs to be. So I can probably remove that - although that would mean changing the Extractors interface. I will circle back to this once the first iteration of Tycho is complete.

src/storage/mod.rs
kayibal: It is; A difference between two points. The points usually being the start and end of a transaction, although there is nothing preventing us to do bigger periods. 

The delta itself does not store information about which start and end points it spans over, atm this is contained in the BlockAccountChanges struct:

```rust
#[derive(Debug)]
pub struct BlockAccountChanges {
    extractor: String,
    chain: Chain,
    pub block: Block,
    pub new_pools: HashMap<H160, SwapPool>,
}
```

With `AccountUpdate` Implementing the `ContractDelta` trait for PostgresGateway.

src/storage/mod.rs
kayibal: You can see here how the AmbientExtractor looks like before that change: https://github.com/propeller-heads/tycho-indexer/blob/b2e2e0a1ec51ee61c83cff0b95b4790c90862130/src/extractor/evm/ambient.rs#L205

src/storage/mod.rs
kayibal: Yes because we currently do not implement `ProtocolGateway` trait for `PostgresGateway`. So for types to work out properly this can't be included yet.

src/storage/mod.rs
kayibal: We have no concept of token atm - we just extract the state of ambient. I don't think we will make all the pool metadata extraction still within this mission. So I commented those bits out.

src/storage/postgres/chain.rs
kayibal: Yes if the contract was directly created by an EOA the to can be empty I think... I have to double check that. In the end this did not work out for ambient, cause the contract was deployed by calling another contract. So yeah.. we'll need to set flags for this in the substreams spkgs.

src/storage/postgres/mod.rs
kayibal: This is a connection Pool, Flo also asked about it: 

https://github.com/propeller-heads/tycho-indexer/pull/29#discussion_r1324369366

src/storage/postgres/orm.rs
kayibal: What do you mean by onboarding?

About the change: So previously I would have an `Account` struct (which owns code bytes `code: Vec<u8>`) and would then directly send the query to the DB, so I could use a reference here (because Account still exists while I send the query). 

Now I batch things and the code is only available within the loop so the `NewContractCode` needs to own the code bytes that it will later insert. At least that was the easiest way without doing more complicated stuff.

src/storage/postgres/orm.rs
kayibal: Same reason as for `NewContract` code, because query and construction of values are decoupled, things are easier if the "New" structs own the data they will insert.

I basically put these structs in vectors and later send all the updates for that table in a single query.

src/main.rs
kayibal: Later this will likely be parsed from a configuration file.

src/extractor/evm/mod.rs
kayibal: We can't else I'd need to consume both values. I don't remember now the exact reason why this wasn't nice though...

src/storage/postgres/contract_state.rs
kayibal: Because that describes the context within this method is used.

src/storage/postgres/contract_state.rs
kayibal: Hm I guess I could mention the trigger stuff there... although that is already mentioned in the module level docs. Not sure what would you like to see here?

src/storage/postgres/contract_state.rs
kayibal: Why? This just parses the types we get from storage into more use case specific types (evm ones in that case).

src/storage/postgres/contract_state.rs
kayibal: There is a lot of ambiguity here. I am not 100% happy with all the namings yet. E.g. slots, store, storage.... I will keep working on this.

Updated could work but then it is confusing because we are already trying to update the database int his case.... I feel modified and changed is not specific enough. "dirty" is usually used in programming to indicate there was a change that has not been fully reflected yet in long term storage ... But I guess that is a preference thing in the end-

src/storage/postgres/contract_state.rs
carloszanella: It sounds like you load from storage? That's why I expected it to do queries. If you mean like a ORM of sorts maybe this interface should take in as parameter a storage struct that contains these storage types?
```
    /// Loads a `ContractDelta` from storage given a chain, address reference,
     /// optional slots, balance, and code.
```

src/storage/postgres/contract_state.rs
carloszanella: Does this link point you to the trait definition of account delta or to the implementation done here? Because the trait definition mentions no "get_balance_delta" so it would kind of lead to a dead end if you went there looking for explanations on this method

src/storage/postgres/orm.rs
carloszanella: I mean about onboarding on the codebase, getting familiar with it. I would see a Vec<u8> on the DB for example and didn't know what type it was supposed to be. I think type aliases help a lot to document it

kayibal: I mean the type `Vec<u8>` is bytes, I guess you mean it is unclear what these bytes represent semantically right? E.g is this `tx: Vec<u8>` a hash or a rlp encoded tx?

src/storage/postgres/contract_state.rs
kayibal: It leads to the trait, but yeah there is no detailed description about balances or code. But it describes what the method does in a higher level. But it basically boils down to exactly what is mentioned in the traits docs:


```
    /// Given start version V1 and end version V2, this method will return the
    /// changes necessary to move from V1 to V2. So if V1 < V2, it will contain
    /// the changes of all accounts that changed between the two versions with the
    /// values corresponding to V2. If V2 < V1 then it will contain all the
    /// slots that changed between the two versions with the values corresponding to V1.
```

I''change the doc to say something like:
```
See [`ContractStateGateway::get_account_deltas`] for more information on the version timestamp semantics.
```

src/storage/postgres/contract_state.rs
kayibal: I will change `"Loads"` to `"Converts"` to make it clear, we can later if needed change this to take a generic type parameter representing the object in storage, but for now I think this is better solved with docs.

PR #30, opened 2023-09-14
Test and document get_accounts_delta

kayibal:
As expected this method is a bit more complex, so while testing I had to add the functionality to handle deleted and created contracts in both directions. This seems to work fine now for most cases. 

## A weird edge case...
While working on this, I encountered an edge case that momentarily made me question the soundness of our table schema.

This edge case involves a contract being created, subsequently deleted, and immediately recreated within the same block. Now, either reverting or forwarding this block presents a challenge. Moreover, our account table is not versioned - it holds only a single version of an account. In such a scenario, we would lose information indicating that the account did not exist, even if it was for a brief period.

Note that the issue persists even if these operations occur across several blocks. In this particular instance, retrieving deltas becomes tricky. If we move forward, the correct action would be to emit only the updates that occurred after the final creation, or simply, the state at target (for EVM, this would indeed be the correct resolution).

A more elegant solution could involve versioning the accounts table. For the EVM situation, it might be even more appropriate to move code attributes to this versioned account table, primarily because each change to the code must involve a self-destruct followed by a create2 operation. This is something we should consider for the future. For now, we will only "see" a code update after the second create2 and would remain oblivious that the contract was deleted for a certain duration... that is, if such a scenario were ever to occur.

### TLDR:
If it happens we recognise it and error. So if we ever run into it we can act if not we are probably ok.

### DIFF
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -23,10 +23,10 @@
 //! ### Version semantics intra block
 //!
 //! ```text
-//! tx            1    2       1  2             3          1  2            3
+//! tx            0    1       0  1             2          0  1            2
 //! B01 ----------x----x---B02-x--x-------------x---B03----x--x------------x-->
 //! 00:00                  00:12                 | |00:24
-//! ____Block(B02), VersionKind::Index(3)________| |
+//! ____Block(B02), VersionKind::Index(2)________| |
 //! ____Block(B02), VersionKind::Last______________+
 //!                                                |
 //! ____Block(B03), VersionKind::First_____________|
@@ -35,8 +35,8 @@
 //! Above you'll find the 3 possible version kinds. Note that Index(N) specifies
 //! the transaction slot after N. VersionKind::First includes the last
 //! transaction of the previous block and VersionKind::Last includes the last
-//! transaction of the specified block. So it is possbile to convert between the
-//! variants first and last
+//! transaction of the specified block. So it is possible to refer to the exact
+//! same state using both Last and First VersionKind variants.
 //!
 //! ## Literal Types
 //!
--- a/src/storage/postgres/contract_state.rs
+++ b/src/storage/postgres/contract_state.rs
@@ -63,8 +63,6 @@ where
         conn: &mut AsyncPgConnection,
     ) -> Result<HashMap<i64, Balance>, StorageError> {
         use schema::account_balance::dsl::*;
-        dbg!(&start_version_ts);
-        dbg!(&target_version_ts);
         let res = if start_version_ts <= target_version_ts {
             let changed_account_ids = account_balance
                 .inner_join(schema::account::table.inner_join(schema::chain::table))
@@ -286,14 +284,14 @@ where
         Ok(result)
     }
 
-    /// Handle deleted or created account deltas
+    /// Fetch deleted or created account deltas
     ///
     /// # Operations
     ///   
     /// 1. Going Forward (`start < target`):
     ///     - a) If an account was deleted, emit an empty delta with [ChangeType::Deletion].
     ///     - b) If an account was created, emit an already collected update delta but with
-    ///       [ChangeType::Creation].. No need to fetches the state at the `target_version` as by
+    ///       [ChangeType::Creation].. No need to fetch the state at the `target_version` as by
     ///       design we emit newly created contracts and their components when going forward.
     ///  
     /// 2. Going Backward (`target < start`):
@@ -314,7 +312,7 @@ where
     /// attribute. We can use this attribute to satisfy the first operation. It also contains new
     /// restored / deleted delta structs withing the `restored` attribute with which we can satisfy
     /// the second rule.
-    async fn created_or_deleted_accounts(
+    async fn get_created_or_deleted_accounts(
         &self,
         chain: Chain,
         start_version_ts: &NaiveDateTime,
@@ -1176,7 +1174,7 @@ where
             .get_slots_delta(chain_id, &start_version_ts, &target_version_ts, conn)
             .await?;
         let account_deltas = self
-            .created_or_deleted_accounts(chain, &start_version_ts, &target_version_ts, conn)
+            .get_created_or_deleted_accounts(chain, &start_version_ts, &target_version_ts, conn)
             .await?;
 
         // We retrieve account addresses separately because this is more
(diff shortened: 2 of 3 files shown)


src/extractor/evm/storage.rs
```diff
@@ -119,8 +119,8 @@ pub mod pg {
             chain: Chain,
             address: AddressRef,
             slots: Option<&ContractStore>,
-            balance: Option<&[u8]>,
-            code: Option<&[u8]>,
+            balance: Option<BalanceRef>,
+            code: Option<CodeRef>,
```
carloszanella, line 135: 🙏

src/storage/mod.rs
```diff
@@ -51,23 +64,50 @@ use thiserror::Error;
...
+
 /// Block hash literal type to uniquely identify a block in the chain and
 /// likely across chains.
 pub type BlockHash = Vec<u8>;
 
+/// Smart contract code is represented as a byte vector containing opcodes.
+pub type Code = Vec<u8>;
+
+/// The chain specific hash digest of an account's code.
+pub type CodeHash = Vec<u8>;
```
carloszanella, line 97: these maybe could be on a types.rs file of their own

src/storage/postgres/mod.rs
```diff
@@ -330,7 +330,7 @@ impl<B, TX, A, D> StateGateway<AsyncPgConnection> for PostgresGateway<B, TX, A,
 where
     B: StorableBlock<orm::Block, orm::NewBlock, i64>,
     TX: StorableTransaction<orm::Transaction, orm::NewTransaction, i64>,
-    D: ContractDelta,
+    D: ContractDelta + From<A>,
```
carloszanella, line 335: this means ContractDelta type must implement the From trait for the `A` type?

src/storage/mod.rs
```diff
@@ -20,6 +20,24 @@
...
+//! ____Block(B02), VersionKind::Last______________+
+//!                                                |
+//! ____Block(B03), VersionKind::First_____________|
+//! ```
+//!
+//! Above you'll find the 3 possible version kinds. Note that Index(N) specifies
+//! the transaction slot after N. VersionKind::First includes the last
+//! transaction of the previous block and VersionKind::Last includes the last
+//! transaction of the specified block. So it is possbile to convert between the
+//! variants first and last
```
carloszanella: nice. the 2nd and 3rd options they look the same in the sketch but they are different no? The last one would be the first index on Block 3

src/storage/postgres/contract_state.rs
```diff
@@ -277,6 +286,197 @@ where
...
     }
 
+    /// Handle deleted or created account deltas
+    ///
+    /// # Operations
+    ///   
+    /// 1. Going Forward (`start < target`):
+    ///     - a) If an account was deleted, emit an empty delta with [ChangeType::Deletion].
+    ///     - b) If an account was created, emit an already collected update delta but with
+    ///       [ChangeType::Creation].. No need to fetches the state at the `target_version` as by
```
carloszanella: looks like a typo on "fetches"

src/storage/postgres/contract_state.rs
```diff
@@ -277,6 +286,197 @@ where
         Ok(result)
     }
 
+    /// Handle deleted or created account deltas
```
carloszanella: "handle" is a bit weird. the function name doesn't tell me what it does either. maybe adding a "get" to function name and here is enough

src/storage/postgres/mod.rs
kayibal: Correct this way I can create a Delta type from a full Account struct.

src/extractor/evm/utils.rs
```diff
@@ -0,0 +1,47 @@
...
+
+pub fn u256_to_bytes(v: &U256) -> Vec<u8> {
+    let mut bytes32 = [0u8; 32];
+    v.to_big_endian(&mut bytes32);
+    bytes32.to_vec()
+}
+
+/// Parses an evm address hash from the db
+///
+/// The db id is required to provide additional error context in case the
```
zizou0x, line 11: This is outdated?

src/storage/mod.rs
kayibal: No they are indeed equivalent as in they refer to the exact same state, it is just a different way to express it. The example here is not ideal but if you are working with hashes and not numbers being able to refer to the beginning (before any transaction happened) or end of block state is useful. 

If you want the state after the first transaction which is what I would understand under _"the first index on Block 3"_ you'd need to use `Block(B03), VersionKind::Index(0)` - just noticed that the indices in the plot above should start at 0. So Index(0) means the state after the first transaction. If you want the state before you'd use First. at least this is how I think it is most useful.

src/storage/postgres/contract_state.rs
kayibal: Yeah that one is hard to name :D

src/storage/mod.rs
```diff
@@ -20,6 +20,37 @@
...
+//! ____Block(B02), VersionKind::Index(3)________| |
+//! ____Block(B02), VersionKind::Last______________+
+//!                                                |
+//! ____Block(B03), VersionKind::First_____________|
+//! ```
+//!
+//! Above you'll find the 3 possible version kinds. Note that Index(N) specifies
+//! the transaction slot after N. VersionKind::First includes the last
+//! transaction of the previous block and VersionKind::Last includes the last
+//! transaction of the specified block. So it is possbile to convert between the
```
zizou0x:
```suggestion
//! transaction of the specified block. So it is possible to convert between the
```

src/storage/postgres/orm.rs
```diff
@@ -299,6 +299,15 @@ impl Account {
             .await
     }
 
+    /// retrieves a account by hash
+    pub async fn by_hash(address: &[u8], conn: &mut AsyncPgConnection) -> QueryResult<Vec<Self>> {
```
zizou0x: Isnt it by address?

src/storage/postgres/contract_state.rs
```diff
@@ -161,24 +31,40 @@ impl<T> Deref for WithTxHash<T> {
...
         &self,
         chain_id: i64,
         start_version_ts: &NaiveDateTime,
         target_version_ts: &NaiveDateTime,
         conn: &mut AsyncPgConnection,
-    ) -> Result<HashMap<i64, Vec<u8>>, StorageError> {
+    ) -> Result<HashMap<i64, Balance>, StorageError> {
         use schema::account_balance::dsl::*;
+        dbg!(&start_version_ts);
+        dbg!(&target_version_ts);
```
zizou0x: Need to remove these

src/storage/mod.rs
carloszanella: ok.. in terms of versioning I get it. what if the logic on the contract takes into account the block number? can this not be the case?

src/extractor/evm/utils.rs
kayibal: already resolved in next PR

PR #31, opened 2023-09-15
Extractor and evm entities tests

kayibal:
Adding unit tests for evm entities and extractor.

### DIFF
--- a/src/extractor/evm/ambient.rs
+++ b/src/extractor/evm/ambient.rs
@@ -273,10 +273,8 @@ where
         let msg = match evm::BlockStateChanges::try_from_message(raw_msg, &self.name, self.chain) {
             Ok(changes) => changes,
             Err(ExtractionError::Empty) => {
-                return {
-                    self.update_cursor(inp.cursor).await;
-                    Ok(None)
-                }
+                self.update_cursor(inp.cursor).await;
+                return Ok(None)
             }
             Err(e) => return Err(e),
         };
--- a/src/extractor/evm/mod.rs
+++ b/src/extractor/evm/mod.rs
@@ -45,7 +45,6 @@ pub struct Transaction {
     pub index: u64,
 }
 
-
 impl Transaction {
     pub fn new(hash: H256, block_hash: H256, from: H160, to: Option<H160>, index: u64) -> Self {
         Transaction { hash, block_hash, from, to, index }
@@ -115,7 +114,7 @@ impl From<&AccountUpdateWithTx> for Account {
         Account::new(
             value.chain,
             value.address,
-            format!("0x{:x}", value.address),
+            format!("{:#020x}", value.address),
             value.slots.clone(),
             value.balance.unwrap_or_default(),
             value.code.clone().unwrap_or_default(),
@@ -180,7 +179,7 @@ impl AccountUpdate {
     fn merge(&mut self, other: AccountUpdate) -> Result<(), ExtractionError> {
         if self.address != other.address {
             return Err(ExtractionError::Unknown(format!(
-                "Can't merge AccountUpdates from differing identities; Expected 0x{:x}, got 0x{:x}",
+                "Can't merge AccountUpdates from differing identities; Expected {:#020x}, got {:#020x}",
                 self.address, other.address
             )))
         }
--- a/src/extractor/evm/utils.rs
+++ b/src/extractor/evm/utils.rs
@@ -1,17 +1,52 @@
-use ethers::types::{H160, U256};
+use ethers::types::{H160, H256, U256};
 
-pub fn u256_to_bytes(v: &U256) -> Vec<u8> {
-    let mut bytes32 = [0u8; 32];
-    v.to_big_endian(&mut bytes32);
-    bytes32.to_vec()
+/// Decoding trait with nice contextual error messages
+pub trait TryDecode: Sized {
+    /// Tries to decode a literal value
+    ///
+    /// In case it is not successfull returns a generic `Err(String)` with an error message
+    /// mentioning the target type, as well as the passed type context. This provides a more
+    /// detailed error message compared to ethers-rs "Invalid data" message.
+    ///
+    /// The error is puprosefully kept as string so it can be casted to a more appropriate type on
+    /// the caller.
+    ///
+    /// ## Example
+    /// ```
+    /// let data = [0;32].to_vec();
+    /// let balance = U256::try_decode(data.as_slice(), "balance").unwrap();
+    /// ```
+    fn try_decode(v: &[u8], type_context: &str) -> Result<Self, String>;
 }
 
-/// Parses an evm address hash from a slice
-pub fn parse_id_h160(v: &[u8]) -> Result<H160, String> {
-    if v.len() != 20 {
-        return Err(format!("H160: Invalid address format: 0x{}", hex::encode(v)))
+impl TryDecode for H160 {
+    fn try_decode(v: &[u8], type_context: &str) -> Result<Self, String> {
+        if v.len() != 20 {
+            return Err(format!("H160: Invalid {} format: 0x{}", type_context, hex::encode(v)))
+        }
+        // This is much simpler compared to AbiDecode::decode, the latter will do
+        // some more complex parsing and required all input to be 32 bytes vs
+        // from_slice simply operates on the underlying bytes directly.
+        Ok(H160::from_slice(v))
     }
-    Ok(H160::from_slice(v))
+}
+
+impl TryDecode for H256 {
+    fn try_decode(v: &[u8], type_context: &str) -> Result<Self, String> {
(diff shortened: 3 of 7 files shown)


src/extractor/evm/ambient.rs
```diff
@@ -270,16 +272,20 @@ where
 
         let msg = match evm::BlockStateChanges::try_from_message(raw_msg, &self.name, self.chain) {
             Ok(changes) => changes,
-            Err(ExtractionError::Empty) => return Ok(None),
+            Err(ExtractionError::Empty) => {
+                return {
+                    self.update_cursor(inp.cursor).await;
```
zizou0x: Why is this line in the return block? Does it change anything?

src/extractor/evm/ambient.rs
```diff
@@ -312,3 +318,121 @@ where
...
+    #[tokio::test]
+    async fn test_get_cursor() {
+        let mut gw = MockAmbientGateway::new();
+        gw.expect_get_cursor()
+            .withf(|name, chain| name == "vm:ambient" && chain == &Chain::Ethereum)
+            .times(1)
+            .returning(|_, _| Ok("cursor".into()));
+        let extractor = AmbientContractExtractor::new("vm:ambient", Chain::Ethereum, gw)
+            .await
+            .expect("extractor init ok");
```
pistomat, line 339: Note: I found the "expect as error message" a bit confusing, but then I read https://doc.rust-lang.org/std/error/index.html#common-message-styles and now the "expect as precondition" makes sense to me now. I should update my expect messages in the ws tests.

kayibal: This little paragraph made me get it:

> Hint: If you’re having trouble remembering how to phrase expect-as-precondition style error messages remember to focus on the word “should” as in “env variable should be set by blah” or “the given binary should be available and executable by the current user”.

src/extractor/evm/ambient.rs
kayibal: Mh no I think I should fix that.

src/extractor/evm/ambient.rs
```diff
@@ -312,3 +326,121 @@ where
...
+        let exp = Ok(Some(()));
+
+        let res = extractor
+            .handle_tick_scoped_data(inp)
+            .await
+            .map(|o| o.map(|_| ()));
+
+        assert_eq!(res, exp);
+        assert_eq!(extractor.get_cursor().await, "cursor@420");
+    }
```
carloszanella, line 380: can't review the tests very well as I am unfamiliar with the extractor logic

src/extractor/evm/utils.rs
```diff
@@ -6,17 +6,31 @@ pub fn u256_to_bytes(v: &U256) -> Vec<u8> {
     bytes32.to_vec()
```
pistomat: Out of scope: Is this an alias for [encode()](https://docs.rs/ethers/latest/ethers/core/types/struct.U256.html#method.encode-1)?

src/extractor/evm/utils.rs
```diff
@@ -6,17 +6,31 @@ pub fn u256_to_bytes(v: &U256) -> Vec<u8> {
...
-///
-/// The db id is required to provide additional error context in case the
-/// parsing fails.
+/// Parses an evm address hash from a slice
 pub fn parse_id_h160(v: &[u8]) -> Result<H160, String> {
     if v.len() != 20 {
-        return Err(format!("Invalid contract address found: {}", hex::encode(v)))
+        return Err(format!("H160: Invalid address format: 0x{}", hex::encode(v)))
     }
     Ok(H160::from_slice(v))
```
pistomat: You can just use `.map_err()`, because wrong length is the only way it panics https://docs.rs/ethers/latest/ethers/core/types/struct.H160.html#panics-1

src/extractor/evm/utils.rs
```diff
@@ -6,17 +6,31 @@ pub fn u256_to_bytes(v: &U256) -> Vec<u8> {
...
 }
 
+// Parse an address potentially zero left padding the value
+pub fn pad_and_parse_h160(v: &[u8]) -> Result<H160, String> {
```
pistomat, line 53: Why are we ever using padding? Do we omit leading zeros somewhere for the sake of storage efficiency?

As H160, Address and all related types are fixed length, I would consider any input with length other than 20 an error and try to fix it in the source.

src/extractor/evm/utils.rs
```diff
@@ -45,3 +59,73 @@ pub fn parse_u256_slot_entry(
     let k = U256::from_big_endian(raw_key);
     Ok((k, v))
 }
+
+// Parse a 32 bytes value potentially zero left padding the value
+pub fn pad_and_parse_32bytes<T>(v: &[u8]) -> Result<T, String>
```
pistomat, line 84: Again, I think padding any fixed size types is a mistake and should be resolved at the source of data.

src/extractor/evm/mod.rs
```diff
@@ -93,12 +100,21 @@ impl Account {
...
+    /// missing, it will use the corresponding types default.
+    /// Will use the associated transaction as creation, balance and code modify
+    /// transaction.
     fn from(value: &AccountUpdateWithTx) -> Self {
         let empty_hash = H256::from(keccak256(Vec::new()));
         Account::new(
             value.chain,
             value.address,
-            format!("{:x}", value.address),
+            format!("0x{:x}", value.address),
```
pistomat: I would prefer using [`encode_hex()`](https://docs.rs/ethers/latest/ethers/core/types/struct.H160.html?#method.encode_hex)

src/extractor/evm/mod.rs
```diff
@@ -137,12 +153,35 @@ impl AccountUpdate {
...
+    ///
+    /// * `other`: An instance of `AccountUpdate`. The attribute values and keys
+    /// of `other` will overwrite those of `self`.
     fn merge(&mut self, other: AccountUpdate) -> Result<(), ExtractionError> {
         if self.address != other.address {
-            return Err(ExtractionError::Unknown(
-                "Can't merge AccountUpdates from differing identities".into(),
-            ))
+            return Err(ExtractionError::Unknown(format!(
+                "Can't merge AccountUpdates from differing identities; Expected 0x{:x}, got 0x{:x}",
```
pistomat: Insead of prefixing `0x` manually, you can use the [LowerHex implementation ](https://stackoverflow.com/a/75509511)

src/extractor/evm/utils.rs
kayibal: Oh I didn't know this existed... well would probably be better to use `encode()` here..

src/extractor/evm/utils.rs
kayibal: But it panics so there is no err, right?

src/extractor/evm/utils.rs
kayibal: I think I need these for protobuf... I think this is probably how we get the data from streamingfasts block storage..

src/extractor/evm/utils.rs
kayibal: Will check where these are coming from and if I have control over it fix them, although I think we don't have control over it. 
So I can either do the padding here .. or within the substreams handlers. To reduce bandwith I would probably prefer to do it here.

src/extractor/evm/utils.rs
pistomat: You are right, we should use `.decode()`, which returns an error

src/extractor/evm/utils.rs
kayibal: So I spent some time looking into it and yes these padding methods are used when reading from streamingfasts protobuf message. I think they do not store leading zeros but this is not so easy to find our right away. (I would need to write a spkg to get the blocks - this is useful anyway for testing so I'll do it sometime soon).
TLDR: I'll leave it for now but I agree that it would be preferrable to have as little ambiguity as possible and streamingfast should probably be changed so it has the correct fixed size in case it hasn't.

src/extractor/evm/utils.rs
kayibal: ~~I went looking, couldn't find a `decode` method on H160, nor on &[u8]. But yeah couldn't find a non panicking way to convert a variable length slice to ethers types.~~

EDIT: Found it

src/extractor/evm/mod.rs
kayibal: If I use encode hex I get `"0x000000000000000000000000e688b84b23f322a994a53dbf8e15fa82cdb71127"`

src/extractor/evm/utils.rs
kayibal: Well turns our the abi encode and decode methods are probably not a good idea to use unless you are doing abi encoding decoding - because they assume everything to be 32 bytes including H160.

PR #32, opened 2023-09-15
ENG-1741 Improve error handling

pistomat:
Includes:
- Remove subscriber from subscriptions on SendError instead of panicking
- Add retries to tick processing

### DIFF
--- a/src/extractor/runner.rs
+++ b/src/extractor/runner.rs
@@ -1,8 +1,7 @@
 use anyhow::{format_err, Context, Result};
 use async_trait::async_trait;
-use once_cell::sync::Lazy;
 use prost::Message;
-use std::{collections::HashMap, env, error::Error, sync::Arc, time::Duration};
+use std::{collections::HashMap, env, error::Error, sync::Arc};
 use tokio::{
     sync::{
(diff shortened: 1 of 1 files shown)


src/extractor/runner.rs
```diff
@@ -95,44 +103,65 @@ where
...
-                                match self.extractor.handle_tick_scoped_data(data).await {
-                                    Ok(msg) => {
-                                        if let Some(msg) = msg {
-                                            debug!("Propagating message: {:?}", msg);
-                                            Self::propagate_msg(&self.subscriptions, msg).await
-                                        }
-                                    },
+                                let extractor = self.extractor.clone();
+                                let result = Retry::spawn(DEFAULT_BACKOFF.clone(), || {
+                                    let data = data.clone();
```
kayibal: Hm at least the latter clone here might be expensive.. maybe not worth to have the retry logic in here after all... After reconsidering I think the extractor should retry if it needs to.. I think it might make more sense as it has more context over what failed and can make a more informed retry decision.

Sorry for noticing this soo late!

pistomat: It makes sense, we can handle the errors better in the Extractors, also makes the code clearer here.

PR #34, opened 2023-09-18
Test AmbientGateway

kayibal:
Adds tests for ambient gateway. 

I expect this to evolve into a more general VMExtractorGateway.

Ambient substreams pkg includes information about account creations. We use that instead of hardcoding the deployment transaction hash.

### DIFF
--- a/src/extractor/evm/ambient.rs
+++ b/src/extractor/evm/ambient.rs
@@ -140,13 +140,13 @@ impl AmbientPgGateway {
 
     async fn backward(
         &self,
-        to: BlockIdentifier,
+        to: &BlockIdentifier,
         new_cursor: &str,
         conn: &mut AsyncPgConnection,
     ) -> Result<evm::BlockAccountChanges, StorageError> {
         let block = self
             .state_gateway
-            .get_block(&to, conn)
+            .get_block(to, conn)
             .await?;
         let target = BlockOrTimestamp::Block(to.clone());
         let address = H160(AMBIENT_CONTRACT);
@@ -159,7 +159,7 @@ impl AmbientPgGateway {
             .collect();
 
         self.state_gateway
-            .revert_contract_state(&to, conn)
+            .revert_contract_state(to, conn)
             .await?;
 
         self.save_cursor(new_cursor, conn)
@@ -175,6 +175,8 @@ impl AmbientPgGateway {
         Result::<evm::BlockAccountChanges, StorageError>::Ok(changes)
     }
 
+    // This method is separate so it can be properly tested
+    // see the tests docstring below to understand more.
     async fn get_last_cursor(&self, conn: &mut AsyncPgConnection) -> Result<Vec<u8>, StorageError> {
         let state = self
             .state_gateway
@@ -216,7 +218,7 @@ impl AmbientGateway for AmbientPgGateway {
         let res = conn
             .transaction(|conn| {
                 async move {
-                    self.backward(to, new_cursor, conn)
+                    self.backward(&to, new_cursor, conn)
                         .await
                 }
                 .scope_boxed()
@@ -630,7 +632,7 @@ mod gateway_test {
         let exp_account = ambient_account(0);
 
         let changes = gw
-            .backward(BlockIdentifier::Number((Chain::Ethereum, 0)), "cursor@2", &mut conn)
+            .backward(&BlockIdentifier::Number((Chain::Ethereum, 0)), "cursor@2", &mut conn)
             .await
             .expect("revert should succeed");
 
--- a/src/storage/postgres/contract_state.rs
+++ b/src/storage/postgres/contract_state.rs
@@ -1016,8 +1016,6 @@ where
 
             let tx_hash = delta.tx.as_ref().unwrap();
             let (tx_id, ts) = *txns.get(tx_hash).ok_or_else(|| {
-                dbg!(&tx_hash);
-                dbg!(&txns);
                 StorageError::NoRelatedEntity(
                     "Transaction".to_owned(),
                     "Account".to_owned(),
--- a/substreams/ethereum-ambient/src/lib.rs
+++ b/substreams/ethereum-ambient/src/lib.rs
@@ -49,12 +49,21 @@ impl From<InterimContractChange> for tycho::ContractChange {
 ///
 /// This implementation has currently two major limitations:
 /// 1. It is hardwired to only care about changes to the ambient main contract, this is ok for this
-///    particular use case but for a more general prupose implementation this is not ideal
+///    particular use case but for a more general purpose implementation this is not ideal
 /// 2. Changes are processed separately, this means that if there are any side effects between each
 ///    other (e.g. if account is deleted and then created again in ethereum all the storage is set
 ///    to 0. So there is a side effect between account creation and contract storage.) these might
 ///    not be properly accounted for. Most of the time this should not be a major issue but may lead
-///    to wrong results so consume this implementation with care.
+///    to wrong results so consume this implementation with care. See example below for a concrete
+///    case where this is problematic.
+///
+/// ## A very contrived example:
+/// 1. Some existing contract receives a transaction that changes it state, the state is updated
+/// 2. Next, this contract has self destruct called on itself
+/// 3. The contract is created again using CREATE2 at the same address
+/// 4. The contract receives a transaction that changes it state
+/// 5. We would emit this as as contract creation with slots set from 1 and from 4, although we
+///    should only emit the slots changed from 4.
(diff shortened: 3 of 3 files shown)


substreams/ethereum-ambient/src/pb/mod.rs
```diff
@@ -1,12 +1,10 @@
 // @generated
```
zizou0x, line 1: What is this file, by what is it generated?

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -38,31 +39,43 @@ impl From<InterimContractChange> for tycho::ContractChange {
...
+            change: value.change.into(),
         }
     }
 }
-
+/// Extracts all contract changes relevant to vm simulations
+///
+/// This implementation has currently two major limitations:
+/// 1. It is hardwired to only care about changes to the ambient main contract, this is ok for this
+///    particular use case but for a more general prupose implementation this is not ideal
```
zizou0x:
```suggestion
///    particular use case but for a more general purpose implementation this is not ideal
```

src/storage/postgres/contract_state.rs
```diff
@@ -1018,6 +1016,8 @@ where
 
             let tx_hash = delta.tx.as_ref().unwrap();
             let (tx_id, ts) = *txns.get(tx_hash).ok_or_else(|| {
+                dbg!(&tx_hash);
+                dbg!(&txns);
```
zizou0x: Need to be removed?

substreams/ethereum-ambient/src/pb/tycho.evm.v1.rs
```diff
@@ -62,4 +65,36 @@ pub struct BlockContractChanges {
     #[prost(message, repeated, tag="2")]
     pub changes: ::prost::alloc::vec::Vec<TransactionChanges>,
 }
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord, ::prost::Enumeration)]
+#[repr(i32)]
+pub enum ChangeType {
+    Unspecified = 0,
```
zizou0x, line 71: What can unspecified be?

substreams/ethereum-ambient/src/pb/mod.rs
kayibal: This is generated by `substreams protogen`

substreams/ethereum-ambient/src/pb/tycho.evm.v1.rs
kayibal: This is the default value if the enum field is e.g. not present (it would mean you are parsing an old message and it's most likely a bug).

src/extractor/evm/ambient.rs
```diff
@@ -140,18 +137,59 @@ impl AmbientPgGateway {
...
+            chain: self.chain,
+            extractor: self.name.clone(),
+            block,
+            account_updates,
+            new_pools: HashMap::new(),
+        };
+        Result::<evm::BlockAccountChanges, StorageError>::Ok(changes)
+    }
+
+    async fn get_last_cursor(&self, conn: &mut AsyncPgConnection) -> Result<Vec<u8>, StorageError> {
```
pistomat, line 180: This is extracted because of testing, right? Maybe add a comment explaining it.

src/extractor/evm/ambient.rs
```diff
@@ -140,18 +137,59 @@ impl AmbientPgGateway {
...
     }
-}
 
-#[async_trait]
-impl AmbientGateway for AmbientPgGateway {
-    async fn get_cursor(&self, name: &str, chain: Chain) -> Result<Vec<u8>, StorageError> {
-        let mut conn = self.pool.get().await.unwrap();
+    async fn backward(
+        &self,
+        to: BlockIdentifier,
```
pistomat: Make this a reference `to: &BlockIdentifier`

src/storage/postgres/contract_state.rs
pistomat: Is this really resolved?

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -156,6 +176,10 @@ fn map_changes(
                             balance: new_balance.bytes.clone(),
                             code: Vec::new(),
                             slots: HashMap::new(),
+                            change: created_accounts
```
pistomat, line 188: This is obfuscated if-else https://rust-lang.github.io/rust-clippy/master/index.html#/obfuscated_if_else

`change: if created_accounts.contains_key(&balance_change.address) { ChangeType::Creation } else { ChangeType::Update },`

It is hinted to me inline in VSCode, but not using the clippy command, I am not sure why...

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -120,6 +136,10 @@ fn map_changes(
                         balance: Vec::new(),
                         code: Vec::new(),
                         slots,
+                        change: created_accounts
+                            .contains_key(&storage_change.address)
+                            .then_some(ChangeType::Creation)
+                            .unwrap_or(ChangeType::Update),
```
pistomat, line 151: This is obfuscated if-else https://rust-lang.github.io/rust-clippy/master/index.html#/obfuscated_if_else

`change: if created_accounts.contains_key(&storage_change.address) { ChangeType::Creation } else { ChangeType::Update },`

It is hinted to me inline in VSCode, but not using the clippy command, I am not sure why...

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -190,6 +214,10 @@ fn map_changes(
                         balance: Vec::new(),
                         code: code_change.new_code.clone(),
                         slots: HashMap::new(),
+                        change: created_accounts
```
pistomat, line 226: This is obfuscated if-else https://rust-lang.github.io/rust-clippy/master/index.html#/obfuscated_if_else

change: if created_accounts.contains_key(&code_change.address) { ChangeType::Creation } else { ChangeType::Update },

It is hinted to me inline in VSCode, but not using the clippy command, I am not sure why...

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -38,31 +39,43 @@ impl From<InterimContractChange> for tycho::ContractChange {
...
         }
     }
 }
-
+/// Extracts all contract changes relevant to vm simulations
+///
+/// This implementation has currently two major limitations:
+/// 1. It is hardwired to only care about changes to the ambient main contract, this is ok for this
+///    particular use case but for a more general prupose implementation this is not ideal
+/// 2. Changes are processed separately, this means that if there are any side effects between each
```
pistomat, line 53: For the example of deleting and creating a contract where all storage is set to 0 that you give, can you explain to me what should happen and doesn't in this function? I am trying to understand this thoroughly.

substreams/ethereum-ambient/src/lib.rs
kayibal: I think I fixed that in my next PR already :)

substreams/ethereum-ambient/src/lib.rs
kayibal: So imagine this very contrived example:

1. Some existing contract receives a transaction that changes it state, the state is updated
2. Next, this contract has self destruct called on itself
3. The contract is created again using CREATE2 at the same address
4. The contract receives a transaction that changes it state

We would emit this as as contract creation with slots set from 1 and from 4, although we should only emit the slots changed from 4. 

That said, I am not 100% sure destroying and recreating at same address is possible and that i behaves like this, but some short googling supported it. Also this would all need to happen within a single block for it to be problematic.

pistomat: Okay, so the solution would be to check the updates not only within the same transaction as noted [here](https://github.com/propeller-heads/tycho-indexer/blob/main/substreams/ethereum-ambient/src/lib.rs#L80-L83) but also across the transactions within the block, right.

Note that [SELFDESTRUCT only in the same block](https://eips.ethereum.org/EIPS/eip-6780) is actually the latest proposal from Vitalik on how to handle this opcode, which would mean the code would not change between blocks.

kayibal: Not sure if that is the solution, I think if you find a SELFDESTRUCT, you basically want to set all storage slots to 0, but within substreams handlers, you only know the slots that have been touched so far in that block... 

So you would need to emit a special message to Tycho indicating that all slots should be reset... for that contract and do not emit any updates that have a lower ordinal then the SELFDESTRUCT call. So yeah quite complex..

PR #35, opened 2023-09-18
ENG-1747: Implement RPC endpoint

pistomat:
Updated the RPC logic to use get_contracts, added a new `/contract_state` route and added a test for the RPC server.

Next step is to start the RPC and WS server in the `main.rs` and properly test the whole integration with extrators.

### DIFF
--- a/src/services/rpc.rs
+++ b/src/services/rpc.rs
@@ -330,8 +330,7 @@ mod tests {
         acc_address.to_string()
     }
 
-    #[tokio::test]
-    async fn test_get_state() {
+    async fn setup_database() -> (AsyncPgConnection, String) {
         let db_url = std::env::var("DATABASE_URL").unwrap();
         let mut conn = AsyncPgConnection::establish(&db_url)
             .await
@@ -341,14 +340,21 @@ mod tests {
             .unwrap();
         let acc_address = setup_account(&mut conn).await;
 
-        let db_gtw = PostgresGateway::<
+        (conn, acc_address)
+    }
+
+    #[tokio::test]
+    async fn test_get_state() {
+        let (mut conn, acc_address) = setup_database().await;
+
+        let db_gw = PostgresGateway::<
             evm::Block,
             evm::Transaction,
             evm::Account,
             evm::AccountUpdate,
         >::from_connection(&mut conn)
         .await;
-        let req_handler = RequestHandler::new(db_gtw, conn);
+        let req_handler = RequestHandler::new(db_gw, conn);
 
         let code = hex::decode("1234").unwrap();
         let code_hash = H256::from_slice(&ethers::utils::keccak256(&code));
@@ -392,23 +398,16 @@ mod tests {
 
     #[actix_web::test]
     async fn test_contract_state_route() {
-        let db_url = std::env::var("DATABASE_URL").unwrap();
-        let mut conn = AsyncPgConnection::establish(&db_url)
-            .await
-            .unwrap();
-        conn.begin_test_transaction()
-            .await
-            .unwrap();
-        let acc_address = setup_account(&mut conn).await;
+        let (mut conn, acc_address) = setup_database().await;
 
-        let db_gtw = PostgresGateway::<
+        let db_gw = PostgresGateway::<
             evm::Block,
             evm::Transaction,
             evm::Account,
             evm::AccountUpdate,
         >::from_connection(&mut conn)
         .await;
-        let req_handler = RequestHandler::new(db_gtw, conn);
+        let req_handler = RequestHandler::new(db_gw, conn);
 
         // Set up the expected account data
         let code = hex::decode("1234").unwrap();


src/services/serde_hex.rs
```diff
@@ -0,0 +1,23 @@
+use serde::{
```
zizou0x: Isn't this file too targeted? What else do we plan to have here? Wouldn't it make more sense to have a `serde_helpers.rs` with this and some others serde helpers?

src/services/mod.rs
```diff
@@ -1,5 +1,5 @@
 //! This module contains Tycho web services implementation
 
-pub mod deserialization_helpers;
 mod rpc;
+pub mod serde_hex;
```
kayibal: Should this be public?

src/services/rpc.rs
```diff
@@ -1,80 +1,138 @@
...
+    fn new(accounts: Vec<Account>) -> Self {
+        Self { accounts }
+    }
+}
+
 #[derive(Error, Debug)]
 pub enum RpcError {
     #[error("Failed to parse JSON: {0}")]
-    ParseError(serde_json::Error),
+    ParseError(#[from] serde_json::Error),
```
zizou0x, line 110: What does `#[from]` do?

src/services/rpc.rs
```diff
@@ -1,80 +1,138 @@
...
 use thiserror::Error;
+use tokio::sync::Mutex;
 use tracing::error;
 
+use super::serde_hex::{deserialize_hex, serialize_hex};
+
 struct RequestHandler {
     db_gw: PostgresGateway<evm::Block, evm::Transaction, evm::Account, evm::AccountUpdate>,
-    db_connection: AsyncPgConnection,
+    db_connection: Arc<Mutex<AsyncPgConnection>>,
```
kayibal, line 24: Would be probably better to use a Pool here, the pool get can retrieve connection reference, which dereferences into AsyncPgConnection. We construct a pool at startup, so I could probably pass it down here. But not super important we can also take out a dedicated connection from the pool that we pass down here..

src/services/rpc.rs
zizou0x: Ah ok I guess it implements the from trait?

src/services/rpc.rs
```diff
@@ -1,80 +1,138 @@
...
+        let mut db_connection_lock = self.db_connection.lock().await;
+
+        // Get the contract IDs from the request
+        let contract_ids = request.contract_ids.clone();
+        let address_slices: Option<Vec<&[u8]>> = contract_ids.as_ref().map(|ids| {
+            ids.iter()
+                .map(|id| id.address.as_slice())
+                .collect()
+        });
+        let addresses: Option<&[&[u8]]> = address_slices.as_deref();
```
kayibal, line 57: I think we should also support settings this to None?

src/services/rpc.rs
```diff
@@ -1,80 +1,138 @@
...
+    // Call the handler to get the state
+    let response = handler
+        .into_inner()
+        .get_state(&body, &query)
+        .await;
+
+    match response {
+        Ok(state) => HttpResponse::Ok().json(state),
+        Err(e) => {
+            error!("Error while getting contract state: {}", e);
```
kayibal: Could you also log the request here?

src/services/serde_hex.rs
```diff
@@ -0,0 +1,23 @@
...
+    de::{self, Deserializer},
+    Deserialize, Serializer,
+};
+
+/// Serializes a Vec<u8> as a hex string with '0x' prefix.
+pub fn serialize_hex<S>(bytes: &Vec<u8>, serializer: S) -> Result<S::Ok, S::Error>
```
pistomat: An alternative is to use the serde_hex crate, but I found it unnecessary at the moment.

src/services/rpc.rs
```diff
@@ -1,80 +1,138 @@
...
-                    .db_gw
-                    .get_contract(contract_id, &Some(&version), false, &mut self.db_connection)
-                    .await
-                {
-                    Ok(contract_state) => accounts.push(contract_state),
-                    Err(e) => {
-                        error!("Error while getting contract state {}", e)
-                    }
-                }
+        let mut db_connection_lock = self.db_connection.lock().await;
```
zizou0x, line 48: That means it can't process requests in parallel right?

src/services/mod.rs
pistomat: The `ContractId` is defined in mod `storage` and we need to de*serialize the `Address` part too.

It would be better to extract all types to a top-level `primitives` mod. And/or move the `serde_hex` one level up.

src/services/rpc.rs
pistomat: It does support setting it to `None`. All of this is just gymnastics to cast an `Option<Vec<ContractId>>` to `Option<&[&[u8]]>`.

src/services/rpc.rs
pistomat: Yes, I was unaware of the better solution of using the fully async and thread-safe bb8 connection pool, which is used elsewhere in the code.

src/services/serde_hex.rs
zizou0x: Yes, if it's just for this I think it's not worth it to use the crate. If we have some other possible usage for it then I would consider switching

src/services/rpc.rs
pistomat: Agree, although using Pool here makes it really hard to test, because you use `begin_test_transaction()` and setup the fixture data on the whole pool. So I lean towards doing it later in a dedicated task.

PR #36, opened 2023-09-18
Preparations for initial release

kayibal:
This branch prepares the repo for the first initial release:

- Added a service builder to run the websocket server.
- Proper task nannying, we exit if any extractor or the http servers task completes. Had to move the JoinHandle out of ExtractorHandle for this.
- Created an outer virtual workspace so everything is handled together.
- Include diesel migrations with the binary so they can be executed on application startup if necessary
- Added a build script (Unfortunately we can't sue the workspace alone for building yet on stable)
- Added a dockerfile & a docker build GH Action.

TODO:

- [x] Add this repo's permissions correctly on AWS
- [x] Add ECR for tycho-indexer
- [x] Make substream_ethereum_ambient conform to our practices (clippy & fmt)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -63,13 +63,17 @@ export SUBSTREAMS_API_TOKEN=$(curl https://auth.streamingfast.io/v1/auth/issue -
 ```bash
 rustup target add wasm32-unknown-unknown
 ```
-5. To compile protobuf files please install [buf](https://docs.buf.build/installation)
-6. To compile the substreams, run:
+5. To compile protobuf files please install [buf](https://docs.buf.build/installation). To generate protobuf definition for a substream crate:
 ```bash
-cd substreams/ethereum-ambient 
-cargo build --release --target wasm32-unknown-unknown
-substreams pack ./substreams.yaml 
+cd substreams/ethereum-abmient
+substreams protogen substreams.yaml --exclude-paths="sf/substreams,google"
 ```
+6. To compile the substreams into a spkgs, run:
+```bash
+cargo build --package substreams-ethereum-ambient --target wasm32-unknown-unknown --profile substreams
+substreams pack substreams/ethereum-ambient/substreams.yaml
+```
+7. Alternatively you can run the `./stable-build.sh` script which will build all substreams crates and pack them. The spkgs artifacts will be placed under `./target`.
 7. To test the substreams (requires the SUBSTREAMS_API_TOKEN env variable set previously), run:
 ```bash
 cd ../..
@@ -113,6 +117,7 @@ export DATABASE_URL=postgres://postgres:mypassword@localhost:5432/tycho_indexer_
 diesel migration run
 ```
 6. We use [pgFormatter](https://github.com/darold/pgFormatter) to keep SQL files consistently formatted.
+
 ### Local development
 1. Please also make sure that the following commands pass if you have changed the code:
 


.dockerignore
```diff
@@ -0,0 +1,11 @@
...
+target
+**/target
+*/target/
+.github
+venv
+*/venv
+.idea
+.git
+**/*.spkg
+Dockerfile
```
pistomat, line 11: Can you also add `.env`?

Cargo.lock
pistomat, line 1: `./Cargo.lock` is in `.gitignore`, but is included in the repo. Either remove it from `.gitignore` or untrack this file.

Cargo.toml
pistomat, line 1: I think this is a good point, where we should define packages used in all members and then reuse them in submodules in order to use the same version of packages all across the workspace.

This is a good example: https://github.com/paradigmxyz/reth/blob/main/Cargo.toml

tycho-indexer/src/extractor/runner.rs
```diff
@@ -91,15 +93,16 @@ where
...
         tokio::spawn(async move {
+            let id = self.extractor.get_id();
             loop {
                 tokio::select! {
                     Some(ctrl) = self.control_rx.recv() =>  {
                         match ctrl {
                             ControlMessage::Stop => {
-                                warn!("Stopping stream.");
-                                break;
+                                warn!("{}: Stop signal received; exiting!", &id);
```
pistomat, line 104: The tracing library is a superset of logs, so you can actually write:
`warn!(?id, "Stop signal received; exiting!");`
which produces structured logs that are easier to filter and work with.

More info in [docs](https://docs.rs/tracing/latest/tracing/index.html#recording-fields)

tycho-indexer/src/main.rs
```diff
@@ -1,6 +1,5 @@
 #![allow(dead_code)] // FIXME: remove after initial development
```
pistomat: This should be removed now that we are doing an initial release.

tycho-indexer/src/services/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    pub fn run(
+        self,
+    ) -> Result<(ServerHandle, JoinHandle<Result<(), ExtractionError>>), ExtractionError> {
+        let app_state = web::Data::new(AppState::<M>::new(self.extractor_handles));
+        let server = HttpServer::new(move || {
+            App::new()
+                .app_data(app_state.clone())
+                .service(
```
pistomat, line 70: Add also the RPC service

tycho-indexer/src/services/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+use self::ws::{AppState, MessageSenderMap, WsActor};
+
+pub mod deserialization_helpers;
+mod rpc;
+mod ws;
+
+pub struct ServicesBuilder<M> {
+    prefix: String,
+    port: u16,
+    bind: String,
```
pistomat, line 21: I would replace this with 
`ip: IpAddr`
to ensure proper values.

tycho-indexer/src/services/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+            .insert(id, Arc::new(handle));
+        self
+    }
+
+    pub fn prefix(mut self, v: &str) -> Self {
+        self.prefix = v.to_owned();
+        self
+    }
+
+    pub fn bind(mut self, v: &str) -> Self {
```
pistomat, line 53: If the variable type would be IpAddr, this would turn into
`pub fn bind(mut self, v: Into<IpAddr>) -> Self`

tycho-indexer/src/services/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+
+pub struct ServicesBuilder<M> {
+    prefix: String,
+    port: u16,
+    bind: String,
+    extractor_handles: MessageSenderMap<M>,
+}
+
+impl<M: NormalisedMessage> ServicesBuilder<M> {
+    pub fn new() -> Self {
```
pistomat: I seems more like a `default()` function
```
impl<M: NormalisedMessage> Default for ServicesBuilder<M> {
    fn default() -> Self {
        Self {
            prefix: "v1".to_owned(),
            port: 4242,
            ip: "0.0.0.0".parse().expect("Invalid IP address"),
            extractor_handles: HashMap::new(),
        }
    }
}
```

This function then would become 
```
pub fn new() -> Self {
    Self::default()
}
```

tycho-indexer/src/storage/postgres/mod.rs
```diff
@@ -391,7 +395,7 @@ pub async fn connect(db_url: &str) -> Result<Pool<AsyncPgConnection>, StorageErr
 /// - If it failed to get a connection from the provided pool.
 /// - If there was an issue ensuring the presence of chains in the database.
 pub async fn ensure_chains(chains: &[Chain], pool: Pool<AsyncPgConnection>) {
-    info!("Ensured chain enum presence for: {:?}", chains);
+    debug!("Ensured chain enum presence for: {:?}", chains);
```
pistomat: This should move after the chain enums are ensured

tycho-indexer/src/substreams/stream.rs
```diff
@@ -63,15 +63,18 @@ fn stream_blocks(
...
 
     try_stream! {
         'retry_loop: loop {
-            info!("Blockstreams disconnected, connecting (endpoint {}, start block {}, cursor {})",
-                &endpoint,
-                start_block_num,
-                &latest_cursor
-            );
+            if retry_count > 0 {
+                warn!("Blockstreams disconnected, connecting (endpoint {}, start block {}, cursor {})",
```
pistomat: Log the `retry_count`

stable-build.sh
pistomat, line 1: This build script is great.

Can you update the Readme.md to follow this?

tycho-indexer/src/main.rs
pistomat, line 1: There are two env variables `DB_URL` and `DATABASE_URL`, we should stick to one and remove the other one.

tycho-indexer/src/extractor/runner.rs
kayibal: Let's do that in a separate PR, I only read about it after doing this work but totally agree that it will be much nicer by using structured logs.

tycho-indexer/src/main.rs
kayibal: I actually separated those on purpose the `DATABASE_URL` is what diesel uses and what we use for testing. You don't want to mix it with DB_URL which is what your executable will use. 
E.g. I have those configured to use different databases, `tycho_0` for diesel and `tycho_1` for running the executable. The `tycho_0` db is always empty and only used in unittests and `tycho_1` will actually contain data of local runs.
I will add this to the readme - maybe I'll also rename `DB_URL` to sth like `TYCHO_DB_CONNECTION`. But we should do that work once we have things a bit further developed, I think the CLI interface will still have to change a lot.

Cargo.toml
kayibal: I am aware we can do it. Not sure it is necessary yet. I prefer things to be more free first and later if they need to be more restricted we can do it. 
For Reth it makes sense, because it imports the subcrates as libraries (probably for faster compiles?) but we use it more like a plugin so Tycho is just interested in the compiled binary and ist not looking to interface with it in any way. In case we factor out a library that we depend on, I think it would sense.

tycho-indexer/src/services/mod.rs
kayibal: Unfortunately I think we'll need a separate PR here where we merge the websocket AppData struct with the RequestHandler struct.

tycho-indexer/src/services/mod.rs
kayibal: I don't have the capacity to do that atm sorry. I want to get this merged quick cause there were a lot of files moved. Happy to come back to this later.

PR #39, opened 2023-09-22
ENG-1923: Add RPC to tycho service builder

pistomat:


tycho-indexer/src/main.rs
```diff
@@ -44,24 +44,47 @@ extern crate pretty_assertions;
...
+    #[clap(long)]
+    spkg: String,
 
     /// Substreams Module name
-    #[clap(name = "module", short, long)]
-    module_name: String,
+    #[clap(long)]
+    module: String,
+
+    /// Substreams start block
```
pistomat, line 68: It seemed like a good idea to add this as an argument, better than having magic values in the code and changing them for each run for testing purposes (and having to rebuild the package...).

tycho-indexer/src/services/mod.rs
```diff
@@ -63,14 +68,22 @@ impl<M: NormalisedMessage> ServicesBuilder<M> {
...
         self,
     ) -> Result<(ServerHandle, JoinHandle<Result<(), ExtractionError>>), ExtractionError> {
-        let app_state = web::Data::new(AppState::<M>::new(self.extractor_handles));
+        let ws_data = web::Data::new(ws::WsData::<M>::new(self.extractor_handles));
+        let rpc_data =
+            web::Data::new(rpc::RpcHandler::new(self.db_gateway, self.db_connection_pool));
         let server = HttpServer::new(move || {
             App::new()
-                .app_data(app_state.clone())
+                .app_data(rpc_data.clone())
```
pistomat, line 76: Added the RPC service here, the actix-web allows me to have the AppData and logic neatly separated, which is cool.

tycho-indexer/src/services/rpc.rs
```diff
@@ -3,40 +3,54 @@
...
 }
 
-impl RequestHandler {
+impl RpcHandler {
     #[allow(dead_code)]
     pub fn new(
-        db_gw: PostgresGateway<evm::Block, evm::Transaction, evm::Account, evm::AccountUpdate>,
-        db_connection: AsyncPgConnection,
+        db_gateway: Arc<EvmPostgresGateway>,
+        db_connection_pool: Pool<AsyncPgConnection>,
```
pistomat, line 34: I replaced the single connection behind a mutex with the connection pool as we discussed previously.

tycho-indexer/src/services/rpc.rs
```diff
@@ -391,81 +395,35 @@ mod tests {
...
             .await
             .unwrap();
 
         assert_eq!(state.accounts.len(), 1);
         assert_eq!(state.accounts[0], expected);
     }
 
-    #[actix_web::test]
-
-    async fn test_contract_state_route() {
```
pistomat, line 404: I had to remove the route test, because I did not find and easy way to create a test connection on the `Pool`. But the logic is still tested and the routing is standard `actix-web` functionality, so I don't mind the tradeoff.

tycho-indexer/src/main.rs
```diff
@@ -44,24 +44,47 @@ extern crate pretty_assertions;
 )]
 struct Args {
```
zizou0x: FMI, is it a common Rust practice to call a struct `Args`? I would specify what args or what it is used for in the name here to make it clearer

tycho-indexer/src/main.rs
```diff
@@ -44,24 +44,47 @@ extern crate pretty_assertions;
...
+            let increment: i64 = self.stop_block[1..]
+                .parse()
+                .expect("Invalid increment value");
+            self.start_block + increment
+        } else {
+            self.stop_block
+                .parse()
+                .expect("Invalid stop block value")
+        }
+    }
```
zizou0x: Nice, do we have a doc somewhere specifying that stop_block can either be a number or an increment value?

tycho-indexer/src/services/rpc.rs
zizou0x: Nice, thanks

tycho-indexer/src/main.rs
pistomat: It seems like it, but quite often with some prefix. I will rename it to `CliArgs`

tycho-indexer/src/services/mod.rs
kayibal: Nice good to know!

tycho-indexer/src/services/mod.rs
```diff
@@ -63,14 +68,22 @@ impl<M: NormalisedMessage> ServicesBuilder<M> {
...
+                        .route(web::post().to(rpc::contract_state)),
+                )
+                .app_data(ws_data.clone())
                 .service(
-                    web::resource(format!("/{}/ws/", self.prefix))
-                        .route(web::get().to(WsActor::<M>::ws_index)),
+                    web::resource(format!("/{}/ws", self.prefix))
+                        .route(web::get().to(ws::WsActor::<M>::ws_index)),
                 )
+                .wrap(RequestTracing::new())
```
kayibal, line 86: This enables tracing for both the RPC and WS?

tycho-indexer/src/services/rpc.rs
kayibal: Yeah mocking the pools is kind of impossible. Which most of the times turns out ok anyway.

tycho-indexer/src/services/mod.rs
pistomat: It adds a tracing wrapper in the actix Middleware, so I think it should add tracing to both, but tbh I am not sure. There is work left to do on exporting the tracing to our monitoring services etc. so I postponed the testing and polishing to then.

PR #40, opened 2023-09-25
ENG-1740: Test ExtractorRunnerBuilder

pistomat:


tycho-indexer/src/extractor/mod.rs
```diff
@@ -30,10 +31,10 @@ pub enum ExtractionError {
     ServiceError(String),
 }
 
+#[automock]
 #[async_trait]
-pub trait Extractor<G, M>: Send + Sync
+pub trait Extractor<M>: Send + Sync
 where
-    G: Send + Sync,
```
zizou0x, line 36: So this wasn't necessary?

pistomat: It used to be necessary, but I think Alan refactored it so it is gateway agnostic and it was just leftover.

PR #41, opened 2023-09-26
ENG-1993: Refactor Tycho types

pistomat:
Replace the `Vec<u8>` as various blockchain types with a custom wrapper of `bytes::Bytes` with implementation of various helpes, serde functionality and diesel ORM representation.

Also fixed some parameter referencing.

### DIFF
--- a/tycho-indexer/src/hex_bytes.rs
+++ b/tycho-indexer/src/hex_bytes.rs
@@ -354,12 +354,12 @@ mod tests {
         let example_bytes = Bytes::from_str("0x0123456789abcdef").unwrap();
 
         conn.batch_execute(
-            r#"
+            r"
             CREATE TEMPORARY TABLE bytes_table (
                 id SERIAL PRIMARY KEY,
                 data BYTEA NOT NULL
             );
-        "#,
+        ",
         )
         .await
         .unwrap();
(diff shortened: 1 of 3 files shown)


tycho-indexer/src/storage/postgres/contract_state.rs
```diff
@@ -509,14 +506,15 @@ where
             .await?
             .into_iter()
             .collect();
+        #[allow(clippy::mutable_key_type)]
```
pistomat, line 509: `bytes::Bytes` is by default ignored, as it does not really allow mutation, but we use a wrapper around it and also this is actually a reference `HashSet<&Bytes>`, which clippy somehow does not pick up as being ignored.

kayibal: What is it complaining about? Bytes being used as key, although it is mutable?

pistomat: Yes, it is bad practice to use internally mutable types as keys HashSet/HashMaps, because if you mutate them, their hash changes and it breaks the container.

If I understand it correctly, `bytes::Bytes` uses internally an `AtomicPtr` which although it is mutable, is is not used to mutate the content. There is a mutable version `bytes::BytesMut` that allows it, but we don't use that.

The `bytes::Bytes` is default as being ignored by this lint, which also works on `&bytes::Bytes`. Because our `hex_bytes::Bytes` is only a wrapper around it, I also added it to the ignore list, but somehow a reference is still being triggered, so I added the allow macro here as well.

PR #51, opened 2023-11-13
Fix ids conflicts

zizou0x:
Issue: Previously, the system assigned IDs to new subscriptions depending on length of current subscriptions. However, a conflict arose when a subscription was removed. For example, with initial subscriptions IDs [0, 1, 2], after removing ID 0, the system would incorrectly attempt to assign ID 2 to the next subscription, creating a conflict as ID 2 already exists.

Fix: This pull request introduces a mechanism to ensure that new subscriptions are always assigned a unique ID, one that has not been used previously, even if earlier IDs have been removed. This change prevents conflicts and ensures the integrity of subscription ID assignment.

### DIFF
--- a/tycho-indexer/src/extractor/runner.rs
+++ b/tycho-indexer/src/extractor/runner.rs
@@ -87,6 +87,7 @@ pub struct ExtractorRunner<M> {
     extractor: Arc<dyn Extractor<M>>,
     substreams: SubstreamsStream,
     subscriptions: Arc<Mutex<SubscriptionsMap<M>>>,
+    next_subscriber_id: u64,
     control_rx: Receiver<ControlMessage<M>>,
 }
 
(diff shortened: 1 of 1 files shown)


tycho-indexer/src/extractor/runner.rs
```diff
@@ -166,9 +175,10 @@ where
 
     #[instrument(skip_all)]
     async fn subscribe(&mut self, sender: Sender<Arc<M>>) {
-        let subscriber_id = self.subscriptions.lock().await.len() as u64;
+        let subscriber_id = self.next_subscriber_id;
```
pistomat, line 178: Oh I am so sorry this is the cause, I wanted to change this to `Uuid`, which would fix it (although I didn't know of the bug) but didn't.

zizou0x: Of course, there's no problem. To be honest, it was difficult to foresee

PR #52, opened 2023-11-14
Improve get contract RPC request

zizou0x:
This PR aims to make `/contract_state` easier. We currently require all of these fields to be able to make a request.

```
"version": {
            "timestamp": "2069-01-01T04:20:00",
            "block": {
                "hash": "0x24101f9cb26cd09425b52da10e8c2f56ede94089a8bbe0f31f1cda5f4daa52c4",
                "parentHash": "0x8d75152454e60413efe758cc424bfd339897062d7e658f302765eb7b50971815",
                "number": 213,
                "chain": "ethereum"
            }
        }
```

But we only partially use them. This PR change the logic to be able to get contract state with only the minimal information. For example, only the timestamp or only the hash of the target block. Here are some example of what we should be able to do:

```
"version": {
            "block": {
                "hash": "0x24101f9cb26cd09425b52da10e8c2f56ede94089a8bbe0f31f1cda5f4daa52c4",
            }
        }
```

```
"version": {
            "block": {
                "number": 213,
                "chain": "ethereum"
            }
        }
```

```
"version": {
            "timestamp": "2069-01-01T04:20:00",
        }
```

Note: It shouldn't break the current request structure.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -167,7 +167,7 @@ cargo install diesel_cli --no-default-features --features postgres
 ```
 3. Start up a the postgres db service
 ```bash
-docker-compose up -d
+docker-compose up -d db
 ```
 4. Set the env var for the DB connection:
 ```
--- a/tycho-indexer/migrations/2023-08-15-154244_tycho_initial_setup/up.sql
+++ b/tycho-indexer/migrations/2023-08-15-154244_tycho_initial_setup/up.sql
@@ -471,6 +471,17 @@ CREATE OR REPLACE FUNCTION invalidate_previous_entry_contract_storage()
     RETURNS TRIGGER
     AS $$
 BEGIN
+    -- Get previous value from latest storage entry.
+    NEW.previous_value =(
+        SELECT
+            value
+        FROM
+            contract_storage
+        WHERE
+            valid_to IS NULL
+            AND account_id = NEW.account_id
+            AND slot = NEW.slot
+        LIMIT 1);
     -- Update the 'valid_to' field of the last valid entry when a new one is inserted.
     UPDATE
         contract_storage
@@ -480,15 +491,6 @@ BEGIN
         valid_to IS NULL
         AND account_id = NEW.account_id
         AND slot = NEW.slot;
-    NEW.previous_value =(
-        SELECT
-            value
-        FROM
-            contract_storage
-        WHERE
-            account_id = NEW.account_id
-            AND slot = NEW.slot
-        LIMIT 1);
     RETURN NEW;
 END;
 $$
--- a/tycho-indexer/src/storage/postgres/contract_state.rs
+++ b/tycho-indexer/src/storage/postgres/contract_state.rs
@@ -550,6 +550,7 @@ where
                 }
             }
         }
+        new_entries.sort_by_key(|k| k.ordinal);
         diesel::insert_into(schema::contract_storage::table)
             .values(&new_entries)
             .execute(conn)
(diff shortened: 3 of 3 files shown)


tycho-indexer/src/services/rpc.rs
```diff
@@ -114,66 +161,26 @@ pub async fn contract_state(
...
-struct Version {
-    timestamp: NaiveDateTime,
-    block: Option<Block>,
-}
-
-impl Default for Version {
-    fn default() -> Self {
-        Version { timestamp: Utc::now().naive_utc(), block: None }
+fn validate_version(version: &Version) -> Result<BlockOrTimestamp, RpcError> {
+    match (&version.timestamp, &version.block) {
```
pistomat: Nice match pattern!

tycho-indexer/src/services/rpc.rs
```diff
@@ -114,66 +161,26 @@ pub async fn contract_state(
...
+        (_, Some(block)) => {
+            // If a full block is provided, we prioritize hash over number and chain
+            let block_identifier = if let Some(hash) = &block.hash {
+                BlockIdentifier::Hash(hash.clone())
+            } else if let (Some(chain), Some(number)) = (&block.chain, &block.number) {
+                BlockIdentifier::Number((*chain, *number))
+            } else {
+                return Err(RpcError::Parse(DeError::custom("Insufficient block information")));
+            };
```
pistomat: This can also be rewritten using match, right?
```suggestion
let block_identifier = match (&block.hash, &block.chain, &block.number) {
    (Some(hash), _, _) => BlockIdentifier::Hash(hash.clone()),
    (_, Some(chain), Some(number)) => BlockIdentifier::Number((*chain, *number)),
    _ => return Err(RpcError::Parse(DeError::custom("Insufficient block information"))),
};
```

tycho-indexer/src/services/rpc.rs
```diff
@@ -114,66 +161,26 @@ pub async fn contract_state(
...
+            Ok(BlockOrTimestamp::Block(block_identifier))
+        }
+        (Some(timestamp), None) => Ok(BlockOrTimestamp::Timestamp(*timestamp)),
+        (None, None) => {
+            Err(RpcError::Parse(DeError::custom("Missing timestamp or block identifier")))
+        }
     }
 }
```
pistomat: This function does not only validate, but rather parse Version to BlockOrTimestamp, is there a reason you don't want it implemented as TryFrom?

```suggestion
/// Attempts to convert a `Version` into a `BlockOrTimestamp`.
/// If `version` contains a block, it tries to create a `BlockOrTimestamp` based on the block's hash or its chain and number.
/// If only a timestamp is provided in `version`, it creates a `BlockOrTimestamp` based on the timestamp.
/// Returns an error if neither sufficient block information nor a timestamp is provided.
impl TryFrom<Version> for BlockOrTimestamp {
    type Error = RpcError;

    fn try_from(version: Version) -> Result<Self, Self::Error> {
        match (version.timestamp, version.block) {
            (_, Some(block)) => {
                // If a full block is provided, we prioritize hash over number and chain
                let block_identifier = match (&block.hash, &block.chain, &block.number) {
                    (Some(hash), _, _) => BlockIdentifier::Hash(hash.clone()),
                    (_, Some(chain), Some(number)) => BlockIdentifier::Number((*chain, *number)),
                    _ => return Err(RpcError::Parse(DeError::custom("Insufficient block information"))),
                };
                Ok(BlockOrTimestamp::Block(block_identifier))
            }
            (Some(timestamp), None) => Ok(BlockOrTimestamp::Timestamp(timestamp)),
            (None, None) => Err(RpcError::Parse(DeError::custom("Missing timestamp or block identifier"))),
        }
    }
}
```

tycho-indexer/src/services/rpc.rs
```diff
@@ -114,66 +161,26 @@ pub async fn contract_state(
...
+fn validate_version(version: &Version) -> Result<BlockOrTimestamp, RpcError> {
+    match (&version.timestamp, &version.block) {
+        (_, Some(block)) => {
+            // If a full block is provided, we prioritize hash over number and chain
+            let block_identifier = if let Some(hash) = &block.hash {
+                BlockIdentifier::Hash(hash.clone())
+            } else if let (Some(chain), Some(number)) = (&block.chain, &block.number) {
+                BlockIdentifier::Number((*chain, *number))
+            } else {
+                return Err(RpcError::Parse(DeError::custom("Insufficient block information")));
```
pistomat: Throwing serde errors manually is confusing to me. This makes me think we are deserializing some json here, which is not the case. Also RpcError::Parse seems to be used only here, just redefine it to `RpcError::Parse(String)`.

tycho-indexer/src/services/rpc.rs
zizou0x: Yes, you're right, I think it makes sense as it has 3 branches

tycho-indexer/src/services/rpc.rs
zizou0x: Nice, I've done it like this

PR #56, opened 2023-12-06
ENG 2406: add native protocols protobuf messages

louise-poole:
Messages to add:

- [x] Main block message to pass all the changes within a block
- [x] Protocol state changes
- [ ] Protocol component changes (message being added in another PR, to be added to this block message when it's done)

PS: There are a few bogey commits from failed rebase attempts and fixing formatting to make CI happy. I recommend reviewing this PR as a whole and not commit-by-commit

NOTE: many of the files are auto-generated or format fixes. The ones that need reviewing are: 
- `proto/tycho/evm/v1/vm.proto`
- `tycho-indexer/src/models.rs`


proto/tycho/evm/v1/vm.proto
```diff
@@ -49,3 +49,24 @@ message BlockContractChanges {
   Block block = 1;
   repeated TransactionChanges changes = 2;
 }
+
+message Attribute {
+  bytes name = 1;
+  bytes value = 2;
```
zizou0x, line 55: I'm a bit worried about this for dynamic sized values, like structs, mapping, array, ect.. But it might work with a standardized naming and I don't have a better idea atm

tycho-indexer/src/storage/postgres/contract_state.rs
```diff
@@ -743,7 +743,7 @@ where
                 .get_contract_slots(&id.chain, Some(&[account.address().clone()]), version, db)
                 .await?
                 .remove(&id.address)
-                .unwrap_or_else(HashMap::new);
+                .unwrap_or_default();
```
zizou0x, line 746: So coooool

proto/tycho/evm/v1/vm.proto
louise-poole: Remember when we met with Alan to discuss storing ProtocolStates? To be able to handle storing data for multiple protocols without needing to change the table schemas he suggested we use a tall table approach: store only key-value pairs. Which means we'll need to treat each attribute as a key-value pair and in the case of dicts, the attribute name and dict key will probably need to merge in the name here?

PR #57, opened 2023-12-07
ENG-2383: Implement cached DB write logic

zizou0x:
This PR introduces the core functionality for batched writes to the database, significantly enhancing performance and efficiency. The main focus of this update is to optimize data insertion processes by aggregating multiple write operations into a single batch, reducing the overhead associated with individual writes.

Additionally, this PR presents an initial approach to handle blockchain reorgs. We propose a strategy that involves flushing the pending state before getting contract deltas. This method aims to maintain data integrity and consistency in the face of potential reorg events.

Key Concerns and Future Work:

While this implementation lays the groundwork for efficient batch processing and reorg management, there are specific concerns to address in subsequent updates:

Handling Reorgs in Multi-Extractor Scenarios: One of the critical challenges we face is dealing with reorgs when multiple extractors are running concurrently. This scenario requires a more sophisticated approach to ensure that subscribers receive the correct data.

Further Improvements and Testing: The current implementation of reorg handling is in its early stages. We need to refine this approach through rigorous testing and potentially develop more robust mechanisms to handle these events.

- [x] Implement the main logic
- [x] Add tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4983,7 +4983,7 @@ checksum = "bfc13d450dc4a695200da3074dacf43d449b968baee95e341920e47f61a3b40f"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.37",
+ "syn 2.0.39",
 ]
 
 [[package]]
--- a/Readme.md
+++ b/Readme.md
@@ -165,7 +165,7 @@ export LIBRARY_PATH=<PATH-TO-LIB>:$LIBRARY_PATH
 ```bash
 cargo install diesel_cli --no-default-features --features postgres
 ```
-3. Start up a the postgres db service
+3. Start up the postgres db service
 ```bash
 docker-compose up -d db
 ```
--- a/tycho-indexer/src/extractor/evm/mod.rs
+++ b/tycho-indexer/src/extractor/evm/mod.rs
@@ -1,28 +1,30 @@
-pub mod ambient;
-pub mod storage;
-mod utils;
-
-use crate::{
-    hex_bytes::Bytes,
-    models::{Chain, ExtractorIdentity, NormalisedMessage},
-    storage::{ChangeType, StateGatewayType},
-};
 use std::{
     collections::{hash_map::Entry, HashMap},
     ops::Deref,
 };
-use tracing::warn;
-use utils::{pad_and_parse_32bytes, pad_and_parse_h160};
 
-use crate::pb::tycho::evm::v1 as substreams;
 use chrono::NaiveDateTime;
 use ethers::{
     types::{H160, H256, U256},
     utils::keccak256,
 };
 use serde::{Deserialize, Serialize};
+use tracing::warn;
+
+use utils::{pad_and_parse_32bytes, pad_and_parse_h160};
+
+use crate::{
+    hex_bytes::Bytes,
+    models::{Chain, ExtractorIdentity, NormalisedMessage},
+    pb::tycho::evm::v1 as substreams,
+    storage::{ChangeType, StateGatewayType},
+};
 
 use super::ExtractionError;
+
+pub mod ambient;
+pub mod storage;
+mod utils;
 
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
 pub struct SwapPool {}
@@ -176,7 +178,7 @@ impl AccountUpdate {
     ///
     /// # Errors
     ///
-    /// It returns an `ExtractionError::Unknown` error if `self.address` and
+    /// It returns an `ExtractionError::MergeError` error if `self.address` and
     /// `other.address` are not identical.
     ///
     /// # Arguments
@@ -185,7 +187,7 @@ impl AccountUpdate {
     /// of `other` will overwrite those of `self`.
     fn merge(&mut self, other: AccountUpdate) -> Result<(), ExtractionError> {
         if self.address != other.address {
-            return Err(ExtractionError::Unknown(format!(
+            return Err(ExtractionError::MergeError(format!(
                 "Can't merge AccountUpdates from differing identities; Expected {:#020x}, got {:#020x}",
                 self.address, other.address
             )));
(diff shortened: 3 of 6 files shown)


tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -99,83 +130,181 @@ impl WriteThroughCache {
                     // TODO: handle errors
```
zizou0x: Shouldn't we just panic?

tycho-indexer/src/extractor/evm/ambient.rs
```diff
@@ -155,7 +155,7 @@ impl AmbientPgGateway {
             .collect();
 
         self.state_gateway
-            .revert_contract_state(to, conn)
+            .revert_state(to, conn)
```
louise-poole, line 158: I think changing this might actually be bad 👀  I popped a message on slack to explain why.

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -62,25 +77,84 @@ struct Transaction {
...
 /// The class does provide read operations for completeness, but it will not consider any
 /// cached changes while reading. Any reads are direct pass throughs to the database.
-struct WriteThroughCache {
+struct DBCacheWriteExecutor {
     name: String,
     chain: Chain,
     pool: Pool<AsyncPgConnection>,
     state_gateway: EVMStateGateway<AsyncPgConnection>,
-    persisted_block: evm::Block,
+    persisted_block: Option<evm::Block>,
```
louise-poole, line 86: What exactly is `persisted_block`? Is it the last block written to the db?

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -89,93 +163,494 @@ impl WriteThroughCache {
...
+                            }
+                            Result::<(), StorageError>::Ok(())
                         }
                         .scope_boxed()
                     })
                     .await?;
-                } else if incoming.parent_hash == pending.hash {
-                    // TODO: handle errors
+
+                    //Notify that it
```
louise-poole: Cliffhanger 👀 😂

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -62,25 +77,84 @@ struct Transaction {
...
+    /// connect with the last persisted block).
+    async fn write(&mut self, new_db_tx: DBTransaction) -> Result<(), StorageError> {
         match self.pending_block {
             Some(pending) => {
-                if pending.hash == incoming.hash {
-                    self.pending_txns.push(tx);
-                } else if incoming.number < pending.number {
+                if pending.hash == new_db_tx.block.hash {
+                    // New database transaction for the same block
+                    self.pending_db_txs.push(new_db_tx);
```
louise-poole, line 156: Is it expected to send a write response here too since the write to cache was successful? Or do we only do that on actual db writes?

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -89,93 +163,494 @@ impl WriteThroughCache {
...
+                    //Notify that it
+                    new_db_tx
+                        .tx
+                        .send(Ok(()))
+                        .expect("response");
+                } else if new_db_tx.block.parent_hash == pending.hash {
+                    // New database transaction for the next block
                     self.flush().await;
+                    self.pending_block = Some(new_db_tx.block);
+                    self.pending_db_txs.push(new_db_tx);
```
louise-poole, line 196: Do we not need to send a write response here too? You do it in `run` after a flush - would expect it here too?

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -89,93 +163,494 @@ impl WriteThroughCache {
...
                     })
                     .await?;
-                } else if incoming.parent_hash == pending.hash {
-                    // TODO: handle errors
+
+                    //Notify that it
+                    new_db_tx
+                        .tx
+                        .send(Ok(()))
+                        .expect("response");
```
louise-poole: Can we have a bit more detail in this expect?

tycho-indexer/src/storage/postgres/cache.rs
louise-poole: Would make sense that a response is sent for all write cases - be it a success response or the write erroring.

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -89,93 +163,494 @@ impl WriteThroughCache {
...
-        todo!()
+    // pub async fn upsert_block(&self, new: &evm::Block) -> Result<(), StorageError> {
+    //     todo!()
+    // }
+    pub async fn get_account_delta(
+        &self,
+        chain: &Chain,
+        start_version: Option<&BlockOrTimestamp>,
+        end_version: &BlockOrTimestamp,
+        db: &mut AsyncPgConnection,
```
louise-poole, line 334: I think all the other gateways call this `conn`?

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -89,93 +163,494 @@ impl WriteThroughCache {
...
 
 impl CachedGateway {
     // TODO: implement the usual gateway methods here, but they are translated into write ops, for
     // reads call the gateway directly
-    pub async fn upsert_block(&self, new: &evm::Block) -> Result<(), StorageError> {
-        todo!()
+    // pub async fn upsert_block(&self, new: &evm::Block) -> Result<(), StorageError> {
+    //     todo!()
+    // }
+    pub async fn get_account_delta(
```
louise-poole: Question: this refers to getting changes for all accounts right? Maybe it should be called `get_accounts_delta` so that this is more clear?

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -89,93 +163,494 @@ impl WriteThroughCache {
...
+            .send(DBCacheMessage::Flush(tx))
+            .await;
+        let _ = rx.await;
+        self.state_gateway
+            .get_account_delta(chain, start_version, end_version, db)
+            .await
+    }
+}
+
+// These two implementation allow to inherit EVMStateGateway methods. If CachedGateway doesn't
```
louise-poole:
```suggestion
// These two implementations allow us to inherit EVMStateGateway methods. If CachedGateway doesn't
```

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -89,93 +163,494 @@ impl WriteThroughCache {
...
+
+    fn deref(&self) -> &Self::Target {
+        &self.state_gateway
+    }
+}
+
+impl DerefMut for CachedGateway {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.state_gateway
     }
```
louise-poole: Super cool! I will defs remember and use this!

tycho-indexer/src/storage/postgres/cache.rs
zizou0x: Yes that's it!

tycho-indexer/src/storage/postgres/cache.rs
zizou0x: So at the moment we have chosen to send the response only when it has been correctly received by the db. Later, if necessary, we could modify this design a little to also notify when a response is added to the cache

tycho-indexer/src/storage/postgres/cache.rs
zizou0x: So in `run` I notify the sender of the `Flush` message that it was flushed, here there is no `Flush` message so we don't need to notify anyone else. Concerning the db_tx, they are notified one by one in the `flush()` logic

tycho-indexer/src/storage/postgres/cache.rs
zizou0x: Hm actually I see both, I will improve consistency here

zizou0x: Ok actually I would prefer to do it in another PR when there aren't many opened PRs as it impacts a lot of files and will create a lot of conflicts

tycho-indexer/src/extractor/evm/ambient.rs
zizou0x: https://propellerswap.slack.com/archives/C05MPCJUQPM/p1702327191753739

PR #58, opened 2023-12-07
ENG-2032: Extend extractors and messages

SmartMonkey-git:
[Task](https://datarevenue.atlassian.net/browse/ENG-2038)

Added protobuf messages in `proto/tycho/evm/v1/vm.proto` and their respective `try_from_message()` functions in `tycho-indexer/src/extractor/evm/mod.rs`. The rest is formatting and autogenerated files.

One disclaimer - I replaced the `tycho-indexer/migrations/2023-08-15-154244_tycho_initial_setup/up.sql` with its master equivalent, because the sketched version had a bug I could not figure out. It caused the CI to fail. I would say we postpone it until we actually need the sql file.

Also please don't mind the bad commit messages, some of the should be squashed, but somehow that did not happen. Will not occur again!

### DIFF
--- a/env.sh
+++ b/env.sh
@@ -1,0 +1,2 @@
+export STREAMINGFAST_KEY=server_6e3a98b2aa01c779e44e2d9b80b337be
+export SUBSTREAMS_API_TOKEN=$(curl https://auth.streamingfast.io/v1/auth/issue -s --data-binary '{"api_key":"'$STREAMINGFAST_KEY'"}' | jq -r .token)
--- a/.gitignore
+++ b/.gitignore
@@ -76,7 +76,7 @@ Temporary Items
 # Substreams spkg files are build artifacts
 *.spkg
 
-env.sh
+.env
 .vscode
 .idea
 *.log


.dockerignore
```diff
@@ -9,4 +9,4 @@ venv
 .git
 **/*.spkg
 Dockerfile
-.env
+env.sh
```
louise-poole, line 12: Why not .env? 👀

substreams/ethereum-ambient/src/pb/tycho.evm.v1.rs
```diff
@@ -51,11 +51,31 @@ pub struct ContractChange {
 }
 #[allow(clippy::derive_partial_eq_without_eq)]
 #[derive(Clone, PartialEq, ::prost::Message)]
+pub struct ProtocolComponent {
+    #[prost(bytes="vec", repeated, tag="1")]
+    pub tokens: ::prost::alloc::vec::Vec<::prost::alloc::vec::Vec<u8>>,
+    #[prost(bytes="vec", repeated, tag="2")]
+    pub contracts: ::prost::alloc::vec::Vec<::prost::alloc::vec::Vec<u8>>,
+}
```
louise-poole, line 59: Hmmmm... these don't seem to match your `vm.proto` message definitions - I don't see the ID field? They probably just need to be re-generated?

proto/tycho/evm/v1/vm.proto
```diff
@@ -40,9 +40,22 @@ message ContractChange {
   ChangeType change = 5;
 }
 
+message ProtocolComponent {
+  repeated bytes tokens = 1;
+  repeated bytes contracts = 2;
+  bytes id = 3;
+}
```
louise-poole, line 47: I'd prefer is `id` was the first field since it's the unique identifier.

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -28,11 +30,46 @@ pub struct ProtocolComponent {
...
+            contracts: msg
+                .contracts
+                .into_iter()
+                .map(|inner_vec| {
+                    pad_and_parse_h160(inner_vec.as_slice())
+                        .map_err(|err| err.to_string())
+                        .unwrap()
+                })
+                .collect::<Vec<_>>(),
+            attributes: None,
```
louise-poole, line 69: What about `attributes`? How are these to be determined/set? I suppose this isn't integral to what you need this object for so I guess I can look into this part myself if you don't already have a plan here?

proto/tycho/evm/v1/vm.proto
```diff
@@ -40,9 +40,22 @@ message ContractChange {
...
+
+message TVLUpdate {
+  bytes token = 2;
+  bytes balance = 3;
```
louise-poole, line 51: Field indexes should start from 1 here.

proto/tycho/evm/v1/vm.proto
```diff
@@ -40,9 +40,22 @@ message ContractChange {
...
+}
+
 message TransactionChanges {
   Transaction tx = 1;
   repeated ContractChange contract_changes = 2;
+  repeated ProtocolComponent components = 3;
+  repeated TVLUpdate tvl = 4;
 }
```
louise-poole, line 59: These seem more like they belong to `BlockContractChanges`? None of the structs you build from these two fields need tx info or to be linked to a specific tx. I think it makes more sense for them to be collected per block than per tx.

proto/tycho/evm/v1/vm.proto
louise-poole: Also, these need to be linked to a protocol component somehow? I suggest adding a component_id field.

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -902,4 +960,22 @@ mod test {
 
         assert_eq!(res, block_account_changes());
     }
+
+    #[rstest]
+    fn test_try_from_message_tvl_change() {
```
louise-poole, line 965: We need tests for interpreting ProtocolComponent messages too. And the updated BlockContractChanges messages.

.dockerignore
pistomat: Or both?

env.sh
pistomat, line 1: Private keys should not be uploaded to git repository.

proto/sf/substreams/intern/v2/deltas.proto
```diff
@@ -25,12 +23,12 @@ message StoreDelta {
...
 message ModuleOutput {
-  string module_name = 1;
-  oneof data {
-    google.protobuf.Any map_output = 2;
-    StoreDeltas store_deltas = 3;
-  }
-  repeated string logs = 4;
-  bool debug_logs_truncated = 5;
-  bool cached = 6;
+    string module_name = 1;
```
pistomat, line 26: FMI, what formatter do you use? Is the style specified somewhere?

It would be great if there was a config and if the CI would be updated to check the style of protobuf files.

proto/tycho/evm/v1/vm.proto
pistomat: Neither `ContractChange` has a link to the specific tx. Afaik all changes and updates should be linked to a tx by default (i.e. included in TransactionChanges) and only afterwards maybe aggregated into BlockChanges to save on compute/db access in Tycho, but it would be great if someone could clarify this for me.

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -28,11 +30,46 @@ pub struct ProtocolComponent {
...
+                .into_iter()
+                .map(|inner_vec| {
+                    pad_and_parse_h160(inner_vec.as_slice())
+                        .map_err(|err| err.to_string())
+                        .unwrap()
+                })
+                .collect::<Vec<_>>(),
+            attributes: None,
+            chain,
+        });
```
pistomat, line 71: You should avoid using `.unwrap()` if possible at all. The function signature is a `Result<Self, ExtractionError>`, so you should propagate any errors you encounter up. For this the most legible way is to parse the tokens and contracts separately and use the `?` to propagate the error to the function itself.

```suggestion
        let tokens = msg
            .tokens
            .iter()
            .map(|inner_vec| pad_and_parse_h160(inner_vec))
            .collect::<Result<Vec<_>, _>>()
            .map_err(ExtractionError::DecodeError)?;

        let contracts = msg
            .contracts
            .iter()
            .map(|inner_vec| pad_and_parse_h160(inner_vec))
            .collect::<Result<Vec<_>, _>>()
            .map_err(ExtractionError::DecodeError)?;

        Ok(Self {
            id: msg.id,
            protocol_type_name,
            protocol_system,
            tokens,
            contracts,
            attributes: None,
            chain,
        })
```

env.sh
SmartMonkey-git: No problem for this one. We are not going to use it. I just gonna delete it on the steamfast side

proto/sf/substreams/intern/v2/deltas.proto
SmartMonkey-git: Sounds like a nice cooldown task!

PR #59, opened 2023-12-07
ENG 2407: deserialise protocol messages

louise-poole:
TODO: update once ProtocolComponent message handling is merged here

### DIFF
--- a/tycho-indexer/src/extractor/evm/mod.rs
+++ b/tycho-indexer/src/extractor/evm/mod.rs
@@ -178,7 +178,7 @@ impl AccountUpdate {
     ///
     /// # Errors
     ///
-    /// It returns an `ExtractionError::Unknown` error if `self.address` and
+    /// It returns an `ExtractionError::MergeError` error if `self.address` and
     /// `other.address` are not identical.
     ///
     /// # Arguments
@@ -187,7 +187,7 @@ impl AccountUpdate {
     /// of `other` will overwrite those of `self`.
     fn merge(&mut self, other: AccountUpdate) -> Result<(), ExtractionError> {
         if self.address != other.address {
-            return Err(ExtractionError::Unknown(format!(
+            return Err(ExtractionError::MergeError(format!(
                 "Can't merge AccountUpdates from differing identities; Expected {:#020x}, got {:#020x}",
                 self.address, other.address
             )));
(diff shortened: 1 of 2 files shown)


tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -704,16 +921,16 @@ mod test {
 
     #[rstest]
     #[case::diff_block(
-        fixtures::transaction02(HASH_256_1, HASH_256_1, 11),
-        Err(ExtractionError::Unknown(format!("Can't merge AccountUpdates from different blocks: 0x{:x} != {}", H256::zero(), HASH_256_1)))
+    fixtures::transaction02(HASH_256_1, HASH_256_1, 11),
+    Err(ExtractionError::Unknown(format ! ("Can't merge AccountUpdates from different blocks: 0x{:x} != {}", H256::zero(), HASH_256_1)))
```
zizou0x: Lol what is this formatting? Space in macros?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -847,4 +1064,240 @@ mod test {
...
+            attributes: attributes2.clone(),
+            modify_tx: tx,
+        };
+
+        let res = state1.merge(state2);
+
+        assert_eq!(res, exp);
+    }
+
+    #[rstest]
```
zizou0x: I think we don't need rstest when there aren't several cases

tycho-indexer/src/extractor/evm/mod.rs
louise-poole: My plan was to fix formatting once I merge Rouven's changes in 😂  Maybe I should just look into it now. No clue what's up though - I've set it up exactly like we have in the docs etc.

pistomat: Maybe it is messed up because it is inside an attribute?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -488,6 +490,166 @@ impl BlockStateChanges {
...
+        };
+
+        let mut attributes = HashMap::new();
+        for attribute in msg.attributes.into_iter() {
+            let name = match String::from_utf8(attribute.name) {
+                Ok(s) => s,
+                Err(err) => return Err(ExtractionError::DecodeError(err.to_string())),
+            };
+            attributes.insert(name, Bytes::from(attribute.value));
+        }
```
pistomat: You can be a bit more idiomatic and avoid a mutable HashMap like this
```suggestion
        
        let component_id = String::from_utf8(msg.component_id)
            .map_err(|err| ExtractionError::DecodeError(err.to_string()))?;

        let attributes = msg.attributes.into_iter().map(|attribute| {
            let name = String::from_utf8(attribute.name)
                .map_err(|err| ExtractionError::DecodeError(err.to_string()))?;
            Ok((name, Bytes::from(attribute.value)))
        }).collect::<Result<HashMap<_, _>, ExtractionError>>()?;
```
but your original way is a bit more legible.

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -488,6 +490,166 @@ impl BlockStateChanges {
...
+    ///   the method will return an error.
+    ///
+    /// The merged update keeps the transaction of `other`.
+    ///
+    /// # Errors
+    /// This method will return `ExtractionError::Unknown` if any of the above
+    /// conditions is violated.
+    pub fn merge(&mut self, other: ProtocolState) -> Result<(), ExtractionError> {
+        if self.component_id != other.component_id {
+            return Err(ExtractionError::Unknown(format!(
```
pistomat: The `ExtractionError` is not actually `Unknown`, isn't it? Maybe add another error type?

louise-poole: I tried to stick closely to how `AccountUpdateWithTx` implemented its `merge` and they used `Unknown` errors there. The `Unknown` errors simply mean they are unexpected errors and if `.merge` is used correctly these should never appear. I can add a new error type if you still feel it's necessary?

PR #61, opened 2023-12-11
ENG-2038: Extend Protobuf messages

SmartMonkey-git:


tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -22,7 +23,7 @@ enum WriteOp {
     UpdateContracts(Vec<(H256, evm::AccountUpdate)>),
     RevertContractState(BlockIdentifier),
 }
-
+#[allow(dead_code)]
```
louise-poole: Instead of repeating this a million times in the same file just put #![allow(dead_code)] at the top of the file.

proto/tycho/evm/v1/vm.proto
```diff
@@ -40,9 +40,23 @@ message ContractChange {
   ChangeType change = 5;
 }
 
+message ProtocolComponent {
+  bytes id = 1;
+  repeated bytes tokens = 2;
+  repeated bytes contracts = 3;
```
dianacarvalho1: why do we need several contracts here?

tycho-indexer/src/models.rs
```diff
@@ -114,3 +190,101 @@ pub struct ProtocolState {
     // via transaction, we can trace back when this state became valid
     modify_tx: Bytes,
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    #[allow(unused_imports)]
```
dianacarvalho1: why do we want unused imports?

tycho-indexer/src/models.rs
```diff
@@ -114,3 +190,101 @@ pub struct ProtocolState {
...
+        let msg = substreams::ProtocolComponent {
+            id: b"component_id".to_vec(),
+            tokens: vec![b"token1".to_vec(), b"token2".to_vec()],
+            contracts: vec![b"contract1".to_vec(), b"contract2".to_vec()],
+        };
+
+        // Sample parameters for testing
+        let protocol_system = ProtocolSystem::Ambient;
+        let protocol_type = ProtocolType {
+            name: "Pool".to_string(),
```
dianacarvalho1: what is usually in `"name"`?

tycho-indexer/src/models.rs
```diff
@@ -114,3 +190,101 @@ pub struct ProtocolState {
...
+        // Call the try_from_message method
+        let result = ProtocolComponent::<String>::try_from_message(
+            msg,
+            ProtocolSystem::Ambient,
+            ProtocolType {
+                name: "Pool".to_string(),
+                attribute_schema: serde_json::Value::default(),
+                financial_type: crate::models::FinancialType::Psm,
+                implementation_type: crate::models::ImplementationType::Custom,
+            },
```
dianacarvalho1: you have it defined above already

substreams/ethereum-ambient/src/pb/tycho.evm.v1.rs
```diff
@@ -2,67 +2,91 @@
...
 }
 #[allow(clippy::derive_partial_eq_without_eq)]
 #[derive(Clone, PartialEq, ::prost::Message)]
+pub struct ProtocolComponent {
+    #[prost(bytes = "vec", tag = "1")]
+    pub id: ::prost::alloc::vec::Vec<u8>,
+    #[prost(bytes = "vec", repeated, tag = "2")]
+    pub tokens: ::prost::alloc::vec::Vec<::prost::alloc::vec::Vec<u8>>,
+    #[prost(bytes = "vec", repeated, tag = "3")]
+    pub contracts: ::prost::alloc::vec::Vec<::prost::alloc::vec::Vec<u8>>,
```
tamaralipows, line 60: Question @dianacarvalho1 @SmartMonkey-git @louise-poole , but don't let this block from merging if everyone already approved. For some protocols like ambient, we need some additional info here (like pool index) I guess. Are we just going to make a massive `ProtocolComponent` object that is supposed to encompass all of these protocols, and add the protocol-specific attributes optionally?

tycho-indexer/src/models.rs
```diff
@@ -100,12 +105,83 @@ pub struct ProtocolComponent<T> {
     protocol_system: ProtocolSystem,
     // more metadata information about the components general type (swap, lend, bridge, etc.)
     protocol_type: ProtocolType,
+    // Blockchain the component belongs to
+    chain: Chain,
     // holds the tokens tradable
     tokens: Vec<T>,
+    // ID's referring to related contracts
+    contract_ids: Vec<String>,
```
tamaralipows: What does this mean? Are these addresses?

tycho-indexer/src/models.rs
```diff
@@ -100,12 +105,83 @@ pub struct ProtocolComponent<T> {
...
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
+pub struct TvlChange<T> {
+    token: T,
+    new_balance: f64,
+    // tx where the this balance was observed
+    tx: String,
+}
+
+impl TvlChange<String> {
```
tamaralipows: What's the difference between the `TvlChange` and `TvlUpdate` classes? One of them is a message, right? Wouldn't it be more intuitive to call it `TvlUpdate` and `TvlUpdateMessage`? I genuinely don't know though, let me know your reason for doing it this way

tycho-indexer/src/models.rs
```diff
@@ -100,12 +105,83 @@ pub struct ProtocolComponent<T> {
...
+    #[allow(dead_code)]
+    pub fn try_from_message(
+        msg: substreams::ProtocolComponent,
+        protocol_system: ProtocolSystem,
+        protocol_type: ProtocolType,
+        chain: Chain,
+    ) -> Result<Self, ExtractionError> {
+        let id = String::from_utf8(msg.id)
+            .map_err(|error| ExtractionError::DecodeError(error.to_string()))
+            .unwrap();
```
louise-poole: I think it is probably best to keep this file for struct definitions only and to add the `try_from_message` implementations to the mod file where the others are?

tycho-indexer/src/models.rs
```diff
@@ -100,12 +105,83 @@ pub struct ProtocolComponent<T> {
...
+            })
+            .collect::<Vec<_>>();
+
+        Ok(Self {
+            id,
+            protocol_type,
+            protocol_system,
+            tokens,
+            contract_ids,
+            attribute_schema: Bytes::default(),
```
louise-poole: The ER diagram suggests the attribute schema will be in the `ProtocolType` struct instead of here as it's a protocol/exchange specific field and not a pool specific one. As mentioned [here](https://propellerswap.slack.com/archives/C05MPCJUQPM/p1702288711608589), we should probably change this to `static_attributes`?

tycho-indexer/src/models.rs
```diff
@@ -100,12 +105,83 @@ pub struct ProtocolComponent<T> {
...
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
+pub struct TvlChange<T> {
+    token: T,
+    new_balance: f64,
+    // tx where the this balance was observed
+    tx: String,
+}
```
louise-poole: You will need a component_id field here, no? We need to link the tvl update to a pool.

tycho-indexer/src/models.rs
louise-poole: Yeah, TVLUpdate is a substream message. We will always see it's type given with `substreams::` in front of it so that should help with identifying it. I'm not sure about appending `Message` on the end of each message struct, but we can if people feel strongly about it? My thoughts are, this is the only place these two types overlap. In the substreams they only care about the message structs and having `Message` on the end of every single struct used there feels excessive. On the extractor and gateways they only care about these structs created here and don't deal with messages/message types.

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -112,16 +114,16 @@ impl WriteThroughCache {
 
     async fn execute_tx(
         &mut self,
-        ops: Vec<WriteOp>,
-        conn: &mut AsyncPgConnection,
+        #[allow(unused_variables)] ops: Vec<WriteOp>,
```
louise-poole: You can also put this just once as `#![allow(unused_variables)]` at the top of this file. This will make it much nicer for Flo merging his PR as he'll only have 1 line to remove instead of all of these.

tycho-indexer/src/services/ws.rs
```diff
@@ -1,4 +1,5 @@
 //! This module contains Tycho Websocket implementation
+#[allow(unused_imports)]
```
louise-poole: We should remove unused imports instead of ignoring the message. I did notice sometimes if you removed an import marked as unused a test would complain. You should then just import that same file inside the test module only then.

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -592,6 +592,8 @@ pub mod fixtures {
                         change: ChangeType::Update.into(),
                     },
                 ],
+                components: vec![],
+                tvl: vec![],
```
louise-poole: Would be better to add the actual messages here and update the related test to make sure they're handled correctly when a `BlockContractChanges` message is deserialised.

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -438,7 +438,7 @@ impl BlockStateChanges {
                 block,
                 tx_updates,
                 new_pools: HashMap::new(),
-            })
+            });
         }
         Err(ExtractionError::Empty)
```
louise-poole: You need to update the `BlockContractChanges` message deserialiser to decode the tvl and protocol component messages found in the TransactionChanges part of the message. The protocol components deserialised will be the `new_pools` you need to pass here.

louise-poole: Keep in mind, we are ultimately adding these structs and methods for the extractor to be able to decode and handle substream messages. Noting that, the main struct the extractor emits is the `BlockAccountChanges` objects that it creates from `BlockContractChanges.aggregate_updates()`. This means all info you want to be made available to users needs to be in the `BlockContractChanges` struct and by extension, the `BlockAccountChanges` struct. `new_pools` will house the new ProtocolComponents you receive from the substream. TVL is missing in these structs though. You need to still add them to the structs and then make sure to deserialise the relevant part of the block message so that the info isn't lost.

substreams/ethereum-ambient/src/pb/tycho.evm.v1.rs
louise-poole: Good question - I also left a comment on this PR relating to this. I assume pool_index is a static pool attribute? Basically in the struct dependency plan Alan created he has an `attributes` property in `ProtocolComponent`. This is for any protocol attributes that will never change. I proposed renaming it `static_attributes` so that this is clear. This will need to be added to the protocol components message too so that we can get the info from somewhere.

tycho-indexer/src/extractor/evm/mod.rs
louise-poole: Just had another thought too: have a look at the `BlockContractChanges` `aggregate_updates` function. It is used to aggregate all updates made to the same account so that you have only one `account_update` entry per account. You will need to do the same for tvl updates.

proto/tycho/evm/v1/vm.proto
```diff
@@ -40,9 +40,23 @@ message ContractChange {
...
+}
+
+message TVLUpdate {
```
kayibal: Lets call this BalanceChange please.

proto/tycho/evm/v1/vm.proto
SmartMonkey-git: I'm quoting alan here:

> Ok now ProtocolComponent and Contracts often have a 1-1 relationship - e.g. UniswapV2 and UniswapV3. Other times they have a many to one relationship (e.g. Ambient only has a single contract). So the protocol id does not necessarily have to be an address, it could also be sth like "USDC-ETH" and link to the ambient contract. The same way it may be that multiple contracts are involved in a swap operation although so far I have not come across such a protocol.

[Original Thread](https://propellerswap.slack.com/archives/C05LQALE9L1/p1702033413523879?thread_ts=1702032292.896259&cid=C05LQALE9L1)

tycho-indexer/src/models.rs
SmartMonkey-git: When I remove `use actix_web::body::MessageBody` and run the tests I get:

```
error[E0599]: no method named `try_into_bytes` found for reference `&str` in the current scope
   --> tycho-indexer/src/models.rs:274:14
    |
273 |           let msg_token = expected_token
    |  _________________________-
274 | |             .try_into_bytes()
    | |             -^^^^^^^^^^^^^^ method not found in `&str`
    | |_____________|
    | 
    |
    = help: items from traits can only be used if the trait is in scope
help: the following trait is implemented but not in scope; perhaps add a `use` for it:
    |
196 +     use actix_web::body::MessageBody;
    |
```

I'm not sure how they relate and how an import in my files influences the presence of a function in another file.
@pistomat Do you have an explanation for this?

louise-poole: I left a comment below to explain how to do this, but will copy it here too for reference:
> We should remove unused imports instead of ignoring the message. I did notice sometimes if you removed an import marked as unused a test would complain. You should then just import that same file inside the test module only then.

tycho-indexer/src/models.rs
SmartMonkey-git: Yes, that is true, however we are in not python but in rust. :)
Where things take ownership. If I pass the object here `ProtocolType` will take ownership of it, to avoid that I would need to copy it, but to copy it I would need to implement the copy trait.
At least from my limited knowledge this is fine.

tycho-indexer/src/models.rs
SmartMonkey-git: I would guess that it is something similar to `pool`.
Protocol types are a way to categories components behaviour. Another type might be "PoolFactory".

Let me quote from the [planning](https://datarevenue.atlassian.net/browse/ENG-1947), maybe that is better:

> A swap pool component should describe which tokens it allows to be exchanged, it can include additional information e.g. if it suffers price impact or how much of a fee it charges. Components that behave the same are grouped under a ProtocolType.[...] We expect protocol types and therefore also components to be highly heterogeneous. Meaning there will be many different kinds of them (e.g. uniswap-v2 pools, uniswap-v3 pools, curve pools, etc.).

tycho-indexer/src/models.rs
louise-poole: You can add `#[derive(Clone)]` above the definition of the `ProtocoType` struct and then just use `protocol_type.clone()` here.

tycho-indexer/src/models.rs
SmartMonkey-git: In other words `MessageBody` is never used directly, but I need the import to have access to `try_into_bytes` on the string class. That is why `I want` an unused import.

louise-poole: Ah, soz! Misunderstood that 😅  That's very weird - I'm curious to the answer to this too 👀

proto/tycho/evm/v1/vm.proto
dianacarvalho1: ok, I see several things that are not reflected on this protobuf message.
he says:
- ProtocolComponent and Contracts are often 1-to-1 but they can also be many-to-1. 
Right now we have 1-to-many.
- the protocol id does not necessarily have to be an address, it could also be sth like "USDC-ETH" 
We have `bytes id`, can that be a string too?

tycho-indexer/src/models.rs
dianacarvalho1: but then wouldn't it better to make it an enum?

proto/tycho/evm/v1/vm.proto
SmartMonkey-git: > ProtocolComponent and Contracts are often 1-to-1 but they can also be many-to-1.
Right now we have 1-to-many.

Not really, you can just have 1 element in that list. I'm not sure what change you are suggesting.

> the protocol id does not necessarily have to be an address, it could also be sth like "USDC-ETH"
We have bytes id, can that be a string too?

Hmmm... good observation. I oriented my self on the address attribute of `ContractChange`, which gets casted to a `H160` in `extractor/evm/mod.rs` and `H160` is a hash type that can represent an address, but could not represent something like `USDC-ETH`.
So, I guess string would be better than the bytes. 🤔 
No, sure if the hexer-decimal representation of an address would give us problems when converting.

tycho-indexer/src/models.rs
pistomat: I understand you need to import, but locally when I remove the `#[allow(unused_imports)]` I don't get any complaints...

tycho-indexer/src/models.rs
SmartMonkey-git: I'm not sure how dynamic these types need to be and where they ultimately get declared and discovered.

tycho-indexer/src/models.rs
pistomat: I would actually name both `TvlUpdate` and alias the substream one like `use substreams::TvlUpdate as TvlUpdateMessage` or something like that. I agree with Louise that it does not make sense to suffix every message with `Message`, we can do it on conflicting imports only.

proto/tycho/evm/v1/vm.proto
louise-poole: > ProtocolComponent and Contracts are often 1-to-1 but they can also be many-to-1.
Right now we have 1-to-many.

Right now you have 1 ProtocolComponent that can have many contracts. Alan's write up said we could have a situation where you have many ProtocolComponents that all belong to the same 1 contract.

louise-poole: > We have bytes id, can that be a string too?

ProtocolComponent's `id` field is already String. It's shows as bytes here in the message struct as everything is represented as bytes in our messages and converted when decoded.

tycho-indexer/src/models.rs
SmartMonkey-git: Ok, false alarm. It was a product of my CI-Pleasing marathon yesterday.
My mistake.
image

proto/tycho/evm/v1/vm.proto
SmartMonkey-git: > ProtocolComponent's id field is already String. It's shows as bytes here in the message struct as everything is represented as bytes in our messages and converted when decoded.

Ok, but protobuf also allows for a string type, why should we not choose that one?

tycho-indexer/src/models.rs
SmartMonkey-git: Let me know if [this](https://github.com/propeller-heads/tycho-indexer/pull/61#discussion_r1422891224) answers your question

tycho-indexer/src/models.rs
SmartMonkey-git: That does not sound intuitive to me. Spreading the implementation of a struct through the code base, but that is the python developer speaking and he does not know or understand rust norms.
@pistomat Can I have your opinion?

tycho-indexer/src/models.rs
SmartMonkey-git: Renamed the protosim message to `BalanceChange`

tycho-indexer/src/extractor/evm/mod.rs
dianacarvalho1: Updating `BlockContractChanges` will be done in separate PR! I'm working on it now

tycho-indexer/src/models.rs
dianacarvalho1: well all the other implementations are in `mod.rs` file

SmartMonkey-git: Yes, but so are their structs

dianacarvalho1: and considering that these methods will be used inside the `mod.rs` file (_I think_), it makes sense to put them there.

tycho-indexer/src/extractor/evm/mod.rs
louise-poole: Oh really? Weird that it's not just done here as it's highly coupled with this PR, but cool cool 👍 😁

tycho-indexer/src/models.rs
SmartMonkey-git: I think the struct is closely related to the functionality of the implementation, because to modify the `try_from_message` you need the attributes. 
I also did a search to see what other rust community thinks. I found different philosophies on it, but this [reddit post ](https://www.reddit.com/r/rust/comments/zzumcy/is_it_good_practice_to_have_impl_blocks_in/) (I know maybe not be best source of authority) describes splitting them as horrifying.

> be used inside the mod.rs file

Then you can import them as usual

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -70,7 +70,12 @@ fn map_changes(
 ) -> Result<tycho::BlockContractChanges, substreams::errors::Error> {
     let mut block_changes = tycho::BlockContractChanges { block: None, changes: Vec::new() };
 
-    let mut tx_change = tycho::TransactionChanges { tx: None, contract_changes: Vec::new() };
+    let mut tx_change = tycho::TransactionChanges {
+        tx: None,
+        contract_changes: Vec::new(),
+        components: Vec::new(),
+        tvl: Vec::new(),
+    };
```
pistomat: There is a really nice way to init this

```suggestion
    let mut tx_change = tycho::TransactionChanges::default();
```

Prost messages apparently define the Default trait, Option defaults to None and Vec defaults to an empty vec, so you can skip this.

tycho-indexer/src/models.rs
pistomat: Yeah it would be great to extract the type to `pub struct ContractId(String)` and add proper documentation of this.

tycho-indexer/src/models.rs
pistomat: It does not sound very idiomatic to split the implementation from the definition to me. Even (unit) tests in rust are usually in the same file.

dianacarvalho1: but then why do we have these two files to begin with? `tycho-indexer/src/extractor/evm/mod.rs` and `tycho-indexer/src/models.rs`

proto/tycho/evm/v1/vm.proto
pistomat: If the ContractId is always a String (afaik it is, only sometimes it is a converted address), then please follow the defined types https://github.com/tokio-rs/prost?tab=readme-ov-file#scalar-values to make it more legible.

tycho-indexer/src/models.rs
```diff
@@ -116,3 +186,99 @@ pub struct ProtocolState {
...
+        let tx = create_transaction();
+        let expected_balance: f64 = 3000.0;
+        let msg_balance = expected_balance.to_le_bytes().to_vec();
+
+        let expected_token = "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2";
+        let msg_token = expected_token
+            .try_into_bytes()
+            .unwrap()
+            .to_vec();
+        let expected_component_id = "DIANA-THALES";
```
pistomat: The best component there is! 😀

tycho-indexer/src/models.rs
pistomat: Okay, looking at this more closely, I can see that there is an attempt to split the extractor logic and the model definition. Is it true that the `try_from_message` fn will be used only within the extractor module to parse substreams messages and convert them to the models? If there is no need to expose that function to the rest of the `tycho-indexer` package, then I agree with the slit. We don't want to expose internal parsing functions unnecessarily.

proto/tycho/evm/v1/vm.proto
pistomat: I can see that Alan uses `bytes` most of the time to save on message size for tokens and balances, as they are always parsed from hex (i.e. Address/H160 or Uint256), but for actual strings I would just use `string`. @SmartMonkey-git

louise-poole: No idea, maybe we should? 👀

tycho-indexer/src/models.rs
```diff
@@ -93,18 +93,88 @@ pub trait NormalisedMessage: std::fmt::Debug + std::fmt::Display + Send + Sync +
...
     protocol_type: ProtocolType,
+    // Blockchain the component belongs to
+    chain: Chain,
     // holds the tokens tradable
     tokens: Vec<T>,
+    // ID's referring to related contracts
+    contract_ids: Vec<String>,
     // allows to express some validation over the attributes if necessary
-    attribute_schema: Bytes,
+    static_attributes: Bytes,
```
louise-poole: You'll need to update the type and comment here too. Type should probably be the same as `attributes` in `ProtocolState`?

tycho-indexer/src/models.rs
```diff
@@ -93,18 +93,88 @@ pub trait NormalisedMessage: std::fmt::Debug + std::fmt::Display + Send + Sync +
...
+        let id = String::from_utf8(msg.id)
+            .map_err(|error| ExtractionError::DecodeError(error.to_string()))
+            .unwrap();
+        let tokens = msg
+            .tokens
+            .into_iter()
+            .map(|t| {
+                String::from_utf8(t)
+                    .map_err(|error| ExtractionError::DecodeError(error.to_string()))
+                    .unwrap()
```
louise-poole: I'm not such a fan of using `unwrap` outside of tests. There's always a risk it will lead to a panic when your program should be able to handle the errors. I think we can handle this more gracefully with something like:
```
match String::from_utf8(t) {
    Ok(string) => string,
    Err(error) => return Err(ExtractionError::DecodeError(error.to_string())),
}
```
Or since you're not doing anything further to the value returned we can just use `?`:
```
String::from_utf8(t).map_err(|error| ExtractionError::DecodeError(error.to_string()))?
```

substreams/ethereum-ambient/src/pb/tycho.evm.v1.rs
louise-poole: I unresolved this because it hasn't been fully addressed? The message needs to have this info in it so that it can be set when you create a ProtocolComponent from the message.

>  I proposed renaming it static_attributes so that this is clear. This will need to be added to the protocol components message too so that we can get the info from somewhere.

tycho-indexer/src/models.rs
```diff
@@ -93,18 +93,88 @@ pub trait NormalisedMessage: std::fmt::Debug + std::fmt::Display + Send + Sync +
...
+        })
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
+pub struct TvlChange<T> {
+    token: T,
+    new_balance: f64,
+    // tx where the this balance was observed
+    tx: String,
```
louise-poole: In ProtocolState this is called `modify_tx` (the tx that modified this state/balance). For consistency maybe we should also name this one `modify_tx`? I don't feel strongly about this - just an observation.

tycho-indexer/src/models.rs
```diff
@@ -93,18 +93,88 @@ pub trait NormalisedMessage: std::fmt::Debug + std::fmt::Display + Send + Sync +
...
+}
+
+impl TvlChange<String> {
+    pub fn try_from_message(
+        msg: substreams::BalanceChange,
+        tx: &Transaction,
+    ) -> Result<Self, ExtractionError> {
+        Ok(Self {
+            token: String::from_utf8(msg.token)
+                .map_err(|error| ExtractionError::DecodeError(error.to_string()))?,
```
louise-poole: Question: is this not meant to be of type T instead of String? Also, do we feel it's necessary to have the whole Token inside this tvl object, maybe the address is enough since TVLChange is already linked to a ProtocolComponent which has the whole token objects if needed.

tycho-indexer/src/models.rs
louise-poole: Also, are tokens not meant to be type `T` here too? This looks like they're just strings to me.

pistomat: 100% agree with no `unwrap` outside tests if possible.

SmartMonkey-git: > I'm not such a fan of using unwrap outside of tests.

Then let me tell you - you are not and neither am I.

I have already tried the suggestions you propose and I was unable to write the code in such a way, that I am can circumvent this error:
```
error[E0277]: the `?` operator can only be used in a closure that returns `Result` or `Option` (or another type that implements `FromResidual`)
   --> tycho-indexer/src/models.rs:144:31
    |
144 |             .map(|res_t| res_t?)
    |                  -------      ^ cannot use the `?` operator in a closure that returns `std::string::String`
    |                  |
    |                  this function should return `Result` or `Option` to accept `?`
    |
    = help: the trait `FromResidual<std::result::Result<Infallible, extractor::ExtractionError>>` is not implemented for `std::string::String`
```

SmartMonkey-git: > Also, are tokens not meant to be type T here too? This looks like they're just strings to me.

No, It can not be T. Otherwise you would not be able to implement, because you need to implement for a type if your class uses generics.

tycho-indexer/src/models.rs
SmartMonkey-git: It can not be T, because T is the generic substitute. You need to choose a type for T to implement. If not string what else?
It is not a whole token it is a string.

tycho-indexer/src/models.rs
louise-poole: Okay, but then have you tried using `?` outside of the map instead then? Something like this:
```
let tokens = msg
            .tokens
            .into_iter()
            .map(|t| {
                String::from_utf8(t)
                    .map_err(|error| ExtractionError::DecodeError(error.to_string()))
            })
            .collect::<Result<Vec<_>, _>>()?;
```

louise-poole: > No, It can not be T. Otherwise you would not be able to implement, because you need to implement for a type if your class uses generics.

Hmmm... okay, yeah this is tricky. Looks like tokens are their own separate struct in the ER diagram though so the whole token probably isn't necessary here? We just need the token id? So we likely can make this `String` from the get go instead of a generic.

tycho-indexer/src/models.rs
dianacarvalho1: aaaaaahh I think it's because tycho-indexer/src/models.rs is general for all chains and tycho-indexer/src/extractor/evm/mod.rs is evm specific. ok ok, got it. AG

proto/tycho/evm/v1/vm.proto
SmartMonkey-git: > No idea, maybe we should? 👀

I also dont know. :D

But would say lets do it like the other structs doin' it

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -592,6 +592,30 @@ pub mod fixtures {
                         change: ChangeType::Update.into(),
                     },
                 ],
+                components: vec![ProtocolComponent {
+                    id: vec![99, 111, 109, 112, 111, 110, 101, 110, 116, 95, 105, 100],
+                    tokens: vec![
+                        vec![116, 111, 107, 101, 110, 49],
+                        vec![116, 111, 107, 101, 110, 50],
```
dianacarvalho1: what do these arrays actually mean?

tycho-indexer/src/models.rs
```diff
@@ -92,19 +92,115 @@ impl ExtractionState {
 pub trait NormalisedMessage: std::fmt::Debug + std::fmt::Display + Send + Sync + 'static {
     fn source(&self) -> ExtractorIdentity;
 }
+/// A type representing the unique identifier for a contract. It can represent an on-chain address
+/// or in the case of a one-to-many relationship it could be something like 'USDC-ETH'. This is for
+/// example the case with arbiturm, where one component is responsible for multiple contracts.
```
dianacarvalho1: ambient not arbiturm

tycho-indexer/src/models.rs
```diff
@@ -92,19 +92,115 @@ impl ExtractionState {
...
+            tokens,
+            contract_ids,
+            static_attributes: attribute_map,
+            chain,
+        })
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
+pub struct TvlChange<T> {
```
dianacarvalho1: didn't we want to rename them all to BalanceUpdate?

tycho-indexer/src/models.rs
```diff
@@ -92,19 +92,115 @@ impl ExtractionState {
...
+    token: T,
```
dianacarvalho1: why is token generic though? why not just string from the start? so the token address I assume

tycho-indexer/src/models.rs
louise-poole: Ah, okay. This is misleading and maybe unnecessary to have a generic type here then? If we expect only token addresses this could be set as String?

tycho-indexer/src/models.rs
dianacarvalho1: I think making it a string is also better for what's to come (meaning my task xD)

tycho-indexer/src/extractor/evm/mod.rs
dianacarvalho1: I'm not sure if I'm misunderstanding something her though

dianacarvalho1: how can we have so many tokens and ids inside the same `ProtocolComponent`? I thought we would have one `ProtocolComponent` per pool created (in ambient's case). so with that in mind, how can we have a vec of vecs of tokens?? and so many ids? I'm just confused overall with this fixture

dianacarvalho1: I thought it could be some rust magic but it doesn't look like it.. 
image

louise-poole: @dianacarvalho1 a `vec[u8]` is how you represent bytes in rust. So this line:
```
id: vec![99, 111, 109, 112, 111, 110, 101, 110, 116, 95, 105, 100],
```
Represents just 1 id that is stored as bytes.

louise-poole: I personally prefer tests assets to be more human readable/easily updatable. So I like doing something like this instead: https://github.com/propeller-heads/tycho-indexer/pull/59/files#diff-c8a481153061f79091542b6d75ee23476ea651b936f5dd92ed8504a54f675eadR761

pistomat: I cannot comment on the number of tokens and ids, but instead of the vecs I would prefer using something like `hex::decode("0xaaaaaaaaa24eeeb8d57d431224f73832bc34f688".trim_start_matches("0x"))` to make it more legible

tycho-indexer/src/models.rs
louise-poole: Wait, thinking about it our token addresses are always hex strings. So can be stored as `H160`? We only use custom ids/addresses for pools.

tycho-indexer/src/models.rs
louise-poole: Since they're always hex strings/addresses we should actually store them as `H160`?

tycho-indexer/src/extractor/evm/mod.rs
dianacarvalho1: yeeees please!! I would prefer that tooo 🙏🏼

dianacarvalho1: me no speak bytes

tycho-indexer/src/models.rs
```diff
@@ -92,19 +93,39 @@ impl ExtractionState {
...
+    // allows to express some validation over the static attributes if necessary
+    pub static_attributes: HashMap<String, Bytes>,
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
+pub struct TvlChange {
+    pub token: H160,
+    pub new_balance: f64,
+    // tx where the this balance was observed
+    pub modify_tx: String,
```
louise-poole: This should be of type `Transaction`, no? I could be wrong - but if this is `Transaction` then it can be used to link a block to the tvl change and allow us to revert changes on reorgs.

tycho-indexer/src/models.rs
```diff
@@ -92,19 +93,39 @@ impl ExtractionState {
...
+    // Blockchain the component belongs to
+    pub chain: Chain,
     // holds the tokens tradable
-    tokens: Vec<T>,
-    // allows to express some validation over the attributes if necessary
-    attribute_schema: Bytes,
+    pub tokens: Vec<T>,
+    // ID's referring to related contracts
+    pub contract_ids: Vec<ContractId>,
+    // allows to express some validation over the static attributes if necessary
```
louise-poole: This comment is still incorrect. We're not doing any validation here, just storing static attributes.

tycho-indexer/src/models.rs
SmartMonkey-git: But if this is the transaction hash, then you can do the same, no?

proto/tycho/evm/v1/vm.proto
SmartMonkey-git: @pistomat I trust you

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -397,6 +401,86 @@ impl AccountUpdateWithTx {
...
+        })
+    }
+}
+
+impl ProtocolComponent<String> {
+    pub fn try_from_message(
+        msg: substreams::ProtocolComponent,
+        protocol_system: ProtocolSystem,
+        protocol_type: ProtocolType,
+        chain: Chain,
```
dianacarvalho1, line 460: why are these passed here and not contained inside the ProtocolComponent message?

louise-poole: I would think it's because we have a separate extractor per protocol which means the extractor already knows the chain, protocolSystem etc and does not need it passed in the message.

tycho-indexer/src/models.rs
louise-poole: Very true - then I think your type here can be `H256` instead of String?

louise-poole: COMMENT #100! 🥳 🎈

tycho-indexer/src/models.rs
SmartMonkey-git: I think alans suggestion was to only rename the message 
https://github.com/propeller-heads/tycho-indexer/pull/61#discussion_r1423220841

PR #64, opened 2023-12-13
ENG-2039: Detect new Ambient token pairs.

tamaralipows:
[Issue](https://datarevenue.atlassian.net/jira/software/projects/ENG/boards/24?selectedIssue=ENG-2039)

I closely followed the calls of [this function](https://github.com/CrocSwap/croc-subgraph/blob/main/subgraphs/croc-mainnet/src/mapping.ts#L332) 

- [x] Debug substreams issue with ethers: switched back to ethabi
- [x] Add `attributes` to ProtocolComponent in `vm.proto` and include `pool_index`
- [x] Properly name the protocol component id
- [x] Ensure correct results by running locally (found several created pools)
- [x] What should I set the contracts attribute to? Just the ambient contract?

**Notes**
- Can't use uuid in the substreams. Doesn't compile to wasm. I thought an okay way to overcome this is to use the transaction hash instead as the pool id.
```
2023-12-14T18:40:39.770582Z ERROR run:extractor_runner::run{id=ethereum:vm:ambient}: tycho_indexer::substreams::stream: Received tonic error status: Internal, message: "error during init_stores_and_backprocess: run_parallel_process failed: parallel processing run: scheduler run: receiving stream resp: rpc error: code = Unknown desc = internal: step new irr: handler step new: execute modules: applying executor results \"map_changes\": execute: maps wasm call: block 17361000: module \"map_changes\": general wasm execution failed: could not instantiate wasm module: \"__getrandom_custom\" is not exported in module \"env\"", details:
```

**Question:**
Some pools are included which have a base token of ZERO_ADDRESS, like the [pool created at this tx](https://etherscan.io/tx/0x288248a6a2d4fc551cc28391597346f26364151074c75d824495aff337291a47), for example. Are we okay with this? We hypothesized that this could be a native token pool, but are not sure.

**I made a commit mess, sorry**
Please review with all commits together except for aedf6ee - re-generate protobuf.
Reviewing the remaining commits separately might be confusing. I don't know what happened here. Has anyone else had problems automatically squashing commits using rust rover?

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -77,3 +77,15 @@ jobs:
           RUSTFLAGS: -Dwarnings
 
       - run: cargo +nightly fmt --all --check
+
+  buf_formatting:
+    name: Protobuf Files Formatting
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - uses: bufbuild/buf-setup-action@v1.28.1
+
+      - uses: bufbuild/buf-lint-action@v1
+          
--- a/Readme.md
+++ b/Readme.md
@@ -154,6 +154,17 @@ buf export buf.build/streamingfast/substreams -o ./proto
 buf generate 
 ```
 
+##### Protobuf Linting
+1. To install the linter execute:
+```
+brew install bufbuild/buf/buf@1.28.1
+```
+2. Linting the files works like:
+```
+buf lint
+```
+It will return a list of errors that you need to fix. If not, then all the changes you made conform to the linter.
+
 ### Postgres & Diesel
 1. If you are on a mac, you might need to install postgres library first and add it to the library path:
 ```bash
--- a/proto/buf.yaml
+++ b/proto/buf.yaml
@@ -4,4 +4,4 @@ breaking:
     - FILE
 lint:
   use:
-    - DEFAULT
+    - BASIC
--- a/proto/sf/substreams/internal/v2/deltas.proto
+++ b/proto/sf/substreams/internal/v2/deltas.proto
@@ -4,7 +4,7 @@ package sf.substreams.internal.v2;
 
 import "google/protobuf/any.proto";
 
-option go_package = "github.com/streamingfast/substreams/pb/sf/substreams/intern/v2;pbssinternal";
+option go_package = "github.com/streamingfast/substreams/pb/sf/substreams/internal/v2;pbssinternal";
 
 message StoreDeltas {
   repeated StoreDelta store_deltas = 1;
--- a/proto/sf/substreams/internal/v2/service.proto
+++ b/proto/sf/substreams/internal/v2/service.proto
@@ -2,10 +2,9 @@ syntax = "proto3";
 
 package sf.substreams.internal.v2;
 
-import "google/protobuf/any.proto";
 import "sf/substreams/v1/modules.proto";
 
-option go_package = "github.com/streamingfast/substreams/pb/sf/substreams/intern/v2;pbssinternal";
+option go_package = "github.com/streamingfast/substreams/pb/sf/substreams/internal/v2;pbssinternal";
 
 service Substreams {
   rpc ProcessRange(ProcessRangeRequest) returns (stream ProcessRangeResponse);
--- a/proto/sf/substreams/sink/service/v1/service.proto
+++ b/proto/sf/substreams/sink/service/v1/service.proto
@@ -2,7 +2,6 @@ syntax = "proto3";
 
 package sf.substreams.sink.service.v1;
 
-import "sf/substreams/options.proto";
 import "sf/substreams/v1/package.proto";
 
 option go_package = "github.com/streamingfast/substreams/pb/sf/substreams/sink/service/v1;pbsinksvc";
--- a/proto/tycho/evm/v1/vm.proto
+++ b/proto/tycho/evm/v1/vm.proto
@@ -78,10 +78,10 @@ message StateChanges {
 message TransactionStateChanges {
   Transaction tx = 1;
   repeated StateChanges state_changes = 2;
+  repeated ProtocolComponent components = 3;
 }
 
 message BlockEntityChanges {
   Block block = 1;
   repeated TransactionStateChanges changes = 2;
-  // TODO: add component changes
 }
--- a/scripts/download-ethereum-block-to-s3.sh
+++ b/scripts/download-ethereum-block-to-s3.sh
@@ -14,7 +14,6 @@ CLEAN_JSON=$(echo "$OUTPUT" | perl -0777 -pe 's/^.*?({.*}).*$/$1/s')
 
 # Extract the data from "@data" key using jq
 DATA=$(echo "$CLEAN_JSON" | jq -r '.["@data"]')
-echo $DATA
 
 echo "$DATA" | aws s3 cp - s3://defibot-data/test-assets/tycho/block_$BLOCK_NUMBER.json
 
--- a/tycho-indexer/src/extractor/evm/ambient.rs
+++ b/tycho-indexer/src/extractor/evm/ambient.rs
@@ -1,4 +1,8 @@
 #![allow(unused_variables)]
+
+use std::{str::FromStr, sync::Arc};
+
+use async_trait::async_trait;
 use diesel_async::{
     pooled_connection::deadpool::Pool, scoped_futures::ScopedFutureExt, AsyncConnection,
     AsyncPgConnection,
@@ -6,22 +10,20 @@ use diesel_async::{
 use ethers::types::{H160, H256};
 use mockall::automock;
 use prost::Message;
-use std::{collections::HashMap, str::FromStr, sync::Arc};
+use tokio::sync::Mutex;
 use tracing::{debug, info, instrument};
 
-use async_trait::async_trait;
-use tokio::sync::Mutex;
-
-use super::EVMStateGateway;
 use crate::{
     extractor::{evm, ExtractionError, Extractor, ExtractorMsg},
-    models::{Chain, ExtractionState, ExtractorIdentity},
+    models::{Chain, ExtractionState, ExtractorIdentity, ProtocolSystem},
     pb::{
         sf::substreams::rpc::v2::{BlockScopedData, BlockUndoSignal, ModulesProgress},
         tycho::evm::v1::BlockContractChanges,
     },
     storage::{BlockIdentifier, BlockOrTimestamp, StorageError},
 };
+
+use super::EVMStateGateway;
 
 const AMBIENT_CONTRACT: [u8; 20] = hex_literal::hex!("aaaaaaaaa24eeeb8d57d431224f73832bc34f688");
 
@@ -33,6 +35,7 @@ pub struct AmbientContractExtractor<G> {
     gateway: G,
     name: String,
     chain: Chain,
+    protocol_system: ProtocolSystem,
     // TODO: There is not reason this needs to be shared
     // try removing the Mutex
     inner: Arc<Mutex<Inner>>,
@@ -94,7 +97,7 @@ impl AmbientPgGateway {
         Ok(())
     }
 
-    #[instrument(skip_all, fields(chain = %self.chain, name = %self.name, block_number = %changes.block.number))]
+    #[instrument(skip_all, fields(chain = % self.chain, name = % self.name, block_number = % changes.block.number))]
     async fn forward(
         &self,
         changes: &evm::BlockStateChanges,
@@ -134,7 +137,7 @@ impl AmbientPgGateway {
         Result::<(), StorageError>::Ok(())
     }
 
-    #[instrument(skip_all, fields(chain = %self.chain, name = %self.name, block = ?to))]
+    #[instrument(skip_all, fields(chain = % self.chain, name = % self.name, block = ? to))]
     async fn backward(
         &self,
         to: &BlockIdentifier,
@@ -162,13 +165,16 @@ impl AmbientPgGateway {
         self.save_cursor(new_cursor, conn)
             .await?;
 
-        let changes = evm::BlockAccountChanges {
-            chain: self.chain,
-            extractor: self.name.clone(),
+        let changes = evm::BlockAccountChanges::new(
+            &self.name,
+            self.chain,
             block,
             account_updates,
-            new_pools: HashMap::new(),
-        };
+            // TODO: get protocol components from gateway (in ENG-2049)
+            Vec::new(),
+            Vec::new(),
+            Vec::new(),
+        );
         Result::<evm::BlockAccountChanges, StorageError>::Ok(changes)
     }
 
@@ -188,7 +194,7 @@ impl AmbientGateway for AmbientPgGateway {
         self.get_last_cursor(&mut conn).await
     }
 
-    #[instrument(skip_all, fields(chain = %self.chain, name = %self.name, block_number = %changes.block.number))]
+    #[instrument(skip_all, fields(chain = % self.chain, name = % self.name, block_number = % changes.block.number))]
     async fn upsert_contract(
         &self,
         changes: &evm::BlockStateChanges,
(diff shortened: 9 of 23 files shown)


substreams/ethereum-ambient/src/pb/tycho.evm.v1.rs
```diff
@@ -2,93 +2,128 @@
 #[allow(clippy::derive_partial_eq_without_eq)]
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct Block {
-    #[prost(bytes = "vec", tag = "1")]
+    #[prost(bytes="vec", tag="1")]
```
SmartMonkey-git: Hmmm.... my rustfmt formats that differently. :/
Do you have the nightly version selected?
image

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -99,6 +99,91 @@ fn map_changes(
             .collect::<Vec<_>>();
         storage_changes.sort_unstable_by_key(|change| change.ordinal);
 
+        // Extract token pair creations
+        const INIT_POOL_CODE: u8 = 71;
+
+        if block_tx.to == AMBIENT_CONTRACT {
+            let user_cmd: String = "a15112f9".to_string();
```
SmartMonkey-git: Whats a user cmd?
And don't you want to make it a constant like you did with `INIT_POOL_CODE`?

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -99,6 +99,91 @@ fn map_changes(
...
+                ];
+
+                // Decode external call to UserCmd
+                if let Ok(external_params) =
+                    decode(user_cmd_external_abi_types, &block_tx.input[4..])
+                {
+                    let cmd_bytes = match &external_params[1] {
+                        Token::Bytes(bytes) => bytes.clone(),
+                        _ => {
+                            panic!("Unexpected type for cmd_bytes: {:?}", &external_params[1]);
```
SmartMonkey-git: I'm not sure if you actually want to panic in all these cases. A panic is different from a raise in python, if you panic the program is going down, you can not catch a panic.
I think what you are looking for is the `Err` macro. Which is the opposite of the `Ok` macro. If you return either of them from a function they will be returned as a `Result` object.

In this case, I THINK, you want to write:
```
return Err(anyhow!(
    "Unexpected type for base: {:?}",
    &internal_params[1]
));
```
However, `anyhow!` would be another lib we need to include. Dont know if thats ok.
If you do `⌘+left mouse` on `Error` in `Result<tycho::BlockContractChanges, substreams::errors::Error>` you will see some documentation on how to throw errors in the `map_changes` function. Maybe you find a way to not use anyhow, if we do not want it.

SmartMonkey-git: I would say this goes for all panics, I dont think we want them.

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -99,6 +99,91 @@ fn map_changes(
...
+
+                            let quote = match &internal_params[2] {
+                                Token::Address(addr) => addr.to_fixed_bytes().to_vec(),
+                                _ => {
+                                    panic!("Unexpected type for quote: {:?}", &internal_params[2]);
+                                }
+                            };
+
+                            let pool_index = match &internal_params[3] {
+                                Token::Uint(uint) => uint.as_u64() as u32,
```
SmartMonkey-git: Why not?
```suggestion
                                Token::Uint(uint) => uint.as_u32(),
```

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -99,6 +99,91 @@ fn map_changes(
...
         storage_changes.sort_unstable_by_key(|change| change.ordinal);
 
+        // Extract token pair creations
+        const INIT_POOL_CODE: u8 = 71;
+
+        if block_tx.to == AMBIENT_CONTRACT {
+            let user_cmd: String = "a15112f9".to_string();
+            let block_tx_hex_string: String = hex::encode(&block_tx.input)[0..8].to_string();
+            if block_tx_hex_string == user_cmd {
+                let user_cmd_external_abi_types = &[
```
SmartMonkey-git, line 117: Why not declare them as constants as well?

I think I would actually prefer these constants in the global scope, no? Like we did in the good old day's in python

https://doc.rust-lang.org/rust-by-example/custom_types/constants.html

substreams/ethereum-ambient/src/lib.rs
SmartMonkey-git: Or maybe you can leave out the match entirely:
```
let pool_index = internal_params[3]
                                .to_owned()
                                .into_uint()
                                .ok_or_else(|| anyhow!("Failed to convert to u32".to_string()))?
                                .as_u32();
```

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -99,6 +99,91 @@ fn map_changes(
...
+                    ParamType::Uint(256), // pool index
+                    ParamType::Uint(128), // price
+                ];
+
+                // Decode external call to UserCmd
+                if let Ok(external_params) =
+                    decode(user_cmd_external_abi_types, &block_tx.input[4..])
+                {
+                    let cmd_bytes = match &external_params[1] {
+                        Token::Bytes(bytes) => bytes.clone(),
```
dianacarvalho1: why do we need this match statement? how would this not match? (same question for the other matches below)

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -99,6 +99,91 @@ fn map_changes(
...
+
+                            let static_attribute = tycho::Attribute {
+                                name: String::from("pool_index")
+                                    .as_bytes()
+                                    .to_vec(),
+                                value: pool_index.to_be_bytes().to_vec(),
+                            };
+
+                            let new_component = tycho::ProtocolComponent {
+                                id: block_tx.hash.clone(),
```
dianacarvalho1: maybe the ID could be something like "ETH-DAI", but at this point we don't know that right? 
can't we know?

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -99,6 +99,91 @@ fn map_changes(
...
+                                tokens: vec![base, quote],
+                                contracts: vec![],
```
dianacarvalho1: huumm so yeah. I think this is not clear yet what is gonna go in contracts :/

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -99,6 +99,91 @@ fn map_changes(
             .collect::<Vec<_>>();
         storage_changes.sort_unstable_by_key(|change| change.ordinal);
 
+        // Extract token pair creations
```
kayibal: I don't understand how this comment, relates to the constant INIT_POOL_CODE. Maybe constants could also be defined on module level instead?

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -99,6 +99,91 @@ fn map_changes(
...
+        const INIT_POOL_CODE: u8 = 71;
+
+        if block_tx.to == AMBIENT_CONTRACT {
+            let user_cmd: String = "a15112f9".to_string();
+            let block_tx_hex_string: String = hex::encode(&block_tx.input)[0..8].to_string();
```
kayibal: Please work with bytes not with hex encoded strings, this is much more precise.

You basically allocate memory here for a new string, then hex encode the bytes into the string, next you create a string slice of the first 8 characters and then allocate memory for that again and put it once again in a new string. Sounds complicated no?

Afaik `block_tx.input` is already a vector of uint8 (so it is a byte array). Now you basically want a constant uint8 array of size 4 with these contents: a15112f9, you can then simply use that to compare to `block_tx.input[0...4]`

substreams/ethereum-ambient/src/lib.rs
kayibal: Yes this should return an error instead

substreams/ethereum-ambient/src/lib.rs
kayibal: I think `if let Token::Bytes(bytes) = ...` pattern matching would be better here. Actually clippy should complain about this code.

There are more of these unnecessary matches further down

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -99,6 +99,91 @@ fn map_changes(
...
+                                    );
+                                }
+                            };
+
+                            let static_attribute = tycho::Attribute {
+                                name: String::from("pool_index")
+                                    .as_bytes()
+                                    .to_vec(),
+                                value: pool_index.to_be_bytes().to_vec(),
+                            };
```
kayibal: Interesting that the name here is bytes oO, can I get a link to the PR that "decided" this?

substreams/ethereum-ambient/src/lib.rs
kayibal: This won't work if a transaction creates two pools, I see that the original indexer computes a poolHash from base address, quote address and some poolIndex. Please use these 3 values to create a unique id for the pool (doen't need to be hash, you can simply concatenate them for example)

substreams/ethereum-ambient/src/lib.rs
dianacarvalho1: aaah @louise-poole to the rescue! we should pass the ambient mega contract address here!

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -99,6 +99,91 @@ fn map_changes(
             .collect::<Vec<_>>();
         storage_changes.sort_unstable_by_key(|change| change.ordinal);
 
+        // Extract token pair creations
+        const INIT_POOL_CODE: u8 = 71;
+
+        if block_tx.to == AMBIENT_CONTRACT {
```
kayibal: You are only checking the first call of the transaction here, you have to also check all internal calls. You'll have to study the structure of Block to find out how to iterate over all internal calls.

substreams/ethereum-ambient/src/lib.rs
kayibal: The contracts that are involved in the swap logic of this component (so the ambient contract and nothing else)

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -99,6 +99,91 @@ fn map_changes(
...
+                            let static_attribute = tycho::Attribute {
+                                name: String::from("pool_index")
+                                    .as_bytes()
+                                    .to_vec(),
+                                value: pool_index.to_be_bytes().to_vec(),
+                            };
+
+                            let new_component = tycho::ProtocolComponent {
+                                id: block_tx.hash.clone(),
+                                tokens: vec![base, quote],
```
kayibal: I think we should sort the tokens by their address.

substreams/ethereum-ambient/src/lib.rs
louise-poole: For the protobuf message? I did it in my PR, but did it in response to other String message fields (such as `component_id`, `contract_id` etc) being set as `bytes` and not `string` and I followed suit. It was later discussed in this PR: https://github.com/propeller-heads/tycho-indexer/pull/61 (I unresolved the discussion so it should be easy to spot). I think this discussion was marked resolved before I saw the conclusion though and my PR was merged before this one. 
TDLR: we should update the types to string for Attribute's `name` and ProtocolComponent's `id`  and subsequently also the `component_id` in StateChanges. I can make a PR for this quick.

louise-poole: Here are the updated messages: https://github.com/propeller-heads/tycho-indexer/pull/68

substreams/ethereum-ambient/src/pb/tycho.evm.v1.rs
tamaralipows: Screenshot 2023-12-15 at 09 22 09
idk whats up here, looks the same

substreams/ethereum-ambient/src/lib.rs
tamaralipows: I've put these in the same place as `AMBIENT_ADDRESS` is. Let me know if that's okay!

substreams/ethereum-ambient/src/lib.rs
tamaralipows: Idk why my clippy doesn't complain here :/ if there is a problem it normally does.  Are you suggesting the matches are useless because of what @dianacarvalho1 mentioned? Or is there a bigger problem with match, such that it is less concise? I just reread the chapter about match and I'm leaning towards the second one?

I agree with @dianacarvalho1 and I would say that an `expect` would be enough here, since the abi decoding should fail long before if something is wrong with the encoding. I did it the way that @SmartMonkey-git suggested using `.ok_or_else(|| {` but I'm not fully convinced. Let me know your opinions.

substreams/ethereum-ambient/src/lib.rs
tamaralipows: This is what is it called in the ambient contract. I've renamed it, let me know how that looks to you!

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -99,6 +102,107 @@ fn map_changes(
...
+                                contracts: vec![hex::encode(AMBIENT_CONTRACT)],
+                                static_att: vec![static_attribute],
+                            };
+                            tx_change.components.push(new_component);
+                        } else {
+                            return Err(anyhow!("Failed to decode ABI internal call.".to_string()));
+                        }
+                    }
+                } else {
+                    return Err(anyhow!("Failed to decode ABI external call.".to_string()));
```
tamaralipows: @SmartMonkey-git is this the right way to "return an error"? Seems fishy to me. I tried to just leave it as `Err(` without the return but got a compile error:

`^^^ cannot infer type of the type parameter `T` declared on the enum `Result` `

SmartMonkey-git: Interesting.
I had a look into it, it seems like the right use of the `anyhow` library in this case is to use the [bail! macro](https://docs.rs/anyhow/latest/anyhow/macro.bail.html).

```suggestion
                    bail!("Failed to decode ABI external call.".to_string());
```

There is also an [ensure! macro](https://docs.rs/anyhow/latest/anyhow/macro.ensure.html) that lets you return early by checking a condition. Just so you know.

substreams/ethereum-ambient/src/pb/tycho.evm.v1.rs
pistomat: I think the latest nightly version changed the formatting, try updating local rust with `rustup update`

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -99,6 +102,107 @@ fn map_changes(
...
+        let ambient_calls = block_tx
+            .calls
+            .iter()
+            .filter(|call| !call.state_reverted)
+            .filter(|call| call.address == AMBIENT_CONTRACT)
+            .collect::<Vec<_>>();
+
+        for call in ambient_calls {
+            if call.input.len() < 4 {
+                continue;
```
dianacarvalho1, line 114: what does it mean in this case?

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -99,6 +102,107 @@ fn map_changes(
...
+                    // Call data is structured differently depending on the cmd code, so only
+                    // decode if this is an init pool code.
+                    if cmd_bytes[31] == INIT_POOL_CODE {
+                        // Decode internal call to UserCmd
+                        if let Ok(internal_params) = decode(user_cmd_internal_abi_types, &cmd_bytes)
+                        {
+                            let base = internal_params[1]
+                                .to_owned()
+                                .into_address()
+                                .ok_or_else(|| {
```
dianacarvalho1, line 149: ok or else sounds so threatening 😂

substreams/ethereum-ambient/src/lib.rs
tamaralipows: I think it's just like the python continue. Move to the next loop element if condition satisfied

dianacarvalho1: no no, I mean like what does it mean for it to have a len() < 4?

tamaralipows: Good question. I don't know why this happens.
```
execute: maps wasm call: block 17361664: module \"map_changes\": wasm execution failed deterministically: panic in the wasm: \"range end index 4 out of range for slice of length 0\"
```

I can't find the tx, but one tx in that block has a call to the ambient contract without any calldata. Could be a number of reasons... not sure it's worth narrowing it down.

PR #65, opened 2023-12-13
ENG-2043: Extend BlockAccountChanges

dianacarvalho1:
[Issue](https://datarevenue.atlassian.net/jira/software/projects/ENG/boards/24?selectedIssue=ENG-2043)

- Add missing derives and defaults to structs 
- Use protocol_type_id in ProtocolComponent instead of the complete object 
- Add new_protocol_components, deleted_protocol_components and tvl_changes
- Adapt AmbientContractExtractor to handle components

Note: remember we decided to skip tvl_changes for now so I'm not actually doing anything with them.

TODO:
- Run whole system and see if the BlockAccountChanges are emitted properly

### DIFF
--- a/tycho-indexer/src/extractor/evm/mod.rs
+++ b/tycho-indexer/src/extractor/evm/mod.rs
@@ -999,7 +999,7 @@ mod test {
     use crate::extractor::evm::fixtures::transaction01;
 
     use super::*;
-    use crate::{models::ProtocolSystem, pb::tycho::evm::v1::Attribute};
+    use crate::models::ProtocolSystem;
     use actix_web::body::MessageBody;
     use rstest::rstest;
 


tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -225,7 +222,31 @@ pub struct BlockAccountChanges {
...
+
+impl BlockAccountChanges {
+    pub fn new(
+        extractor: &str,
+        chain: Chain,
+        block: Block,
+        account_updates: HashMap<H160, AccountUpdate>,
+        new_components: Vec<ProtocolComponent>,
+        deleted_protocol_components: Vec<ProtocolComponent>,
+        tvl_change: HashMap<String, TvlChange>,
```
SmartMonkey-git: What are we mapping here?
Component to TvlChange, because the TvlChange already has a component id

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -1116,6 +1140,24 @@ mod test {
...
+                ContractId("DIANA-THALES".to_string()),
+                ContractId("0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2".to_string()),
+            ],
+            static_attributes: [
+                ("key1".to_string(), Bytes::from(b"value1".to_vec())),
+                ("key2".to_string(), Bytes::from(b"value2".to_vec())),
+            ]
+            .iter()
+            .cloned()
+            .collect(),
```
SmartMonkey-git: Why do you need to clone them here?

tycho-indexer/src/extractor/evm/mod.rs
dianacarvalho1: I can make it a vec as well.. tbh this was what was on the plan. not sure why and I don't see an obvious reason for it rn

tycho-indexer/src/extractor/evm/mod.rs
dianacarvalho1: full disclosure: this cloning logic is not 1005 clear to me. but without it I get
```

error[E0277]: a value of type `std::collections::HashMap<std::string::String, hex_bytes::Bytes>` cannot be built from an iterator over elements of type `&(std::string::String, hex_bytes::Bytes)`
    --> tycho-indexer/src/extractor/evm/mod.rs:1158:14
     |
1158 |             .collect(),
     |              ^^^^^^^ value of type `std::collections::HashMap<std::string::String, hex_bytes::Bytes>` cannot be built from `std::iter::Iterator<Item=&(std::string::String, hex_bytes::Bytes)>`
     |
     = help: the trait `FromIterator<&(std::string::String, hex_bytes::Bytes)>` is not implemented for `std::collections::HashMap<std::string::String, hex_bytes::Bytes>`
     = help: the trait `FromIterator<(K, V)>` is implemented for `std::collections::HashMap<K, V, S>`
note: the method call chain might not have had the expected associated types
    --> tycho-indexer/src/extractor/evm/mod.rs:1157:14
     |
1153 |               static_attributes: [
     |  ________________________________-
1154 | |                 ("key1".to_string(), Bytes::from(b"value1".to_vec())),
1155 | |                 ("key2".to_string(), Bytes::from(b"value2".to_vec())),
1156 | |             ]
     | |_____________- this expression has type `[(String, Bytes); 2]`
1157 |               .iter()
     |                ^^^^^^ `Iterator::Item` is `&(String, Bytes)` here
note: required by a bound in `std::iter::Iterator::collect`
    --> /Users/dianacarvalho/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/iter/traits/iterator.rs:2049:19
     |
2049 |     fn collect<B: FromIterator<Self::Item>>(self) -> B
     |                   ^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `Iterator::collect`


```

SmartMonkey-git: When I remove 

```
.iter()
.cloned()
.collect(),
```
I get a quick fix suggestion from clippy to do this:
```
HashMap::from([
                ("key1".to_string(), Bytes::from(b"value1".to_vec())),
                ("key2".to_string(), Bytes::from(b"value2".to_vec())),
            ])
```
This works when I run the test, do you have clippy setup?

dianacarvalho1: ooohh. yes I do

tycho-indexer/src/extractor/evm/ambient.rs
```diff
@@ -281,7 +289,15 @@ where
 
         debug!(?raw_msg, "Received message");
 
-        let msg = match evm::BlockStateChanges::try_from_message(raw_msg, &self.name, self.chain) {
+        // TODO: figure out how/where to get this ID from (in ENG-2049)
+        let protocol_type_id = String::from("id-1");
```
tamaralipows, line 293: I have the same question for my task but at a different level lol

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -225,7 +222,31 @@ pub struct BlockAccountChanges {
...
+        extractor: &str,
+        chain: Chain,
+        block: Block,
+        account_updates: HashMap<H160, AccountUpdate>,
+        new_components: Vec<ProtocolComponent>,
+        deleted_protocol_components: Vec<ProtocolComponent>,
+        tvl_change: Vec<TvlChange>,
+    ) -> Self {
+        BlockAccountChanges {
+            extractor: extractor.to_owned(),
```
tamaralipows, line 242: I'm curious why we use this `.to_owned()` on an obejct instead of cloning it like we do in other places

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -225,7 +222,31 @@ pub struct BlockAccountChanges {
...
+}
+
+impl BlockAccountChanges {
+    pub fn new(
+        extractor: &str,
+        chain: Chain,
+        block: Block,
+        account_updates: HashMap<H160, AccountUpdate>,
+        new_components: Vec<ProtocolComponent>,
+        deleted_protocol_components: Vec<ProtocolComponent>,
```
tamaralipows: For consistency? 
```suggestion
        new_components: Vec<ProtocolComponent>,
        deleted_components: Vec<ProtocolComponent>,
```

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -499,7 +523,7 @@ impl ProtocolComponent {
 
         Ok(Self {
             id,
-            protocol_type,
+            protocol_type_id,
```
tamaralipows, line 532: What's the difference between a protocol type and a protocol type id?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -225,7 +222,31 @@ pub struct BlockAccountChanges {
...
     pub account_updates: HashMap<H160, AccountUpdate>,
-    pub new_pools: HashMap<H160, SwapPool>,
+    pub new_protocol_components: Vec<ProtocolComponent>,
+    pub deleted_protocol_components: Vec<ProtocolComponent>,
+    pub tvl_changes: Vec<TvlChange>,
+}
+
+impl BlockAccountChanges {
+    pub fn new(
+        extractor: &str,
```
tamaralipows, line 233: I'm still confused by we send the pointer to the string instead of the actual string here, maybe some rust guru can explain to me...

tycho-indexer/src/extractor/evm/ambient.rs
dianacarvalho1: We will handle this in the gtw (and consequently in the extractors). It's my next task. why do you need it for your task though?

tycho-indexer/src/extractor/evm/mod.rs
dianacarvalho1: oh you are right. let me fix it everywhere though

tycho-indexer/src/extractor/evm/mod.rs
dianacarvalho1: protocol type would be the complete object, while the id is just the reference to it

tycho-indexer/src/extractor/evm/mod.rs
dianacarvalho1: yeah it's also not clear to me. it's what clippy suggested

tycho-indexer/src/extractor/evm/mod.rs
dianacarvalho1: it's  used as an explicit way to convert borrowed type, &str, into owned type String

tycho-indexer/src/extractor/evm/ambient.rs
tamaralipows: Because the protocol component message has an id attribute

dianacarvalho1: that is just an ID for the component.. it can be anything, you can make it random I think

tycho-indexer/src/models.rs
```diff
@@ -20,26 +20,29 @@ pub enum Chain {
...
 pub enum ImplementationType {
+    #[default]
     Vm,
     Custom,
 }
 
-#[derive(PartialEq, Debug, Clone)]
+#[derive(PartialEq, Debug, Clone, Default, Deserialize, Serialize)]
 pub enum FinancialType {
+    #[default]
```
louise-poole, line 38: What does this line do? 👀

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -223,7 +223,31 @@ pub struct BlockAccountChanges {
...
+        deleted_protocol_components: Vec<ProtocolComponent>,
+        tvl_change: Vec<TvlChange>,
+    ) -> Self {
+        BlockAccountChanges {
+            extractor: extractor.to_owned(),
+            chain,
+            block,
+            account_updates,
+            new_protocol_components,
+            deleted_protocol_components,
```
louise-poole, line 247: Hmmmm. How do we know if it's an addition or a deletion? Will we get it from the message somehow? If so, we probably need these to be separate vecs on `BlockStateChanges` too as it's when we decode it that we have this info. Also, an idea is that instead of 2 vecs for components, we can have a `protocol_components` Hashmap with 2 keys `add` and `delete` that map to `Vec<ProtocolComponent>`?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -1111,6 +1141,21 @@ mod test {
...
+            tokens: vec!["token1".to_string(), "token2".to_string()],
+            contract_ids: vec![
+                ContractId("DIANA-THALES".to_string()),
+                ContractId("0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2".to_string()),
+            ],
+            static_attributes: HashMap::from([
+                ("key1".to_string(), Bytes::from(b"value1".to_vec())),
+                ("key2".to_string(), Bytes::from(b"value2".to_vec())),
+            ]),
+        };
```
louise-poole, line 1158: Would be useful to make this a test fixture so that it can be used in multiple tests - I see it's used twice here and in my PR it's used a again a few times. I made a test fixture in my PR so can actually update this for you when I merge after you?

tycho-indexer/src/models.rs
dianacarvalho1: defined the default value as the first one

tycho-indexer/src/extractor/evm/mod.rs
dianacarvalho1: no no, we can only have a deletion if there is a reorg of some blocks that had new components.

tycho-indexer/src/extractor/evm/mod.rs
dianacarvalho1: yeees thank yoouu 🙏🏼

tycho-indexer/src/extractor/evm/mod.rs
dianacarvalho1: so from substreams we will only ever get new components. it's only in a possible re org that we might have to delete components

tycho-indexer/src/models.rs
louise-poole: Ah, okay! This is enum specific. Nice thanks!

tycho-indexer/src/extractor/evm/mod.rs
louise-poole: Ah nice! Okay, I accept 👍

PR #67, opened 2023-12-14
ENG 2407: add protocol component to native protocol msgs

louise-poole:
What was done:
- add ProtocolComponent messages to BlockEntityChanges message. Update deserialisation and handing accordingly
- change `modify_tx` in ProtocolState struct to be a hash. This lead to the need for a ProtocolStateWithTx struct that can be used to merge and aggregate state updates with validation checks on the tx that modified the states updated.

### DIFF
--- a/tycho-indexer/src/extractor/evm/mod.rs
+++ b/tycho-indexer/src/extractor/evm/mod.rs
@@ -710,7 +710,7 @@ pub struct ProtocolStatesWithTx {
 }
 
 impl ProtocolStatesWithTx {
-    /// Parses protocol state from tychos protobuf TransactionStateChanges message
+    /// Parses protocol state from tychos protobuf StateChanges message
     pub fn try_from_message(
         msg: Vec<substreams::StateChanges>,
         tx: Transaction,
@@ -784,7 +784,7 @@ pub struct BlockEntityChangesResult {
     chain: Chain,
     pub block: Block,
     pub state_updates: HashMap<String, ProtocolState>,
-    pub new_pools: HashMap<String, ProtocolComponent>,
+    pub new_protocol_components: HashMap<String, ProtocolComponent>,
 }
 
 /// A container for state updates grouped by transaction
@@ -797,7 +797,7 @@ pub struct BlockEntityChanges {
     chain: Chain,
     pub block: Block,
     pub state_updates: Vec<ProtocolStatesWithTx>,
-    pub new_pools: HashMap<String, ProtocolComponent>,
+    pub new_protocol_components: HashMap<String, ProtocolComponent>,
 }
 
 // TODO: remove dead code check skip once extractor is implemented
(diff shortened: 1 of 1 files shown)


proto/tycho/evm/v1/vm.proto
```diff
@@ -78,10 +78,10 @@ message StateChanges {
 message TransactionStateChanges {
   Transaction tx = 1;
   repeated StateChanges state_changes = 2;
+  repeated ProtocolComponent components = 3;
```
zizou0x, line 81: Here we receive and update or a change right? If yes, I feel it could be great to add this aspect to the name

louise-poole: No, we only receive new pools here. ProtocolComponents are static. If the dynamic attributes within the pool change, we will receive that as a StateChange to indicate that. We will only ever receive 1 ProtocolComponent message per pool created.

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -678,30 +678,97 @@ impl ProtocolState {
...
+
+/// Updates grouped by their respective transaction.
+#[derive(Debug, Clone, PartialEq, Default)]
+pub struct ProtocolStatesWithTx {
+    pub protocol_states: HashMap<String, ProtocolState>,
+    pub tx: Transaction,
+}
+
+impl ProtocolStatesWithTx {
+    /// Parses protocol state from tychos protobuf TransactionStateChanges message
```
dianacarvalho1: it's called only `StateChanges´ no?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -678,30 +678,97 @@ impl ProtocolState {
...
+///
+/// Hold a single update per component. This is a condensed form of
+/// [BlockEntityChanges].
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, Default)]
+pub struct BlockEntityChangesResult {
+    extractor: String,
+    chain: Chain,
+    pub block: Block,
+    pub state_updates: HashMap<String, ProtocolState>,
+    pub new_pools: HashMap<String, ProtocolComponent>,
```
dianacarvalho1: Maybe `new_protocol_components` is more general than `new_pools`?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -723,27 +790,39 @@ impl BlockEntityChanges {
         msg: substreams::BlockEntityChanges,
         extractor: &str,
         chain: Chain,
+        protocol_system: ProtocolSystem,
+        protocol_type: ProtocolType,
```
dianacarvalho1: In this [PR](https://github.com/propeller-heads/tycho-indexer/pull/65), I changed the `protocol_type` to a `protocol_type_id`

proto/tycho/evm/v1/vm.proto
zizou0x: Ahh I see, ok I'm still confused with the naming sorry 😅

PR #69, opened 2023-12-18
ENG-2385: implement CachedGateway

zizou0x:
This PR's goal is to run Tycho with Ambient extractor running with the CachedGateway. CachedGateway and EVMGateway have different interfaces so I also had to do some modification inside Ambient extractor. 

Please ignore modifications on `tycho.evm.v1.rs`.

### DIFF
--- a/tycho-indexer/src/storage/postgres/cache.rs
+++ b/tycho-indexer/src/storage/postgres/cache.rs
@@ -377,7 +377,7 @@ impl CachedGateway {
             .expect("Send message to receiver ok");
 
         match rx.await {
-            Ok(result) => result, // Flatten the Result here
+            Ok(result) => result,
             Err(_) => Err(StorageError::WriteCacheGoneAway()),
         }
     }
@@ -394,7 +394,7 @@ impl CachedGateway {
             .await
             .expect("Send message to receiver ok");
         match rx.await {
-            Ok(result) => result, // Flatten the Result here
+            Ok(result) => result,
             Err(_) => Err(StorageError::WriteCacheGoneAway()),
         }
     }
@@ -411,7 +411,7 @@ impl CachedGateway {
             .await
             .expect("Send message to receiver ok");
         match rx.await {
-            Ok(result) => result, // Flatten the Result here
+            Ok(result) => result,
             Err(_) => Err(StorageError::WriteCacheGoneAway()),
         }
     }
@@ -428,7 +428,7 @@ impl CachedGateway {
             .await
             .expect("Send message to receiver ok");
         match rx.await {
-            Ok(result) => result, // Flatten the Result here
+            Ok(result) => result,
             Err(_) => Err(StorageError::WriteCacheGoneAway()),
         }
     }
@@ -445,7 +445,7 @@ impl CachedGateway {
             .await
             .expect("Send message to receiver ok");
         match rx.await {
-            Ok(result) => result, // Flatten the Result here
+            Ok(result) => result,
             Err(_) => Err(StorageError::WriteCacheGoneAway()),
         }
     }
@@ -457,7 +457,7 @@ impl CachedGateway {
             .await
             .expect("Send message to receiver ok");
         match rx.await {
-            Ok(result) => result, // Flatten the Result here
+            Ok(result) => result,
             Err(_) => Err(StorageError::WriteCacheGoneAway()),
         }
     }


tycho-indexer/src/main.rs
```diff
@@ -120,8 +116,30 @@ async fn main() -> Result<(), ExtractionError> {
 
     info!("Starting Tycho");
     let mut extractor_handles = Vec::new();
+    let (tx, rx) = mpsc::channel(10);
+    let (err_tx, mut err_rx) = mpsc::channel(10);
+
+    // Spawn a new task that listens to err_rx
+    task::spawn(async move {
+        while let Some(msg) = err_rx.recv().await {
+            //TODO: panic?
```
zizou0x, line 125: What should we do if we receive an error from the `DBCacheWriteExecutor`? Panic?

tycho-indexer/src/extractor/evm/ambient.rs
```diff
@@ -618,66 +640,89 @@ mod gateway_test {
...
+    //             None,
+    //             true,
+    //             &mut conn,
+    //         )
+    //         .await
+    //         .expect("test successfully retrieved ambient contract");
+    //     assert_eq!(account, exp_account);
+    //     // Assert no error happened
+    //     assert_eq!(maybe_err, Empty);
+    // }
```
zizou0x: This test is stuck for the reason I explained in https://github.com/propeller-heads/tycho-indexer/pull/76#discussion_r1431708684

tycho-indexer/src/main.rs
kayibal: Ideally the extractors affected should be stopped, we inform clients that this extractor is down, send an alert and continue processing the remaining extractors as if nothing had happened.

tycho-indexer/src/extractor/evm/ambient.rs
```diff
@@ -74,26 +74,18 @@ pub trait AmbientGateway: Send + Sync {
...
-        new_cursor: &str,
-        conn: &mut AsyncPgConnection,
-    ) -> Result<(), StorageError> {
+    async fn save_cursor(&self, block: &Block, new_cursor: &str) -> Result<(), StorageError> {
         let state =
             ExtractionState::new(self.name.to_string(), self.chain, None, new_cursor.as_bytes());
-        self.state_gateway
-            .save_state(&state, conn)
-            .await?;
+        let _ = self
```
tvinagre: Why not handling the error?

tycho-indexer/src/main.rs
tvinagre: Hmm yeah I think so, so we can restart and replay the block. It's crucial we don't skip a block so it makes sense to panic IMO

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -107,27 +119,31 @@ impl DBCacheWriteExecutor {
...
+                        self.write(db_tx).await;
                     }
                     DBCacheMessage::Flush(sender) => {
                         // Flush the current state and send back the result
                         let flush_result = self.flush().await;
                         let _ = sender.send(Ok(flush_result));
                     }
+                    DBCacheMessage::Revert(to, sender) => {
+                        let revert_result = self.revert(&to).await;
+                        let _ = sender.send(revert_result);
```
tvinagre, line 148: No error handling on the 3 cases?

tvinagre: Ah, ok, errors are not raised, they're sent to the channel, right?

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -144,66 +160,98 @@ impl DBCacheWriteExecutor {
...
+                        .transaction(|conn| {
+                            async {
+                                for op in new_db_tx.operations {
+                                    match self.execute_write_op(op, conn).await {
+                                        Err(StorageError::DuplicateEntry(entity, id)) => {
+                                            // As this db transaction is old. It can contain already
+                                            // stored blocks or txs, we log the duplicate entry
+                                            // error and
+                                            // continue
+                                            debug!("Duplicate entry for {} with id {}", entity, id);
```
tvinagre: Nice

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -144,66 +160,98 @@ impl DBCacheWriteExecutor {
...
-                            Result::<(), StorageError>::Ok(())
+                            .scope_boxed()
+                        })
+                        .await
+                    {
+                        Ok(_) => {
+                            // Notify that this transaction was correctly send to the db
+                            new_db_tx
+                                .tx
+                                .send(Ok(()))
```
tvinagre: Does it make sense to notify in different places the outcome of the same operation?
We have one error_transmitter channel and each tx has its channel. Shouldn't we send the errors to the `new_db_tx.tx` channel as well?

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -144,66 +160,103 @@ impl DBCacheWriteExecutor {
...
+                            self.error_transmitter
+                                .send(e)
+                                .await
+                                .expect("Should successfully notify error");
+                        }
+                    }
+                } else if new_db_tx.block.parent_hash == pending.hash {
+                    debug!("New block received {} !", new_db_tx.block.parent_hash);
+                    // New database transaction for the next block, we flush and cache it
+                    self.flush().await;
```
tvinagre: In this case, db state will always be 1 block old, correct? Not sure how this is going to work in Tycho itself, since it consumes from the database.
@kayibal maybe you can help me understand

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -144,66 +160,103 @@ impl DBCacheWriteExecutor {
...
-                    return Err(StorageError::Unexpected("Invalid cache state!".into()));
+                    // Other cases send unexpected error
+                    self.error_transmitter
+                        .send(StorageError::Unexpected("Invalid cache state!".into()))
+                        .await
+                        .expect("Should successfully notify error");
                 }
             }
             None => {
+                // This case happens at the start or after a Flush message
```
tvinagre: Can you explain more? I didn't understand

tycho-indexer/src/extractor/evm/ambient.rs
zizou0x: This was a mistake from a previous implementation, I changed it

tycho-indexer/src/storage/postgres/cache.rs
zizou0x: This is what's tricky here, `new_db_tx.tx` needs to be awaited by the sender (for example in `upsert_block`). But the moment when we try to send this db_tx to the database happens later, when we receive message from a new block. So most of the time, we can't use these channels to forward storage error because the errors happens a long time after the db_tx is sent and the sender can't await for it.

tycho-indexer/src/storage/postgres/cache.rs
zizou0x: `self.pending_block == None` can only happen when we start Tycho or after we received a `DBCacheMessage::Flush`. I will improve this comment

tycho-indexer/src/storage/postgres/cache.rs
zizou0x: Yes exactly

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -21,27 +21,37 @@ use crate::{
...
+        operations: Vec<WriteOp>,
+        tx: oneshot::Sender<Result<(), StorageError>>,
+    ) -> Self {
+        Self { block, operations, tx }
+    }
+}
+
 /// Represents different types of messages that can be sent to the DBCacheWriteExecutor.
-enum DBCacheMessage {
+pub enum DBCacheMessage {
```
kayibal: Same here it seems that this could be private to this module.

tycho-indexer/src/storage/postgres/cache.rs
kayibal: I think this will spam a lot, especially on block when an extractor is catching up (initial sync). I would suggest to simply filter block insertions if block.number < pending.number. It can still happen for duplicated transactions, in this case we should then ignore this error. An easier alternative would be to always ignore it.

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -144,72 +164,97 @@ impl DBCacheWriteExecutor {
...
+                        .expect("Flush should succeed");
                     self.pending_block = Some(new_db_tx.block);
-                    self.pending_db_txs.push(new_db_tx);
+
+                    self.pending_db_txs
+                        .append(&mut new_db_tx.operations);
+                    new_db_tx
+                        .tx
+                        .send(Ok(()))
+                        .expect("Should successfully notify sender");
```
kayibal: Same as above receiver of this may be gone at this point, that shouldn't crash this thread...

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -21,27 +21,37 @@ use crate::{
 
 /// Represents different types of database write operations.
 #[derive(PartialEq, Clone)]
-enum WriteOp {
+pub enum WriteOp {
```
kayibal: Can this be made private?

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -144,72 +164,97 @@ impl DBCacheWriteExecutor {
...
     /// and order of blockchain data.
     ///
-    /// Errors are raised if the incoming transaction's block does not logically follow
-    /// the sequence of the blockchain (e.g., if the block is too far ahead or does not
-    /// connect with the last persisted block).
-    async fn write(&mut self, new_db_tx: DBTransaction) -> Result<(), StorageError> {
+    /// Errors are sent to the error channel if the incoming transaction's block does not logically
+    /// follow the sequence of the blockchain (e.g., if the block is too far ahead or does not
+    /// connect with the last persisted block) or if we fail to send a transaction to the database.
+    async fn write(&mut self, mut new_db_tx: DBTransaction) {
```
kayibal: From the comment it is unclear whether an old tx or old block will cause an error (.. not logically follow).

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -221,47 +266,55 @@ impl DBCacheWriteExecutor {
...
-                            .tx
-                            .send(Err(e))
-                            .expect("send error back to db tx ok")
-                    } else {
-                        db_tx.tx.send(Ok(())).expect("response");
+                for op in db_txs.into_iter() {
+                    if !seen_operations.contains(&op) {
+                        // Only executes if it is not already in seen_operations
+                        match self
+                            .execute_write_op(op.clone(), conn)
```
kayibal: I checked this method and I think most of the cases in there just use a reference to op, maybe the interface should also accept a reference and then we can get rid of the clone here.

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -221,47 +266,55 @@ impl DBCacheWriteExecutor {
...
+                            .await
+                        {
+                            Ok(_) => {
+                                seen_operations.push(op.clone());
```
kayibal: That would also let us get rid of this second clone here.

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -320,11 +368,95 @@ pub struct CachedGateway {
...
+    }
+    pub async fn upsert_block(&self, new: &evm::Block) -> Result<(), StorageError> {
+        let (tx, rx) = oneshot::channel();
+        let db_tx = DBTransaction::new(*new, vec![WriteOp::UpsertBlock(*new)], tx);
+        self.tx
+            .send(DBCacheMessage::Write(db_tx))
+            .await
+            .expect("Send message to receiver ok");
+        rx.await
+            .expect("Receive confirmation ok")
```
kayibal: I don't think we should panic on errors, instead they should be returned so they can be handled by the extractor.

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -333,16 +465,23 @@ impl CachedGateway {
...
         //TODO: handle multiple extractors reverts
-        let (tx, rx) = oneshot::channel();
-        let _ = self
-            .tx
-            .send(DBCacheMessage::Flush(tx))
-            .await;
-        let _ = rx.await;
+        self.flush()
+            .await
+            .expect("Flush should succeed");
```
kayibal: Currently this won't work for multiple extractors running at the same time but I saw there is another PR available for this right?

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -411,86 +550,120 @@ mod test {
...
-        let os_rx_flush = send_flush_message(&tx).await;
-
         os_rx
             .await
             .expect("Response from channel ok")
-            .expect("DB transaction not sent");
+            .expect("Transaction cached");
 
+        // Send flush message
+        let os_rx_flush = send_flush_message(&tx).await;
```
kayibal: Why not use the CacheGateway struct here directly to send the messages?

tycho-indexer/src/storage/postgres/cache.rs
zizou0x: Yes there is a draft for it https://github.com/propeller-heads/tycho-indexer/pull/76

tycho-indexer/src/storage/postgres/cache.rs
zizou0x: I was expecting to use this logic in several places so I wrapped it in a function to avoid duplication but it seems in the end it's only used once

tycho-indexer/src/storage/postgres/cache.rs
zizou0x: I've added logic to ignore block upserts in this case

tycho-indexer/src/storage/postgres/cache.rs
zizou0x: Can we make this restricted only to this module even if it's used in `main.rs`? I tried and it's complaining, I've made it `pub(crate)`

tycho-indexer/src/storage/postgres/cache.rs
zizou0x: Above it says 
> If a transaction belongs to an older block, it is immediately applied to the database.

Here I removed the "logically follow" to be more precise

tycho-indexer/src/storage/postgres/cache.rs
zizou0x: Same I above, I couldn't make it fully private, only `pub(crate)`

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -376,85 +368,85 @@ impl CachedGateway {
...
         rx.await
-            .expect("Receive confirmation ok")
     }
 
     pub async fn upsert_tx(
         &self,
         block: &evm::Block,
         new: &evm::Transaction,
-    ) -> Result<(), StorageError> {
+    ) -> Result<Result<(), StorageError>, RecvError> {
```
kayibal: Small suggestion: make a new type for this called `CachedGatewayResult`?

kayibal: Alternatively we could unpack this and cast RecvError to `StorageError::WriteCacheGoneAway`

EDIT: Yup favouring this option the return type of this is currently not so nice tbh

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -144,72 +163,93 @@ impl DBCacheWriteExecutor {
...
+                                        // Ignore block upserts, we expect it to already be stored
+                                        continue;
+                                    } else {
+                                        match self.execute_write_op(&op, conn).await {
+                                            Err(StorageError::DuplicateEntry(entity, id)) => {
+                                                // As this db transaction is old. It can contain
+                                                // already stored txs, we log the duplicate entry
+                                                // error and continue
+                                                debug!(
+                                                    "Duplicate entry for {} with id {}",
```
kayibal: I would maybe add here that we are ignoring it, else it sounds kind of serious.

tycho-indexer/src/storage/postgres/cache.rs
zizou0x: That's a really cool idea, I also don't like this type

PR #72, opened 2023-12-18
Clean up protobuf messages

louise-poole:


proto/tycho/evm/v1/shared.proto
kayibal, line 1: Can we call it common instead of shared? Nitpick but if you don't mind^^

proto/tycho/evm/v1/shared.proto
```diff
@@ -0,0 +1,35 @@
...
+  repeated bytes tokens = 2;
+  repeated string contracts = 3;
+  repeated Attribute static_att = 4;
+}
+
+message BalanceChange {
+  bytes token = 1;
+  bytes balance = 2;
+  bytes component_id = 3;
+}
```
kayibal: It seems we are not using these anywhere? Could you please add them to the transaction changes please?

louise-poole: It's already used in the `TransactionBalanceChanges` message. I'll add it to the entity one 👍

proto/tycho/evm/v1/common.proto
```diff
@@ -0,0 +1,35 @@
...
+  string name = 1;
+  bytes value = 2;
+}
+
+message ProtocolComponent {
+  string id = 1;
+  repeated bytes tokens = 2;
+  repeated string contracts = 3;
+  repeated Attribute static_att = 4;
+}
```
kayibal: I think this will need a change type to handle deletes.

proto/tycho/evm/v1/common.proto
```diff
@@ -0,0 +1,35 @@
...
+  bytes hash = 1;
+  bytes from = 2;
+  bytes to = 3;
+  uint64 index = 4;
+}
+
+message Attribute {
+  string name = 1;
+  bytes value = 2;
+}
```
kayibal: This also will need a change type, no? Mostly to handle deletes...

proto/tycho/evm/v1/vm.proto
```diff
@@ -40,48 +28,14 @@ message ContractChange {
...
-  bytes token = 1;
-  bytes balance = 2;
-  bytes component_id = 3;
-}
-
-message TransactionChanges {
+message TransactionContractChanges {
   Transaction tx = 1;
   repeated ContractChange contract_changes = 2;
   repeated ProtocolComponent components = 3;
```
kayibal: To be consequent this should also be called component changes...

proto/tycho/evm/v1/common.proto
louise-poole: Do you have an example of when attributes are deleted instead of just updated? Or are we doing this to future proof it?

PR #73, opened 2023-12-18
CI: Protobuf liniting

SmartMonkey-git:


substreams/ethereum-ambient/proto/buf.yaml
louise-poole, line 1: What's this file here for? I don't think we define any protobuf messages in this `substreams` directory.

.github/workflows/ci.yaml
```diff
@@ -85,11 +85,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
-      - name: Install Buf
-        run: |
-          npm install @bufbuild/buf@1.28.1
+      - uses: bufbuild/buf-setup-action@v1.28.1
```
louise-poole, line 88: Nice! Do we not need to install it if we use it like this?

SmartMonkey-git: Install it on you mac book or in the CI? I'm not sure if I understand the question. :)
This actions installs it in the CI runner. How to install it on your mac is in the read me.

louise-poole: On the CI 😆 You've removed `npm install @bufbuild/buf@1.28.1` so I assume it's no longer necessary?

louise-poole: (I'm a noob with these CI configs so have lots for learn/understand with what you're changing 👀 )

SmartMonkey-git: Basically, `npm install @bufbuild/buf@1.28.1` has the same outcome as `- uses: bufbuild/buf-setup-action@v1.28.1`. The difference is that the former is a normal console command. You can determine that by the "run" keyword. The latter is a dedicated github-actions action. Actions are wraped code so that its easily usable in a workflow.
There is even an official page for those actions -> https://github.com/marketplace/actions/buf-setup

PR #74, opened 2023-12-19
ENG 2435: Implement StorableProtocolState for ProtocolState

louise-poole:
Implement the StorableProtocolState trait for ProtocolState. We do this for 2 reasons:
1) we can make the gateway generic over the StorableProtocolState trait (for now it is not as we think our current struct will cover all use cases, but if we need it one day then this is ready)
2) to add ease of converting between the db and tycho ProtocolState structs.

### DIFF
--- a/tycho-indexer/src/extractor/evm/mod.rs
+++ b/tycho-indexer/src/extractor/evm/mod.rs
@@ -1072,6 +1072,12 @@ pub mod fixtures {
                         },
                     ],
                     change: ChangeType::Creation.into(),
+                    protocol_type: Some(ProtocolType {
+                        name: "WeightedPool".to_string(),
+                        financial_type: 0,
+                        attribute_schema: vec![],
+                        implementation_type: 0,
+                    }),
                 }],
                 balance_changes: vec![BalanceChange {
                     token: hex::decode(
@@ -1179,6 +1185,12 @@ pub mod fixtures {
                             change: ChangeType::Creation.into(),
                         }],
                         change: ChangeType::Creation.into(),
+                        protocol_type: Some(ProtocolType {
+                            name: "WeightedPool".to_string(),
+                            financial_type: 0,
+                            attribute_schema: vec![],
+                            implementation_type: 0,
+                        }),
                     }],
                     balance_changes: vec![],
                 },
@@ -1248,6 +1260,12 @@ pub mod fixtures {
                 },
             ],
             change: ChangeType::Creation.into(),
+            protocol_type: Some(ProtocolType {
+                name: "WeightedPool".to_string(),
+                financial_type: 0,
+                attribute_schema: vec![],
+                implementation_type: 0,
+            }),
         }
     }
 }
--- a/proto/tycho/evm/v1/common.proto
+++ b/proto/tycho/evm/v1/common.proto
@@ -46,6 +46,26 @@ message Attribute {
   ChangeType change = 3;
 }
 
+enum FinancialType{
+  SWAP = 0;
+  LEND = 1;
+  LEVERAGE = 2;
+  PSM = 3;
+}
+
+
+enum ImplementationType {
+  VM = 0;
+  CUSTOM = 1;
+}
+
+message ProtocolType{
+  string name = 1;
+  FinancialType financial_type = 2;
+  repeated Attribute attribute_schema = 3;
+  ImplementationType implementation_type = 4;
+}
+
 // A struct describing a part of the protocol.
 message ProtocolComponent {
   // A unique identifier for the component within the protocol.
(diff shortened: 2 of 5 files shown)


tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -683,6 +683,15 @@ pub struct ProtocolState {
 // TODO: remove dead code check skip once extractor is implemented
 #[allow(dead_code)]
 impl ProtocolState {
+    pub fn new(component_id: String, attributes: HashMap<String, Bytes>, modify_tx: H256) -> Self {
+        Self {
+            component_id,
+            updated_attributes: attributes,
+            deleted_attributes: HashMap::new(),
```
dianacarvalho1: just because we don't know yet if we even need this or how it would look like right?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -683,6 +683,15 @@ pub struct ProtocolState {
...
+            modify_tx,
```
dianacarvalho1, line 691: what is the modify_tx?

dianacarvalho1: ah block ts, got it

tycho-indexer/src/storage/mod.rs
```diff
@@ -466,11 +466,13 @@ pub trait StorableToken<S, N, I>: Sized + Send + Sync + 'static {
...
-
-    fn to_storage(&self, contract_id: I) -> N;
+    fn from_storage(
+        val: S,
+        component_id: ContractId,
+        tx_hash: &TxHash,
+    ) -> Result<Self, StorageError>;
 
-    fn contract_id(&self) -> ContractId;
+    fn to_storage(&self, protocol_component_id: I, tx_id: I, block_ts: NaiveDateTime) -> N;
```
dianacarvalho1: I'm confused about these generics:
- In `from_storage` `component_id` is a `ContractId` but then in `to_storage` `protocol_component_id` is an `I`
- why is `tx_id` also an `I`?

tycho-indexer/src/extractor/evm/mod.rs
louise-poole: Yes. I have it because I think we'll potentially need it when handling reverts/deltas. But I think it'll need a proper think-through at that point and will likely change type etc by then.

tycho-indexer/src/storage/mod.rs
louise-poole: In `from_storage` I can probably just change `component_id` to `String`. I had it as `ContractId` initially as that's what all other `from_storage` implementations used to pass contract addresses etc. However, component_id isn't necessarily an address and I should change this.
In `to_storage` we use arguments of type `I` for all foreign keys expected in the db entry we are creating. Here we need the db id for both the linked protocol component and the linked tx.

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -675,28 +675,37 @@ pub struct ProtocolState {
     // the update protocol specific attributes, validated by the components schema
     pub updated_attributes: HashMap<String, Bytes>,
     // the deleted protocol specific attributes
-    pub deleted_attributes: HashMap<String, Bytes>,
+    pub deleted_attributes: HashSet<String>,
```
kayibal, line 678: This struct kind of mixes the message ProtocolStateChanges, with the actual ProtocolState as this is the current complete state of the protocol, the latter would be slightly different.

I am not sure but seems problematic to me going forward.

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -727,8 +736,8 @@ impl ProtocolState {
             )));
         }
         self.modify_tx = other.modify_tx;
-        for attr in other.deleted_attributes.keys() {
-            self.updated_attributes.remove(attr);
+        for attr in other.deleted_attributes.clone() {
+            self.updated_attributes.remove(&attr);
```
kayibal, line 740: Why do you clone and immediately take the reference again here?

louise-poole: Thanks! Will fix in the new PR 👍

tycho-indexer/src/extractor/evm/mod.rs
louise-poole: Okay, I see your point. We can definitely split this responsibility between 2 structs. Though I am unsure it will actually present an issue, it's probably best to be explicit.
Strictly speaking though, a ProtocolState is merely the ProtocolStateChanges from between the beginning of the contract's existence to now 👀 (hehe)
I will open a new PR to implement this 👍

PR #75, opened 2023-12-19
ENG 2436: Implement ProtocolGateway get_states

louise-poole:
Implement get_protocol_states (renamed from get_states to remove any ambiguity as the word 'state' is also used in the `StateGateway` to represent extractor states etc).

Note - I removed the db triggers to automatically update `valid_to` for protocol states. The approach we decided to use to manage states and their attributes in the db is to store a new state entity per attribute update. Therefore automatically updating `valid_to` in the way it was done no longer makes sense. Also, since it is planned to remove all of these triggers ([here](https://datarevenue.atlassian.net/browse/ENG-2544)), I went ahead and removed them for protocol states now already.

### DIFF
--- a/tycho-indexer/src/storage/postgres/protocol.rs
+++ b/tycho-indexer/src/storage/postgres/protocol.rs
@@ -56,9 +56,9 @@ where
                         let transaction = data.2;
                         let latest_tx = match latest_tx {
                             Some(latest)
-                                if latest.block_id < transaction.block_id
-                                    || (latest.block_id == transaction.block_id
-                                        && latest.index < transaction.index) =>
+                                if latest.block_id < transaction.block_id ||
+                                    (latest.block_id == transaction.block_id &&
+                                        latest.index < transaction.index) =>
                             {
                                 Some(transaction)
                             }


tycho-indexer/src/storage/postgres/mod.rs
```diff
@@ -733,4 +769,112 @@ pub mod db_fixtures {
...
+    ) -> i64 {
+        diesel::insert_into(schema::protocol_type::table)
+            .values((
+                schema::protocol_type::name.eq(name),
+                schema::protocol_type::financial_type.eq(finantial_type),
+                schema::protocol_type::implementation.eq(implementation_type),
+            ))
+            .returning(schema::protocol_type::id)
+            .get_result(conn)
+            .await
```
dianacarvalho1: didn't @tamaralipows and @SmartMonkey-git already did this?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -1,22 +1,83 @@
...
+                    .filter(schema::transaction::id.eq(state.modify_tx))
+                    .select(schema::transaction::hash)
+                    .first::<Bytes>(conn)
+                    .await
+                    .map_err(|err| {
+                        StorageError::NoRelatedEntity(
+                            "Transaction".to_owned(),
+                            "ProtocolState".to_owned(),
+                            state.id.to_string(),
+                        )
```
dianacarvalho1: huumm it isn't so nice that we need to query so many extra things outside of the gtw :/ 
why isn't this a gtw method actually?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -180,22 +291,178 @@ mod test {
...
+        let result = gateway
+            .get_protocol_states(&Chain::Ethereum, None, None, None, &mut conn)
+            .await
+            .unwrap();
+
+        assert_eq!(result, expected)
+    }
+
+    #[tokio::test]
+    async fn test_get_protocol_states_at() {
```
dianacarvalho1, line 423: we are missing testing with filter at system and ids no?

tycho-indexer/src/extractor/evm/storage.rs
```diff
@@ -408,16 +409,21 @@ pub mod pg {
...
-        ) -> orm::NewProtocolState {
-            orm::NewProtocolState {
-                protocol_component_id,
-                state: self.convert_attributes_to_json(),
-                modify_tx: tx_id,
-                tvl: None,
-                inertias: None,
-                valid_from: block_ts,
-                valid_to: None,
+        ) -> Vec<orm::NewProtocolState> {
```
tvinagre, line 417: Would be nice to add docs here, since we're altering the way things are saved in the db

tycho-indexer/src/extractor/evm/storage.rs
```diff
@@ -408,16 +409,21 @@ pub mod pg {
...
+            let mut protocol_states = Vec::new();
+            for (name, value) in &self.updated_attributes {
+                protocol_states.push(orm::NewProtocolState {
+                    protocol_component_id,
+                    attribute_name: Some(name.clone()),
+                    attribute_value: Some(value.clone()),
+                    modify_tx: tx_id,
+                    valid_from: block_ts,
+                    valid_to: None,
+                    tvl: None,
```
tvinagre: Do `tvl` and `inertias` need to be repeated for each protocol_state? I assume they'd be one per-protocol, not per-state

tycho-indexer/src/storage/postgres/protocol.rs
tvinagre: bc we only save the ids on the database, but I agree you could do this by JOINS in the single query?

tycho-indexer/src/storage/postgres/mod.rs
louise-poole: They added the gateway methods that do this - here I added a test helper function to `db_fixtures` that allows me to add entities directly to my test db. If you look at the full `db_fixtures` you'll see there's a function like this for almost all the tables in the db. This allows us more control over our test set up.

tycho-indexer/src/storage/postgres/protocol.rs
louise-poole: @dianacarvalho1 I wanted to make this a gw function, but encountered a lot of rust red tape with doing it like that. I also tried to make it a nested function but cannot use generics when declaring the attribute types in a nested function... so that sucked. I'll have another go at moving it onto the gateway though now that I'm older and wiser than when I first attempted (hehe)

tycho-indexer/src/extractor/evm/storage.rs
louise-poole: It wasn't in the scope of my task to handle tvl and inertias so I left them exactly as they were given in the plans/sketch. If I'm not mistaken, adding tvl etc is part of the Purpeller's current task? @dianacarvalho1

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -67,30 +128,75 @@ where
...
+                decode_protocol_states(
+                    orm::ProtocolState::by_protocol_system(system, chain_db_id, version_ts, conn)
+                        .await,
+                    system.to_string().as_str(),
+                    conn,
+                )
+                .await
+            }
+            _ => {
+                dbg!(chain_db_id);
```
SmartMonkey-git: Did you leave this here intentionally?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -67,30 +128,75 @@ where
...
+        conn: &mut Self::DB,
+    ) -> Result<Vec<Self::ProtocolState>, StorageError> {
+        let chain_db_id = self.get_chain_id(chain);
+        let version_ts = match &at {
+            Some(version) => Some(version.to_ts(conn).await?),
+            None => None,
+        };
+
+        match (ids, system) {
+            (Some(ids), _) => {
```
SmartMonkey-git: Isn't there a 4th case, where Id's and systems are set?

tycho-indexer/src/storage/postgres/protocol.rs
louise-poole: Nope 👀  Well spotted!

tycho-indexer/src/storage/postgres/protocol.rs
louise-poole: To me that case doesn't make sense - if you are providing something as specific as ids, then you don't need a blanket filter like system as well. I should make this clearer in the docs here though - if ids are provided, system is ignored. Unless others disagree and think this is a scenario I should cater for?

tycho-indexer/src/storage/postgres/mod.rs
dianacarvalho1: aaaaahhh cooool! sorry, I didn't know about this and I'm still getting my bearing in all of these files 👀

tycho-indexer/src/extractor/evm/storage.rs
dianacarvalho1: yes, handling tvls is in our task but we are not planning on handling the protocol states 😬

dianacarvalho1: I'm adding an extra task at end of us to double check this

tycho-indexer/src/extractor/evm/storage.rs
```diff
@@ -382,42 +382,51 @@ pub mod pg {
             })
         }
     }
+
     impl StorableProtocolState<orm::ProtocolState, orm::NewProtocolState, i64> for evm::ProtocolState {
         fn from_storage(
             val: orm::ProtocolState,
```
kayibal: I think we'll need to change this interface, it should probably take a list of `orm::ProtocolState`, now that each row there will only contain a single attribute and value, else you'll need to do a lot of unnecessary allocations.

tycho-indexer/src/extractor/evm/storage.rs
```diff
@@ -382,42 +382,51 @@ pub mod pg {
...
-        ) -> orm::NewProtocolState {
-            orm::NewProtocolState {
-                protocol_component_id,
-                state: self.convert_attributes_to_json(),
-                modify_tx: tx_id,
-                tvl: None,
-                inertias: None,
-                valid_from: block_ts,
-                valid_to: None,
+        ) -> Vec<orm::NewProtocolState> {
```
kayibal, line 417: You altered the interface here correctly already, I think the same should go for the `fn from_storage` method.

tycho-indexer/src/extractor/evm/storage.rs
kayibal: Hmmm tvl can change with each transaction but not with each attribute - indeed we'll probably need a separate table for them. Inertias will be dropped completely.

tycho-indexer/src/storage/postgres/mod.rs
```diff
@@ -274,6 +275,38 @@ impl StorageError {
...
+    pub async fn to_ts(&self, conn: &mut AsyncPgConnection) -> Result<NaiveDateTime, StorageError> {
+        if !matches!(self.1, VersionKind::Last) {
+            return Err(StorageError::Unsupported(format!(
+                "Unsupported version kind: {:?}",
+                self.1
+            )));
+        }
+        self.0.to_ts(conn).await
+    }
+}
```
kayibal, line 308: This is not an orm object but acore entity right here? 

Will rust make this method available even if I don't have the postgres DB module imported? If it does (which I'd expect) then I think this is not so nice because we adding DB specific behaviour to a core model.

Same for BlockOrTimestamp above.

EDIT: I researched this further and I am against doing this. The reason is that rust will simply merge this with the main implementation block. So apart from making it harder to see a structs behaviour in a single file this adds a database dependency to our core models which I think is a bad idea - even though this looks smarter and more elegant sometimes the dumb way (simple function) is the way to go :]

tycho-indexer/src/storage/postgres/protocol.rs
kayibal: Alternative would be to throw an error if both are set. It comes down to taste here... I think I agree with you that if ids are provided we can skip other filters.

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -67,30 +130,77 @@ where
...
         chain: &Chain,
         at: Option<Version>,
         system: Option<ProtocolSystem>,
-        id: Option<&[&str]>,
-    ) -> Result<Vec<ProtocolState>, StorageError> {
-        let block_chain_id = self.get_chain_id(chain);
-        todo!()
+        ids: Option<&[&str]>,
+        conn: &mut Self::DB,
+    ) -> Result<Vec<Self::ProtocolState>, StorageError> {
```
kayibal, line 149: The return type will not be a protocol state, but all changes to protocol attributes that happened at that version, right?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -1,22 +1,85 @@
...
 
+impl<B, TX, A, D, T> PostgresGateway<B, TX, A, D, T>
+where
+    B: StorableBlock<orm::Block, orm::NewBlock, i64>,
+    TX: StorableTransaction<orm::Transaction, orm::NewTransaction, i64>,
+    D: ContractDelta + From<A>,
+    A: StorableContract<orm::Contract, orm::NewContract, i64>,
+    T: StorableToken<orm::Token, orm::NewToken, i64>,
+{
+    // decode Protocol State query results
```
kayibal: Would appreciate more details here.

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -1,22 +1,85 @@
...
+    async fn _decode_protocol_states(
+        &self,
+        result: Result<Vec<orm::ProtocolState>, diesel::result::Error>,
+        context: &str,
+        conn: &mut <Self as ProtocolGateway>::DB,
```
kayibal: We need to do queries to decode???

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -1,22 +1,85 @@
...
+
+                    let protocol_state =
+                        ProtocolState::from_storage(state, component_id.clone(), &tx_hash)?;
+
+                    if let Some(existing_state) = protocol_states.get_mut(&component_id) {
+                        // found a protocol state with a matching component id - merge states
+                        dbg!(&existing_state);
+                        dbg!(&protocol_state);
+                        existing_state
+                            .merge(protocol_state)
```
kayibal: Seems like misuse of merge to me...

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -1,22 +1,85 @@
...
+                            schema::transaction::table
+                                .on(schema::transaction::id.eq(schema::protocol_state::modify_tx)),
+                        )
+                        .filter(schema::protocol_state::id.eq(state.id))
+                        .select((
+                            schema::protocol_component::external_id,
+                            schema::transaction::hash,
+                        ))
+                        .first::<(String, Bytes)>(conn)
+                        .await?;
```
kayibal: Doing a (very expensice) query for each individual attribute sounds like it would be extremely slow once we run this under actual load? I think we can't proceed before we fix this.

This is a result of splitting the responsibilities too much, I think you will need to restructure things quite a lot.

tycho-indexer/src/storage/postgres/mod.rs
louise-poole: I did some trial and error and you cannot access these implemented functions outside of the postgres module or file that imports from the postgres module. While you are right, I am technically extending the core implementation of `Version` here, these added functions are not actually visible or accessible outside of their implementation scope. They exist, but are hidden. To be able to access the new functions you need to first bring their implementation into scope.
ChatGPT helped me with this explanation:

>In Rust, traits and their implementations (impls) are global. If you add an implementation for a type anywhere in your code, that implementation is accessible throughout the entire project.
However, to use these extended implementations in other modules, the trait itself must be in scope. The visibility of a trait's methods does not follow the visibility of the trait object, but rather the visibility of the trait itself.
Here is an example to illustrate this:
```
// In core/models.rs
pub struct Version {
  // fields here
}

// In specific/module.rs
use crate::core::models::Version;

trait SpecificTrait {
    fn specific_fn(&self);
}

impl SpecificTrait for Version {
    fn specific_fn(&self) {
        // Implementation here
    }
}

// This method is now available for all `Version` objects,
// but only in modules where `SpecificTrait` is brought into scope.
```

tycho-indexer/src/storage/postgres/protocol.rs
louise-poole: No, this is returning the full valid protocol state at that version. I was under the impression one would use `get_protocol_state_delta` to get changes at a given version.

tycho-indexer/src/storage/postgres/protocol.rs
louise-poole: Nope, I will optimise this 👍

tycho-indexer/src/extractor/evm/storage.rs
louise-poole: I take this to mean I can remove both `tvl` and `inertias` from the protocol state table and leave it as that for now 👍

tycho-indexer/src/storage/postgres/mod.rs
kayibal: Yes if this was done with a trait I think it would be alright because you have to import a trait to use it. So in that case it would indeed as you said an extension.

Afaik you can't import an impl block... but then having a `ToTs` trait is more complex than just calling an internal util function.

louise-poole: I tried it out and can access the core functions for `Version` outside of this mod, but I cannot access `to_ts` - the exact behaviour we want. This functionality is used by 2 gateways: I originally did this implementation in the file of one of the gateways and could not access it from the other. The only way to get this to work was to put it in a mod file that both had access to/imported from. So in practice what we want to happen is happening. It comes down to if we're okay with having an implementation hidden/separated from the actual struct definition and core implementation. I can change it to be a gateway util function if we strongly prefer? I just really liked how much easier and cleaner this was to use.

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -369,18 +376,19 @@ impl ProtocolState {
         }
 
         query
-            .select(Self::as_select())
-            .get_results::<Self>(conn)
+            .select((Self::as_select(), protocol_component::external_id, Transaction::as_select()))
+            .get_results::<(Self, String, Transaction)>(conn)
```
kayibal, line 380: Imagine we have to retrieve all usv3 pools. Each pool will have an row in the table for each tick some pools have 100s of ticks and we may have 1000s of pools, so you can expect about 100k - 1M rows retrieved from that. Now apart from having to join such a huge table, you'll also have to transfer all the data for the transactions and the component ids, these will be duplicated for each attribute (row) of the same pool. So you are incurring a huge unnecessary data transfer with this strategy.

That is why I usually prefer to issue separate queries for each of these and then piece them together using hash maps application side.

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -28,52 +33,51 @@ where
...
+                                        latest.index < transaction.index) =>
+                            {
+                                Some(transaction)
+                            }
+                            None => Some(transaction),
+                            _ => latest_tx,
+                        };
+                        (states, latest_tx)
+                    },
+                );
```
kayibal, line 70: Sorry not a big a fan of this block^^. It would probably would be more performant and easier to read without the hash map. 

You are basically doing a groupby, but you don't need the complete result of it (as in you don't need the hashmap as whole that you are producing here). 

I'd probably sort the data by component, block and transaction index, then iterate over it creating slices when the id changes (you have to look ahead one element to detect that). Then use a slice into the vector you are iterating over to call `from_storage` (thus avoiding additional allocations), the latest row in your group should also contain the latest modifying transaction for this state.

tycho-indexer/src/storage/postgres/orm.rs
kayibal: Just want to raise awareness about that, for now probably ok and we have to monitor closely and potentially refactor later on.

kayibal: EDIT just chatgpt'ed this and seems like join is alright here :)

kayibal: It seems this blogpost also confirms that:
https://medium.com/@clementgrimault/optimize-the-way-you-fetch-relationships-with-postgresql-7711fe6457d2

PR #76, opened 2023-12-19
ENG-2419: Handle reorgs with multiple extractors

zizou0x:


tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -411,35 +430,49 @@ impl CachedGateway {
...
+            self.last_revert_request = Some((
+                RevertParameters {
+                    start_version: start_version.cloned(),
+                    end_version: end_version.clone(),
+                },
+                delta.clone(),
+            ));
+        }
+
+        Ok(delta)
```
zizou0x: Here is how we currently handle multiple extractors revert: we get the account delta for a specific revert and cache it. Then the db will be reverted and if another extractor ask for the same account delta (so same request parameters) we return the one that was cached.
At the moment this is called only during reorgs so it's fine. But what is we want to use this function somewhere else? It could erase the `last_revert_request` in the middle of a revert?

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -731,6 +874,66 @@ mod test {
...
+            .get()
+            .await
+            .expect("Failed to get a connection from the pool");
+        connection
+            .begin_test_transaction()
+            .await
+            .expect("Failed to start test transaction");
+
+
+        setup_data(&mut connection).await;
```
zizou0x: This is annoying because in our code we use a pool of db connections. But `begin_test_transaction()` can be used only on a single connection. That means, we're stuck with a pool with a single connection. But in some cases (like here) my test require a pool of two connections (because I use 2 at the same time at some point in my logic). Here I managed to make it work because one of the two connections is just created but not used at all. But it's confusing for readers no?

kayibal: Would be good to find a solution for this... but it would probably mean mocking the pool. Note that you do not want to mock the AsyncPgConnection type (it is close to impossible to get the traits right for this...)

tycho-indexer/src/storage/postgres/cache.rs
kayibal: I have been thinking about this and I think for now the caching will work for us. But it is important to note that we are assuming that if reverts happen, they happen over the same "span" of blocks (please document this). I think that is safe as long as all your substreams runs on the same geographical firehose network. So hopefully we are good for now - (should probably double check this in discord though). 

Another stability issue we may get with the application if we crashes during a revert (somewhat likely especially on new releases). Once we restart, some extractors might still have to process the revert message but the db was already reverted. I do think though that this is a special case that we can probably handle on the extractor level - in this case it is probably save to ignore the revert message from substreams.

A more elegant way to handle this would be to allow for forks to exist in the DB. So you can always request state diffs of reverts, without requiring any in memory caches. Probably worth exploring a bit in that direction as that would have big positive benefits on the entire applications flexibility and stability.

Last but not least about this implementation: I think you should probably keep multiple deltas cached. Maybe use something like LRU for this?

tycho-indexer/src/storage/postgres/cache.rs
zizou0x: I agree, made a cooldown week task for this https://datarevenue.atlassian.net/browse/ENG-2635

tycho-indexer/src/storage/postgres/cache.rs
zizou0x: I've spent a small amount of my time exploring db forking but I couldn't find an obvious solution for this. It might be worth to come back to this later, but in the meantime I've implemented the LRU solution you suggested.

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -467,16 +484,39 @@ impl CachedGateway {
...
-            .expect("Received signal ok")
-            .expect("Flush should succeed");
-        self.state_gateway
-            .get_accounts_delta(chain, start_version, end_version, db)
-            .await
+        let mut lru_cache = self.lru_cache.lock().await;
+
+        // Construct a key for the LRU cache
+        let key = RevertParameters {
+            start_version: start_version.cloned(),
```
kayibal, line 499: Please think about it: Is it safe for this to be None, basically 100% of the time? Or maybe we should resolve to the latest version here, so we get a absolute value?

zizou0x: I've already put some thought while implementing. It's not 100% safe but really close as it's very unlikely to have two forks at the same block (that would mean three different state on the network at the same time).

To make this fully reliable we would need to use block hashes, but I don't know from where I could get them, I think from Substream is the only option here, right?

kayibal: I think you could query the DB for a block header or return block headers from the revert method?

zizou0x: I'm not sure we can rely on the database in this logic, the first revert would be able to get the last block header, but the next one couldn't (because it would have already been reverted). I think the only thing we can rely on is Substream returning two block header for reverts (from and to)

tvinagre: > very unlikely to have two forks at the same block

This is not a safe assumption. It has happened and the likelihood is increased on Blockchains with slower block times (like Polygon)

zizou0x: Link to the rest of the discussion: https://propellerswap.slack.com/archives/C05MPCJUQPM/p1705679679491189

tycho-indexer/src/extractor/evm/ambient.rs
```diff
@@ -317,9 +342,19 @@ where
                 block_ref.id, err
             ))
         })?;
+
+        let current = self
+            .get_last_processed_block()
+            .await
+            .map(|block| BlockIdentifier::Hash(block.hash.into()));
+
```
kayibal, line 350: What should we do if current is None here? I feel here we might potentially request an unnecessary revert...

I am ok to simply skip the revert if it is at the start of an extractors lifetime? Not sure if save though... else we can panic if this happens...

Another option would be to initialise this block also from the DB same as we do for cursor.

zizou0x: The best way would be to have the last processed block in the ExtractionState, that way we could retrieve it when we start a new extractor and it fixes this issue. 

But I prefer to do this in another PR as it seems out of scope for this one.

tycho-indexer/src/storage/postgres/chain.rs
```diff
@@ -60,6 +60,7 @@ where
             }
 
             BlockIdentifier::Hash(block_hash) => orm::Block::by_hash(block_hash, conn).await,
+            BlockIdentifier::Latest => orm::Block::most_recent(conn).await,
```
kayibal: Don't you need the chain id here?

zizou0x: Absolutely, we need it

PR #80, opened 2023-12-21
Adjust protocol component types

SmartMonkey-git:


tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -502,7 +502,7 @@ pub struct ProtocolComponent {
     // ids of the tokens tradable
     tokens: Vec<String>,
```
pistomat: tokens is also `tokens: ::prost::alloc::vec::Vec<::prost::alloc::vec::Vec<u8>>,` in the protobuf message, should it be here also Vec<H160> as it represents the addresses?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -502,7 +502,7 @@ pub struct ProtocolComponent {
...
     // ids of the related contracts
```
pistomat:
```suggestion
    // addresses of the related contracts
```

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -1357,8 +1378,8 @@ mod test {
             chain: Chain::Ethereum,
             tokens: vec!["token1".to_string(), "token2".to_string()],
```
pistomat: This should also be H160, no?

PR #84, opened 2024-01-08
ENG-2028: Save ProtocolSystem to DB

SmartMonkey-git:
[Task](https://datarevenue.atlassian.net/browse/ENG-2028)

### DIFF
--- a/tycho-indexer/src/extractor/evm/mod.rs
+++ b/tycho-indexer/src/extractor/evm/mod.rs
@@ -1,7 +1,7 @@
 #![allow(dead_code)]
 
 use std::{
-    collections::{hash_map::Entry, HashMap},
+    collections::{hash_map::Entry, HashMap, HashSet},
     ops::Deref,
 };
 
@@ -675,7 +675,7 @@ pub struct ProtocolState {
     // the update protocol specific attributes, validated by the components schema
     pub updated_attributes: HashMap<String, Bytes>,
     // the deleted protocol specific attributes
-    pub deleted_attributes: HashMap<String, Bytes>,
+    pub deleted_attributes: HashSet<String>,
     // via transaction, we can trace back when this state became valid
     pub modify_tx: H256,
 }
@@ -683,12 +683,21 @@ pub struct ProtocolState {
 // TODO: remove dead code check skip once extractor is implemented
 #[allow(dead_code)]
 impl ProtocolState {
+    pub fn new(component_id: String, attributes: HashMap<String, Bytes>, modify_tx: H256) -> Self {
+        Self {
+            component_id,
+            updated_attributes: attributes,
+            deleted_attributes: HashSet::new(),
+            modify_tx,
+        }
+    }
+
     /// Parses protocol state from tychos protobuf EntityChanges message
     pub fn try_from_message(
         msg: substreams::EntityChanges,
         tx: &Transaction,
     ) -> Result<Self, ExtractionError> {
-        let (mut updates, mut deletions) = (HashMap::new(), HashMap::new());
+        let (mut updates, mut deletions) = (HashMap::new(), HashSet::new());
 
         for attribute in msg.attributes.into_iter() {
             match attribute.change().into() {
@@ -696,7 +705,7 @@ impl ProtocolState {
                     updates.insert(attribute.name, Bytes::from(attribute.value));
                 }
                 ChangeType::Deletion => {
-                    deletions.insert(attribute.name, Bytes::from(attribute.value));
+                    deletions.insert(attribute.name);
                 }
             }
         }
@@ -727,8 +736,8 @@ impl ProtocolState {
             )));
         }
         self.modify_tx = other.modify_tx;
-        for attr in other.deleted_attributes.keys() {
-            self.updated_attributes.remove(attr);
+        for attr in other.deleted_attributes.clone() {
+            self.updated_attributes.remove(&attr);
         }
         for attr in other.updated_attributes.keys() {
             self.deleted_attributes.remove(attr);
@@ -1578,10 +1587,9 @@ mod test {
         ]
         .into_iter()
         .collect();
-        let del_attributes1: HashMap<String, Bytes> =
-            vec![("to_add_back".to_owned(), Bytes::from(U256::from(0)))]
-                .into_iter()
-                .collect();
+        let del_attributes1: HashSet<String> = vec!["to_add_back".to_owned()]
+            .into_iter()
+            .collect();
         let mut state1 = ProtocolState {
             component_id: "State1".to_owned(),
             updated_attributes: up_attributes1,
@@ -1597,10 +1605,9 @@ mod test {
         ]
         .into_iter()
         .collect();
-        let del_attributes2: HashMap<String, Bytes> =
-            vec![("to_be_removed".to_owned(), Bytes::from(U256::from(0)))]
-                .into_iter()
-                .collect();
+        let del_attributes2: HashSet<String> = vec!["to_be_removed".to_owned()]
+            .into_iter()
+            .collect();
         let state2 = ProtocolState {
             component_id: "State1".to_owned(),
             updated_attributes: up_attributes2.clone(),
@@ -1621,10 +1628,9 @@ mod test {
         .into_iter()
         .collect();
         assert_eq!(state1.updated_attributes, expected_up_attributes);
-        let expected_del_attributes: HashMap<String, Bytes> =
-            vec![("to_be_removed".to_owned(), Bytes::from(U256::from(0)))]
-                .into_iter()
-                .collect();
+        let expected_del_attributes: HashSet<String> = vec!["to_be_removed".to_owned()]
+            .into_iter()
+            .collect();
         assert_eq!(state1.deleted_attributes, expected_del_attributes);
     }
 
@@ -1641,7 +1647,7 @@ mod test {
                 ProtocolState {
                     component_id: "State1".to_owned(),
                     updated_attributes: attributes.clone(),
-                    deleted_attributes: HashMap::new(),
+                    deleted_attributes: HashSet::new(),
                     modify_tx: H256::zero(),
                 },
             ),
@@ -1650,7 +1656,7 @@ mod test {
                 ProtocolState {
                     component_id: "State2".to_owned(),
                     updated_attributes: attributes,
-                    deleted_attributes: HashMap::new(),
+                    deleted_attributes: HashSet::new(),
                     modify_tx: H256::zero(),
                 },
             ),
@@ -1676,7 +1682,7 @@ mod test {
             ProtocolState {
                 component_id: "State1".to_owned(),
                 updated_attributes: new_attributes,
-                deleted_attributes: HashMap::new(),
+                deleted_attributes: HashSet::new(),
                 modify_tx: new_tx.hash,
             },
         )]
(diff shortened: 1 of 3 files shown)


tycho-indexer/src/storage/postgres/schema.rs
```diff
@@ -235,6 +238,24 @@ diesel::table! {
...
+diesel::table! {
+    use diesel::sql_types::Int8;
+    use super::sql_types::ProtocolSystemType;
+    protocol_system_type(id) {
+        id -> Int8,
+        protocol_enum -> ProtocolSystemType,
+    }
+}
+
+//diesel::table! {
```
SmartMonkey-git: Keeping this here to test something. Will remove it before merging.

tycho-indexer/migrations/2024-01-10_tycho_patch/down.sql
```diff
@@ -0,0 +1,7 @@
+-- dropping tables will drop any triggers or indices along with it
```
dianacarvalho1, line 1: we need to be careful when merging these two migrations no @tamaralipows @SmartMonkey-git ? Maybe we could actually merge everything into one?

tycho-indexer/migrations/2024-01-10_tycho_patch/up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TYPE protocol_system_type_enum AS ENUM(
+    'ambient'
+);
+
+CREATE TABLE protocol_system_type (
+    "id" BIGSERIAL PRIMARY KEY,
+    "protocol_enum" protocol_system_type_enum UNIQUE NOT NULL
```
dianacarvalho1: why "enum"?

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -222,6 +222,35 @@ pub struct ProtocolSystem {
     pub modified_ts: NaiveDateTime,
 }
 
+#[derive(Debug, DbEnum, Clone)]
+#[ExistingTypePath = "crate::storage::postgres::schema::sql_types::ProtocolSystemTypeEnum"]
+pub enum ProtocolSystemTypeEnum {
```
dianacarvalho1: do we need Enum in the name?

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -222,6 +222,35 @@ pub struct ProtocolSystem {
...
+#[diesel(table_name=protocol_system_type)]
+#[diesel(check_for_backend(diesel::pg::Pg))]
+pub struct NewProtocolSystemType {
+    pub protocol_enum: ProtocolSystemTypeEnum,
+}
+#[derive(Queryable, Selectable, Debug)]
+#[diesel(table_name=protocol_system_type)]
+#[diesel(primary_key(id))]
+#[diesel(check_for_backend(diesel::pg::Pg))]
+pub struct ProtocolSystemType {
```
dianacarvalho1: why not just ProtocolSystem?

tycho-indexer/src/storage/postgres/orm.rs
dianacarvalho1: I think this could be just `ProtocolSystemType`

tycho-indexer/migrations/2024-01-10_tycho_patch/down.sql
```diff
@@ -0,0 +1,7 @@
...
+DROP TABLE protocol_system_type;
+
+-- custom types
+DROP TYPE protocol_system_type;
```
dianacarvalho1, line 6: why did you decide to make a new table?? we already have one `protocol_system` table in the initial migration 🤔

tycho-indexer/migrations/2024-01-10_tycho_patch/up.sql
```diff
@@ -1,8 +1,6 @@
...
     'ambient'
 );
 
-CREATE TABLE protocol_system_type (
-    "id" BIGSERIAL PRIMARY KEY,
-    "protocol_enum" protocol_system_type_enum UNIQUE NOT NULL
-);
\ No newline at end of file
+ALTER TABLE protocol_system
+ALTER COLUMN "name" TYPE protocol_system_type USING "name"::protocol_system_type;
```
dianacarvalho1: so every time we add a new protocol_system_type we need to add a new migration like this right?

tycho-indexer/migrations/2024-01-10_tycho_patch/down.sql
tamaralipows: It's also an important question... how are these migrations applied in dev/prod/wherever this goes?

tycho-indexer/migrations/2024-01-10_tycho_patch/up.sql
SmartMonkey-git: I would like to say no, but the answer is yes. :(

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -217,11 +217,32 @@ pub struct NewTransaction {
...
+}
+
+impl From<models::ProtocolSystem> for ProtocolSystemType {
+    fn from(value: models::ProtocolSystem) -> Self {
+        match value {
+            models::ProtocolSystem::Ambient => ProtocolSystemType::Ambient,
+        }
+    }
+}
+
```
tamaralipows, line 245: There is one reason this concerns me, which is that you are now transforming this here, instead of using the `Storable` type that all of the other db elements are using. Is there a reason why? I've also been looking at this PR for reference
https://github.com/propeller-heads/tycho-indexer/pull/74/files

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -87,4 +89,78 @@ where
...
+    ) -> Result<i64, StorageError> {
+        use super::schema::protocol_system::dsl::*;
+        let new_system = orm::ProtocolSystemType::from(new);
+
+        let existing_entry = protocol_system
+            .filter(name.eq(new_system.clone()))
+            .first::<orm::ProtocolSystem>(conn)
+            .await;
+
+        if let Ok(entry) = existing_entry {
```
tamaralipows, line 106: Don't think I understand this line - what is `entry` here?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -87,4 +89,78 @@ where
...
+            ._get_or_create_protocol_system_id(ProtocolSystem::Ambient, &mut conn)
+            .await
+            .unwrap();
+        assert_eq!(protocol_system_id, 1);
+
+        let protocol_system_id = gw
+            ._get_or_create_protocol_system_id(ProtocolSystem::Ambient, &mut conn)
+            .await
+            .unwrap();
+        assert_eq!(protocol_system_id, 1);
```
tamaralipows, line 164: Can we also test here that we didn't make a new db entry?

SmartMonkey-git: We test that, dont we?
Given that the ID is unique and it still 1, after a second insert attempt. Then there should not be a new entry. 
Or are you suggestion something else?

tycho-indexer/src/storage/postgres/orm.rs
tamaralipows: @louise-poole I understand it may be overkill here, but do we have a clear guideline about when to implement this `Storable` trait and when not to?

tycho-indexer/src/storage/postgres/protocol.rs
SmartMonkey-git: `existing_entry` is a `Result<orm::ProtocolSystem, Error>`.
This syntax checks if the result is ok, unpacks it and makes it available in the scope of the if under the name `entry`. The else branch is taken, if the result is not ok.
It is kind of like the `match` operator, but with if and else.

You can think of it like the walrus operator `:=` in python. If you need an example, there is one in our solver API.

tamaralipows: ahhhhh thank you thank you

tycho-indexer/src/storage/postgres/protocol.rs
tamaralipows: My bad, misunderstood!

tycho-indexer/migrations/2024-01-10_tycho_patch/down.sql
```diff
@@ -0,0 +1,8 @@
+-- dropping tables will drop any triggers or indices along with it
+ALTER TABLE protocol_system
+ALTER COLUMN "name" TYPE varchar(255);
+
+-- custom types
+DROP TYPE protocol_system_type;
```
louise-poole, line 6: For my understanding (I have no db management experience in industry so migrations are a new concept to me): we need to add a `DROP` command in `down.sql` for any new things we create in `up.sql`?

tycho-indexer/migrations/2024-01-10_tycho_patch/down.sql
```diff
@@ -0,0 +1,8 @@
+-- dropping tables will drop any triggers or indices along with it
+ALTER TABLE protocol_system
+ALTER COLUMN "name" TYPE varchar(255);
```
louise-poole, line 3: Altering tables should be in `up.sql` no?

louise-poole: Ah, I see you have a conflicting alteration in `up.sql`. I assume that's the correct one and this should be removed?

tycho-indexer/src/storage/postgres/orm.rs
louise-poole: From my understanding - the Storable trait is so that we can use generics in our gateway definitions (on implementation of a gateway we can use any token type as long as it satisfies the trait `StorableToken` etc). Knowing this, I would only implement `Storable` traits for types we might want to make the gateways generic over or types that are complicated to convert to db format and back.
This is a very simple conversion and isn't needed for a gateway generic so I don't think it's necessary to implement the `Storable` trait here.

tycho-indexer/migrations/2024-01-10_tycho_patch/down.sql
SmartMonkey-git: Yes, that statement sounds correct to me.
If you add a table in up.sql, you need to drop it in down.sql.

tycho-indexer/migrations/2024-01-10_tycho_patch/down.sql
SmartMonkey-git: No, I'm altering the columns name back to what is was before the alteration in the up.sql.

tycho-indexer/migrations/2024-01-10_tycho_patch/down.sql
SmartMonkey-git: More generally, everything you do in up.sql, you need to revert in the down.sql

tycho-indexer/migrations/2024-01-10_tycho_patch/down.sql
louise-poole: Oooooooh! So `down.sql` is if you want to reverse the migration and `up.sql` is if you want to apply the migration?Wow, this makes a lot more sense! 😆

tycho-indexer/migrations/2024-01-10_tycho_patch/down.sql
dianacarvalho1: 🤷🏼‍♀️

PR #85, opened 2024-01-09
ENG-2029: ProtocolGateway: add method to upsert ProtocolTypes.

tamaralipows:
[Issue](https://datarevenue.atlassian.net/browse/ENG-2029) (No additional info.)

### DIFF
--- a/tycho-indexer/src/storage/postgres/orm.rs
+++ b/tycho-indexer/src/storage/postgres/orm.rs
@@ -244,7 +244,7 @@ impl From<models::ProtocolSystem> for ProtocolSystemType {
 }
 
 #[derive(Debug, DbEnum, Clone, PartialEq)]
-#[ExistingTypePath = "crate::storage::postgres::schema::sql_types::FinancialProtocolType"]
+#[ExistingTypePath = "crate::storage::postgres::schema::sql_types::FinancialType"]
 pub enum FinancialType {
     Swap,
     Psm,
@@ -253,7 +253,7 @@ pub enum FinancialType {
 }
 
 #[derive(Debug, DbEnum, Clone, PartialEq)]
-#[ExistingTypePath = "crate::storage::postgres::schema::sql_types::ProtocolImplementationType"]
+#[ExistingTypePath = "crate::storage::postgres::schema::sql_types::ImplementationType"]
 pub enum ImplementationType {
     Custom,
     Vm,
--- a/tycho-indexer/src/storage/postgres/schema.rs
+++ b/tycho-indexer/src/storage/postgres/schema.rs
@@ -1,16 +1,16 @@
 // @generated automatically by Diesel CLI.
 pub mod sql_types {
     #[derive(diesel::query_builder::QueryId, diesel::sql_types::SqlType)]
-    #[diesel(postgres_type(name = "financial_protocol_type"))]
-    pub struct FinancialProtocolType;
+    #[diesel(postgres_type(name = "financial_type"))]
+    pub struct FinancialType;
 
     #[derive(diesel::query_builder::QueryId, diesel::sql_types::SqlType)]
     #[diesel(postgres_type(name = "hstore"))]
     pub struct Hstore;
 
     #[derive(diesel::query_builder::QueryId, diesel::sql_types::SqlType)]
-    #[diesel(postgres_type(name = "protocol_implementation_type"))]
-    pub struct ProtocolImplementationType;
+    #[diesel(postgres_type(name = "implementation_type"))]
+    pub struct ImplementationType;
 
     #[derive(diesel::query_builder::QueryId, diesel::sql_types::SqlType)]
     #[diesel(postgres_type(name = "protocol_system_type"))]
@@ -199,16 +199,16 @@ diesel::table! {
 
 diesel::table! {
     use diesel::sql_types::*;
-    use super::sql_types::FinancialProtocolType;
-    use super::sql_types::ProtocolImplementationType;
+    use super::sql_types::FinancialType;
+    use super::sql_types::ImplementationType;
 
     protocol_type (id) {
         id -> Int8,
         #[max_length = 255]
         name -> Varchar,
-        financial_type -> FinancialProtocolType,
+        financial_type -> FinancialType,
         attribute_schema -> Nullable<Jsonb>,
-        implementation -> ProtocolImplementationType,
+        implementation -> ImplementationType,
         inserted_ts -> Timestamptz,
         modified_ts -> Timestamptz,
     }


tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -88,3 +125,77 @@ where
...
+        gw.upsert_protocol_types(protocol_types_slice.as_slice(), &mut conn)
+            .await
+            .unwrap();
+
+        let inserted_data = schema::protocol_type::table
+            .filter(schema::protocol_type::name.eq("Protocol"))
+            .select(schema::protocol_type::all_columns)
+            .first::<orm::ProtocolType>(&mut conn)
+            .await // Use optional() to handle the case when the record is not found
+            .unwrap();
```
tamaralipows: AFAIK we aren't adding a method to get these from the db, so I'm manually testing the retrieval. Pls tell me if I'm blind and missed something

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -251,6 +251,16 @@ pub struct ProtocolType {
     pub modified_ts: NaiveDateTime,
 }
 
+#[derive(AsChangeset, Insertable)]
+#[diesel(table_name = protocol_type)]
+#[diesel(check_for_backend(diesel::pg::Pg))]
+pub struct NewProtocolType {
```
SmartMonkey-git, line 278: Interesting.
Does that mean, that there are different structs for different database related functionalities?
`NewProtocolType` for inserting into the database and `ProtocolType` for querying/selecting?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -41,6 +47,38 @@ where
...
 
+    async fn upsert_protocol_types(
+        &self,
+        new: &[&ProtocolType],
+        conn: &mut Self::DB,
+    ) -> Result<(), StorageError> {
+        use super::schema::protocol_type::dsl::*;
+
+        for new_protocol_type in new {
+            let values = NewProtocolType {
```
SmartMonkey-git: You could implement `From<ProtocolType> for NewProtocolType`.

tycho-indexer/src/storage/mod.rs
```diff
@@ -502,6 +502,21 @@ pub trait ProtocolGateway {
...
 
+    /// Stores new found ProtocolTypes or update if existing.
+    ///
+    /// # Parameters
+    /// - `new`  The new protocol types.
+    ///
+    /// # Returns
+    /// Ok if stored successfully, may error if:
+    /// - related entities are not in store yet.
+    /// - type with same is id already present.
```
dianacarvalho1:
```suggestion
    /// - type with same id is already present.
```

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -41,6 +47,38 @@ where
     //     todo!()
     // }
 
+    async fn upsert_protocol_types(
+        &self,
+        new: &[&ProtocolType],
```
dianacarvalho1: does it need to be a list? 🤔 will we ever want to upsert more than one at the same time?

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -251,6 +251,16 @@ pub struct ProtocolType {
...
+    pub name: String,
+    pub financial_type: FinancialProtocolType,
```
dianacarvalho1: this is an enum right? and it works just like that? 🤔 @SmartMonkey-git didn't you need to do something more because of enums?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -88,3 +126,77 @@ where
...
+        evm::Block,
+        evm::Transaction,
+        evm::Account,
+        evm::AccountUpdate,
+        evm::ERC20Token,
+    >;
+
+    async fn setup_db() -> AsyncPgConnection {
+        // let db_url = std::env::var("DATABASE_URL").unwrap();
+        let db_url = "postgres://postgres:mypassword@localhost:5432/tycho_indexer_0".to_string();
```
dianacarvalho1: 👀 👀 👀 👀 👀 👀 👀 👀 👀 👀 👀 👀 👀 👀 👀

tycho-indexer/src/storage/postgres/protocol.rs
dianacarvalho1: why aren't we?? 👀 I think we need one no? I think we agreed that the extractor would upsert the protocol types on startup and then use them for the following components. 
Anyway, I think it's good practice not to use gtw methods in tests of other methods!

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -88,3 +126,77 @@ where
...
+        conn
+    }
+
+    #[tokio::test]
+    async fn test_add_protocol_type() {
+        let mut conn = setup_db().await;
+        let gw = EVMGateway::from_connection(&mut conn).await;
+
+        let d = NaiveDate::from_ymd_opt(2015, 6, 3).unwrap();
+        let t = NaiveTime::from_hms_milli_opt(12, 34, 56, 789).unwrap();
```
dianacarvalho1, line 207: ew

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -88,3 +126,77 @@ where
...
+            id: 1,
+            name: "Protocol".to_string(),
+            financial_type: FinancialProtocolType::Debt,
+            attribute_schema: None,
+            implementation: ProtocolImplementationType::Custom,
+            inserted_ts: dt,
+            modified_ts: dt,
+        }];
+
+        let protocol_types_slice: Vec<&ProtocolType> = protocol_types.iter().collect();
```
dianacarvalho1: what does this mean? could you explain please?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -88,3 +126,77 @@ where
...
+    async fn test_add_protocol_type() {
+        let mut conn = setup_db().await;
+        let gw = EVMGateway::from_connection(&mut conn).await;
+
+        let d = NaiveDate::from_ymd_opt(2015, 6, 3).unwrap();
+        let t = NaiveTime::from_hms_milli_opt(12, 34, 56, 789).unwrap();
+        let dt = NaiveDateTime::new(d, t);
+
+        let protocol_types: Vec<ProtocolType> = vec![ProtocolType {
+            id: 1,
```
dianacarvalho1: we define the id on our own? 🤔

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -88,3 +126,77 @@ where
...
+            .filter(schema::protocol_type::name.eq("Protocol"))
+            .select(schema::protocol_type::all_columns)
+            .first::<orm::ProtocolType>(&mut conn)
+            .await // Use optional() to handle the case when the record is not found
+            .unwrap();
+
+        assert_eq!(inserted_data.name, "Protocol".to_string());
+        assert_eq!(inserted_data.financial_type, FinancialProtocolType::Debt);
+        assert_eq!(inserted_data.attribute_schema, None);
+        assert_eq!(inserted_data.implementation, ProtocolImplementationType::Custom);
```
dianacarvalho1: could we test the update too please? and with something in the attribute_schema

tycho-indexer/src/storage/postgres/orm.rs
SmartMonkey-git: > didn't you need to do something more because of enums?

Yes.
If you follow `FinancialProtocolType` you can see that it is not the same as the enum `FinancialType`.
`FinancialProtocolType` or rather `financial_protocol_type` gets declared as a custom postgres type in `up.sql`. 
For the ProtocolSystem enum I followed the instructions of the [diesel-derive-enum](https://github.com/adwhit/diesel-derive-enum) repo. They a different from what you found here, which surprises me, as i'm sure we are using that library already.

SmartMonkey-git: Also looking at the enum and the postgres type, there might already be a bug, as the values are different.
image

SmartMonkey-git: @tamaralipows could you check if it is possible to insert `lend`?

tycho-indexer/src/storage/postgres/orm.rs
tamaralipows: Yes, that's how the other components are done

tycho-indexer/src/storage/postgres/protocol.rs
tamaralipows: fml

tycho-indexer/src/storage/postgres/protocol.rs
tamaralipows: This is the literal example given in the NativeDateTime docstring xD

tycho-indexer/src/storage/postgres/orm.rs
tamaralipows: lend is not a type there? Or what do you mean @SmartMonkey-git . I am overall confused by this thread tbh can someone summarize more clearly for me? @dianacarvalho1

tycho-indexer/src/storage/postgres/protocol.rs
tamaralipows: I had the same question, I found this paradigm on the other methods. I'm going to remove this, so unless someone cries it is no longer a list :)

tycho-indexer/src/storage/postgres/protocol.rs
tamaralipows: Hmm I don't see this done in `NewTransaction`, `NewState` or `NewBlock`. I'm afraid of inconsistency here. Unless they are implemented and I've missed it.

tycho-indexer/src/storage/postgres/protocol.rs
tamaralipows: some black magic to make it passable as a list. I'm removing the list though, so I'm removing this.

tycho-indexer/src/extractor/evm/storage.rs
```diff
@@ -113,6 +119,20 @@ pub mod pg {
...
+        }
+
+        fn to_storage(&self) -> orm::NewProtocolType {
+            orm::NewProtocolType {
+                name: self.name.into(),
+                implementation: ProtocolImplementationType::Custom,
+                attribute_schema: self.attribute_schema.into(),
+                financial_type: FinancialProtocolType::Debt,
+            }
+        }
```
tamaralipows: These implementations are fixed in the next commit.

tycho-indexer/src/models.rs
```diff
@@ -22,34 +22,6 @@ pub enum ProtocolSystem {
...
-/// similarly are grouped under a specific `ProtocolType` (i.e. Pool, Factory) within a
-/// `ProtocolSystem`.
-#[derive(PartialEq, Debug, Clone, Default, Deserialize, Serialize)]
-pub struct ProtocolType {
-    pub name: String,
-    pub attribute_schema: serde_json::Value,
-    pub financial_type: FinancialType,
-    pub implementation_type: ImplementationType,
-}
-
```
tamaralipows, line 53: Don't know why these were in models.rs before, all other relevant objects were in the evm subdir. I moved them to match the others.

tycho-indexer/migrations/2024-01-11-195345_protocol_type_unique_name/down.sql
```diff
@@ -0,0 +1 @@
+-- This file should undo anything in `up.sql`
```
SmartMonkey-git: I think you should drop the constraint that you added in the up.sql here.

tycho-indexer/src/storage/mod.rs
```diff
@@ -224,6 +224,27 @@ pub trait StorableTransaction<S, N, I>: Sized + Send + Sync + 'static {
...
+/// Generics:
+/// * `S`: This represents the storage-specific data type used when converting from storage to the
+///   transaction.
+/// * `N`: This represents the storage-specific data type used when converting from the protocol
+///   type to storage.
+/// * `I`: Represents the type of the database identifier, which is used as an argument in the
+///   conversion function. This facilitates the passage of database-specific foreign keys to the
+///   `to_storage` method, thereby providing a flexible way for different databases to interact with
+///   the transaction.
+pub trait StorableProtocolType<S, N, I>: Sized + Send + Sync + 'static {
```
SmartMonkey-git, line 239: Do we really need generic `I` in the interface or did you add it for consistency sake?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -1,30 +1,39 @@
...
 #[async_trait]
-impl<B, TX, A, D, T> ProtocolGateway for PostgresGateway<B, TX, A, D, T>
+impl<B, TX, A, D, T, PT> ProtocolGateway for PostgresGateway<B, TX, A, D, T, PT>
 where
     B: StorableBlock<orm::Block, orm::NewBlock, i64>,
     TX: StorableTransaction<orm::Transaction, orm::NewTransaction, i64>,
     D: ContractDelta + From<A>,
     A: StorableContract<orm::Contract, orm::NewContract, i64>,
     T: StorableToken<orm::Token, orm::NewToken, i64>,
+    PT: StorableProtocolType<orm::ProtocolType, orm::NewProtocolType, i64>,
```
SmartMonkey-git: Do you understand why it is necessary to add the table classes as generics? Can you explain it to me, I dont think I understand it.

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -88,3 +118,105 @@ where
...
+            .await
+            .unwrap();
+        conn.begin_test_transaction()
+            .await
+            .unwrap();
+        conn
+    }
+
+    #[tokio::test]
+    async fn test_add_protocol_type() {
```
SmartMonkey-git, line 202: Test looks good!

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -77,6 +77,40 @@ impl Transaction {
     }
 }
 
+#[derive(PartialEq, Debug, Clone, Default, Deserialize, Serialize)]
+pub enum FinancialType {
+    #[default]
+    Swap,
+    Psm,
+    Debt,
+    Leverage,
```
dianacarvalho1: ah you changed them to match the definition in the migration correct?

tycho-indexer/src/extractor/evm/storage.rs
```diff
@@ -113,6 +117,44 @@ pub mod pg {
             self.hash.into()
         }
     }
+    impl StorableProtocolType<orm::ProtocolType, orm::NewProtocolType, i64> for evm::ProtocolType {
+        fn from_storage(val: orm::ProtocolType) -> Result<Self, StorageError> {
+            let financial_type: FinancialType = match val.financial_type {
+                FinancialProtocolType::Swap => FinancialType::Swap,
```
dianacarvalho1: why do they have different names? 👀  could we make this consistent? 🥺

tycho-indexer/src/extractor/evm/storage.rs
```diff
@@ -113,6 +117,44 @@ pub mod pg {
...
+                FinancialProtocolType::Psm => FinancialType::Psm,
+                FinancialProtocolType::Debt => FinancialType::Debt,
+                FinancialProtocolType::Leverage => FinancialType::Leverage,
+            };
+            let implementation_type: ImplementationType = match val.implementation {
+                ProtocolImplementationType::Custom => ImplementationType::Custom,
+                ProtocolImplementationType::Vm => ImplementationType::Vm,
```
dianacarvalho1: same here.. or there might be a FinancialType or a ImplementationType that is not protocol specific? 🤔

tycho-indexer/src/models.rs
dianacarvalho1: because maybe these are not evm specific?

tycho-indexer/src/storage/mod.rs
```diff
@@ -502,6 +526,21 @@ pub trait ProtocolGateway {
     //     ids: Option<&[&str]>,
     // ) -> Result<Vec<Self::ProtocolComponent>, StorageError>;
 
+    /// Stores new found ProtocolTypes or update if existing.
```
dianacarvalho1:
```suggestion
    /// Stores new found ProtocolTypes or updates if existing.
```

tycho-indexer/src/storage/mod.rs
```diff
@@ -502,6 +526,21 @@ pub trait ProtocolGateway {
...
+    ///
+    /// # Parameters
+    /// - `new`  The new protocol types.
+    ///
+    /// # Returns
+    /// Ok if stored successfully, may error if:
+    /// - related entities are not in store yet.
```
dianacarvalho1: huum.. will it really fail? in your test you aren't adding those entries before hand and it is working 👀

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -88,3 +118,105 @@ where
...
+            .unwrap();
+
+        let inserted_data = schema::protocol_type::table
+            .filter(schema::protocol_type::name.eq("Protocol"))
+            .select(schema::protocol_type::all_columns)
+            .first::<orm::ProtocolType>(&mut conn)
+            .await
+            .unwrap();
+
+        assert_eq!(inserted_data.name, "Protocol".to_string());
```
SmartMonkey-git, line 228: Could you also check the ID and run this test twice?
For me, the ID increments per test run.

tycho-indexer/src/storage/postgres/protocol.rs
dianacarvalho1: yaaas very comprehensive test!! 🙏🏼

tycho-indexer/src/storage/mod.rs
```diff
@@ -502,6 +526,21 @@ pub trait ProtocolGateway {
...
+    /// - type with same id is already present.
```
dianacarvalho1: this is not true no? if we try to upsert with the same id (same name) it will just update it

tycho-indexer/src/storage/postgres/protocol.rs
louise-poole: I think we can actually avoid this. Have a look at how I added `ProtocolState` just a few lines below here... it's not a generic added to the implementation type, but rather a hardcoded type. This is because `ProtocolState` is already generic enough and the same `ProtocolState` struct is expected to be used for all protocols we implement. I would assume this also applies to your ProtocolType.

louise-poole: I'm not 100% sure my thought process was correct though... I just felt that this was best as a start and if later we need to make it a trait generic it's easy to do. In my project brief it said to implement what works and we'll make things generic later as necessary.

tycho-indexer/src/models.rs
louise-poole: Yes, these aren't EVM specific so should stay here with the generic models. We'll add non evm chains at some point and need to reuse these structs there too.

tycho-indexer/src/storage/postgres/protocol.rs
tamaralipows: is that bad?

tycho-indexer/src/storage/postgres/protocol.rs
tamaralipows: @louise-poole I can't add this like this:

```
    type DB = AsyncPgConnection;
    type Token = T;
    type ProtocolState = ProtocolState;
    type ProtocolType = StorableProtocolType<orm::ProtocolType, orm::NewProtocolType, i64>;
```

Because then I need to add the `dyn` type in front of it, which then gives me

```
he size for values of type `(dyn storage::StorableProtocolType<storage::postgres::orm::ProtocolType, storage::postgres::orm::NewProtocolType, i64> + 'static)` cannot be known at compilation time [E0277] doesn't have a size known at compile-time
```

Or are you suggesting to eliminate the `StorableProtocolType` completely?

I don't think I understand the difference between your ProtocolState and the generic types that have already been added? What does it mean to "be generic enough"?

tycho-indexer/src/storage/postgres/protocol.rs
SmartMonkey-git: Yes, that means the first run will insert id 1 and the second run will insert id 2.
When you assert for a certain id, then you can only run the test once with success, before restarting the database.

SmartMonkey-git: But it seems to be not a problem of your implementation but a general one. I just want to confirm that it also concerns you or if i did something wrong

tycho-indexer/src/storage/postgres/protocol.rs
louise-poole: I'm suggesting you use the struct and not the trait implementation. So you have 
```
type ProtocolType = ProtocolType;
```
Because you should have implemented the storable trait for that ProtocolType struct already:
```
impl StorableProtocolType<orm::ProtocolType, orm::NewProtocolType, i64> for evm::ProtocolType {
```
which means you can call these StorableProtocolType functions for any ProtocolType object.

louise-poole: By 'generic enough' I mean the ProtocolState struct isn't protocol specific. Like in python we have a new class per protocol with specific attribute fields etc. In our rust implementation we have a hashmap for attributes so the same ProtocolState struct/class can be used for all protocols.

louise-poole: When you used the `StorableProtocolType` as a trait generic (what you were doing) you were telling it that on initialisation it needs to set PT type as a struct that implements the StorableProtocolType traits. I'm telling you that you can hardcode it to a suitable struct (`ProtocolType`) that we already know implements the `StorableProtocolType` trait.

tycho-indexer/src/extractor/evm/mod.rs
tamaralipows: oui. Not the migration but the orm type

tycho-indexer/src/storage/mod.rs
tamaralipows: Consistency's sake. Should I remove it? I don't know tbh

tycho-indexer/src/storage/postgres/protocol.rs
tamaralipows: We do drop the db after every test though, no? I've been running this test like crazy

tycho-indexer/src/extractor/evm/storage.rs
tamaralipows: These were here before I arrived T_T I don't know if there is a good reason? Just can get confusing with the imports I guess. I will change this in a separate commit though so it's easier to undo if anyone cries

tycho-indexer/src/storage/mod.rs
tamaralipows: copy pasta

tycho-indexer/src/extractor/evm/storage.rs
tamaralipows: Same as other comment - I will change this in a separate commit so it's easier to undo if anyone cries

tycho-indexer/src/extractor/evm/mod.rs
tamaralipows: wait a sec. Do I need to make a migration when I change these? No, right?

tycho-indexer/src/storage/mod.rs
SmartMonkey-git: I don't know, as well.
I would say before this will cause a big interface change in the future, lets carrie it.

WDYT?

tycho-indexer/src/storage/postgres/protocol.rs
SmartMonkey-git: > We do drop the db after every test though, no?

From my limited understand `begin_test_transaction`, does not drop the db with conclusion of the test, but it prevents your test from committing changes. I think, those are different approaches of db hygiene .

> I've been running this test like crazy

The problem arise, when you have an auto-incrementing column that you assert. For some reason, the database instance keeps incrementing those columns, despite a need for committing.

tycho-indexer/src/extractor/evm/mod.rs
dianacarvalho1: if they don't match with the current migration, then yes, you do

tycho-indexer/src/storage/mod.rs
tamaralipows: I agree!

tycho-indexer/src/storage/postgres/protocol.rs
tamaralipows: I think this isn't a problem in my case?

tamaralipows: ((we talked about this in our meeting and decided to add another task for this))

PR #86, opened 2024-01-11
[WIP] ENG 2437: Implement ProtocolGateway update_states

louise-poole:
Implemented the logic to update protocol states in the db. 

The rough outline of the logic implemented is as follows:
Given a list of ProtocolStateDeltas (changes to the protocol state by either updating an attribute value or deleting an attribute), we first find all linked txns and component entities already in the db (we need their db ids to be able to insert protocolStates correctly).
Next we loop through all these deltas and prep the attribute updates for storage. If there are any deleted attributes, we update their related db entity that is currently acknowledged as valid (valid_to is null) and set their valid_to to this ts. 
After that we sort all updates: first group them by component id and attribute name, then sort by block and txn index. This allows us to loop through all updates and easily determine if any updates overwrite a previous one. If there is an attribute that is to be immediately overwritten, we set their "valid_to" appropriately. 
Finally, we insert all new updates into the db.

Note: the logic to invalidate old db entities that are now outdated by the new entities added has not been implemented. This is planned [here](https://datarevenue.atlassian.net/browse/ENG-2682).

### DIFF
--- a/tycho-indexer/src/extractor/evm/mod.rs
+++ b/tycho-indexer/src/extractor/evm/mod.rs
@@ -498,8 +498,11 @@ impl TvlChange {
 /// smart contract.
 #[derive(Debug, Clone, PartialEq, Default, Deserialize, Serialize)]
 pub struct ProtocolComponent {
-    // an id for this component, could be hex repr of contract address
-    pub id: ContractId,
+    /// Is the unique identifier of a contract. It can represent an on-chain
+    /// address or in the case of a one-to-many relationship it could be something like
+    /// 'USDC-ETH'. This is for example the case with ambient, where one contract is
+    /// responsible for multiple components.
+    pub id: String,
     // what system this component belongs to
     pub protocol_system: String,
     // more metadata information about the components general type (swap, lend, bridge, etc.)
@@ -520,15 +523,6 @@ pub struct ProtocolComponent {
     pub created_at: NaiveDateTime,
 }
 
-/// A type representing the unique identifier for a contract. It can represent an on-chain address
-/// or in the case of a one-to-many relationship it could be something like 'USDC-ETH'. This is for
-/// example the case with ambient, where one component is responsible for multiple contracts.
-///
-/// `ContractId` is a simple wrapper around a `String` to ensure type safety
-/// and clarity when working with contract identifiers.
-#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, Default)]
-pub struct ContractId(pub String);
-
 impl ProtocolComponent {
     pub fn try_from_message(
         msg: substreams::ProtocolComponent,
@@ -538,8 +532,6 @@ impl ProtocolComponent {
         tx_hash: H256,
         creation_ts: NaiveDateTime,
     ) -> Result<Self, ExtractionError> {
-        let id = ContractId(msg.id.clone());
-
         let tokens = msg
             .tokens
             .clone()
@@ -562,7 +554,7 @@ impl ProtocolComponent {
             .collect::<Result<HashMap<_, _>, ExtractionError>>()?;
 
         Ok(Self {
-            id,
+            id: msg.id.clone(),
             protocol_type_id: protocol_type_id.to_owned(),
             protocol_system: protocol_system.to_owned(),
             tokens,
@@ -912,7 +904,7 @@ impl BlockEntityChanges {
                             tx.hash,
                             block.ts,
                         )?;
-                        new_protocol_components.insert(pool.clone().id.0, pool);
+                        new_protocol_components.insert(pool.id.clone(), pool);
                     }
                 }
             }
@@ -1455,7 +1447,7 @@ mod test {
             index: 2,
         };
         let protocol_component = ProtocolComponent {
-            id: ContractId("0xaaaaaaaaa24eeeb8d57d431224f73832bc34f688".to_owned()),
+            id: "0xaaaaaaaaa24eeeb8d57d431224f73832bc34f688".to_owned(),
             protocol_system: "ambient".to_string(),
             protocol_type_id: String::from("id-1"),
             chain: Chain::Ethereum,
@@ -1542,7 +1534,7 @@ mod test {
     fn block_account_changes() -> BlockAccountChanges {
         let address = H160::from_low_u64_be(0x0000000000000000000000000000000061626364);
         let protocol_component = ProtocolComponent {
-            id: ContractId("0xaaaaaaaaa24eeeb8d57d431224f73832bc34f688".to_owned()),
+            id: "0xaaaaaaaaa24eeeb8d57d431224f73832bc34f688".to_owned(),
             protocol_system: "ambient".to_string(),
             protocol_type_id: String::from("id-1"),
             chain: Chain::Ethereum,
@@ -1868,7 +1860,7 @@ mod test {
         let new_protocol_components: HashMap<String, ProtocolComponent> = vec![(
             "Pool".to_owned(),
             ProtocolComponent {
-                id: ContractId("Pool".to_owned()),
+                id: "Pool".to_owned(),
                 protocol_system: "ambient".to_string(),
                 protocol_type_id: "Pool".to_owned(),
                 chain: Chain::Ethereum,
@@ -1978,7 +1970,7 @@ mod test {
         let new_protocol_components: HashMap<String, ProtocolComponent> = vec![(
             "Pool".to_owned(),
             ProtocolComponent {
-                id: ContractId("Pool".to_owned()),
+                id: "Pool".to_owned(),
                 protocol_system: "ambient".to_string(),
                 protocol_type_id: "Pool".to_owned(),
                 chain: Chain::Ethereum,
@@ -2076,7 +2068,7 @@ mod test {
         let protocol_component = result.unwrap();
 
         // Assert specific properties of the protocol component
-        assert_eq!(protocol_component.id, ContractId("component_id".to_string()));
+        assert_eq!(protocol_component.id, "component_id".to_string());
         assert_eq!(protocol_component.protocol_system, expected_protocol_system);
         assert_eq!(protocol_component.protocol_type_id, protocol_type_id);
         assert_eq!(protocol_component.chain, expected_chain);
--- a/tycho-indexer/src/storage/postgres/protocol.rs
+++ b/tycho-indexer/src/storage/postgres/protocol.rs
@@ -505,10 +505,7 @@ where
 mod test {
     use super::*;
     use crate::{
-        extractor::{
-            evm,
-            evm::{ContractId, ERC20Token},
-        },
+        extractor::evm::{self, ERC20Token},
         storage::ChangeType,
     };
     use chrono::{NaiveDate, NaiveDateTime, NaiveTime};
@@ -1047,7 +1044,7 @@ mod test {
         let protocol_system = "ambient".to_string();
         let chain = Chain::Ethereum;
         let original_component = ProtocolComponent {
-            id: ContractId("test_contract_id".to_string()),
+            id: "test_contract_id".to_string(),
             protocol_system,
             protocol_type_id: protocol_type_id_1.to_string(),
             chain,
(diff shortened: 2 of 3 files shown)


tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -26,7 +26,7 @@ use super::ExtractionError;
 
 pub mod ambient;
 pub mod storage;
-mod utils;
+pub mod utils;
```
kayibal: Why? Where are you using this outside of here?

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -299,6 +311,19 @@ pub struct NewProtocolComponent {
     pub attributes: Option<serde_json::Value>,
 }
 
+impl ProtocolComponent {
+    pub async fn by_external_ids(
+        external_ids: &[&str],
+        conn: &mut AsyncPgConnection,
+    ) -> QueryResult<Vec<Self>> {
```
kayibal: I don't love that you retrieve all rows, and then further down only end up using a few.

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -199,6 +199,18 @@ impl Transaction {
             .first::<Self>(conn)
             .await
     }
+
+    pub async fn by_hashes(
+        hashes: &[&[u8]],
+        conn: &mut AsyncPgConnection,
+    ) -> QueryResult<Vec<(Self, NaiveDateTime)>> {
```
kayibal: Same here a lot of data retrieved and then usually only a single attribute accessed. I think the correct way to deal with this would be to pass into there also the columns that you would like to have but that would require the use of generics. You could try this though:

```
pub async fn by_hashes<C, R>(
        hashes: &[&[u8]],
        columns: R,
        conn: &mut AsyncPgConnection,
    ) -> QueryResult<Vec<(R, NaiveDateTime)>> {
```

Most likely Rust will complain about some missing trait bounds (at least Selectable), just add them and then hopefully it will work^^

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -176,13 +171,101 @@ where
...
+                    // 'valid_to' is None
+                    if let Some((matching_state, matching_index)) =
+                        state_data
+                            .iter_mut()
+                            .find(|(existing_state, existing_index)| {
+                                existing_state.protocol_component_id ==
+                                    new_state.protocol_component_id &&
+                                    existing_state.attribute_name == new_state.attribute_name &&
+                                    existing_state.valid_to.is_none()
+                            })
```
kayibal: Noooo haha, as you already knew this is very inefficient it has quadratic runtime. The strategy for this is:

Sort by entity id, here `component_id`, and time, here `(block, transaction index)`. Then use a for loop and iterate with a lookahead index. If the the entry at i+1 has the same component id, you can directly invalidate the current one. 

That is it and this will have `O(nlogn)` complexity, so it is much faster.

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -176,13 +171,101 @@ where
...
+                        if matching_state.valid_from < new_state.valid_from {
+                            matching_state.valid_to = Some(new_state.valid_from);
+                        } else if matching_state.valid_from > new_state.valid_from {
+                            new_state.valid_to = Some(matching_state.valid_from);
+                        } else if matching_index < tx_index {
+                            matching_state.valid_to = Some(new_state.valid_from);
+                        } else {
+                            new_state.valid_to = Some(matching_state.valid_from);
+                        }
+                    }
```
kayibal: Sorting data avoids this.

tycho-indexer/src/storage/postgres/orm.rs
kayibal: Just noticed you'll probably want to also retrieve the block number here (to correctly sort later on)

tycho-indexer/src/extractor/evm/mod.rs
louise-poole: I use it to decode bytes into a tx hash in `update_states`

tycho-indexer/src/storage/postgres/orm.rs
louise-poole: Adding the generics started to get quite complicated. Since for now this is only used in one place I decided to just adjust the code to return what I need. If one day we need a generic version we can spend time adjusting it then. Hope this is fine?

tycho-indexer/src/extractor/evm/mod.rs
tvinagre: Maybe these functions should be in a less nested directory. Seems like they're very general

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -218,6 +218,19 @@ impl Transaction {
 
         Ok(results.into_iter().collect())
     }
+
+    // fetches the transaction id, hash, index and block timestamp for a given set of hashes
+    pub async fn id_by_hashes(
```
tvinagre: ids_by_hashes

tvinagre: Or even, `get_txs_by_hashes`. Since we're returning a bunch of transactions. Or `get_transaction_details_by_hashes` if you want to be more verbose

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -257,43 +260,63 @@ where
...
+            // invalidated db entities for deleted attributes
+            for attr in &state.deleted_attributes {
+                diesel::update(schema::protocol_state::table)
+                    .filter(schema::protocol_state::protocol_component_id.eq(component_db_id))
+                    .filter(schema::protocol_state::attribute_name.eq(attr))
+                    .filter(schema::protocol_state::valid_to.is_null())
+                    .set(schema::protocol_state::valid_to.eq(tx_db.2))
+                    .execute(conn)
+                    .await?;
+            }
```
kayibal: Potentially problematic cause of the nested loop, but I know that diesel has currently no other way of doing this. 
Can you please add a comment here with `// PERF: slow but required due to diesel restrictions`

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -257,43 +260,63 @@ where
...
+                if sub_order == Ordering::Equal {
+                    // Sort by block ts and tx_index as well
+                    a.1.cmp(&b.1)
+                } else {
+                    sub_order
+                }
+            } else {
+                order
+            }
+        });
```
kayibal, line 304: I am almost done generalising this, all you'll need to do here is to implement a `sort_key_id()` method on the `orm::ProtocolState` struct and a few other getters and setters and then some generic methods will do this work here for you.

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -217,13 +217,121 @@ where
...
+            new.iter()
+                .map(|state| state.modify_tx.as_bytes())
+                .collect::<Vec<&[u8]>>()
+                .as_slice(),
+            conn,
+        )
+        .await?
+        .into_iter()
+        .map(|(id, hash, index, ts)| {
+            (H256::try_decode(&hash, "tx hash").expect("Failed to decode tx hash"), (id, index, ts))
```
kayibal: Ah ok this is where you use `evm::utils`, I think something is off here you should not be using evm specific logic.

Hmm ok it's because we have no abstraction on ProtocolStateDelta, could you maybe move this to `ProtocolStateDelta.txhash_as_bytes` add a comment that this serves as a placeholder for adding sth like `StorableProtocolStateDelta` for now.

Starknet is likely going to be using slightly different types here that is why the input models should give you access to retrieve these kind of values directly as bytes but adding such an abstraction is a lot of work so we should better skip it for now.

tycho-indexer/src/extractor/evm/mod.rs
kayibal: This module is on purpose tucked away like this :D, I added a comment below explaining how we could resolve this for now.

tycho-indexer/src/storage/postgres/protocol.rs
louise-poole: Will do! I spent a while trying to find a better way to do it. Felt like this wasn't the worst option since all db writes are collected by the cache gateway anyway so it'll still be executed in a single db transaction.

tycho-indexer/src/storage/postgres/protocol.rs
louise-poole: Hmm... Okay, to clarify we are converting from Bytes to H256 here. It didn't seem appropriate to put this functionality in `ProtocolStateDelta`, so I added it to `Transaction` and called it `hash_from_bytes`.

tycho-indexer/src/storage/postgres/orm.rs
louise-poole: True. Considering we're just getting the ids and the timestamp info (block ts and index) I'll maybe rename it `ids_and_ts_by_hashes`?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -10,7 +10,7 @@ use ethers::types::H256;
 use tracing::warn;
 
 use crate::{
-    extractor::evm::{utils::TryDecode, ProtocolComponent, ProtocolState, ProtocolStateDelta},
+    extractor::evm::{ProtocolComponent, ProtocolState, ProtocolStateDelta, Transaction},
```
kayibal: Now you had to add this unrelated `Transaction` import, I still believe this is better under ProtocolComponent

tycho-indexer/src/storage/postgres/protocol.rs
kayibal: Mh no the cache gateway are just delayed calls of this methods so that is not how t works, the transaction just ensures atomicity there round trip overheads still exist.

tycho-indexer/src/storage/postgres/protocol.rs
kayibal: Yeah still would prefer to see this on `ProtocolState` and potentially duplicated on `ProtocolStateDelta` with clear comments that this is a temporary workaround until we implement and add the traits `StorableProtocolState` and `StorableProtocolStateDelta`, afaik we have storage abstractions for `Transaction` already so putting it there is even less appropriate.

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -217,13 +217,120 @@ where
...
+        .into_iter()
+        .map(|(id, external_id)| (external_id, id))
+        .collect();
+
+        let mut state_data: Vec<(orm::NewProtocolState, i64)> = Vec::new();
+
+        for state in new {
+            let tx_db = txns
+                .get(&state.modify_tx)
+                .expect("Failed to find tx");
```
kayibal: I was wondering so badly why we need the H256 type here... and I think we don't need it. You could just index with `&state.modify_ty.as_bytes()` here, is that possible? It's alright if not... cause this isn't great either: `as_bytes` is a method on H256 but at least the txns `HashMap` would be chain independent.

The main issue here is that `ProtocolState` and `Delta` are chain specific so their transaction hash types may differ per chain. It seems like currently we are hardcoding things to ethereum. Which is alright.

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -217,13 +217,120 @@ where
...
+        let txns: HashMap<H256, (i64, i64, NaiveDateTime)> = orm::Transaction::ids_and_ts_by_hash(
+            new.iter()
+                .map(|state| state.modify_tx.as_bytes())
+                .collect::<Vec<&[u8]>>()
+                .as_slice(),
+            conn,
+        )
+        .await?
+        .into_iter()
+        .map(|(id, hash, index, ts)| (Transaction::hash_from_bytes(hash), (id, index, ts)))
```
kayibal: Here you could instead keep Bytes in the keys to avoid using H256.

tycho-indexer/src/storage/postgres/protocol.rs
louise-poole: Yeah, we took the hardcoding approach since it was emphasised that we shouldn't complicate things now and can introduce generics later when needed. 
As for why I opted to use the H256 type for the key: I simply did this because doing one conversion per key at HashMap initialisation is more efficient than doing repeated conversions within a loop whenever we need to fetch values from the HashMap. However I think the difference is minor - I am happy to make this part of the code more generic and use Bytes instead 👍

PR #87, opened 2024-01-12
ENG-1717: Implement StorableToken

dianacarvalho1:
[Issue](https://datarevenue.atlassian.net/browse/ENG-1717)

### DIFF
--- a/tycho-indexer/src/extractor/evm/storage.rs
+++ b/tycho-indexer/src/extractor/evm/storage.rs
@@ -19,14 +19,18 @@ use crate::{
 pub mod pg {
     use crate::extractor::evm::utils::pad_and_parse_h160;
     use ethers::types::{H160, H256, U256};
+    use serde_json::Value;
 
-    use crate::storage::{
-        postgres::{
-            orm,
-            orm::{NewProtocolState, NewToken, Token},
+    use crate::{
+        hex_bytes::Bytes,
+        storage::{
+            postgres::{
+                orm,
+                orm::{NewToken, Token},
+            },
+            Address, Balance, BlockHash, ChangeType, Code, StorableProtocolState, StorableToken,
+            TxHash,
         },
-        Address, Balance, BlockHash, ChangeType, Code, StorableProtocolState, StorableToken,
-        TxHash,
     };
 
     use super::*;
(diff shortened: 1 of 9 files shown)


tycho-indexer/src/extractor/evm/storage.rs
```diff
@@ -287,3 +300,52 @@ pub mod pg {
...
+        assert!(result.is_ok());
+
+        let token = result.unwrap();
+        assert_eq!(token.address, pad_and_parse_h160(&token_address).unwrap());
+        assert_eq!(token.symbol, String::from("WETH"));
+        assert_eq!(token.decimals, 18);
+    }
+    #[test]
+    fn test_storable_token_to_storage() {
+        let token_address = "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2".into();
```
tamaralipows, line 383: what does into do in this scenario?

SmartMonkey-git: The compiler checks where `token_address` is used and derives the typing from that. In this case it automatically casts the string to `Bytes`, because thats what `pad_and_parse_h160` asks for.

tycho-indexer/src/extractor/evm/storage.rs
```diff
@@ -287,3 +300,52 @@ pub mod pg {
...
+    #[test]
+    fn test_storable_token_from_storage() {
+        let token_address: Address = "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2".into();
+        let orm_token = Token {
+            id: 1,
+            account_id: 1,
+            symbol: String::from("WETH"),
+            decimals: 18,
+            tax: 0,
+            gas: vec![],
```
SmartMonkey-git: Adding
```suggestion
            gas: vec![Some(64), None],
```

Would improve your test a little, because as it is right now, your map lambda is not tested. Same for the `to_storage` test.

tycho-indexer/src/extractor/evm/storage.rs
```diff
@@ -251,22 +252,34 @@ pub mod pg {
 
     impl StorableToken<orm::Token, orm::NewToken, i64> for evm::ERC20Token {
         fn from_storage(val: Token, contract: ContractId) -> Result<Self, StorageError> {
```
SmartMonkey-git, line 258: Why do we need to pass the `ContractId` additionally to the token orm?
Shouldn't the `id` or `account_id` of the token orm already be that address?

SmartMonkey-git: Something seems to be off with the interfaces. 
If `id` or `account_id` are not the address of the token, then it seems odd that we are passing addresses in the `get_token` interface to presumably to filter for tokens.
Further, the address input of the `get_token` function is optional and if the interface of your `from_storage` function is correct, then you would be unable to even convert from orm to object in `get_token`, because you need to pass a contract id to the `from_storage` function.

On top of all of this the `id` or `account_id` are i64, which seems confusing to me. :confused:

dianacarvalho1: I'm not fully sure I understand your comment. 
But the issue is that the orm.Token doesn't have the token address to begin with. So only from a orm.Token we can't construct the struct Token. That's why we need the ContractId here as well.. I guess the token gtw will have to fetch both

PR #88, opened 2024-01-15
ENG-1728: StorableComponent trait

SmartMonkey-git:


tycho-indexer/src/extractor/evm/storage.rs
```diff
@@ -269,11 +269,70 @@ pub mod pg {
             todo!()
         }
 
-        fn contract_id(&self) -> ContractId {
+        fn contract_id(&self) -> storage::ContractId {
             todo!()
         }
     }
 
+    impl StorableComponent<orm::ProtocolComponent, orm::NewProtocolComponent, i64>
```
dianacarvalho1: `StorableProtocolComponent` no?

tycho-indexer/src/extractor/evm/storage.rs
```diff
@@ -269,11 +269,70 @@ pub mod pg {
...
+
+            if let Some(Value::Object(map)) = val.attributes {
+                static_attributes = map
+                    .into_iter()
+                    .map(|(key, value)| (key, Bytes::from(bytes::Bytes::from(value.to_string()))))
+                    .collect();
+            }
+
+            Ok(evm::ProtocolComponent {
+                id: evm::ContractId(val.external_id),
```
dianacarvalho1, line 347: aah this is yet the other `ContractId` right? the one that is just a string.. this is really confusing :/

tycho-indexer/src/extractor/evm/storage.rs
```diff
@@ -269,11 +269,70 @@ pub mod pg {
...
+            creation_ts: NaiveDateTime,
+        ) -> Result<orm::NewProtocolComponent, StorageError> {
+            let protocol_type_id = self
+                .protocol_type_id
+                .parse::<i64>()
+                .map_err(|err| {
+                    StorageError::DecodeError(
+                        "Could not parse protocol type id in StorableComponent".to_string(),
+                    )
+                })?;
```
dianacarvalho1, line 371: why is this needed?

tycho-indexer/src/extractor/evm/storage.rs
```diff
@@ -333,3 +392,136 @@ pub mod pg {
...
+mod test {
+    use super::*;
+    use crate::{models::ProtocolSystem, storage::postgres::orm};
+
+    use crate::{hex_bytes::Bytes, storage::StorableComponent};
+    use chrono::Utc;
+    use ethers::prelude::H160;
+    use std::str::FromStr;
+
+    #[tokio::test]
```
dianacarvalho1: tokio? why?

tycho-indexer/src/extractor/evm/storage.rs
```diff
@@ -333,3 +392,136 @@ pub mod pg {
...
+            attributes: Some(atts.clone()),
+            protocol_system_id: 0,
+            created_at: Default::default(),
+            deleted_at: None,
+            inserted_ts: Default::default(),
+            modified_ts: Default::default(),
+        };
+
+        let tokens = vec![
+            H160::from_str("0x6B175474E89094C44Da98b954EedeAC495271d0F").unwrap(),
```
dianacarvalho1, line 526: aaaahh didn't know you could do this 👀

tycho-indexer/src/extractor/evm/storage.rs
```diff
@@ -333,3 +392,136 @@ pub mod pg {
...
+            "{:?}",
+            std::str::from_utf8(
+                &protocol_component
+                    .static_attributes
+                    .get("key1")
+                    .unwrap()
+                    .0
+            )
+            .unwrap()
+        );
```
dianacarvalho1: print?

tycho-indexer/src/extractor/evm/storage.rs
```diff
@@ -333,3 +392,136 @@ pub mod pg {
...
+    use super::*;
+    use crate::{models::ProtocolSystem, storage::postgres::orm};
+
+    use crate::{hex_bytes::Bytes, storage::StorableComponent};
+    use chrono::Utc;
+    use ethers::prelude::H160;
+    use std::str::FromStr;
+
+    #[tokio::test]
+    async fn test_from_storage() {
```
dianacarvalho1: can you give the tests better names please? say that is related to the storable protocol. there are more of these methods in this file

tycho-indexer/src/extractor/evm/storage.rs
dianacarvalho1: to convert from the enum to an i64? 🤔

SmartMonkey-git: It is actually from String to i64.
I'm not sure why one is a string and the other an i64. Should we adjust either?

tycho-indexer/src/extractor/evm/storage.rs
SmartMonkey-git: Yeah, your are right.
There is actually a thread on this, where I suggest do merged both ContractId's

tycho-indexer/src/extractor/evm/storage.rs
SmartMonkey-git: Life can be so simple! :D

tycho-indexer/src/extractor/evm/storage.rs
dianacarvalho1: I vote yes on the merge!!

tycho-indexer/src/extractor/evm/storage.rs
dianacarvalho1: huuuuuummmmmmm I find this very confusing tbh :/ aaah wait, I think this needs to point to the row/entry ID in the database of the corresponding `ProtocolType`

tycho-indexer/src/extractor/evm/storage.rs
dianacarvalho1: well with rust I highly doubt that

tycho-indexer/src/extractor/evm/storage.rs
SmartMonkey-git: I would suggest to do it in an additional PR though, because It will involve some additional changes.

wdyt?

tycho-indexer/src/extractor/evm/storage.rs
SmartMonkey-git: Then should we change `evm::ProtocolComponent.protocol_type_id` to `evm::Protocoltype` or you think just `i64` would be sufficient?

dianacarvalho1: no no, just the correct ID pointing to the correct object should be enough. But how would you know the ID at this point? 🤔

tycho-indexer/src/extractor/evm/storage.rs
dianacarvalho1: yes ok ok

tycho-indexer/src/extractor/evm/storage.rs
```diff
@@ -333,3 +392,123 @@ pub mod pg {
...
+        let new_protocol_component = result.unwrap();
+
+        assert_eq!(new_protocol_component.external_id, protocol_component.id.0);
+        assert_eq!(new_protocol_component.chain_id, chain_id);
+
+        let parsed_protocol_type_id = protocol_component
+            .protocol_type_id
+            .parse::<i64>()
+            .unwrap();
+        assert_eq!(new_protocol_component.protocol_type_id, parsed_protocol_type_id);
```
tamaralipows: Aren't we repeating the code here that was implemented in the method? Shouldn't we just checked a hardcoded value instead?

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -290,6 +290,19 @@ pub struct ProtocolComponent {
...
+#[diesel(belongs_to(ProtocolType))]
+#[diesel(belongs_to(ProtocolSystem))]
+#[diesel(table_name = protocol_component)]
+pub struct NewProtocolComponent {
+    pub external_id: String,
+    pub chain_id: i64,
+    pub protocol_type_id: i64,
+    pub protocol_system_id: i64,
+    pub attributes: Option<serde_json::Value>,
+    pub created_at: NaiveDateTime,
```
tamaralipows: I don't think we add this `created_at` manually? Wouldn't this be whatever time it is inserted into the db? Or am I wrong

SmartMonkey-git: No, I dont think so, but I was also unsure. What makes me doubt is that there is an `inserted_ts` and `created_at`.

tamaralipows: I would remove this line from here (see how I've done it in the `NewProtocolType`)

tycho-indexer/src/extractor/evm/storage.rs
SmartMonkey-git: We would get all the ID's in the extract. Only thing we are missing to do so, is a `get_protocol_type_id` function. Which is not planned to do.
But I can do it in the future.

tycho-indexer/src/extractor/evm/storage.rs
SmartMonkey-git: https://datarevenue.atlassian.net/browse/ENG-2700

tycho-indexer/src/extractor/evm/storage.rs
dianacarvalho1: can you add a task for it please?

SmartMonkey-git: https://datarevenue.atlassian.net/browse/ENG-2715

PR #89, opened 2024-01-16
ENG-2042: Substreams TVL extraction on Ambient swaps

tamaralipows:
**Goal:** Emit TVL updates in substreams
I’ve looked at how it’s done in [mapping.ts](https://github.com/CrocSwap/croc-subgraph/blob/main/subgraphs/croc-mainnet/src/mapping.ts#L187) and found the following info:

- `AggEvent.baseFlow` and .`quoteFlow` are not null in `handleSwap` and `modifyLiquidity` method. So we need to deal with both of these methods, eventually.
- In the scope of this task, we care about the handleSwap method - this method is called 5 times:

1. `handleDirectSwapCall` - Handler for a swap() call made to CrocSwapDex
2. `handleDirectSwapEvent` - event _CrocSwap (address indexed base, address indexed quote, uint256 poolIdx, bool isBuy, bool inBaseQty, uint128 qty, uint16 tip, uint128 limitPrice, uint128 minOut, uint8 reserveFlags, int128 baseFlow, int128 quoteFlow)_;
**Comment:** I don't see this event emitted on their Ethereum [contracts]. On other chains it's emitted every time the swap method is called. So this shall be left out.
3. `handleHotProxy` for HotProxy swaps - i.e. whenever someone calls` userCmd(bytes)` on the [HotProxy contract](https://etherscan.io/address/0x37e00522Ce66507239d59b541940F99eA19fF81F#code)
**Question:** I have implemented this as is done in the mapping.ts code, but I ran it on **700k blocks** , and didn't find anything. Not sure how to test this...
4. `handleSweepSwapCall` - Handler for a [MicroPaths](https://etherscan.io/address/0xf241bEf0Ea64020655C70963ef81Fea333752367#writeContract) `sweepSwap()` call (as part of a long-form order) that performs a swap operation on a single pool 
5. `handleSweepSwapEvent` - event CrocMicroSwap(bytes input, bytes output).
**Comment:** I don’t even find where this event is emitted on their Ethereum [contracts](https://etherscan.io/address/aaaaaaaaa24eeeb8d57d431224f73832bc34f688#code) so I am leaving this out.  This seems to not be emitted on Ethereum mainnet.

Note: If you are wondering where I got these contract addresses, see [here](https://github.com/CrocSwap/croc-subgraph/blob/main/subgraphs/croc-mainnet/subgraph.yaml)

So all in all I am left with three cases:
- [x]  `handleDirectSwapCall` -> tested
- [x]  `handleHotProxy`  -> untestable?
- [x] `handleSweepSwapCall` -> tested

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -125,7 +125,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e01ed3140b2f8d422c68afa1ed2e85d996ea619c988ac834d255db32138655cb"
 dependencies = [
  "quote",
- "syn 2.0.40",
+ "syn 2.0.39",
 ]
 
 [[package]]
@@ -300,7 +300,7 @@ dependencies = [
  "actix-router",
  "proc-macro2",
  "quote",
- "syn 2.0.40",
+ "syn 2.0.39",
 ]
 
 [[package]]
@@ -325,7 +325,7 @@ checksum = "7c7db3d5a9718568e4cf4a537cfd7070e6e6ff7481510d0237fb529ac850f6d3"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.40",
+ "syn 2.0.39",
 ]
 
 [[package]]
@@ -408,9 +408,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.5"
+version = "0.6.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d664a92ecae85fd0a7392615844904654d1d5f5514837f471ddef4a057aba1b6"
+checksum = "2ab91ebe16eb252986481c5b62f6098f3b698a45e34b5b98200cf20dd2484a44"
 dependencies = [
  "anstyle",
  "anstyle-parse",
@@ -428,30 +428,30 @@ checksum = "7079075b41f533b8c61d2a4d073c4676e1f8b249ff94a393b0595db304e0dd87"
 
 [[package]]
 name = "anstyle-parse"
-version = "0.2.3"
+version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c75ac65da39e5fe5ab759307499ddad880d724eed2f6ce5b5e8a26f4f387928c"
+checksum = "317b9a89c1868f5ea6ff1d9539a69f45dffc21ce321ac1fd1160dfa48c8e2140"
 dependencies = [
  "utf8parse",
 ]
 
 [[package]]
 name = "anstyle-query"
-version = "1.0.2"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e28923312444cdd728e4738b3f9c9cac739500909bb3d3c94b43551b16517648"
+checksum = "5ca11d4be1bab0c8bc8734a9aa7bf4ee8316d462a08c6ac5052f888fef5b494b"
 dependencies = [
- "windows-sys 0.52.0",
+ "windows-sys",
 ]
 
 [[package]]
 name = "anstyle-wincon"
-version = "3.0.2"
+version = "3.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1cd54b81ec8d6180e24654d0b371ad22fc3dd083b6ff8ba325b72e00c87660a7"
+checksum = "f0699d10d2f4d628a98ee7b57b289abbc98ff3bad977cb3152709d4bf2330628"
 dependencies = [
  "anstyle",
- "windows-sys 0.52.0",
+ "windows-sys",
 ]
 
 [[package]]
@@ -494,7 +494,7 @@ checksum = "16e62a023e7c117e27523144c5d2459f4397fcc3cab0085af8e2224f643a0193"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.40",
+ "syn 2.0.39",
 ]
 
 [[package]]
@@ -505,7 +505,7 @@ checksum = "a66537f1bb974b254c98ed142ff995236e81b9d0fe4db0575f46612cb15eb0f9"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.40",
+ "syn 2.0.39",
 ]
 
 [[package]]
@@ -818,18 +818,18 @@ dependencies = [
 
 [[package]]
 name = "cargo-platform"
-version = "0.1.5"
+version = "0.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e34637b3140142bdf929fb439e8aa4ebad7651ebf7b1080b3930aa16ac1459ff"
+checksum = "12024c4645c97566567129c204f65d5815a8c9aecf30fcbe682b2fe034996d36"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "cargo_metadata"
-version = "0.18.1"
+version = "0.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2d886547e41f740c616ae73108f6eb70afe6d940c7bc697cb30f13daec073037"
+checksum = "e7daec1a2a2129eeba1644b220b4647ec537b0b5d4bfd6876fcc5a540056b592"
 dependencies = [
  "camino",
  "cargo-platform",
@@ -867,7 +867,7 @@ dependencies = [
  "num-traits",
  "serde",
  "wasm-bindgen",
- "windows-targets 0.48.5",
+ "windows-targets",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


substreams/ethereum-ambient/src/lib.rs
```diff
@@ -213,6 +213,96 @@ fn map_changes(
...
+                    {
+                        let base_flow = external_outputs[0]
+                            .to_owned()
+                            .into_int() // Needs conversion into bytes for next step
+                            .ok_or_else(|| anyhow!("Failed to convert to i128".to_string()))?;
+
+                        let quote_flow = external_outputs[1]
+                            .to_owned()
+                            .into_int() // Needs conversion into bytes for next step
+                            .ok_or_else(|| anyhow!("Failed to convert to i128".to_string()))?;
```
tamaralipows: @SmartMonkey-git I could use your expertise here. 
1. How on earth can I get this to an `I128`?
2. Should I really then proceed to store it as a bytes array? Why not store it as an int?

SmartMonkey-git: Ok, I have a suggestion for you, but I dont know if it is the best one or even if it persists the value of the U256.

```
let byte_array: Vec<u8> = quote_flow
    .0
    .map(|u| u.to_le_bytes().to_vec())
    .into_iter()
    .flatten()
    .collect();

let byte_obj = bytes::Bytes::from(byte_array);
```

I think the only thing that is non-intuitive is `.0`. It originates from the concept of [tuple structs](https://doc.rust-lang.org/1.9.0/book/structs.html#tuple-structs), which are like structs, but their attributes  are unnamed. Calling `.0` is similar to retrieving the first element of a tuple. In this case a [u64; 4], representing the U256.

SmartMonkey-git: Where I'm unsure is `to_le_bytes`. Could also be `to_be_bytes`.
le = little edina
be = big edina

You can check that by reconstructing the `U256`.

tamaralipows: I think I may push to save this as an int instead of bytes in the `BalanceChange.balance` attribute. However this is now out of the scope of this PR (https://datarevenue.atlassian.net/jira/software/projects/ENG/boards/24?selectedIssue=ENG-2711) so I'll close this thread. Thank you for your suggestions though! I've noted them

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -8,10 +8,10 @@ use substreams_ethereum::pb::eth::{self};
 use pb::tycho::evm::v1::{self as tycho};
 
 mod pb;
-
 const AMBIENT_CONTRACT: [u8; 20] = hex!("aaaaaaaaa24eeeb8d57d431224f73832bc34f688");
 const INIT_POOL_CODE: u8 = 71;
 const USER_CMD_FN_SIG: [u8; 4] = [0xA1, 0x51, 0x12, 0xF9];
+const SWAP_FN_SIG: [u8; 4] = [0x3d, 0x71, 0x9c, 0xd9];
```
kayibal: Can't we use the `hex!(...)` macro here instead? I find it more readable.

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -213,6 +213,78 @@ fn map_changes(
...
+                ];
+
+                let swap_external_abi_output_types = &[
+                    // The token base and quote token flows associated with this swap action.
+                    // Negative indicates a credit paid to the user (token balance of pool
+                    // decreases), positive a debit collected from the user (token balance of pool
+                    // increases).
+                    ParamType::Int(128), // baseFlow
+                    ParamType::Int(128), // quoteFlow
+                ];
```
kayibal: These seem like constants? Should we be redefining these always?

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -213,6 +239,110 @@ fn map_changes(
...
+            .collect::<Vec<_>>();
+
+        for call in ambient_hotproxy_calls {
+            if call.input.len() < 4 {
+                continue;
+            }
+            if call.input[0..4] == USER_CMD_HOTPROXY_FN_SIG {
+                // Handle TVL changes caused by calling the userCmd method on the HotProxy contract
+
+                if let Ok(external_input_params) = decode(SWAP_ABI_INPUT, &call.input[4..]) {
```
tamaralipows: I know this is repeated code - I'll move it into its own fn tomorrow

substreams/ethereum-ambient/src/lib.rs
SmartMonkey-git: I understand that i64 is much more approachable than bytes are. But given that U256 is quadruple i64's resolution, wouldn't you, at the very least, lose accuracy through that conversion?

kayibal: You can not convert a 128 bit integer into a 64 bit - it is simply not possible. Rust supports up to 128 so if this is at most 128 then it is alright and we should use u128 or i128. If this is int256 we probably need to use bytes.

substreams/ethereum-ambient/src/lib.rs
kayibal: Maybe a nicer flow would be to iterate over calls just once (without filters) and then use a match statement on `call.address` with a default that just skips? Then handle each call with a private method?

substreams/ethereum-ambient/src/lib.rs
tamaralipows: it's `i128` :)

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -393,3 +439,146 @@ fn map_changes(
...
+            bail!("Failed to decode call outputs.".to_string());
+        }
+    } else {
+        bail!("Failed to decode ABI internal call.".to_string());
+    }
+}
+
+fn decode_sweep_swap_call(
+    call: &Call,
+) -> Result<(Vec<u8>, ethabi::Int, ethabi::Int), anyhow::Error> {
```
tamaralipows, line 508: About this method, reposting a summarized conversation from slack:

There is no way to extract the pool token addresses from the transaction call itself when it comes to this sweep swap method. We only have the poolHash from the input to work with. This is made from the token addresses and the pool index, but it’s keccak encoded.

The way it’s done in the [mapping.ts](https://github.com/CrocSwap/croc-subgraph/blob/main/subgraphs/croc-mainnet/src/mapping.ts#L187) every time a pool is created, they store it on their end using GraphQL, and then query it later using this poolHash for the relevant token info.

**We have decided:** We will add this information to the [store module](https://datarevenue.atlassian.net/browse/ENG-2711), as this may be cleaner than doing these operations on the tycho side. This means that the store module will also be responsible for initializing the existing pools, in addition to updating their balances. 

For this reason, I have decided to always just return the keccak encoded pool hashes from the helper methods in the context of this PR.

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -393,3 +439,146 @@ fn map_changes(
...
+
+fn decode_direct_swap_call(
+    call: &Call,
+) -> Result<([u8; 32], ethabi::Int, ethabi::Int), anyhow::Error> {
+    if let Ok(external_input_params) = decode(SWAP_ABI_INPUT, &call.input[4..]) {
+        let base_token = external_input_params[0]
+            .to_owned()
+            .into_address()
+            .ok_or_else(|| {
+                anyhow!("Failed to convert to address: {:?}", &external_input_params[1])
```
SmartMonkey-git:
```suggestion
                anyhow!("Failed to convert to base token address: {:?}", &external_input_params[1])
```

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -393,3 +439,146 @@ fn map_changes(
...
+            })?
+            .to_fixed_bytes()
+            .to_vec();
+
+        let quote_token = external_input_params[1]
+            .to_owned()
+            .into_address()
+            .ok_or_else(|| {
+                anyhow!("Failed to convert to address: {:?}", &external_input_params[1])
```
SmartMonkey-git:
```suggestion
                anyhow!("Failed to convert to quote token address: {:?}", &external_input_params[1])
```

SmartMonkey-git: I think all of these anyhow's could state what value they were failing to convert. Makes debugging a bit easier.

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -213,6 +243,22 @@ fn map_changes(
...
+            {
+                // TODO: aggregate these with the previous balances to get new balances:
+                let (_pool_hash, _base_flow, _quote_flow) = decode_direct_swap_call(call)?;
+            } else if call.address == AMBIENT_MICROPATHS_CONTRACT &&
+                call.input[0..4] == SWEEP_SWAP_FN_SIG
+            {
+                // Handle TVL changes caused by calling the sweepSwap method on the MicroPaths
+                // contract
+                // TODO: aggregate these flows with the previous balances to get new balances:
+                let (_base_flow, _quote_flow, _pool_hash) = decode_sweep_swap_call(call)?;
```
SmartMonkey-git, line 261: Whats a sweep swap? Reading it for the first time.

tamaralipows: It's just one of three ways that tokens can be swapped on Ambient .In the typescript code, it says that this happens when the swap is "part of a long-form order"

PR #90, opened 2024-01-17
ENG-1718 and ENG-1729: Add and retrieve tokens

dianacarvalho1:
Add and retrieve tokens from the DB

Initially it was supposed to be an upsert tokens instead of add, but I couldn't find a way to batch upsert :/ Changed it to add for now

### DIFF
--- a/tycho-indexer/src/storage/postgres/protocol.rs
+++ b/tycho-indexer/src/storage/postgres/protocol.rs
@@ -10,9 +10,9 @@ use ethers::types::{transaction, Transaction};
 use tracing::warn;
 
 use crate::{
-    extractor::evm::{ERC20Token, ProtocolState, ProtocolStateDelta},
+    extractor::evm::{ERC20Token, ProtocolComponent, ProtocolState, ProtocolStateDelta},
     hex_bytes::Bytes,
-    models::{Chain, ProtocolSystem, ProtocolType},
+    models::{Chain, ProtocolType},
     storage::{
         postgres::{
             orm,
@@ -22,8 +22,8 @@ use crate::{
             PostgresGateway,
         },
         Address, BlockIdentifier, BlockOrTimestamp, ContractDelta, ContractId, ProtocolGateway,
-        StorableBlock, StorableContract, StorableProtocolState, StorableProtocolType,
-        StorableToken, StorableTransaction, StorageError,
+        StorableBlock, StorableContract, StorableProtocolComponent, StorableProtocolState,
+        StorableProtocolType, StorableToken, StorableTransaction, StorageError,
         StorageError::DecodeError,
         TxHash, Version,
     },
@@ -411,7 +411,6 @@ mod test {
     use hex::FromHex;
     use rstest::rstest;
     use serde_json::{json, Value};
-    use std::str::FromStr;
 
     use crate::{
         models,


tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -85,13 +91,49 @@ where
...
+            };
+            let account_id = diesel::insert_into(schema::account::table)
+                .values(&new_account)
+                .on_conflict((schema::account::address, schema::account::chain_id))
+                .do_update()
+                .set(&new_account)
+                .returning(schema::account::id)
+                .get_result::<i64>(conn)
+                .await
+                .map_err(|err| StorageError::from_diesel(err, "Account", &hex_addr, None))?;
```
kayibal: Got to be careful with this upsert, some pools are also tokens, so they might already exist in the account table. If they do we should not overwrite these attributes:
```
                creation_tx: None,
                created_at: None,
                deleted_at: None,
```

with `NULL`, so probably better opt for a second query in case the we get a duplicate key error, unless we can ignore updating these columns with the upsert.

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -85,13 +91,49 @@ where
...
     }
 
-    async fn add_tokens(&self, chain: Chain, token: &[&Self::Token]) -> Result<(), StorageError> {
-        todo!()
+    async fn add_tokens(
+        &self,
+        tokens: &[&Self::Token],
+        conn: &mut Self::DB,
+    ) -> Result<(), StorageError> {
+        for token in tokens {
```
kayibal: I would like to avoid doing a single query per entity. Maybe you can make this more efficient by retrieving inserting multiple values at once using batch inserts?

dianacarvalho1: I tried that 😕 and I couldn't find a way to batch upsert. Are you suggesting to retrieve and then insert the missing ones instead of upserting?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -27,7 +36,7 @@ where
     T: StorableToken<orm::Token, orm::NewToken, i64>,
 {
     type DB = AsyncPgConnection;
-    type Token = T;
+    type Token = ERC20Token;
```
dianacarvalho1: I'm not so sure about this :/

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -85,13 +94,110 @@ where
...
+                ERC20Token::from_storage(orm_token, contract_id)
+                    .map_err(|err| StorageError::DecodeError(err.to_string()))
+            })
+            .collect();
+        tokens
     }
 
-    async fn add_tokens(&self, chain: Chain, token: &[&Self::Token]) -> Result<(), StorageError> {
-        todo!()
+    async fn add_tokens(
```
dianacarvalho1, line 273: in the end I needed 3 queries for this method:
- insert new accounts
- retrieve the accounts (I couldn't find a way to return them in the previous action - the issue was to return also the ones that are already in the DB)
- insert new tokens with the correct account_id set

if anyone has any idea other idea on how to do this, then please let me know 🙏🏼 
-

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -85,13 +94,110 @@ where
...
+                    deleted_at: None,
+                }
+            })
+            .collect();
+
+        diesel::insert_into(schema::account::table)
+            .values(&new_accounts)
+            .on_conflict((schema::account::address, schema::account::chain_id))
+            .do_nothing()
+            .execute(conn)
```
SmartMonkey-git, line 310: I think you might be able to use `get_result` instead of execute. If that is possible you could omit the second query.

Check the diesel doc's [here](https://diesel.rs/guides/all-about-inserts.html) and search for `The RETURNING Clause`

dianacarvalho1: yes I tried that but it doesn't return me the conflicting accounts I would like to have as well :/ at least I couldn't find a way to do that. the `get_results` only returns the entries that were actually inserted I think. do you know how to that?

tycho-indexer/src/storage/postgres/protocol.rs
kayibal: Can't it stay T?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -188,19 +196,110 @@ where
...
+            .await
+            .map_err(|err| StorageError::from_diesel(err, "Token", &chain.to_string(), None))?;
+
+        let tokens: Result<Vec<Self::Token>, StorageError> = results
+            .into_iter()
+            .map(|(orm_token, chain_id_, address_)| {
+                let chain = self.get_chain(&chain_id_);
+                let contract_id = ContractId::new(chain, address_);
+
+                ERC20Token::from_storage(orm_token, contract_id)
```
kayibal: You should use `Self::Token` here instead of `ERC20Token`. This way we can keep the exact token type generic.

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -188,19 +196,110 @@ where
...
+        let titles: Vec<String> = tokens
+            .iter()
+            .map(|token| format!("{}_{}", token.chain, token.symbol))
+            .collect();
+
+        let new_accounts: Vec<NewAccount> = tokens
+            .iter()
+            .zip(titles.iter())
+            .map(|(token, title)| {
+                let chain_id = self.get_chain_id(&token.chain);
```
kayibal: Can't you produce the title directly here? Would save you the zip...

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -188,19 +196,110 @@ where
...
+            .get_results::<Account>(conn)
+            .await
+            .map_err(|err| StorageError::from_diesel(err, "Account", "retrieve", None))?;
+
+        let new_tokens: Vec<orm::NewToken> = tokens
+            .iter()
+            .map(|token| {
+                let account_id = accounts
+                    .iter()
+                    .find(|account| {
```
kayibal: Please use a HashSet here instead of find for performance.

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -188,19 +196,110 @@ where
...
+                    .expect("Account ID not found");
+
+                token.to_storage(account_id)
+            })
+            .collect();
+
+        diesel::insert_into(schema::token::table)
+            .values(&new_tokens)
+            .on_conflict(schema::token::account_id)
+            .do_nothing()
```
kayibal, line 344: Can you explain maybe with a comment why we need `on_conflict` here?

tycho-indexer/src/storage/postgres/protocol.rs
dianacarvalho1: no, this was a pain because of lifetimes 😩 full explanation:
Initially I had this:
```
        let new_accounts: Vec<NewAccount> = tokens
            .iter()
            .map(|token| {
                let chain_id = self.get_chain_id(&token.chain);
                NewAccount {
                    title: &format!("{}_{}", token.chain, token.symbol), <----------- problems here
                    address: token.address.as_ref(),
                    chain_id,
                    creation_tx: None,
                    created_at: None,
                    deleted_at: None,
                }
            })
            .collect();
````
when I tried to build it gave me
```
error[E0515]: cannot return value referencing temporary value
   --> tycho-indexer/src/storage/postgres/protocol.rs:143:17
    |
143 | /                 NewAccount {
144 | |                     title: &format!("{}_{}", token.chain, token.symbol),
    | |                             ------------------------------------------- temporary value created here
145 | |                     address: token.address.as_ref(),
146 | |                     chain_id,
...   |
149 | |                     deleted_at: None,
150 | |                 }
    | |_________________^ returns a value referencing data owned by the current function
````
Making the titles before hand was the best solution I could find 😕

File: tycho-indexer/src/storage/postgres/protocol.rs (hunk 3 above)
Comment (dianacarvalho1) [reply tycho-indexer/src/storage/postgres/protocol.rs]: No 😞 If I change this back I get a looooot of failing things like:

```
error[E0609]: no field `chain` on type `&&T`
   --> tycho-indexer/src/storage/postgres/protocol.rs:238:49
    |
96  | impl<B, TX, A, D, T> ProtocolGateway for PostgresGateway<B, TX, A, D, T>
    |                   - type parameter 'T' declared here
...
238 |             .map(|token| format!("{}_{}", token.chain, token.symbol))
    |                                                 ^^^^^
```
Unless if I need to change something else somewhere that it isn't clear to me 😕

File: tycho-indexer/src/storage/postgres/mod.rs [hunk 10]
```diff
@@ -874,4 +875,36 @@ pub mod db_fixtures {
...
+    }
+
+    pub async fn get_token_by_symbol(conn: &mut AsyncPgConnection, symbol: String) -> orm::Token {
+        schema::token::table
+            .filter(schema::token::symbol.eq(symbol.clone()))
+            .select(schema::token::all_columns)
+            .first::<orm::Token>(conn)
+            .await
+            .unwrap()
+    }
```
tamaralipows, line 967: Question, should we not care about the chain id here? I remember a similar problem in the defibot repo, where we would get tokens for different chains?

dianacarvalho1: oohh good catch. but this only for testing purposes anyway, so I think it's fine to keep it simple for now

tamaralipows: okk let's just keep this in mind if ever used for more

tycho-indexer/src/storage/postgres/protocol.rs
kayibal: Yeah you need to expose the chain via a method on `StorableToken` and then use `token.chain()`

tycho-indexer/src/storage/postgres/protocol.rs
kayibal: Does title need to be a reference? Or can it be String? If it has to be a reference the zip makes sense yes.

tycho-indexer/src/storage/postgres/protocol.rs
SmartMonkey-git: Oh, ok. Then I sadly, dont know how to do it efficiently. :/

tycho-indexer/src/storage/postgres/protocol.rs
dianacarvalho1: aah huum I tried that. And now I'm getting issues because of lifetimes 😕 I had to add that for chain, address and symbol. and now for example here
image
I'm getting this 😕 do I need to loop through the tokens before like I did for the titles? that seems very hacky

tycho-indexer/src/storage/postgres/protocol.rs
dianacarvalho1: not sure why it is a reference. In the commit that introduced it is just says "Avoid some unnecessary clones"

tycho-indexer/src/storage/mod.rs
```diff
@@ -642,15 +642,14 @@ pub trait ProtocolGateway {
...
     /// - `token` The tokens to insert.
     ///
     /// # Return
     /// Ok if all tokens could be inserted, Err if at least one token failed to
     /// insert.
     async fn add_tokens(
         &self,
-        chain: Chain,
+
         token: &[&Self::Token],
```
tamaralipows: One more thing, shouldn't this be called `tokens` not `token`

tycho-indexer/src/storage/postgres/protocol.rs
dianacarvalho1: I went with the loop outside of this to create the list of addresses

tycho-indexer/src/storage/mod.rs
```diff
@@ -473,6 +473,12 @@ pub trait StorableToken<S, N, I>: Sized + Send + Sync + 'static {
     fn from_storage(val: S, contract: ContractId) -> Result<Self, StorageError>;
 
     fn to_storage(&self, contract_id: I) -> N;
+
+    fn chain(&self) -> Chain;
+
+    fn address(&self) -> H160;
+
+    fn symbol(&self) -> String;
```
kayibal, line 481: OK I wonder if we still need to_storage after we have this three^^ But probably yes.

tycho-indexer/src/storage/postgres/protocol.rs
kayibal: Yeah rust is weird in that regard sometimes, the solution you came up with should work. I also really have problem with the Byte types and orm structs.

tycho-indexer/src/storage/mod.rs
dianacarvalho1: yes, I think we do

PR #91, opened 2024-01-17
Database: Remove protocol_system_type and use string instead

SmartMonkey-git:
[Slack](https://propellerswap.slack.com/archives/C05LQALE9L1/p1705498231068279)

Using `to_string` entry looks like:
image

### DIFF
--- a/tycho-indexer/src/storage/postgres/orm.rs
+++ b/tycho-indexer/src/storage/postgres/orm.rs
@@ -217,7 +217,7 @@ pub struct NewTransaction {
 #[diesel(check_for_backend(diesel::pg::Pg))]
 pub struct ProtocolSystem {
     pub id: i64,
-    pub name: ProtocolSystemType,
+    pub name: String,
     pub inserted_ts: NaiveDateTime,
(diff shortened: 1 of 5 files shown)


tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -226,21 +226,7 @@ pub struct ProtocolSystem {
...
 #[diesel(check_for_backend(diesel::pg::Pg))]
 pub struct NewProtocolSystem {
-    pub name: ProtocolSystemType,
-}
-
-#[derive(Debug, DbEnum, Clone)]
-#[ExistingTypePath = "crate::storage::postgres::schema::sql_types::ProtocolSystemType"]
-pub enum ProtocolSystemType {
-    Ambient,
-}
```
kayibal, line 236: Ok we can debate whether we want this enum to exist or not - it certainly can while on the DB side this is modelled as a string. But actually - I think I might prefer to not keep it as an enum given that this way we can just add new protocols without needing to recompile Tycho!

The downside is ofc that we can't assume anything about which protocols we are dealing with here.

PR #92, opened 2024-01-17
ENG-2031: Implement insert ProtocolComponent

SmartMonkey-git:
[Ticket](https://datarevenue.atlassian.net/browse/ENG-2031)

This PR does multiple things besides achieving its initial task.
#### ProtocolComponent leftovers
- Completes `orm::ProtocolComponent` attributes, so it is possible to retrieve rows
- Elevate `created_tx` and `created_ts` through code base
#### Cache ProtocolSystems as response to [this](https://github.com/propeller-heads/tycho-indexer/pull/92#discussion_r1456446061)
- Adds an EnumCache for the `ProtocolSystem` enum. So, time can be saved by retrieving from the cache and not from the db. ProtocolComponent need the System ID.
- Ensures the cache is filled on start
- Augment with_cache function to add cache
#### Get Transaction ids
- Add `db_id_by_hash` to orm::Transaction for retrieving ids

Besides that, it of course, implements an `add_protocol_components` to add protocol_components. Not an upsert, because it did not seem possible to implement that for a batch operation.

### DIFF
--- a/tycho-indexer/src/main.rs
+++ b/tycho-indexer/src/main.rs
@@ -105,7 +105,8 @@ async fn main() -> Result<(), ExtractionError> {
 
     let pool = postgres::connect(&args.database_url).await?;
     postgres::ensure_chains(&[Chain::Ethereum], pool.clone()).await;
-    postgres::ensure_protocol_systems(pool.clone()).await;
+    // TODO: Find a dynamic way to load protocol systems into the application.
+    postgres::ensure_protocol_systems(&["ambient".to_string()], pool.clone()).await;
     let evm_gw = PostgresGateway::<
         evm::Block,
         evm::Transaction,
--- a/tycho-indexer/src/storage/mod.rs
+++ b/tycho-indexer/src/storage/mod.rs
@@ -81,7 +81,7 @@ use thiserror::Error;
 use crate::{
     extractor::evm::{ProtocolComponent, ProtocolState, ProtocolStateDelta},
     hex_bytes::Bytes,
-    models::{Chain, ExtractionState, ProtocolSystem, ProtocolType},
+    models::{Chain, ExtractionState, ProtocolType},
     storage::postgres::orm,
 };
 
@@ -557,7 +557,7 @@ pub trait ProtocolGateway {
     async fn get_protocol_components(
         &self,
         chain: &Chain,
-        system: Option<ProtocolSystem>,
+        system: Option<String>,
         ids: Option<&[&str]>,
     ) -> Result<Vec<Self::ProtocolComponent>, StorageError>;
 
@@ -617,7 +617,7 @@ pub trait ProtocolGateway {
         &self,
         chain: &Chain,
         at: Option<Version>,
-        system: Option<ProtocolSystem>,
+        system: Option<String>,
         id: Option<&[&str]>,
         conn: &mut Self::DB,
     ) -> Result<Vec<ProtocolState>, StorageError>;
@@ -679,7 +679,7 @@ pub trait ProtocolGateway {
     async fn get_state_delta(
         &self,
         chain: &Chain,
-        system: Option<ProtocolSystem>,
+        system: Option<String>,
         id: Option<&[&str]>,
         start_version: Option<&BlockOrTimestamp>,
         end_version: &BlockOrTimestamp,
@@ -703,7 +703,7 @@ pub trait ProtocolGateway {
 
     async fn _get_or_create_protocol_system_id(
         &self,
-        protocol_system: ProtocolSystem,
+        protocol_system: String,
         conn: &mut Self::DB,
     ) -> Result<i64, StorageError>;
 }
@@ -781,7 +781,7 @@ pub trait StorableProtocolComponent<S, N, I>: Sized + Send + Sync + 'static {
         tokens: Vec<H160>,
         contract_ids: Vec<H160>,
         chain: Chain,
-        protocol_system: ProtocolSystem,
+        protocol_system: String,
         transaction_hash: H256,
     ) -> Result<Self, StorageError>;
 
--- a/tycho-indexer/src/storage/postgres/mod.rs
+++ b/tycho-indexer/src/storage/postgres/mod.rs
@@ -11,8 +11,7 @@
 //! This decision stems from an understanding that while extending the Rust
 //! codebase to include more enums is a straightforward task, modifying the type
 //! of a SQL column can be an intricate process. By representing enums as
-//! tables, we circumvent unnecessary migrations when modifying Chain or
-//! ProtocolSystem enums.
+//! tables, we circumvent unnecessary migrations when modifying, for example the Chain enum.
 //!
 //! With this representation, it's important to synchronize them whenever the
 //! enums members changed. This can be done automatically once at system
@@ -139,10 +138,9 @@ use diesel_async::{
     AsyncPgConnection, RunQueryDsl,
 };
 use diesel_migrations::{embed_migrations, EmbeddedMigrations, MigrationHarness};
-use strum::IntoEnumIterator;
 use tracing::{debug, info};
 
-use crate::models::{Chain, ProtocolSystem};
+use crate::models::Chain;
 
 use super::{
     BlockIdentifier, BlockOrTimestamp, ContractDelta, StateGateway, StorableBlock,
@@ -159,7 +157,7 @@ pub mod schema;
 
 const MIGRATIONS: EmbeddedMigrations = embed_migrations!("./migrations/");
 
-pub struct EnumTableCache<E> {
+pub struct ValueIdTableCache<E> {
     map_id: HashMap<E, i64>,
     map_enum: HashMap<i64, E>,
 }
@@ -167,9 +165,9 @@ pub struct EnumTableCache<E> {
 /// Provides caching for enum and its database ID relationships.
 ///
 /// Uses a double sided hash map to provide quick lookups in both directions.
-impl<'a, E> EnumTableCache<E>
+impl<'a, E> ValueIdTableCache<E>
 where
-    E: Eq + Hash + Copy + FromStr + std::fmt::Debug,
+    E: Eq + Hash + Clone + FromStr + std::fmt::Debug,
     <E as FromStr>::Err: std::fmt::Debug,
 {
     pub async fn from_pool(pool: Pool<AsyncPgConnection>) -> Result<Self, StorageError> {
@@ -199,7 +197,7 @@ where
         let mut cache = Self { map_id: HashMap::new(), map_enum: HashMap::new() };
         for (id_, name_) in entries {
             let val = E::from_str(&name_).expect("valid enum value");
-            cache.map_id.insert(val, id_);
+            cache.map_id.insert(val.clone(), id_);
             cache.map_enum.insert(id_, val);
         }
         cache
@@ -223,18 +221,21 @@ where
     /// # Arguments
     ///
     /// * `id` - The database ID to lookup.
-    fn get_enum(&self, id: &i64) -> E {
-        *self
-            .map_enum
+    fn get_value(&self, id: &i64) -> E {
+        self.map_enum
             .get(id)
             .unwrap_or_else(|| {
                 panic!("Unexpected cache miss for id {}, entries: {:?}", id, self.map_enum)
             })
+            .to_owned()
     }
 }
 
-type ChainEnumCache = EnumTableCache<Chain>;
-type ProtocolSystemEnumCache = EnumTableCache<ProtocolSystem>;
+type ChainEnumCache = ValueIdTableCache<Chain>;
+/// ProtocolSystem is not handled as an Enum, because that would require us to restart the whole
+/// application every time we want to add another System. Hence, to diverge from the implementation
+/// of the Chain enum was a conscious decision.
+type ProtocolSystemEnumCache = ValueIdTableCache<String>;
 
 impl From<diesel::result::Error> for StorageError {
     fn from(value: diesel::result::Error) -> Self {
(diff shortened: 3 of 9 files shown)


tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -30,21 +31,56 @@ where
...
+    ) -> Result<(), StorageError> {
+        use super::schema::protocol_component::dsl::*;
+        let mut values: Vec<orm::NewProtocolComponent> = vec![];
+        //let values: Vec<NewProtocolComponent> = new
+        //    .into_iter()
+        //    .map(|pc| async {
+        //        pc.to_storage(
+        //            self.get_chain_id(&pc.chain),
+        //            self._get_or_create_protocol_system_id(pc.protocol_system, conn)
+        //                .await?,
```
kayibal: You probably noticed that async in map doesn't work, but more importantly in terms of strategy: 
- Please retrieve only - if a component has a protocol system assigned that we don't know then you should error. We should not be creating systems here.
- It probably makes sense to have a method that retrieves all protocol system ids directly, then cast that to a hash map from name -> id, this will help you in the loop below to correctly call to_storage on each protocol components.

SmartMonkey-git: Added a cache [here](https://github.com/propeller-heads/tycho-indexer/pull/92/commits/b7253a345789e429f14c1632f19ebca704375110)

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -530,4 +572,66 @@ mod test {
...
+
+        let result = gw
+            .add_protocol_components(&[&original_component.clone()], &mut conn)
+            .await;
+
+        assert!(result.is_ok());
+
+        use crate::storage::postgres::schema::protocol_component::dsl::*;
+        let inserted_data = protocol_component
+            .filter(external_id.eq("test_contract_id"))
```
louise-poole: I think this should fix the failing CI:
```suggestion
            .filter(external_id.eq("test_contract_id".to_string()))
```

SmartMonkey-git: Sadly, thats not it.

tycho-indexer/src/storage/mod.rs
```diff
@@ -339,6 +339,12 @@ pub trait ChainGateway {
         hash: &TxHash,
         db: &mut Self::DB,
     ) -> Result<Self::Transaction, StorageError>;
+
+    async fn _get_tx_ids(
+        &self,
+        hashes: &[TxHash],
+        conn: &mut Self::DB,
+    ) -> Result<HashMap<TxHash, i64>, StorageError>;
```
kayibal: Is this a private method on a trait?? 

Please move this to `orm::Transaction::by_hash()` or `orm::Transaction::db_id_by_hash()`

tycho-indexer/src/storage/postgres/mod.rs
```diff
@@ -324,8 +327,12 @@ where
     A: StorableContract<orm::Contract, orm::NewContract, i64>,
     T: StorableToken<orm::Token, orm::NewToken, i64>,
 {
-    pub fn with_cache(cache: Arc<ChainEnumCache>) -> Self {
+    pub fn with_cache(
+        cache: Arc<ChainEnumCache>,
+        protocol_system_cache: Arc<ProtocolSystemEnumCache>,
+    ) -> Self {
         Self {
+            protocol_system_id_cache: protocol_system_cache,
```
kayibal, line 336: Just realised that this might be problematic. If we add a new extractor on startup, the extractor will register the new protocol system - to do this it will need this gateway. So if we want to keep the cache here we'd need a way to refresh it and do that if a new protocol system is added.

Not saying to change it - maybe we end up registering these protocol systems differently. Just want to raise awareness.

tycho-indexer/src/storage/postgres/mod.rs
```diff
@@ -347,22 +354,49 @@ where
...
-        let cache = Arc::new(ChainEnumCache::from_tuples(results));
-        Self::with_cache(cache)
+
+        let protocol_system_id_mapping: Vec<(i64, String)> = async {
+            use schema::protocol_system::dsl::*;
+            protocol_system
+                .select((id, name))
+                .load(conn)
+                .await
+                .expect("Failed to load chain ids!")
```
kayibal:
```suggestion
                .expect("Failed to load protocol systems!")
```

tycho-indexer/src/storage/postgres/mod.rs
```diff
@@ -447,6 +481,27 @@ pub async fn ensure_chains(chains: &[Chain], pool: Pool<AsyncPgConnection>) {
     debug!("Ensured chain enum presence for: {:?}", chains);
 }
 
+pub async fn ensure_protocol_systems(pool: Pool<AsyncPgConnection>) {
+    let mut conn = pool.get().await.expect("connection ok");
+    let protocol_system_strings: Vec<String> = ProtocolSystem::iter()
```
kayibal: Oh when did this become an enum again? I thought it will remain a string.... :/

https://github.com/propeller-heads/tycho-indexer/pull/91#discussion_r1456157103

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -98,21 +94,61 @@ where
...
+        let tx_hash_id_mapping: HashMap<TxHash, i64> = self
+            ._get_tx_ids(&tx_hashes, conn)
+            .await
+            .unwrap();
+
+        for pc in new {
+            let tx_hash = match tx_hash_id_mapping.get::<TxHash>(&pc.creation_tx.into()) {
+                Some(hash) => hash.to_owned(),
+                None => return Err(StorageError::DecodeError("TxHash not found".to_string())),
+            };
```
kayibal: Nit: probably don't need a match here. I am surprised clippy does not complain:
```
let txh = tx_hash_id_mapping.get().ok_or_else(StorgeError::...)?
```

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -98,21 +94,61 @@ where
...
+        todo!()
+    }
+
+    async fn add_protocol_components(
+        &self,
+        new: &[&Self::ProtocolComponent],
+        conn: &mut Self::DB,
+    ) -> Result<(), StorageError> {
+        use super::schema::protocol_component::dsl::*;
+        let mut values: Vec<orm::NewProtocolComponent> = vec![];
```
kayibal:
```suggestion
        let mut values: Vec<orm::NewProtocolComponent> = Vec::with_capacity(new.length);
```

tycho-indexer/src/storage/mod.rs
SmartMonkey-git: FYI.
There is another private method in the trait `ProtocolGateway._get_or_create_protocol_system_id`

But i moved this one, as you asked, because it makes more sense. Do you want me to move this one as well?

tycho-indexer/src/storage/postgres/mod.rs
SmartMonkey-git: It follows the same pattern as Chain.
It is an enum in `tycho-indexer/src/models.rs`, but in the database it is a varchar.

tycho-indexer/src/main.rs
```diff
@@ -105,6 +105,7 @@ async fn main() -> Result<(), ExtractionError> {
 
     let pool = postgres::connect(&args.database_url).await?;
     postgres::ensure_chains(&[Chain::Ethereum], pool.clone()).await;
+    postgres::ensure_protocol_systems(pool.clone()).await;
```
dianacarvalho1: what is `pool`? 🤔

tycho-indexer/src/storage/mod.rs
```diff
@@ -541,6 +541,11 @@ pub trait ProtocolGateway {
 
     // TODO: uncomment below when StorableProtocolComponent is implemented (ENG 1728)
     // type ProtocolComponent;
```
dianacarvalho1: comment no longer needed

tycho-indexer/src/storage/postgres/mod.rs
```diff
@@ -222,7 +223,7 @@ where
     /// # Arguments
     ///
     /// * `id` - The database ID to lookup.
-    fn get_chain(&self, id: &i64) -> E {
+    fn get_enum(&self, id: &i64) -> E {
```
dianacarvalho1: I actually liked `get_chain` better. `get_enum`? which `enum`?

tycho-indexer/src/main.rs
dianacarvalho1: oh it's a `Pool<AsyncPgConnection>`, that is confusing.. considering what pool normally means for us

tycho-indexer/src/storage/postgres/mod.rs
```diff
@@ -447,6 +481,27 @@ pub async fn ensure_chains(chains: &[Chain], pool: Pool<AsyncPgConnection>) {
...
+        .collect();
+
+    diesel::insert_into(schema::protocol_system::table)
+        .values(
+            protocol_system_strings
+                .iter()
+                .map(|ps| schema::protocol_system::name.eq(ps))
+                .collect::<Vec<_>>(),
+        )
+        .on_conflict_do_nothing()
```
dianacarvalho1, line 495: oh I didn't know there was a one liner for this, cool cool. noted

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -199,6 +203,21 @@ impl Transaction {
             .first::<Self>(conn)
             .await
     }
+
+    pub async fn db_id_by_hash(
```
dianacarvalho1: you think it's necessary to call it `db_id`? won't `id_by_hash` be enough?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -98,21 +94,60 @@ where
...
-    //     ids: Option<&[&str]>,
-    // ) -> Result<Vec<ProtocolComponent>, StorageError> {
-    //     todo!()
-    // }
-
-    // TODO: uncomment to implement in ENG 2049
-    // async fn upsert_components(&self, new: &[&ProtocolComponent]) -> Result<(), StorageError> {
-    //     todo!()
-    // }
+    async fn get_components(
```
dianacarvalho1: can we call this `get_protocol_components`? the other one is `add_protocol_components` and not `add_components`

tycho-indexer/src/storage/postgres/mod.rs
dianacarvalho1: but actually.. on conflict of what?

tycho-indexer/src/storage/postgres/mod.rs
SmartMonkey-git: Its a function of a class called `EnumTableCache`. There is a more specific type called `ChainEnumCache` that "derives" from `EnumTableCache`, but the later is general and not specific to the chain enum.
That is why I renamed it.

tycho-indexer/src/storage/postgres/mod.rs
SmartMonkey-git: The DB is persistent. The state of the application is not. If you restart the application this code will be called again, while the protocol systems are still in the table.
That is one scenario for a conflict.

tycho-indexer/src/storage/postgres/mod.rs
kayibal: Yes yes but I changed my opinion on the last PR, I think it should be string also application side - sorry. This will avoid us having to make a new release for each new protocol. So the protocol will be like plug ins will save us a lot of work to do this now. Could you change it back please?

Once again sorry for changing my opinion so last minute here.

tycho-indexer/src/main.rs
kayibal: If seen in IDE, you'll have the types which give you more context.

tycho-indexer/src/storage/postgres/mod.rs
tamaralipows: So we are planning use this for other applications other than the chain?

SmartMonkey-git: Right now its in use for ProtocolSystems and Chains

tycho-indexer/src/storage/postgres/mod.rs
SmartMonkey-git: https://github.com/propeller-heads/tycho-indexer/pull/95

tycho-indexer/src/storage/postgres/mod.rs
dianacarvalho1: ok. I withdraw my comment

tycho-indexer/src/storage/postgres/mod.rs
dianacarvalho1: yes but then the conflict will be if there is anything in the DB at all and it will skip the insert? I'm confused. or will it check the current entries vs the new ones? (if yes, then what does it check for the conflict?)

SmartMonkey-git: Good question. 
I doubted it will skip the whole insert, when there is one conflict. I would say it checks per entry if it already exists. It does so for every value on each column and rejects when all columns conflict.
At least, that is how I interpret the [diesel documentation](https://docs.rs/diesel/latest/diesel/query_builder/struct.InsertStatement.html#examples-1), but it does not make a concrete statement about that. 
Will investigate further, have not checked the postgres docu, yet.

SmartMonkey-git: Here is a little explanation for you.
https://www.postgresql.org/docs/current/sql-insert.html#:~:text=ON%20CONFLICT%20DO%20NOTHING%20simply,can%20perform%20unique%20index%20inference.

You've made a great discovery here. The name columns was missing a unique constraint, which caused the table to accept duplicate ProtocolSystems.

image

SmartMonkey-git: Fixed it [here](https://github.com/propeller-heads/tycho-indexer/pull/92/commits/1aac058090f63f501443e6f496bdc243605a4748)

PR #93, opened 2024-01-18
Separate ProtocolState and ProtocolStateUpdate structs

louise-poole:
In response to the post-merge comments on this PR: https://github.com/propeller-heads/tycho-indexer/pull/74
In terms of the gateway this means:
- `get_states` will return whole `ProtocolState`s
- `get_state_delta` will return a `ProtocolStateUpdate`
- `update_state` expects a `ProtocolStateUpdate` argument

Some of the necessary logic updates here are already somewhat handled in [this](https://github.com/propeller-heads/tycho-indexer/pull/75) PR so I have left them as 'TODOs' here.


tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -728,16 +742,16 @@ impl ProtocolState {
     /// # Errors
     /// This method will return `ExtractionError::MergeError` if any of the above
     /// conditions is violated.
-    pub fn merge(&mut self, other: ProtocolState) -> Result<(), ExtractionError> {
+    pub fn merge(&mut self, other: ProtocolStateUpdate) -> Result<(), ExtractionError> {
```
kayibal: The order of merges is quite important, I guess there is no way we can double check this order here?

kayibal: In that case maybe adding it to the docs would be a good idea.

tycho-indexer/src/storage/mod.rs
```diff
@@ -579,7 +606,12 @@ pub trait ProtocolGateway {
         id: Option<&[&str]>,
     ) -> Result<Vec<ProtocolState>, StorageError>;
 
-    async fn update_state(&self, chain: Chain, new: &[(TxHash, ProtocolState)], db: &mut Self::DB);
+    async fn update_state(
+        &self,
+        chain: Chain,
+        new: &[(TxHash, ProtocolStateUpdate)],
```
kayibal: Should we call it `Update` or `Delta`? Should we be consistent (even if we might not have been previously)?

louise-poole: I followed the lead of how it's done for the contract gateway. They have an Account and an AccountUpdate there and ContractDelta for the 'Storable' trait for it. I'm happy to make it more consistent though - which do you prefer?

tycho-indexer/src/extractor/evm/mod.rs
louise-poole: I do have this line in the doc string: 
```
 /// The merged update keeps the transaction of `other`.
```
Though I can definitely make it more explicit. Note, this merge is designed to be called from within the `ProtocolStateUpdatesWithTx`'s merge. There the order is checked and an error thrown if incorrect.

tycho-indexer/src/storage/mod.rs
kayibal: I prefer `Delta` I think because `Update` kind of has a forward connotation for me.

PR #94, opened 2024-01-18
ENG-2042: Emit TVL updates on all liquidity change paths.

tamaralipows:
**Goal:** Emit TVL updates in substreams
I’ve looked at how it’s done in [mapping.ts](https://github.com/CrocSwap/croc-subgraph/blob/main/subgraphs/croc-mainnet/src/mapping.ts#L187) and found the following info:

- In the scope of this task, we care about the `modifyLiquidity` method - this method is called the following situations:

1. `handleWarmPathCall` - Handler for a userCmd() call made to the WarmPath contract.
2.  `handleMicroPathsLiquidity` called the following times:
   a. `handleMintRangeCall` - Handler for a MicroPaths mintRange() call (as part of a long-form order) that mints a concentrated liquidity position
   b. `handleMintRangeEvent` - event CrocMicroMintRange(bytes input, bytes output); **[not emitted on mainnet]**
   c. `handleMintAmbientCall` -  a MicroPaths mintAmbient() call (as part of a long-form order) that mints an ambient liquidity position
   d. `handleMintAmbientEvent` - event CrocMicroMintAmbient(bytes input, bytes output); **[not emitted on mainnet]**
   e. `handleBurnRangeCall` - Handler for a MicroPaths mintRange() call (as part of a long-form order) that burns a concentrated liquidity position
   f. `handleBurnRangeEvent` - event CrocMicroBurnRange(bytes input, bytes output); **[not emitted on mainnet]**
   g. `handleBurnAmbientCall` -  Handler for a MicroPaths burnAmbient() call (as part of a long-form order) that burns an ambient liquidity position
   h. `handleBurnAmbientEvent` - event CrocMicroBurnAmbient(bytes input, bytes output); **[not emitted on mainnet]**
3. `handleKnockoutCmd` -  Handler for  KnockoutLiqPath knockoutCmd() calls
4. `handleKnockoutCross` - Handler for a CrocKnockoutCross event emission. This is only emitted when the sweepSwap method is called on the MicroPaths contract. We have already handled this case.

So all in all I am left with these SIX cases (which still need to be tested):
- [x]  `handleWarmPathCall` -> tested
- [x] `handleMintRangeCall` -> tested
- [x] `handleMintAmbientCall`  -> tested
- [x] `handleBurnRangeCall` -> tested
- [x] `handleBurnAmbientCall` -> untestable? Ran on 700k blocks and nothing found.
- [x] `handleKnockoutCmd` -> tested

Also in this PR:

- [x]  Fix overflow error when getting pool index


substreams/ethereum-ambient/src/lib.rs
```diff
@@ -43,6 +54,131 @@ const SWAP_ABI_OUTPUT: &[ParamType] = &[
...
+    ParamType::Uint(256), // poolIdx
+    ParamType::Int(24),
+    ParamType::Int(24),
+    ParamType::Bool,
+    ParamType::Uint(8),
+    ParamType::Uint(256),
+    ParamType::Uint(256),
+    ParamType::Uint(32),
+];
+
```
kayibal: Could you split all of these into a separate abi module in a separate file please?

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -258,7 +393,53 @@ fn map_changes(
                 // Handle TVL changes caused by calling the sweepSwap method on the MicroPaths
                 // contract
                 // TODO: aggregate these flows with the previous balances to get new balances:
-                let (_base_flow, _quote_flow, _pool_hash) = decode_sweep_swap_call(call)?;
+                let (_pool_hash, _base_flow, _quote_flow) = decode_sweep_swap_call(call)?;
+            } else if call.address == AMBIENT_WARMPATH_CONTRACT &&
+                call.input[0..4] == USER_CMD_WARMPATH_FN_SIG
```
kayibal: I think it would be nicer if you also made this a bit more structured:

One file per contract, then one decoding function per method. Import these here, then have a match on the contract address and the put the decoding logic for that contract into the match block (could be another match on `call.input[0..4] `, btw please put this into a variable and call it `selector`.). The return value of the entire match block, so it seems can be `(pool, base_flow, quote_flow)`.

I think this will make things a bit nicer to read because you can focus on one contract at a time.

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -258,7 +393,53 @@ fn map_changes(
...
+            {
+                // Handle TVL changes on burnAmbient() calls to the MicroPaths contract
+                // TODO: aggregate these flows with the previous balances to get new balances:
+                let (_pool_hash, _base_flow, _quote_flow) = decode_burn_ambient_call(call)?;
+            } else if call.address == AMBIENT_KNOCKOUT_CONTRACT &&
+                call.input[0..4] == USER_CMD_KNOCKOUT_FN_SIG
+            {
+                // Handle TVL changes on userCmd() calls to the KnockoutLiqPath contract
+                // TODO: aggregate these flows with the previous balances to get new balances:
+                let (_pool_hash, _base_flow, _quote_flow) = decode_knockout_call(call)?;
```
kayibal: Why do you underscore these varaible names?

tamaralipows: I will get a lint error if I don't, because they are currently unused.

substreams/ethereum-ambient/src/contracts/mod.rs
```diff
@@ -0,0 +1,16 @@
+// @generated
```
dianacarvalho1, line 1: how is this generated??

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -145,121 +126,51 @@ fn map_changes(
...
-                // contract
-                // TODO: aggregate these flows with the previous balances to get new balances:
-                let (_base_flow, _quote_flow, _pool_hash) = decode_sweep_swap_call(call)?;
             }
+
+            // Extract TVL changes
+            let _result = match (address, selector) {
+                (AMBIENT_CONTRACT, SWAP_FN_SIG) => Some(decode_direct_swap_call(call)?),
+                (AMBIENT_HOTPROXY_CONTRACT, USER_CMD_HOTPROXY_FN_SIG) => {
+                    Some(decode_direct_swap_hotproxy_call(call)?)
```
dianacarvalho1, line 146: before for this case we were calling `decode_direct_swap_call` instead

tamaralipows: Yeah they are decoded in identical ways basically. But I wanted to separate them for the sake of consistency, organization, and error logging.

substreams/ethereum-ambient/src/contracts/mod.rs
tamaralipows: At first I copied this from the other file, and then it just started auto updating at every save or run or compile, not really sure sorry...

tamaralipows: (It's entirely possible I did this wrong)

dianacarvalho1: magic

PR #95, opened 2024-01-22
Remove protocol system enum

SmartMonkey-git:
This PR has its origin in this [comment](https://github.com/propeller-heads/tycho-indexer/pull/92#discussion_r1458897991)

The merge will be postponed until the [base branch](https://github.com/propeller-heads/tycho-indexer/pull/92) is ready to merge to dodge review redundancies.

### DIFF
--- a/tycho-indexer/src/storage/postgres/mod.rs
+++ b/tycho-indexer/src/storage/postgres/mod.rs
@@ -232,6 +232,9 @@ where
 }
 
 type ChainEnumCache = ValueIdTableCache<Chain>;
+/// ProtocolSystem is not handled as an Enum, because that would require us to restart the whole
+/// application every time we want to add another System. Hence, to diverge from the implementation
+/// of the Chain enum was a conscious decision.
 type ProtocolSystemEnumCache = ValueIdTableCache<String>;
 
 impl From<diesel::result::Error> for StorageError {


tycho-indexer/src/main.rs
```diff
@@ -105,7 +105,8 @@ async fn main() -> Result<(), ExtractionError> {
 
     let pool = postgres::connect(&args.database_url).await?;
     postgres::ensure_chains(&[Chain::Ethereum], pool.clone()).await;
-    postgres::ensure_protocol_systems(pool.clone()).await;
+    // TODO: Find a dynamic way to load protocol systems into the application.
```
dianacarvalho1, line 108: should we have a config where we define the protocol systems to be used?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -606,7 +606,7 @@ impl BlockContractChanges {
                         let component = ProtocolComponent::try_from_message(
                             component_msg,
                             chain,
-                            protocol_system,
+                            protocol_system.clone(),
```
kayibal: I think it would be better if the interface accepted &str and the actual method clones if needed (which it likely does here).

tycho-indexer/src/main.rs
kayibal: Yeah we'll have to add a config that will specify all the extractors to instantiate, each extractor will belongs to a protocol system.

But that is out of scope here.

tycho-indexer/src/storage/postgres/mod.rs
```diff
@@ -224,17 +222,17 @@ where
...
             .unwrap_or_else(|| {
                 panic!("Unexpected cache miss for id {}, entries: {:?}", id, self.map_enum)
             })
+            .to_owned()
     }
 }
 
 type ChainEnumCache = EnumTableCache<Chain>;
-type ProtocolSystemEnumCache = EnumTableCache<ProtocolSystem>;
+type ProtocolSystemEnumCache = EnumTableCache<String>;
```
kayibal: Ok I understand why we do this and it makes sense, but putting string here is super weird, because String is not an enum. Please add a comment here documenting why we do this and our design decisions about treating ProtocolSystems also application side as Strings.
Also should we consider renaming the `EnumTableCache` to a `ValueIdTableCache` or so?

PR #96, opened 2024-01-22
ENG-2030 Retrieving  ProtocolComponents

SmartMonkey-git:
[Task](https://datarevenue.atlassian.net/browse/ENG-2030)
- Adds a function to get ProtocolComponents from the Database. 
- Additionally, add a little function to get transactions hash by id.
- Interface change of planned get_protocol_components function - makes ProtocolSystem mandatory, because the from_storage function demands it.

### DIFF
--- a/tycho-indexer/src/storage/postgres/protocol.rs
+++ b/tycho-indexer/src/storage/postgres/protocol.rs
@@ -1,11 +1,11 @@
 #![allow(unused_variables)]
 
 use async_trait::async_trait;
-use std::{collections::HashMap, hash::Hash};
+use chrono::NaiveDateTime;
+use std::{cmp::Ordering, collections::HashMap};
 
 use diesel::prelude::*;
 use diesel_async::{AsyncPgConnection, RunQueryDsl};
-use ethers::types::{transaction, Transaction};
 use tracing::warn;
 
 use crate::{
@@ -639,6 +639,7 @@ mod test {
             "0x3108322284d0a89a7accb288d1a94384d499504fe7e04441b0706c7628dee7b7".to_string(),
             "0x50449de1973d86f21bfafa7c72011854a7e33a226709dc3e2e4edcca34188388".to_string(),
         ];
+
         let txn = db_fixtures::insert_txns(
             conn,
             &[
@@ -651,10 +652,12 @@ mod test {
             ],
         )
         .await;
+
         let protocol_system_id_ambient =
             db_fixtures::insert_protocol_system(conn, "ambient".to_owned()).await;
         let protocol_system_id_zz =
             db_fixtures::insert_protocol_system(conn, "zigzag".to_owned()).await;
+
         let protocol_type_id = db_fixtures::insert_protocol_type(
             conn,
             "Pool",
@@ -674,14 +677,13 @@ mod test {
         .await;
         let protocol_component_id2 = db_fixtures::insert_protocol_component(
             conn,
-            "state2",
+            "state3",
             chain_id,
-            protocol_system_id,
+            protocol_system_id_ambient,
             protocol_type_id,
             txn[0],
         )
         .await;
-
         db_fixtures::insert_protocol_component(
             conn,
             "state2",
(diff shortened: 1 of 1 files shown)


tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -99,10 +100,67 @@ where
...
+                protocol_component
+                    .filter(
+                        chain_id
+                            .eq(chain_id_value)
+                            .and(protocol_system_id.eq(protocol_system)),
+                    )
+                    .select(orm::ProtocolComponent::as_select())
+                    .load::<orm::ProtocolComponent>(conn)
+                    .await?
+            }
```
dianacarvalho1: I think this can be simplified.. like:
```
query = protocol_component
                    .filter(
                        chain_id
                            .eq(chain_id_value)
                            .and(protocol_system_id.eq(protocol_system)),
                    )
                    .select(orm::ProtocolComponent::as_select();

if let Some(ids) = addresses {
    query = query.filter(external_id.eq_any(ids));
}
// execute it
```

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -642,4 +710,23 @@ mod test {
...
+        let system = "ambient".to_string();
+        let ids = Some(["state1", "state2"].as_slice());
+
+        let result = gw
+            .get_protocol_components(&chain, system, ids, &mut conn)
+            .await;
+
+        assert!(result.is_ok());
+        let components = result.unwrap();
+        assert_eq!(components.len(), 2);
```
dianacarvalho1: can you add some more asserts to make sure everything is ok? like that the objects were cast correctly and etc

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -99,10 +100,67 @@ where
...
+                    .select(orm::ProtocolComponent::as_select())
+                    .load::<orm::ProtocolComponent>(conn)
+                    .await?
+            }
+        };
+        let tx_ids: Vec<i64> = orm_protocol_components
+            .iter()
+            .map(|pc| pc.creation_tx)
+            .collect();
+        let tx_hashes = orm::Transaction::hash_by_id(tx_ids.as_slice(), conn).await?;
```
dianacarvalho1: why not make this with a join?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -99,10 +100,45 @@ where
     async fn get_protocol_components(
         &self,
         chain: &Chain,
-        system: Option<String>,
+        system: String,
```
kayibal: Why did you decide to remove Option here?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -99,10 +100,45 @@ where
...
+            .load::<(orm::ProtocolComponent, TxHash)>(conn)
+            .await?;
+
+        orm_protocol_components
+            .into_iter()
+            .map(|(pc, tx_hash)| {
+                ProtocolComponent::from_storage(
+                    pc,
+                    vec![],
+                    vec![],
```
kayibal, line 160: Add a TODO here so we don't forget once tokens are ready? Shouldn't at least the latter (the related contracts) already be available?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -341,33 +377,23 @@ mod test {
...
-                    1i64,
-                    "0xbb7e16d797a9e2fbc537e30f91ed3d27a254dd9578aa4c3af3e5f0d3e8130945",
-                ),
-                (
-                    blk[0],
-                    2i64,
-                    "0x794f7df7a3fe973f1583fbb92536f9a8def3a89902439289315326c04068de54",
-                ),
+                (blk[0], 1i64, &tx_hashes[0]),
+                (blk[0], 2i64, &tx_hashes[1]),
```
kayibal, line 647: I find it confusing that you use tx hashes as block hashes...

SmartMonkey-git: This uses the tx_hashes to create transactions.

tycho-indexer/src/storage/postgres/protocol.rs
SmartMonkey-git: Will be done in the next PR!

PR #97, opened 2024-01-22
ENG 2438: implement ProtocolGateway get_protocol_state_deltas

louise-poole:
Also optimised `get_protocol_states` and helper functions.

To summarise: we need to support getting protocol state deltas in both the forward and reverse directions. The reverse direction is typically used for reverts.

FORWARDS: 
We collect and aggregate all state updates applied during the provided timeframe. We also collect all attributes that were deleted during that time (their last entry before the end of the timeframe has a 'valid_to' value (they are now outdated) and there is no new entry for that attribute).

BACKWARDS (reverts):
We collect all state updates applied during the provided timeframe and for each component attribute that changed, we collect their previous state and return that. If their previous state was "None", this indicates that this attribute didn't exist before that entry and we list that attribute as a deleted attribute. We also collect all attributes that were deleted during that timeframe so that they can be reinstated at with their value at the start of the timeframe.

### DIFF
--- a/tycho-indexer/src/storage/postgres/orm.rs
+++ b/tycho-indexer/src/storage/postgres/orm.rs
@@ -682,6 +682,7 @@ impl ProtocolState {
         conn: &mut AsyncPgConnection,
     ) -> QueryResult<Vec<(ComponentId, AttrStoreKey)>> {
         // subquery to exclude entities that have a valid version at end_ts (haven't been deleted)
+        // TODO: use parameter binding instead of string interpolation
         let sub_query = format!("NOT EXISTS (
                                 SELECT 1 FROM protocol_state ps2
                                 WHERE ps2.protocol_component_id = protocol_state.protocol_component_id
@@ -755,6 +756,7 @@ impl ProtocolState {
             .distinct_on((protocol_state::protocol_component_id, protocol_state::attribute_name));
 
         // subquery to exclude entities that have a valid version at start_ts (weren't deleted)
+        // TODO: use parameter binding instead of string interpolation
         let sub_query = format!("NOT EXISTS (
                                 SELECT 1 FROM protocol_state ps2
                                 WHERE ps2.protocol_component_id = protocol_state.protocol_component_id


tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -386,44 +387,58 @@ pub struct ProtocolState {
...
+    /// either only an end timestamp or no timestamp (latest state).
     pub async fn by_id(
         component_ids: &[&str],
         chain_id: i64,
-        version_ts: Option<NaiveDateTime>,
+        start_ts: Option<NaiveDateTime>,
+        end_ts: Option<NaiveDateTime>,
         conn: &mut AsyncPgConnection,
-    ) -> QueryResult<Vec<(Self, String, Transaction)>> {
+    ) -> QueryResult<Vec<(Self, String, Bytes)>> {
```
kayibal: I find this interface quite complex, don't you think we should have two separate functions? One to get deltas and one to get states? 

Also not sure we need the filtering here, usually if we do retrieve deltas we get it for the complete chain and extractors would then filter the deltas application side.

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -386,44 +387,58 @@ pub struct ProtocolState {
...
             )
             .into_boxed();
 
-        if let Some(ts) = version_ts {
+        // if end timestamp is provided, we want to filter by valid_from <= end_ts
+        if let Some(ts) = end_ts {
             query = query.filter(protocol_state::valid_from.le(ts));
         }
 
+        // if start timestamp is provided, we want to filter by valid_from >= start_ts
```
kayibal: One of those two should not include the equal case.

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -386,44 +387,58 @@ pub struct ProtocolState {
...
         query
-            .select((Self::as_select(), protocol_component::external_id, Transaction::as_select()))
-            .get_results::<(Self, String, Transaction)>(conn)
+            .order_by((
+                protocol_state::protocol_component_id,
+                transaction::block_id,
+                transaction::index,
+            ))
+            .select((Self::as_select(), protocol_component::external_id, transaction::hash))
+            .get_results::<(Self, String, Bytes)>(conn)
```
kayibal: This will return a lot of duplicated entities if the range is big. Idk if you looked at the ContractStorage example, but there we manage to do the deduplication on the db side. But it required including the `ordinal` column from the transaction table into the `contract_store` table.

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -386,44 +387,58 @@ pub struct ProtocolState {
...
+    /// either only an end timestamp or no timestamp (latest state).
     pub async fn by_protocol_system(
         system: String,
         chain_id: i64,
-        version_ts: Option<NaiveDateTime>,
+        start_ts: Option<NaiveDateTime>,
+        end_ts: Option<NaiveDateTime>,
         conn: &mut AsyncPgConnection,
-    ) -> QueryResult<Vec<(Self, String, Transaction)>> {
+    ) -> QueryResult<Vec<(Self, String, Bytes)>> {
```
kayibal: Ok now I honestly didn't expect that we would have two versions of this interface...

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -435,44 +450,308 @@ impl ProtocolState {
...
+            .await
+    }
+
+    // retrieves old values for matching protocol states
+    pub async fn reverted_by_chain(
+        chain_id: i64,
+        start_ts: NaiveDateTime,
+        end_ts: NaiveDateTime,
+        conn: &mut AsyncPgConnection,
+    ) -> QueryResult<Vec<(String, String, Option<Bytes>)>> {
```
kayibal: Ok I am honestly surprised about this. Why do we need all these different flavours of the same thing?

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -435,44 +446,143 @@ impl ProtocolState {
...
+            .into_boxed();
+
+        // Subquery to exclude entities that have a valid at version at end_ts
+        let sub_query = format!("NOT EXISTS (
+                                SELECT 1 FROM protocol_state ps2
+                                WHERE ps2.protocol_component_id = protocol_state.protocol_component_id
+                                AND ps2.attribute_name = protocol_state.attribute_name
+                                AND ps2.valid_from <= '{}'
+                                AND (ps2.valid_to > '{}' OR ps2.valid_to IS NULL)
+                            )", end_ts, end_ts);
```
kayibal, line 692: Please use bind for parameters.

```suggestion
        let sub_query = format!("NOT EXISTS (
                                SELECT 1 FROM protocol_state ps2
                                WHERE ps2.protocol_component_id = protocol_state.protocol_component_id
                                AND ps2.attribute_name = protocol_state.attribute_name
                                AND ps2.valid_from <= $1
                                AND (ps2.valid_to > $1 OR ps2.valid_to IS NULL)
                            )").bind::<_, Timestamp>(end_ts);
```

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -33,55 +39,167 @@ where
...
+
+                    let mut updates = HashMap::new();
+                    let mut deleted = HashSet::new();
+                    for (component, attribute, value) in states_slice {
+                        if let Some(value) = value {
+                            updates.insert(attribute.clone(), value.clone());
+                        } else {
+                            deleted.insert(attribute.clone());
+                        }
+                    }
```
kayibal: So the value is None if the attribute was deleted? Does this come from previous value?

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -435,44 +446,143 @@ impl ProtocolState {
...
+                protocol_state::protocol_component_id,
+                transaction::block_id,
+                transaction::index,
+            ))
+            .select((Self::as_select(), protocol_component::external_id, transaction::hash))
+            .get_results::<(Self, String, Bytes)>(conn)
+            .await
+    }
+
+    // retrieves old values for matching protocol states
```
kayibal: It would be nice to extend this comment and describe the return value.

tycho-indexer/src/storage/postgres/protocol.rs
louise-poole: Yes, these values here are the previous values of all updated states. A 'None' value indicates there was no previous state and so the attribute was created. The reverse of this change is to delete that attribute. (I added comments to make this clearer)

tycho-indexer/src/storage/postgres/orm.rs
louise-poole: I attempted multiple times and am struggling to get bind to work for this query

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -394,55 +398,52 @@ pub struct ProtocolState {
...
+    ) -> QueryResult<Vec<(Self, ComponentId, TxHash)>> {
         let mut query = protocol_state::table
             .inner_join(protocol_component::table)
             .inner_join(transaction::table.on(transaction::id.eq(protocol_state::modify_tx)))
             .filter(protocol_component::external_id.eq_any(component_ids))
             .filter(protocol_component::chain_id.eq(chain_id))
             .filter(
                 protocol_state::valid_to
-                    .gt(end_ts)
+                    .gt(version_ts)
```
kayibal: What happens here with the `gt` is this is `None`? Did you ever debug that query? Is it just skipped or does it evaluate to `<= NULL`

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -394,55 +398,52 @@ pub struct ProtocolState {
...
     }
 
-    /// retrieves all matching protocol states along with their linked component ids and transaction
-    /// hashes. To get state deltas, provide a start and end timestamp. To get full states, provide
-    /// either only an end timestamp or no timestamp (latest state).
+    /// Used to fetch the full state of a component at a given version.
+    ///
+    /// Retrieves all matching protocol states, filtered by protocol system, along with their linked
+    /// component ids and transaction hashes. If no version is provided, the latest state is
+    /// returned. The results are grouped by component id and ordered by transaction.
```
kayibal: Maybe mention that this is basically the same as by_id but instead allows filtering by system. Going forward, we should be returning boxed queries here and only have something like `snapshot_by_version(Option<NaiveDataTime>)` and then the caller wll just call filter on the result. But absolutely not in scope of this PR ;)

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -521,75 +513,156 @@ impl ProtocolState {
...
-        protocol_state::table
+    ) -> QueryResult<Vec<(ComponentId, AttrStoreKey, Option<StoreVal>)>> {
+        // subquery to exclude entities that have a valid version at start_ts (weren't deleted)
+        let sub_query = format!("NOT EXISTS (
+                                SELECT 1 FROM protocol_state ps2
+                                WHERE ps2.protocol_component_id = protocol_state.protocol_component_id
+                                AND ps2.attribute_name = protocol_state.attribute_name
+                                AND ps2.valid_from <= '{}'
+                                AND (ps2.valid_to > '{}' OR ps2.valid_to IS NULL)
+                            )", start_ts, start_ts);
```
kayibal: Still not using bind?

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -521,75 +513,156 @@ impl ProtocolState {
...
 
-    // retrieves old values for matching protocol states
+    /// Used to retrieve the original state of all component attributes that were updated within the
+    /// given timeframe.
+    ///
+    /// Retrieves the previous values (reverse deltas) of all state updates that were applied after
+    /// target_ts and before start_ts, filtered by chain. Please note - this function is intended to
+    /// be used to fetch beckwards changes/revert deltas. Start_ts represents the more recent ts
+    /// and target_ts represents the ts of the block to be reverted to. The results are grouped by
+    /// component id and ordered by transaction.
```
kayibal: Should we add the same notice to the deltas_by_chain? And maybe name the `forward_delta_by_chain` and `reverse_delta_by_chain`?

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -521,75 +513,156 @@ impl ProtocolState {
...
             )
-            .into_boxed();
+            .filter(protocol_state::valid_from.le(end_ts))
+            .filter(protocol_state::valid_from.gt(start_ts))
+            .order_by((
+                protocol_state::protocol_component_id,
+                transaction::block_id,
+                transaction::index,
+            ))
+            .select((Self::as_select(), protocol_component::external_id, transaction::hash))
```
kayibal, line 666: I think we could deduplicate here by `component_id, attribute` using `DISTINCT`, what do you think? If an attribute changed multiple times within the time frame we are only interested in it's most recent change.

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -521,75 +513,156 @@ impl ProtocolState {
...
                 transaction::index,
             ))
-            .select((
-                protocol_component::external_id,
+            .select(
+                sql::<(sql_types::Text, sql_types::Text, sql_types::Nullable<sql_types::Bytea>)>(
+                    "external_id, attribute_name, previous_value AS value",
+                ),
+            )
+            .distinct_on((protocol_state::protocol_component_id, protocol_state::attribute_name));
```
kayibal, line 756: Interesting here we have the distinct clause I was missing in the forward case.

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -521,75 +513,156 @@ impl ProtocolState {
...
+        // states need to be reinstated so we return the component id, attribute name and latest
+        // value of each component-attribute pair here.
+        let deleted_query = query
+            // validity ends during the timeframe (potentially deleted)
+            .filter(protocol_state::valid_to.le(start_ts))
+            .filter(protocol_state::valid_to.gt(target_ts))
+            // validity starts before the timeframe (is valid at target_ts)
+            .filter(protocol_state::valid_from.le(target_ts))
+            // subquery to remove those that weren't deleted (valid version exists at start_ts)
+            .filter(sql::<Bool>(&sub_query))
```
kayibal, line 778: nit: `sub_query` defined too far away from where it is used.

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -521,75 +513,156 @@ impl ProtocolState {
...
-            .distinct_on((protocol_state::protocol_component_id, protocol_state::attribute_name))
+            .select(
+                sql::<(sql_types::Text, sql_types::Text, sql_types::Nullable<sql_types::Bytea>)>(
+                    "external_id, attribute_name, attribute_value AS value",
+                ),
+            );
+
+        // query and merge results for both reverted updates and deleted states
+        reverted_query
+            .union_all(deleted_query)
```
kayibal, line 793: Ok fancy hehe. That explains the weirdness with the Selects. Slightly would have preferred two queries since we do not to any additional things with this Union on the db side. But let's keep it like this we will later probably have to do some query performance measuring.

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -676,37 +566,153 @@ where
...
+                chain_db_id,
+                start_ts,
+                end_ts,
+                conn,
+            )
+            .await;
+
+            // fetch updated component attributes
+            let state_updates =
+                orm::ProtocolState::deltas_by_chain(chain_db_id, start_ts, end_ts, conn).await;
```
kayibal: Want to get rid of that ugly match below?
```suggestion
            let state_updates =
                orm::ProtocolState::deltas_by_chain(chain_db_id, start_ts, end_ts, conn)
                .await
                .map_err(|err| StorageError::from_diesel(
                    err,
                    "ProtocolStates",
                    chain.to_string().as_str(),
                    None
                )?;
```

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -676,37 +566,153 @@ where
...
+
+            // decode final state deltas
+            match state_updates {
+                Ok(data_vec) => {
+                    let mut protocol_states = Vec::new();
+
+                    let deleted_vec = match deleted_attrs {
+                        Ok(value) => value,
+                        Err(_) => Vec::new(),
+                    };
```
kayibal: This seems a bit out place, can't you make the method return empty vec?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -676,37 +566,153 @@ where
...
+
+                    // Variable to keep track of current position in deleted_vec
+                    let mut j = 0;
+
+                    // Variable to keep track of current position in data_vec
+                    let mut i = 0;
+
+                    while i < data_vec.len() {
```
kayibal: If I see this concentration mode turns on, can you please make me less scared by adding a comment about what will happen here? The i,j comments are not very helpful :D

```suggestion
                   // Comment about logic below

                   let mut protocol_states = Vec::new();
                   let (mut i, mut j) = (0, 0)
                   
                   while i < data_vec.len() {
```

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -676,37 +566,153 @@ where
...
+
+                        let deleted_start = j;
+                        // Iterate over deleted attributes until the component_id changes
+                        while j < deleted_vec.len() && &deleted_vec[j].0 == current_component_id {
+                            j += 1;
+                        }
+
+                        let states_slice = &data_vec[component_start..i];
+                        let tx_hash = &states_slice.last().unwrap().2; // Last element has the latest transaction
+                        let deleted_slice = &deleted_vec[deleted_start..j];
```
kayibal: Looks like maybe sth like this may work and would be a bit more verbose, maybe worth spending max 5-10m on:

```
state_group = data_vec.iter().groupby(|state| state.1)
deleted_group = deleted_vec.iter().groupby(|state| state.1).collect::<HashMap<i64, Vec<_>>>()
state_group.into_iter().map(|| // contruct deltas)
```

It's worse in performance though, cause this uses a hash map allocation but in case the current code is buggy the above might be easier.

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -676,37 +566,153 @@ where
...
+                                // if prev_value is null, then the attribute was created and must be
+                                // deleted on revert
+                                deleted.insert(attribute.clone());
+                            }
+                        }
+                        let state_delta = ProtocolStateDelta {
+                            component_id: current_component_id.clone(),
+                            updated_attributes: updates,
+                            deleted_attributes: deleted,
+                        };
```
kayibal: Why do we not use from_storage here?

tycho-indexer/src/storage/postgres/protocol.rs
kayibal: Actually while writing this I realised that if both are ordered by component ids, we don't need the hash map, we can just rely on the order and then we can use zip:
```
state_group = data_vec.iter().groupby(|state| state.1)
deleted_group = deleted_vec.iter().groupby(|state| state.1)
state_group.into_iter().zip_longest(deleted_group.into_iter()).map(|state_slice, deleted_slice| // contruct deltas)
```

tycho-indexer/src/storage/postgres/orm.rs
louise-poole: Yeah it evaluates to `> NULL`. I should conditionally add this filter - well spotted!

tycho-indexer/src/storage/postgres/orm.rs
louise-poole: Yeah, I responded to your bind suggestion above - I am struggling to get it to work in this context.

tycho-indexer/src/storage/postgres/orm.rs
kayibal: The fact that tests did not catch this is not great ;)

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -501,7 +501,7 @@ pub struct ProtocolComponent {
     /// address or in the case of a one-to-many relationship it could be something like
     /// 'USDC-ETH'. This is for example the case with ambient, where one contract is
     /// responsible for multiple components.
-    pub id: String,
+    pub id: ComponentId,
```
tvinagre, line 504: Thanks! Makes things way easier to comprehend

tycho-indexer/src/extractor/evm/storage.rs
```diff
@@ -441,60 +438,25 @@ pub mod pg {
...
     impl StorableProtocolStateDelta<orm::ProtocolState, orm::NewProtocolState, i64>
         for evm::ProtocolStateDelta
     {
         fn from_storage(
-            val: orm::ProtocolState,
+            vals: Vec<orm::ProtocolState>,
             component_id: String,
             tx_hash: &TxHash,
-            change: ChangeType,
+            deleted_attributes: Vec<AttrStoreKey>,
```
tvinagre, line 473: Sorry I don't understand this change. Isn't the `StateDelta` two-ways? (created or deleted)?

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -416,22 +421,29 @@ impl ProtocolState {
...
 
+        // if a version timestamp is provided, we want to filter by valid_from <= version_ts
         if let Some(ts) = version_ts {
             query = query.filter(protocol_state::valid_from.le(ts));
         }
 
         query
-            .select((Self::as_select(), protocol_component::external_id, Transaction::as_select()))
-            .get_results::<(Self, String, Transaction)>(conn)
+            .order_by((protocol_component::external_id, transaction::block_id, transaction::index))
```
tvinagre, line 542: Maybe comment why the ordering was structured like this / performance impacts

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -416,22 +421,29 @@ impl ProtocolState {
...
+            .select((Self::as_select(), protocol_component::external_id, transaction::hash))
+            .get_results::<(Self, String, Bytes)>(conn)
             .await
     }
 
+    /// Used to fetch the full state of a component at a given version.
```
tvinagre: , filtered by protocol_system

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -448,21 +460,32 @@ impl ProtocolState {
...
+                protocol_state::protocol_component_id,
+                transaction::block_id,
+                transaction::index,
+            ))
+            .select((Self::as_select(), protocol_component::external_id, transaction::hash))
+            .get_results::<(Self, String, Bytes)>(conn)
             .await
     }
 
+    /// Used to fetch the full state of a component at a given version.
```
tvinagre: , filtered by chain

tycho-indexer/src/storage/postgres/orm.rs
tvinagre: Noob question but why is bind better here?

louise-poole: It provides more control (or at least type validation) over the vars you're inserting into the sql statement - helps protect against sql injection attacks etc. Also increases query performance as the db caches the sql statement's execution plan and can reuse it with different parameters using the bind. The way I currently have it implemented (build the sql statement as a formatted string), a new execution plan needs to be created every time the params change.

tycho-indexer/src/extractor/evm/storage.rs
louise-poole: `ProtocolStateDelta` can have changes in both directions at once. It does not represent merely a single attribute change (like we have in the db), but rather can hold all changes for a given component id. Before the above change, I would use `from_storage` to decode all updates at once, and then use it again to decode all deletes at once. Then I'd merge the outputs together into one delta. This was not optimal - especially because we do not need to waste resources decoding full protocol states just to get the deleted attribute's name. Therefore I adjusted this to have all the info it needs to create the full state at once.

tycho-indexer/src/storage/postgres/orm.rs
louise-poole: We already get the most recent change, using distinct here is unnecessary - no duplicates should appear in the results as it is. We know only one attribute version is valid at a time. Here I filter for states where valid_from <= end_ts < valid_to - only one state per attribute should pass this filter as it's looking for the version actively valid _at_ end_ts.

tycho-indexer/src/storage/postgres/orm.rs
louise-poole: Yes - here it is necessary as I do not have a exact moment in time I can filter the state versions for, merely need to get the oldest.

tycho-indexer/src/storage/postgres/protocol.rs
louise-poole: `from_storage` decodes `orm::ProtocolStates`. The query used to get reverse deltas does not return full protocol states, only the component id, attribute and prev_value (do not want to transfer excess data like current attribute info and transaction info when it's not needed). Therefore we need an adjusted version of the `from_storage` logic that can build states from the info provided.

tycho-indexer/src/storage/postgres/protocol.rs
louise-poole: These `group_by`'s aren't quite working for me unfortunately...

tycho-indexer/src/storage/postgres/orm.rs
louise-poole: I just had another look here - the line after this one is `.or(protocol_state::valid_to.is_null()),` which means if `version_ts` is None and `.gt(version_ts)` fails, the 'or' is checked (looking for the current version).

PR #98, opened 2024-01-22
ENG-2711: Ambient Substreams: Balance and pool store

dianacarvalho1:
Add map_pool_flows, store_pools_balances and store_pool modules to ambient substreams
This is necessary because we need to know the balance messages we have are relative to the previous balances. This means that we need to index the pools from its creation to have the correct balances

### DIFF
--- a/tycho-indexer/src/extractor/evm/ambient.rs
+++ b/tycho-indexer/src/extractor/evm/ambient.rs
@@ -1,6 +1,6 @@
 #![allow(unused_variables)]
 
-use std::{str::FromStr, sync::Arc};
+use std::{collections::HashMap, str::FromStr, sync::Arc};
 
 use async_trait::async_trait;
 use diesel_async::{
@@ -17,7 +17,7 @@ use super::{AccountUpdate, Block};
 use crate::{
     extractor::{evm, ExtractionError, Extractor, ExtractorMsg},
     hex_bytes::Bytes,
-    models::{Chain, ExtractionState, ExtractorIdentity},
+    models::{Chain, ExtractionState, ExtractorIdentity, ProtocolType},
     pb::{
         sf::substreams::rpc::v2::{BlockScopedData, BlockUndoSignal, ModulesProgress},
         tycho::evm::v1::BlockContractChanges,
@@ -40,6 +40,7 @@ pub struct AmbientContractExtractor<G> {
     // TODO: There is not reason this needs to be shared
     // try removing the Mutex
     inner: Arc<Mutex<Inner>>,
+    protocol_types: HashMap<String, ProtocolType>,
 }
 
 impl<DB> AmbientContractExtractor<DB> {
@@ -66,6 +67,9 @@ pub struct AmbientPgGateway {
 #[async_trait]
 pub trait AmbientGateway: Send + Sync {
     async fn get_cursor(&self) -> Result<Vec<u8>, StorageError>;
+
+    async fn ensure_protocol_types(&self, new_protocol_types: &[ProtocolType]);
+
     async fn upsert_contract(
         &self,
         changes: &evm::BlockContractChanges,
@@ -83,6 +87,14 @@ pub trait AmbientGateway: Send + Sync {
 impl AmbientPgGateway {
     pub fn new(name: &str, chain: Chain, pool: Pool<AsyncPgConnection>, gw: CachedGateway) -> Self {
         AmbientPgGateway { name: name.to_owned(), chain, pool, state_gateway: gw }
+    }
+
+    async fn ensure_protocol_types(&self, new_protocol_types: &[ProtocolType]) {
+        let mut conn = self.pool.get().await.unwrap();
+        self.state_gateway
+            .add_protocol_types(new_protocol_types, &mut *conn)
+            .await
+            .expect("Couldn't insert protocol types");
     }
 
     #[instrument(skip_all)]
@@ -194,6 +206,10 @@ impl AmbientGateway for AmbientPgGateway {
         self.get_last_cursor(&mut conn).await
     }
 
+    async fn ensure_protocol_types(&self, new_protocol_types: &[ProtocolType]) {
+        todo!()
+    }
+
     #[instrument(skip_all, fields(chain = % self.chain, name = % self.name, block_number = % changes.block.number))]
     async fn upsert_contract(
         &self,
@@ -233,7 +249,12 @@ impl<G> AmbientContractExtractor<G>
 where
     G: AmbientGateway,
 {
-    pub async fn new(name: &str, chain: Chain, gateway: G) -> Result<Self, ExtractionError> {
+    pub async fn new(
+        name: &str,
+        chain: Chain,
+        gateway: G,
+        protocol_types: HashMap<String, ProtocolType>,
+    ) -> Result<Self, ExtractionError> {
         // check if this extractor has state
         let res = match gateway.get_cursor().await {
             Err(StorageError::NotFound(_, _)) => AmbientContractExtractor {
@@ -245,6 +266,7 @@ where
                     last_processed_block: None,
                 })),
                 protocol_system: "ambient".to_string(),
+                protocol_types,
             },
             Ok(cursor) => AmbientContractExtractor {
                 gateway,
@@ -252,9 +274,12 @@ where
                 chain,
                 inner: Arc::new(Mutex::new(Inner { cursor, last_processed_block: None })),
                 protocol_system: "ambient".to_string(),
+                protocol_types,
             },
             Err(err) => return Err(ExtractionError::Setup(err.to_string())),
         };
+
+        res.ensure_protocol_types().await;
         Ok(res)
     }
 }
@@ -266,6 +291,18 @@ where
 {
     fn get_id(&self) -> ExtractorIdentity {
         ExtractorIdentity::new(self.chain, &self.name)
+    }
+
+    /// Make sure that the protocol types are present in the database.
+    async fn ensure_protocol_types(&self) {
+        let protocol_types: Vec<ProtocolType> = self
+            .protocol_types
+            .values()
+            .cloned()
+            .collect();
+        self.gateway
+            .ensure_protocol_types(&protocol_types)
+            .await;
     }
 
     async fn get_cursor(&self) -> String {
@@ -296,14 +333,16 @@ where
 
         debug!(?raw_msg, "Received message");
 
-        // TODO: figure out how/where to get this ID from (in ENG-2049)
-        let protocol_type_id = String::from("id-1");
+        let protocol_type = self
+            .protocol_types
+            .get("vm:pool")
+            .expect("Couldn't find Protocol Type");
         let msg = match evm::BlockContractChanges::try_from_message(
             raw_msg,
             &self.name,
             self.chain,
             self.protocol_system.clone(),
-            protocol_type_id,
+            protocol_type.name.clone(),
         ) {
             Ok(changes) => {
                 tracing::Span::current().record("block_number", changes.block.number);
@@ -374,19 +413,45 @@ where
 
 #[cfg(test)]
 mod test {
-    use crate::{extractor::evm, pb::sf::substreams::v1::BlockRef};
+    use crate::{
+        extractor::evm,
+        models::{FinancialType, ImplementationType},
+        pb::sf::substreams::v1::BlockRef,
+    };
 
     use super::*;
 
+    fn ambient_protocol_types() -> HashMap<String, ProtocolType> {
+        let mut ambient_protocol_types = HashMap::new();
+        ambient_protocol_types.insert(
+            "vm:pool".to_string(),
+            ProtocolType::new(
+                "ambient_pool".to_string(),
+                FinancialType::Swap,
+                None,
+                ImplementationType::Vm,
+            ),
+        );
+        ambient_protocol_types
+    }
     #[tokio::test]
     async fn test_get_cursor() {
         let mut gw = MockAmbientGateway::new();
+        gw.expect_ensure_protocol_types()
+            .times(1)
+            .returning(|_| ());
         gw.expect_get_cursor()
             .times(1)
             .returning(|| Ok("cursor".into()));
-        let extractor = AmbientContractExtractor::new("vm:ambient", Chain::Ethereum, gw)
-            .await
-            .expect("extractor init ok");
+
+        let extractor = AmbientContractExtractor::new(
+            "vm:ambient",
+            Chain::Ethereum,
+            gw,
+            ambient_protocol_types(),
+        )
+        .await
+        .expect("extractor init ok");
 
         let res = extractor.get_cursor().await;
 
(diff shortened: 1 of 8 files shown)


substreams/ethereum-ambient/src/lib.rs
```diff
@@ -168,10 +138,101 @@ fn map_changes(
...
+/// 1. It is hardwired to only care about changes to the ambient main contract, this is ok for this
+///    particular use case but for a more general purpose implementation this is not ideal
+/// 2. Changes are processed separately, this means that if there are any side effects between each
+///    other (e.g. if account is deleted and then created again in ethereum all the storage is set
+///    to 0. So there is a side effect between account creation and contract storage.) these might
+///    not be properly accounted for. Most of the time this should not be a major issue but may lead
+///    to wrong results so consume this implementation with care. See example below for a concrete
+///    case where this is problematic.
+///
+/// ## A very contrived example:
```
tamaralipows: This diff is weird... why is this all showing up as a change? You didn't change this right?

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -168,10 +138,101 @@ fn map_changes(
                 }
                 _ => None,
             };
-
-            // TODO: if not None, unwrap result into `(pool_hash, base_flow, quote_flow)` and
-            // aggregate these with the previous balances to get new balances
+            let (_pool_hash, _base_flow, _quote_flow) = match _result {
```
tamaralipows: Small thing, I underscored these because they were unused so they gave lint errors. It seems you are using them now, so no need to underscore I think

substreams/ethereum-ambient/src/lib.rs
dianacarvalho1: yeah this diff is really weird. I didn't change it no

proto/tycho/evm/v1/vm.proto
```diff
@@ -47,3 +47,24 @@ message BlockContractChanges {
   // The set of transaction changes observed in the specified block.
   repeated TransactionContractChanges changes = 2;
 }
+
+
+// A change to a pool's balance. Ambient specific.
```
kayibal: Please put ambient specific models into their own separate module.

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -62,46 +72,12 @@ impl From<InterimContractChange> for tycho::ContractChange {
...
-            })
-        })
-        .collect();
-
+) -> Result<tycho::BlockPoolChanges, substreams::errors::Error> {
+    let mut protocol_components = Vec::new();
+    let mut balance_deltas = Vec::new();
     for block_tx in block.transactions() {
         // extract storage changes
         let mut storage_changes = block_tx
```
kayibal: This seems unused?

dianacarvalho1: no no, it is used!

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -168,10 +141,112 @@ fn map_changes(
...
+
+#[substreams::handlers::store]
+pub fn store_pools_balances(changes: tycho::BlockPoolChanges, balance_store: StoreAddBigInt) {
+    for component in changes.protocol_components {
+        let pool_hash_hex = hex::encode(&component.id);
+        balance_store.add(0, format!("{}{}", pool_hash_hex, "base"), BigInt::from(0));
+        balance_store.add(0, format!("{}{}", pool_hash_hex, "quote"), BigInt::from(0));
+    }
+    let mut deltas = changes.balance_deltas.clone();
+    deltas.sort_by_key(|delta| delta.ordinal);
```
dianacarvalho1: they need to be sorted so that the insert into the store works

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -168,10 +141,112 @@ fn map_changes(
...
+    for balance_delta in deltas {
+        let pool_hash_hex = hex::encode(&balance_delta.pool_hash);
+        balance_store.add(
+            balance_delta.ordinal + 1,
```
dianacarvalho1: to make sure it won't be 0 and conflict with the additions above on pool creation

substreams/ethereum-ambient/src/contracts/warmpath.rs
```diff
@@ -23,53 +26,64 @@ const LIQUIDITY_CHANGE_ABI: &[ParamType] = &[
...
+        let input_bytes = external_cmd[0]
+            .to_owned()
+            .into_bytes() // Convert Bytes32 to Vec<u8>
+            .ok_or_else(|| anyhow!("Failed to hotproxy userCmd input data.".to_string()))?;
+
+        let code = input_bytes[31];
+        let is_mint =
+            code == 1 || code == 11 || code == 12 || code == 3 || code == 31 || code == 32;
+        let is_burn =
+            code == 2 || code == 21 || code == 22 || code == 4 || code == 41 || code == 42;
```
SmartMonkey-git: I think, this also works and is a bit more appealing:
```suggestion
let is_burn = [2, 21, 22, 4, 41, 42].contains(&code);
```
```

File: substreams/ethereum-ambient/src/contracts/warmpath.rs [hunk 8]
```diff
@@ -23,53 +26,64 @@ const LIQUIDITY_CHANGE_ABI: &[ParamType] = &[
...
+        let is_harvest = code == 5;
+        if is_mint || is_burn || is_harvest {
```
SmartMonkey-git, line 39: Why do we need to differentiate between them? It seems like there is no specific action taken for any of them.

substreams/ethereum-ambient/src/contracts/warmpath.rs
```diff
@@ -23,53 +26,64 @@ const LIQUIDITY_CHANGE_ABI: &[ParamType] = &[
...
+            if let Ok(liquidity_change_calldata) = decode(LIQUIDITY_CHANGE_ABI, &input_bytes) {
+                let base_token = liquidity_change_calldata[1]
+                    .to_owned()
+                    .into_address()
+                    .ok_or_else(|| {
+                        anyhow!(
                         "Failed to convert base token to address for WarmPath userCmd call: {:?}",
```
SmartMonkey-git, line 46: Is the error message correct here?

SmartMonkey-git: Same question for the quote_token

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -309,6 +384,45 @@ fn map_changes(
...
+            tx_change
+                .component_changes
+                .push(component.clone());
+        }
+
+        for balance_delta in &block_pool_changes.balance_deltas {
+            let pool_hash_hex = hex::encode(&balance_delta.pool_hash);
+            let pool = match pool_store.get_last(pool_hash_hex.clone()) {
+                Some(pool) => pool,
+                None => panic!("Pool not found in store for given hash: {}", pool_hash_hex),
```
SmartMonkey-git: Why do we want to panic here?

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -309,6 +384,45 @@ fn map_changes(
...
+            };
+            let base_balance = balance_store.get_last(format!("{}{}", pool_hash_hex, "base"));
+            let quote_balance = balance_store.get_last(format!("{}{}", pool_hash_hex, "quote"));
+
+            let base_balance_change = BalanceChange {
+                component_id: balance_delta.pool_hash.clone(),
+                token: pool.tokens[0].clone(),
```
SmartMonkey-git: How do you know its the first token of the list?

Also, I think you can do:
```suggestion
                token: pool.tokens.first().clone(),
```

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -309,6 +384,45 @@ fn map_changes(
...
+                balance: base_balance
+                    .clone()
+                    .expect("Couldn't get balance from store")
```
SmartMonkey-git:
```suggestion
                    .expect("Couldn't get base balance from store")
```

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -309,6 +384,45 @@ fn map_changes(
...
+                    .to_bytes_be()
+                    .1,
+            };
+            let quote_balance_change = BalanceChange {
+                component_id: balance_delta.pool_hash.clone(),
+                token: pool.tokens[1].clone(),
+                balance: quote_balance
+                    .expect("Couldn't get balance from store")
```
SmartMonkey-git:
```suggestion
                    .expect("Couldn't get quote balance from store")
```

proto/tycho/evm/v1/ambient.proto
```diff
@@ -0,0 +1,25 @@
+syntax = "proto3";
+
+package tycho.evm.v1;
+
+import "tycho/evm/v1/common.proto";
```
tvinagre, line 5: hmm, I thought this folder was only for generic protos, not protocol specific. But I'm fine with it
Might move my UniswapV2 here

tvinagre: And does this protofub work if you're compiling from a subpackage? Does it acknowledge the correct path?

proto/tycho/evm/v1/ambient.proto
```diff
@@ -0,0 +1,25 @@
...
+// A change to a pool's balance. Ambient specific.
+message BalanceDelta {
+  // The address of the ERC20 token whose balance changed.
+  bytes pool_hash = 1;
+  // The delta of the base token.
+  bytes base_token_delta = 2;
+  // The delta of the quote token.
+  bytes quote_token_delta = 3;
+  // Used to determine the order of the balance changes. Necessary for the balance store.
+  uint64 ordinal = 4;
```
tamaralipows, line 16: This fixed a problem you had? I'm surprised that this matters tbh

substreams/ethereum-ambient/src/contracts/warmpath.rs
tamaralipows: Diana didn't change this code, this was from a previous PR. I only made a small fix to it in this PR and now it shows up as all changed :/ This is taken directly from the mapping.ts method that we based this off of, and I think this is fine for readability or if we ever need to debug.

substreams/ethereum-ambient/src/contracts/warmpath.rs
tamaralipows: What is wrong with it?

substreams/ethereum-ambient/src/lib.rs
tamaralipows: so strange...

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -62,46 +72,12 @@ impl From<InterimContractChange> for tycho::ContractChange {
...
-/// ## A very contrived example:
-/// 1. Some existing contract receives a transaction that changes it state, the state is updated
-/// 2. Next, this contract has self destruct called on itself
-/// 3. The contract is created again using CREATE2 at the same address
-/// 4. The contract receives a transaction that changes it state
-/// 5. We would emit this as as contract creation with slots set from 1 and from 4, although we
-///    should only emit the slots changed from 4.
 #[substreams::handlers::map]
-fn map_changes(
+fn map_pool_changes(
```
tvinagre: Maybe would be nice to have one file per handler?

substreams/ethereum-ambient/src/lib.rs
tamaralipows: Because this should never happen. It means something is wrong with our pool encoding, or it was not inserted properly upon creation.

substreams/ethereum-ambient/src/lib.rs
tamaralipows: We have agreed to always put the base token first and the quote token second

substreams/ethereum-ambient/src/lib.rs
tvinagre: Nice solution but, if the pool was created before the balance update (same block), then for sure the ordinal of the update would never be 0, no?

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -168,10 +141,112 @@ fn map_changes(
...
+    }
+    let pool_changes = tycho::BlockPoolChanges { protocol_components, balance_deltas };
+    Ok(pool_changes)
+}
+
+#[substreams::handlers::store]
+pub fn store_pools_balances(changes: tycho::BlockPoolChanges, balance_store: StoreAddBigInt) {
+    for component in changes.protocol_components {
+        let pool_hash_hex = hex::encode(&component.id);
+        balance_store.add(0, format!("{}{}", pool_hash_hex, "base"), BigInt::from(0));
```
tvinagre: If the key doesn't exist, if you call `add` it will do a `set`?

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -168,10 +141,112 @@ fn map_changes(
...
+/// 1. It is hardwired to only care about changes to the ambient main contract, this is ok for this
+///    particular use case but for a more general purpose implementation this is not ideal
+/// 2. Changes are processed separately, this means that if there are any side effects between each
+///    other (e.g. if account is deleted and then created again in ethereum all the storage is set
+///    to 0. So there is a side effect between account creation and contract storage.) these might
+///    not be properly accounted for. Most of the time this should not be a major issue but may lead
+///    to wrong results so consume this implementation with care. See example below for a concrete
+///    case where this is problematic.
+///
+/// ## A very contrived example:
```
tvinagre: I didn't understand if this example is for a single-block or if it happens on multiple blocks

substreams/ethereum-ambient/src/lib.rs
```diff
@@ -309,6 +384,45 @@ fn map_changes(
...
+            let quote_balance_change = BalanceChange {
+                component_id: balance_delta.pool_hash.clone(),
+                token: pool.tokens[1].clone(),
+                balance: quote_balance
+                    .expect("Couldn't get balance from store")
+                    .to_bytes_be()
+                    .1,
+            };
+
+            tx_change
```
tvinagre: I'm assuming you did this above bc you just want the final balance change for each pool - you don't care about the individual changes

proto/tycho/evm/v1/ambient.proto
kayibal: It is meant as that we should not add any protocol specific logic to it, these types will probably make it into a sdk lib later. So shouldn't have protocol specific things in there.

substreams/ethereum-ambient/src/lib.rs
kayibal: I like it

substreams/ethereum-ambient/src/lib.rs
dianacarvalho1: that is very true actually 🤔 so the +1 is unnecessary (will leave it anyway 👀  shouldn't hurt)

substreams/ethereum-ambient/src/lib.rs
dianacarvalho1: yes I think so

substreams/ethereum-ambient/src/lib.rs
dianacarvalho1: (I didn't change this docstring - git is crazy sometimes)

substreams/ethereum-ambient/src/lib.rs
dianacarvalho1: yes exactly

substreams/ethereum-ambient/src/lib.rs
dianacarvalho1: that suggestions doesn't actually work
````
Attempted to access method first as a field for type `Vec<Vec<u8, Global>, Global>`
````

proto/tycho/evm/v1/ambient.proto
dianacarvalho1: @kayibal I'm confused with your answer. Should I move them?

PR #99, opened 2024-01-23
Generic application side table versioning

kayibal:
We need to handle the versioning application side as triggers have not been effective for batch inserts.

More context: https://datarevenue.atlassian.net/browse/ENG-2544

There was quite some back and forth for me on what to do with the types. I summarised the situation under Design Decisions.

This is Part I and adds the main application side logic to achieve our goal. In Part II we will integrate this with the CachedGateway and change to a repeatable read isolation mode.

### DIFF
--- a/tycho-indexer/src/storage/postgres/versioning.rs
+++ b/tycho-indexer/src/storage/postgres/versioning.rs
@@ -11,8 +11,8 @@
 //!
 //! * `DeltaVersionedRow`: Same as above but will also set `previous_value`` attributes.
 //!
-//! * `StoredVersionedRow`: This trait provides to set the end version on currently active version
-//!   in the db based on the currently incoming data.
+//! * `StoredVersionedRow`: Enables setting the end version on currently active version in the db
+//!   based on new incoming entries.
 //!
 //! ## Notes
 //! To use the apply_versioning function defined here VersionRow::EntityId and
@@ -22,12 +22,12 @@
 //! # Design Decisions
 //!
 //! Initially we would support references in EntityId, to reduce the number of clones necessary for
-//! compley entity id types. This would lead to a strange situation, where these the trait bounds
-//! for the `apply_versioning` method would not be expressable. Reasons fot this are not 100% clear
-//! but `latest_versions_by_ids` referring to the `StoredVersionedRow::EntityId`` but actually being
-//! used with `VersionedRow::EntityId` is most likely related. Previous iterations had lifetimes on
-//! `StoredVersionedRow<'a>` but as said the mangement of lifetimes become increasingly complex to a
-//! point where apply_versioning was not always usable.
+//! complex entity id types. This would lead to a strange situation, where these trait bounds
+//! for the `apply_versioning` method would not be expressible. Reasons for this are not 100% clear,
+//! however, `latest_versions_by_ids` referring to the `StoredVersionedRow::EntityId`` but actually
+//! being used with `VersionedRow::EntityId` is most likely related. Previous iterations had
+//! lifetimes on `StoredVersionedRow<'a>` but as said, the management of lifetimes became
+//! increasingly complex to a point where apply_versioning was not always usable.
 //!
 //! Instead we removed support for references in the EntityId type for now and just accept the high
 //! number of clones necessary. This may be revisited later again in case the clones become a
@@ -53,7 +53,7 @@ use std::fmt::Debug;
 ///
 /// This trait enables querying the struct for its current state and allows to set the `valid_to``
 /// column in case we are inserting a historical row (row that is outdated at the time of insertion,
-/// but contributes to the historiy of the entity).
+/// but contributes to the history of the entity).
 pub trait VersionedRow {
     /// Rust type to use as key to sort a collection of structs by entity and time.
     type SortKey: Ord + Clone + Debug + Send + Sync;
@@ -75,7 +75,7 @@ pub trait VersionedRow {
     fn get_valid_from(&self) -> Self::Version;
 }
 
-/// Trait indicating that a struct can be insered in a delta versioned table.
+/// Trait indicating that a struct can be inserted in a delta versioned table.
 ///
 /// Delta versioned records require the previous value present in one of their columns. This serves
 /// to build both forward and backward delta changes while avoiding self joins.
@@ -89,7 +89,7 @@ pub trait DeltaVersionedRow {
     fn set_previous_value(&mut self, previous_value: Self::Value);
 }
 
-/// Indicates the struct relates to a stored entry in a versioned table
+/// Trait indicating that a struct relates to a stored entry in a versioned table.
 ///
 /// This struct is used to invalidate rows that are currently valid on the db side before inserting
 /// new versions for those entities.
@@ -192,6 +192,8 @@ fn build_batch_update_query<'a, O: StoredVersionedRow>(
     table_name: &str,
     end_versions: &'a HashMap<O::EntityId, O::Version>,
 ) -> BoxedSqlQuery<'a, Pg, SqlQuery> {
+    // Generate bind parameter 2-tuples the result will look like '($1, $2), ($3, $4), ...'
+    // These are later subsituted with the primary key and valid to values.
     let bind_params = (1..=objects.len() * 2)
         .map(|i| if i % 2 == 0 { format!("${}", i) } else { format!("(${}", i) })
         .collect::<Vec<String>>()


tycho-indexer/src/storage/postgres/versioning.rs
```diff
@@ -0,0 +1,277 @@
...
+//!
+//! ## Notes
+//! To use the apply_versioning function defined here VersionRow::EntityId and
+//! StoredVersionedRow::EntityId must be of the same type. Keep that in mind while implementing
+//! these traits on your structs.
+//!
+//! # Design Decisions
+//!
+//! Initially we would support references in EntityId, to reduce the number of clones necessary for
+//! compley entity id types. This would lead to a strange situation, where these the trait bounds
```
dianacarvalho1: compley?

tycho-indexer/src/storage/postgres/versioning.rs
```diff
@@ -0,0 +1,277 @@
...
+//! for the `apply_versioning` method would not be expressable. Reasons fot this are not 100% clear
```
dianacarvalho1:
```suggestion
//! for the `apply_versioning` method would not be expressible. Reasons for this are not 100% clear
```

tycho-indexer/src/storage/postgres/versioning.rs
```diff
@@ -0,0 +1,277 @@
...
+use std::{collections::HashMap, hash::Hash};
+
+use crate::storage::StorageError;
+use std::fmt::Debug;
+
+/// Trait indicating that a struct can be inserted into a versioned table.
+///
+/// This trait enables querying the struct for its current state and allows to set the `valid_to``
+/// column in case we are inserting a historical row (row that is outdated at the time of insertion,
+/// but contributes to the historiy of the entity).
```
dianacarvalho1:
```suggestion
/// but contributes to the history of the entity).
```

tycho-indexer/src/storage/postgres/versioning.rs
```diff
@@ -0,0 +1,277 @@
...
+    fn get_sort_key(&self) -> Self::SortKey;
+
+    /// Allows setting `valid_to`` column, thereby invalidating this version.
+    fn set_valid_to(&mut self, end_version: Self::Version);
+
+    /// Exposes the starting version.
+    fn get_valid_from(&self) -> Self::Version;
+}
+
+/// Trait indicating that a struct can be insered in a delta versioned table.
```
dianacarvalho1:
```suggestion
/// Trait indicating that a struct can be inserted in a delta versioned table.
```

tycho-indexer/src/storage/postgres/versioning.rs
```diff
@@ -0,0 +1,277 @@
...
+//!
+//! The module exposes three main traits that can be implemented to provide versioning logic:
+//!
+//! * `VersionedRow`: Gives this module access to versioning attributes such as valid_to. Implement
+//!   this trait to enable setting these attributes automatically to use batch insertion.
+//!
+//! * `DeltaVersionedRow`: Same as above but will also set `previous_value`` attributes.
+//!
+//! * `StoredVersionedRow`: This trait provides to set the end version on currently active version
+//!   in the db based on the currently incoming data.
```
louise-poole: "This trait provides to set" sounds a bit weird.

tycho-indexer/src/storage/postgres/versioning.rs
louise-poole: Maybe?
```suggestion
//! complex entity id types. This would lead to a strange situation, where these trait bounds
```

tycho-indexer/src/storage/postgres/versioning.rs
```diff
@@ -0,0 +1,277 @@
...
+//! This module provides access to versioning tools.
+//!
+//! # Traits
+//!
+//! The module exposes three main traits that can be implemented to provide versioning logic:
+//!
+//! * `VersionedRow`: Gives this module access to versioning attributes such as valid_to. Implement
+//!   this trait to enable setting these attributes automatically to use batch insertion.
+//!
+//! * `DeltaVersionedRow`: Same as above but will also set `previous_value`` attributes.
```
dianacarvalho1, line 12: why is the DeltaVersionedRow necessary? I see you implemented it for NewSlot. Why do we need to know the previous value there?

tycho-indexer/src/storage/postgres/versioning.rs
```diff
@@ -0,0 +1,277 @@
...
+//!
+//! # Design Decisions
+//!
+//! Initially we would support references in EntityId, to reduce the number of clones necessary for
+//! compley entity id types. This would lead to a strange situation, where these the trait bounds
+//! for the `apply_versioning` method would not be expressable. Reasons fot this are not 100% clear
+//! but `latest_versions_by_ids` referring to the `StoredVersionedRow::EntityId`` but actually being
+//! used with `VersionedRow::EntityId` is most likely related. Previous iterations had lifetimes on
+//! `StoredVersionedRow<'a>` but as said the mangement of lifetimes become increasingly complex to a
+//! point where apply_versioning was not always usable.
```
louise-poole: Grammer nazi time (soz, couldn't resist)
```suggestion
//! for the `apply_versioning` method would not be expressable. Reasons for this are not 100% clear,
//! however, `latest_versions_by_ids` referring to the `StoredVersionedRow::EntityId`` but actually being
//! used with `VersionedRow::EntityId` is most likely related. Previous iterations had lifetimes on
//! `StoredVersionedRow<'a>` but as said, the management of lifetimes became increasingly complex to a
//! point where apply_versioning was not always usable.
```

tycho-indexer/src/storage/postgres/versioning.rs
```diff
@@ -0,0 +1,277 @@
...
+    type Value: Clone + Debug;
+
+    /// Exposes the current value.
+    fn get_value(&self) -> Self::Value;
+
+    /// Sets the previous value.
+    fn set_previous_value(&mut self, previous_value: Self::Value);
+}
+
+/// Indicates the struct relates to a stored entry in a versioned table
```
louise-poole:
```suggestion
/// Trait indicating that a struct relates to a stored entry in a versioned table.
```

tycho-indexer/src/storage/postgres/versioning.rs
```diff
@@ -0,0 +1,277 @@
...
+///
+/// ## Note
+/// The associated types of this trait need to match with the types defined for the corresponding
+/// `VersionedRow` trait.
+#[async_trait]
+pub trait StoredVersionedRow {
+    /// The entity identifier type.
+    type EntityId: Ord + Hash + Debug + Send + Sync;
+    /// The primary key on the table for this row.
+    type PrimaryKey: Into<i64> + Debug + Send + Sync;
```
louise-poole, line 105: Ah, okay. So `EntityId` could be both the component id and the attribute name together as a pair? To uniquely identify each attribute state per component?

tycho-indexer/src/storage/postgres/versioning.rs
```diff
@@ -0,0 +1,277 @@
...
+    table_name: &str,
+    end_versions: &'a HashMap<O::EntityId, O::Version>,
+) -> BoxedSqlQuery<'a, Pg, SqlQuery> {
+    let bind_params = (1..=objects.len() * 2)
+        .map(|i| if i % 2 == 0 { format!("${}", i) } else { format!("(${}", i) })
+        .collect::<Vec<String>>()
+        .chunks(2)
+        .map(|chunk| chunk.join(", ") + ")")
+        .collect::<Vec<String>>()
+        .join(", ");
```
louise-poole, line 203: Okay whaaaaaaat is this 😂  EDIT-  took lots of back and forth and some brain power but I get it now 👍  This is creating a string of placeholders for you to populate later when building the final update query.

tycho-indexer/src/storage/postgres/versioning.rs
```diff
@@ -0,0 +1,277 @@
...
+}
+
+/// Builds a update query that updates multiple rows at once.
+///
+/// Builds a query that will take update multiple rows end versions. The rows are identified by
+/// their primary key and the version is retrieved from the `end_versions` parameter.
+///
+/// Building such a query with pure diesel is currently not supported as this query updates each
+/// primary key with a unique value. See: https://github.com/diesel-rs/diesel/discussions/2879
+fn build_batch_update_query<'a, O: StoredVersionedRow>(
```
louise-poole, line 190: Nice, this function holds some very clever coding to work around the lack of diesel batch update support. It's a bit complicated to try understand at first, adding comments might help with that?

tycho-indexer/src/storage/postgres/versioning.rs
kayibal: If we revert, e.g. we go 1 or more blocks back, we need to emit messages that reset the entities we are about to reset to their previous state. This is what this field is used for. So on reverts, we take all the affected rows, sort them by version and then take the earliest version to be removed and emit the previous value of that entity row.

You could achieve the same thing by doing self joins on the table but that is less efficient and quickly gets very complex with diesel.

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -659,6 +722,43 @@ impl ContractCode {
...
+        ids: I,
+        conn: &mut AsyncPgConnection,
+    ) -> Result<Vec<Box<Self>>, StorageError> {
+        Ok(contract_code::table
+            .filter(contract_code::account_id.eq_any(ids))
+            .select(Self::as_select())
+            .get_results::<Self>(conn)
+            .await?
+            .into_iter()
+            .map(Box::new)
```
louise-poole, line 753: Using `Box` is a memory optimisation trick if I'm remembering correctly?

tycho-indexer/src/storage/postgres/versioning.rs
```diff
@@ -0,0 +1,277 @@
...
+    fn get_valid_to(&self) -> Self::Version;
+
+    /// Exposes the entity id.
+    fn get_entity_id(&self) -> Self::EntityId;
+
+    /// Retrieves the latest versions for the passed entity ids from the database.
+    async fn latest_versions_by_ids<I: IntoIterator<Item = Self::EntityId> + Send + Sync>(
+        ids: I,
+        conn: &mut AsyncPgConnection,
+    ) -> Result<Vec<Box<Self>>, StorageError>;
```
louise-poole, line 122: We could make this more versatile and support getting the version at a given timestamp/block? If the version requested is None, we get the latest.

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -747,16 +869,103 @@ pub struct ContractStorage {
...
+            .filter(
+                contract_storage::account_id
+                    .eq_any(&accounts)
+                    .and(contract_storage::slot.eq_any(&slots))
+                    .and(contract_storage::valid_to.is_null()),
+            )
+            .get_results(conn)
+            .await?
+            .into_iter()
+            .filter(|cs| tuple_ids.contains(&(&cs.account_id, &cs.slot)))
```
louise-poole, line 913: Ah, clever! Never thought of doing this.

tycho-indexer/src/storage/postgres/orm.rs
kayibal: This is a rust constrain on traits, you can only returned sized types from trait objects. We don't know the size of Self before actually implementing it so I boxed it this way the size is fixed.

tycho-indexer/src/storage/postgres/orm.rs
kayibal: This is actually a workaround^^ ideally we would be able to do this with diesel but currently it is not supported.

tycho-indexer/src/storage/postgres/versioning.rs
kayibal: Yes correct!

tycho-indexer/src/storage/postgres/versioning.rs
kayibal: I had so many problems with this :D and getting the traits right, but feel free to try this. Adding this doesn't sound too hard though.

tycho-indexer/src/storage/postgres/versioning.rs
kayibal: Yes I want to use `bind` for dynamic parameters because that will check types and avoid sql injection issues (in case we would be attacked at some point via a hacked substreams pkg.. although very unlikely)

tycho-indexer/src/storage/postgres/versioning.rs
louise-poole: I'll add it to my cool down list 👍  Since it's not something necessary, but something I'd like to try get working anyway

tycho-indexer/src/storage/postgres/versioning.rs
kayibal: Added a comment there explaining this block it's quite something to take in...

tycho-indexer/src/storage/postgres/versioning.rs
kayibal: After thinking about this I'd probably not do it here, these traits are only used for versioning and it doesn't need this. Doesn't mean we could implement this and then satisfy the trait by calling the other implementation.

tycho-indexer/src/storage/postgres/versioning.rs
tvinagre: next level, congrats

PR #100, opened 2024-01-23
ENG-2034 Implement deleting components

SmartMonkey-git:
[Task](https://datarevenue.atlassian.net/browse/ENG-2034)

### DIFF
--- a/tycho-indexer/src/storage/postgres/protocol.rs
+++ b/tycho-indexer/src/storage/postgres/protocol.rs
@@ -1,8 +1,8 @@
 #![allow(unused_variables)]
 
 use async_trait::async_trait;
-use chrono::{NaiveDateTime, Utc};
-use std::{cmp::Ordering, collections::HashMap, hash::Hash, time::SystemTime};
+use chrono::NaiveDateTime;
+use std::{cmp::Ordering, collections::HashMap};
 
 use diesel::prelude::*;
 use diesel_async::{AsyncPgConnection, RunQueryDsl};
@@ -527,7 +527,7 @@ mod test {
         extractor::evm::{self, ERC20Token},
         storage::ChangeType,
     };
-    use chrono::{NaiveDate, NaiveDateTime, NaiveTime};
+    use chrono::{NaiveDate, NaiveDateTime, NaiveTime, Utc};
     use diesel_async::AsyncConnection;
     use ethers::{prelude::H160, types::U256};
     use rstest::rstest;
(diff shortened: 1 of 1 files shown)


tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -642,4 +673,51 @@ mod test {
...
+            .iter()
+            .collect::<Vec<_>>();
+
+        let res = gw
+            .delete_protocol_components(&to_delete, &mut conn)
+            .await;
+
+        assert!(res.is_ok());
+        for (index, component) in test_components.iter().enumerate() {
+            let updated_ts = schema::protocol_component::table
```
tamaralipows: Is this the `updated_ts`? It seems to be the `deleted_at`

SmartMonkey-git: It is the deleted_at timestamp that got updated.
The column that indicates an update of the table is called modified_ts

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -642,4 +673,51 @@ mod test {
...
+                .filter(schema::protocol_component::external_id.eq(component.id.0.to_string()))
+                .select(schema::protocol_component::deleted_at)
+                .first::<Option<NaiveDateTime>>(&mut conn)
+                .await
+                .unwrap();
+
+            assert!(updated_ts.is_some());
```
tamaralipows: Should we also check the length here? I think we expect 2, no?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -149,6 +150,25 @@ where
...
+            .iter()
+            .map(|c| c.id.0.to_string())
+            .collect();
+        let current_time: NaiveDateTime = Utc::now().naive_utc();
+
+        diesel::update(protocol_component.filter(external_id.eq_any(ids_to_delete)))
+            .set(deleted_at.eq(current_time))
+            .execute(conn)
+            .await?;
+        Ok(())
```
dianacarvalho1: this is the delete behaviour we want?? 🤔 why not real delete them?

kayibal: Backtesting

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -149,6 +150,25 @@ where
...
+        use super::schema::protocol_component::dsl::*;
+
+        let ids_to_delete: Vec<String> = to_delete
+            .iter()
+            .map(|c| c.id.0.to_string())
+            .collect();
+        let current_time: NaiveDateTime = Utc::now().naive_utc();
+
+        diesel::update(protocol_component.filter(external_id.eq_any(ids_to_delete)))
+            .set(deleted_at.eq(current_time))
```
kayibal: This can't be current_time it must be related to a block timestamp

PR #101, opened 2024-01-23
ENG-2700: Clear up ContractID structs

SmartMonkey-git:
[Task](https://datarevenue.atlassian.net/browse/ENG-2700)
It was more confusing, then actually helping. Documentation has been inherited.

### DIFF
--- a/tycho-indexer/src/extractor/evm/mod.rs
+++ b/tycho-indexer/src/extractor/evm/mod.rs
@@ -31,6 +31,7 @@ mod utils;
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
 pub struct SwapPool {}
 
+#[derive(Debug)]
 pub struct ERC20Token {
     pub address: H160,
     pub symbol: String,
--- a/tycho-indexer/migrations/2024-01-10_tycho_patch/down.sql
+++ b/tycho-indexer/migrations/2024-01-10_tycho_patch/down.sql
@@ -43,10 +43,10 @@ END;
 $$
 LANGUAGE plpgsql;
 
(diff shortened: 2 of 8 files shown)


tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -490,8 +490,11 @@ impl TvlChange {
 /// smart contract.
 #[derive(Debug, Clone, PartialEq, Default, Deserialize, Serialize)]
 pub struct ProtocolComponent {
-    // an id for this component, could be hex repr of contract address
-    pub id: ContractId,
+    /// Represents the unique identifier of a contract. It can represent an on-chain
+    /// address or in the case of a one-to-many relationship it could be something like
+    /// 'USDC-ETH'. This is for example the case with ambient, where one component is
+    /// responsible for multiple contracts.
```
dianacarvalho1: is it though??? didn't we decide to have multiple components for each of the ambient pools (token pair) anyway. ai I feel like we keep having this discussion. maybe I'm confused too. what did we decide before? @tamaralipows

tamaralipows: Yeah we decided to have multiple components for each of the pools (I think you told me that the most recently) haha. That's how I set it up in substreams at least.

dianacarvalho1: @SmartMonkey-git can you just fix the docstring so it isn't misleading please?

kayibal: I don't get why we would have multiple components per pool? I understand why we would have multiple components per contract.

PR #102, opened 2024-01-24
ENG-2037: Insert Component Balances

tamaralipows:
[Issue](https://datarevenue.atlassian.net/browse/ENG-2037)


tycho-indexer/migrations/2024-01-10_tycho_patch/up.sql
```diff
@@ -52,4 +52,19 @@ ADD CONSTRAINT name_unique UNIQUE (name);
...
+--  Saves the TVL change of a protocol component.
+
+CREATE TABLE IF NOT EXISTS tvl_change(
+    "id" bigserial PRIMARY KEY,
+    -- id of the token whose tvl changed
+    "token_id" bigint REFERENCES "token"(id) NOT NULL,
+    -- new balance of the token for this component
+    "new_balance" varchar(255) NOT NULL,
+    -- the transaction that modified the tvl of this component
+    "modify_tx" bigint REFERENCES "transaction"(id) ON DELETE CASCADE NOT NULL,
```
kayibal, line 72: Unfortunately I think this version will need to be versioned as well, so we'll need valid_to and valid_from columns here.

tycho-indexer/migrations/2024-01-10_tycho_patch/up.sql
```diff
@@ -52,4 +52,19 @@ ADD CONSTRAINT name_unique UNIQUE (name);
 
 -- Make the "account_id" column of the token unique
 ALTER TABLE token
-    ADD CONSTRAINT unique_account_id_constraint UNIQUE (account_id);
\ No newline at end of file
+    ADD CONSTRAINT unique_account_id_constraint UNIQUE (account_id);
+--  Saves the TVL change of a protocol component.
+
+CREATE TABLE IF NOT EXISTS tvl_change(
```
kayibal: Should we rename to `component_balance`?

tamaralipows: Yeah, this will indeed be renamed. Will do it in its own commit as the final step

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -308,6 +308,55 @@ pub struct ProtocolType {
     pub modified_ts: NaiveDateTime,
 }
 
+#[derive(Identifiable, Queryable, Selectable)]
+#[diesel(table_name = component_balance)]
+#[diesel(belongs_to(ProtocolComponent))]
```
tamaralipows, line 313: @dianacarvalho1 I know we mentioned also referencing the Token and Transaction here, but in the AccountBalance object, I noticed we are not doing that for Transaction, and only referencing Account there. So I did it this way for consistency. Pls let me know if you see a problem.

kayibal: I think it depends on what joins you want to do and depending on that you have to declare relationship here.

tamaralipows: Ok, so far no joins on this objects so I think we are fine.

tycho-indexer/src/storage/mod.rs
```diff
@@ -252,6 +252,29 @@ pub trait StorableProtocolType<S, N, I>: Sized + Send + Sync + 'static {
...
+/// Generics:
+/// * `S`: This represents the storage-specific data type used when converting from storage to the
+///   transaction.
+/// * `N`: This represents the storage-specific data type used when converting from the protocol
+///   type to storage.
+/// * `I`: Represents the type of the database identifier, which is used as an argument in the
+///   conversion function. This facilitates the passage of database-specific foreign keys to the
+///   `to_storage` method, thereby providing a flexible way for different databases to interact with
+///   the transaction.
+pub trait StorableComponentBalance<S, N, I>: Sized + Send + Sync + 'static {
```
kayibal, line 267: I think you forgot to use the generics?

tycho-indexer/src/storage/mod.rs
```diff
@@ -252,6 +252,29 @@ pub trait StorableProtocolType<S, N, I>: Sized + Send + Sync + 'static {
     fn to_storage(&self) -> N;
 }
 
+/// Lays out the necessary interface needed to store and retrieve component balances
+/// from storage.
+///
+/// Generics:
+/// * `S`: This represents the storage-specific data type used when converting from storage to the
+///   transaction.
```
kayibal: copy pasta?

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -308,6 +308,55 @@ pub struct ProtocolType {
...
+    type SortKey = (i64, NaiveDateTime, i64);
+    type EntityId = i64;
+    type Version = NaiveDateTime;
+
+    fn get_entity_id(&self) -> Self::EntityId {
+        self.protocol_component_id
+    }
+
+    fn get_sort_key(&self) -> Self::SortKey {
+        (self.protocol_component_id, self.valid_from, self.modify_tx)
```
kayibal: I think the token id is missing here. Think about it we will basically group by the entities and then assign versions by increasing version (see module for clearer explanation) so basically this tuple is  `(entitiy_id, version)`

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -308,6 +308,55 @@ pub struct ProtocolType {
...
+    pub valid_to: Option<NaiveDateTime>,
+}
+
+impl VersionedRow for NewComponentBalance {
+    type SortKey = (i64, NaiveDateTime, i64);
+    type EntityId = i64;
+    type Version = NaiveDateTime;
+
+    fn get_entity_id(&self) -> Self::EntityId {
+        self.protocol_component_id
```
kayibal: Also here entity id is `(self.protocol_component_id, self.token_id)`

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -340,6 +389,22 @@ pub struct ProtocolComponent {
     pub deletion_tx: Option<i64>,
 }
 
+impl ProtocolComponent {
+    pub async fn id_by_external_id(
```
kayibal: Comment pls.

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -528,6 +529,67 @@ where
...
+        &self,
+        chain: Chain,
+        component_balances: &[&Self::ComponentBalance],
+        block_ts: NaiveDateTime,
+        conn: &mut Self::DB,
+    ) -> Result<(), StorageError> {
+        use super::schema::{account::dsl::*, token::dsl::*};
+        let token_addresses: Vec<Bytes> = component_balances
+            .iter()
+            .map(|component_balance| Bytes::from(component_balance.token.as_bytes()))
```
kayibal: Hmm again H160 ethereum specific type making it into here - it shouldn't. Probably need a method for this on the StorableTrait...

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -528,6 +529,67 @@ where
...
+            let new_component_balance = component_balance.to_storage(
+                current_account_id,
+                transaction_id,
+                protocol_component_id,
+                block_ts,
+            );
+            new_component_balances.push(new_component_balance);
+        }
+
+        diesel::insert_into(schema::component_balance::table)
```
kayibal: Shouldn't you be calling `apply_versioning` here on `new_component_balances`?

tycho-indexer/migrations/2024-01-10_tycho_patch/up.sql
```diff
@@ -61,3 +61,23 @@ DROP TRIGGER invalidate_previous_contract_code ON contract_code;
...
+    -- id of the token whose tvl changed
+    "token_id" bigint REFERENCES "token"(id) NOT NULL,
+    -- new balance of the token for this component
+    "new_balance" bytea NOT NULL,
+    -- the transaction that modified the tvl of this component
+    "modify_tx" bigint REFERENCES "transaction"(id) ON DELETE CASCADE NOT NULL,
+    -- Reference to static attributes of the protocol.
+    "protocol_component_id" bigint REFERENCES protocol_component(id) NOT NULL,
+    -- Timestamp this entry was inserted into this table.
+    "inserted_ts" timestamptz NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
SmartMonkey-git, line 76: Most tables also feature a `modified_ts`

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -528,6 +529,67 @@ where
...
+            .iter()
+            .map(|component_balance| Bytes::from(component_balance.modify_tx.as_bytes()))
+            .collect();
+        let transaction_ids = orm::Transaction::ids_by_hash(&transaction_hashes, conn).await?;
+
+        let external_ids: Vec<String> = component_balances
+            .iter()
+            .map(|component_balance| {
+                component_balance
+                    .component_id
```
SmartMonkey-git: I could see that `component_id` might lead to misunderstandings in the future, as it is not clear wether it is the tables id or the `external_id`.
I would suggest renaming it to `external_component_id`

tycho-indexer/src/storage/mod.rs
```diff
@@ -664,6 +693,26 @@ pub trait ProtocolGateway {
         conn: &mut Self::DB,
     ) -> Result<Vec<Self::Token>, StorageError>;
 
+    /// Saves multiple component balances to storage.
+    ///
+    /// Inserts token into storage. Tokens and their properties are assumed to
+    /// be immutable.
```
dianacarvalho1: ?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -1115,6 +1177,58 @@ mod test {
...
+    #[tokio::test]
+    async fn test_add_component_balances() {
+        let mut conn = setup_db().await;
+        setup_data(&mut conn).await;
+        let gw = EVMGateway::from_connection(&mut conn).await;
+
+        let tx_hash =
+            H256::from_str("0xbb7e16d797a9e2fbc537e30f91ed3d27a254dd9578aa4c3af3e5f0d3e8130945")
+                .unwrap();
+        let protocol_component_id: String = String::from("state1");
```
dianacarvalho1, line 1271: can we call it "pool ETH-USDC" or something? "state1" doesn't make much sense

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -528,6 +529,67 @@ where
...
+            .iter()
+            .map(|component_balance| Bytes::from(component_balance.token.as_bytes()))
+            .collect();
+
+        let account_ids = token
+            .inner_join(account)
+            .select(schema::token::id)
+            .filter(schema::account::address.eq_any(token_addresses))
+            .get_results::<i64>(conn)
+            .await?;
```
dianacarvalho1: what if this doesn't return the number of tokens you expect? it's gonna die later. same for the txs.
for the same txs we are gonna have at least two balance updates, will the returned transaction_ids be repeated? I don't really know tbh 🤔  
same for the component actually

tycho-indexer/src/storage/postgres/protocol.rs
tamaralipows: This is taken directly from the setup :/ not sure if there was a deeper motivation for that naming (it seems so, looking at the comments) @louise-poole

tycho-indexer/migrations/2024-01-10_tycho_patch/up.sql
tamaralipows: We won't ever be updating, so is that necessary?

tycho-indexer/src/storage/postgres/protocol.rs
tamaralipows: Hmmm but I think that's pretty clear by the fact that it is a string type? And this is the `ComponentBalance` object, which doesn't yet have access to the tables. Unless I misunderstood

tycho-indexer/src/storage/postgres/protocol.rs
tamaralipows: Oh, you mean that I may expect the same token to appear multiple times? This is true, I think I can't use a batch get query here :( :( :(

tycho-indexer/src/storage/postgres/protocol.rs
tamaralipows: I want to avoid having to add new components if I don't need to - wanted to keep this test simple.

tycho-indexer/migrations/2024-01-10_tycho_patch/up.sql
SmartMonkey-git: In that case, its not.

tycho-indexer/src/storage/postgres/protocol.rs
dianacarvalho1: huuummm I just find it misleading (and maybe a mistake of copy pasting).. it is a protocol component not a state, why should it be called state?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -528,6 +531,55 @@ where
...
+        &self,
+        chain: Chain,
+        component_balances: &[&Self::ComponentBalance],
+        block_ts: NaiveDateTime,
+        conn: &mut Self::DB,
+    ) -> Result<(), StorageError> {
+        use super::schema::{account::dsl::*, token::dsl::*};
+
+        let mut new_component_balances = Vec::new();
+        for component_balance in component_balances.iter() {
```
dianacarvalho1: ooh this could be a lot of queries 😕 
what if we did something like we did before but saved the results in a hashmap instead of looping through the vecs?

PR #103, opened 2024-01-24
ENG-2373: Implement Uniswap V2 Substreams

tvinagre:
Ok, this was a little harder than expected because it became very different than the original substream.

Please note that there are 2 packages that were copied from another repo: https://github.com/messari/substreams - feel free to skip them, I've just realized I can use them as a package and reduce this repeated code.

## Context

This PR implements an UniswapV2 substream to map the created pools and the reserve updates to our `Tycho's BlockEntityChanges` format.

Although Substreams previous knowledge is not required for reviewing this task, if you have any doubts about Substreams you can check this doc: https://substreams.streamingfast.io/documentation/develop
 

Task description: https://datarevenue.atlassian.net/browse/ENG-2373

### Questions
- `Attribute` value is `Vec<u8> (bytes)`. In this case the value is a BigNumber, should I convert it using plain bytes conversion or should I hex encode it before?
If not hex encoded, should we do little endian or big endian conversion?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,6 @@ thiserror = "1"
 tracing = "0.1.37"
 hyper = "0.14.27"
 async-trait = "0.1.73"
-anyhow = "1"
 tokio = { version = "1.27", features = [
     "time",
     "sync",


substreams/common/proto/erc20.proto
```diff
@@ -0,0 +1,38 @@
+syntax = "proto3";
```
tvinagre, line 1: For future reference, not needed rn, you can skip reviewing this one

substreams/common/src/codegen.rs
```diff
@@ -0,0 +1,234 @@
+use std::collections::{HashMap, HashSet};
```
tvinagre: Copied from: https://github.com/messari/substreams - no need to review

tvinagre: This files generate Rust code from ABI files

substreams/ethereum-uniswap-v2/src/abi/factory.rs
```diff
@@ -0,0 +1,904 @@
+    const INTERNAL_ERR: &'static str = "`ethabi_derive` internal error";
```
tvinagre, line 1: Generated code, can skip

substreams/ethereum-uniswap-v2/src/abi/pool.rs
```diff
@@ -0,0 +1,3554 @@
+    const INTERNAL_ERR: &'static str = "`ethabi_derive` internal error";
```
tvinagre, line 1: Generated code, can skip

substreams/ethereum-uniswap-v2/src/store_key.rs
```diff
@@ -0,0 +1,75 @@
...
+    TotalValueLockedUSD,
+    Volume,
+}
+
+impl StoreKey {
+    pub fn get_unique_pool_key(&self, key: &str) -> String {
+        format!("{}:{}", self.unique_id(), key)
+    }
+
+    pub fn get_unique_pair_key(&self, key1: &str, key2: &str) -> String {
```
tvinagre: I'll remove these useless keys tomorrow

substreams/substreams-helper/src/block.rs
```diff
@@ -0,0 +1,194 @@
+use std::ops::Add;
```
tvinagre: Also copied from https://github.com/messari/substreams. Won't clean bc a lot of methods can be useful later

substreams/substreams-helper/src/event_handler.rs
```diff
@@ -0,0 +1,108 @@
+use std::collections::HashMap;
```
tvinagre, line 1: Copied from https://github.com/messari/substreams - worth reading

substreams/substreams-helper/src/price.rs
```diff
@@ -0,0 +1,144 @@
+use std::ops::Div;
```
tvinagre: K this is useless, will remove

proto/tycho/evm/v1/entity.proto
```diff
@@ -30,3 +30,8 @@ message BlockEntityChanges {
   // The set of transaction changes observed in the specified block.
   repeated TransactionEntityChanges changes = 2;
 }
+
+message SameTypeTransactionContractChanges {
+  // The set of transaction changes observed in the specified block.
+  repeated TransactionEntityChanges changes = 1;
+}
```
louise-poole: What's this message for?

substreams/ethereum-uniswap-v2/src/modules/1_map_pool_created.rs
```diff
@@ -0,0 +1,42 @@
...
+pub fn map_pool_created(block: eth::Block) -> Result<Pools, substreams::errors::Error> {
+    let mut pools: Vec<Pool> = vec![];
+
+    get_pools(&block, &mut pools);
+    Ok(Pools { pools })
+}
+
+fn get_pools(block: &eth::Block, pools: &mut Vec<Pool>) {
+    // Extract new pools from PairCreated events
+    let mut on_pair_created = |event: PairCreated, _tx: &eth::TransactionTrace, _log: &eth::Log| {
```
louise-poole: Was there a purpose to splitting this into 2 functions? Looks like the only purpose of `map_pool_created` is to call `get_pools`?

louise-poole: Also since we are dealing with multiple pools maybe we can name it `map_pools_created`?

substreams/ethereum-uniswap-v2/src/modules/3_map_pool_events.rs
```diff
@@ -0,0 +1,77 @@
...
+    pools_store: StoreGetProto<Pool>,
+) -> Result<SameTypeTransactionContractChanges, substreams::errors::Error> {
+    let mut events = vec![];
+
+    // Sync event is sufficient for our use-case. Since it's emitted on
+    // every reserve-altering function call, we can use it as the only event
+    // to update the reserves of a pool.
+    handle_sync(&block, &pools_store, &mut events);
+
+    Ok(SameTypeTransactionContractChanges { changes: events })
```
louise-poole: Ah, is it some type of internal substream message? I'm hesitant on the name including the word "contract" as that is primarily used for vm messages. Could we just call it `SameTypeTransactionChanges`?

substreams/ethereum-uniswap-v2/src/modules/3_map_pool_events.rs
```diff
@@ -0,0 +1,77 @@
...
+fn handle_sync(
+    block: &eth::Block,
+    store: &StoreGetProto<Pool>,
+    events: &mut Vec<TransactionEntityChanges>,
+) {
+    let mut on_sync = |event: Sync, tx: &eth::TransactionTrace, log: &eth::Log| {
+        let pool_address = log.address.to_hex();
+
+        // Convert reserves to hex
+        // TODO: Verify this conversion. Value needs to be bytes, should it be simple conversion or hex?
```
louise-poole: Good question - I initially assumed simple conversion, but I'm not actually sure. On the extractor/db side it doesn't matter, so it's down to what we want to send to the clients that subscribe to us. @kayibal do you have a preference?

substreams/ethereum-uniswap-v2/src/modules/4_extract_protocol_components.rs
```diff
@@ -0,0 +1,81 @@
...
+                        Attribute {
+                            name: "reserve1".to_string(),
+                            value: BigInt::zero().to_signed_bytes_le(),
+                            change: i32::from(ChangeType::Creation),
+                        },
+                    ],
+                    implementation_type: i32::from(ImplementationType::Custom),
+                }),
+            }],
+            balance_changes: vec![],
```
dianacarvalho1: you are not going to implement balance changes? 🥺

substreams/ethereum-uniswap-v2/src/modules/4_extract_protocol_components.rs
```diff
@@ -0,0 +1,81 @@
...
+                        name: "reserve0".to_string(),
+                        value: BigInt::zero().to_signed_bytes_le(),
+                        change: i32::from(ChangeType::Creation),
+                    },
+                    Attribute {
+                        name: "reserve1".to_string(),
+                        value: BigInt::zero().to_signed_bytes_le(),
+                        change: i32::from(ChangeType::Creation),
+                    },
+                ],
```
louise-poole: `reserve0` and `reserve1` are dynamic attributes (they are expected to change on events) so will belong to ProtocolStates (entity_changes message) and not to ProtocolComponent. Here we need static attributes only. For usv2 I think only `trading_fee` would apply here?

substreams/ethereum-uniswap-v2/src/modules/5_block_sink.rs
```diff
@@ -0,0 +1,29 @@
+use substreams_ethereum::pb::eth::v2::{self as eth};
+
+use crate::pb::tycho::evm::v1::{
+    Block, BlockEntityChanges, SameTypeTransactionContractChanges, TransactionEntityChanges,
+};
+
+#[substreams::handlers::map]
+pub fn block_sink(
+    block: eth::Block,
+    balance_changes: SameTypeTransactionContractChanges,
```
louise-poole: Is this not meant to `state_changes` or `entity_changes` maybe?

substreams/ethereum-uniswap-v2/substreams.yaml
```diff
@@ -0,0 +1,62 @@
...
+    kind: map
+    initialBlock: 10008300
+    inputs:
+      - source: sf.ethereum.type.v2.Block
+    output:
+      type: proto:tycho.evm.uniswap.v2.Pools
+
+  - name: store_pools
+    kind: store
+    updatePolicy: set_if_not_exists
```
dianacarvalho1, line 32: I think you need an initialBlock in the store too otherwise it will try to get aaalll the blocks ever

substreams/ethereum-uniswap-v2/substreams.yaml
```diff
@@ -0,0 +1,62 @@
...
+    output:
+      type: proto:tycho.evm.v1.SameTypeTransactionContractChanges
+
+  - name: block_sink
+    kind: map
+    initialBlock: 10008300
+    inputs:
+      - source: sf.ethereum.type.v2.Block
+      - map: extract_protocol_components
+      - map:
```
dianacarvalho1: ?

dianacarvalho1: is it the `map_pool_events` here?

substreams/ethereum-uniswap-v2/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+substreams-helper = { path = "../substreams-helper" }
+num-bigint = "0.4.4"
+
+[target.wasm32-unknown-unknown.dependencies]
+getrandom = { version = "0.2", features = ["custom"] }
+
+[build-dependencies]
+anyhow = "1"
+substreams-ethereum = "0.9"
+substreams-common = { path = "../common" }
```
zizou0x: Why not use the one defined in the workspace here?

substreams/ethereum-uniswap-v2/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "substreams-ethereum-uniswap-v2"
+version = "0.1.0"
+edition = "2021"
```
zizou0x, line 4:
```suggestion
edition = "2024"
```

substreams/ethereum-uniswap-v2/src/modules/1_map_pool_created.rs
```diff
@@ -0,0 +1,42 @@
...
+            address: Vec::from(event.pair.as_slice()),
+            token0: Vec::from(event.token0.as_slice()),
+            token1: Vec::from(event.token1.as_slice()),
+            created_timestamp: block.timestamp_seconds() as i64,
+            created_block_number: block.number as i64,
+            created_tx_hash: block.hash.clone(),
+        })
+    };
+
+    let mut eh = EventHandler::new(&block);
```
zizou0x: You're such a Go developer https://go.dev/wiki/CodeReviewComments#variable-names

substreams/ethereum-uniswap-v2/src/modules/1_map_pool_created.rs
```diff
@@ -0,0 +1,42 @@
...
+
+    // TODO: Parametrize Factory Address
+    eh.filter_by_address(vec![
+        Address::from_str("0x5c69bee701ef814a2b6a3edd4b1652cb9cc5aa6f").unwrap()
+    ]);
+
+    eh.on::<PairCreated, _>(&mut on_pair_created);
```
zizou0x, line 69: Really nice

substreams/ethereum-uniswap-v2/src/modules/2_store_pools.rs
```diff
@@ -0,0 +1,21 @@
...
+use substreams_helper::hex::Hexable;
+
+use crate::{
+    pb::tycho::evm::uniswap::v2::{Pool, Pools},
+    store_key::StoreKey,
+};
+
+#[substreams::handlers::store]
+pub fn store_pools(pools_created: Pools, store: StoreSetIfNotExistsProto<Pool>) {
+    // Store pools. Required so the next maps can match any event to an known pool by their address
```
dianacarvalho1:
```suggestion
    // Store pools. Required so the next maps can match any event to a known pool by their address
```

substreams/ethereum-uniswap-v2/src/modules/2_store_pools.rs
```diff
@@ -0,0 +1,21 @@
...
+
+    for pool in pools_created.pools {
+        //  Use ordinal 0 because the address should be unique, so ordering doesn't matter.
+        let pool_address: &str = &pool.address.to_hex();
+        store.set_if_not_exists(0, StoreKey::Pool.get_unique_pool_key(pool_address), &pool);
```
dianacarvalho1: why is the set_if_not_exists necessary? do you expect the same pool to appear here more than once?

substreams/ethereum-uniswap-v2/src/modules/3_map_pool_events.rs
kayibal: Not sure what a hex conversion would be? Encode as string with hex encoding and then cast that utf-8 string to bytes? I think I prefer to go to bytes directly.

substreams/ethereum-uniswap-v2/src/modules/3_map_pool_events.rs
```diff
@@ -0,0 +1,77 @@
...
+        // Events
+        events.push(TransactionEntityChanges {
+            tx: Option::from(tycho_tx),
+            entity_changes: vec![EntityChanges {
+                component_id: pool_address.clone(),
+                attributes: vec![
+                    Attribute {
+                        name: "reserve0".to_string(),
+                        value: reserve0,
+                        change: i32::from(ChangeType::Update),
```
kayibal:
```suggestion
                        change: ChangeType::Update.into(),
```

substreams/ethereum-uniswap-v2/src/modules/4_extract_protocol_components.rs
```diff
@@ -0,0 +1,81 @@
...
+                    Attribute {
+                        name: "reserve0".to_string(),
+                        value: BigInt::zero().to_signed_bytes_le(),
+                        change: i32::from(ChangeType::Creation),
+                    },
+                    Attribute {
+                        name: "reserve1".to_string(),
+                        value: BigInt::zero().to_signed_bytes_le(),
+                        change: i32::from(ChangeType::Creation),
+                    },
```
kayibal: Reserves are not static attributes, but pool fee is.

substreams/ethereum-uniswap-v2/src/modules/4_extract_protocol_components.rs
```diff
@@ -0,0 +1,81 @@
...
+                        Attribute {
+                            name: "reserve0".to_string(),
+                            value: BigInt::zero().to_signed_bytes_le(),
+                            change: i32::from(ChangeType::Creation),
+                        },
+                        Attribute {
+                            name: "reserve1".to_string(),
+                            value: BigInt::zero().to_signed_bytes_le(),
+                            change: i32::from(ChangeType::Creation),
+                        },
```
kayibal: This is to validate the bytes coming in for the static attributes. So I think this should not be set by substreams but we'll have to set these in a config file.

substreams/ethereum-uniswap-v2/src/modules/4_extract_protocol_components.rs
```diff
@@ -0,0 +1,81 @@
...
+pub fn extract_protocol_components(
+    block: eth::Block,
+    pools_created: Pools,
+) -> Result<SameTypeTransactionContractChanges, substreams::errors::Error> {
+    let mut tx_entity_changes: Vec<TransactionEntityChanges> = vec![];
+
+    handle_created_pools(block, pools_created, &mut tx_entity_changes);
+
+    Ok(SameTypeTransactionContractChanges { changes: tx_entity_changes })
+}
```
kayibal: To be honest I have doubts about the structure of this substream:

Since you went quite custom anyway - maybe we should already instantiate the `ProtocolComponents` struct in `map_pools_created` and then store components right away in the store in `store_pools`, next we can check in `map_pool_events` for any sync log received, if the corresponding component was created in this block, and if it was, emit in the message.

As for balance changes they are simply emitted here with every message as each sync means a balance change.

proto/tycho/evm/v1/entity.proto
tvinagre: Substreams doesn't allow a return type to be a vector, needs to be inside a single struct, so I just create a single-field struct in which the element is a vector of TransactionEntityChanges

substreams/ethereum-uniswap-v2/src/modules/1_map_pool_created.rs
tvinagre: No specific reason tbh, I just followed the suggested implementation and I think it looks cleaner this way tbh, do you want me to change?

substreams/ethereum-uniswap-v2/src/modules/3_map_pool_events.rs
tvinagre: Oh yeah this was an error - I've based the name from the wrong buf file. Thanks!

substreams/ethereum-uniswap-v2/src/modules/4_extract_protocol_components.rs
tvinagre: nice! Will change :)

substreams/ethereum-uniswap-v2/src/modules/4_extract_protocol_components.rs
tvinagre: hmm was not in the plan tbh - what's it needed for?

substreams/ethereum-uniswap-v2/substreams.yaml
tvinagre: somehow it worked without it in the store but you're right, will add, nice catch! Thanks

substreams/ethereum-uniswap-v2/substreams.yaml
tvinagre: ouais, I think I've deleted it by mistake lol, fixed

substreams/ethereum-uniswap-v2/src/modules/2_store_pools.rs
tvinagre: Hmm no I don't tbh, this one I just followed the example and didn't question, will check

substreams/ethereum-uniswap-v2/src/modules/4_extract_protocol_components.rs
tvinagre: Sorry I don't understand - should I change it to an empty vec?

substreams/ethereum-uniswap-v2/Cargo.toml
tvinagre: That was a mistake that will be corrected :) tnx

substreams/ethereum-uniswap-v2/src/modules/1_map_pool_created.rs
tvinagre: hehehehe

substreams/ethereum-uniswap-v2/src/modules/4_extract_protocol_components.rs
tvinagre: Yeah I guess we could simplify it, I just wanted to keep the store simple.

But I don't understand this part:
> next we can check in map_pool_events for any sync log received, if the corresponding component was created in this block, and if it was, emit in the message.

Why is it relevant if this component was created in this block? If the TXs are different, than it should be in different TransactionEntityChanges, no? Or you mean to aggregate if, by any chance, there's a TX where you create a pool and add liquidity (in the same transaction)

Also, I thought `balance changes` meant only for native tokens, I guess I've misunderstood. But shouldn't they be emitted in `map_pool_events` instead of `extract_protocol_components` ?

substreams/ethereum-uniswap-v2/src/modules/4_extract_protocol_components.rs
dianacarvalho1: For the TVL tracking

substreams/ethereum-uniswap-v2/src/modules/4_extract_protocol_components.rs
kayibal: Yes please

substreams/ethereum-uniswap-v2/src/modules/4_extract_protocol_components.rs
kayibal: > Why is it relevant if this component was created in this block? If the TXs are different, than it should be in different TransactionEntityChanges, no? Or you mean to aggregate if, by any chance, there's a TX where you create a pool and add liquidity (in the same transaction)

Indeed you are right here, but at the same time we would expect all changes to be grouped by transaction. As you can see [here](https://github.com/propeller-heads/tycho-indexer/blob/main/proto/tycho/evm/v1/entity.proto#L17) All changes of a single tx are emitted as one message. Is this the case here? It seems that if someone deploys a pool and adds liquidity in the same tx we might get two `TransactionEntityChanges` messages.
In short: Yes I mean that these messages should be aggregated.

About balances: We need to know tvl for each protocol components to we need to emit both erc20 and native balances with each transaction that changes those. As v2 is ERC20 only, the entries there will basically be a duplicate of the reserves.

tvinagre: ah nice, will add them - thanks!

substreams/ethereum-uniswap-v2/src/modules/4_extract_protocol_components.rs
tvinagre: Aaaaadded

substreams/ethereum-uniswap-v2/src/modules/2_store_pools.rs
tvinagre: Doesn't make any diff:
https://substreams.streamingfast.io/documentation/develop/manifest-modules/types#updatepolicy-property

substreams/ethereum-uniswap-v2/src/modules/1_map_pool_created.rs
louise-poole: If I remember correctly, these "pass through" functions were listed as a red flag in the software design book we did in book club. It's adding to the interface of the class without adding functionality (makes a shallow class). If we are to follow the design principles of the book, we should avoid these 👀

Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-members = ["tycho-indexer", "substreams/ethereum-ambient"]
+members = ["tycho-indexer", "substreams/common", "substreams/substreams-helper", "substreams/ethereum-ambient", "substreams/ethereum-uniswap-v2"]
```
kayibal: Once you start on uniswap-v3 let's add it directly to propeller protocol lib?

proto/tycho/evm/v1/entity.proto
```diff
@@ -30,3 +30,8 @@ message BlockEntityChanges {
   // The set of transaction changes observed in the specified block.
   repeated TransactionEntityChanges changes = 2;
 }
+
+message SameTypeTransactionChanges {
+  // The set of transaction changes observed in the specified block.
+  repeated TransactionEntityChanges changes = 1;
+}
```
kayibal: Why is this here without a comment? And what is this? It looks superfluous, you know that all fields on protobuf models are optional? Maybe move to uniswap-v2 models if you want to keep it.

substreams/ethereum-uniswap-v2/src/modules/1_map_pool_created.rs
```diff
@@ -0,0 +1,68 @@
...
+    Ok(SameTypeTransactionChanges { changes: new_pools })
+}
+
+fn get_pools(block: &eth::Block, new_pools: &mut Vec<TransactionEntityChanges>) {
+    // Extract new pools from PairCreated events
+    let mut on_pair_created = |event: PairCreated, _tx: &eth::TransactionTrace, _log: &eth::Log| {
+        let tycho_tx: Transaction = _tx.into();
+
+        new_pools.push(TransactionEntityChanges {
+            tx: Option::from(tycho_tx),
```
kayibal: What does this do? Do you mean `Some(tycho_tx)` ?

substreams/ethereum-uniswap-v2/src/modules/2_store_pools.rs
```diff
@@ -0,0 +1,28 @@
...
+    pb::tycho::evm::{uniswap::v2::Pool, v1::SameTypeTransactionChanges},
+    store_key::StoreKey,
+};
+
+#[substreams::handlers::store]
+pub fn store_pools(
+    pools_created: SameTypeTransactionChanges,
+    store: StoreSetIfNotExistsProto<Pool>,
+) {
+    // Store pools. Required so the next maps can match any event to a known pool by their address
```
kayibal: you mean next step?

substreams/ethereum-uniswap-v2/src/modules/3_map_pool_events.rs
```diff
@@ -0,0 +1,127 @@
...
+        },
+    },
+    store_key::StoreKey,
+    traits::PoolAddresser,
+};
+
+#[substreams::handlers::map]
+pub fn map_pool_events(
+    block: eth::Block,
+    created_pairs: SameTypeTransactionChanges,
```
kayibal: Feels weird to see this type with this variable name? I think you can get the deltas from the store, this gives you the keys that changed in this block. See playgrounds balancer implementation.

substreams/ethereum-uniswap-v2/src/modules/3_map_pool_events.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    handle_sync(&block, &mut created_pairs.clone(), &mut tx_changes_map, &pools_store);
+
+    // Make a list of all HashMap values:
+    let tx_entity_changes: Vec<TransactionEntityChanges> = tx_changes_map.into_values().collect();
+
+    let tycho_block: Block = block.into();
+
+    let block_entity_changes =
+        BlockEntityChanges { block: Option::from(tycho_block), changes: tx_entity_changes };
```
kayibal: Brrrr pls use `Some(tycho_block)` this hurts my eyes :D

substreams/ethereum-uniswap-v2/src/modules/2_store_pools.rs
```diff
@@ -0,0 +1,28 @@
...
+
+use crate::{
+    pb::tycho::evm::{uniswap::v2::Pool, v1::SameTypeTransactionChanges},
+    store_key::StoreKey,
+};
+
+#[substreams::handlers::store]
+pub fn store_pools(
+    pools_created: SameTypeTransactionChanges,
+    store: StoreSetIfNotExistsProto<Pool>,
```
kayibal: I'd probably be storing ProtocolComponents right away...

substreams/ethereum-uniswap-v2/src/modules/3_map_pool_events.rs
```diff
@@ -0,0 +1,127 @@
...
+        let transaction = change.tx.as_ref().unwrap(); // get reference instead of value
+        tx_changes_map.insert(transaction.hash.clone(), change.clone()); // clone because we're
+                                                                         // working with references
+    }
+
+    let mut on_sync = |event: Sync, _tx: &eth::TransactionTrace, _log: &eth::Log| {
+        let pool_address = _log.address.to_hex();
+
+        let pool = store.must_get_last(StoreKey::Pool.get_unique_pool_key(pool_address.as_str()));
+        // Convert reserves to hex
```
kayibal: you mean to bytes

substreams/ethereum-uniswap-v2/src/store_key.rs
```diff
@@ -0,0 +1,16 @@
...
+    pub fn get_unique_pool_key(&self, key: &str) -> String {
+        format!("{}:{}", self.unique_id(), key)
+    }
+
+    pub fn unique_id(&self) -> String {
+        match self {
+            StoreKey::Pool => "Pool".to_string(),
+        }
+    }
+}
```
kayibal, line 16: Does this do anything useful?

substreams/ethereum-uniswap-v2/src/traits.rs
```diff
@@ -0,0 +1,60 @@
...
+            .store
+            .get_last(StoreKey::Pool.get_unique_pool_key(&key.to_hex()));
+
+        pool.is_some()
+    }
+}
+
+impl<'a> HasAddresser for PoolAddresser<'a> {
+    fn has_address(&self, key: Address) -> bool {
+        self.has_address(key)
```
kayibal: inline aboves implementation here?

substreams/ethereum-uniswap-v2/src/modules/1_map_pool_created.rs
tvinagre: Ah yeah - blame copilot autocomplete sorry

proto/tycho/evm/v1/entity.proto
tvinagre: This is here because you can't return a vector from Substreams - it always needs to be incapsulated.
I've added here instead of USv2 models because it shall be used on every native implementation

tvinagre: I agree I should add a comment

substreams/ethereum-uniswap-v2/src/traits.rs
tvinagre: Yeah I just copied this one but they could be merged, tnx

substreams/ethereum-uniswap-v2/src/store_key.rs
tvinagre: Just keeps keys organized - and serves as a reference for more complex implementations (like USv3 you might need)

substreams/ethereum-uniswap-v2/src/modules/2_store_pools.rs
tvinagre: Hmm yeah Idk why I didn't - I wanted to keep it small but PCs are not much bigger so I'll make the change

proto/tycho/evm/v1/entity.proto
kayibal: I think you can just use `TransactionChanges`, no?

substreams/ethereum-uniswap-v2/src/modules/3_map_pool_events.rs
```diff
@@ -0,0 +1,169 @@
...
+    },
+    store_key::StoreKey,
+    traits::PoolAddresser,
+};
+
+#[derive(Clone, Hash, Eq, PartialEq)]
+struct ComponentKey<T> {
+    component_id: String,
+    name: T,
+}
```
kayibal, line 26: It took me quite long to figure out that this is supposed to work as a tuple. in HashMap.

substreams/ethereum-uniswap-v2/src/modules/3_map_pool_events.rs
```diff
@@ -0,0 +1,169 @@
...
+
+#[substreams::handlers::map]
+pub fn map_pool_events(
+    block: eth::Block,
+    block_entity_changes: BlockEntityChanges,
+    pools_store: StoreGetProto<ProtocolComponent>,
+) -> Result<BlockEntityChanges, substreams::errors::Error> {
+    // Sync event is sufficient for our use-case. Since it's emitted on every reserve-altering
+    // function call, we can use it as the only event to update the reserves of a pool.
+    let mut block_entity_changes = block_entity_changes;
```
kayibal, line 67: Unnecessary reassign? I think you can declare it as mut in the functions sig

substreams/ethereum-uniswap-v2/src/modules/3_map_pool_events.rs
```diff
@@ -0,0 +1,169 @@
...
+                    token: token.clone(),
+                    balance: balance.clone(),
+                    component_id: pool_address_hex.as_bytes().to_vec(),
+                },
+            );
+        }
+    };
+
+    let mut eh = EventHandler::new(block);
+    eh.filter_by_address(PoolAddresser { store });
```
kayibal, line 140: If we don't have this we may get false positives right? Maybe worth adding a comment here.

substreams/ethereum-uniswap-v2/src/modules/3_map_pool_events.rs
```diff
@@ -0,0 +1,169 @@
...
+    let mut block_entity_changes = block_entity_changes;
+    let mut tx_changes: HashMap<Vec<u8>, PartialChanges> = HashMap::new();
+
+    handle_sync(&block, &mut tx_changes, &pools_store);
+    merge_block(&mut tx_changes, &mut block_entity_changes);
+
+    Ok(block_entity_changes)
+}
+
+fn handle_sync(
```
kayibal, line 86: Please add docstrings to functions

substreams/ethereum-uniswap-v2/src/modules/3_map_pool_events.rs
```diff
@@ -0,0 +1,169 @@
...
+
+fn merge_block(
+    tx_changes: &mut HashMap<Vec<u8>, PartialChanges>,
+    block_entity_changes: &mut BlockEntityChanges,
+) {
+    // First, iterate through the previously created transactions, extracted from the
+    // map_pool_created step. If there are sync events for this transaction, add them to the
+    // block_entity_changes and the corresponding balance changes.
+    for change in block_entity_changes.changes.iter_mut() {
+        let tx = change.tx.as_mut().unwrap();
```
kayibal: Why does the transaction need to be mutable? Why is the unwrap ok here? Can you use expect instead please?

substreams/ethereum-uniswap-v2/src/modules/3_map_pool_events.rs
```diff
@@ -0,0 +1,169 @@
...
+
+    let mut eh = EventHandler::new(block);
+    eh.filter_by_address(PoolAddresser { store });
+    eh.on::<Sync, _>(&mut on_sync);
+    eh.handle_events();
+}
+
+fn merge_block(
+    tx_changes: &mut HashMap<Vec<u8>, PartialChanges>,
+    block_entity_changes: &mut BlockEntityChanges,
```
kayibal, line 160: Hard to follow here by the names what contains what... So BlockEntityChanges contain only new components and the `PartialChanges` contain attribute and balance changes, right?

substreams/ethereum-uniswap-v2/src/modules/3_map_pool_events.rs
```diff
@@ -0,0 +1,169 @@
...
+                    .clone()
+                    .consolidate_entity_changes(),
+                balance_changes: partial_changes
+                    .balance_changes
+                    .clone()
+                    .into_values()
+                    .collect(),
+                component_changes: vec![],
+            });
+    }
```
kayibal, line 203: Ok and these are the changes that had not component created. I see.. comment would be nice here too I was like wtf why is he iterating twice..

substreams/ethereum-uniswap-v2/src/modules/3_map_pool_events.rs
```diff
@@ -0,0 +1,169 @@
...
+impl PartialChanges {
+    fn consolidate_entity_changes(self) -> Vec<EntityChanges> {
+        let mut entity_changes_map: HashMap<String, Vec<Attribute>> = HashMap::new();
+
+        for (key, attribute) in self.entity_changes {
+            entity_changes_map
+                .entry(key.component_id)
+                .or_default()
+                .push(attribute);
+        }
```
kayibal: Ah ok weird af way to do a groupby :D. 
Literally took me 5m to figure out this is a groupby. You know iterators have a `groupby` method? Docstring would also have helped.

substreams/ethereum-uniswap-v2/src/modules/3_map_pool_events.rs
tvinagre: Yes, but that also why makes it inefficient to merge `pool_creation` transactions and `sync` transactions. We don't have the same subset of TXs to work with.

I'll add a comment but I will remove this in a future PR hopefully, when I refactor this

substreams/ethereum-uniswap-v2/src/modules/3_map_pool_events.rs
tvinagre: Hmm I think I've added that for a reason - I've tried to add as a mut before and it didn't work - will check later

substreams/ethereum-uniswap-v2/src/modules/3_map_pool_events.rs
tvinagre: Because it comes from change, which is a mutable reference (we update it in the next line)

tvinagre: But cloning seems to work

PR #104, opened 2024-01-24
ENG-2343: Add OpenAPI for Tycho RPC

zizou0x:
This PR aims to add OpenAPI support for Tycho RPC. We use the [utoipa](https://docs.rs/utoipa/latest/utoipa/#traits) crate for this.

Run locally:

1. Run Tycho and get some data in your local Docker DB (for this part, refer to the readme)
2. Run `cargo run --example rpc`
3. Visit http://0.0.0.0:4242/docs/#/

### DIFF
--- a/tycho-indexer/src/extractor/evm/mod.rs
+++ b/tycho-indexer/src/extractor/evm/mod.rs
@@ -11,7 +11,6 @@ use std::{
     ops::Deref,
 };
 use tracing::warn;
-use utoipa::ToSchema;
 
 use utils::{pad_and_parse_32bytes, pad_and_parse_h160};
 
@@ -85,25 +84,17 @@ impl Transaction {
     }
 }
 
-#[derive(PartialEq, Debug, Clone, Serialize, Deserialize, ToSchema)]
+#[derive(PartialEq, Debug, Clone)]
 pub struct Account {
     pub chain: Chain,
-    #[schema(value_type=String)]
     pub address: H160,
     pub title: String,
-    #[schema(value_type=HashMap<String, String>)]
     pub slots: HashMap<U256, U256>,
-    #[schema(value_type=String)]
     pub balance: U256,
-    #[schema(value_type=String)]
     pub code: Bytes,
-    #[schema(value_type=String)]
     pub code_hash: H256,
-    #[schema(value_type=String)]
     pub balance_modify_tx: H256,
-    #[schema(value_type=String)]
     pub code_modify_tx: H256,
-    #[schema(value_type=Option<String>)]
     pub creation_tx: Option<H256>,
 }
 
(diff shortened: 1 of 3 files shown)


tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -85,17 +85,25 @@ impl Transaction {
...
 }
 
-#[derive(PartialEq, Debug, Clone, Serialize, Deserialize)]
+#[derive(PartialEq, Debug, Clone, Serialize, Deserialize, ToSchema)]
 pub struct Account {
     pub chain: Chain,
+    #[schema(value_type=String)]
     pub address: H160,
     pub title: String,
+    #[schema(value_type=HashMap<String, String>)]
```
kayibal: Hm I dislike having to import utopia so deep in the core. Could you instead just copy this struct into RPC and call it EVMAccount? Then implement `From<evm::Account>` trait on it there? This way RPC stuff stays nicely separated.

tycho-indexer/src/services/rpc.rs
```diff
@@ -15,20 +15,70 @@ use diesel_async::{
...
+    pub code_modify_tx: String,
+    pub creation_tx: Option<String>,
+}
+
+fn trim_hex(hex_number: &str) -> String {
+    let trimmed = hex_number
+        .trim_start_matches("0x")
+        .trim_start_matches('0');
+    if trimmed.is_empty() {
+        "0x0".to_string()
```
kayibal: hex must always be at least `0x00`

tycho-indexer/src/services/rpc.rs
```diff
@@ -15,20 +15,70 @@ use diesel_async::{
...
+            chain: account.chain,
+            address: trim_hex(&account.address.encode_hex()),
+            title: account.title,
+            slots: account
+                .slots
+                .into_iter()
+                .map(|(key, value)| (trim_hex(&key.encode_hex()), trim_hex(&value.encode_hex())))
+                .collect(),
+            balance: trim_hex(&account.balance.encode_hex()),
+            code: account.code.to_string(),
```
kayibal: Why is this str while everything else is hex? Is it already hex encoded?

zizou0x: I've changed it to match the other struct types. I think it's prettier and less confusing.
 
But to give you an answer, this one is a `Bytes` type and `.to_string()` converts it directly to a hex encoded string yes. For the others which are ethers type (like U256, H256..) `.to_string()` converts them to decimal encoded string so I had to add and extra step to convert them to hex using `.encode_hex()`

tycho-indexer/src/services/rpc.rs
zizou0x: Do you mean for the `u8` conversion later? I've compared it to what we currently have and it seems it's fine with only `0x0` for example on the current Tycho we have this slot `"0x0": "0x100000000000000000000000000000000000000000000"`. Do you think it's gonna be problematic in the future?

PR #106, opened 2024-01-25
Use repeatable read transaction isolation.

kayibal:
This transaction type will ensure that our state remaing consistent with application side versioning.

Properly testing this isolation mode was impossible using the test transaction. To work around this I added a new kind of db test that will purge all data after it finished running. This test type can be used for tests that require actually committing data.

### DIFF
--- a/tycho-indexer/src/storage/postgres/cache.rs
+++ b/tycho-indexer/src/storage/postgres/cache.rs
@@ -41,7 +41,7 @@ pub(crate) enum WriteOp {
 
 /// Represents a transaction in the database, including the block information,
 /// a list of operations to be performed, and a channel to send the result.
-pub(crate) struct DBTransaction {
+pub struct DBTransaction {
     block: evm::Block,
     operations: Vec<WriteOp>,
     tx: oneshot::Sender<Result<(), StorageError>>,
@@ -58,7 +58,7 @@ impl DBTransaction {
 }
 
 /// Represents different types of messages that can be sent to the DBCacheWriteExecutor.
-pub(crate) enum DBCacheMessage {
+pub enum DBCacheMessage {
     Write(DBTransaction),
     Flush(oneshot::Sender<Result<(), StorageError>>),
     Revert(BlockIdentifier, oneshot::Sender<Result<(), StorageError>>),
@@ -97,7 +97,7 @@ pub(crate) enum DBCacheMessage {
 /// Read Operations
 /// The class does provide read operations for completeness, but it will not consider any
 /// cached changes while reading. Any reads are direct pass through to the database.
-pub(crate) struct DBCacheWriteExecutor {
+pub struct DBCacheWriteExecutor {
     name: String,
     chain: Chain,
     pool: Pool<AsyncPgConnection>,
@@ -114,7 +114,7 @@ pub(crate) struct DBCacheWriteExecutor {
 }
 
 impl DBCacheWriteExecutor {
-    pub(crate) fn new(
+    pub fn new(
         name: String,
         chain: Chain,
         pool: Pool<AsyncPgConnection>,
--- a/tycho-indexer/examples/rpc.rs
+++ b/tycho-indexer/examples/rpc.rs
@@ -2,15 +2,13 @@
 /// Usage: cargo run --example rpc
 use futures03::future::select_all;
 
-use crate::{
+use actix_web::dev::ServerHandle;
+use tracing::info;
+use tycho_indexer::{
     extractor::{evm, ExtractionError},
     services::ServicesBuilder,
     storage::postgres::{self, PostgresGateway},
 };
-use actix_web::dev::ServerHandle;
-use tracing::info;
-
-use tycho_indexer::{extractor, services, storage};
 
 #[tokio::main]
 async fn main() -> Result<(), ExtractionError> {
--- a/tycho-indexer/src/lib.rs
+++ b/tycho-indexer/src/lib.rs
@@ -6,3 +6,7 @@ pub mod serde_helpers;
 pub mod services;
 pub mod storage;
 pub mod substreams;
+
+#[cfg(test)]
+#[macro_use]
+extern crate pretty_assertions;
--- a/tycho-indexer/src/main.rs
+++ b/tycho-indexer/src/main.rs
@@ -9,29 +9,20 @@ use extractor::{
 };
 use models::Chain;
 
-use crate::{
-    extractor::{evm, ExtractionError},
-    services::ServicesBuilder,
-    storage::postgres::{self, cache::CachedGateway, PostgresGateway},
-};
 use actix_web::dev::ServerHandle;
 use clap::Parser;
 use std::sync::Arc;
 use tokio::{sync::mpsc, task, task::JoinHandle};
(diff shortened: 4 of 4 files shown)


tycho-indexer/src/storage/postgres/mod.rs
```diff
@@ -514,6 +514,112 @@ fn run_migrations(db_url: &str) {
...
+    ///
+    /// ## Interference with other tests
+    /// While this function runs, the db will actually contain data. This means two things:
+    ///
+    ///     1. It is likely to interfere with other tests using this same function. To mitigate
+    ///        this, you can use the `#[serial]` macro which will execute these tests sequentially.
+    ///     2. Other tests that rely on a empty db (most tests unsing test transactions) will likely
+    ///        be affected if run in parrallel with this function. To keep running these tests in
+    ///        parallel but avoid running them during the sequential tests use the macro
+    ///        #[parallel].
```
kayibal: We'll need to update everyone about this. I am thinking of writing a db tests section in the readme for this...

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -191,9 +190,12 @@ impl DBCacheWriteExecutor {
                         .expect("pool should be connected");
 
                     let res = conn
-                        .transaction(|conn| {
+                        .build_transaction()
+                        .repeatable_read()
+                        .run(|conn| {
                             async {
                                 for op in new_db_tx.operations {
+                                    dbg!("WRITE", &op);
```
zizou0x:
```suggestion
```

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -31,7 +30,7 @@ use crate::{
 };
 
 /// Represents different types of database write operations.
-#[derive(PartialEq, Clone)]
+#[derive(PartialEq, Clone, Debug)]
```
zizou0x, line 33:
```suggestion
#[derive(PartialEq, Clone)]
```

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -267,27 +269,30 @@ impl DBCacheWriteExecutor {
...
-                            }
-                        };
+        conn.build_transaction()
+            .repeatable_read()
+            .run(|conn| {
+                async {
+                    for op in db_txs.into_iter() {
+                        if !seen_operations.contains(&op) {
+                            // Only executes if it is not already in seen_operations
+                            dbg!("FLUSH", &op);
```
zizou0x:
```suggestion
```

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -191,9 +190,12 @@ impl DBCacheWriteExecutor {
                         .expect("pool should be connected");
 
                     let res = conn
-                        .transaction(|conn| {
+                        .build_transaction()
+                        .repeatable_read()
```
zizou0x, line 194: Why do we need this in that case?

tycho-indexer/src/storage/postgres/mod.rs
```diff
@@ -514,6 +514,112 @@ fn run_migrations(db_url: &str) {
...
+                .await
+                .expect(&format!("Error truncating {} table", t));
+        }
+        dbg!("Teardown completed");
+    }
+
+    /// Run tests that require committing data to the db.
+    ///
+    /// This function will run tests that are expected to commit data into the database, e.g.
+    /// because the test setups are to complex for using `begin_test_transaction`. Please only use
```
zizou0x:
```suggestion
    /// because the test setups are too complex for using `begin_test_transaction`. Please only use
```

tycho-indexer/src/storage/postgres/mod.rs
```diff
@@ -514,6 +514,112 @@ fn run_migrations(db_url: &str) {
...
+    /// this as a last resort as these tests are slow and have to be run serially. Using a test
+    /// transaction is preferred were possible.
```
zizou0x:
```suggestion
    /// transaction is preferred where possible.
```

tycho-indexer/src/storage/postgres/cache.rs
kayibal: This would also potentially update versions, no? I mean ideally the repeatable reads are not needed, they are just additional validation that protects us against application side bugs (concurrent writes to the same rows).

tycho-indexer/src/storage/postgres/contract_state.rs
```diff
@@ -1987,6 +1987,7 @@ mod test {
     }
 
     #[tokio::test]
+
```
zizou0x, line 1990: Is this a new formatting?

tycho-indexer/src/extractor/evm/ambient.rs
```diff
@@ -505,7 +505,7 @@ mod test {
 }
 
 #[cfg(test)]
-mod gateway_test {
+mod test_serial_db {
```
zizou0x, line 508: So this mod naming is important now? And all tests inside this will be run in a serial way, is that it?

Can we scope it to only a single test with a decorator or something? 

For example `test_arg_parsing_long_from_env` would need to be run alone also

tycho-indexer/src/storage/postgres/cache.rs
zizou0x: I think it shouldn't be needed if we stick to write only through this cachedGW, but it can makes sense if we need to manually update something (ideally it should never happen but for debugging purpose it could).

Do you know if there is any drawback (in terms of perfomance maybe?) to this?

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -644,7 +644,11 @@ impl StoredVersionedRow for AccountBalance {
         conn: &mut AsyncPgConnection,
     ) -> Result<Vec<Box<Self>>, StorageError> {
         Ok(account_balance::table
-            .filter(account_balance::account_id.eq_any(ids))
+            .filter(
+                account_balance::account_id
+                    .eq_any(ids)
+                    .and(account_balance::valid_to.is_null()),
```
zizou0x, line 650: Ugh, how comes we never had issue with this yet

tycho-indexer/src/storage/postgres/cache.rs
kayibal: > I think it shouldn't be needed if we stick to write only through this cachedGW

Yes correct we actually don't need it. But having it here will protect our DB from reaching an invalid state there is probably a small performance penalty from using.  I think at least in the beginning we should use it and later we can remove it if we want to prioritize faster writes - in cases we really need that.

tycho-indexer/src/extractor/evm/ambient.rs
kayibal: You can scope it to single functions by naming that function `test_serial_db....`

tycho-indexer/src/storage/postgres/orm.rs
kayibal: Because previously trigger would do this correctly :D

PR #108, opened 2024-01-25
ENG 2435: Implement ProtocolGateway's revert_protocol_states

louise-poole:


tycho-indexer/src/storage/postgres/chain.rs
```diff
@@ -121,16 +122,93 @@ where
...
+        // child entries.
+        diesel::delete(
+            schema::block::table
+                .filter(schema::block::number.gt(block.number))
+                .filter(schema::block::chain_id.eq(block.chain_id)),
+        )
+        .execute(conn)
+        .await?;
+
+        // Any versioned table's rows, which have `valid_to` set to "> block.ts"
```
tvinagre, line 150: Shouldn't it be `gte`?
If a transaction was overriden in this block, `valid_to` will be set to `this_block` and should become `None` since this block was rolled-back

louise-poole: Nope, the block we're reverting to still exists - all the blocks after it are the ones that are reverted.

tycho-indexer/src/storage/postgres/chain.rs
```diff
@@ -121,16 +122,93 @@ where
...
+        // To revert all changes of a chain, we need to delete & modify entries
+        // from a big number of tables. Reverting state, signifies deleting
+        // history. We will not keep any branches in the db only the main branch
+        // will be kept.
+        let block = orm::Block::by_id(to, conn).await?;
+
+        // All entities and version updates are connected to the block via a
+        // cascade delete, this ensures that the state is reverted by simply
+        // deleting the correct blocks, which then triggers cascading deletes on
+        // child entries.
```
tvinagre: Here is not clear by the comments that you will revert every block *after* the reference block `to`. I think it's worth a warning

tycho-indexer/src/storage/postgres/chain.rs
```diff
@@ -121,16 +122,93 @@ where
...
+            schema::block::table
+                .filter(schema::block::number.gt(block.number))
+                .filter(schema::block::chain_id.eq(block.chain_id)),
+        )
+        .execute(conn)
+        .await?;
+
+        // Any versioned table's rows, which have `valid_to` set to "> block.ts"
+        // need, to be updated to be valid again (thus, valid_to = NULL).
+        diesel::update(
```
tvinagre, line 152: Is there a Rustonic way of doing all these repeated calls with less code?
Like Python's 
```python
for schema in [contract_storage, account_balance, contract_code...]:
    diesel.update(schema.filter(schema.valid_to.gt(block.ts)),
    ...
```

I know the premises of the two languages are completely different but I'm just wondering

tvinagre: Also, rolling back should be atomic, otherwise we'd have dangling references.

tvinagre: Don't know if it works, but maybe something like:
```rust
async fn revert_state(
    &self,
    to: &BlockIdentifier,
    conn: &mut AsyncPgConnection,
    let block = orm::Block::by_id(to, conn).await?;

    // Start a transaction
    conn.transaction(|conn| {
        async {
            // Delete blocks
            self.delete_blocks_greater_than(block.number, block.chain_id, conn).await?;

            // Reset valid_to for various tables
            self.reset_valid_to(schema::contract_storage::table, block.ts, conn).await?;
            self.reset_valid_to(schema::account_balance::table, block.ts, conn).await?;
            self.reset_valid_to(schema::contract_code::table, block.ts, conn).await?;
            self.reset_valid_to(schema::protocol_state::table, block.ts, conn).await?;
            self.reset_valid_to(schema::protocol_calls_contract::table, block.ts, conn).await?;
            self.reset_valid_to(schema::account::table, block.ts, conn).await?;
            self.reset_valid_to(schema::protocol_component::table, block.ts, conn).await?;

            Ok(())
        }
        .await
    })
}

async fn delete_blocks_greater_than(
    &self,
    number: i64,
    chain_id: i32,
    conn: &mut AsyncPgConnection,
    diesel::delete(
        schema::block::table
            .filter(schema::block::number.gt(number))
            .filter(schema::block::chain_id.eq(chain_id)),
    )
    .execute(conn)
    .await?;

    Ok(())
}

async fn reset_valid_to<T>(
    &self,
    table: T,
    ts: NaiveDateTime,
    conn: &mut AsyncPgConnection,
) -> Result<(), StorageError>
where
    T: Table,
    T::AllColumns: Expression,
{
    diesel::update(table.filter(T::valid_to.gt(ts)))
        .set(T::valid_to.eq(Option::<NaiveDateTime>::None))
        .execute(conn)
        .await?;

    Ok(())
}

```

tycho-indexer/src/storage/postgres/chain.rs
louise-poole: This is already given/noted in the function doc string in the trait definition. I will update the docstring to make it clearer though. And add it to these comments for good measure.

tycho-indexer/src/storage/postgres/chain.rs
louise-poole: I had a look into this - using a generic helper function affects performance as it will need to use dynamic dispatch (`dyn`) and boxing (which uses the heap) since the function isn't statically typed at compile time. The code isn't excessively repeated so I would prefer to lean towards performance here over readability here. I will update the comments inside the code block to hopefully help improve this.

kayibal: I think performance should not be an issue here dyn objects use one more pointer to do the lookup, the heap things are more problematic but as we only call this once per hour or so performance is not a big issues. Dyn objects and heap lookup are relevant if you expect to call sth 10k or more times within a second.

Nevertheless making this code more dynamic is probably hard and would require familiarity with a bunch of diesel traits... So not against leaving it as is.

PR #109, opened 2024-01-25
ENG-2715: Ensure protocol types on startup

dianacarvalho1:
[Task](https://datarevenue.atlassian.net/jira/software/projects/ENG/boards/24?selectedIssue=ENG-2715)

Original issue: We need to know the protocol type when we convert the protobuf message of `ProtocolComponent` into an object `ProtocolComponent.

Before we had already decided that the `ProtocolType` would be inserted by the Ambient Extractor. So the ideia was to insert the `ProtocolType`s on startup and then pass the correct ID to the `ProtocolComponent`.

My issue: I don't like much that we are passing around a database ID on the extractor level 😕  so I preferred to pass the protocol type name instead and find the ID further down inside the gtw. 

I'm very unsure about these changes.. 😕 they seem very brittle. Let me know what you think before I continue with this (Fix tests and polish methods)

### DIFF
--- a/tycho-indexer/src/extractor/evm/ambient.rs
+++ b/tycho-indexer/src/extractor/evm/ambient.rs
@@ -293,6 +293,7 @@ where
         ExtractorIdentity::new(self.chain, &self.name)
     }
 
+    /// Make sure that the protocol types are present in the database.
     async fn ensure_protocol_types(&self) {
         let protocol_types: Vec<ProtocolType> = self
             .protocol_types
--- a/tycho-indexer/src/main.rs
+++ b/tycho-indexer/src/main.rs
@@ -2,7 +2,6 @@
 
 use diesel_async::{pooled_connection::deadpool::Pool, AsyncPgConnection};
 use futures03::future::select_all;
-use std::collections::HashMap;
 
 use extractor::{
     evm::ambient::{AmbientContractExtractor, AmbientPgGateway},
@@ -10,12 +9,6 @@ use extractor::{
 };
 use models::Chain;
 
-use crate::{
-    extractor::{evm, ExtractionError},
-    models::{FinancialType, ImplementationType, ProtocolType},
-    services::ServicesBuilder,
-    storage::postgres::{self, cache::CachedGateway, PostgresGateway},
-};
 use actix_web::dev::ServerHandle;
 use clap::Parser;
 use std::sync::Arc;
@@ -26,9 +19,12 @@ use tycho_indexer::{
     extractor,
     extractor::{evm, ExtractionError},
     models,
+    models::{FinancialType, ImplementationType, ProtocolType},
     services::ServicesBuilder,
-    storage,
-    storage::postgres::{self, cache::CachedGateway, PostgresGateway},
+    storage::{
+        postgres,
+        postgres::{cache::CachedGateway, PostgresGateway},
+    },
 };
 
 /// Tycho Indexer using Substreams
@@ -126,7 +122,7 @@ async fn main() -> Result<(), ExtractionError> {
         }
     });
 
-    let write_executor = crate::storage::postgres::cache::DBCacheWriteExecutor::new(
+    let write_executor = postgres::cache::DBCacheWriteExecutor::new(
         "ethereum".to_owned(),
         Chain::Ethereum,
         pool.clone(),
@@ -167,8 +163,7 @@ async fn start_ambient_extractor(
 ) -> Result<(JoinHandle<Result<(), ExtractionError>>, ExtractorHandle), ExtractionError> {
     let ambient_name = "vm:ambient";
     let ambient_gw = AmbientPgGateway::new(ambient_name, Chain::Ethereum, pool, cached_gw);
-    let mut ambient_protocol_types = HashMap::new();
-    ambient_protocol_types.insert(
+    let ambient_protocol_types = [(
         "vm:pool".to_string(),
         ProtocolType::new(
             "ambient_pool".to_string(),
@@ -176,7 +171,9 @@ async fn start_ambient_extractor(
             None,
             ImplementationType::Vm,
         ),
-    );
+    )]
+    .into_iter()
+    .collect();
     let extractor = AmbientContractExtractor::new(
         ambient_name,
         Chain::Ethereum,


tycho-indexer/src/extractor/evm/ambient.rs
```diff
@@ -268,6 +288,22 @@ where
         ExtractorIdentity::new(self.chain, &self.name)
     }
 
+    async fn ensure_protocol_types(&mut self) {
+        let pt = ProtocolType::new(
+            "ambient_pool".to_string(),
+            FinancialType::Swap,
+            None,
+            ImplementationType::Vm,
+        );
```
dianacarvalho1: this will be specific per each protocol system. in ambient we only have one. in balancer we will have several

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -229,23 +234,24 @@ where
             .await?;
         Ok(())
     }
-    async fn upsert_protocol_type(
+    async fn add_protocol_types(
```
dianacarvalho1, line 237: I had to change it from an update to an add 😕  I don't think we can do bulk updates

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -1000,25 +1006,25 @@ mod test {
...
-
-        assert_eq!(newly_inserted_data.len(), 1);
-        assert_eq!(newly_inserted_data[0].name, "Protocol".to_string());
-        assert_eq!(newly_inserted_data[0].financial_type, orm::FinancialType::Leverage);
-        assert_eq!(
-            newly_inserted_data[0].attribute_schema,
-            Some(json!({"attribute": "another_schema"}))
-        );
-        assert_eq!(newly_inserted_data[0].implementation, orm::ImplementationType::Vm);
+        // gw.add_protocol_types(&updated_protocol_type, &mut conn)
```
dianacarvalho1: to be deleted if you like the idea

tycho-indexer/src/storage/postgres/protocol.rs
SmartMonkey-git: I can confirm

tycho-indexer/src/extractor/evm/ambient.rs
kayibal: Yeah remember though that we plan to make the ambient extractor generic, so I think this would be better hardcoded in `main.rs` and then passed into the extractor constructor.

tycho-indexer/src/extractor/evm/ambient.rs
```diff
@@ -296,14 +332,16 @@ where
 
         debug!(?raw_msg, "Received message");
 
-        // TODO: figure out how/where to get this ID from (in ENG-2049)
-        let protocol_type_id = String::from("id-1");
+        let protocol_type_id = self
+            .protocol_type_names
+            .get("pool")
+            .expect("Couldn't fing Protocol Type");
```
kayibal: typo

tycho-indexer/src/extractor/evm/ambient.rs
```diff
@@ -296,14 +332,16 @@ where
...
         let msg = match evm::BlockContractChanges::try_from_message(
             raw_msg,
             &self.name,
             self.chain,
             self.protocol_system.clone(),
-            protocol_type_id,
+            protocol_type_id.clone(),
```
kayibal: Better to clone within `try_from_message`

tycho-indexer/src/extractor/evm/ambient.rs
```diff
@@ -268,6 +288,22 @@ where
...
+            None,
+            ImplementationType::Vm,
+        );
+
+        self.gateway
+            .ensure_protocol_types(&[pt])
+            .await;
+
+        self.protocol_type_names
+            .insert("pool".to_string(), "ambient_pool".to_string());
```
kayibal: Why do we need this mapping? The unique identifier should be a single string no? I would simply name this "vm:pool" or so and the extractor has a `HashMap<String, ProtocolType>`

tycho-indexer/src/extractor/evm/ambient.rs
```diff
@@ -268,6 +288,22 @@ where
         ExtractorIdentity::new(self.chain, &self.name)
     }
 
+    async fn ensure_protocol_types(&mut self) {
```
kayibal: Kind of strange that this mutates it's own state no?

tycho-indexer/src/extractor/evm/ambient.rs
dianacarvalho1: why?? or how? it says:
"cannot move out of `protocol_type.name` which is behind a shared reference [E0507] move occurs because `protocol_type.name` has type `std::string::String`, which does not implement the `Copy` trait"

tycho-indexer/src/main.rs
```diff
@@ -169,8 +171,23 @@ async fn start_ambient_extractor(
...
+            ImplementationType::Vm,
+        ),
+    );
+    let extractor = AmbientContractExtractor::new(
+        ambient_name,
+        Chain::Ethereum,
+        ambient_gw,
+        ambient_protocol_types,
+    )
+    .await?;
```
kayibal: We can make hashmap literals with this trick, avoids having a mut variable.

```suggestion
    let extractor = AmbientContractExtractor::new(
        ambient_name,
        Chain::Ethereum,
        ambient_gw,
        [(
         "vm:pool".to_string(),
          ProtocolType::new(
              "ambient_pool".to_string(),
              FinancialType::Swap,
              None,
              ImplementationType::Vm,
          ),
       )].into_iter().collect(),
    )
    .await?;
```

tycho-indexer/src/extractor/evm/ambient.rs
kayibal: Functions or methods should prefer to take a reference: `&str`, because this way the method can decide if the .value needs to be cloned or not. It is also more flexible because it can take both `&str` and `&String`

tycho-indexer/src/extractor/evm/ambient.rs
```diff
@@ -268,6 +293,17 @@ where
...
+    async fn ensure_protocol_types(&self) {
+        let protocol_types: Vec<ProtocolType> = self
+            .protocol_types
+            .values()
+            .cloned()
+            .collect();
+        self.gateway
+            .ensure_protocol_types(&protocol_types)
+            .await;
+    }
```
tamaralipows, line 306: I only have one request that's not mandatory but would make m really happy. Could we add a comment or a docstring? I have a feeling if I come back to this later, I won't remember what we meant by "ensure" in this context. Right now I interpreted this as - ensuring they are inserted?

PR #110, opened 2024-01-25
ENG-2030: Adjust Token-Component relation

SmartMonkey-git:
Updated: Mon Jan 29 2024 11:29:05
[Slack](https://propellerswap.slack.com/archives/C05LQALE9L1/p1706176481423879)

The goal is to match the ids of token table and the protocol_component table.

The procedure is:
- Insert new components as before, but new retrieve id, external_id, protocol_system_id and chain_id. The three latter make the unique key of the table.
- Convert these retrieved results into a hashmap<unique_key, id>. Having the unique_key as key lets me match the orm to emv object. Which lets me match orm to token, because the evm object has the token address. 
- Get all unique token addresses.
- Create HashMap that matches protocol_component id's to token addresses
- Do a joined query on token and accounts to retrieve `token.id` and `account.address`. Retrieve a HashMap that has token address as key and token id as value.
- Match `evm::ProtocolComponent.token_addresses` to `account.address` and you to a relation.

### DIFF
--- a/tycho-indexer/migrations/2024-01-10_tycho_patch/down.sql
+++ b/tycho-indexer/migrations/2024-01-10_tycho_patch/down.sql
@@ -146,3 +146,4 @@ CREATE TRIGGER invalidate_previous_contract_code
     AFTER INSERT ON contract_code
     FOR EACH ROW
     EXECUTE PROCEDURE invalidate_previous_entry_contract_code();
+


tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -171,41 +172,100 @@ where
...
-            .execute(conn)
+        let inserted_protocol_components: Vec<orm::ProtocolComponent> =
+            diesel::insert_into(protocol_component)
+                .values(&values)
+                .on_conflict((
+                    schema::protocol_component::chain_id,
+                    protocol_system_id,
+                    external_id,
+                ))
+                .do_nothing()
```
kayibal, line 217: I think you can use `returning((id, name))` instead of the zip further down although if the order is guaranteed we probably should save us that additional data transfer.

I didn't read this but if this sections mentions the order we can probably rely on it:
https://www.postgresql.org/docs/current/dml-returning.html#DML-RETURNING

If it is not mentioned though, it is probably an postgres implementation detail and we shouldn't rely on it.

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -171,41 +172,100 @@ where
...
+                .get_results(conn)
+                .await
+                .map_err(|err| StorageError::from_diesel(err, "ProtocolComponent", "", None))?;
+
+        let component_id_by_token_addresses: HashMap<i64, Vec<H160>> = inserted_protocol_components
```
kayibal: We shouldn't use H160 type here... (because it is ethereum specific) can you add a method on ProtocolComponent that returns the token addresses as bytes please?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -171,41 +172,100 @@ where
...
+            .select((schema::account::address, schema::token::id))
+            .filter(schema::account::address.eq_any(token_addresses))
+            .into_boxed();
+
+        let token_add_by_id: HashMap<Address, i64> = query
+            .load::<(Address, i64)>(conn)
             .await
-            .map_err(|err| StorageError::from_diesel(err, "ProtocolComponent", "", None))
-            .unwrap();
+            .map_err(|err| {
```
kayibal: Why do you split the query here?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -171,41 +172,100 @@ where
...
+        let token_addresses: Vec<Address> = new
+            .iter()
+            .flat_map(|pc| {
+                pc.tokens
+                    .clone()
+                    .into_iter()
+                    .map(|token_address| Address::from(token_address.0))
+                    .collect::<Vec<Address>>()
+            })
+            .collect();
```
kayibal: These bits indeed feel a bit chunky with the amount of hashmaps and vectors required. Maybe you can try to create a vector that will hold `(component_id, token_address)`? Since this is close of what will go into the final table.

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -171,41 +172,100 @@ where
...
+
+        let query = token
+            .inner_join(account)
+            .select((schema::account::address, schema::token::id))
+            .filter(schema::account::address.eq_any(token_addresses))
```
kayibal, line 272: The here you can collect that vector into a HashSet to deduplicate addresses like this:
```
let unique_token_addresses = cid_token_addr.iter().map(|(_, addr)| addr).unique()
```

The benefit of this that you use references (that refer back to the vector of tuples) and you should be able to pass this into here.

tycho-indexer/src/storage/postgres/protocol.rs
SmartMonkey-git: The documentation does not mention the return row order, it only mentions the column order. Additionally, I found a discussion on [stack overflow](https://stackoverflow.com/questions/5439293/is-insert-returning-guaranteed-to-return-things-in-the-right-order) about it, which seems rather pessimistic.

So, I would not like to rely on it.

tycho-indexer/migrations/2024-01-10_tycho_patch/down.sql
```diff
@@ -61,7 +61,4 @@ CREATE TRIGGER invalidate_previous_protocol_state
 ALTER TABLE protocol_system
 DROP CONSTRAINT name_unique;
 
-ALTER TABLE token
-DROP CONSTRAINT unique_account_id_constraint;
-
-DROP TABLE IF EXISTS protocol_component_token;
\ No newline at end of file
+ALTER TABLE token
```
kayibal: Do we alter it or not?

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -337,8 +337,8 @@ pub struct ProtocolComponent {
 }
 
 #[derive(Insertable)]
-#[diesel(table_name = protocol_component_token_junction)]
-pub struct NewProtocolComponentToken {
+#[diesel(table_name = protocol_holds_token)]
+pub struct NewProtocolComponentTokenRelation {
```
kayibal: There is also already an orm for this table, did you see that? We should at least remove the old one if we want to use the new one or if we need to use both make the naming consistent.

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -207,63 +211,75 @@ where
...
+                        self.get_chain(&chain_db_id),
+                    )
+            }) {
+                pc.get_byte_token_addresses()
+                    .into_iter()
+                    .for_each(|t_byte_address| {
+                        component_id_by_token_address.push((db_id, t_byte_address))
+                    });
+            }
+        }
```
kayibal: We need to work on this, apart from looking quite terrifying, it also uses quadratic complexity which I think is not necessary (most of the time prefer to use more hash maps or sets to avoid quadratic complexity). 

So the end result is a list of `(component.id, token.id)` and you correctly realised that we need to use the entity id: (ext_id, ps_id, chain_id) to identify our corresponding component. I should work with a HelperHashMaps and a HashSet here:

```
unique_tokens = HashSet<&Address> ) = new.iter().flatmap(|pc| pc.tokens.iter()).collect()
pc_entity_tokens_map = HashMap<(&String, i64. i64), Vec<&Address>> = new.iter().map(....)
token_pk_map = HashMap<Address, i64> = token.inner_join()....filter(unique_tokens).collect()

protocol_holds_token_data = inserted.iter().flatmap(| (db_id, ext_id, ps_id, chain_db_id) | {
    pc_entity_tokens_map.get(&(&ext_id, *ps_id, *chain_db_id)).iter().map(|address| (db_id, address)
})
```

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -393,10 +409,10 @@ where
         for state in new {
             let tx_db = txns
                 .get(state.modify_tx.as_bytes())
-                .expect("Failed to find tx");
+                .expect("Failed to find tx"); // TODO: Found out that these are panics in disgues. Should we panic here?
             let component_db_id = *components
                 .get(&state.component_id)
-                .expect("Failed to find component");
+                .expect("Failed to find component"); // TODO: Found out that these are panics in disgues. Should we panic here?
```
kayibal: Nope we should probably return an error

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -171,41 +173,109 @@ where
...
+            .iter()
+            .flat_map(|pc| pc.get_byte_token_addresses())
+            .collect();
+
+        let pc_entity_tokens_map = new
+            .iter()
+            .flat_map(|pc| {
+                let pc_id = protocol_db_id_map
+                    .get(&(pc.id.clone(), pc.protocol_system.clone(), pc.chain))
+                    .expect("Could not find Protocol Component. Even though it should have."); //Because we just insderted the protocol systems, there should not be any missing.
```
dianacarvalho1:
```suggestion
                    .expect("Could not find Protocol Component. Even though it should have."); //Because we just inserted the protocol systems, there should not be any missing.
```

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -171,41 +173,109 @@ where
...
+                                                                                               // However, trying to handel this via Results is needlessly difficult, because you
```
dianacarvalho1:
```suggestion
                                                                                               // However, trying to handle this via Results is needlessly difficult, because you
```

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -171,41 +173,109 @@ where
...
+        let inserted_protocol_components: Vec<(i64, String, i64, i64)> =
+            diesel::insert_into(protocol_component)
+                .values(&values)
+                .on_conflict((
+                    schema::protocol_component::chain_id,
+                    protocol_system_id,
+                    external_id,
+                ))
+                .do_nothing()
+                .returning((
```
dianacarvalho1, line 218: I think this won't return protocol components that already exist. I think it's ok in the flow of things though. This basically mean that we won't be able to add/update existing protocol components. even if the change was in the tokens only

tycho-indexer/src/storage/postgres/protocol.rs
dianacarvalho1: ahahahahah why should it have?

dianacarvalho1: ah yeah, it will fail here if the protocol component already exists actually 😕

dianacarvalho1: do we really want it to fail?

tycho-indexer/src/storage/postgres/protocol.rs
SmartMonkey-git: > This basically mean that we won't be able to add/update existing protocol components. even if the change was in the tokens only

I would not expect an add function to update any entries, so I think its ok.

tycho-indexer/src/storage/postgres/protocol.rs
SmartMonkey-git: > ah yeah, it will fail here if the protocol component already exists actually 😕

Fixed it.

> do we really want it to fail?

No, but it should never fail.

kayibal: I am ok with an expect here because as @SmartMonkey-git mentioned, it should basically never happen so no need to deal with a Result.
Buuuut @SmartMonkey-git dealing with result here is not hard if you know this trick: https://doc.rust-lang.org/rust-by-example/error/iter_result.html#fail-the-entire-operation-with-collect

PR #111, opened 2024-01-26
ENG-2381: Add native protocol extractor

tvinagre:
Context: https://datarevenue.atlassian.net/browse/ENG-2371

Slack thread about decisions: https://propellerswap.slack.com/archives/C05LQALE9L1/p1706272378050749

### DIFF
--- a/tycho-indexer/src/extractor/evm/ambient.rs
+++ b/tycho-indexer/src/extractor/evm/ambient.rs
@@ -626,12 +626,16 @@ mod test {
         .expect("extractor init ok");
         let inp = evm::fixtures::pb_block_scoped_data(());
 
-        let _res = extractor
+        let res = extractor
             .handle_tick_scoped_data(inp)
             .await;
 
-        // TODO: fix this assert
-        // assert_eq!(res, Ok(None));
+        match res {
+            Ok(Some(_)) => panic!("Expected Ok(None) but got Ok(Some(..))"),
+            Ok(None) => (), // This is the expected case
+            Err(_) => panic!("Expected Ok(None) but got Err(..)"),
+        }
+
         assert_eq!(extractor.get_cursor().await, "cursor@420");
     }
 
--- a/tycho-indexer/src/extractor/evm/native.rs
+++ b/tycho-indexer/src/extractor/evm/native.rs
@@ -500,31 +500,33 @@ mod test {
         assert_eq!(extractor.get_cursor().await, "cursor@420");
     }
 
-    // TODO: What is this testing?
-    // #[tokio::test]
-    // async fn test_handle_tick_scoped_data_skip() {
-    //     let mut gw = MockNativeGateway::new();
-    //     gw.expect_ensure_protocol_types()
-    //         .times(1)
-    //         .returning(|_| ());
-    //     gw.expect_get_cursor()
-    //         .times(1)
-    //         .returning(|| Ok("cursor".into()));
-    //     gw.expect_advance()
-    //         .times(0)
-    //         .returning(|_, _| Ok(()));
+    #[tokio::test]
+    async fn test_handle_tick_scoped_data_skip() {
+        let mut gw = MockNativeGateway::new();
+        gw.expect_ensure_protocol_types()
+            .times(1)
+            .returning(|_| ());
+        gw.expect_get_cursor()
+            .times(1)
+            .returning(|| Ok("cursor".into()));
+        gw.expect_advance()
+            .times(0)
+            .returning(|_, _| Ok(()));
 
-    //     let protocol_types = HashMap::from([("WeightedPool".to_string(),
-    // ProtocolType::default())]);     let extractor = create_extractor(gw).await;
+        let extractor = create_extractor(gw).await;
 
-    //     let inp = evm::fixtures::pb_block_scoped_data(());
-    //     let res = extractor
-    //         .handle_tick_scoped_data(inp)
-    //         .await;
+        let inp = evm::fixtures::pb_block_scoped_data(());
+        let res = extractor
+            .handle_tick_scoped_data(inp)
+            .await;
 
-    //     assert_eq!(res, Ok(None));
-    //     assert_eq!(extractor.get_cursor().await, "cursor@420");
-    // }
+        match res {
+            Ok(Some(_)) => panic!("Expected Ok(None) but got Ok(Some(..))"),
+            Ok(None) => (), // This is the expected case
+            Err(_) => panic!("Expected Ok(None) but got Err(..)"),
+        }
+        assert_eq!(extractor.get_cursor().await, "cursor@420");
+    }
 
     #[tokio::test]
     async fn test_handle_revert() {
@@ -573,59 +575,6 @@ mod test {
         assert!(matches!(res, Ok(None)));
         assert_eq!(extractor.get_cursor().await, "cursor@400");
     }
-
-    // THALES TESTS
-
-    // #[tokio::test]
-    // async fn test_update_cursor() {
-    //     let extractor = create_extractor();
-    //     let new_cursor = "new_cursor".to_string();
-
-    //     // Update the cursor.
-    //     extractor
-    //         .update_cursor(new_cursor.clone())
-    //         .await;
-
-    //     // Lock the state to retrieve the cursor.
-    //     let state = extractor.inner.lock().await;
-
-    //     // Check if the cursor has been updated correctly.
-    //     assert_eq!(state.cursor, new_cursor.into_bytes());
-    // }
-
-    // #[tokio::test]
-    // async fn test_update_last_processed_block() {
-    //     let extractor = create_extractor();
-    //     // pub struct Block {
-    //     //     pub number: u64,
-    //     //     pub hash: H256,
-    //     //     pub parent_hash: H256,
-    //     //     pub chain: Chain,
-    //     //     pub ts: NaiveDateTime,
-    //     // }
-
-    //     let new_block = Block {
-    //         number: 1,
-    //         hash: H256::zero(),
-    //         parent_hash: H256::zero(),
-    //         chain: Chain::Ethereum,
-    //         ts: "2020-01-01T01:00:00"
-    //             .parse()
-    //             .expect("Invalid timestamp"),
-    //     };
-
-    //     // Update the last processed block.
-    //     extractor
-    //         .update_last_processed_block(new_block.clone())
-    //         .await;
-
-    //     // Lock the state to retrieve the last processed block.
-    //     let state = extractor.inner.lock().await;
-
-    //     // Check if the last processed block has been updated correctly.
-    //     assert!(state.last_processed_block.is_some());
-    //     assert_eq!(state.last_processed_block.unwrap(), new_block);
-    // }
 }
 
 #[cfg(test)]


tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -288,6 +289,19 @@ impl NormalisedMessage for BlockAccountChanges {
...
+    }
+}
+
+#[typetag::serde]
+impl NormalisedMessage for BlockEntityChangesResult {
+    fn source(&self) -> ExtractorIdentity {
+        ExtractorIdentity::new(self.chain, &self.extractor)
+    }
+}
+
```
louise-poole, line 373: Is this needed for your PR? 🤔  I was adding it in mine too 😅

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -783,22 +797,80 @@ impl ProtocolStateDelta {
 /// Updates grouped by their respective transaction.
 #[derive(Debug, Clone, PartialEq, Default)]
 pub struct ProtocolStateDeltasWithTx {
```
louise-poole: Can we change this name? Since it's no longer holding just ProtocolStateDeltas... maybe we can just call it `ProtocolChangesWithTx`?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -783,22 +797,80 @@ impl ProtocolStateDelta {
...
+    pub new_protocol_components: HashMap<String, ProtocolComponent>,
     pub protocol_states: HashMap<String, ProtocolStateDelta>,
+    pub balance_changes: HashMap<String, HashMap<H160, TvlChange>>,
```
louise-poole: I think this branch is out of date - the `TvlChange` object is now `BalanceChanges` (fyi for when you update this branch)

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -783,22 +797,80 @@ impl ProtocolStateDelta {
 /// Updates grouped by their respective transaction.
 #[derive(Debug, Clone, PartialEq, Default)]
 pub struct ProtocolStateDeltasWithTx {
+    pub new_protocol_components: HashMap<String, ProtocolComponent>,
```
louise-poole, line 898: @dianacarvalho1 not all ProtocolComponent messages are created components, right? 
There should be logic checking the change type of this component and we populate 2 lists - `new_protocol_components` and `deleted_protocol_components` accordingly. Looks like this is missing from the vm structs/message handling too.

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -783,22 +797,80 @@ impl ProtocolStateDelta {
...
+        for state_msg in msg.entity_changes.into_iter() {
             let state = ProtocolStateDelta::try_from_message(state_msg)?;
-            protocol_states.insert(state.clone().component_id, state);
+            // Check if a state update for the same component already exists
+            // If it exists, overwrite the existing state update with the new one and log a warning
+            if state_updates.contains_key(&state.component_id) {
+                warn!(
+                    "Overwriting state update for component {} with a new one",
+                    state.component_id
+                );
```
louise-poole: Hmmm... I don't think this is doing what you're hoping it does. If we have 2 `ProtocolStateDelta` for the same component, they could very well be updating different attributes and we will need both deltas instead of overriding one. I think we should avoid aggregating the info here before it is sent to the db - there is already logic to correctly aggregate and merge deltas when we create the `BlockEntityChangesResult`.

louise-poole: Therefore it probably makes sense to just have a `Vec<ProtocolStateDelta>` here. Same for components and balances.

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -783,22 +797,80 @@ impl ProtocolStateDelta {
...
+            balance_changes
+                .get_mut(&tvl_change.component_id)
+                .unwrap()
+                .insert(tvl_change.token, tvl_change);
+        }
+
+        Ok(Self { new_protocol_components, protocol_states: state_updates, balance_changes, tx })
     }
 
     /// Merges this update with another one.
```
louise-poole, line 981: This merge method is used by `BlockEntityChanges` to aggregate all changes so that only final changes are emitted by our extractor. Right now it's only merging/aggregating protocol states - you will need to add merge logic for components and balances here too. (I suppose merging components just means making sure a component isn't listed in both the `new` and `deleted` lists?)

louise-poole: @dianacarvalho1 Looks like this is missing from `BlockContractChanges::aggregate_updates` too - the balance changes info is lost when the block data is aggregated into `BlockAccountChanges`.

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -965,6 +1057,30 @@ impl BlockEntityChanges {
...
         })
     }
+
+    pub fn get_new_protocols(&self) -> Vec<ProtocolStateDeltasWithTx> {
+        self.state_updates
+            .iter()
+            .filter(|update| update.has_new_protocols())
+            .cloned()
+            .collect()
+    }
```
louise-poole: Shouldn't this just return `Vec<ProtocolStateDelta>`?

tycho-indexer/src/extractor/evm/mod.rs
louise-poole: UPDATE: chatted with Diana - for now all ProtocolComponent messages are indeed created components. We do not know of any protocols that delete pools yet and will add this functionality when/if we encounter them.

tycho-indexer/src/extractor/evm/mod.rs
tvinagre: Ah yeah I've already fixed this one locally - thanks!

tycho-indexer/src/extractor/evm/mod.rs
tvinagre: Yup! Feel free to cherry-pick to yours if it's needed rn :)

tycho-indexer/src/extractor/evm/mod.rs
tvinagre: Sure!

tycho-indexer/src/extractor/evm/mod.rs
tvinagre: Hmmm, yeah I understand, but previously it was already a `HashMap` and we'd do:
`protocol_states.insert(state.clone().component_id, state);`
I just added this warning - I'll check now if there was a bug in the previous logic

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -862,8 +945,50 @@ impl ProtocolStateDeltasWithTx {
...
+                Entry::Vacant(entry) => {
+                    entry.insert(value);
+                }
+            }
+        }
+
         Ok(())
     }
+
+    pub fn has_new_protocols(&self) -> bool {
```
tvinagre: Future Thales, pls check if this is still needed.

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -973,6 +1106,30 @@ impl BlockEntityChanges {
             new_protocol_components: self.new_protocol_components,
         })
     }
+
+    pub fn get_new_protocols(&self) -> Vec<ProtocolChangesWithTx> {
```
tvinagre: Future Thales, pls check if this is still needed.

tycho-indexer/src/pb/tycho.evm.v1.rs
kayibal, line 1: Formatting party

tvinagre: Yeah, no clue why, will just delete these changes

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -888,7 +1010,7 @@ pub struct BlockEntityChanges {
     extractor: String,
     chain: Chain,
     pub block: Block,
-    pub state_updates: Vec<ProtocolStateDeltasWithTx>,
+    pub state_updates: Vec<ProtocolChangesWithTx>,
     pub new_protocol_components: HashMap<String, ProtocolComponent>,
```
kayibal, line 1088: I think this attribute is now duplicated (already covered in `ProtocolChangesWithTx`)

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -904,14 +1026,19 @@ impl BlockEntityChanges {
...
-            let mut state_updates = Vec::new();
+            let mut state_updates: Vec<ProtocolChangesWithTx> = Vec::new();
             let mut new_protocol_components = HashMap::new();
 
             for change in msg.changes.into_iter() {
-                if let Some(tx) = change.tx {
-                    let tx = Transaction::try_from_message(tx, &block.hash)?;
-                    let tx_update =
-                        ProtocolStateDeltasWithTx::try_from_message(change.entity_changes, tx)?;
+                if let Some(tx) = change.clone().tx {
```
kayibal: Cloning the tx here is not enough? Do you need to clone the entire message?

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -0,0 +1,565 @@
...
+        conn: &mut AsyncPgConnection,
+    ) -> Result<evm::BlockEntityChangesResult, StorageError> {
+        let block = self
+            .state_gateway
+            .get_block(to, conn)
+            .await?;
+        let start = current.map(BlockOrTimestamp::Block);
+        let target = BlockOrTimestamp::Block(to.clone());
+
+        // CHECK: Here there's an assumption that self.name == protocol_system
```
kayibal, line 179: Yeah would be better via an attribute but the general assumption is correct (chain, extractor) : system = 1 : 1. Good point we should document that somewhere.

docker-compose.yaml
```diff
@@ -2,7 +2,7 @@ version: '3.1'
 services:
   db:
     image: postgres:15.3
-    restart: always
+    restart: "always"
```
zizou0x: FMI, why?

tycho-indexer/src/models.rs
```diff
@@ -22,7 +22,7 @@ use utoipa::ToSchema;
 #[strum(serialize_all = "lowercase")]
 pub enum Chain {
     #[default]
-    Ethereum,
+    /**/Ethereum,
```
zizou0x: This isn't intentional, right?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -240,9 +241,11 @@ where
                 let tokens_by_pc: &Vec<Address> = protocol_component_tokens
                     .get(&pc.id)
                     .expect("Could not find Tokens for Protocol Component."); // We expect all protocol components to have tokens.
+                // TODO: This is down for discussion - don't forget to remove this if not needed or deleting this comment.
```
zizou0x: The discussion is about if we allow protocol components to not have any contract? I would say yes, even though I don't see any case for this atm

tycho-indexer/src/extractor/evm/ambient.rs
```diff
@@ -278,14 +274,8 @@ impl AmbientGateway for AmbientPgGateway {
...
-        let res = conn
-            .transaction(|conn| {
-                async move {
-                    self.backward(current, to, new_cursor, conn)
-                        .await
-                }
-                .scope_boxed()
-            })
+        let res = self
+            .backward(current, to, new_cursor, &mut conn)
```
zizou0x, line 355: It has nothing to do with your PR but I think it's a bit weird that we need a `conn` for `backward()` and not `forward()`.

tycho-indexer/src/extractor/evm/ambient.rs
```diff
@@ -262,11 +261,8 @@ impl AmbientGateway for AmbientPgGateway {
         changes: &evm::BlockContractChanges,
         new_cursor: &str,
     ) -> Result<(), StorageError> {
-        let mut conn = self.pool.get().await.unwrap();
-        conn.transaction(|_conn| {
-            async move { self.forward(changes, new_cursor).await }.scope_boxed()
-        })
-        .await?;
+        self.forward(changes, new_cursor)
```
zizou0x, line 341: Not required anymore because we have the cached gw, right?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -809,23 +823,88 @@ impl ProtocolStateDelta {
...
             let state = ProtocolStateDelta::try_from_message(state_msg)?;
-            protocol_states.insert(state.clone().component_id, state);
+            // Check if a state update for the same component already exists
+            // If it exists, overwrite the existing state update with the new one and log a warning
+            match state_updates.entry(state.component_id.clone()) {
+                Entry::Vacant(e) => {
+                    e.insert(state);
+                }
+                Entry::Occupied(mut e) => {
+                    warn!("Overwriting state update for component {} with a new one", e.key());
```
zizou0x: Nice, maybe we can be a little bit more explicit on this warning, like "received two state updates for one component: overwriting..." or something like this?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -809,23 +823,88 @@ impl ProtocolStateDelta {
...
+            // warning
+            let token_balances = component_balances
+                .entry(component_balance.component_id.clone())
+                .or_default();
+
+            if let Some(existing_balance) =
+                token_balances.insert(component_balance.token, component_balance)
+            {
+                warn!(
+                    "Overwriting balance change for component {} and token {} with a new one",
```
zizou0x: Same here

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -809,23 +823,88 @@ impl ProtocolStateDelta {
...
 
-impl ProtocolStateDeltasWithTx {
+impl ProtocolChangesWithTx {
     /// Parses protocol state from tychos protobuf EntityChanges message
     pub fn try_from_message(
-        msg: Vec<substreams::EntityChanges>,
-        tx: Transaction,
+        msg: substreams::TransactionEntityChanges,
+        block: &Block,
+        protocol_system: String,
```
zizou0x: Can we make this a `&String`?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -919,14 +1041,19 @@ impl BlockEntityChanges {
...
-                    let tx = Transaction::try_from_message(tx, &block.hash)?;
-                    let tx_update =
-                        ProtocolStateDeltasWithTx::try_from_message(change.entity_changes, tx)?;
+                if let Some(tx) = change.tx.clone() {
+                    let tycho_tx = Transaction::try_from_message(tx, &block.hash)?;
+
+                    let tx_update = ProtocolChangesWithTx::try_from_message(
+                        change.clone(),
+                        &block,
+                        protocol_system.clone(),
```
zizou0x: Then if we can change it in `ProtocolChangesWithTx::try_from_message` we probably also can do it here and avoid this `clone()`

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -0,0 +1,954 @@
...
+    pb::{
+        sf::substreams::rpc::v2::{BlockScopedData, BlockUndoSignal, ModulesProgress},
+        tycho::evm::v1::BlockEntityChanges,
+    },
+    storage::{
+        postgres::cache::CachedGateway, BlockIdentifier, BlockOrTimestamp, StorageError, TxHash,
+    },
+};
+
+// TODO: Use the same Inner as AmbientExtractor
```
zizou0x: Looks like it's done already

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -0,0 +1,954 @@
...
+    protocol_system: String,
+    inner: Arc<Mutex<Inner>>,
+    protocol_types: HashMap<String, ProtocolType>,
+}
+
+impl<DB> NativeContractExtractor<DB> {
+    async fn update_cursor(&self, cursor: String) {
+        let cursor_bytes: Vec<u8> = cursor.into();
+        let mut state = self.inner.lock().await;
+        state.cursor = cursor_bytes;
```
zizou0x: I think we can do this to simplify

```suggestion
        let mut state = self.inner.lock().await;
        state.cursor = cursor.into();
```

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -0,0 +1,954 @@
...
+pub struct Inner {
+    cursor: Vec<u8>,
+    last_processed_block: Option<Block>,
+}
+
+pub struct NativeContractExtractor<G> {
+    gateway: G,
+    name: String,
+    chain: Chain,
+    protocol_system: String,
```
zizou0x, line 32: What is the difference between `name` and `protocol_system`?

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -0,0 +1,954 @@
...
+            self.state_gateway
+                .upsert_tx(&changes.block, &tx.tx)
+                .await?;
+
+            let hash: TxHash = tx.tx.hash.into();
+
+            for (_component_id, new_protocol_component) in tx.new_protocol_components.iter() {
+                new_protocol_components.push(new_protocol_component.clone());
+            }
+
```
zizou0x: I think we can simplify this

```suggestion
            new_protocol_components.extend(tx.new_protocol_components.values().cloned());
```

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -0,0 +1,954 @@
...
+            .await
+            .last_processed_block
+    }
+
+    #[instrument(skip_all, fields(chain = % self.chain, name = % self.name))]
+    async fn handle_tick_scoped_data(
+        &self,
+        inp: BlockScopedData,
+    ) -> Result<Option<ExtractorMsg>, ExtractionError> {
+        let _data = inp
```
zizou0x:
```suggestion
        let data = inp
```

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -0,0 +1,954 @@
...
+    async fn ensure_protocol_types(&self, new_protocol_types: &[ProtocolType]) {
+        let mut conn = self.pool.get().await.unwrap();
+        self.state_gateway
+            .add_protocol_types(new_protocol_types, &mut *conn)
+            .await
+            .expect("Couldn't insert protocol types");
+    }
+
+    #[instrument(skip_all, fields(chain = % self.chain, name = % self.name, block_number = % changes.block.number))]
+    async fn upsert_contract(
```
zizou0x: I feel like this function name is not really accurate

tycho-indexer/src/extractor/evm/ambient.rs
louise-poole: I assume it's because `forward()` just adds to the write cache and doesn't actually connect to the db, but `backward()` always reads from the db?

tycho-indexer/src/extractor/evm/mod.rs
louise-poole: Hi future Thales! How you doing?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -919,14 +1041,19 @@ impl BlockEntityChanges {
...
+            let mut state_updates: Vec<ProtocolChangesWithTx> = Vec::new();
             let mut new_protocol_components = HashMap::new();
 
             for change in msg.changes.into_iter() {
-                if let Some(tx) = change.tx {
-                    let tx = Transaction::try_from_message(tx, &block.hash)?;
-                    let tx_update =
-                        ProtocolStateDeltasWithTx::try_from_message(change.entity_changes, tx)?;
+                if let Some(tx) = change.tx.clone() {
+                    let tycho_tx = Transaction::try_from_message(tx, &block.hash)?;
```
louise-poole, line 1108: Curious: why name is `tycho-tx`? 👀 It feels purposefully specific. Why not just `tx` or `txn`?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -935,15 +1062,17 @@ impl BlockEntityChanges {
...
-                            tx.hash,
+                            tycho_tx.hash,
                             block.ts,
                         )?;
                         new_protocol_components.insert(pool.id.clone(), pool);
                     }
                 }
             }
 
+            // Sort transactions by index
```
louise-poole:
```suggestion
            // Sort updates by transactions index
```

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -919,14 +1041,19 @@ impl BlockEntityChanges {
     ) -> Result<Self, ExtractionError> {
         if let Some(block) = msg.block {
             let block = Block::try_from_message(block, chain)?;
-            let mut state_updates = Vec::new();
+            let mut state_updates: Vec<ProtocolChangesWithTx> = Vec::new();
```
louise-poole: These are no longer just state updates, so can probably rename this?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -970,7 +1100,7 @@ impl BlockEntityChanges {
     /// This returns an error if there was a problem during merge. The error
     /// type is `ExtractionError`.
     pub fn aggregate_updates(self) -> Result<BlockEntityChangesResult, ExtractionError> {
-        let base = ProtocolStateDeltasWithTx::default();
+        let base = ProtocolChangesWithTx::default();
 
         let aggregated_states = self
```
louise-poole: Can probably rename this to `aggregated_changes`? It's no longer just state deltas...

tycho-indexer/src/extractor/evm/native.rs
louise-poole: From what I understand, `name` should be unique to each `protocol_system`/`chain` pair.

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -0,0 +1,954 @@
...
+}
+
+#[automock]
+#[async_trait]
+pub trait NativeGateway: Send + Sync {
+    async fn get_cursor(&self) -> Result<Vec<u8>, StorageError>;
+
+    async fn ensure_protocol_types(&self, new_protocol_types: &[ProtocolType]);
+
+    async fn upsert_contract(
```
louise-poole: But we don't have contracts on native protocols?

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -0,0 +1,954 @@
...
+        self.save_cursor(&block, new_cursor)
+            .await?;
+
+        Ok(evm::BlockEntityChangesResult {
+            extractor: self.name.clone(),
+            chain: self.chain,
+            block,
+            revert: true,
+            state_updates,
+            // TODO: Map new_protocol_components
```
louise-poole: You can remove this TODO - there shouldn't be any new components on a revert. Potentially deleted components, but we're not supporting that just yet. 
Side note: we should add `deleted_protocol_components` to `BlockEntityChangesResult` for when we do one day implement handling their reverts correctly.

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -0,0 +1,954 @@
...
+            last_valid_cursor: "cursor@400".into(),
+        };
+
+        let res = extractor.handle_revert(inp).await;
+
+        assert!(matches!(res, Ok(None)));
+        assert_eq!(extractor.get_cursor().await, "cursor@400");
+    }
+
+    // THALES TESTS
```
louise-poole: I assume these are to be removed?

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -0,0 +1,954 @@
...
+        let res = extractor
+            .handle_tick_scoped_data(inp)
+            .await
+            .map(|o| o.map(|_| ()));
+
+        assert_eq!(res, exp);
+        assert_eq!(extractor.get_cursor().await, "cursor@420");
+    }
+
+    // TODO: What is this testing?
```
louise-poole: You tell me? 😂

tycho-indexer/src/pb/tycho.evm.v1.rs
```diff
@@ -6,59 +6,59 @@
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct Block {
     /// The blocks hash.
-    #[prost(bytes="vec", tag="1")]
+    #[prost(bytes = "vec", tag = "1")]
```
louise-poole, line 9: Note sure which formatting is the agreed upon one... 👀

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -520,7 +523,7 @@ where
                     version_ts,
                     conn,
                 )
-                .await,
+                    .await,
```
louise-poole: Some of these formatting changes seem quite odd to me - we should standardise this, is it not already?

louise-poole: I had a look through the repo - there's examples of both formatting rules, but yours introduced here is used the most it seems. Looking at the `rustfmt.toml` file I don't see the indent rules set - where did you set them?

docker-compose.yaml
tvinagre: Docker was complaining that `restart` needs to be a string

tycho-indexer/src/models.rs
tvinagre: No idea how this appeared here

tycho-indexer/src/storage/postgres/protocol.rs
tvinagre: Yes, this was discussed [here](https://propellerswap.slack.com/archives/C05LQALE9L1/p1707273092274649)

tycho-indexer/src/extractor/evm/ambient.rs
tvinagre: Yeah @louise-poole was on point

tycho-indexer/src/extractor/evm/ambient.rs
tvinagre: Yup

tycho-indexer/src/extractor/evm/mod.rs
tvinagre: Sure

tycho-indexer/src/extractor/evm/native.rs
tvinagre: Not really, this TODO was to remove code duplication, but we decided to leave this to another PR so I'll remove the TODO and open an issue

tycho-indexer/src/extractor/evm/native.rs
tvinagre: I think I needed this bc the compiler was annoying about having to know the types beforehand but I'm not sure

tycho-indexer/src/extractor/evm/native.rs
tvinagre: Also not clear to me, but yeas. Something like: `native:uniswap_v2`

tycho-indexer/src/extractor/evm/native.rs
tvinagre: The 🦀 👑

tycho-indexer/src/extractor/evm/native.rs
tvinagre: Yeah I agree, I just followed the `ambient` naming - will see if I can rename without breaking the whole world

tycho-indexer/src/extractor/evm/native.rs
tvinagre: Yeah just followed ambient naming, will fix!

tycho-indexer/src/pb/tycho.evm.v1.rs
tvinagre: Blame clippyyyy

tycho-indexer/src/storage/postgres/protocol.rs
tvinagre: I've never set them 😕 I think it was RustRover

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -1082,12 +1070,12 @@ impl BlockEntityChanges {
...
         extractor: &str,
         chain: Chain,
-        protocol_system: String,
+        protocol_system: &str,
         protocol_types: &HashMap<String, ProtocolType>,
     ) -> Result<Self, ExtractionError> {
         if let Some(block) = msg.block {
             let block = Block::try_from_message(block, chain)?;
-            let mut state_updates: Vec<ProtocolChangesWithTx> = Vec::new();
+            let mut update_txs: Vec<ProtocolChangesWithTx> = Vec::new();
```
louise-poole: To be picky: this name feels misleading. It sounds like it's just txs? Maybe `tx_updates` would be better? Or `updates_with_tx`?

tycho-indexer/src/extractor/evm/mod.rs
louise-poole: Don't forget to remove this

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -942,13 +1054,13 @@ pub struct BlockEntityChangesResult {
...
 /// component changes.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Default)]
 pub struct BlockEntityChanges {
     extractor: String,
     chain: Chain,
     pub block: Block,
     pub revert: bool,
-    pub state_updates: Vec<ProtocolStateDeltasWithTx>,
+    pub state_updates: Vec<ProtocolChangesWithTx>,
```
louise-poole: This can also get a name change - it's no longer just state updates here.

tycho-indexer/src/extractor/evm/mod.rs
tvinagre: Yeah I'll do that in a separate PR though

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -1039,12 +1159,13 @@ impl BlockEntityChanges {
...
 
-        let aggregated_states = self
-            .state_updates
-            .iter()
-            .try_fold(base, |mut acc_state, new_state| {
+        // Use unwrap_or_else to provide a default state if iter.next() is None
+        let first_state = iter.next().unwrap_or_default();
+
+        let aggregated_changes = iter
+            .try_fold(first_state, |mut acc_state, new_state| {
```
kayibal, line 1168: Looks like it could be done with reduce but not a blocker.

PR #112, opened 2024-01-29
Split Tycho into crates

kayibal:
The main motivation for this PR was to factor out a tycho-client crate that can be used by protosim.

A nice benefit of this is that we also get faster compilation times as a crate is only compiled again if there were any changes in it.

### DIFF
--- a/tycho-indexer/src/extractor/evm/storage.rs
+++ b/tycho-indexer/src/extractor/evm/storage.rs
@@ -27,8 +27,9 @@ pub mod pg {
                 orm,
                 orm::{NewToken, Token},
             },
-            Address, Balance, BlockHash, ChangeType, Code, StorableProtocolComponent,
-            StorableProtocolState, StorableProtocolType, StorableToken, TxHash,
+            Address, Balance, BlockHash, ChangeType, Code, StorableComponentBalance,
+            StorableProtocolComponent, StorableProtocolState, StorableProtocolType, StorableToken,
+            TxHash,
         },
     };
     use ethers::types::{H160, H256, U256};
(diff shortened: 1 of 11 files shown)


tycho-client/src/lib.rs
kayibal, line 1: This crate is already included here but it is not being used atm. It is still very much WIP.

tycho-indexer/src/services/rpc.rs
kayibal, line 1: The types I removed from here are not imported from tycho-types. I also added utoipa schema definitions to them to the OpenAPI spec derivation still works.

tycho-types/src/serde_primitives.rs
```diff
@@ -0,0 +1,202 @@
...
+            Some(val) => decode_hex_with_prefix(&val)
+                .map(Into::into)
+                .map(Some)
+                .map_err(|e| serde::de::Error::custom(e.to_string())),
+            None => Ok(None),
+        }
+    }
+}
+
+pub mod hex_hashmap_key {
```
kayibal, line 82: We need this and the next one because we decode/serialize to bytes directly not via U256 anymore.

tycho-indexer/src/extractor/evm/storage.rs
```diff
@@ -659,8 +646,8 @@ mod test {
             contract_ids: vec![H160::from_low_u64_be(2), H160::from_low_u64_be(3)],
             static_attributes: {
                 let mut map = HashMap::new();
-                map.insert("key1".to_string(), Bytes::from(bytes::Bytes::from("value1")));
-                map.insert("key2".to_string(), Bytes::from(bytes::Bytes::from("value2")));
+                map.insert("key1".to_string(), Bytes::from("badbabe1"));
```
tvinagre, line 677: ❤️

PR #118, opened 2024-01-30
ENG-2045: Implement get_balance_deltas

tamaralipows:
[Issue (no additional info)](https://datarevenue.atlassian.net/browse/ENG-2045)

Note: Code inspiration is mostly taken from `get_balance_deltas` in `contract_state.rs`, which in turn is used in `get_accounts_delta`. Please make sure you understand those methods if you are confused about anything in this PR.

**Questions:**
Is this the correct type that we want to return? It is currently just a hashmap from the protocol ids and token ids to the correct balance, similar to `contract_state.rs`

**TODO:**
- [x] Test forward case
- [x] This test fails only in CI ^ fix it.
- [x] Test backward case

### DIFF
--- a/tycho-indexer/src/storage/postgres/protocol.rs
+++ b/tycho-indexer/src/storage/postgres/protocol.rs
@@ -1561,7 +1561,7 @@ mod test {
             .expect("Failed to fetch protocol component id");
         let token_id = schema::token::table
             .filter(schema::token::symbol.eq("WETH"))
-            .select(schema::token::account_id)
+            .select(schema::token::id)
             .first::<i64>(&mut conn)
             .await
             .expect("Failed to fetch token id");


tycho-indexer/src/storage/postgres/mod.rs
```diff
@@ -965,6 +965,34 @@ pub mod db_fixtures {
...
+        tx_id: i64,
+        protocol_component_id: i64,
+    ) {
+        let ts: NaiveDateTime = schema::transaction::table
+            .inner_join(schema::block::table)
+            .filter(schema::transaction::id.eq(tx_id))
+            .select(schema::block::ts)
+            .first::<NaiveDateTime>(conn)
+            .await
+            .expect("setup tx id not found");
```
SmartMonkey-git, line 982: I'm sure you added this 'except', because of my PR, but you should only ever use 'except' when you are sure it will never be triggered. If that is the case here, go on; if not, please use the 'map_err' function. :)

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -777,6 +777,97 @@ where
         Ok(())
     }
 
+    #[instrument(skip(self, conn))]
```
SmartMonkey-git, line 848: What does this do? :O

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -777,6 +777,97 @@ where
...
+            // Going forward
+            //                  ]     changes to update   ]
+            // -----------------|--------------------------|
+            //                start                     target
+            // We query for balance updates between start and target version.
+            let changed_component_balances = component_balance
+                .inner_join(schema::protocol_component::table.inner_join(schema::chain::table))
+                .filter(schema::chain::id.eq(chain_id))
+                .filter(valid_from.gt(start_ts))
+                .filter(valid_from.le(target_ts))
```
SmartMonkey-git: Little syntactical sugar.

```suggestion
    .filter(
        schema::chain::id.eq(chain_id)
        .and(valid_from.gt(start_ts))
        .and(valid_from.le(target_ts))
    )
```

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -777,6 +777,97 @@ where
...
+                .order_by((
+                    protocol_component_id,
+                    token_id,
+                    valid_from.desc(),
+                    schema::transaction::index.desc(),
+                ))
+                .distinct_on((protocol_component_id, token_id))
+                .get_results::<(i64, i64, Balance)>(conn)
+                .await?
+                .into_iter()
```
SmartMonkey-git, line 898: That is interesting. Other uses of `get_result` call `map_err` after the `await` how are you able to just call `into_iter()`?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -777,6 +777,97 @@ where
...
+            //                target                     start
+            // We query for the previous values of all (protocol_component, token) pairs updated
+            // between start and target version.
+            let changed_component_balances = component_balance
+                .inner_join(schema::protocol_component::table.inner_join(schema::chain::table))
+                .filter(schema::chain::id.eq(chain_id))
+                .filter(valid_from.gt(target_ts))
+                .filter(valid_from.le(start_ts))
+                .select((protocol_component_id, token_id))
+                .distinct();
```
SmartMonkey-git: Instead of having mostly duplicate code in your if-statements, maybe you can use the fact that queries can be stored in a variable and augmented down the road.

For example:
```suggestion
            let mut query = component_balance
                .inner_join(schema::protocol_component::table.inner_join(schema::chain::table))
                .filter(schema::chain::id.eq(chain_id));
                
                if start_ts <= target_ts{
                        query = query.filter(valid_from.gt(start_ts))
                        .filter(valid_from.le(target_ts))
                }
                else {
                         query = query.filter(valid_from.gt(target_ts))
                        .filter(valid_from.le(start_ts))
                }
                query.select((protocol_component_id, token_id))
                .distinct();
```

Its not fully thought through and I haven't checkt, if you can also do that with your second query, but you might be able to eliminate some excess code.

SmartMonkey-git: Omg, the formatting of the suggestion change is atrocious.

tycho-indexer/src/storage/postgres/mod.rs
```diff
@@ -965,6 +965,34 @@ pub mod db_fixtures {
         }
     }
 
+    // Insert a new Component Balance
+    pub async fn insert_component_balance(
```
dianacarvalho1, line 969: I have an overall question: do we prefer insert or add? because in some places we use one and in others the other

tycho-indexer/src/storage/postgres/mod.rs
dianacarvalho1: these are the db fixtures.. I think it's fine

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -777,6 +777,97 @@ where
         Ok(())
     }
 
+    #[instrument(skip(self, conn))]
+    async fn get_balance_deltas(
```
dianacarvalho1, line 849: I'm overall confused... we would want to get ALL balances for a certain chain? why?

tycho-indexer/src/storage/postgres/mod.rs
SmartMonkey-git: > these are the db fixtures.. I think it's fine

Ah, sorry. I missed that!

tycho-indexer/src/storage/postgres/mod.rs
SmartMonkey-git: Good question.
I think insert is associated with databases, while add points more towards an array.

tycho-indexer/src/storage/postgres/protocol.rs
dianacarvalho1: aaah ok I think this is necessary for a revert right??

tamaralipows: Yes exactly. And in that case we don't just want to get the balances for all protocol components because that would be passing around a lot of data. Instead, we want the balances that changed only (hence the query for all balances between start and end time).

tycho-indexer/src/storage/postgres/mod.rs
tamaralipows: i would be down for a whole renaming PR

tycho-indexer/src/storage/postgres/protocol.rs
tamaralipows: Yeah I understand it will be nice for reducing repeated code, but I feel this may be less readable overall because we would then need two mini if blocks as opposed to just one big one

tycho-indexer/src/storage/postgres/protocol.rs
tamaralipows: Don't understand this question - doesn't the `?` unwrap it?

tycho-indexer/src/storage/postgres/protocol.rs
tamaralipows: Took this from `get_balance_deltas` in `contract_state.rs`. Here is an explanation :)

> #[instrument]: This is the main attribute for instrumentation in the context of the tracing crate. It enables automatic generation of tracing events for the annotated item.

> skip(self, conn): This is a method provided by the instrument attribute. It is used to specify which arguments should be skipped or omitted from the generated instrumentation code. In this case, it's skipping the self (implicitly the first argument in a method) and conn arguments.

tycho-indexer/src/storage/postgres/protocol.rs
SmartMonkey-git: Ups, yes

tycho-indexer/src/storage/mod.rs
```diff
@@ -756,12 +756,32 @@ pub trait ProtocolGateway {
...
+    /// # Return
+    /// A hashmap containing a map of (protocol_component_id, token_id) to its respective balance
+    /// change, Err if no changes were found.
+    async fn get_balance_deltas(
+        &self,
+        chain: &Chain,
+        start_version: &BlockOrTimestamp,
+        target_version: &BlockOrTimestamp,
+        conn: &mut Self::DB,
+    ) -> Result<HashMap<(i64, i64), Balance>, StorageError>;
```
kayibal, line 777: I assume that this (database ids leaked) was already taken care of in another PR?

tycho-indexer/src/storage/postgres/mod.rs
```diff
@@ -965,6 +965,34 @@ pub mod db_fixtures {
...
+        conn: &mut AsyncPgConnection,
+        balance: Balance,
+        token_id: i64,
+        tx_id: i64,
+        protocol_component_id: i64,
```
kayibal, line 974: Again db details leaked?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -845,6 +845,104 @@ where
...
+                .inner_join(schema::protocol_component::table.inner_join(schema::chain::table))
+                .filter(
+                    schema::chain::id
+                        .eq(chain_id)
+                        .and(valid_from.gt(start_ts))
+                        .and(valid_from.le(target_ts)),
+                )
+                .select((protocol_component_id, token_id))
+                .distinct();
+
```
kayibal, line 878: Why do you split the query here? Looks like you wanted to use a subquery...

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -845,6 +845,104 @@ where
...
+            let result: HashMap<(i64, i64), Balance> = changed_component_balances
+                .inner_join(schema::transaction::table)
+                .filter(
+                    valid_from.le(target_ts).and(
+                        valid_to
+                            .gt(target_ts)
+                            .or(valid_to.is_null()),
+                    ),
+                )
+                .select((protocol_component_id, token_id, new_balance))
```
kayibal, line 888: Are you aware that the above query already had a select clause? You are just overriding it here... I think this query does not do what you expected it to do.

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -845,6 +845,104 @@ where
...
+
+            let result: HashMap<(i64, i64), Balance> = changed_component_balances
+                .inner_join(schema::transaction::table)
+                .filter(
+                    valid_from.le(target_ts).and(
+                        valid_to
+                            .gt(target_ts)
+                            .or(valid_to.is_null()),
+                    ),
+                )
```
kayibal, line 887: Why do we need to version filter things twice? Again seems like you wanted to do a subquery join subquery. Which btw is not the right way to approach this.

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -845,6 +845,104 @@ where
...
+
+            let result: HashMap<(i64, i64), Balance> = changed_component_balances
+                .inner_join(schema::transaction::table)
+                .filter(valid_from.le(target_ts))
+                .filter(
+                    valid_to
+                        .gt(target_ts)
+                        .or(valid_to.is_null()),
+                )
+                .select((protocol_component_id, token_id, new_balance))
```
kayibal, line 928: Apart from this query suffering from the same fundamental issues like the above one, you need to select the previous balance here - not the new one, else your deltas will have the last balance of the record that got reverted... so I think you need a previous balance on your table.

tycho-indexer/src/storage/postgres/mod.rs
tamaralipows: this is just a db fixture for the test? why does that matter

PR #119, opened 2024-01-30
ENG-2734: Retrieve Token and Contract Addresses via junction tables when querying ProtocolComponents

SmartMonkey-git:
[Task](https://datarevenue.atlassian.net/browse/ENG-2734)
After inserting the tokens and contracts related to the ProtocolComponent, we now also retrieve them.

### DIFF
--- a/tycho-indexer/src/storage/postgres/protocol.rs
+++ b/tycho-indexer/src/storage/postgres/protocol.rs
@@ -1194,15 +1194,6 @@ mod test {
             Some(vec![contract_code_id]),
         )
         .await;
-        db_fixtures::insert_protocol_component(
-            conn,
-            "state3",
-            chain_id,
-            protocol_system_id_ambient,
-            protocol_type_id,
-            txn[0],
-        )
-        .await;
 
         // protocol state for state1-reserve1
         db_fixtures::insert_protocol_state(


tycho-indexer/src/extractor/evm/storage.rs
```diff
@@ -386,13 +386,25 @@ pub mod pg {
                 Default::default()
             };
 
+            fn convert_addresses_to_h160(addresses: &[Address]) -> Result<Vec<H160>, StorageError> {
```
dianacarvalho1: maybe this function could be in a utils somewhere?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -155,14 +155,77 @@ where
...
+        let protocol_component_contracts = map_addresses(protocol_component_contracts);
+
         orm_protocol_components
             .into_iter()
             .map(|(pc, tx_hash)| {
                 let ps = self.get_protocol_system(&pc.protocol_system_id);
+                let default_vec = vec![];
+                let tokens_by_pc = protocol_component_tokens
+                    .get(&pc.id)
+                    .unwrap_or(&default_vec);
```
dianacarvalho1: there should always be something here no? why would we allow an empty vec? 🤔

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -155,14 +155,77 @@ where
...
+                    schema::account::address,
+                ))
+                .filter(
+                    schema::protocol_component_holds_contract::protocol_component_id
+                        .eq_any(protocol_component_ids),
+                )
+                .load::<(i64, Address)>(conn)
+                .await?;
+
+        fn map_addresses(
```
dianacarvalho1: this is more map addresses per protocol component

tycho-indexer/src/storage/postgres/mod.rs
```diff
@@ -1005,11 +1008,51 @@ pub mod db_fixtures {
...
+                                    .eq(component_id),
+                                schema::protocol_component_holds_contract::contract_code_id
+                                    .eq(cc_id),
+                            )
+                        })
+                        .collect::<Vec<_>>(),
+                )
+                .execute(conn)
+                .await
+                .expect("protocol component holds contract code insert ok");
```
tamaralipows, line 1071: Shouldn't this be the panic/error message of the insert?

tycho-indexer/src/storage/postgres/mod.rs
```diff
@@ -1005,11 +1008,51 @@ pub mod db_fixtures {
...
+                                schema::protocol_component_holds_token::protocol_component_id
+                                    .eq(component_id),
+                                schema::protocol_component_holds_token::token_id.eq(t_id),
+                            )
+                        })
+                        .collect::<Vec<_>>(),
+                )
+                .execute(conn)
+                .await
+                .expect("protocol component holds token insert ok");
```
tamaralipows, line 1051: Same here

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -155,14 +155,78 @@ where
...
         orm_protocol_components
             .into_iter()
             .map(|(pc, tx_hash)| {
                 let ps = self.get_protocol_system(&pc.protocol_system_id);
+                let tokens_by_pc: &Vec<Address> = protocol_component_tokens
+                    .get(&pc.id)
+                    .expect("Could not find Protocol Component.");
+                let contracts_by_pc: &Vec<Address> = protocol_component_contracts
+                    .get(&pc.id)
+                    .expect("Could not find Protocol Component.");
```
tamaralipows: Maybe specify that this is in the protocol component contracts, and the one above is in the protocol component tokens?

tycho-indexer/src/storage/postgres/mod.rs
tamaralipows: Just realized it's done this way literally everywhere in the code. I am outnumbered lol. Resolving this comment

SmartMonkey-git: We only use these under the assumption that they will never trigger. Any other time it should indeed be handled by results.

PR #120, opened 2024-01-30
ENG-2738: Add protocol components and tokens methods to cached gateway

dianacarvalho1:
[Issue](https://datarevenue.atlassian.net/browse/ENG-2738)

### DIFF
--- a/tycho-indexer/migrations/2024-01-10_tycho_patch/down.sql
+++ b/tycho-indexer/migrations/2024-01-10_tycho_patch/down.sql
@@ -43,11 +43,9 @@ $$
 LANGUAGE plpgsql;
 
 ALTER TABLE protocol_state
-    DROP CONSTRAINT check_attribute_fields;
-
-ALTER TABLE protocol_state
     DROP COLUMN attribute_name,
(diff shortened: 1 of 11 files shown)


tycho-indexer/src/storage/postgres/cache.rs
louise-poole, line 1: I assume on revert we need to get balance deltas correct? If so - I have adjusted the cached deltas logic [here](https://github.com/propeller-heads/tycho-indexer/pull/116). Let's first get some reviews there to see if I did it correctly, then add balance to it? (this is only needed if balances are emitted to our ws subscribers - I'm not sure if they're just an internal thing?)

dianacarvalho1: right now they are only an internal thing for TVL I think 🤔

dianacarvalho1: ah no, I think you are right @louise-poole ! But then this is blocked by @tamaralipows 's [PR](https://github.com/propeller-heads/tycho-indexer/pull/118) I think

PR #123, opened 2024-01-31
Update native protocol structs

louise-poole:
The native protocol extractor structs were a bit behind on the latest changes - component balances, splitting the protocol components by change (new/deleted) etc. This is to get them on par with the vm struct changes.


tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -920,7 +923,8 @@ impl BlockEntityChanges {
                 chain,
                 block,
                 state_updates,
-                new_protocol_components,
+                protocol_components,
+                component_balances: Vec::new(),
```
dianacarvalho1, line 927: oh we are missing this in the other objects too. gonna add a new task for this!

PR #124, opened 2024-01-31
ENG-2738: Add get tokens endpoint to RPC

dianacarvalho1:
Questions:
- Should the endpoint return all tokens if no addresses are passed? I want to say yes

### DIFF
--- a/tycho-types/src/dto.rs
+++ b/tycho-types/src/dto.rs
@@ -369,6 +369,7 @@ impl StateRequestParameters {
 #[derive(Serialize, Deserialize, Debug, Default, PartialEq, ToSchema)]
 pub struct TokensRequestBody {
     #[serde(rename = "tokenAddresses")]
+    #[schema(value_type=Option<Vec<String>>)]
     pub token_addresses: Option<Vec<Bytes>>,
 }
 


tycho-types/src/dto.rs
```diff
@@ -366,6 +366,65 @@ impl StateRequestParameters {
     }
 }
 
+#[derive(Serialize, Deserialize, Debug, Default, PartialEq, ToSchema)]
```
dianacarvalho1, line 369: I'm not so sure we need all of these structs. @zizou0x maybe you can advice me here? 🙏🏼

tycho-indexer/src/services/rpc.rs
```diff
@@ -168,6 +182,55 @@ impl RpcHandler {
...
+        request: &dto::TokensRequestBody,
+    ) -> Result<dto::TokensRequestResponse, RpcError> {
+        let mut conn = self.db_connection_pool.get().await?;
+
+        info!(?chain, ?request, "Getting tokens.");
+        self.get_tokens_inner(chain, request, &mut conn)
+            .await
+    }
+
+    async fn get_tokens_inner(
```
tamaralipows, line 201: inner? what does that mean?

tycho-indexer/src/services/rpc.rs
```diff
@@ -424,4 +519,59 @@ mod tests {
...
+                USDT.parse::<Bytes>().unwrap(),
+            )]),
+        };
+
+        let tokens = req_handler
+            .get_tokens_inner(&Chain::Ethereum, &request, &mut conn)
+            .await
+            .unwrap();
+
+        assert_eq!(tokens.tokens.len(), 0);
```
tamaralipows, line 574: I think I could use a small explanation huddle for this PR tomorrow (don't let this block you from merging though). The flow isn't totally clear to me when it comes down to the actual code? I thought that we would want to get from the RPC if the token is not in the db, though not sure what exactly this is test is doing then

tycho-types/src/dto.rs
zizou0x: I think we do

tycho-types/src/dto.rs
```diff
@@ -366,6 +366,65 @@ impl StateRequestParameters {
...
+pub struct TokensRequestBody {
+    #[serde(rename = "contractIds")]
+    pub contract_ids: Option<Vec<ContractId>>,
```
zizou0x: What happens if it's None?

tycho-types/src/dto.rs
```diff
@@ -366,6 +366,65 @@ impl StateRequestParameters {
...
+#[serde(rename = "Token")]
+/// Token struct for the response from Tycho server for a tokens request.
+pub struct ResponseToken {
+    pub chain: Chain,
+    #[schema(value_type=String, example="0xc9f2e6ea1637E499406986ac50ddC92401ce1f58")]
+    #[serde(with = "hex_bytes")]
+    pub address: Bytes,
+    #[schema(value_type=String, example="WETH")]
+    pub symbol: String,
+    #[schema(value_type=HashMap<String, String>, example="0xBADBABE")]
```
zizou0x: I don't think we need this line

tycho-indexer/src/services/mod.rs
```diff
@@ -104,6 +104,10 @@ impl ServicesBuilder {
                     web::resource(format!("/{}/{{execution_env}}/contract_state", self.prefix))
                         .route(web::post().to(rpc::contract_state)),
                 )
+                .service(
```
zizou0x, line 111: You will need to add schemas to the `components` above in the openapi attribute. Also add your new path please.

It's really annoying that it doesn't error at compile time 😩. Please use `examples/rpc.rs` and visit https://0.0.0.0:4242/docs/ to see if openAPI is not broken.

tycho-indexer/src/services/rpc.rs
```diff
@@ -168,6 +182,55 @@ impl RpcHandler {
...
+        request: &dto::TokensRequestBody,
+        db_connection: &mut AsyncPgConnection,
+    ) -> Result<dto::TokensRequestResponse, RpcError> {
+        let contract_ids = request.contract_ids.clone();
+        let addresses: Option<Vec<Address>> = contract_ids.map(|ids| {
+            ids.into_iter()
+                .map(|id| Address::from(id.address))
+                .collect()
+        });
+
```
zizou0x: Something feels a bit weird here. We receive `Option<Vec<ContractId>>` but then we only use the address of it because the chain is now part of the url right? So maybe we could directly use `Option<Vec<Address>>`?

tycho-indexer/src/services/rpc.rs
```diff
@@ -168,6 +182,55 @@ impl RpcHandler {
...
+            .await
+        {
+            Ok(tokens) => Ok(dto::TokensRequestResponse::new(
+                tokens
+                    .into_iter()
+                    .map(dto::ResponseToken::from)
+                    .collect(),
+            )),
+            Err(err) => {
+                error!(error = %err, "Error while getting contract states.");
```
zizou0x:
```suggestion
                error!(error = %err, "Error while getting tokens.");
```

tycho-indexer/src/services/rpc.rs
dianacarvalho1: yeah.. it's just a passthrough method to make testing easier 😕 I don't like it much either

tycho-indexer/src/services/rpc.rs
dianacarvalho1: oooh giiiirrlll that's a completely different task 😂 this is our RPC. This is like an API endpoint to get tokens from our DB

tycho-indexer/src/services/rpc.rs
dianacarvalho1: upsi

tycho-indexer/src/services/rpc.rs
dianacarvalho1: yeah that's true. I will simplify it

tycho-types/src/dto.rs
dianacarvalho1: it returns all tokens

tycho-indexer/src/services/rpc.rs
tamaralipows: ufffffffff

tycho-types/src/dto.rs
```diff
@@ -366,6 +366,58 @@ impl StateRequestParameters {
...
+pub struct ResponseToken {
+    pub chain: Chain,
+    #[schema(value_type=String, example="0xc9f2e6ea1637E499406986ac50ddC92401ce1f58")]
+    #[serde(with = "hex_bytes")]
+    pub address: Bytes,
+    #[schema(value_type=String, example="WETH")]
+    pub symbol: String,
+    pub decimals: u32,
+    pub tax: u64,
+    pub gas: Vec<Option<u64>>,
```
kayibal, line 406: Type is questionable but won't block for now I think.

tycho-indexer/src/services/rpc.rs
```diff
@@ -202,6 +259,34 @@ pub async fn contract_state(
     }
 }
 
+#[utoipa::path(
+    post,
+    path = "/v1/{execution_env}/tokens",
```
kayibal, line 268: Oh god I think I made a mistake here by asking you to add this as a path parameter. I just recently decided that all chains will use the same models - but ok we'll change it later.

dianacarvalho1: oh nooooo 😐

tycho-indexer/src/services/mod.rs
dianacarvalho1: AG ✨

tycho-types/src/dto.rs
dianacarvalho1: why?

PR #125, opened 2024-01-31
Tycho client library

kayibal:
Adds a library to interface with tycho via HTTP and Websocket.

We will use this to build our new data feed for the NTQ and later also for the solver (if I can manage to wrap this with PyO3):

https://datarevenue.atlassian.net/browse/ENG-2596

### DIFF
--- a/tycho-client/src/deltas.rs
+++ b/tycho-client/src/deltas.rs
@@ -140,6 +140,7 @@ enum SubscriptionInfo {
     RequestedUnsubscription(oneshot::Sender<()>),
 }
 
+/// Internal struct containing shared state between of WsDeltaClient instances.
 struct Inner {
     /// Websocket sender handle.
     sink: WebSocketSink,
@@ -184,7 +185,7 @@ impl Inner {
 
     /// Transitions a pending subscription to active.
     ///
-    /// Will ignore andy request to do so for subscriptions that are not pending.
+    /// Will ignore any request to do so for subscriptions that are not pending.
     fn mark_active(&mut self, extractor_id: &ExtractorIdentity, subscription_id: Uuid) {
         if let Some(info) = self.pending.remove(extractor_id) {
             if let SubscriptionInfo::RequestedSubscription(ready_tx) = info {
@@ -278,7 +279,7 @@ impl Inner {
         } else {
             error!(
                 ?subscription_id,
-                "Received `SubscriptionEnded` but the never subscribed 
+                "Received `SubscriptionEnded`, but was never subscribed 
                 to it. This is likely a bug!"
             );
         }
@@ -332,7 +333,7 @@ impl WsDeltasClient {
 
     /// Waits for the client to be connected
     ///
-    /// This method will acquires the lock for inner for a short period then waits until the
+    /// This method acquires the lock for inner for a short period, then waits until the  
     /// connection is established if not already connected.
     async fn ensure_connection(&self) {
         if !self.is_connected().await {
@@ -451,7 +452,6 @@ impl WsDeltasClient {
 
 #[async_trait]
 impl DeltasClient for WsDeltasClient {
-    #[allow(unused_variables)]
     #[instrument(skip(self))]
     async fn subscribe(
         &self,
--- a/tycho-client/src/rpc.rs
+++ b/tycho-client/src/rpc.rs
@@ -88,6 +88,7 @@ impl RPCClient for HttpRPCClient {
             chain,
             filters.to_query_string()
         );
+        dbg!(&uri);
         debug!(%uri, "Sending contract_state request to Tycho server");
         let body =
             serde_json::to_string(&request).map_err(|e| RPCError::FormatRequest(e.to_string()))?;
--- a/tycho-types/src/dto.rs
+++ b/tycho-types/src/dto.rs
@@ -363,7 +363,7 @@ impl StateRequestParameters {
         }
 
         let mut res = parts.join("&");
-        if res.len() > 0 {
+        if !res.is_empty() {
             res = format!("?{res}");
         }
         res
--- a/tycho-indexer/src/storage/postgres/schema.rs
+++ b/tycho-indexer/src/storage/postgres/schema.rs
@@ -171,6 +171,15 @@ diesel::table! {
 }
 
 diesel::table! {
+    protocol_component_holds_contract (protocol_component_id, contract_code_id) {
+        protocol_component_id -> Int8,
+        contract_code_id -> Int8,
+        inserted_ts -> Timestamptz,
+        modified_ts -> Timestamptz,
+    }
+}
+
+diesel::table! {
     protocol_component_holds_token (protocol_component_id, token_id) {
         protocol_component_id -> Int8,
         token_id -> Int8,
@@ -248,15 +257,6 @@ diesel::table! {
     }
 }
 
-diesel::table! {
-    protocol_component_holds_contract (protocol_component_id, contract_code_id) {
-        protocol_component_id -> Int8,
-        contract_code_id -> Int8,
-        inserted_ts -> Timestamptz,
-        modified_ts -> Timestamptz,
-    }
-}
-
 diesel::joinable!(account -> chain (chain_id));
 diesel::joinable!(account_balance -> account (account_id));
 diesel::joinable!(account_balance -> transaction (modify_tx));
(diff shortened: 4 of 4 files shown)


tycho-client/src/rpc.rs
```diff
@@ -0,0 +1,182 @@
...
+#[async_trait]
+pub trait RPCClient {
+    /// Retrieves a snapshot of contract state.
+    async fn get_contract_state(
+        &self,
+        chain: Chain,
+        filters: &StateRequestParameters,
+        request: &StateRequestBody,
+    ) -> Result<StateRequestResponse, RPCError>;
+}
```
louise-poole, line 44: Is this only for vm protocols? What about native protocols? Or is that to be added later? If so, a TODO here would be nice.

tycho-client/src/deltas.rs
```diff
@@ -0,0 +1,976 @@
...
+    /// Subscribe to an extractor and receive realtime messages
+    ///
+    /// Will request a subscription from tycho and wait for confirmation of it. If the caller
+    /// cancels while waiting for confirmation the subscription may still be registered. If the
+    /// receiver was deallocated though, the first message from the subscription will remove it
+    /// again - since there is no one to inform about these messages.
+    async fn subscribe(
+        &self,
+        extractor_id: ExtractorIdentity,
+    ) -> Result<(Uuid, Receiver<BlockAccountChanges>), DeltasError>;
```
louise-poole, line 96: What about `BlockEntityChanges`? Or maybe we can have the `ExtractorMsg` type here?

tycho-client/src/deltas.rs
```diff
@@ -0,0 +1,976 @@
...
+/// O ---> requested subscribe ----> active ----> requested unsub ---> ended
+/// ```
+///
+/// We use oneshot channels to inform the client struct about when these transition happened. E.g.
+/// because for `subscribe`` to finish, we want the state to have transition to `active` and similar
+/// for `unsubscribe`.
+#[derive(Debug)]
+enum SubscriptionInfo {
+    /// Subscription was requested we wait for server confirmation and uuid assignment.
+    RequestedSubscription(oneshot::Sender<(Uuid, Receiver<BlockAccountChanges>)>),
```
louise-poole, line 136: Same here - this is vm specific, `ExtractorMsg` is probably better?

tycho-client/src/deltas.rs
```diff
@@ -0,0 +1,976 @@
...
+    /// Subscription is active.
+    Active,
+    /// Unsubscription was requested, we wait for server confirmation.
+    RequestedUnsubscription(oneshot::Sender<()>),
+}
+
+struct Inner {
```
louise-poole, line 144: Not such a fan of the oversimplified name here - its hard to figure out its purpose without digging through implementation comments etc. At the same time I don't have a better name to suggest right this moment so... 👀 😂 
EDIT - going through more of the code "Inner" does make sense, but can maybe be slightly more descriptive? Like `InnerClient`?

tycho-client/src/deltas.rs
```diff
@@ -0,0 +1,976 @@
...
+            return Err(DeltasError::SubscriptionAlreadyPending);
+        }
+        self.pending
+            .insert(id.clone(), SubscriptionInfo::RequestedSubscription(ready_tx));
+        Ok(())
+    }
+
+    /// Transitions a pending subscription to active.
+    ///
+    /// Will ignore andy request to do so for subscriptions that are not pending.
```
louise-poole:
```suggestion
    /// Will ignore any request to do so for subscriptions that are not pending.
```

tycho-client/src/deltas.rs
```diff
@@ -0,0 +1,976 @@
...
+                self.sender.insert(subscription_id, tx);
+                self.subscriptions
+                    .insert(subscription_id, SubscriptionInfo::Active);
+                let _ = ready_tx
+                    .send((subscription_id, rx))
+                    .map_err(|_| {
+                        warn!(
+                            ?extractor_id,
+                            ?subscription_id,
+                            "Subscriber for has gone away. Ignoring."
```
louise-poole, line 202: This warning message isn't too clear to me - has the subscriber dropped their connection?
```suggestion
                            "Subscriber has gone away. Ignoring."
```

tycho-client/src/deltas.rs
```diff
@@ -0,0 +1,976 @@
...
+                    "Pending subscription was not in the correct state to 
+                    transition to active. Ignoring!"
+                )
+            }
+        } else {
+            error!(
+                ?extractor_id,
+                ?subscription_id,
+                "Tried to mark an unkown subscription as active. Ignoring!"
+            );
```
louise-poole, line 218: Question: this just logs the error instead of raising one, correct?

tycho-client/src/deltas.rs
```diff
@@ -0,0 +1,976 @@
...
+        }
+    }
+
+    /// Sends a message to a subscriptions receiver.
+    async fn send(&mut self, id: &Uuid, msg: BlockAccountChanges) -> Result<(), DeltasError> {
```
louise-poole, line 223: Vm specific here too 😞

tycho-client/src/deltas.rs
```diff
@@ -0,0 +1,976 @@
...
+                warn!(?subscription_id, "Subscription ended unexpectedly!");
+                self.sender
+                    .remove(&subscription_id)
+                    .expect("sender channel missing");
+            }
+        } else {
+            error!(
+                ?subscription_id,
+                "Received `SubscriptionEnded` but the never subscribed 
+                to it. This is likely a bug!"
```
louise-poole:
```suggestion
                "Received `SubscriptionEnded`, but was never subscribed 
                to it. This is likely a bug!"
```

tycho-client/src/deltas.rs
```diff
@@ -0,0 +1,976 @@
...
+        Ok(Self {
+            uri,
+            inner: Arc::new(Mutex::new(None)),
+            conn_notify: Arc::new(Notify::new()),
+            max_reconnects: 5,
+        })
+    }
+
+    // Construct a new client with a custom number of reconnection attempts.
+    pub fn new_with_reconnects(ws_uri: &str, max_reconnects: u32) -> Result<Self, DeltasError> {
```
louise-poole, line 313: What's the benefit to doing this over having an optional `max_reconnects` argument in the `new` function above?

tycho-client/src/deltas.rs
```diff
@@ -0,0 +1,976 @@
...
+    ///
+    /// This method will acquire the lock for inner.
+    async fn is_connected(&self) -> bool {
+        let guard = self.inner.as_ref().lock().await;
+        guard.is_some()
+    }
+
+    /// Waits for the client to be connected
+    ///
+    /// This method will acquires the lock for inner for a short period then waits until the
```
louise-poole:
```suggestion
    /// This method acquires the lock for inner for a short period, then waits until the
```

tycho-client/src/deltas.rs
```diff
@@ -0,0 +1,976 @@
...
+                serde_json::to_string(&cmd).expect("serialize cmd encode error"),
+            ))
+            .await?;
+        Ok(())
+    }
+}
+
+#[async_trait]
+impl DeltasClient for WsDeltasClient {
+    #[allow(unused_variables)]
```
louise-poole: Are there unused variables here that this applies to? I think this can likely be removed..

tycho-client/src/deltas.rs
```diff
@@ -0,0 +1,976 @@
...
+
+                            warn!(
+                                ?error,
+                                ?retry_count,
+                                "Connection dropped unexpectedly; Reconnecting"
+                            );
+                            break
+                        } else {
+                            // Other errors are considered fatal
+                            break 'retry;
```
louise-poole, line 559: Interesting! Learnt something new here - labelled loops 👀

tycho-types/src/dto.rs
```diff
@@ -76,7 +76,7 @@ pub enum Response {
 #[derive(Serialize, Deserialize, Debug)]
 #[serde(untagged)]
 pub enum WebSocketMessage {
-    BlockAccountChanges(BlockAccountChanges),
+    BlockAccountChanges { subscription_id: Uuid, data: BlockAccountChanges },
     Response(Response),
 }
```
louise-poole, line 81: Hmmmm -  we probably need `BlockEntityChangesResult` here too.

tycho-client/src/rpc.rs
kayibal: Yes only vm protocols now, I ported this from protosim repo. So old code. But it should be easy to extend this to also support native protocol messages / rpc endpoints.

tycho-client/src/deltas.rs
kayibal: Hmm probably can work with an enum here to support both message types.

tycho-client/src/deltas.rs
kayibal: Inner usually means some inner state that is shared via synchronisation primitives - or there is another reason for separating state. I can rename this to `InnerClient` but I think it won't help much?

tycho-client/src/deltas.rs
kayibal: Yes when someone calls subscribe we create a oneshot channel this channel is used to notify the caller that the subscription is now active. If this caller dies or panics or for any other reasons manages to close that channel, this case can happen.

tycho-client/src/deltas.rs
kayibal: Yes

tycho-client/src/deltas.rs
kayibal: I am aware this is currently vm specific but it's not a big deal we will just wrap the message into an enum and everything will work as it used to :D

tycho-client/src/deltas.rs
kayibal: Lazyness :D of the one who calls the constructor.

tycho-client/src/deltas.rs
louise-poole: Cool - I don't feel too strongly about this. Was confused at first glance but am happy with it after going through the code.

tycho-types/src/dto.rs
louise-poole: I see this is planned for a later task 👍

PR #126, opened 2024-01-31
Extract protocol type from substreams instead of hardcode it

tvinagre:
Context:
Whenever we parse Tycho's `BlockContractChanges` we need to build `ProtocolComponent` (by calling a chain of `try_from_message` methods).

Currently, on Ambient, the `protocol_type` is being passed as a value from the `AmbientExtractor`. This is undesirable, because leaves us with a hardcoded value for `protocol_type` and unifyies multiple `protocol_types` under a single type.

This PR makes the following changes:
1. Extract `protocol_type_name` from Substream's message `ProtocolComponent.protocol_type.name`
2. Validate if `protocol_type_name` is a subset of allowed protocol_type names specified when initializing the ProtocolExtractor
3. Raise an error if, by any chance, `ProtocolComponent.protocol_type` is `None`, as every Protobuf field can be nullified

### DIFF
--- a/substreams/ethereum-ambient/src/contracts/main.rs
+++ b/substreams/ethereum-ambient/src/contracts/main.rs
@@ -146,7 +146,7 @@ pub fn decode_pool_init(call: &Call) -> Result<Option<ProtocolComponent>, anyhow
                     static_att: vec![static_attribute],
                     change: crate::pb::tycho::evm::v1::ChangeType::Creation.into(),
                     protocol_type: Some(crate::pb::tycho::evm::v1::ProtocolType {
-                        name: "Ambient".to_string(),
+                        name: "ambient_pool".to_string(),
                         attribute_schema: vec![],
                         financial_type: 0,
                         implementation_type: 0,


tycho-indexer/src/extractor/evm/ambient.rs
```diff
@@ -423,7 +425,7 @@ mod test {
         ambient_protocol_types.insert(
             "vm:pool".to_string(),
             ProtocolType::new(
-                "ambient_pool".to_string(),
+                "WeightedPool".to_string(),
```
dianacarvalho1, line 428: why? 👀

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -527,7 +527,7 @@ impl ProtocolComponent {
         msg: substreams::ProtocolComponent,
         chain: Chain,
         protocol_system: &str,
-        protocol_type_name: &str,
+        protocol_type_names: Vec<String>,
```
kayibal, line 530: @tvinagre sorry but why so complicated :D Just pass `&HashMap<String, ProtocolType>` in here. 

Yeah we may not use the ProtocolType value of the map but:
- no clones
- O(1) lookup vs O(n) for a Vec
- If we ever want to have the full protocol type vs just the name on ProtocolComponent it is an easy change.

tvinagre: Yeah you're right - that was not smart 😅 
Fixing!

tycho-indexer/src/extractor/evm/ambient.rs
tvinagre: Ah it's just a test fixture - it's to follow the same format as the other fixtures

tycho-indexer/src/extractor/evm/mod.rs
tvinagre: Actually, just noticed. The reason is because the value of `ProtocolType.name` is in the Vec, not the HashMap keys. 
And that's what we expect, no? That the `ProtocolType.name` is an "allowed" value

PR #128, opened 2024-02-01
ENG-2755: Adjust AccountUpdateWithTx

SmartMonkey-git:


tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -371,8 +378,8 @@ pub struct BlockContractChanges {
     chain: Chain,
     pub block: Block,
     pub tx_updates: Vec<AccountUpdateWithTx>,
-    pub protocol_components: Vec<ProtocolComponent>,
-    pub component_balances: Vec<ComponentBalance>,
+    pub protocol_components: Vec<ProtocolComponentsWithTx>,
+    pub component_balances: Vec<ComponentBalancesWithTx>,
```
louise-poole: We also thought about taking this approach (discussed it [here](https://propellerswap.slack.com/archives/C05LQALE9L1/p1706272378050749)), however decided that this leads to duplicated data being passed and we would prefer to only pass a txn once and have all changes linked to it passed with it.

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -290,26 +341,24 @@ impl NormalisedMessage for BlockAccountChanges {
...
 /// Updates grouped by their respective transaction.
 #[derive(Debug, Clone, PartialEq)]
-pub struct AccountUpdateWithTx {
+pub struct TransactionUpdates {
     // TODO: for ambient it works to have only a single update here but long
     // term we need to be able to store changes to multiple accounts per
     // transactions.
-    pub update: AccountUpdate,
+    pub account_updates: Vec<AccountUpdate>,
+    pub protocol_componets: Vec<ProtocolComponent>,
```
dianacarvalho1: components

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -351,15 +400,78 @@ impl AccountUpdateWithTx {
...
+        self.account_updates
+            .into_iter()
+            .map(|update| (update.address, update))
+            .collect()
     }
 }
 
-impl Deref for AccountUpdateWithTx {
-    type Target = AccountUpdate;
+impl Deref for TransactionUpdates {
```
SmartMonkey-git: Should we keep this?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -143,27 +142,31 @@ impl From<&AccountUpdateWithTx> for Account {
...
+            .clone()
+            .into_iter()
+            .map(|update| {
+                let acc = Account::new(
+                    update.chain,
+                    update.address,
+                    format!("{:#020x}", update.address),
+                    update.slots.clone(),
+                    update.balance.unwrap_or_default(),
+                    update.code.clone().unwrap_or_default(),
```
louise-poole: Are the clones here necessary?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -190,6 +193,54 @@ impl AccountUpdate {
         Self { address, chain, slots, balance, code, change }
     }
 
+    // Converting into via self as references saves us from cloning the whole struct of
+    // BlockContractChanges in forward.
```
louise-poole: I'm not sure this comment make sense to me... Maybe we can just say "Using a self reference save us from cloning... etc". Also why mention the BlockContractChanges struct? And what do you mean by "in forward"? I think we can just mention the "AccountUpdates" struct here and then it should be clear to anyone using this struct and conversion function?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -190,6 +193,54 @@ impl AccountUpdate {
...
+    pub fn ref_into_account(&self, tx: &Transaction) -> Account {
```
louise-poole, line 196: Also, could we not maybe implement something like 
```
fn from(value: &AccountUpdate) -> Self {
```
in the Account struct?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -190,6 +193,54 @@ impl AccountUpdate {
...
+        let code = &self.code.clone().unwrap_or_default();
+        Account::new(
+            self.chain,
+            self.address,
+            format!("{:#020x}", self.address),
+            self.slots.clone(),
+            self.balance.unwrap_or_default(),
+            code.clone(),
```
louise-poole: Wasn't the purpose to avoid cloning?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -290,28 +341,26 @@ impl NormalisedMessage for BlockAccountChanges {
...
+impl TransactionUpdates {
     pub fn new(
-        address: H160,
-        chain: Chain,
-        slots: HashMap<U256, U256>,
-        balance: Option<U256>,
-        code: Option<Bytes>,
-        change: ChangeType,
+        account_updates: Vec<AccountUpdate>,
+        protocol_componets: Vec<ProtocolComponent>,
```
louise-poole:
```suggestion
        protocol_components: Vec<ProtocolComponent>,
```

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -331,7 +380,7 @@ impl AccountUpdateWithTx {
     /// # Errors
     /// This method will return `ExtractionError::MergeError` if any of the above
     /// conditions is violated.
-    pub fn merge(&mut self, other: AccountUpdateWithTx) -> Result<(), ExtractionError> {
+    pub fn merge(&mut self, other: &TransactionUpdates) -> Result<(), ExtractionError> {
         if self.tx.block_hash != other.tx.block_hash {
             return Err(ExtractionError::MergeError(format!(
                 "Can't merge AccountUpdates from different blocks: 0x{:x} != 0x{:x}",
```
louise-poole: You can probably update these error messages since we're merging more than just AccountUpdates now.

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -331,7 +380,7 @@ impl AccountUpdateWithTx {
     /// # Errors
     /// This method will return `ExtractionError::MergeError` if any of the above
     /// conditions is violated.
-    pub fn merge(&mut self, other: AccountUpdateWithTx) -> Result<(), ExtractionError> {
+    pub fn merge(&mut self, other: &TransactionUpdates) -> Result<(), ExtractionError> {
```
louise-poole: I see the protocol components missing from the merge? The two vecs can probably just be appended to each other - we shouldn't have clashes.

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -673,31 +748,35 @@ impl BlockContractChanges {
...
-        for update in self.tx_updates.into_iter() {
-            match account_updates.entry(update.address) {
-                Entry::Occupied(mut e) => {
-                    e.get_mut().merge(update)?;
-                }
-                Entry::Vacant(e) => {
-                    e.insert(update);
-                }
-            }
+    pub fn aggregate_updates(mut self) -> Result<BlockAccountChanges, ExtractionError> {
```
louise-poole: Why are you making this mutable for self? Self should remain unchanged, the returned struct has all the aggregated info.

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -673,31 +748,35 @@ impl BlockContractChanges {
...
+        self.tx_updates
+            .sort_unstable_by_key(|update| update.tx.index);
```
louise-poole: Curious - why did we not sort in the previous version of this function? Could we assume the updates are already in order?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -673,31 +748,35 @@ impl BlockContractChanges {
...
+        let mut tx_update = self
+            .tx_updates
+            .first()
+            .cloned()
+            .ok_or(ExtractionError::Empty)?;
+
+        let mut protocol_components = self
```
louise-poole: Hmmm... okay I see you're dealing with protocol component collecting here... why not just include it in the merge logic though?

tycho-indexer/src/extractor/evm/ambient.rs
```diff
@@ -501,14 +517,7 @@ mod test {
...
 
     fn block_contract_changes_ok() -> BlockContractChanges {
-        let mut data = evm::fixtures::pb_block_contract_changes();
-        // TODO: make fixtures configurable through parameters so they can be
-        // properly reused. Will need fixture to easily assemble contract
-        // change objects.
-        data.changes[0]
-            .contract_changes
-            .remove(1);
-        data
```
louise-poole, line 512: What was this for in the old code? 👀

tycho-indexer/src/extractor/evm/mod.rs
louise-poole: I'm not actually sure what it's for TBH 👀

tycho-indexer/src/extractor/evm/mod.rs
SmartMonkey-git: To construct `Account` you need a `Transaction`, which `AccountUpdate` does not feature.

tycho-indexer/src/extractor/evm/mod.rs
SmartMonkey-git: Yes, cloning the whole BlockContractChanges object in the forward function

tycho-indexer/src/extractor/evm/mod.rs
SmartMonkey-git: This was previously `fn from(value: &AccountUpdateWithTx)`

tycho-indexer/src/extractor/evm/mod.rs
SmartMonkey-git: Please, be aware that this is not a doc sting, but a simple comment. It does not have the aspiration to explain this method's functionality. I added it to clairify why I cast into account via a reference and not as usual by a full object.

> why mention the BlockContractChanges struct?

Because converting into account by reference saves us from cloning the whole BlockContractChanges object.

> And what do you mean by "in forward"?

Forward is a function in ambient.rs

tycho-indexer/src/extractor/evm/mod.rs
SmartMonkey-git: Because of the sort

tycho-indexer/src/extractor/evm/mod.rs
SmartMonkey-git: I'm also unsure. Do you think we can trust the order?

tycho-indexer/src/extractor/evm/mod.rs
SmartMonkey-git: Cloning the slots was not necessary, but cloning the code is, as its used twice.

dianacarvalho1: slut cloning is very frowned upon. you shouldn't do it

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -190,6 +190,54 @@ impl AccountUpdate {
         Self { address, chain, slots, balance, code, change }
     }
 
+    // Converting AccountUpdate into Account with a references saves us from cloning the whole
+    // struct of BlockContractChanges in the forward function in ambient.rs.
```
dianacarvalho1:
```suggestion
    // Converting AccountUpdate into Account with references saves us from cloning the whole
    // struct of BlockContractChanges in the forward function in ambient.rs.
```

tycho-indexer/src/extractor/evm/ambient.rs
SmartMonkey-git: I dont know for sure why it removed that change, but basically the structure of `BlockContractChanges.changes `changed to contain the TransactionVMUpdates and this removed an additional AccountWithTx that is now part of TransactionVMUpdates.

SmartMonkey-git: It is hard to explain, but other changes made this redundant.

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -335,35 +379,66 @@ impl AccountUpdateWithTx {
...
+            .clone()
+            .into_iter()
+        {
+            match self
+                .component_balances
+                .entry(component_id)
+            {
+                Entry::Occupied(mut e) => {
+                    e.get_mut().extend(balances);
+                }
```
dianacarvalho1: why are we extending? shouldn't we replace?

dianacarvalho1: or are you accounting for more than one token? 🤔 I mean these balances look like {pool_id: {token1: balance, token2: balance}} right? and we just want to keep the last balance of the block for each component and token

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -420,21 +420,23 @@ impl TransactionVMUpdates {
...
+                .get_mut(&component_id)
             {
-                Entry::Occupied(mut e) => {
-                    e.get_mut().extend(balances);
-                }
-                Entry::Vacant(e) => {
-                    e.insert(balances);
+                // Iterate through the inner map and update values
+                for (inner_key, value) in balance_by_token_map {
+                    existing_inner_map.insert(inner_key, value);
```
dianacarvalho1, line 443: this changing in place works fine? 👀

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -335,35 +379,68 @@ impl AccountUpdateWithTx {
     /// # Errors
     /// This method will return `ExtractionError::MergeError` if any of the above
     /// conditions is violated.
-    pub fn merge(&mut self, other: AccountUpdateWithTx) -> Result<(), ExtractionError> {
+    pub fn merge(&mut self, other: &TransactionVMUpdates) -> Result<(), ExtractionError> {
         if self.tx.block_hash != other.tx.block_hash {
             return Err(ExtractionError::MergeError(format!(
-                "Can't merge AccountUpdates from different blocks: 0x{:x} != 0x{:x}",
+                "Can't merge TransactionUpdates from different blocks: 0x{:x} != 0x{:x}",
```
tamaralipows: I guess we should update this log to `TransactionVMUpdates`? It looks like there was a rename?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -335,35 +379,68 @@ impl AccountUpdateWithTx {
...
+            }
+        }
+
+        // Add new protocol components
+        self.protocol_components
+            .extend(other.protocol_components.clone());
 
-    fn deref(&self) -> &Self::Target {
-        &self.update
+        // Add new component balances and overwrite existing ones
```
tamaralipows, line 430: In what case do we overwrite the existing ones? If this happens multiple times a tx only right?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -190,6 +190,54 @@ impl AccountUpdate {
...
     }
 
+    // Converting AccountUpdate into Account with references saves us from cloning the whole
+    // struct of BlockContractChanges in the forward function in ambient.rs.
+    pub fn ref_into_account(&self, tx: &Transaction) -> Account {
+        let code = &self.code.clone().unwrap_or_default();
+        Account::new(
+            self.chain,
+            self.address,
+            format!("{:#020x}", self.address),
```
tamaralipows, line 205: So the title will always just be the string formatted address basically? Why do we even have a title?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -190,6 +190,54 @@ impl AccountUpdate {
...
+            self.slots.clone(),
+            self.balance.unwrap_or_default(),
+            code.clone(),
+            H256::from(keccak256(code)),
+            tx.hash,
+            tx.hash,
+            Some(tx.hash),
```
tamaralipows, line 215: Does this imply that we will only ever be passing an account creation tx here?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -680,31 +725,26 @@ impl BlockContractChanges {
...
-            }
-        }
+        let mut tx_update = sorted_tx_updates
+            .first()
+            .cloned()
+            .ok_or(ExtractionError::Empty)?;
 
+        for update in sorted_tx_updates.into_iter().skip(1) {
+            tx_update.merge(&update.clone())?;
+        }
```
tamaralipows, line 752: What does this skip do?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -249,9 +297,9 @@ pub struct BlockAccountChanges {
     pub block: Block,
     pub revert: bool,
     pub account_updates: HashMap<H160, AccountUpdate>,
-    pub new_protocol_components: Vec<ProtocolComponent>,
-    pub deleted_protocol_components: Vec<ProtocolComponent>,
-    pub component_balances: Vec<ComponentBalance>,
+    pub new_protocol_components: HashMap<ComponentId, ProtocolComponent>,
+    pub deleted_protocol_components: HashMap<ComponentId, ProtocolComponent>,
+    pub component_balances: HashMap<ComponentId, HashMap<H160, ComponentBalance>>,
```
tamaralipows, line 310: Is it safe to use ethereum specific types like H160 here? It's a bit weird that it's `ComponentId` but not something like `TokenAddress` for example (but please correct me because idk).

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -2243,4 +2251,50 @@ mod test {
...
+    #[rstest]
+    fn test_merge() {
+        let block_changes = block_state_changes();
+        let mut tx_update = block_changes.tx_updates[0].clone();
+
+        tx_update
+            .merge(block_changes.tx_updates.last().unwrap())
+            .unwrap();
+
+        println!("{:?}", tx_update.account_updates);
```
dianacarvalho1: print?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -2243,4 +2251,50 @@ mod test {
...
+
+        assert_eq!(
+            tx_update.component_balances,
```
dianacarvalho1: are there even any balances in the first tx_update?

tycho-indexer/src/extractor/evm/mod.rs
dianacarvalho1: when we merge tx changes (for the same block) we will keep only the latest balances per component

tycho-indexer/src/extractor/evm/mod.rs
dianacarvalho1: these are EVM specific methods, so yeah I think it's fine to use H160

tycho-indexer/src/extractor/evm/mod.rs
SmartMonkey-git: Oh, nice finding. The up.sql file actually states: `-- Succinct title of this account e.g. "maker psm"`.
So, I guess, it is suppose to be a combination of the ProtocolType and the ProtocolSystem, don't know if they are available at that point.

tycho-indexer/src/extractor/evm/mod.rs
SmartMonkey-git: Its the equivalent to `for a in b[1:]` in python

tycho-indexer/src/extractor/evm/mod.rs
SmartMonkey-git: No, there is not. Adding them to `block_state_changes()` fixture would require me to implement extracting balance changes in `BlockContractChanges.try_from_message()`, which is beyond the scope of this task. Further, other tests depend of the result of `BlockContractChanges.try_from_message()` for assertions which would make them fail.

I made the merge test independen by creating its own test values.

tycho-indexer/src/extractor/evm/mod.rs
SmartMonkey-git: Another nice finding. Yes, I accidentally removed the warning that was here. I added it back [Add warning to ref_into_account](https://github.com/propeller-heads/tycho-indexer/pull/128/commits/f0315d93054b625a14b9998db7b8bef47e9c11a7)

PR #129, opened 2024-02-01
ENG-2047: Add protocol components endpoint

dianacarvalho1:
Similar to the token endpoint.

Notes: 
- For now I didn't add the tvl filter - just a placeholder to be used later on
- Neither endpoint has pagination


tycho-indexer/src/services/rpc.rs
```diff
@@ -583,4 +691,94 @@ mod tests {
 
         assert_eq!(tokens.tokens.len(), 3);
     }
+
+    pub async fn setup_components(conn: &mut AsyncPgConnection) {
```
dianacarvalho1, line 695: I would have liked to make a nicer test fixtures for these DB inserts, but time is scarce so I added a task for it [here](https://datarevenue.atlassian.net/browse/ENG-2756) for cooldown

tycho-indexer/src/services/rpc.rs
```diff
@@ -80,6 +80,35 @@ impl From<evm::ERC20Token> for ResponseToken {
...
+                    let bytes = h.as_bytes().to_vec();
+                    Bytes::from(bytes)
+                })
+                .collect(),
+            contract_ids: protocol_component
+                .contract_ids
+                .into_iter()
+                .map(|h| {
+                    let bytes = h.as_bytes().to_vec();
+                    Bytes::from(bytes)
```
tamaralipows, line 103: I think we can do a one-liner as `Bytes::copy_from_slice(h.as_bytes())` ?

tycho-indexer/src/services/rpc.rs
```diff
@@ -228,6 +257,52 @@ impl RpcHandler {
...
+            .await
+        {
+            Ok(components) => Ok(dto::ProtocolComponentRequestResponse::new(
+                components
+                    .into_iter()
+                    .map(dto::ResponseProtocolComponent::from)
+                    .collect(),
+            )),
+            Err(err) => {
+                error!(error = %err, "Error while getting tokens.");
```
tamaralipows: hehe copypasta

tycho-indexer/src/services/rpc.rs
```diff
@@ -80,6 +80,35 @@ impl From<evm::ERC20Token> for ResponseToken {
...
+            id: protocol_component.id,
+            protocol_system: protocol_component.protocol_system,
+            protocol_type_name: protocol_component.protocol_type_name,
+            tokens: protocol_component
+                .tokens
+                .into_iter()
+                .map(|h| {
+                    let bytes = h.as_bytes().to_vec();
+                    Bytes::from(bytes)
+                })
```
SmartMonkey-git:
```suggestion
                .map(|token| {
                    let bytes = token.as_bytes().to_vec();
                    Bytes::from(bytes)
                })
```

tycho-indexer/src/services/rpc.rs
SmartMonkey-git:
```suggestion
                error!(error = %err, "Error while getting protocol components.");
```

tycho-types/src/dto.rs
```diff
@@ -406,19 +406,56 @@ pub struct ResponseToken {
...
-        decimals: u32,
-        tax: u64,
-        gas: Vec<Option<u64>>,
-    ) -> Self {
-        Self { chain, address, symbol, decimals, tax, gas }
+#[derive(Serialize, Deserialize, Debug, Default, PartialEq, ToSchema)]
+pub struct ProtocolComponentsRequestBody {
+    pub protocol_system: Option<String>,
+    #[serde(rename = "componentAddresses")]
+    pub component_ids: Option<Vec<String>>,
```
kayibal, line 493: Could you name it addresses here as well, please?

tycho-types/src/dto.rs
```diff
@@ -406,19 +406,56 @@ pub struct ResponseToken {
...
+
+impl ProtocolComponentsRequestBody {
+    pub fn new(protocol_system: Option<String>, component_ids: Option<Vec<String>>) -> Self {
+        Self { protocol_system, component_ids }
     }
 }
 
+/// Response from Tycho server for a protocol components request.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, ToSchema)]
+pub struct ProtocolComponentRequestResponse {
```
kayibal, line 510: Do we really call these structs `RequestResponse`? Maybe a better naming would be `RPCResponse`. But alright I don't mind it too much.

tycho-types/src/dto.rs
dianacarvalho1: they won't necessarily be addresses though. for ambient they might be like "ambient_USDC_ETH"

tycho-types/src/dto.rs
dianacarvalho1: yeah :/ I find it mind numbing tbh. I kept confusing them. I can rename them all though 👀

PR #130, opened 2024-02-01
ENG-2747: Get and save balance deltas in CachedGateway

tamaralipows:
[Issue](https://datarevenue.atlassian.net/browse/ENG-2747)

Read full commit messages for explanations.


tycho-indexer/src/storage/mod.rs
```diff
@@ -785,15 +785,15 @@ pub trait ProtocolGateway {
     /// - `target_version` The version at which to stop looking for changes.
     ///
     /// # Return
-    /// A hashmap containing a map of (protocol_component_id, token_id) to its respective balance
-    /// change, Err if no changes were found.
+    /// A hashmap containing a map of (protocol_external_id, token_address) to its respective
+    /// balance change, Err if no changes were found.
```
tamaralipows: Will fix this docstring in the morning sorry

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -1543,20 +1557,26 @@ mod test {
     async fn test_get_balance_deltas() {
         let mut conn = setup_db().await;
         setup_data(&mut conn).await;
-
+        let protocol_external_id = String::from("state1");
```
dianacarvalho1, line 1568: why statteeeeeeeeee? 😩

tamaralipows: Once again this is from the setup, I am not inserting anything new here (and I'd like to avoid doing so)

PR #131, opened 2024-02-01
ENG-2741: Add get_new_tokens to AmbientPgGateway

dianacarvalho1:
[Issue](https://datarevenue.atlassian.net/jira/software/projects/ENG/boards/24?selectedIssue=ENG-2741)

### DIFF
--- a/tycho-indexer/src/extractor/evm/ambient.rs
+++ b/tycho-indexer/src/extractor/evm/ambient.rs
@@ -114,15 +114,13 @@ impl AmbientPgGateway {
             .expect("pool should be connected");
 
         let mut tokens_set = HashSet::new();
+        let mut addresses = HashSet::new();
         for component in protocol_components {
             for token in &component.tokens {
                 tokens_set.insert(*token);
+                let byte_slice = token.as_bytes();
+                addresses.insert(Bytes::from(byte_slice.to_vec()));
             }
-        }
-        let mut addresses = HashSet::new();
-        for token in &tokens_set {
-            let byte_slice = token.as_bytes();
-            addresses.insert(Bytes::from(byte_slice.to_vec()));
         }
 
         let address_refs: Vec<&Bytes> = addresses.iter().collect();


tycho-indexer/src/extractor/evm/ambient.rs
```diff
@@ -99,13 +103,52 @@ impl AmbientPgGateway {
         Ok(())
     }
 
+    async fn get_new_tokens(&self, protocol_components: Vec<evm::ProtocolComponent>) -> Vec<H160> {
+        let mut conn = self
```
dianacarvalho1, line 110: I'm not sure about this 😬

tycho-indexer/src/extractor/evm/ambient.rs
```diff
@@ -99,13 +103,52 @@ impl AmbientPgGateway {
...
+        }
+
+        let address_refs: Vec<&Bytes> = addresses.iter().collect();
+        let addresses_option = Some(address_refs.as_slice());
+
+        let db_tokens = self
+            .state_gateway
+            .get_tokens(self.chain, addresses_option, &mut conn)
+            .await
+            .unwrap(); // TODO be better
```
tamaralipows: 😂

tycho-indexer/src/extractor/evm/ambient.rs
tamaralipows: This looks correct to me? Though tbh I'm not sure about the db connection.. I will need to look deeper tomorrow

dianacarvalho1: yes I think this is fine in the end

tycho-indexer/src/extractor/evm/ambient.rs
```diff
@@ -99,13 +103,57 @@ impl AmbientPgGateway {
...
+        for component in protocol_components {
+            for token in &component.tokens {
+                tokens_set.insert(*token);
+            }
+        }
+        let mut addresses = Vec::new();
+        for token in &tokens_set {
+            let byte_slice = token.as_bytes();
+            addresses.push(Bytes::from(byte_slice.to_vec()));
+        }
```
SmartMonkey-git:
```suggestion
        let mut tokens_set = HashSet::new();
        let mut addresses = Vec::new();
        for component in protocol_components {
            for token in &component.tokens {
                if tokens_set.contains(token) {
                    continue;
                }
                tokens_set.insert(*token);
                let byte_slice = token.as_bytes();
                addresses.push(Bytes::from(byte_slice.to_vec()));
            }
        }
```

tycho-indexer/src/extractor/evm/ambient.rs
```diff
@@ -99,13 +103,57 @@ impl AmbientPgGateway {
...
+
+        let address_refs: Vec<&Bytes> = addresses.iter().collect();
+        let addresses_option = Some(address_refs.as_slice());
+
+        let db_tokens = self
```
SmartMonkey-git, line 129: If the ProtocolComponent has not tokens at all, do you still query and if we do so, should we not just return early?

tycho-indexer/src/extractor/evm/ambient.rs
dianacarvalho1: do you really think this is better? to kinda reimplement the set logic instead of looping (not even) twice?

tycho-indexer/src/extractor/evm/ambient.rs
dianacarvalho1: would this ever even happen? to have no tokens?

SmartMonkey-git: Well, that is a question worth asking.
I did not consider that this is ambient specific. If ambient only has components that have always tokens, then this should be ok. I only know of ambient pools, I dont know if there is a factory for example.

tycho-indexer/src/extractor/evm/ambient.rs
SmartMonkey-git: What if you make addresses a set as well?

dianacarvalho1: addresses should already only have unique values. does it make sense to make it a set as well? well it shouldn't hurt I guess 🤷🏼‍♀️

tycho-indexer/src/extractor/evm/ambient.rs
dianacarvalho1: rn this would never happen, so I think it's fine to leave it like this tbh. I feel uncomfortable in adding early returns that may mask other bugs

tycho-indexer/src/extractor/evm/ambient.rs
SmartMonkey-git: Yes, the addresses should already be unique, but the point was to eliminate the second loop without `kinda reimplement the set logic`, since addresses is also a set now you can just insert the values in the first loop and not reimplement the set logic, no?

dianacarvalho1: aaaahh, sorry. I misunderstood. ok ok

PR #132, opened 2024-02-02
ENG 2758: Support native protocol deltas on WsDeltasClient

louise-poole:
TODOs: update block structs when https://github.com/propeller-heads/tycho-indexer/pull/128 and https://github.com/propeller-heads/tycho-indexer/pull/111 are merged.

### DIFF
--- a/tycho-indexer/src/services/rpc.rs
+++ b/tycho-indexer/src/services/rpc.rs
@@ -392,8 +392,8 @@ mod tests {
             .version
             .timestamp
             .unwrap()
-            .timestamp_millis()
-            - result
+            .timestamp_millis() -
+            result
                 .version
                 .timestamp
                 .unwrap()


tycho-types/src/dto.rs
```diff
@@ -72,11 +74,18 @@ pub enum Response {
     SubscriptionEnded { subscription_id: Uuid },
 }
 
+#[derive(Deserialize, Serialize, Debug)]
+#[serde(untagged)]
+pub enum Deltas {
+    VM(BlockAccountChanges),
+    Native(BlockEntityChangesResult),
```
kayibal, line 81: Why the `Result` naming? Would be nicer if it was not here no?

louise-poole: So on for the VM structs everyone was getting very confused between the `BlockContractChanges` proto message, `BlockContractChanges` struct for the decoded message and the `BlockAccountChanges` struct for the aggregated result of `BlockContractChanges`. To us `BlockContractChanges` and `BlockAccountChanges` sound like they should be similar structs representing block changes of two different things, but one is actually just the aggregated result of the other. We tried to make that clearer on the native protocol side by naming the proto message and it's decoded struct both `BlockEntityChanges` and the aggregated result of that struct as `BlockEntityChangesResult`.

PR #133, opened 2024-02-02
ENG-2045: Fix get_balance_deltas query

tamaralipows:
Please see commit messages.

For more context you can also see [this thread](https://propellerswap.slack.com/archives/C05LQALE9L1/p1706881958012079)

The last 4 commits refer to [this task](https://datarevenue.atlassian.net/jira/software/projects/ENG/boards/24?selectedIssue=ENG-2765).

### DIFF
--- a/tycho-indexer/src/extractor/evm/ambient.rs
+++ b/tycho-indexer/src/extractor/evm/ambient.rs
@@ -873,7 +873,7 @@ mod test_serial_db {
                             base_token,
                             ComponentBalance {
                                 token: base_token,
-                                new_balance: Bytes::from(&[0u8]),
+                                balance: Bytes::from(&[0u8]),
                                 balance_float: 10.0,
                                 modify_tx: TX_HASH_0.parse().unwrap(),
                                 component_id: component_id.clone(),
@@ -903,7 +903,7 @@ mod test_serial_db {
                             base_token,
                             ComponentBalance {
                                 token: base_token,
-                                new_balance: Bytes::from(&[0u8]),
+                                balance: Bytes::from(&[0u8]),
                                 balance_float: 10.0,
                                 modify_tx: TX_HASH_1.parse().unwrap(),
                                 component_id: component_id.clone(),
--- a/tycho-indexer/src/extractor/evm/mod.rs
+++ b/tycho-indexer/src/extractor/evm/mod.rs
@@ -2301,7 +2301,7 @@ mod test {
                     H160::from_low_u64_be(0x0000000000000000000000000000000061626364),
                     ComponentBalance {
                         token: H160::from_low_u64_be(0x0000000000000000000000000000000066666666),
-                        new_balance: Bytes::from(0_i32.to_le_bytes()),
+                        balance: Bytes::from(0_i32.to_le_bytes()),
                         modify_tx: Default::default(),
                         component_id: protocol_component_first_tx.id.clone(),
                         balance_float: 0.0,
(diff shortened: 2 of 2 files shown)


tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -825,10 +825,27 @@ where
                 .component_id
                 .to_string()];
 
+            let previous_component_balance: Option<Balance> = schema::component_balance::table
```
kayibal: We shouldn't do queries in a loop! You can use the `apply_delta_versioning` function to set the previous balances. If some component's balance here is updated multiple times.

Buuut I just noticed that it doesn't update the `previous_value` of the latest version currently in the database . .. I'll have to fix that.

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -868,25 +885,20 @@ where
             // -----------------|--------------------------|
             //                start                     target
             // We query for balance updates between start and target version.
-            let changed_component_balances = component_balance
+            component_balance
                 .inner_join(schema::protocol_component::table.inner_join(schema::chain::table))
```
kayibal: Why do we need this join? You already have the chain id as a local var, might just use that to filter directly on the transaction table...

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -908,7 +919,7 @@ where
                 .map(|(external_id, address, balance, tx)| ComponentBalance {
                     component_id: external_id,
                     token: address.into(),
-                    new_balance: balance,
+                    balance,
                     modify_tx: tx.into(),
```
kayibal, line 907: Why do you emit this? I think we might not need the tx for deltas.

tycho-indexer/migrations/2024-01-10_tycho_patch/up.sql
```diff
@@ -80,6 +80,8 @@ CREATE TABLE IF NOT EXISTS component_balance(
     "token_id" bigint REFERENCES "token"(id) NOT NULL,
     -- new balance of the token for this component
     "new_balance" bytea NOT NULL,
+    -- new balance of the token for this component
+    "previous_balance" bytea NOT NULL,
```
SmartMonkey-git: Does it make sense for this to be `NOT NULL`. What if there is no previous balances and this balance is in fact the first balance ever recorded?

SmartMonkey-git: Actually, `StorableComponentBalance.to_storage` takes `previous_balance` as Option.

SmartMonkey-git: Ah, I see you just insert `0`, but isn't not having any liquidity different from not knowing the liquidity?

tamaralipows: At least with Ambient, I don't think there will be a case where we don't know the liquidity, since we are tracking the pool balances since creation. I'm not sure how this will be in other cases (i.e. when we start tracking the pool balances)

tycho-indexer/src/storage/postgres/protocol.rs
tamaralipows: I chose to emit this simply to not have to create a new object without this. Does it hurt to keep the modify_tx?

PR #135, opened 2024-02-03
ENG-2374: implement uniswap V3 substream module

zizou0x:
# Description

This PR introduces a submodule for extracting information about Uniswap-V3 pools, including their state and Total Value Locked (TVL).

# Tech details

## For new pools:

Extracting new created pool is done with the `PoolCreated` event. Whenever this event is emitted by the pool factory address, the corresponding pool is recorded in a storage module.

## For pools data:

Unfortunately, UniswapV3 events don't return all the data we need. To be able to extract everything we need to look storage changes within transactions and identify any alterations relevant to the data we require.

## For TVL:

TVL tracking is managed by a storage module. We employ a map module that monitors every event emitted by a Uniswap pool and accordingly adjusts the token balance within the storage module. _Note on this: This approach may lead to a balance mismatch if tokens are transferred into the pool through unconventional means (i.e., in a manner not aligned with the pool's standard operations). But in my opinion this is very fine because we need to track the actual TVL and not the balances._

### DIFF
--- a/substreams/ethereum-uniswap-v3/src/modules/1_map_pool_created.rs
+++ b/substreams/ethereum-uniswap-v3/src/modules/1_map_pool_created.rs
@@ -13,22 +13,25 @@ use crate::{
     },
 };
 
-// TODO: Parametrize Factory Address
-const UNISWAP_V3_FACTORY_ADDRESS: &str = "0x1F98431c8aD98523631AE4a59f267346ea31F984";
-
 #[substreams::handlers::map]
 pub fn map_pools_created(
+    params: String,
     block: eth::Block,
 ) -> Result<BlockEntityChanges, substreams::errors::Error> {
     let mut new_pools: Vec<TransactionEntityChanges> = vec![];
+    let factory_address = params.as_str();
 
-    get_new_pools(&block, &mut new_pools);
+    get_new_pools(&block, &mut new_pools, factory_address);
 
     Ok(BlockEntityChanges { block: Some(block.into()), changes: new_pools })
 }
 
 // Extract new pools from PoolCreated events
-fn get_new_pools(block: &eth::Block, new_pools: &mut Vec<TransactionEntityChanges>) {
+fn get_new_pools(
+    block: &eth::Block,
+    new_pools: &mut Vec<TransactionEntityChanges>,
+    factory_address: &str,
+) {
     // Extract new pools from PoolCreated events
     let mut on_pair_created = |event: PoolCreated, _tx: &eth::TransactionTrace, _log: &eth::Log| {
         let tycho_tx: Transaction = _tx.into();
@@ -66,7 +69,7 @@ fn get_new_pools(block: &eth::Block, new_pools: &mut Vec<TransactionEntityChange
 
     let mut eh = EventHandler::new(block);
 
-    eh.filter_by_address(vec![Address::from_str(UNISWAP_V3_FACTORY_ADDRESS).unwrap()]);
+    eh.filter_by_address(vec![Address::from_str(factory_address).unwrap()]);
 
     eh.on::<PoolCreated, _>(&mut on_pair_created);
     eh.handle_events();
--- a/substreams/ethereum-uniswap-v3/src/lib.rs
+++ b/substreams/ethereum-uniswap-v3/src/lib.rs
@@ -1,3 +1,5 @@
+#![allow(clippy::not_unsafe_ptr_arg_deref)]
+
 mod abi;
 mod modules;
 mod pb;
--- a/substreams/ethereum-uniswap-v3/substreams.yaml
+++ b/substreams/ethereum-uniswap-v3/substreams.yaml
@@ -22,6 +22,7 @@ modules:
     kind: map
     initialBlock: 12369621
     inputs:
+      - params: string
       - source: sf.ethereum.type.v2.Block
     output:
       type: proto:tycho.evm.v1.BlockEntityChanges
@@ -61,3 +62,6 @@ modules:
       - store: store_pools_balances
     output:
       type: proto:tycho.evm.v1.BlockEntityChanges
+
+params:
+  map_pools_created: "1F98431c8aD98523631AE4a59f267346ea31F984"


substreams/ethereum-uniswap-v3/src/storage/utils.rs
```diff
@@ -0,0 +1,178 @@
+use substreams::scalar::BigInt;
```
zizou0x, line 1: This was taken from the subtreams implementation

substreams/ethereum-uniswap-v3/src/events/burn.rs
```diff
@@ -0,0 +1,45 @@
...
+        let filtered_storage_changes = storage_vec
+            .filter_by_address(pool_address)
+            .into_iter()
+            .cloned()
+            .collect();
+
+        let pool_storage = UniswapPoolStorage::new(&filtered_storage_changes);
+
+        let mut changed_attributes =
+            pool_storage.get_changed_attributes(TRACKED_SLOTS.to_vec().iter().collect());
```
zizou0x, line 29: As an improvement, we could only track the slots that could change when this event happens. But for simplicity I've decided to always track every slots we need.

substreams/ethereum-uniswap-v3/src/events/burn.rs
```diff
@@ -0,0 +1,45 @@
...
+        let changed_ticks =
+            pool_storage.get_ticks_changes(vec![&self.tick_upper, &self.tick_lower]);
+
+        changed_attributes.extend(changed_ticks);
+
+        changed_attributes
+    }
+
+    fn get_balance_delta(&self, _pool: &Pool, _ordinal: u64) -> Vec<BalanceDelta> {
+        // Burn event balances deltas are accounted for by the Collect event.
```
zizou0x, line 40: This was very surprising but it works if we track only `Collect` events

substreams/ethereum-uniswap-v3/src/modules/2_store_pools.rs
```diff
@@ -0,0 +1,22 @@
+use std::str;
+
+use substreams::store::{StoreNew, StoreSetIfNotExists, StoreSetIfNotExistsProto};
+
+use crate::pb::tycho::evm::{uniswap::v3::Pool, v1::BlockEntityChanges};
```
kayibal: Why do we need this Pool model? Could you not store ProtocolComponent model instead which is already included with tycho? If not, why do we need it to by under tycho package - seems like it is v3 specific.

substreams/ethereum-uniswap-v3/src/modules/3_map_balance_changes.rs
```diff
@@ -0,0 +1,40 @@
...
+
+#[substreams::handlers::map]
+pub fn map_balance_changes(
+    block: eth::Block,
+    pools_store: StoreGetProto<Pool>,
+) -> Result<BalanceDeltas, anyhow::Error> {
+    let mut balances_deltas = Vec::new();
+    for trx in block.transactions() {
+        let mut tx_deltas = Vec::new();
+        for (log, _) in trx.logs_with_calls() {
```
kayibal: If you don't need the calls maybe there is a more specific logs method?

substreams/ethereum-uniswap-v3/src/events/swap.rs
```diff
@@ -0,0 +1,50 @@
...
+        let filtered_storage_changes = storage_vec
+            .filter_by_address(pool_address)
+            .into_iter()
+            .cloned()
+            .collect();
+
+        let pool_storage = UniswapPoolStorage::new(&filtered_storage_changes);
+
+        pool_storage.get_changed_attributes(TRACKED_SLOTS.to_vec().iter().collect())
+    }
```
kayibal, line 31: Except for updating ticks this method seems to be the same and has basically nothing to do with the event itself (you don't use self). Wouldn't this be better as a function vs on a trait?

EDIT: ok in some cases you get the tick positions from the events and the balance one seems alright too. I was just worried about the amount of duplication here but I think it is alright.

substreams/ethereum-uniswap-v3/src/events/burn.rs
kayibal: Is it possible for someone to call burn without collect?

substreams/ethereum-uniswap-v3/src/events/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+        Swap::match_and_decode(event).map(EventType::Swap),
+        Flash::match_and_decode(event).map(EventType::Flash),
+        Mint::match_and_decode(event).map(EventType::Mint),
+        Burn::match_and_decode(event).map(EventType::Burn),
+        Collect::match_and_decode(event).map(EventType::Collect),
+        SetFeeProtocol::match_and_decode(event).map(EventType::SetFeeProtocol),
+        CollectProtocol::match_and_decode(event).map(EventType::CollectProtocol),
+    ]
+    .into_iter()
+    .find_map(std::convert::identity)
```
kayibal, line 106: Why not use a match on the event hash?

substreams/ethereum-uniswap-v3/src/modules/1_map_pool_created.rs
```diff
@@ -0,0 +1,73 @@
...
+
+use crate::{
+    abi::factory::events::PoolCreated,
+    pb::tycho::evm::v1::{
+        Attribute, BlockEntityChanges, ChangeType, FinancialType, ImplementationType,
+        ProtocolComponent, ProtocolType, Transaction, TransactionEntityChanges,
+    },
+};
+
+// TODO: Parametrize Factory Address
```
kayibal: Forgot about this?

substreams/ethereum-uniswap-v3/src/modules/4_store_pools_balances.rs
```diff
@@ -0,0 +1,27 @@
...
+    store::{StoreAdd, StoreAddBigInt, StoreNew},
+};
+
+#[substreams::handlers::store]
+pub fn store_pools_balances(balances_deltas: BalanceDeltas, store: StoreAddBigInt) {
+    balances_deltas
+        .deltas
+        .iter()
+        .for_each(|delta| {
+            store.add(
```
kayibal, line 14: Is there a benefit of using `store.add` vs aggregating the deltas per transaction in the previous map step?

substreams/ethereum-uniswap-v3/src/modules/5_map_pool_events.rs
```diff
@@ -0,0 +1,219 @@
...
+                    pool.address
+                        .clone()
+                        .as_slice()
+                        .try_into()
+                        .expect("Pool address is not 20 bytes long"),
+                );
+
+                let mut balance_changes: Vec<BalanceChange> = vec![];
+
+                if !(get_log_changed_balances(log, &pool).is_empty()) {
```
kayibal, line 50: Why do you need to call this again? Can't you use the store to get the balance changes?
Afaik the store should be able to provide you with the keys that changed in the last block.

substreams/ethereum-uniswap-v3/src/modules/5_map_pool_events.rs
```diff
@@ -0,0 +1,219 @@
...
+        }
+    }
+
+    // Make a list of all HashMap values:
+    let tx_entity_changes: Vec<TransactionEntityChanges> = tx_changes_map.into_values().collect();
+
+    let tycho_block: Block = block.into();
+
+    let block_entity_changes =
+        BlockEntityChanges { block: Option::from(tycho_block), changes: tx_entity_changes };
```
kayibal: `Some(tycho_block)`, why is everyone using `Option::from` these days? 

My guess is CoPilot, perfect example why I am becoming more and more convinced that using it will make you a worse coder..

substreams/ethereum-uniswap-v3/src/modules/5_map_pool_events.rs
```diff
@@ -0,0 +1,219 @@
...
+
+    Ok(block_entity_changes)
+}
+
+fn update_tx_changes_map(
+    entity_changes: Vec<EntityChanges>,
+    balance_changes: Vec<BalanceChange>,
+    tx_changes_map: &mut HashMap<Vec<u8>, TransactionEntityChanges>,
+    tx_trace: &TransactionTrace,
```
kayibal, line 112: You only get the hash from here, no? Maybe just pass in here the hash in that case.

substreams/ethereum-uniswap-v3/src/storage/utils.rs
```diff
@@ -0,0 +1,178 @@
...
+        )
+    }
+
+    #[test]
+    #[should_panic]
+    fn left_pad_gt_32_bytes() {
+        let input = vec![
+            7u8, 0u8, 0u8, 0u8, 0u8, 10u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 93u8, 0u8, 0u8, 0u8, 0u8,
+            0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 221u8, 98u8, 237u8, 62u8,
+        ];
```
kayibal: Very cryptic^^ I'd probably prefer using `hex!()` here instead...

substreams/ethereum-uniswap-v3/src/events/mod.rs
zizou0x: If I do this I would then need to cast it to the corresponding `abi::pool::events` I think? 
This `match_and_decode` does this for me basically, it compare the event topic0 and if it match, it cast it into the event struct

substreams/ethereum-uniswap-v3/src/events/swap.rs
zizou0x: Yes, that part i'm not a big big fan tbh, also `get_changed_attributes` is transaction and not event scoped here. It means that if in the same tx two events change the same attribute twice, we would get the latest state even when calling this on the first event.

But now that we know we will change to rely on events only and not storage changes, I would keep it like this as it makes it easier to change afterward

substreams/ethereum-uniswap-v3/src/modules/1_map_pool_created.rs
zizou0x: I need to look into how to do this, do you have any input maybe?

substreams/ethereum-uniswap-v3/src/modules/5_map_pool_events.rs
zizou0x: I think this was copy pasted this from our uniswapV2 module 🙈 I agree its dirty, I changed it

substreams/ethereum-uniswap-v3/src/modules/5_map_pool_events.rs
zizou0x: No I also use it to create the `TransactionEntityChanges` if the transaction isn't already in the map `tx: Some(tx_trace.into())`

substreams/ethereum-uniswap-v3/src/storage/utils.rs
zizou0x: True, changed it for the hex! macro

substreams/ethereum-uniswap-v3/src/modules/4_store_pools_balances.rs
zizou0x: Not really, it just simplifies our logic by using what's implemented by `StoreAddBigInt`

substreams/ethereum-uniswap-v3/src/modules/3_map_balance_changes.rs
zizou0x: I don't think there is one, I've used a `flat_map` to get only the logs here

substreams/ethereum-uniswap-v3/src/modules/2_store_pools.rs
zizou0x: I guess we could use `ProtocolComponent` but wouldn't we carry around a lot of unnecessary data with it?

Also Pool is defined in `ethereum-uniswap-v3/proto/v1/uniswap.proto` but the package name was weird. I've renamed it

substreams/ethereum-uniswap-v3/src/events/burn.rs
zizou0x: I think so, if someone doesn't use the frontend. But in that case it wouldn't transfer the tokens out, just remove the position. Also in all the cases I've checked, both came together

PR #138, opened 2024-02-06
ENG-2049: Save tokens, components and balances in Ambient gtw

dianacarvalho1:


tycho-indexer/src/extractor/evm/ambient.rs
```diff
@@ -874,6 +938,37 @@ mod test_serial_db {
...
+                    None,
+                    &BlockOrTimestamp::Block(BlockIdentifier::Number((
+                        Chain::Ethereum,
+                        msg.block.number as i64,
+                    ))),
+                    &mut conn,
+                )
+                .await
+                .unwrap();
+            println!("BALNACES {:?}", component_balances);
```
tamaralipows: leftover print

dianacarvalho1: upsi

PR #142, opened 2024-02-06
Add protocol state delta endpoint

zizou0x:
This PR adds a get protocol state delta endpoint. This endpoint is equivalent to get_contracts_delta but for protocol state. It's used to get the protocol state deltas from a list a component ids between two versions (could be block or timestamp).

### DIFF
--- a/tycho-client/src/deltas.rs
+++ b/tycho-client/src/deltas.rs
@@ -682,6 +682,7 @@ mod tests {
                             "chain": "ethereum",             
                             "ts": "2023-09-14T00:00:00"
                         },
+                        "revert": false,
                         "account_updates": {
                             "0x7a250d5630b4cf539739df2c5dacb4c659f2488d": {
                                 "address": "0x7a250d5630b4cf539739df2c5dacb4c659f2488d",
@@ -733,6 +734,7 @@ mod tests {
                             "chain": "ethereum",             
                             "ts": "2023-09-14T00:00:00"
                         },
+                        "revert": false,
                         "state_updates": {
                             "component_1": {
                                 "component_id": "component_1",
@@ -989,6 +991,7 @@ mod tests {
                             "chain": "ethereum",             
                             "ts": "2023-09-14T00:00:00"
                         },
+                        "revert": false,
                         "account_updates": {
                             "0x7a250d5630b4cf539739df2c5dacb4c659f2488d": {
                                 "address": "0x7a250d5630b4cf539739df2c5dacb4c659f2488d",
--- a/tycho-indexer/src/extractor/evm/ambient.rs
+++ b/tycho-indexer/src/extractor/evm/ambient.rs
@@ -222,6 +222,7 @@ impl AmbientPgGateway {
             &self.name,
             self.chain,
             block,
+            true,
             account_updates,
             // TODO: get protocol components from gateway (in ENG-2049)
             Vec::new(),
@@ -770,6 +771,7 @@ mod test_serial_db {
             extractor: "vm:ambient".to_owned(),
             chain: Chain::Ethereum,
             block: evm::Block::default(),
+            revert: false,
             tx_updates: vec![
                 evm::AccountUpdateWithTx::new(
                     H160(AMBIENT_CONTRACT),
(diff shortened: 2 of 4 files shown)


tycho-indexer/src/services/rpc.rs
```diff
@@ -467,6 +479,70 @@ impl RpcHandler {
             }
         }
     }
+
+    #[instrument(skip(self, chain, request, params))]
+    async fn get_protocol_delta(
+        &self,
+        chain: &Chain,
+        request: &dto::ProtocolDeltaRequestBody,
+        params: &dto::StateRequestParameters,
```
zizou0x: Do we need this?

louise-poole: I would say no

tycho-indexer/src/services/rpc.rs
```diff
@@ -467,6 +479,70 @@ impl RpcHandler {
...
+    }
+
+    async fn get_protocol_delta_inner(
+        &self,
+        chain: &Chain,
+        request: &dto::ProtocolDeltaRequestBody,
+        params: &dto::StateRequestParameters,
+        db_connection: &mut AsyncPgConnection,
+    ) -> Result<dto::ProtocolDeltaRequestResponse, RpcError> {
+        #![allow(unused_variables)]
```
louise-poole: Do we need this here?

tycho-indexer/src/services/rpc.rs
```diff
@@ -467,6 +479,70 @@ impl RpcHandler {
...
+        //TODO: handle when no contract is specified with filters
```
louise-poole: Do you mean component?

tycho-indexer/src/services/rpc.rs
```diff
@@ -467,6 +479,70 @@ impl RpcHandler {
...
+        // Get the components IDs from the request
+        let ids = request.component_ids.clone();
+        let component_ids: Option<HashSet<String>> = ids.map(|ids| {
+            ids.into_iter()
+                .collect::<HashSet<String>>()
+        });
+        debug!(?component_ids, "Getting protocol states delta.");
+        let component_ids: Option<&HashSet<String>> = component_ids.as_ref();
+
+        // Get the contract deltas from the database
```
```suggestion
        // Get the protocol state deltas from the database
```

tycho-indexer/src/services/rpc.rs
```diff
@@ -467,6 +479,70 @@ impl RpcHandler {
...
+        match self
+            .db_gateway
+            .get_protocol_states_delta(chain, Some(&start), &end, db_connection)
+            .await
+        {
+            Ok(mut components) => {
+                // Filter by contract addresses if specified in the request
```
```suggestion
                // Filter by component id if specified in the request
```

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -1267,7 +1265,7 @@ mod test {
         let (account_id_weth, weth_id) =
             db_fixtures::insert_token(conn, chain_id, WETH.trim_start_matches("0x"), "WETH", 18)
                 .await;
-        let (account_id_usdc, usdc_id) =
+        let (_account_id_usdc, _usdc_id) =
```
louise-poole: We can remove the let part here

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -1290,7 +1288,7 @@ mod test {
             Some(vec![contract_code_id]),
         )
         .await;
-        let protocol_component_id2 = db_fixtures::insert_protocol_component(
+        let _protocol_component_id2 = db_fixtures::insert_protocol_component(
```
louise-poole: Same here

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -2080,7 +2071,7 @@ mod test {
         let protocol_type_id_1 =
             db_fixtures::insert_protocol_type(&mut conn, &protocol_type_name_1, None, None, None)
                 .await;
-        let protocol_type_id_2 =
+        let _protocol_type_id_2 =
```
louise-poole: And here

tycho-indexer/src/services/rpc.rs
zizou0x: We don't if we remove `params`, so I removed it

PR #143, opened 2024-02-06
Client Feed

kayibal:
Ok this is reaching the end stadium, I recommend reviewing files in this order:

- feed/block_history.rs
- feed/component_tracker.rs
- feed/state_synchronizer.rs
- feed/mod.rs

For context: The goal is to take multiple extractor streams of the same chain. And emit state update messages (including snapshot + update) logic that are merged on the block.

To do this we have to synchronize messages across two dimensions: State and Blocks. By synchronizing state I mean ensuring that we retrieve a snapshot in case the client has not seen a component before, then apply any updates on top of that.

As for block synchronization I mean that we try to aggregate different streams that are live by their block number. This allows client to spend more time doing work as they get all updates in one transactional message compared to many messages arriving at slightly different times. 
This logic is now part of the client as whether we want to use that or no depends on the use case.

### DIFF
--- a/tycho-client/src/feed/mod.rs
+++ b/tycho-client/src/feed/mod.rs
@@ -45,51 +45,54 @@ impl Header {
 
 type BlockSyncResult<T> = anyhow::Result<T>;
 
-/// Synchronizes multiple StateSynchronizers on the time dimension.
+/// Aligns multiple StateSynchronizers on the block dimension.
+///
+/// ## Purpose
+///
+/// The purpose of this component is to handle streams from multiple state synchronizers and
+/// basically align/merge them according to their blocks. Ideally this should be done in a
+/// fault-tolerant way meaning we can recover from a state synchronizer suffering from timing
+/// issues. E.g. a delayed or unresponsive state synchronizer might recover again. An advanced state
+/// synchronizer can be included again once we reach the block it is at.
+///
+/// ## Limitations
+/// - Supports only chains with fixed blocks time for now due to the lock step mechanism
 ///
 /// ## Initialisation
-/// Queries all registered synchronizers for the first header. It expects all synchronizers to
-/// return the same header. If this is not the case the run method will error.
+/// Queries all registered synchronizers for their first message. It expects all synchronizers to
+/// return a message from the same header. If this is not the case the run method will error.
 ///
 /// ## Main loop
-/// Once started, the synchronizers are queried for new blocks in lock step: the main loop queries
-/// all ready synchronizers for the last emitted data, builds the `FeedMessage` and sends it through
-/// the channel, then it schedule the wait procedure for the next block.
+/// Once started, the synchronizers are queried concurrently for messages in lock step:
+/// the main loop queries all synchronizers in ready for the last emitted data, builds the
+/// `FeedMessage` and emits it, then it schedules the wait procedure for the
+/// next block.
 ///
-/// ### Synchronization Logic
-/// We know the next block number we expect - this helps us detect gaps (and reverts, but
-/// they don't require special handling)
-/// So the wait procedure consists in waiting for any of the receivers to emit a new block
-/// we check if that block corresponds with what we expected - if that is the case we deem
-/// the respective synchronizer as ready.
-/// At this point a timeout starts for the remaining channels, as soon as they reply, we
-/// again check against the expected block number and if they all reply with the expected
-/// next block within the timeout they are all considered as ready.
+/// ## Synchronization Logic
 ///
-/// #### Note
+/// To classify a synchronizer as delayed or advanced, we need to first define the current
+/// block. Currently, we simply expect all synchronizer to be at the same block at startup. The
+/// block they are at is then simply taken as the current block.
+///
+/// Once we heave the current block we can easily determine which block we expect next. And if a
+/// state synchronizer delivers a different block from that we can classify it as delayed or
+/// advanced.
+///
+/// If any synchronizer is in not ready state we will try to bring it back to the ready state. This
+/// is done by trying to empty any buffers of a delayed synchronizer or waiting to reach
+/// the height of an advanced synchronizer (and flagging it as such in the meantime).
+///
+/// Of course, we can't wait forever for a synchronizer to reply/recover. All of this must happen
+/// within the block production step of the blockchain:
+/// The wait procedure consists in waiting for any of the receivers to emit a new message (within a
+/// max timeout - several multiples of the block time). Once a message is received a very short
+/// timeout start for the remaining synchronizers, to deliver a message. Any synchronizer failing to
+/// do so is transitioned to delayed.
+///
+/// ### Note
 /// The described process above is the goal. It is currently not implemented like that. Instead we
 /// simply wait `block_time` + `wait_time`. Synchronizers are expected to respond within that
-/// timeout. This is simpler but only works on chains with fixed block times.
-///
-/// ### Synchronization State
-/// Now when either all receivers emitted a block or the timeout passed, we are ready for the
-/// next step. At this point each synchronizer can be in one of the following states:
-/// 1. Ready, it emitted the expected block, or a block of same height within the timeout
-/// 2. Advanced, it emitted a future block within the timeout: likely a reconnect happened and there
-///    is some significant back pressure on our channels. This basically means we are too slow to
-///    process messages.
-/// 3. Delayed, it did not emit anything within the timeout, or an older block than expected:
-///    Extractor syncing, connection issue, or firehose too slow.
-///
-/// We will only query synchronizer for state if they are in the ready state or if they are
-/// advanced in which case we simply take the advanced state. For others, we will inform
-/// clients about their current sync state but not actually emit old or advanced data.
-///
-/// Any extractors that have been stale or delayed for too long we will stop waiting for - so
-/// we will end their synchronizers thread and drop the handle.
-///
-/// Any advanced extractors, we will not wait for in the future, they will be assumed ready
-/// until we reach their height and from then on are handled as if they were ready.
+/// timeout. This is simpler but only works well on chains with fixed block times.
 pub struct BlockSynchronizer<S> {
     synchronizers: Option<HashMap<ExtractorIdentity, S>>,
     block_time: std::time::Duration,
@@ -110,31 +113,37 @@ pub enum SynchronizerState {
     Ended,
 }
 
-pub struct SynchronizerHandle<S> {
+pub struct SynchronizerStream {
     extractor_id: ExtractorIdentity,
     state: SynchronizerState,
-    sync: S,
     modify_ts: NaiveDateTime,
-    rx: Receiver<Header>,
+    rx: Receiver<StateSyncMessage>,
 }
 
-impl<S> SynchronizerHandle<S> {
-    async fn try_advance(&mut self, block_history: &BlockHistory, max_wait: std::time::Duration) {
+impl SynchronizerStream {
+    async fn try_advance(
+        &mut self,
+        block_history: &BlockHistory,
+        max_wait: std::time::Duration,
+    ) -> Option<StateSyncMessage> {
         let extractor_id = &self.extractor_id;
         let latest_block = block_history.latest();
         match &self.state {
             SynchronizerState::Started | SynchronizerState::Ended => {
                 warn!(state=?&self.state, "Advancing Synchronizer in this state not supported!");
+                None
             }
             SynchronizerState::Advanced(b) => {
                 let future_block = b.clone();
                 // Transition to ready once we arrived at the expected height
                 self.transition(future_block, block_history);
+                None
             }
             SynchronizerState::Ready(previous_block) => {
                 // Try to recv the next expected block, update state accordingly.
                 self.try_recv_next_expected(max_wait, block_history, previous_block.clone())
-                    .await;
+                    .await
+                // TODO: if we entered advanced state we need to buffer the message for a while.
             }
             SynchronizerState::Delayed(old_block) => {
                 // try to catch up all currently queued blocks until the expected block
@@ -145,7 +154,7 @@ impl<S> SynchronizerHandle<S> {
                     "Trying to catch up to latest block"
                 );
                 self.try_catch_up(block_history, max_wait)
-                    .await;
+                    .await
             }
             SynchronizerState::Stale(old_block) => {
                 debug!(
@@ -155,7 +164,7 @@ impl<S> SynchronizerHandle<S> {
                     "Trying to catch up to latest block"
                 );
                 self.try_catch_up(block_history, max_wait)
-                    .await;
+                    .await
             }
         }
     }
@@ -169,23 +178,28 @@ impl<S> SynchronizerHandle<S> {
         max_wait: std::time::Duration,
         block_history: &BlockHistory,
         previous_block: Header,
-    ) {
+    ) -> Option<StateSyncMessage> {
         let extractor_id = &self.extractor_id;
         match timeout(max_wait, self.rx.recv()).await {
-            Ok(Some(b)) => self.transition(b, block_history),
+            Ok(Some(msg)) => {
+                self.transition(msg.header.clone(), block_history);
+                Some(msg)
+            }
             Ok(None) => {
                 error!(
                     %extractor_id,
                     ?previous_block,
-                    "Extractor terminated: closed header channel!"
+                    "Extractor terminated: channel closed!"
                 );
                 self.state = SynchronizerState::Ended;
                 self.modify_ts = Local::now().naive_utc();
+                None
             }
             Err(_) => {
                 // trying to advance a block timed out
                 debug!(%extractor_id, ?previous_block, "Extractor did not check in within time.");
                 self.state = SynchronizerState::Delayed(previous_block.clone());
+                None
             }
         }
     }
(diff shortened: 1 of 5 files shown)


tycho-types/src/dto.rs
```diff
@@ -42,6 +42,16 @@ pub enum ChangeType {
     Unspecified,
 }
 
+impl ChangeType {
+    pub fn merge(&self, other: &Self) -> Self {
+        if matches!(self, Self::Creation) {
```
zizou0x, line 47: I just read about this macro and I don't understand why/how you choose to use it here rather than `==`?

zizou0x: It seems it's more flexible than `==` because it doesn't require the `PartialEq` trait, and also it allows for more complex comparison like `matches!(x, Enum::Value(v) if v < 30)`. Is that it?

tycho-client/src/feed/block_history.rs
```diff
@@ -0,0 +1,287 @@
...
+impl BlockHistory {
+    pub fn new(history: Vec<Header>, size: usize) -> Self {
+        Self {
+            history: VecDeque::from(history),
+            size,
+            reverts: LruCache::new(NonZeroUsize::new(size * 10).expect("cache size is 0")),
+        }
+    }
+
+    // How does this behave if empty?
```
louise-poole: If the block is NextExpected then it'll error with "Reverting block's insert position not found! History exceeded.", no? Otherwise it does nothing?

tycho-client/src/feed/block_history.rs
```diff
@@ -0,0 +1,287 @@
...
+                                    "Reverting block's insert position not found! History exceeded."
+                                )
+                            })?;
+                            // record reverted blocks in cache
+                            self.reverts
+                                .push(reverted_block.hash.clone(), reverted_block);
+                        }
+                    }
+                }
+                // Final sanity check against things going awefully wrong.
```
louise-poole:
```suggestion
                // Final sanity check against things going awfully wrong.
```

tycho-types/src/dto.rs
```diff
@@ -267,6 +439,51 @@ pub struct ProtocolStateDelta {
...
+    ///     - If an attribute was updated and is deleted in the next message, it is removed from
+    ///       updated attributes and kept in deleted. This way the attributes temporary update (or
+    ///       potentially short lived existence) before its deletion is never communicated to the
+    ///       client.
+    pub fn merge(&mut self, other: &Self) {
+        // either updated and then deleted -> keep in deleted, remove from updated
+        self.updated_attributes
+            .retain(|k, _| !other.deleted_attributes.contains(k));
+
+        // or deleted and then updated/recreated -> remove from deleted an keep in updated
```
louise-poole:
```suggestion
        // or deleted and then updated/recreated -> remove from deleted and keep in updated
```

tycho-types/src/dto.rs
```diff
@@ -267,6 +439,51 @@ pub struct ProtocolStateDelta {
     pub deleted_attributes: HashSet<String>,
 }
 
+impl ProtocolStateDelta {
+    /// Merges 'other' into 'self'.
+    ///
+    ///
+    /// During merge of these deltas a special situation can arrise when an attribute is present in
+    /// `self.deleted_attributes` and `other.update_attributes``. If we would just merge the sets
+    /// of deleted attribtues or vice versa, it would be ambiguous and potential lead to a
```
louise-poole:
```suggestion
    /// During merge of these deltas a special situation can arise when an attribute is present in
    /// `self.deleted_attributes` and `other.update_attributes``. If we would just merge the sets
    /// of deleted attributes or vice versa, it would be ambiguous and potential lead to a
```

tycho-client/src/feed/synchronizer.rs
```diff
@@ -0,0 +1,995 @@
...
+pub struct VMSnapshot {
+    state: HashMap<Bytes, ResponseAccount>,
+    component: ProtocolComponent,
+}
+
+#[derive(Clone, PartialEq, Debug)]
+pub struct NativeSnapshot {
+    state: ResponseProtocolState,
+    component: ProtocolComponent,
+}
```
louise-poole, line 58: Can this be moved to `dto.rs`? Would be nice to access this struct definition on protosim via the tycho-types crate.

tycho-client/src/feed/block_history.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    /// Determines the blocks position relative to current history.
+    ///
+    /// If there is no history we'll return an error here. This will also error if we
+    /// have a single block and we encounter a revert as it will be impossible to
+    /// find the fork block.
+    pub fn determine_block_position(&self, block: &Header) -> anyhow::Result<BlockPosition> {
+        let latest = self
+            .latest()
+            .ok_or_else(|| anyhow::format_err!("history empty"))?;
```
zizou0x, line 101: Should we start using anyhow everywhere else?

tycho-client/src/feed/synchronizer.rs
```diff
@@ -0,0 +1,995 @@
...
+#[derive(Clone, PartialEq, Debug)]
+pub struct StateSyncMessage {
+    /// The block number for this update.
+    pub header: Header,
+    /// Snapshot for new components.
+    pub snapshots: HashMap<String, Snapshot>,
+    /// A single delta contains state updates for all tracked components, as well as additional
+    /// information about the system components e.g. newly added components (even below tvl), tvl
+    /// updates, balance updates.
+    pub deltas: Option<Deltas>,
```
louise-poole, line 75: Is this over a single block, or all blocks since the snapshot?

tycho-client/src/feed/block_history.rs
zizou0x: If empty `self.determine_block_position` would return an error

tycho-client/src/feed/synchronizer.rs
```diff
@@ -0,0 +1,995 @@
...
+            is_native: native,
+            rpc_client,
+            deltas_client,
+            min_tvl_threshold: min_tvl,
+            max_retries,
+            shared: Arc::new(Mutex::new(SharedState {
+                last_served_block_hash: None,
+                last_synced_block: None,
+                pending: None,
+            })),
```
louise-poole: Personal preference: if we add `#[derive(Default)]` on the `SharedState` struct, we can simplify this line a lot:
```suggestion
            shared: Arc::new(Mutex::new(SharedState::default())),
```

tycho-client/src/feed/synchronizer.rs
```diff
@@ -0,0 +1,995 @@
...
+                .map(|acc| (acc.address.clone(), acc))
+                .collect::<HashMap<_, _>>();
+
+            trace!(states=?&contract_state, "Retrieved ContractState");
+            Ok(StateSyncMessage {
+                header,
+                // iteration over all component is not ideal for performance but reduces the
+                // required state and state updating. Since we e.g. have no mapping from
+                // contract_address to corresponding protocol component. Snapshots should not be
+                // retrieved too frequently, so we are ok for now this can be optimised should it be
```
louise-poole:
```suggestion
                // retrieved too frequently, so we are ok for now. This can be optimised should it be
```

tycho-client/src/feed/synchronizer.rs
```diff
@@ -0,0 +1,995 @@
...
+                            .iter()
+                            .filter_map(|contract_address| {
+                                // Cloning is essential to prevent mistakenly assuming a component
+                                // lacks associated state due to the m2m relationship between
+                                // contracts and components.
+                                if let Some(state) = contract_state.get(contract_address) {
+                                    Some((contract_address.clone(), state.clone()))
+                                } else if contract_ids.contains(contract_address) {
+                                    // only emit error even if we did actually request this
+                                    // address.
```
louise-poole:
```suggestion
                                    // only emit error event if we requested this address
```

tycho-client/src/feed/block_history.rs
```diff
@@ -0,0 +1,287 @@
...
+/// BlockHistory
+///
+/// Provides lightweight validation and relative positioning of received block headers
+/// emitted by StateSynchronizer structs.
+impl BlockHistory {
+    pub fn new(history: Vec<Header>, size: usize) -> Self {
+        Self {
+            history: VecDeque::from(history),
+            size,
+            reverts: LruCache::new(NonZeroUsize::new(size * 10).expect("cache size is 0")),
```
zizou0x, line 36: Do we need it to be that big? `10*size` will be a lot no?

Edit: I see we init this with 15 as size, 150 blocks in the lru cache is overkill I think

tycho-client/src/feed/component_tracker.rs
```diff
@@ -0,0 +1,261 @@
...
+                .map(|pc| (pc.id.clone(), pc)),
+        );
+        self.update_contracts();
+        debug!("Sucessfully started tracking new components");
+        Ok(())
+    }
+
+    /// Stop tracking components
+    #[instrument(skip(self))]
+    pub fn stop_tracking<'a, I: IntoIterator<Item = &'a String> + std::fmt::Debug>(
```
zizou0x, line 108: I would expect both `start_tracking` and `stop_tracking` to be generic or not, do we need the generics here?

tycho-client/src/feed/mod.rs
```diff
@@ -0,0 +1,641 @@
...
+///
+/// We will only query synchronizer for state if they are in the ready state or if they are
+/// advanced in which case we simply take the advanced state. For others, we will inform
+/// clients about their current sync state but not actually emit old or advanced data.
+///
+/// Any extractors that have been stale or delayed for too long we will stop waiting for - so
+/// we will end their synchronizers thread and drop the handle.
+///
+/// Any advanced extractors, we will not wait for in the future, they will be assumed ready
+/// until we reach their height and from then on are handles is they were ready.
```
louise-poole:
```suggestion
/// until we reach their height and from then on are handled as if they were ready.
```

tycho-client/src/feed/mod.rs
```diff
@@ -0,0 +1,641 @@
...
+                            });
+                        }
+                        _ => continue,
+                    }
+                }
+                let synced_msgs = join_all(pulled_data_fut)
+                    .await
+                    .into_iter()
+                    // TODO: justification to ignore missing data here? IMO indicates a bug if a
+                    //  ready extractor has not data available.
```
louise-poole: Agreed

tycho-client/src/feed/mod.rs
```diff
@@ -0,0 +1,641 @@
...
+                let mut recv_futures = Vec::new();
+                for sh in sync_handles.values_mut() {
+                    recv_futures
+                        .push(sh.try_advance(&block_history, self.block_time + self.max_wait));
+                }
+                join_all(recv_futures).await;
+
+                // Purge any bad synchronizers, respective warnings have already been issued at
+                // transition time.
+                sync_handles.retain(|_, v| match v.state {
```
louise-poole: I really like this `retain` function - wish I knew about it earlier. I always used a filter map, but this is much cleaner

tycho-client/src/feed/mod.rs
```diff
@@ -0,0 +1,641 @@
...
+                            .clone(),
+                    )
+                    .map_err(|err| anyhow::format_err!("Failed processing new block: {}", err))?;
+            }
+        });
+
+        let nanny_jh = tokio::spawn(async move {
+            select! {
+                error = state_sync_tasks.select_next_some() => {
+                    // TODO: close all synchronizers
```
louise-poole: She feels lonely and forgotten ☝️

tycho-client/src/feed/synchronizer.rs
```diff
@@ -0,0 +1,995 @@
...
+        extractor_id: ExtractorIdentity,
+        native: bool,
+        min_tvl: f64,
+        max_retries: u64,
+        rpc_client: R,
+        deltas_client: D,
+    ) -> Self {
+        Self {
+            extractor_id,
+            is_native: native,
```
zizou0x: Small detail, we could name them the same

tycho-client/src/feed/synchronizer.rs
```diff
@@ -0,0 +1,995 @@
...
+                removed_components: HashMap::new(),
+            })
+        } else {
+            let component_ids = tracked_components.get_tracked_component_ids();
+
+            let mut protocol_states = self
+                .rpc_client
+                .get_protocol_states(
+                    self.extractor_id.chain,
+                    &Default::default(),
```
zizou0x, line 253: Now that I see this logic, we can remove this from the RPC

tycho-client/src/feed/synchronizer.rs
kayibal: Good idea. I think we'll move most of these types there yes.

tycho-client/src/feed/synchronizer.rs
kayibal: All deltas since the last time we got an update.

tycho-client/src/feed/synchronizer.rs
```diff
@@ -0,0 +1,995 @@
...
+            .unwrap_or_else(|| {
+                tracked_components
+                    .components
+                    .keys()
+                    .collect::<Vec<_>>()
+            });
+        if !self.is_native {
+            let contract_ids = tracked_components.get_contracts_by_component(ids);
+
+            let contract_state = self
```
zizou0x:
```suggestion
            let contract_states = self
```

tycho-client/src/feed/synchronizer.rs
```diff
@@ -0,0 +1,995 @@
...
+        let first_msg = msg_rx
+            .recv()
+            .await
+            .ok_or_else(|| anyhow::format_err!("Subscription ended too soon"))?;
+
+        // initial snapshot
+        let block = first_msg.get_block().clone();
+        let header = Header::from_block(first_msg.get_block(), first_msg.is_revert());
+        let snapshot = self
+            .get_snapshots::<Vec<&String>>(Header::from_block(&block, false), &tracker, None)
```
zizou0x, line 334: I'm confused here:

What if the first_msg is a revert? 

Also, I'm not sure but I think `first_msg` write operation won't be flushed yet no? Don't we need to request the snapshot at the msg ignored above `let _ = msg_rx` and then apply `first_msg`?

tycho-client/src/feed/synchronizer.rs
```diff
@@ -0,0 +1,995 @@
...
+                            } else {
+                                Some(*k)
+                            }
+                        })
+                        .collect::<Vec<_>>();
+                    tracker
+                        .start_tracking(&requiring_snapshot)
+                        .await?;
+                    let snapshots = self
+                        .get_snapshots(header.clone(), &tracker, Some(requiring_snapshot))
```
zizou0x: Same here, the data won't be flushed yet no?

tycho-client/src/feed/mod.rs
```diff
@@ -0,0 +1,641 @@
...
+            number: block.number,
+            parent_hash: block.parent_hash.clone(),
+            revert,
+        }
+    }
+}
+
+type BlockSyncResult<T> = anyhow::Result<T>;
+
+/// Synchronizes multiple StgteSynchronizers on the time dimension.
```
zizou0x: @louise-poole missed this one hehe

```suggestion
/// Synchronizes multiple StateSynchronizers on the time dimension.
```

tycho-client/src/feed/mod.rs
```diff
@@ -0,0 +1,641 @@
...
+    // StateSynchronizer. This can only happen if we are processing too slow and the one of the
+    // synchronizers restarts e.g. because Tycho ended the subscription.
+    Advanced(Header),
+    Ended,
+}
+
+pub struct SynchronizerHandle<S> {
+    extractor_id: ExtractorIdentity,
+    state: SynchronizerState,
+    sync: S,
```
zizou0x: What is this for?

tycho-client/src/feed/block_history.rs
kayibal: No actually I should not be using it here (since this is a library) but ran out of time... I promise I will later introduce nice error types.

tycho-client/src/feed/block_history.rs
kayibal: Basically the reasoning is we support keeping up to 10 branches of 15 blocks.... But we should probably not hardcode this. making it too low would also not be good. Anyway won't consume too much memory with 150 blocks either so I prefer to big rather then to small.

tycho-client/src/feed/component_tracker.rs
kayibal: I implemented what made the usage easiest in the parent class. I guess we don't need them but they also don't hurt.

tycho-client/src/feed/synchronizer.rs
kayibal: What do you mean?

tycho-client/src/feed/synchronizer.rs
kayibal: Good catch I think you are right!

tycho-client/src/feed/mod.rs
kayibal: Yeah currently unused but it is meant to shut down and deallocate the StateSynchronizer structs once we exit.

tycho-types/src/dto.rs
kayibal: I didn't know you could use == here.

PR #145, opened 2024-02-06
Add balance_float to component_balance table.

kayibal:
This is necessary so we can compute the tvl of each component with a sql query.


tycho-indexer/src/extractor/u256_num.rs
kayibal, line 1: I copied this from protosim, should be safe to use. No need to review this in detail unless you are interested into some f64 binary black magics.

PR #146, opened 2024-02-06
ENG-2740: Token pre processor

dianacarvalho1:
[Issue](https://datarevenue.atlassian.net/browse/ENG-2740)

Question:
- how should we handle errors properly? 🤔  what if we can't find a certain token?? that whole component should be invalid. Should we double check in the ambient gtw later that all tokens are inserted in the DB and the ones that aren't we remove the component? I'm not so sure

Answer:
- If symbol is not available default to the hex encoded address string
- If decimals not available, default to 18 decimals

### DIFF
--- a/tycho-indexer/src/extractor/evm/mod.rs
+++ b/tycho-indexer/src/extractor/evm/mod.rs
@@ -1086,10 +1086,8 @@ pub struct BlockEntityChanges {
     pub block: Block,
     pub revert: bool,
     pub txs_with_update: Vec<ProtocolChangesWithTx>,
-    pub new_protocol_components: HashMap<String, ProtocolComponent>,
 }
 
-// TODO: remove dead code check skip once extractor is implemented
 impl BlockEntityChanges {
     /// Parse from tychos protobuf message
     pub fn try_from_message(
@@ -1101,49 +1099,38 @@ impl BlockEntityChanges {
     ) -> Result<Self, ExtractionError> {
         if let Some(block) = msg.block {
             let block = Block::try_from_message(block, chain)?;
-            let mut txs_with_update: Vec<ProtocolChangesWithTx> = Vec::new();
-            let mut new_protocol_components = HashMap::new();
 
-            for change in msg.changes.into_iter() {
-                if let Some(tx) = change.tx.clone() {
-                    let tycho_tx = Transaction::try_from_message(tx, &block.hash)?;
-
-                    let tx_update = ProtocolChangesWithTx::try_from_message(
-                        change.clone(),
-                        &block,
-                        protocol_system,
-                        protocol_types,
-                    )?;
-
-                    txs_with_update.push(tx_update);
-                    for component in change.component_changes {
-                        let pool = ProtocolComponent::try_from_message(
-                            component,
-                            chain,
+            let mut txs_with_update: Vec<ProtocolChangesWithTx> = msg
+                .changes
+                .into_iter()
+                .filter_map(|change| {
+                    change.tx.as_ref()?;
+                    Some(
+                        ProtocolChangesWithTx::try_from_message(
+                            change,
+                            &block,
                             protocol_system,
                             protocol_types,
-                            tycho_tx.hash,
-                            block.ts,
-                        )?;
-                        new_protocol_components.insert(pool.id.clone(), pool);
-                    }
-                }
-            }
+                        )
+                        .ok(),
+                    )
+                })
+                .collect::<Option<Vec<_>>>()
+                .expect("Failed to parse changes");
 
-            // Sort updates by transaction by index
+            // Sort updates by transaction index
             txs_with_update.sort_unstable_by_key(|update| update.tx.index);
 
-            return Ok(Self {
-                extractor: extractor.to_owned(),
+            Ok(Self {
+                extractor: extractor.to_string(),
                 chain,
                 block,
                 revert: false,
                 txs_with_update,
-                new_protocol_components,
-            });
+            })
+        } else {
+            Err(ExtractionError::Empty)
         }
-
-        Err(ExtractionError::Empty)
     }
 
     /// Aggregates state updates.
@@ -1178,7 +1165,7 @@ impl BlockEntityChanges {
             block: self.block,
             revert: self.revert,
             state_updates: aggregated_changes.protocol_states,
-            new_protocol_components: self.new_protocol_components,
+            new_protocol_components: aggregated_changes.new_protocol_components,
         })
     }
 }
@@ -1780,8 +1767,8 @@ mod test {
                             ChangeType::Update,
                         ),
                     )]
-                    .into_iter()
-                    .collect(),
+                        .into_iter()
+                        .collect(),
                     protocol_components: [(protocol_component.id.clone(), protocol_component)]
                         .into_iter()
                         .collect(),
@@ -1797,11 +1784,11 @@ mod test {
                                 component_id: "WETH-CAI".to_string(),
                             },
                         )]
+                            .into_iter()
+                            .collect(),
+                    )]
                         .into_iter()
                         .collect(),
-                    )]
-                    .into_iter()
-                    .collect(),
 
                     tx,
                 },
@@ -1820,8 +1807,8 @@ mod test {
                             ChangeType::Update,
                         ),
                     )]
-                    .into_iter()
-                    .collect(),
+                        .into_iter()
+                        .collect(),
                     protocol_components: HashMap::new(),
                     component_balances: [(
                         "WETH-CAI".to_string(),
@@ -2249,7 +2236,6 @@ mod test {
                     ..Default::default()
                 },
             ],
-            new_protocol_components,
         }
     }
 
--- a/tycho-indexer/src/extractor/evm/native.rs
+++ b/tycho-indexer/src/extractor/evm/native.rs
@@ -765,7 +765,6 @@ mod test_serial_db {
                     },
                 )]),
             }],
-            new_protocol_components: HashMap::new(),
         }
     }
 
@@ -874,7 +873,6 @@ mod test_serial_db {
                     balance_changes: HashMap::new(),
                     new_protocol_components: HashMap::new(),
                 }],
-                new_protocol_components: HashMap::new(),
             };
 
             gw.forward(&msg0, "cursor@0")


tycho-indexer/src/extractor/evm/token_pre_processor.rs
```diff
@@ -0,0 +1,90 @@
...
+            match (symbol, decimals) {
+                (Ok(symbol), Ok(decimals)) => tokens_info.push(Ok(ERC20Token {
+                    address,
+                    symbol,
+                    decimals: decimals.into(),
+                    tax: 0,
+                    gas: vec![],
+                    chain: Chain::Ethereum,
+                })),
+                (Err(e), _) | (_, Err(e)) => println!("Error fetching token data: {:?}", e),
```
dianacarvalho1: eheh

tycho-indexer/src/extractor/evm/token_pre_processor.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    pub async fn get_tokens(
+        &self,
+        addresses: Vec<H160>,
+    ) -> Vec<Result<ERC20Token, Box<dyn std::error::Error>>> {
+        let mut tokens_info = Vec::new();
+
+        let abi_str = fs::read_to_string("src/extractor/evm/abi/erc20.json")
+            .expect("Unable to read ABI file");
+        let abi = from_str::<Abi>(&abi_str).expect("Unable to parse ABI");
```
SmartMonkey-git, line 17: Is it not possible to do this in the new function and keep the abi as an attribute?

tycho-indexer/src/extractor/evm/token_pre_processor.rs
SmartMonkey-git: I know you are unsure at the moment what to do, but if you want to keep the error message, I would use the tracing module.
`use tracing::{debug, error, info, instrument, warn};`

tycho-indexer/src/extractor/evm/token_pre_processor.rs
```diff
@@ -0,0 +1,90 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use ethers::types::H160;
+    use std::str::FromStr;
+
+    #[tokio::test]
+    async fn test_get_tokens() {
+        let rpc_url = "https://eth-mainnet.g.alchemy.com/v2/OTD5W7gdTPrzpVot41Lx9tJD9LUiAhbs";
+        let processor = TokenPreProcessor::new(rpc_url);
```
SmartMonkey-git: It looks like you are actually querying the RPC in this test.
The ethers crate offers a [MockProvider](https://docs.rs/ethers/latest/ethers/providers/struct.MockProvider.html#). Its also used in [protosim](https://github.com/search?q=repo%3Apropeller-heads%2Fprotosim%20MockProvider&type=code), maybe you can fine some inspiration there.

SmartMonkey-git: I think you would need to make the provider type generic on the `TokenPreProcessor` to use the MockProvider.

tycho-indexer/src/extractor/evm/token_pre_processor.rs
```diff
@@ -0,0 +1,90 @@
...
+            .expect("Error creating HTTP provider")
+            .into();
+
+        TokenPreProcessor { client }
+    }
+
+    pub async fn get_tokens(
+        &self,
+        addresses: Vec<H160>,
+    ) -> Vec<Result<ERC20Token, Box<dyn std::error::Error>>> {
```
SmartMonkey-git: Is it not possible to specify the error here?

tycho-indexer/src/extractor/evm/token_pre_processor.rs
dianacarvalho1: uuhh nice! yeah I will have a look at that! I wanted the test to use the RPC just to make sure everything is working. thank you 🙏🏼

tycho-indexer/src/extractor/evm/token_pre_processor.rs
dianacarvalho1: what do you mean?

tycho-indexer/src/extractor/evm/token_pre_processor.rs
```diff
@@ -0,0 +1,90 @@
...
+
+            let symbol: Result<String, _> = contract
+                .method("symbol", ())
+                .expect("Error preparing request")
+                .call()
+                .await;
+
+            let decimals: Result<u8, _> = contract
+                .method("decimals", ())
+                .expect("Error preparing request")
```
tamaralipows, line 36: Maybe we should specify the method here in case of failure?

tycho-indexer/src/extractor/evm/token_pre_processor.rs
SmartMonkey-git: As far as I know `Box<dyn std::error::Error>>` means this function can return any kind of error. The only error your function can return is `ContractError`, which gets propagated by the `call()` function.

Also do you think it is possible to change the return type to `Result<Vec<ERC20Token>, Error>>`. In this case you will be able to check with a single action if it is an error or a value, in the current case you will need to check it for each entry, i think. The change would lead to a better dev experience.

SmartMonkey-git: If you are struggling to much with implementing the `ContractError`, then I can live with it being the current interface. It does not look straight forward to implement it.

tycho-indexer/src/extractor/evm/token_pre_processor.rs
dianacarvalho1: damn it, I've been trying to make this work for a while now and can't figure it out for the life of me 😩 @SmartMonkey-git maybe you can check this tomorrow morning? you did this for protosim no?

tycho-indexer/src/extractor/evm/token_pre_processor.rs
```diff
@@ -0,0 +1,93 @@
+use crate::{extractor::evm::ERC20Token, models::Chain};
+use ethers::{abi::Abi, contract::Contract, prelude::Middleware, types::H160};
+use serde_json::from_str;
+use std::{fs, sync::Arc};
+
+pub struct TokenPreProcessor<M: Middleware> {
+    pub client: Arc<M>,
```
SmartMonkey-git: I think the pub is a leftover from the generics. Don't think this needs to be public, does it?

tycho-indexer/src/extractor/evm/token_pre_processor.rs
```diff
@@ -0,0 +1,93 @@
...
+
+        let results = processor.get_tokens(addresses).await;
+
+        assert_eq!(results.len(), 3);
+        assert_eq!(results[0].symbol, "WETH");
+        assert_eq!(results[0].decimals, 18);
+        assert_eq!(results[1].symbol, "USDC");
+        assert_eq!(results[1].decimals, 6);
+        assert_eq!(results[2].symbol, "0xa0b8…eb48");
+        assert_eq!(results[2].decimals, 18);
```
SmartMonkey-git: I would rather see something like `asser_eq!(result[0],ERC20Token {....} )`

tycho-indexer/src/extractor/evm/token_pre_processor.rs
```diff
@@ -0,0 +1,93 @@
...
+            .expect("Unable to read ABI file");
+        let abi = from_str::<Abi>(&abi_str).expect("Unable to parse ABI");
+        TokenPreProcessor { client: Arc::new(client), erc20_abi: abi }
+    }
+
+    pub async fn get_tokens(&self, addresses: Vec<H160>) -> Vec<ERC20Token> {
+        let mut tokens_info = Vec::new();
+
+        for address in addresses {
+            let contract = Contract::new(address, self.erc20_abi.clone(), self.client.clone());
```
tvinagre, line 26: Feels quite inefficient having to clone the ABI & Client here for every token

tycho-indexer/src/extractor/evm/token_pre_processor.rs
```diff
@@ -0,0 +1,93 @@
...
+    use super::*;
+    use ethers::{
+        providers::{Http, Provider},
+        types::H160,
+    };
+    use std::str::FromStr;
+
+    #[tokio::test]
+    async fn test_get_tokens() {
+        let rpc_url = "https://eth-mainnet.g.alchemy.com/v2/OTD5W7gdTPrzpVot41Lx9tJD9LUiAhbs";
```
tvinagre: Can we use a public node, env_vars or skip the test on CI?
We're leaking an API key here 😕

tycho-indexer/src/extractor/evm/token_pre_processor.rs
dianacarvalho1: how else could we do this?

kayibal: Client is in an Arc so cloning is super cheap. Not sure about ABI but wouldn't bother for now.

tycho-indexer/src/extractor/evm/token_pre_processor.rs
```diff
@@ -0,0 +1,93 @@
...
+
+impl<M: Middleware> TokenPreProcessor<M> {
+    pub fn new(client: M) -> Self {
+        let abi_str = fs::read_to_string("src/extractor/evm/abi/erc20.json")
+            .expect("Unable to read ABI file");
+        let abi = from_str::<Abi>(&abi_str).expect("Unable to parse ABI");
+        TokenPreProcessor { client: Arc::new(client), erc20_abi: abi }
+    }
+
+    pub async fn get_tokens(&self, addresses: Vec<H160>) -> Vec<ERC20Token> {
```
kayibal, line 22: Could you please instrument this method? I think it would be good to know how often we call it and how long it takes. There is a macro for it: `#[instrument]`

tycho-indexer/src/extractor/evm/token_pre_processor.rs
tvinagre: Could you use `&`? I'm not sure

tvinagre: Since `Contract` is not ours it makes it a little harder - but I was just pointing out, no big deal here

tycho-indexer/src/extractor/evm/token_pre_processor.rs
kayibal: Yes, please convert this into a single assert. If the result changes for whatever reason figuring out why or what changed will be much harder.
I assume you only want to assert symbol and decimals? So you can use :

```
result.iter().map(|t| (t.symbol, t.decimals)).collect()
```

tycho-indexer/src/extractor/evm/token_pre_processor.rs
kayibal: Use the env variable `ETH_RPC_URL` it should be available on the CI - maybe we can also skip the test if the variable is not set.

I would also be ok to simply skip this test by default.

tycho-indexer/src/extractor/evm/token_pre_processor.rs
SmartMonkey-git: > how else could we do this?

Instead of just loading the abi in new, cant you just also initialise the contract there and keep it as an attribute?

dianacarvalho1: @SmartMonkey-git the contract depends on the address

SmartMonkey-git: True

PR #149, opened 2024-02-07
[FIX] add_component_balances: Filter by chain.

tamaralipows:
The component external id isn't chain unique, so before this change we might have been updating balances for a similarly named component on a different chain

### DIFF
--- a/tycho-indexer/src/extractor/evm/ambient.rs
+++ b/tycho-indexer/src/extractor/evm/ambient.rs
@@ -267,14 +267,28 @@ impl AmbientPgGateway {
 
         let target = BlockOrTimestamp::Block(to.clone());
         let address = H160(AMBIENT_CONTRACT);
-        let account_updates = self
+        let (account_updates, _, component_balances) = self
             .state_gateway
             .get_delta(&self.chain, start.as_ref(), &target)
-            .await?
-            .0
+            .await?;
+        let account_updates = account_updates
             .into_iter()
             .filter_map(|u| if u.address == address { Some((u.address, u)) } else { None })
             .collect();
+
+        let mut component_balances_map: HashMap<
+            evm::ComponentId,
+            HashMap<H160, evm::ComponentBalance>,
+        > = HashMap::new();
+        for balance in component_balances {
+            let component_id = balance.component_id.clone();
+            let h160 = balance.token;
+            let inner_map = component_balances_map
+                .entry(component_id)
+                .or_default();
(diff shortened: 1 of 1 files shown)


tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -536,6 +536,7 @@ where
         new: &[(TxHash, &ProtocolStateDelta)],
         conn: &mut Self::DB,
     ) -> Result<(), StorageError> {
+        let chain_id = self.get_chain_id(chain);
```
tamaralipows: @louise-poole I think it was also missing here in `update_protocol_states`

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -475,10 +475,15 @@ pub struct NewProtocolComponent {
...
+        chain_db_id: i64,
         conn: &mut AsyncPgConnection,
     ) -> QueryResult<Vec<(i64, String)>> {
         protocol_component::table
-            .filter(protocol_component::external_id.eq_any(external_ids))
+            .filter(
+                protocol_component::external_id
+                    .eq_any(external_ids)
+                    .and(protocol_component::chain_id.eq(chain_db_id)),
+            )
```
louise-poole: personal preference because I think it's easier to read:
```suggestion
            .filter(protocol_component::external_id.eq_any(external_ids))
            .filter(protocol_component::chain_id.eq(chain_db_id))
```

tycho-indexer/src/storage/postgres/protocol.rs
louise-poole: Thanks! Can we maybe also name this `chain_db_id` so that it's clear it's a db identifier?

tycho-indexer/src/storage/postgres/orm.rs
tamaralipows: @louise-poole You'll have to fight with @SmartMonkey-git on that one

dianacarvalho1: tbh I agree with @louise-poole on this one. it's the weird indentation for me

SmartMonkey-git: In the end its taste. Do which ever you like more. Be it brown, white or even rock candy syntax.

tamaralipows: settled. no more using `and` ever

PR #150, opened 2024-02-07
Add price, tvl tables and enable filtering on components

kayibal:
We will need the token prices table filled and the aggregation query filled for the tvl filter to work. I will add a python task for this. 

Long term we will use the prices produced by the NTQ for this. Unfortunately this will create a cyclical dependency if the NTQ uses tvl information. To mitigate this later we can adjust the NTQ to fall back to a simpler tvl filter methodology e.g. by only requesting components that are paired with WETH and or stables then use that first pass for initial prices.

For now we can take the python tvl script which produces prices to calculate tvl values every 3h.

### DIFF
--- a/tycho-indexer/src/storage/postgres/orm.rs
+++ b/tycho-indexer/src/storage/postgres/orm.rs
@@ -26,7 +26,7 @@ use super::{
         account, account_balance, block, chain, component_balance, component_tvl, contract_code,
         contract_storage, extraction_state, protocol_component, protocol_component_holds_contract,
         protocol_component_holds_token, protocol_state, protocol_system, protocol_type, token,
-        token_price, transaction,
+        transaction,
     },
     versioning::{DeltaVersionedRow, StoredVersionedRow, VersionedRow},
 };
@@ -1381,17 +1381,6 @@ pub struct ProtocolHoldsToken {
 pub struct NewProtocolComponentHoldsToken {
     pub protocol_component_id: i64,
     pub token_id: i64,
-}
-
-#[derive(Identifiable, Queryable, Associations, Selectable, Debug)]
-#[diesel(belongs_to(Token))]
-#[diesel(table_name = token_price)]
-#[diesel(check_for_backend(diesel::pg::Pg))]
-pub struct TokenPrice {
-    id: i64,
-    token_id: i64,
-    pub inserted_ts: NaiveDateTime,
-    pub modified_ts: NaiveDateTime,
 }
 
 #[derive(Identifiable, Queryable, Associations, Selectable, Debug)]
--- a/tycho-indexer/src/storage/postgres/protocol.rs
+++ b/tycho-indexer/src/storage/postgres/protocol.rs
@@ -329,11 +329,7 @@ where
         let inserted_protocol_components: Vec<(i64, String, i64, i64)> =
             diesel::insert_into(protocol_component)
                 .values(&values)
-                .on_conflict((
-                    schema::protocol_component::chain_id,
-                    protocol_system_id,
-                    external_id,
-                ))
+                .on_conflict((schema::protocol_component::chain_id, external_id))
                 .do_nothing()
                 .returning((
                     schema::protocol_component::id,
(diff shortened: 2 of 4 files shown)


tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -967,6 +1001,56 @@ where
...
+                schema::protocol_component::external_id,
+                schema::account::address,
+                schema::component_balance::balance_float,
+            ))
+            .filter(
+                schema::component_balance::valid_to
+                    .gt(version_ts) // if version_ts is None, diesel equates this expression to "False"
+                    .or(schema::component_balance::valid_to.is_null()),
+            )
+            .into_boxed();
```
tamaralipows: Chain is unused here, shouldn't we use it for the filtering?

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -967,6 +1001,56 @@ where
...
+
+        if let Some(external_ids) = ids {
+            q = q.filter(schema::protocol_component::external_id.eq_any(external_ids))
+        }
```
tamaralipows, line 1025: I think I just solved something similar in my most recent PR - the external_id is not unique across chains

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -1146,32 +1230,55 @@ where
...
-                })?;
-            Ok(inserted_protocol_system.id)
-        }
+        let upsert_map: HashMap<_, _> = tvl_values
+            .iter()
+            .filter_map(|(component_id, v)| {
+                if let Some(db_id) = external_db_id_map.get(component_id) {
+                    Some((*db_id, *v))
+                } else {
+                    warn!(?component_id, "Tried to upsert tvl for unknown compoent!");
```
tamaralipows:
```suggestion
                    warn!(?component_id, "Tried to upsert tvl for unknown component!");
```

tycho-indexer/src/storage/postgres/protocol.rs
```diff
@@ -1146,32 +1230,55 @@ where
...
-            return Ok(entry.id);
-        } else {
-            let new_entry = orm::NewProtocolSystem { name: new.to_string() };
+        let external_db_id_map: HashMap<String, i64> = protocol_component
+            .select((external_id, id))
+            .filter(external_id.eq_any(tvl_values.keys()))
+            .get_results::<(String, i64)>(conn)
+            .await?
+            .into_iter()
+            .collect::<HashMap<_, _>>();
```
tamaralipows: We have an existing method for this `ids_by_external_ids` that would work in this case AFAIU

tycho-indexer/src/storage/postgres/orm.rs
```diff
@@ -1323,40 +1326,54 @@ pub struct NewProtocolComponentHoldsToken {
...
-    pub tokens: Vec<Token>,
+#[derive(Identifiable, Queryable, Associations, Selectable, Debug)]
+#[diesel(belongs_to(ProtocolComponent))]
+#[diesel(table_name = component_tvl)]
+#[diesel(check_for_backend(diesel::pg::Pg))]
+pub struct ComponentTVL {
+    id: i64,
+    protocol_component_id: i64,
+    pub inserted_ts: NaiveDateTime,
+    pub modified_ts: NaiveDateTime,
```
tamaralipows: Could you explain these to me? Where are the actual price and tvl values here? Are they hidden?

tycho-indexer/src/storage/postgres/protocol.rs
kayibal: How did clippy not notice?!

tycho-indexer/src/storage/postgres/orm.rs
kayibal: I don't use them so clippy complains. I might actually remove the orm objects but I think it may break some joins...

tycho-indexer/src/storage/postgres/protocol.rs
kayibal: But this method is chain scoped so it should be unique?

tamaralipows: Does that mean that the db we are querying is chain scoped? o_O

PR #151, opened 2024-02-07
ENG-2750: Extracting balance changes 

SmartMonkey-git:
This PR adjusts `BlockContractChanges.try_from_message()` to extract `ComponentBalance` from ProtoBuf messages.

Additionally, test fixtures have been updated to include the extracted `ComponentBalance`.

### DIFF
--- a/tycho-indexer/src/storage/postgres/versioning.rs
+++ b/tycho-indexer/src/storage/postgres/versioning.rs
@@ -110,6 +110,7 @@ pub trait StoredVersionedRow {
     fn get_pk(&self) -> Self::PrimaryKey;
 
     /// Exposes the end version.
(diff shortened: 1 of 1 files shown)


tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -697,10 +700,24 @@ impl BlockContractChanges {
                         )?;
                         protocol_components.insert(component.id.clone(), component);
                     }
+
+                    for balance_change in change.balance_changes.into_iter() {
+                        let component_id =
```
kayibal, line 712: Could you please double check if this is also done on the sibling native protocols struct?

SmartMonkey-git: Jup, it is

PR #153, opened 2024-02-08
feat(vm): realtime token & tvl tracking.

kayibal:
We emit now absolute tvl values and token from the extractors. This is done with the help of a cache. The cache provides protocol components, tokens and prices. 

The cache is shared between extractors since it might be very memory intensive. For simplicity cache does not do any special revert handling. I could not come up with a serious example where this may be dangerous. 

Worst case a component is created, the change is reverted and then it is created again with different attributes/tokens. 
In this case, the moment the component is seen for the second time the cache should have the correct component available -  in the meantime tvl might have been wrong and clients might have received a unrelated token.


tycho-indexer/src/extractor/evm/ambient.rs
```diff
@@ -398,19 +433,117 @@ where
...
+        // might be slow. We should potentially skip this step while the extractor
+        // is syncing, especially if we don't need historical tvl values. Additionally
+        // a smart cache can help here in case it becomes a bottleneck.
+        let mut balances = self
+            .gateway
+            .get_component_balances(&component_ids)
+            .await?;
+
+        // Update with balances of this message, the result could be cached in an
+        // LRU fashion if needed.
```
dianacarvalho1: couldn't we do this at the end when we already processed the block?

tycho-indexer/src/extractor/evm/ambient.rs
```diff
@@ -398,19 +433,117 @@ where
...
+                        });
+                })
+            });
+
+        // collect token decimals and prices to calculate tvl in the next step
+        // most of this data should be in the cache.
+        let addresses = balances
+            .iter()
+            .flat_map(|(_, b)| b.keys())
+            .collect::<Vec<_>>();
```
dianacarvalho1: should be a `Vec<Bytes>` no?

dianacarvalho1: this can have duplicated token addresses and then all the following code might be done repeatedly for the same token. We could gather the prices and the decimals only for a unique set of tokens and then use the corresponding one when calculating the TVL

tycho-indexer/src/extractor/evm/ambient.rs
kayibal: I am doing this here on the block aggregated data so it basically happens at the very end.

tycho-indexer/src/extractor/evm/ambient.rs
kayibal: This should return `HashMap<String, HashMap<Bytes, f64>>`

tycho-indexer/src/extractor/evm/protocol_cache.rs
```diff
@@ -0,0 +1,450 @@
...
+                .await?
+                .into_iter()
+                .for_each(|c| {
+                    n_fetched += 1;
+                    cached_components.insert(c.id.clone(), c);
+                });
+            debug!(n_missing = missing.len(), n_fetched, resource = "component", "CacheMiss");
+        }
+        let guard = self.components.read().await;
+        let cached_components = guard.get(system).unwrap_or(&empty);
```
louise-poole, line 245: There is one cache per extractor, no? If so, there will only ever be components for one protocol system in the cache and this line in unnecessary. Or is this a shared cache?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -500,6 +502,9 @@ pub struct BlockContractChanges {
     pub block: Block,
     pub finalized_block_height: u64,
     pub revert: bool,
+    /// Required here, so it is part of the revert buffer and thus inserted into storage once
+    /// finalized.
+    pub new_tokens: HashMap<Address, CurrencyToken>,
```
louise-poole, line 510: Not sure I understand this... are protocol components not also in the revert buffer? And they're in `TransactionVMUpdates`.

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -298,6 +192,8 @@ where
                 .await?;
         }
 
+        // TODO: send new tokens to db
+
```
louise-poole: 👀 isn't this already done using the state_gateway above?

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -442,12 +336,307 @@ where
...
+    #[allow(clippy::mutable_key_type)]
+    async fn handle_tvl_changes(
+        &self,
+        msg: &mut evm::BlockEntityChangesResult,
+    ) -> Result<(), ExtractionError> {
+        trace!("Calculating tvl changes");
+        if msg.component_balances.is_empty() {
+            return Ok(());
+        }
+        dbg!("A");
```
louise-poole: I think these can either be removed or made more useful by logging actual variables?

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -442,12 +336,307 @@ where
...
+                )
+            })
+            .fold(HashMap::new(), |mut acc, (c_id, b_changes)| {
+                acc.entry(c_id)
+                    .or_default()
+                    .extend(b_changes);
+                acc
+            });
+        Ok(combined_balances)
+    }
```
louise-poole, line 563: Breaking this up into smaller steps/sections would help a lot with readability

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -442,12 +336,307 @@ where
...
+
+    async fn process_tokens(
```
tvinagre: Also would be nice to have a smol doc explaining what `process` does

tycho-indexer/src/extractor/evm/protocol_cache.rs
kayibal: @louise-poole third sentence in the PRs description :D

tycho-indexer/src/extractor/evm/native.rs
kayibal: Upsie, left over from a deadlock that I had...

tycho-indexer/src/extractor/evm/mod.rs
kayibal: So I need to save the tokens that I extracted in the message, so it can be later inserted into the DB, once this block is final. Note that ProtocolComponent does not contain the complete token data but instead only the address of the tokens.

tycho-indexer/src/extractor/evm/native.rs
kayibal: It has inline comments with steps? Is that not enough?

tycho-indexer/src/extractor/evm/native.rs
kayibal: I mean I only moved things, I did not add this functionality here^^ but I can add this doc..

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -511,28 +693,44 @@ where
             Err(e) => return Err(e),
         };
 
-        let msg =
+        let mut msg =
             if let Some(post_process_f) = self.post_processor { post_process_f(msg) } else { msg };
 
+        msg.new_tokens = self
+            .construct_currency_tokens(&msg)
```
zizou0x, line 699: We could also do this

```
construct_currency_tokens(&mut msg)
```

PR #158, opened 2024-02-08
ENG-2743: Use token pre processor in ambient gtw

dianacarvalho1:
To be able to do this, I had to be able to mock it for the ambient gtw tests

### DIFF
--- a/tycho-client/src/rpc.rs
+++ b/tycho-client/src/rpc.rs
@@ -12,7 +12,11 @@ use tracing::{debug, error, info, instrument, warn};
 
 use async_trait::async_trait;
 
-use tycho_types::dto::{Chain, StateRequestBody, StateRequestParameters, StateRequestResponse};
+use tycho_types::dto::{
+    Chain, ProtocolComponentRequestParameters, ProtocolComponentRequestResponse,
+    ProtocolComponentsRequestBody, ProtocolStateRequestBody, ProtocolStateRequestResponse,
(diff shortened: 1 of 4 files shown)


tycho-indexer/src/extractor/evm/ambient.rs
```diff
@@ -151,35 +174,13 @@ impl AmbientPgGateway {
                     .cloned()
             })
             .collect();
-        let new_tokens = self
+        let new_tokens_addresses = self
             .get_new_tokens(protocol_components)
```
tvinagre, line 178: I've handled it in a slightly different way in my PR. My `get_new_tokens` just filters out the tokens that are already on DB, from a Vec<H160> - instead of `protocol_components`.
JFYI, I don't think it's a huge difference, but I feel like it's cleaner imo

tycho-indexer/src/extractor/evm/ambient.rs
```diff
@@ -735,9 +738,44 @@ mod test_serial_db {
     const TX_HASH_2: &str = "0xcf574444be25450fe26d16b85102b241e964a6e01d75dd962203d4888269be3d";
     const BLOCK_HASH_0: &str = "0x98b4a4fef932b1862be52de218cc32b714a295fae48b775202361a6fa09b66eb";
 
+    fn get_mocked_token_pre_processor() -> MockTokenPreProcessorTrait {
```
tvinagre, line 743: sweeeet, I've been waiting for it :)

PR #159, opened 2024-02-08
ENG-2775: Save new tokens to db on native extractor

tvinagre:


tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -684,14 +770,20 @@ mod test_serial_db {
...
-        let gw = NativePgGateway::new("test", Chain::Ethereum, pool.clone(), cached_gw);
-        (gw, err_rx, pool)
+        let gw = NativePgGateway::new(
+            "test",
+            Chain::Ethereum,
+            pool.clone(),
+            cached_gw,
+            get_mocked_token_pre_processor(),
+        );
+        (gw, err_rx, pool, chain_id)
```
louise-poole, line 780: This `chain_id` is only used in one test, no? Minor preference: I'd prefer we just query for this id in the one test that needs it than pass it to all tests on set up. But I think this is down to personal preference, not sure which is actually best practice.

PR #160, opened 2024-02-08
ENG-2776: Run tycho with ambient locally fixes

dianacarvalho1:
Fixes:
- Allow for tx_updates to be empty:
The previous implementation threw an error if it was empty. This was wrong because it can indeed be empty. There might be blocks that there were no txs related to ambient
- Change ValueIdTableCache.from_pool() to accept the name of the table
- Handle component_id as a hex string in ComponentBalance

### DIFF
--- a/tycho-indexer/src/extractor/evm/mod.rs
+++ b/tycho-indexer/src/extractor/evm/mod.rs
@@ -1547,7 +1547,6 @@ pub mod fixtures {
 mod test {
     use std::str::FromStr;
 
-    use actix_web::body::MessageBody;
     use prost::Message;
 
     use rstest::rstest;


tycho-indexer/src/storage/postgres/mod.rs
```diff
@@ -416,9 +434,12 @@ where
     }
 
     pub async fn new(pool: Pool<AsyncPgConnection>) -> Result<Arc<Self>, StorageError> {
-        let cache = ValueIdTableCache::<Chain>::from_pool(pool.clone()).await?;
+        println!("is it in new???");
```
SmartMonkey-git: And is it? :P

tycho-indexer/src/storage/postgres/mod.rs
```diff
@@ -174,22 +174,40 @@ where
     E: Eq + Hash + Clone + FromStr + std::fmt::Debug,
     <E as FromStr>::Err: std::fmt::Debug,
 {
-    pub async fn from_pool(pool: Pool<AsyncPgConnection>) -> Result<Self, StorageError> {
+    pub async fn from_pool(
```
SmartMonkey-git: Wouldn't it be cleaner to have a `FromPool` trait and to implement that for each table?
Not saying you should do that right now, but maybe a cooldown task?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -2411,7 +2433,10 @@ mod test {
         let protocol_component = result.unwrap();
 
         // Assert specific properties of the protocol component
-        assert_eq!(protocol_component.id, "component_id".to_string());
+        assert_eq!(
+            protocol_component.id,
+            "d417ff54652c09bd9f31f216b1a2e5d1e28c1dce1ba840c40d16f2b4d09b5902".to_string()
```
SmartMonkey-git, line 2429: Where did you get these hex-string? If they originate from ambient, then I would have expected them to end up as something readable, no?

tycho-indexer/src/storage/postgres/mod.rs
SmartMonkey-git: Something like:

```
trait FromPool {
    fn from_pool(pool: Pool<AsyncPgConnection>) -> Self;
}

impl FromPool for ValueIdTableCache<Chain> {
    fn from_pool(pool: Pool<AsyncPgConnection>) -> Self {
        unimplemented!()
    }
}

impl FromPool for ValueIdTableCache<ProtocolSystem> {
    fn from_pool(pool: Pool<AsyncPgConnection>) -> Self {
        unimplemented!()
    }
}
```

tycho-indexer/src/storage/postgres/mod.rs
dianacarvalho1: upsi

tycho-indexer/src/storage/postgres/mod.rs
dianacarvalho1: yes but I didn't want to do it because I'm lazy and I hate rust and I just wanted this to work. but oookkk

tycho-indexer/src/extractor/evm/mod.rs
dianacarvalho1: this is a real pool hash

dianacarvalho1: from substreams the component ID is the pool hash

PR #163, opened 2024-02-10
Run Uniswaps locally

zizou0x:
This PR contains all the modifications I had to do to run Tycho locally.

### DIFF
--- a/tycho-indexer/src/storage/postgres/cache.rs
+++ b/tycho-indexer/src/storage/postgres/cache.rs
@@ -104,6 +104,7 @@ pub struct DBCacheWriteExecutor {
     pool: Pool<AsyncPgConnection>,
     state_gateway: EVMStateGateway<AsyncPgConnection>,
     persisted_block: Option<evm::Block>,
+    latest_rpc_block_number: u64,
     /// The `pending_block` field denotes the most recent block that is pending processing in this
     /// cache context. It's important to note that this is distinct from the blockchain's
     /// concept of a pending block. Typically, this block corresponds to the latest block that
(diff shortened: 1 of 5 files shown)


tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -228,8 +222,10 @@ impl DBCacheWriteExecutor {
 
                     // Forward the result to the sender
                     let _ = new_db_tx.tx.send(res);
-                } else if new_db_tx.block.parent_hash == pending.hash {
-                    debug!("New block received {} !", new_db_tx.block.parent_hash);
+                } else if new_db_tx.block.parent_hash == pending.hash ||
+                    new_db_tx.block.number > pending.number
```
zizou0x: This was required during sync: if we sync one extractor (for example uniswapv2) and it reaches the write cache first, then all other extractors with a later cursor are going to fall in the `else` below and send an error to error_transmitter. 

Note: now I've changed this I'm not sure the `else` can still happen, maybe if we mix up blocks from two different chains.

zizou0x: Also it probably makes sense to simplify this with only `if new_db_tx.block.number > pending.number`

proto/tycho/evm/v1/common.proto
```diff
@@ -87,7 +87,7 @@ message ProtocolComponent {
 // A struct for following the changes of Total Value Locked (TVL) of a protocol component.
 // Note that if the ProtocolComponent contains multiple contracts, the TVL is tracked for the component as a whole.
 message BalanceChange {
-  // The address of the ERC20 token whose balance changed.
+  // The address of the ERC20 token whose balance changed. Note: This MUST be utf8 encoded.
```
kayibal: What? Why I think since this is a hash we should not utf-8 encode this. Maybe you meant to add this comment to `component_id`?

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -199,25 +201,21 @@ impl DBCacheWriteExecutor {
                         .run(|conn| {
                             async {
                                 for op in new_db_tx.operations {
-                                    if let WriteOp::UpsertBlock(_) = op {
-                                        // Ignore block upserts, we expect it to already be stored
-                                        continue;
```
kayibal, line 204: So we don't skip those anymore?

tycho-indexer/src/storage/postgres/cache.rs
kayibal: Will we still enter this branch if we get a revert, let; say we are at block 100:
`100, hash: 0x100, parent: 0x99`

Now we get a revert to 98:
`98, hash 0x98a, parent 0x97`

Our state at this point should be `pending.hash = 0x100`, `pending.number = 100`

kayibal: I currently don't remember if we enter this function on reverts...

tycho-types/src/dto.rs
```diff
@@ -573,6 +573,8 @@ pub struct ProtocolId {
 #[derive(Debug, Clone, PartialEq, Default, Deserialize, Serialize, ToSchema)]
 pub struct ResponseProtocolState {
     pub component_id: String,
+    /// Attributes of the component. If an attribute's value is a `bigint`,
+    /// it will be encoded as a little endian signed hex string.
```
kayibal, line 577: Mh ok we should probably unify this for all substreams. The benefit of this method is that if the int is <=128 bit we can convert it cheaper vs when it is 256 we pay probably some miniscule penalty for reversing the byte array.

We probably need a section for byte encoding values in the docs.

proto/tycho/evm/v1/common.proto
zizou0x: Yes I meant `component_id `, changed it

tycho-indexer/src/storage/postgres/cache.rs
zizou0x: No, we try to insert them anyway and ignore if it fails with DuplicateEntry. 

This is because we might add a protocol that started before what we currently have. For example currently in our database our blocks start from when Ambient was created, but we don't have previous blocks. So if we add uniswapV2, we can't ignore old blocks insertion

PR #166, opened 2024-02-14
[DO NOT MERGE] Ambient: Notes about multiple changes per block issue

tamaralipows:


substreams/ethereum-ambient/src/modules/3_map_changes.rs
```diff
@@ -252,6 +252,11 @@ fn map_changes(
                 .push(component.clone());
```
tamaralipows, line 252: The component changes here a line before also has the same issue.

substreams/ethereum-ambient/src/modules/3_map_changes.rs
```diff
@@ -252,6 +252,11 @@ fn map_changes(
...
         }
 
+        // TODO We add all of these ComponentBalance changes once per transaction per block.
+        // (look where this is in that for loop). No good. We have many duplicated entries because
+        // of this. We should probably also store in the balance_store the transaction hash
+        // where the change happened, and check then if this transaction corresponds.
+        // Or add them once per block? Not sure if that would work.
```
kayibal, line 259: No unfortunately that would not work well with the versioning :/

PR #168, opened 2024-02-15
ENG-2787: Bugfix: Components and balances were duplicated in ambient substreams

dianacarvalho1:
Fix:
- Add tx to BalanceDelta and ProtocolComponent protobuf messages
- Make BalanceDelta hold only one balance per object (not both base and quote token balances together)
- Use StoreDeltas in map_changes
- Add components only once and in the correct transaction change
- Add balances only once and in the correct transaction change

### DIFF
--- a/substreams/common/src/codegen.rs
+++ b/substreams/common/src/codegen.rs
@@ -5,7 +5,6 @@ use std::{
     fs,
     path::{Path, PathBuf},
     process::Command,
-    string::ToString,
 };
 use substreams_ethereum::Abigen;
 
--- a/tycho-client/src/deltas.rs
+++ b/tycho-client/src/deltas.rs
@@ -24,7 +24,6 @@ use futures03::{stream::SplitSink, SinkExt, StreamExt};
 use hyper::Uri;
 use std::{
     collections::{hash_map::Entry, HashMap},
-    string::ToString,
     sync::Arc,
 };
 use thiserror::Error;
@@ -36,9 +35,7 @@ use tokio::{
     },
     task::JoinHandle,
 };
-use tokio_tungstenite::{
-    connect_async, tungstenite, tungstenite::protocol::Message, MaybeTlsStream, WebSocketStream,
-};
+use tokio_tungstenite::{connect_async, tungstenite, MaybeTlsStream, WebSocketStream};
 use tracing::{debug, error, info, instrument, warn};
 use tycho_types::dto::{Command, Deltas, ExtractorIdentity, Response, WebSocketMessage};
 use uuid::Uuid;
@@ -115,7 +112,8 @@ pub struct WsDeltasClient {
     inner: Arc<Mutex<Option<Inner>>>,
 }
 
-type WebSocketSink = SplitSink<WebSocketStream<MaybeTlsStream<TcpStream>>, Message>;
+type WebSocketSink =
+    SplitSink<WebSocketStream<MaybeTlsStream<TcpStream>>, tungstenite::protocol::Message>;
 
 /// Subscription State
 ///
@@ -284,7 +282,7 @@ impl Inner {
     }
 
     /// Sends a message through the websocket.
-    async fn ws_send(&mut self, msg: Message) -> Result<(), DeltasError> {
+    async fn ws_send(&mut self, msg: tungstenite::protocol::Message) -> Result<(), DeltasError> {
         self.sink.send(msg).await.map_err(|e| {
             DeltasError::TransportError(format!("Failed to send message to websocket: {e}"))
         })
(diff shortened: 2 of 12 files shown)


substreams/ethereum-ambient/src/modules/2_store_pool_balances.rs
```diff
@@ -1,30 +1,19 @@
...
 
 #[substreams::handlers::store]
 pub fn store_pool_balances(changes: BlockPoolChanges, balance_store: StoreAddBigInt) {
-    for component in changes.protocol_components {
-        let pool_hash_hex = hex::encode(&component.id);
-        balance_store.add(0, format!("{}{}", pool_hash_hex, "base"), BigInt::from(0));
-        balance_store.add(0, format!("{}{}", pool_hash_hex, "quote"), BigInt::from(0));
-    }
     let mut deltas = changes.balance_deltas.clone();
     deltas.sort_by_key(|delta| delta.ordinal);
```
kayibal: I would do that in the module before, also I think we should sort by `(ordinal, token_type)`, so base fore quote just to make things very deterministic.

substreams/ethereum-ambient/src/modules/3_map_changes.rs
```diff
@@ -316,6 +272,79 @@ fn map_changes(
...
+    for (tx_hash, components) in grouped_components {
+        if let Some(tx_change) = block_changes
+            .changes
+            .iter_mut()
+            .find(|tx_change| {
+                tx_change
+                    .tx
+                    .as_ref()
+                    .map_or(false, |tx| tx.hash == tx_hash)
+            })
```
kayibal, line 298: Quadratic complexity here. Not great please add at least a comment to come back to later.

substreams/ethereum-ambient/src/modules/3_map_changes.rs
```diff
@@ -316,6 +272,79 @@ fn map_changes(
...
+    for (tx_hash, grouped_balance_changes) in balance_changes {
+        if let Some(tx_change) = block_changes
+            .changes
+            .iter_mut()
+            .find(|tx_change| {
+                tx_change
+                    .tx
+                    .as_ref()
+                    .map_or(false, |tx| tx.hash == tx_hash)
+            })
```
kayibal, line 344: Another quadratic complexity.

substreams/ethereum-ambient/src/modules/3_map_changes.rs
```diff
@@ -316,6 +272,79 @@ fn map_changes(
...
+        .deltas
+        .into_iter()
+        .zip(block_pool_changes.balance_deltas)
+        .for_each(|(store_delta, balance_delta)| {
+            let pool_hash_hex = hex::encode(balance_delta.pool_hash);
+            let pool = match pool_store.get_last(pool_hash_hex.clone()) {
+                Some(pool) => pool,
+                None => panic!("Pool not found in store for given hash: {}", pool_hash_hex),
+            };
+            let token_index = if balance_delta.token_type == "quote" { 1 } else { 0 };
```
kayibal: maybe we can use store_delta.key to get either quote or base? I think that would be safer to rely on order matching even if the ordinal is the same. E.g. if both quote and balance changed in the same transaction.

kayibal: You can use: `substream::key::segment_at(&store_delta.key, 1);`

kayibal: I think if the order is undeterministic, we may still match with the wrong tx here, no?

dianacarvalho1: yes 😕 playgrounds balancer implementation also assumes this though 😕 we could just add here a panic in case it happens so we can sleep better at night

substreams/ethereum-ambient/src/modules/1_map_pool_changes.rs
```diff
@@ -95,15 +101,24 @@ fn map_pool_changes(block: eth::v2::Block) -> Result<BlockPoolChanges, substream
...
                 ordinal: call.index as u64,
+                tx: Some(tx.clone()),
             };
-            balance_deltas.push(balance_delta.clone());
+            let quote_balance_delta = BalanceDelta {
+                pool_hash: Vec::from(pool_hash),
+                token_type: "quote".to_string(),
+                token_delta: from_u256_to_vec(quote_flow),
+                ordinal: call.index as u64,
+                tx: Some(tx.clone()),
```
tamaralipows, line 116: Don't really understand why this is `Some`, it doesn't look optional in the protobuf message? But I'm not super familiar with that part

PR #169, opened 2024-02-15
Add logs and parametrize usv2 factory

zizou0x:


substreams/ethereum-uniswap-v2/src/modules/1_map_pool_created.rs
```diff
@@ -67,9 +74,7 @@ fn get_pools(block: &eth::Block, new_pools: &mut Vec<TransactionEntityChanges>)
     let mut eh = EventHandler::new(block);
 
     // TODO: Parametrize Factory Address
```
tvinagre: useless comment now

PR #170, opened 2024-02-16
fix: skip first message from substream on restart

zizou0x:
We notice that on restart, Tycho will send the last processed block again which could break things because of conflicts. This PR fixes this by ignoring the first msg from substream on restart.

### DIFF
--- a/tycho-indexer/src/substreams/stream.rs
+++ b/tycho-indexer/src/substreams/stream.rs
@@ -68,7 +68,7 @@ fn stream_blocks(
     // For example, if before the restart we received block 100, 101 and crashed,
     // block 100 and its cursor was sent to the db. After the restart substreams will send
     // us block 100 again, and we want to skip it to avoid conflicts.
-    let mut skip_first_msg = cursor != Some("".to_string());
+    let mut skip_first_msg = cursor.as_deref() != Some("");
     let mut latest_cursor = cursor.unwrap_or_default();
     let mut retry_count = 0;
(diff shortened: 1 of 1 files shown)


tycho-indexer/src/substreams/stream.rs
```diff
@@ -62,6 +62,13 @@ fn stream_blocks(
     start_block_num: i64,
     stop_block_num: u64,
 ) -> impl Stream<Item = Result<BlockResponse, Error>> {
+    // When restarting a Tycho stream, we want to skip the first message we receive from
+    // substreams. This is because the first message is the last message we processed before the
+    // restart.
+    // For example, if before the restart we received block 100, 101 and crashed,
+    // block 100 and its cursor was sent to the db. After the restart substreams will send
+    // us block 100 again, and we want to skip it to avoid conflicts.
+    let mut skip_first_msg = cursor != Some("".to_string());
```
louise-poole: Interesting thing I just learnt - this:
```
let mut skip_first_msg = cursor.as_deref() != Some("");
```
is considered more efficient and better coding practice as it doesn't allocate a new String just for a comparison. Looks like this line won't be executed often so it doesn't make much of a difference here. Anyway, just wanted someone to share my new findings with 😂

zizou0x: https://github.com/propeller-heads/tycho-indexer/pull/170/commits/a3b97269f3f4b87d4a9f7741595abf67fffe80d2

PR #175, opened 2024-02-17
Improve Sync UX & Performance

kayibal:
- Add column indices for faster versioning queries, aimed to speed up inserts.
- Batch collect operations in CachedGateway while the extractor is syncing.
- Add a progress metric to each extractor that reports its sync status.

### DIFF
--- a/tycho-indexer/src/extractor/evm/native.rs
+++ b/tycho-indexer/src/extractor/evm/native.rs
@@ -601,11 +601,8 @@ where
 
 #[cfg(test)]
 mod test {
-    use std::collections::HashMap;
 
-    use tycho_types::Bytes;
-
-    use crate::{extractor::evm, pb::sf::substreams::v1::BlockRef};
+    use crate::pb::sf::substreams::v1::BlockRef;
 
     use super::*;
 
@@ -760,9 +757,7 @@ mod test_serial_db {
     //!
     //! Note that it is ok to use higher level db methods here as there is a layer of abstraction
     //! between this component and the actual db interactions
-    use std::collections::{HashMap, HashSet};
 
-    use ethers::prelude::H160;
     use mpsc::channel;
     use tokio::sync::{
         mpsc,
@@ -770,7 +765,6 @@ mod test_serial_db {
     };
 
     use test_serial_db::evm::ProtocolChangesWithTx;
-    use tycho_types::Bytes;
 
     use crate::{
         extractor::evm::{
--- a/tycho-indexer/src/extractor/evm/token_pre_processor.rs
+++ b/tycho-indexer/src/extractor/evm/token_pre_processor.rs
@@ -70,10 +70,6 @@ impl TokenPreProcessorTrait for TokenPreProcessor {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use ethers::{
-        providers::{Http, Provider},
-        types::H160,
-    };
     use std::{env, str::FromStr};
 
     #[tokio::test]
--- a/tycho-indexer/src/storage/postgres/cache.rs
+++ b/tycho-indexer/src/storage/postgres/cache.rs
@@ -825,13 +825,10 @@ mod test_serial_db {
         prelude::H256,
         types::{H160, U256},
     };
-    use std::{collections::HashMap, str::FromStr, sync::Arc};
+    use std::{collections::HashMap, str::FromStr};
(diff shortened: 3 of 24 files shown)


.github/workflows/docker.yaml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - main
       - patch/*
+      - ah/sync-performance
```
tvinagre: Let's remember to remove later

substreams/ethereum-uniswap-v2/src/modules/3_map_pool_events.rs
```diff
@@ -159,15 +159,34 @@ fn merge_block(
     tx_changes: &mut HashMap<Vec<u8>, PartialChanges>,
     block_entity_changes: &mut BlockEntityChanges,
 ) {
+    let mut tx_entity_changes_map = HashMap::new();
+
+    // Add created pools to the tx_changes_map
+    for change in block_entity_changes
```
tvinagre, line 165: I don't understand the benefit of this change, since `block_entity_changes.changes` already has the create transactions - why not just modify it?

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -33,17 +34,40 @@ use super::token_pre_processor::{TokenPreProcessor, TokenPreProcessorTrait};
...
+}
+
+impl ChainState {
+    pub fn new(start: NaiveDateTime, block_number: u64) -> Self {
+        Self { start, block_number }
+    }
+    pub async fn current_block(&self) -> u64 {
+        let now = chrono::Local::now().naive_utc();
+        let diff = now.signed_duration_since(self.start);
+        let blocks_passed = (diff.num_seconds() / 12) as u64;
```
tvinagre, line 66: Hacky aff

tvinagre: Maybe add a note that this needs to be improved if we want to move to other EVM chains?

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -54,6 +78,49 @@ impl<DB> NativeContractExtractor<DB> {
         let mut state = self.inner.lock().await;
         state.last_processed_block = Some(block);
     }
+
+    async fn report_progress(&self, block: Block) {
```
tvinagre, line 82: Nice

tycho-indexer/src/extractor/evm/token_pre_processor.rs
```diff
@@ -54,7 +55,12 @@ impl TokenPreProcessorTrait for TokenPreProcessor {
             };
             tokens_info.push(ERC20Token {
                 address,
-                symbol,
+                symbol: symbol
+                    .replace('\0', "")
+                    .graphemes(true)
+                    .take(230) // Limit to 230 graphemes because we need to prepend this by "{chain_name}_" and
```
tvinagre: Also would be nice to have a TODO as 230 is calculated considering ethereum

tycho-indexer/src/storage/postgres/cache.rs
```diff
@@ -448,6 +554,9 @@ type DeltasCache = LruCache<
 type OpenTx = (DBTransaction, oneshot::Receiver<Result<(), StorageError>>);
 
 pub struct CachedGateway {
+    // Can we batch multiple block in here without breaking things?
```
tvinagre, line 557: If we maintain the block ordering upon insertion, yeah. We'd have to change a little the commit rules, bc rn it flushes if we find a newer block.

Shouldn't be a problem bc we wouldn't have rollbacks / forks, so I think it'd be a nice improvment

.github/workflows/docker.yaml
kayibal: Will do before merging

substreams/ethereum-uniswap-v2/src/modules/3_map_pool_events.rs
kayibal: This is to aggregate components by their transactions, I think transactions were duplicated previously so we would only emit the last change component.

tycho-indexer/src/extractor/evm/native.rs
kayibal: I'll start a todo list for other chains now

PR #183, opened 2024-02-26
Increase DeltasClient Buffers.

kayibal:
Before these changes it was possible for the `handle_message` method to deadlock the websocket heartbeat messages. This could happen when the subscriber channel was full and the `handle_message` method would block until subscriber messages had been consumed. Any messages arriving to a full channel are now dropped, this is not ideal but prevents the application from deadlocks.

### DIFF
--- a/tycho-client/src/deltas.rs
+++ b/tycho-client/src/deltas.rs
@@ -387,7 +387,7 @@ impl WsDeltasClient {
                         Err(DeltasError::BufferFull) => {
                             error!(?subscription_id, "Buffer full, message dropped!");
                         }
-                        Err(e) => {
+                        Err(_) => {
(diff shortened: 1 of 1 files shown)


tycho-client/src/deltas.rs
```diff
@@ -317,6 +336,8 @@ impl WsDeltasClient {
         Ok(Self {
             uri,
             inner: Arc::new(Mutex::new(None)),
+            ws_buffer_size: 128,
+            subscription_buffer_size: 128,
```
louise-poole, line 340: This is quite a large increase from what we had before - are we able to support such a large buffer? What happens when the messages get quite large (as I assume they easily will become if we use low tvl to filter components)

PR #185, opened 2024-02-27
Client side pagination

kayibal:
This PR aims to speed up protocol state retrieval as this is currently pretty slow.

Ideally we would also remove the join on transaction then the query plan is a pure index lookup without any joins but that would require changing a few models as well. So we will only do that if this is still not fast enough.

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -36,8 +36,10 @@ jobs:
         uses: aws-actions/amazon-ecr-login@v1
 
       - name: Get next version
-        uses: reecetech/version-increment@2023.10.1
         id: version
+        uses: reecetech/version-increment@2023.10.1
+        with:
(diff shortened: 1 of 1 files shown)


tycho-client/src/feed/synchronizer.rs
```diff
@@ -249,10 +249,12 @@ where
...
                 .rpc_client
-                .get_protocol_states(
+                .get_protocol_states_paginated(
                     self.extractor_id.chain,
-                    &Default::default(),
-                    &ProtocolStateRequestBody::id_filtered(component_ids),
+                    &component_ids,
+                    &version,
+                    50,
+                    4,
```
louise-poole, line 257: How much did you play around with these pagination params? Have these been optimised?

kayibal: No I haven't player around with these at all.

PR #188, opened 2024-03-08
Refactor storage

kayibal:
This PR factors out a storage crate. Original motivation was that diesel macros are quite heavy on the CPU and we don't expect too much more changes there. 

In this PR the DB interface was also strongly simplified. It does not use any generics types nor associated types anymore and instead consumes cross-chain agnostic models. Chain specific attributes of our models are simply handled as bytes. The chain specific validation should be handled before Tycho ingestion on the substreams layer. It is therefore not necessary to use chain specific types. That said, the extractor subcrate does still use the evm specific models, these are converted using the From trait into the general purpose structs from `tycho_core::models`.

This PR also adds a mocked gateway that can be used for unittests. This included updating rpc tests to use this mocked gateway (unfortunately test quality decreased a bit but we were anyway testing mostly the DB within the rpc tests).

I also already removed the substream packages from the workspace as a preparation for moving the into propeller-protocol-lib.

### DIFF
--- a/tycho-client/src/feed/mod.rs
+++ b/tycho-client/src/feed/mod.rs
@@ -3,6 +3,7 @@ use std::collections::HashMap;
 use chrono::{Local, NaiveDateTime};
 use futures03::{future::join_all, stream::FuturesUnordered, StreamExt};
 
+use serde::Serialize;
 use tokio::{
     select,
     sync::mpsc::{self, Receiver},
@@ -24,7 +25,7 @@ mod block_history;
 pub mod component_tracker;
 pub mod synchronizer;
 
-#[derive(Debug, Clone, PartialEq, Default)]
+#[derive(Debug, Clone, PartialEq, Default, Serialize)]
 pub struct Header {
     pub hash: Bytes,
     pub number: u64,
@@ -99,7 +100,8 @@ pub struct BlockSynchronizer<S> {
     max_wait: std::time::Duration,
 }
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Serialize)]
+#[serde(tag = "status", rename_all = "lowercase")]
 pub enum SynchronizerState {
     Started,
     Ready(Header),
@@ -299,7 +301,7 @@ impl SynchronizerStream {
     }
 }
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Serialize)]
 pub struct FeedMessage {
     pub state_msgs: HashMap<String, StateSyncMessage>,
     pub sync_states: HashMap<String, SynchronizerState>,
--- a/tycho-client/src/feed/synchronizer.rs
+++ b/tycho-client/src/feed/synchronizer.rs
@@ -1,4 +1,5 @@
 use async_trait::async_trait;
+use serde::Serialize;
 use std::{collections::HashMap, sync::Arc, time::Duration};
 
 use tokio::{
@@ -45,25 +46,25 @@ struct SharedState {
     last_synced_block: Option<Header>,
 }
 
-#[derive(Clone, PartialEq, Debug)]
+#[derive(Clone, PartialEq, Debug, Serialize)]
 pub struct VMSnapshot {
     pub state: HashMap<Bytes, ResponseAccount>,
     pub component: ProtocolComponent,
 }
 
-#[derive(Clone, PartialEq, Debug)]
+#[derive(Clone, PartialEq, Debug, Serialize)]
 pub struct NativeSnapshot {
     pub state: ResponseProtocolState,
     pub component: ProtocolComponent,
 }
 
-#[derive(Clone, PartialEq, Debug)]
+#[derive(Clone, PartialEq, Debug, Serialize)]
 pub enum Snapshot {
     VMSnapshot(VMSnapshot),
     NativeSnapshot(NativeSnapshot),
 }
 
-#[derive(Clone, PartialEq, Debug, Default)]
+#[derive(Clone, PartialEq, Debug, Default, Serialize)]
 pub struct StateSyncMessage {
     /// The block number for this update.
     pub header: Header,


tycho-indexer/src/services/rpc.rs
```diff
@@ -849,180 +759,52 @@ mod tests {
...
-
-        let expected = evm::Account {
-            chain: Chain::Ethereum,
-            address: "0x6b175474e89094c44da98b954eedeac495271d0f"
+        let expected = Contract::new(
+            Chain::Ethereum,
+            "0x6b175474e89094c44da98b954eedeac495271d0f"
                 .parse()
                 .unwrap(),
-            title: "account0".to_owned(),
```
tvinagre, line 992: Is it bad practice to have the attribute name when calling the `new` func?

kayibal: No but it is a syntax error :D

tvinagre: lol I didn't realize it - tnx

tycho-indexer/src/models/blockchain.rs
```diff
@@ -0,0 +1,173 @@
...
+    pub decimals: u32,
+    pub tax: u64,
+    pub gas: Vec<Option<u64>>,
+    pub chain: Chain,
+    /// Quality is between 0-100, where:
+    ///  - 100: Normal token
+    ///  - 75: Rebase token
+    ///  - 50: Fee token
+    ///  - 0: Scam token that we shouldn't use
+    pub quality: u32,
```
louise-poole: How is this measured?

tycho-indexer/src/models/blockchain.rs
```diff
@@ -0,0 +1,173 @@
...
+
+pub struct TransactionDeltaGroup<T> {
+    changes: T,
+    protocol_component: HashMap<String, ProtocolComponent>,
+    component_balances: HashMap<String, ComponentBalance>,
+    component_tvl: HashMap<String, f64>,
+    tx: Transaction,
+}
+
+pub struct BlockAggregatedDeltas<T> {
```
louise-poole: I really like all of the renaming you did 👍

tycho-indexer/src/testing.rs
```diff
@@ -0,0 +1,514 @@
...
+        /// # Parameters
+        /// - `chain` The chain of the component
+        /// - `system` Allows to optionally filter by system.
+        /// - `id` Allows to optionally filter by id.
+        ///
+        /// # Returns
+        /// Ok, if found else Err
+
+        #[allow(clippy::type_complexity, clippy::type_repetition_in_bounds)]
+        fn get_protocol_components<'life0, 'life1, 'life2, 'life3, 'async_trait>(
```
louise-poole: Just to nit-pick - I'd prefer if there wasn't a space between the doc string and the function it's describing

tycho-indexer/src/testing.rs
```diff
@@ -0,0 +1,514 @@
...
+        /// Stores new found ProtocolTypes.
+        ///
+        /// # Parameters
+        /// - `new`  The new protocol types.
+        ///
+        /// # Returns
+        /// Ok if stored successfully.
+
+        #[allow(clippy::type_complexity, clippy::type_repetition_in_bounds)]
+        fn add_protocol_types<'life0, 'life1, 'async_trait>(
```
louise-poole: Same here. I'd also prefer a space above so it's clear it's not linked to the code above too.

louise-poole: (repeated for all doc strings in the file)

tycho-core/src/models/contract.rs
```diff
@@ -0,0 +1,117 @@
...
+    ) -> Self {
+        Self {
+            chain: *chain,
+            address: address.clone(),
+            change,
+            slots: slots
+                .map(|storage| {
+                    storage
+                        .iter()
+                        // TODO: unwrap or default converts to 0x0 instead of a 32 byte 0 string
```
louise-poole: I see a couple TODOs still here - do they still need addressing?

tycho-core/src/models/mod.rs
```diff
@@ -0,0 +1,199 @@
...
+        ExtractionState {
+            name,
+            chain,
+            attributes: attributes.unwrap_or_default(),
+            cursor: cursor.to_vec(),
+        }
+    }
+}
+
+// TODO: replace with types from dto on extractor
```
louise-poole, line 126: And here?

tycho-core/src/models/protocol.rs
```diff
@@ -0,0 +1,108 @@
...
+pub struct ProtocolComponent {
+    pub id: String,
+    pub protocol_system: String,
+    pub protocol_type_name: String,
+    pub chain: Chain,
+    pub tokens: Vec<Bytes>,
+    pub contract_addresses: Vec<Bytes>,
+    pub static_attributes: HashMap<String, Bytes>,
+    // TODO: decide for a module. dto or storage. -> probably promote change type eventually
+    pub change: ChangeType,
```
louise-poole: What do you mean by promote change type?

tycho-core/src/models/protocol.rs
```diff
@@ -0,0 +1,108 @@
...
+        Self { component_id: component_id.to_string(), attributes, modify_tx }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct ProtocolComponentStateDelta {
+    pub component_id: String,
+    pub updated_attributes: HashMap<String, Bytes>,
+    // TODO: rename back to deleted_attributes
+    pub removed_attributes: HashSet<String>,
```
louise-poole: Why would this need to be changed back?

tycho-indexer/src/extractor/evm/convert.rs
```diff
@@ -0,0 +1,207 @@
...
+        }
+    }
+}
+
+// TODO: remove this, only here so the old revert method compiles.
+impl From<ComponentBalance> for evm::ComponentBalance {
+    fn from(_value: ComponentBalance) -> Self {
+        todo!()
+    }
+}
```
louise-poole: Looks like this can be removed?

tycho-core/Cargo.toml
```diff
@@ -1,11 +1,12 @@
 [package]
-name = "tycho-types"
+name = "tycho-core"
 version = "0.3.0"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+# TODO: remove or keep only for testing
```
zizou0x: It could be a dev-dependency then I think

tycho-core/src/dto.rs
```diff
@@ -33,6 +38,33 @@ pub enum Chain {
     ZkSync,
 }
 
+impl From<models::contract::Contract> for ResponseAccount {
+    fn from(value: Contract) -> Self {
```
zizou0x, line 42: Is it on purpose/a good practice that on the first line full path is specified and on the second line only the final type?

Edit: ok I see that sometimes it's done like this and sometimes not, I think we can use full path everywhere and remove some imports. Not a big comment but just to make this cleaner

tycho-core/src/models/contract.rs
```diff
@@ -0,0 +1,117 @@
...
+    pub code: Bytes,
+    pub code_hash: Bytes,
+    pub balance_modify_tx: Bytes,
+    pub code_modify_tx: Bytes,
+    pub creation_tx: Option<Bytes>,
+}
+
+impl Contract {
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
```
zizou0x, line 23: Seeing this reminds me that I heard about the [derive_more](https://crates.io/crates/derive_more) crate. I didn't look at it in detail but it provides a derive for this, which looks cool and much cleaner for simple cases.

```
#[derive(Constructor)]
struct Point {
  x: f32,
  y: f32,
}

fn main() {
  let p = Point::new(3, 12);
}
```

tycho-core/src/models/mod.rs
```diff
@@ -0,0 +1,199 @@
+#![allow(dead_code)]
```
zizou0x: Is this needed here? Can't we move it to be targeted to only some types?

tycho-core/src/models/protocol.rs
zizou0x: I didn't get this too

tycho-core/src/storage.rs
```diff
@@ -0,0 +1,577 @@
...
+/// blocks of their respective chain if applicable. This means while indexing we
+/// need to keep a lightweight and cross chain compatible representation of
+/// blocks and transactions in storage.
+///
+/// It's defined generically over two associated types:
+///
+/// * `Block`: represents a block in the blockchain.
+/// * `Transaction`: represents a transaction within a block.
+#[async_trait]
+pub trait ChainGateway {
```
zizou0x, line 80: Nice, you removed the generics

tycho-indexer/src/testing.rs
```diff
@@ -0,0 +1,514 @@
...
+        Address, Chain, ContractId, ExtractionState, ProtocolType, TxHash,
+    },
+    storage::{
+        BlockIdentifier, BlockOrTimestamp, ChainGateway, ContractStateGateway,
+        ExtractionStateGateway, Gateway, ProtocolGateway, StorageError, Version,
+    },
+    Bytes,
+};
+
+mock! {
```
zizou0x, line 26: What does this do?

tycho-storage/src/postgres/orm.rs
```diff
@@ -539,6 +584,31 @@ pub struct NewProtocolComponent {
     pub attributes: Option<serde_json::Value>,
 }
 
+impl NewProtocolComponent {
+    #[allow(unused_variables, clippy::too_many_arguments)]
```
zizou0x: Looks like all variables are used, we can probably remove `allow(unused_variables)`

tycho-indexer/src/models/blockchain.rs
kayibal: This is more of a placeholder for a potential enum later, the values you can see above. At the moment simply all tokens for which we can get decimals are 100, anything that makes problems when reading decimals is set to 0.

tycho-indexer/src/testing.rs
kayibal: Yeah these comments should actually not be there ^^

tycho-core/src/models/mod.rs
kayibal: Didn't want to do this one yet since I was afraid it the changes might spread too much. I will takle it once I move all the core types into the extractor.

tycho-core/src/models/protocol.rs
kayibal: Hmm for consistency reasons - e.g. dto models have this as deleted attributes - I don't even remember why I changed it. But will change it back.

tycho-core/src/dto.rs
kayibal: I did initially work with the full path everywhere because I didn't want to get confused since some models have the same name as before. So this can be removed it is actually best practice to not do this and just use the shortest path possible that is not ambiguous.

tycho-indexer/src/testing.rs
kayibal: This allows me to define a MockGateway that implements several traits. It does basically the same as automock does.

tycho-core/src/models/protocol.rs
kayibal: This type used to be under storage, I promoted it to core already so this is an outdated comment.

tycho-core/src/models/contract.rs
kayibal: If it is a proc macro I'd probably prefer to type it out^^

tycho-core/Cargo.toml
kayibal: We use it in hex_bytes to derive some traits. We'll have to get rid of this crate in indexer first, else it will be a pain to remove from here.

tycho-core/src/dto.rs
kayibal: I tried removing it but we still have some naming conflicts, so I'll keep it like this until we get rid of etheres and the evm specific models.

PR #189, opened 2024-03-08
Add release steps to Github CI

tvinagre:


.github/workflows/release.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+jobs:
+  create-release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: taiki-e/create-gh-release-action@v1
+        with:
+          draft: true
+          token: ${{ secrets.GITHUB_TOKEN }}
```
kayibal, line 23: So this creates a release if we push a tag right? And what if the release is already present, e.g. if I create a release via the UI?

Dockerfile
```diff
@@ -29,6 +29,7 @@ RUN rustup target add wasm32-unknown-unknown && ./stable-build.sh
 FROM debian:bookworm
 WORKDIR /opt/tycho-indexer
 COPY --from=build /build/target/release/tycho-indexer ./tycho-indexer
+COPY --from=build /build/target/release/tycho-client ./tycho-client
```
kayibal, line 32: Do we still need this then?

stable-build.sh
```diff
@@ -19,3 +20,6 @@ done
 
 echo "Building tycho: $TYCHO_INDEXER"
 cargo build --package $TYCHO_INDEXER --release
+
+echo "Building tycho-client: $TYCHO_CLIENT}"
+cargo build --package $TYCHO_CLIENT --release
```
kayibal, line 25: And this?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,39 @@
+# When a release is created, this workflow will build and upload the tycho-client binary to the release.
+name: Release
+
+on:
+  release:
+    types: [ created ]
+  push:
+    tags:
+      - v[0-9]+.*
```
kayibal: I our tags don't have the `v` prefix

louise-poole: Yeah - I think only frontend has this...

Dockerfile
tvinagre: Ah yeah I'd keep it - you can build and copy the binaries to linux-arm this way locally very easy

stable-build.sh
tvinagre: Same for linux-arm

PR #191, opened 2024-03-13
Add Balancer substreams

zizou0x:
This PR adds balancer extraction logic. Along the way I also completely refactored how we create and configure our extractors. It's now handled by a yaml config file and some general functions. 

Spkg are downloaded (if not already there) at start from S3 (we still need to add env var for this in our pods). I've manually uploaded all our current version of substreams modules to S3. 

We will later need to move the remaining substreams folder to the lib repo to allow for automatic push to s3 on changes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,6 +30,7 @@ FROM debian:bookworm
 WORKDIR /opt/tycho-indexer
 COPY --from=build /build/target/release/tycho-indexer ./tycho-indexer
 COPY --from=build /build/target/release/tycho-client ./tycho-client
+COPY --from=build /build/tycho-indexer/extractors.yaml ./extractors.yaml
 COPY --from=build /build/substreams/ethereum-ambient/substreams-ethereum-ambient-v0.4.0.spkg ./substreams/substreams-ethereum-ambient-v0.4.0.spkg
 COPY --from=build /build/substreams/ethereum-uniswap-v2/substreams-ethereum-uniswap-v2-v0.1.0.spkg ./substreams/substreams-ethereum-uniswap-v2-v0.1.0.spkg
 COPY --from=build /build/substreams/ethereum-uniswap-v3/substreams-ethereum-uniswap-v3-v0.1.0.spkg ./substreams/substreams-ethereum-uniswap-v3-v0.1.0.spkg


substreams/ethereum-ambient/src/contracts/main.rs
```diff
@@ -151,8 +151,9 @@ pub fn decode_pool_init(
                     protocol_type: Some(crate::pb::tycho::evm::v1::ProtocolType {
                         name: "ambient_pool".to_string(),
                         attribute_schema: vec![],
-                        financial_type: 0,
-                        implementation_type: 0,
+                        financial_type: crate::pb::tycho::evm::v1::FinancialType::Swap.into(),
+                        implementation_type: crate::pb::tycho::evm::v1::ImplementationType::Vm
+                            .into(),
```
kayibal: Oh no :/

tycho-indexer/src/main.rs
```diff
@@ -164,185 +162,99 @@ async fn main() -> Result<(), ExtractionError> {
...
-    };
-    builder.run().await
+async fn load_extractors_config<P: AsRef<Path>>(
+    path: P,
+) -> Result<ExtractorConfigs, Box<dyn std::error::Error>> {
+    let mut file = File::open(path)?;
+    let mut contents = String::new();
+    file.read_to_string(&mut contents)?;
+    let config: ExtractorConfigs = serde_yaml::from_str(&contents)?;
+    Ok(config)
```
kayibal: Could be a method on `ExtractorConfigs`?
```
impl ExtractorConfigs {
     fn from_yaml<P>(path: P) -> anyhow::Result<Self>{
        ....
     }
}
```

tycho-indexer/src/main.rs
```diff
@@ -164,185 +162,99 @@ async fn main() -> Result<(), ExtractionError> {
...
+        download_file_from_s3("repo.propellerheads", &spkg_path, Path::new(&spkg_path))
+            .await
+            .map_err(|e| {
+                ExtractionError::Setup(format!("Failed to download {} from s3. {}", &spkg_path, e))
+            })?;
+
+        let builder = ExtractorRunnerBuilder::new(&spkg_path, extractor)
+            .start_block(start_block)
+            .module_name(&module_name)
+            .only_final_blocks();
```
kayibal: I wonder if we can merge this with the `ExtractorBuilder`?

I think they can be merged, they actually look very related, atm the ExtractorRunnerBuilder would need the Extractor passed to it. I think we can simply make the extractor attribute optional on the ExtractorRunnerBuilder, and add a method like:

```rust
fn with_extractor(mut self, config: ExtractorConfig) {
    let self.extractor = match  ....
    
    self
}
```

This way you go directly from ExtractorConfig to running Extractor.

tycho-indexer/src/extractor/builder.rs
```diff
@@ -0,0 +1,130 @@
...
+        cached_gw: CachedGateway,
+        chain_state: ChainState,
+    ) -> Self {
+        Self { token_pre_processor, cached_gw, chain_state }
+    }
+
+    pub async fn build(
+        &self,
+        config: ExtractorConfig,
+    ) -> Result<Arc<dyn Extractor>, ExtractionError> {
```
kayibal: This `&self`, is a bit a violation of the Builder pattern. The builder should consume itself to yield the type that it is building. In this case it's not necessary because all types a clone. But I think I would prefer sth like:

```
let (extractor_handle, jh) = ExtractorBuilder::new(config)
    .set_gateway(&gateway)
    .set_token_processor(&token_processor)
    .set_chain_state(&chain_state)
    .run()
    .await?;
```

So basically this class merged with extractor ExtractorRunner.  This way starting and initialising an extractor is all handled in a single place.

tycho-indexer/src/main.rs
```diff
@@ -164,185 +162,99 @@ async fn main() -> Result<(), ExtractionError> {
...
-    builder.run().await
+) -> Result<Vec<(JoinHandle<Result<(), ExtractionError>>, ExtractorHandle)>, ExtractionError> {
+    let mut extractor_handles = Vec::new();
+    let extractor_builder =
+        ExtractorBuilder::new(token_pre_processor, cached_gw.clone(), chain_state);
+
+    for (_, extractor_config) in config.extractors.into_iter() {
+        let spkg_path = extractor_config.spkg();
+        let start_block = extractor_config.start_block();
+        let module_name = extractor_config.module_name();
```
kayibal: It feels weird that we have to do this....

tycho-indexer/src/main.rs
```diff
@@ -164,185 +162,99 @@ async fn main() -> Result<(), ExtractionError> {
...
+        let extractor = extractor_builder
+            .build(extractor_config)
+            .await?;
+
+        // Pull spkg from s3 and copy it at `spkg_path`
+        download_file_from_s3("repo.propellerheads", &spkg_path, Path::new(&spkg_path))
+            .await
+            .map_err(|e| {
+                ExtractionError::Setup(format!("Failed to download {} from s3. {}", &spkg_path, e))
+            })?;
```
kayibal: I think this should be part of starting the extractor, so should be part of `ExtractorRunnerBuilder::run`. 

Would also be nice if we could skip this in case the file already exists btw. This way we can add a persistent directory, so that in case of a restart it doesn't take forever to download all the spkgs, I am expecting to have quite a few of them in the future.

tycho-indexer/src/main.rs
kayibal: Also no need for this to be async imo?

tycho-indexer/src/main.rs
```diff
@@ -125,35 +122,36 @@ async fn main() -> Result<(), ExtractionError> {
...
 
-    let (uniswap_v2_task, uniswap_v2_handle) =
-        start_uniswap_v2_extractor(&args, chain_state, cached_gw.clone(), token_processor.clone())
-            .await?;
-    extractor_handles.push(uniswap_v2_handle.clone());
-    info!("Extractor {} started!", uniswap_v2_handle.get_id());
+    for (task, extractor_handle) in extractor_map {
+        extractor_handles.push(extractor_handle);
+        tasks.push(task);
+    }
```
kayibal:
```suggestion
    let (tasks, extractor_handles): (Vec<_>, Vec<_>) =
        build_all_extractors(config, chain_state, cached_gw.clone(), token_processor.clone())
            .await
            .map_err(|e| ExtractionError::Setup(format!("Failed to create extractors: {}", e)))?
            .into_iter()
            .unzip();
```

substreams/ethereum-ambient/src/contracts/main.rs
zizou0x: Do you prefer 0? I feel like it's shadier with just the number. Or is it because we might need to go through the whole sync thing again?

tycho-indexer/src/extractor/builder.rs
zizou0x: Yeah I would also like to merge these two, I wanted to have a look into how doable this would be. I'll have a look

tycho-indexer/src/main.rs
zizou0x: This is very cool!

tycho-indexer/src/main.rs
zizou0x: Yes I also don't really like it, also the getter functions that clone things is a weird design. I'll improve this

substreams/ethereum-ambient/src/contracts/main.rs
kayibal: I would prefer if we keep that change separate for now, maybe you can add this to the ambient PR that is currently open?

tycho-indexer/src/extractor/runner.rs
```diff
@@ -281,31 +416,58 @@ impl ExtractorRunnerBuilder {
...
-        );
+        let runner =
+            ExtractorRunner::new(extractor, stream, Arc::new(Mutex::new(HashMap::new())), ctrl_rx);
 
         let handle = runner.run();
         Ok((handle, ExtractorHandle::new(id, ctrl_tx)))
     }
 }
 
+pub async fn download_file_from_s3(
```
kayibal: does it need to be public?

tycho-indexer/src/extractor/runner.rs
```diff
@@ -268,10 +297,116 @@ impl ExtractorRunnerBuilder {
         self
     }
 
+    pub async fn ensure_spkg(&self) -> Result<(), ExtractionError> {
```
kayibal: Same here does this need to be public?

tycho-indexer/src/extractor/runner.rs
```diff
@@ -268,10 +297,116 @@ impl ExtractorRunnerBuilder {
...
+        // Pull spkg from s3 and copy it at `spkg_path`
+        if !Path::new(&self.config.spkg).exists() {
```
kayibal, line 308: How can I configure where these spkgs will be stored locally?

tycho-indexer/src/main.rs
```diff
@@ -125,35 +133,29 @@ async fn main() -> Result<(), ExtractionError> {
...
     let chain_state = ChainState::new(chrono::Local::now().naive_utc(), block_number);
 
+    let extractors_config = ExtractorConfigs::from_yaml("./tycho-indexer/extractors.yaml")
+        .map_err(|e| ExtractionError::Setup(format!("Failed to load extractors.yaml. {}", e)))?;
+
+    let protocol_systems: Vec<String> = extractors_config
+        .extractors
+        .keys()
+        .cloned()
+        .collect();
```
kayibal, line 143: Unfortunately this breaks modularity quite a bit. I think I'd prefer if each extractor did create their protocol types at the beginning... I know this is currently done on Gateway startup. So maybe out of scope on this PR.

But this would let us move `ExtractorConfigs` and `build_all_extractors` (btw I would rename this then to `start_extractors(config_path)`) into the builder module.

I think we can go almost all the way there by moving `ExtractorsConfigs` and `build_all_extractors` into the builder module.

tycho-indexer/src/extractor/runner.rs
zizou0x: It's the spkg line in the config file. This line is used for both getting the file from s3 and where it's gonna be saved. So basically it will be stored locally with the same path that it is on s3. `substreams/ethereum-uniswap-v2/ethereum-uniswap-v2-v0.1.0.spkg` for example. If we want to add a "snapshot" of the spkg, to avoid having to download the image on every restart, we can just cp the folder `substreams/` from s3 into the image

tycho-indexer/src/main.rs
zizou0x: I 100% agree with this, that's the only reason the config is read in the main function and it's not cool. 

I think this PR is already big enough though and this can probably be done in another PR 👍

substreams/ethereum-ambient/src/contracts/main.rs
zizou0x: Ok I will do this 👍

PR #193, opened 2024-03-14
Speed up inserts

louise-poole:


tycho-storage/migrations/2024-02-16-203155_v0.2.1/down.sql
kayibal, line 1: Is ok to have this in an old migration? I assume you are applying these manually to the dev db?

louise-poole: Yes. I'm not sure what is best practice though? Should I rather have a separate migration?

PR #196, opened 2024-03-20
Add pagination to tokens RPC endpoint

tvinagre:
Adds pagination to tokens endpoint

### DIFF
--- a/tycho-indexer/src/services/rpc.rs
+++ b/tycho-indexer/src/services/rpc.rs
@@ -335,11 +335,10 @@ where
         debug!(?addresses_slice, "Getting tokens.");
 
         let converted_params: PaginationParams = (&request.pagination).into();
-        let pagination: Option<&PaginationParams> = Some(&converted_params);
 
         match self
             .db_gateway
-            .get_tokens(*chain, addresses_slice, pagination)
+            .get_tokens(*chain, addresses_slice, Some(&converted_params))
             .await
         {
             Ok(tokens) => Ok(dto::TokensRequestResponse::new(


tycho-indexer/src/services/rpc.rs
```diff
@@ -316,16 +315,18 @@ where
...
         request: &dto::TokensRequestBody,
+        pagination: &dto::PaginationRequest,
     ) -> Result<dto::TokensRequestResponse, RpcError> {
         info!(?chain, ?request, "Getting tokens.");
-        self.get_tokens_inner(chain, request)
+        self.get_tokens_inner(chain, request, pagination)
             .await
     }
 
     async fn get_tokens_inner(
```
tvinagre: I don't get why do we have these `inner` functions - anyone knows?

zizou0x: If I recall correctly it was done this way to allow us to pass a db connection in our tests. Like this:

`self.get_tokens_inner(chain, request, &mut conn).await`

It was removed by Alan's big PR so we dont need them at all now

tycho-indexer/src/services/rpc.rs
```diff
@@ -531,13 +536,14 @@ pub async fn contract_delta<G: Gateway>(
 )]
 pub async fn tokens<G: Gateway>(
     execution_env: web::Path<Chain>,
+    pagination: web::Path<dto::PaginationRequest>,
```
zizou0x: I think you need to fix Swaager API also, so this probably need to be added to the utoipa decorator above. 

Also add the new structs to the Schemas in `services/mod.rs`

To double check that SwaggerAPI is not broken you can use `tycho-indexer/examples/rpc.rs`

tycho-indexer/src/services/rpc.rs
kayibal: We should be able to mock the gateway now and that means we can use the actix-web testing utils to directly test these endpoints with a mock request.

https://actix.rs/docs/testing/

tycho-core/src/dto.rs
```diff
@@ -815,11 +815,38 @@ impl TokensRequestBody {
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, ToSchema)]
 pub struct TokensRequestResponse {
     pub tokens: Vec<ResponseToken>,
+    /// The current page number.
+    pub page: i64,
+    /// The number of items per page.
+    pub page_size: i64,
```
louise-poole: Why do we need pagination details in the response? Surely the user knows what they set it to in the request?

tycho-storage/src/postgres/protocol.rs
```diff
@@ -2174,26 +2183,38 @@ mod test {
...
+        let tokens = gw
+            .get_tokens(
+                Chain::Ethereum,
+                None,
+                Some(&PaginationParams { page: 0, page_size: 1 }),
+                &mut conn,
+            )
+            .await
+            .unwrap();
+        assert_eq!(tokens.len(), 1);
```
louise-poole, line 2237: Personally would prefer this be in its own stand-alone test. So that pagination (an optional feature) is tested separately from the core functionality of `get_tokens`. But this test is small enough that it's not really that big a deal here, so up to you.

tycho-core/src/dto.rs
tvinagre: It's a good practice, to indicate to the user that this is a paginated request. Since pagination parameters are optional (they have a default) - this makes it very transparent to the user.
Also, it's nice for verification, specially if you're implementing concurrent requests

tycho-indexer/src/services/rpc.rs
```diff
@@ -334,16 +334,20 @@ where
         let addresses_slice = address_refs.as_deref();
         debug!(?addresses_slice, "Getting tokens.");
 
+        let converted_params: PaginationParams = (&request.pagination).into();
+        let pagination: Option<&PaginationParams> = Some(&converted_params);
+
         match self
             .db_gateway
-            .get_tokens(*chain, addresses_slice)
+            .get_tokens(*chain, addresses_slice, pagination)
```
zizou0x: I think we can do this

```suggestion
            .get_tokens(*chain, addresses_slice, Some(&converted_params))
```

tvinagre: tru tru

PR #198, opened 2024-03-20
ENG-2825: Implement Revert Buffer

zizou0x:
This PR implements a new struct called RevertBuffer. As said in the description of the [task](https://datarevenue.atlassian.net/jira/software/projects/ENG/boards/24?selectedIssue=ENG-2815), this component aims to provide us a way to sync unfinalized blocks again (this was turned off because reverting db state became too slow). It's a buffer that contains all unfinalized block updates (so they are not commited to the db anymore) and a bunch of functions for lookups, committing new finalized block to the db and purge reverted blocks.

### DIFF
--- a/tycho-core/src/models/mod.rs
+++ b/tycho-core/src/models/mod.rs
@@ -129,8 +129,8 @@ pub trait NormalisedMessage: std::fmt::Debug + std::fmt::Display + Send + Sync +
     fn source(&self) -> ExtractorIdentity;
 }
 
-pub trait MessageWithBlock<B> {
-    fn block(&self) -> &B;
+pub trait BlockScoped {
+    fn block(&self) -> blockchain::Block;
 }
 
 #[derive(PartialEq, Debug, Clone, Default, Deserialize, Serialize)]
--- a/tycho-indexer/src/extractor/evm/mod.rs
+++ b/tycho-indexer/src/extractor/evm/mod.rs
@@ -12,7 +12,7 @@ use tracing::log::warn;
 use tycho_core::{
     dto,
     models::{
-        Address, AttrStoreKey, Chain, ChangeType, ComponentId, ExtractorIdentity, MessageWithBlock,
+        Address, AttrStoreKey, BlockScoped, Chain, ChangeType, ComponentId, ExtractorIdentity,
         NormalisedMessage, ProtocolType, StoreVal,
     },
     Bytes,
@@ -483,8 +483,8 @@ pub struct BlockContractChanges {
 }
 
 pub trait FilteredUpdates {
-    type IdType;
-    type KeyType;
+    type IdType: std::hash::Hash + std::cmp::Eq + Clone;
+    type KeyType: std::hash::Hash + std::cmp::Eq + Clone;
     type ValueType;
 
     fn get_filtered_state_update(
@@ -1156,29 +1156,32 @@ pub struct BlockEntityChanges {
     pub txs_with_update: Vec<ProtocolChangesWithTx>,
 }
 
-struct BlockMessageWithCursor<B>(B, String);
+struct BlockMessageWithCursor<B> {
+    block: B,
+    cursor: String,
+}
 
 impl<B> BlockMessageWithCursor<B> {
-    fn new(block: B, cursor: String) -> Self {
-        Self(block, cursor)
+    pub fn new(block: B, cursor: String) -> Self {
+        Self { block, cursor }
     }
 }
 
-impl<B: MessageWithBlock<Block>> MessageWithBlock<Block> for BlockMessageWithCursor<B> {
-    fn block(&self) -> &Block {
-        self.0.block()
+impl<B: BlockScoped> BlockScoped for BlockMessageWithCursor<B> {
+    fn block(&self) -> tycho_core::models::blockchain::Block {
+        self.block.block()
     }
 }
 
-impl MessageWithBlock<Block> for BlockEntityChanges {
-    fn block(&self) -> &Block {
-        &self.block
+impl BlockScoped for BlockEntityChanges {
+    fn block(&self) -> tycho_core::models::blockchain::Block {
+        (&self.block).into()
     }
 }
 
-impl MessageWithBlock<Block> for BlockContractChanges {
-    fn block(&self) -> &Block {
-        &self.block
+impl BlockScoped for BlockContractChanges {
+    fn block(&self) -> tycho_core::models::blockchain::Block {
+        (&self.block).into()
     }
 }
 
--- a/tycho-indexer/src/extractor/evm/native.rs
+++ b/tycho-indexer/src/extractor/evm/native.rs
@@ -479,13 +479,14 @@ where
             false => {
                 let mut revert_buffer = self.revert_buffer.lock().await;
                 revert_buffer
-                    .insert_block(BlockMessageWithCursor::new(msg.clone(), inp.cursor.clone()));
+                    .insert_block(BlockMessageWithCursor::new(msg.clone(), inp.cursor.clone()))
+                    .expect("Error while inserting a block into revert buffer");
                 for msg in revert_buffer
                     .drain_new_finalized_blocks(inp.final_block_height)
                     .expect("Final block height not found in revert buffer")
                 {
                     self.gateway
-                        .advance(&msg.0, &msg.1, is_syncing)
+                        .advance(&msg.block, &msg.cursor, is_syncing)
                         .await?;
                 }
             }
--- a/tycho-indexer/src/extractor/revert_buffer.rs
+++ b/tycho-indexer/src/extractor/revert_buffer.rs
@@ -1,14 +1,13 @@
 use std::collections::{HashMap, HashSet, VecDeque};
 
-use ethers::types::H256;
 use tycho_core::{
     dto::TokenBalances,
-    models::{ComponentId, MessageWithBlock},
+    models::{BlockScoped, ComponentId},
     storage::StorageError,
     Bytes,
 };
 
-use super::evm::{Block, FilteredUpdates};
+use super::evm::FilteredUpdates;
 
 /// This buffer temporarily stores blockchain blocks that are not yet finalized. It allows for
 /// efficient handling of block reverts without requiring database rollbacks.
@@ -19,23 +18,29 @@ use super::evm::{Block, FilteredUpdates};
 ///
 /// In case of revert, we can just purge this buffer.
 pub struct RevertBuffer<BM> {
-    blocks: VecDeque<BM>,
+    block_messages: VecDeque<BM>,
 }
 
-impl<BM: MessageWithBlock<Block>> RevertBuffer<BM> {
+impl<BM: BlockScoped> RevertBuffer<BM> {
     pub fn new() -> Self {
-        Self { blocks: VecDeque::new() }
+        Self { block_messages: VecDeque::new() }
     }
 
     /// Inserts a new block into the buffer. Ensures the new block is the expected next block,
     /// otherwise panics.
-    pub fn insert_block(&mut self, new: BM) {
+    pub fn insert_block(&mut self, new: BM) -> Result<(), StorageError> {
         // Make sure the new block matches the one we expect, panic if not.
-        if let Some(last_block) = self.blocks.back() {
-            assert_eq!(last_block.block().hash, new.block().parent_hash);
+        if let Some(last_message) = self.block_messages.back() {
+            if last_message.block().hash != new.block().parent_hash {
+                return Err(StorageError::Unexpected(
+                    "Unexpected block sequence received in revert buffer".to_string(),
+                ));
+            };
         }
 
-        self.blocks.push_back(new);
+        self.block_messages.push_back(new);
+
+        Ok(())
     }
 
     /// Drains blocks up to the specified finalized block height. The last finalized block is kept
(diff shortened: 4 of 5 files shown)


tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -1086,9 +1102,62 @@ pub struct BlockEntityChanges {
     chain: Chain,
     pub block: Block,
     pub revert: bool,
+    /// Vec of updates at this block, aggregated by tx and sorted by tx index in ascending order
     pub txs_with_update: Vec<ProtocolChangesWithTx>,
 }
 
+impl BlockEntityChanges {
```
kayibal: I think the methods implemented here should be a trait, so the rever buffer can be generic over it or you can also do what you planned and just duplicate it since we'll only have it for two types.... it just means also two tests and remembering to always update two places.

tycho-indexer/src/extractor/revert_buffer.rs
```diff
@@ -0,0 +1,89 @@
...
+use tycho_core::{
+    dto::TokenBalances,
+    models::{AttrStoreKey, ComponentId},
+    Bytes,
+};
+
+use super::evm::{BlockContractChanges, BlockEntityChanges};
+
+struct RevertBuffer<B> {
+    data: Vec<B>,
```
kayibal: Maybe use a VecDeque instead, this way you can push to the end and remove from the start?

tycho-indexer/src/extractor/revert_buffer.rs
```diff
@@ -0,0 +1,89 @@
...
+        for block_change in self.data.iter().rev() {
+            if remaning_keys.is_empty() {
+                break;
+            }
+
+            for (key, val) in block_change.get_filtered_state_update(
+                remaning_keys
+                    .iter()
+                    .map(|k| (&k.0, &k.1))
+                    .collect(),
```
kayibal: Nit: Maybe take anything that implements `IntoIter` as argument?

tycho-indexer/src/extractor/revert_buffer.rs
```diff
@@ -0,0 +1,89 @@
...
+                    res.insert(key, val);
+                }
+            }
+        }
+
+        if !res.is_empty() {
+            Some(res)
+        } else {
+            None
+        }
```
kayibal: Isn't it more ergonomic to simply return an empty hash map. vs handling the None case? I guess you can only know once you actually use this method.

Also should we not somehow also return any remaining keys if there are any? Since we will have to query the DB for the previous values of those keys...

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -1086,9 +1102,62 @@ pub struct BlockEntityChanges {
...
+            for (component_id, protocol_update) in update.protocol_states.iter() {
+                for (attr, val) in protocol_update
+                    .updated_attributes
+                    .iter()
+                {
+                    let key = (component_id, attr);
+                    if keys_set.contains(&key) {
+                        res.entry((component_id.clone(), attr.clone()))
+                            .or_insert(val.clone());
+                    }
```
kayibal: I am realising now, that if substreams messages would return the previous value per update, we could completely save us the db round trip on reverts..

Since it looks like we'll have to do that anyway.... it might be worth doing it right away.

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -1086,9 +1102,62 @@ pub struct BlockEntityChanges {
...
+        let mut res = HashMap::new();
+
+        for update in self.txs_with_update.iter().rev() {
+            for (component_id, protocol_update) in update.protocol_states.iter() {
+                for (attr, val) in protocol_update
+                    .updated_attributes
+                    .iter()
+                {
+                    let key = (component_id, attr);
+                    if keys_set.contains(&key) {
```
kayibal: I wonder if a filter on the iterator above wouldn't look a bit cleaner?

tycho-indexer/src/extractor/evm/mod.rs
zizou0x: Yes true, but it's doable only with substreams "extended" block. And these are only available in some chains so I think it better to stay independent of these

tycho-indexer/src/extractor/revert_buffer.rs
```diff
@@ -0,0 +1,520 @@
...
+            Err(StorageError::Unexpected(format!(
+                "Couldnt find block with number {} in revert buffer",
+                final_block_height
+            )))
+        }
+    }
+
+    /// Purges all blocks following the specified block hash from the buffer. Return an error if the
+    /// target hash is not found.
+    pub fn purge(&mut self, target_hash: H256) -> Result<(), StorageError> {
```
zizou0x: We will probably need to return the purged block to send them to the ws clients right?

tycho-core/src/models/mod.rs
```diff
@@ -129,6 +129,10 @@ pub trait NormalisedMessage: std::fmt::Debug + std::fmt::Display + Send + Sync +
     fn source(&self) -> ExtractorIdentity;
 }
 
+pub trait MessageWithBlock<B> {
```
kayibal: I would prefer sth like `BlockScoped` here as a name.

Also does the trait need to be generic? I would prefer if we returned  the `tycho-core` block model here if possible.

tycho-indexer/src/extractor/revert_buffer.rs
```diff
@@ -0,0 +1,520 @@
...
+    pub fn new() -> Self {
+        Self { blocks: VecDeque::new() }
+    }
+
+    /// Inserts a new block into the buffer. Ensures the new block is the expected next block,
+    /// otherwise panics.
+    pub fn insert_block(&mut self, new: BM) {
+        // Make sure the new block matches the one we expect, panic if not.
+        if let Some(last_block) = self.blocks.back() {
+            assert_eq!(last_block.block().hash, new.block().parent_hash);
```
kayibal: Can you please throw an error instead of using assert_eq?

tycho-indexer/src/extractor/revert_buffer.rs
```diff
@@ -0,0 +1,520 @@
...
+        } else {
+            Err(StorageError::Unexpected(format!(
+                "Couldnt find block with number {} in revert buffer",
+                final_block_height
+            )))
+        }
+    }
+
+    /// Purges all blocks following the specified block hash from the buffer. Return an error if the
+    /// target hash is not found.
```
kayibal, line 74: the block with target hash remains in the buffer right?

tycho-indexer/src/extractor/revert_buffer.rs
kayibal: I don't think we need to send these blocks to the clients, they are basically wrong, we only need them to identify the keys that changed within them, which will give you the set of keys to search backwards for.

tycho-indexer/src/extractor/revert_buffer.rs
```diff
@@ -0,0 +1,520 @@
...
+    /// Looks up buffered state updates for the provided keys. Returns a map of updates and a list
+    /// of keys for which updates were not found in the buffered blocks.
+    #[allow(clippy::type_complexity)] //TODO: use type aliases
+    fn lookup_state(
+        &self,
+        keys: &[(&B::IdType, &B::KeyType)],
+    ) -> (HashMap<(B::IdType, B::KeyType), B::ValueType>, Vec<(B::IdType, B::KeyType)>)
+    where
+        B::KeyType: std::hash::Hash + std::cmp::Eq + Clone,
+        B::IdType: std::hash::Hash + std::cmp::Eq + Clone,
```
kayibal: I'd try to include these already on the FilteredUpdate trait directly as type constraints on the associated types.

tycho-indexer/src/extractor/revert_buffer.rs
```diff
@@ -0,0 +1,520 @@
...
+    {
+        let mut res = HashMap::new();
+        let mut remaning_keys: HashSet<(B::IdType, B::KeyType)> = HashSet::from_iter(
+            keys.iter()
+                .map(|(c_id, attr)| (c_id.to_owned().clone(), attr.to_owned().clone())),
```
kayibal: Do you need both `to_owned` and `clone`?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -1086,9 +1171,91 @@ pub struct BlockEntityChanges {
...
+    fn new(block: B, cursor: String) -> Self {
+        Self(block, cursor)
+    }
+}
+
+impl<B: MessageWithBlock<Block>> MessageWithBlock<Block> for BlockMessageWithCursor<B> {
+    fn block(&self) -> &Block {
+        self.0.block()
+    }
+}
```
kayibal: This feels a bit clunky? Maybe you can simplify here:

Merge the functionality into a single trait. The trait should look sth like
```
trait RevertBufferEntry {
     fn block(&self) -> &tycho_core::models::Block;
     fn cursor(&self) -> &String
     fn get_filtered_update(...) -> ...
     fn get_filtered_balance_update(...) -> ...
}
```

tycho-indexer/src/extractor/revert_buffer.rs
```diff
@@ -0,0 +1,520 @@
...
+        }
+    }
+
+    /// Purges all blocks following the specified block hash from the buffer. Return an error if the
+    /// target hash is not found.
+    pub fn purge(&mut self, target_hash: H256) -> Result<(), StorageError> {
+        let mut target_index = None;
+
+        for (index, block) in self.blocks.iter().rev().enumerate() {
+            if block.block().hash == target_hash {
```
zizou0x: Here we could use `parent_hash` and adapt the `target_index`. But I decided to do it like this for code clarity, not sure it's worth to make this more complex to save a tiny amount of computation

tycho-indexer/src/extractor/revert_buffer.rs
```diff
@@ -0,0 +1,520 @@
...
+        revert_buffer.insert_block(get_block_entity(1));
+        revert_buffer.insert_block(get_block_entity(2));
+        revert_buffer.insert_block(get_block_entity(3));
+
+        let finalized = revert_buffer.drain_new_finalized_blocks(3);
+
+        assert_eq!(revert_buffer.blocks.len(), 1);
+        assert_eq!(finalized, Ok(vec![get_block_entity(1), get_block_entity(2)]));
+
+        let unknown = revert_buffer.drain_new_finalized_blocks(999);
```
kayibal, line 509: What happens if called on an empty buffer?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -471,15 +472,87 @@ impl TransactionVMUpdates {
...
+    ) -> HashMap<(Self::IdType, Self::KeyType), Self::ValueType>;
+
+    #[allow(clippy::mutable_key_type)] // Clippy thinks that tuple with Bytes are a mutable type.
+    fn get_filtered_balance_update(
+        &self,
+        keys: Vec<(&String, &Bytes)>,
+    ) -> HashMap<(String, Bytes), ComponentBalance>;
+}
+
+impl FilteredUpdates for BlockContractChanges {
```
kayibal, line 502: I think we miss a few things here: 
- native balance of a contract could have changed
- the code of a contract could have changed
- contracts could have been created and deleted

Apart from the first these changes are very rare - but I think we should be at least aware of this limitation atm.

Unfortunately the current interface is not expressive enough to let us return these things... here and I think adding another method is not great.... either.... maybe given keys, instead of returning a HashMap, we can return the final type we are interested in. Unfortunately I think this might make the scope of this PR explode so tending to just keep it as it is for now.

tycho-indexer/src/extractor/revert_buffer.rs
zizou0x: Yes it does 👍

tycho-indexer/src/extractor/revert_buffer.rs
zizou0x: If called on an empty buffer it returns an error

tycho-indexer/src/extractor/evm/mod.rs
zizou0x: Yes I've thought about them but I would like to keep this for later, probably top prio for next cd week.

tycho-indexer/src/extractor/evm/mod.rs
zizou0x: I don't have a strong opinion but I kind of like to have a generic trait for structs which are block scoped. It's currently only used there but it might be useful at some point elsewhere no?

Or is it a better design to have a single trait for a well defined purpose (as you suggested above)

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -873,9 +960,9 @@ impl ProtocolStateDelta {
 /// Updates grouped by their respective transaction.
 #[derive(Debug, Clone, PartialEq, Default)]
 pub struct ProtocolChangesWithTx {
-    pub new_protocol_components: HashMap<String, ProtocolComponent>,
-    pub protocol_states: HashMap<String, ProtocolStateDelta>,
-    pub balance_changes: HashMap<String, HashMap<H160, ComponentBalance>>,
+    pub new_protocol_components: HashMap<ComponentId, ProtocolComponent>,
+    pub protocol_states: HashMap<ComponentId, ProtocolStateDelta>,
+    pub balance_changes: HashMap<ComponentId, HashMap<H160, ComponentBalance>>,
```
louise-poole, line 965: ❤️

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -1480,7 +1653,14 @@ pub mod fixtures {
                             implementation_type: 0,
                         }),
                     }],
-                    balance_changes: vec![],
+                    balance_changes: vec![BalanceChange {
+                        token: H160::from_str("0x6B175474E89094C44Da98b954EedeAC495271d0F")
+                            .unwrap()
+                            .0
+                            .to_vec(),
+                        balance: 1_i32.to_le_bytes().to_vec(),
```
louise-poole, line 1661: Balances are big endian encoded, attributes are le encoded.

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -475,7 +499,7 @@ where
 
         self.update_cursor(inp.cursor).await;
 
-        Ok(Some(Arc::new(msg.aggregate_updates()?)))
+        return Ok(Some(Arc::new(msg.aggregate_updates()?)))
```
louise-poole, line 502: Why do we need `return` here? Should work fine without it?

PR #201, opened 2024-03-21
ENG 2906: Add balances to snapshots

louise-poole:
Note: once merged I will need to update the tycho version used by protosim and add `retrieve_balances=False` to the extractor registration for the NTQ.

### DIFF
--- a/tycho-core/src/dto.rs
+++ b/tycho-core/src/dto.rs
@@ -784,7 +784,7 @@ pub struct StateRequestParameters {
     pub tvl_gt: Option<u64>,
     #[param(default = 0)]
     pub inertia_min_gt: Option<u64>,
-    #[param(default = 0)]
+    #[param(default = true)]
     pub balances_flag: bool,
 }
 


tycho-client/src/rpc.rs
```diff
@@ -310,7 +310,10 @@ mod tests {
                     "address": "0x0000000000000000000000000000000000000000",
                     "title": "",
                     "slots": {},
-                    "balance": "0x01f4",
+                    "native_balance": "0x01f4",
+                    "balances": {
+                        "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2": 100000000000
+                    },
```
kayibal, line 316: Aren't you assuming here that the contracts map 1:1 to protocol components? I think for things to be correct you would need to move this key one level up, in that case you could also leave the `account.balance` field with the original name.

But then I guess we would need to merge this endpoint with the protocol state endpoint and return accounts instead of attributes...

tycho-storage/src/postgres/contract.rs
```diff
@@ -854,18 +880,33 @@ impl PostgresGateway {
...
+                            "protocol_component".to_string(),
+                            "contract_code".to_string(),
+                            code.id.to_string(),
+                        )
+                    })?;
+
+                let balances = token_balances
+                    .get(component_id)
+                    .unwrap_or(&HashMap::new())
+                    .clone();
```
kayibal: Ah ok you put the erc20 balance on each contract that belongs to the component. It will probably work for us, but its not a good look^^

tycho-storage/src/postgres/contract.rs
```diff
@@ -854,18 +880,33 @@ impl PostgresGateway {
...
+                let code_tx = code.tx.clone().unwrap();
                 let creation_tx = account.tx;
 
+                let component_id = components
+                    .get(&code.id)
+                    .ok_or_else(|| {
+                        StorageError::NoRelatedEntity(
+                            "protocol_component".to_string(),
+                            "contract_code".to_string(),
+                            code.id.to_string(),
```
kayibal: Is this a copy paste error? I'd expect balances mentioned in here?

louise-poole: No, this error relates to finding a component linked to the current account. This is done with the component/contract code joining table. In the next step I use the component retrieved here to get the balances.

tycho-storage/src/postgres/contract.rs
louise-poole: Any suggestions on how else to do this?

tycho-storage/src/postgres/contract.rs
kayibal: Ah yeah but you need the component to get the balances...

tycho-storage/src/postgres/contract.rs
kayibal: Not really without merging `protocol_state` and `contract_state` endpoints into a `component_state` endpoint.

tycho-storage/src/postgres/contract.rs
louise-poole: Ah, okay I see what you're getting at. I will change component to balance then

tycho-core/src/dto.rs
```diff
@@ -776,11 +783,17 @@ pub struct StateRequestParameters {
     pub tvl_gt: Option<u64>,
     #[param(default = 0)]
```
tvinagre, line 785: By addding a 0 default here, will it automatically return Some(0)?

tycho-core/src/dto.rs
```diff
@@ -776,11 +783,17 @@ pub struct StateRequestParameters {
...
     pub inertia_min_gt: Option<u64>,
+    #[param(default = 0)]
```
tvinagre: Same here, needs to be 0 or can it be `false`?

tycho-indexer/src/extractor/evm/convert.rs
```diff
@@ -80,7 +79,8 @@ impl From<&evm::Account> for Contract {
                 .into_iter()
                 .map(|(u, v)| (Bytes::from(u), Bytes::from(v)))
                 .collect(),
-            balance: Bytes::from(value.balance),
+            native_balance: Bytes::from(value.balance),
+            balances: HashMap::new(), // empty balances when converted from Account
```
tvinagre, line 83: Why?

tycho-core/src/dto.rs
louise-poole: I get errors when I try change it to Some(0) so I will leave it as is. I struggled to find the utopia docs to explain these defaults nicely so am not sure what the correct approach is.

tycho-indexer/src/extractor/evm/convert.rs
louise-poole: Because `evm::Account` does not contain any token balances that can be set here. Also, this conversion is only used on the extractor when updating accounts, balances are updated separately and so them being missing here is of no consequence.

PR #202, opened 2024-03-26
Update protocol state query docs

louise-poole:


tycho-storage/src/postgres/protocol.rs
```diff
@@ -1126,10 +1123,11 @@ impl PostgresGateway {
             })?;
 
             // Decode final state deltas. We can assume both the deleted_attrs and state_updates
-            // are sorted by component_id and transaction index. Therefore we can use slices to
-            // iterate over the data in groups of component_id. To do this we first need to collect
-            // an ordered set of the component ids, then we can loop through deleted_attrs and
-            // state_updates in parallel, creating a slice for each component_id.
+            // are sorted by component_idx. Therefore we can use slices to iterate over the data
```
kayibal:
```suggestion
            // are sorted by component_id. Therefore we can use slices to iterate over the data
```

Did you mean id here?

PR #205, opened 2024-03-27
Handle reverts with RevertBuffer in extractors

zizou0x:
This PR aims to make our extractors use the freshly implemented RevertBuffer to handle reverts. This should allow us to be able to sync unfinalized blocks again without worrying about db reverts being too slow.

Still TODO:

- [x] Test
- [x] Clean, optimize

### DIFF
--- a/tycho-storage/src/postgres/cache.rs
+++ b/tycho-storage/src/postgres/cache.rs
@@ -14,7 +14,7 @@ use tokio::{
     },
     task::JoinHandle,
 };
-use tracing::{debug, error, info, trace};
+use tracing::{debug, info, trace};
 use tycho_core::{
     models::{
         self,
@@ -96,10 +96,6 @@ struct BlockRange {
 impl BlockRange {
     fn new(start: &models::blockchain::Block, end: &models::blockchain::Block) -> Self {
         Self { start: start.clone(), end: end.clone() }
-    }
-
-    fn is_single_block(&self) -> bool {
-        self.start.hash == self.end.hash
     }
 }
 
@@ -280,31 +276,6 @@ impl DBCacheWriteExecutor {
             .get()
             .await
             .expect("pool should be connected");
-
-        // If persisted block is not set we don't have data for this chain yet.
-        if let Some(db_block) = &self.persisted_block {
-            // during sync we insert in batches of blocks.
-            let syncing = !new_db_tx.block_range.is_single_block();
-
-            // if we are not syncing (catching up) we are not allowed to create a separate block
-            // range.
-            if !syncing {
-                let start = &new_db_tx.block_range.start;
-                // if we are advancing a block, while not syncing it must fit on top of the
-                // persisted block.
-                if start.number > db_block.number && start.parent_hash != db_block.hash {
-                    error!(
-                        block_range=?&new_db_tx.block_range,
-                        persisted_block=?&db_block,
-                        "Invalid block range encountered"
-                    );
-                    let _ = new_db_tx
-                        .tx
-                        .send(Err(StorageError::InvalidBlockRange()));
-                    return;
-                }
-            }
-        }
 
         let res = conn
             .build_transaction()


tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -517,32 +535,132 @@ where
...
-            // This behaviour is not correct and we will have to rollback the database
-            // to a good state once the revert issue has been fixed.
-            return Ok(None);
-        }
+        let mut revert_buffer = self.revert_buffer.lock().await;
+
+        // Purge the buffer and create a set of state that was reverted
+        let reverted_state = revert_buffer
+            .purge(block_hash.into())
+            .expect("Failed to purge revert buffer");
```
kayibal: We should make this an error and add a lot more info to the error since we will be pretty clueless if this panics here just with that message.

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -517,32 +535,132 @@ where
...
+                &missing_map
+                    .keys()
+                    .map(AsRef::as_ref)
+                    .collect::<Vec<&str>>(),
             )
-            .await?;
+            .await
+            .expect("Failed to retrieve components states during revert");
+
+        for mut missing_state in missing_components_states {
```
kayibal: In terms of flow, would it be possible to lookup state in the buffer, if anything is missing fetch from the db, then merge values from db and values from revert buffer and only do this loop down here once?

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -517,32 +535,132 @@ where
...
+                            .iter()
+                            .flat_map(|(c_id, delta)| {
+                                delta
+                                    .updated_attributes
+                                    .keys()
+                                    .map(move |key| (c_id, key))
+                            })
+                    })
+            })
+            .collect();
```
kayibal: It would be good to also extract any created and deleted `ProtocolComponents` and report them as deleted and created (so the opposite from what you get from these forward messages)

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -517,32 +535,132 @@ where
...
+            !revert_message
+                .new_protocol_components
+                .is_empty() ||
+            !revert_message
+                .deleted_protocol_components
+                .is_empty() ||
+            !revert_message
+                .component_balances
+                .is_empty() ||
+            !revert_message.component_tvl.is_empty();
```
kayibal: Move to `is_empty` method on `BlockEntityChangesResult` ?

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -138,6 +140,11 @@ pub trait NativeGateway: Send + Sync {
         to: &BlockIdentifier,
         new_cursor: &str,
     ) -> Result<evm::BlockEntityChangesResult, StorageError>;
+
+    async fn get_components_state<'a>(
```
louise-poole: Nit pick: I would feel better if we called this `get_component_states`. Why not name it `get_protocol_states` though?

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -517,32 +533,132 @@ where
...
+            extractor: self.name.clone(),
+            chain: self.chain,
+            block: revert_buffer
+                .get_most_recent_block()
+                .expect("Couldn't find most recent block in buffer during revert")
+                .into(),
+            revert: true,
+            state_updates,
+            new_protocol_components: HashMap::new(),
+            deleted_protocol_components: HashMap::new(),
```
louise-poole: What about components that were created on the blocks we're now reverting? Shouldn't we detect them and list them here?

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -517,32 +533,132 @@ where
...
+                    .flat_map(|update| {
+                        update
+                            .protocol_states
+                            .iter()
+                            .flat_map(|(c_id, delta)| {
+                                delta
+                                    .updated_attributes
+                                    .keys()
+                                    .map(move |key| (c_id, key))
+                            })
```
louise-poole: And how do we handle deleted attributes? I think they need to be added to these reverted_keys as they need to be reinstated to their previous values.

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -555,6 +553,12 @@ impl FilteredUpdates for BlockContractChanges {
     }
 }
 
+impl BlockScoped for BlockContractChanges {
+    fn block(&self) -> tycho_core::models::blockchain::Block {
+        (&self.block).into()
+    }
+}
```
kayibal, line 560: I think we should remvoe this one.

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -1218,7 +1204,7 @@ impl FilteredUpdates for BlockEntityChanges {
     fn get_filtered_balance_update(
         &self,
         keys: Vec<(&ComponentId, &Address)>,
-    ) -> HashMap<(String, Bytes), ComponentBalance> {
+    ) -> HashMap<(String, Bytes), tycho_core::models::protocol::ComponentBalance> {
```
kayibal: make short

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -1386,144 +1380,411 @@ pub mod fixtures {
...
+            final_block_height: final_block_height.unwrap_or(420),
             debug_map_outputs: vec![],
             debug_store_outputs: vec![],
         }
     }
 
-    pub fn pb_block_contract_changes() -> crate::pb::tycho::evm::v1::BlockContractChanges {
+    pub fn pb_block_contract_changes(
+        version: u8,
+    ) -> crate::pb::tycho::evm::v1::BlockContractChanges {
```
kayibal, line 1394: Explain how these are created. Like they are designed to trigger the retrieval of some states from the db and other from the revert buffer...

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -517,32 +538,290 @@ where
...
+                });
+
+        trace!("Missing state keys after buffer lookup {:?}", &missing_map);
+
+        let missing_components_states = self
+            .gateway
+            .get_protocol_states(
+                &missing_map
+                    .keys()
+                    .map(AsRef::as_ref)
```
kayibal: `String::as_str`

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -517,32 +538,290 @@ where
...
+
+        let state_updates: HashMap<String, evm::ProtocolStateDelta> = db_states?
+            .into_iter()
+            .chain(buffered_state)
+            .fold(HashMap::new(), |mut acc, ((c_id, key), value)| {
+                acc.entry(c_id.clone())
+                    .or_insert_with(|| evm::ProtocolStateDelta {
+                        component_id: c_id.clone(),
+                        updated_attributes: HashMap::new(),
+                        deleted_attributes: HashSet::new(),
```
kayibal: How do we handle attributes that were created and didn't exist before? I guess this would mean we wouldn't find them in the DB - in that case these should probably be assumed to have been deleted?

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -517,32 +538,290 @@ where
...
+                        })?;
+
+                    let value = state
+                        .attributes
+                        .get(key)
+                        .ok_or_else(|| {
+                            ExtractionError::Storage(StorageError::NotFound(
+                                "Component storage".to_string(),
+                                format!("component: {}, key {}", component_id.clone(), key),
+                            ))
```
kayibal: I think this will trigger if a recently created attribute is reverted. Instead of raising, this should probably be put into a HashSet that will constitute the deleted attributes of the revert message.

tycho-indexer/src/extractor/evm/vm.rs
```diff
@@ -531,51 +548,331 @@ where
...
+        let account_updates =
+            combined_states
+                .into_iter()
+                .fold(HashMap::new(), |mut acc, ((addr, key), value)| {
+                    acc.entry(addr)
+                        .or_insert_with(|| evm::AccountUpdate {
+                            address: addr,
+                            chain: self.chain,
+                            slots: HashMap::new(),
+                            balance: None, //TODO: handle balance changes
```
kayibal, line 740: I think you might already have this available here in the response from the db. Since the db returns contract objects, it should be very straightforward to get any code and native balances changes as well.

tycho-indexer/src/extractor/evm/mod.rs
zizou0x: This impl is required by `RevertBuffer` for `BlockContractChanges` (so in vm)

tycho-indexer/src/extractor/evm/vm.rs
zizou0x: I would prefer to leave it as a TODO and do this in another PR later

PR #208, opened 2024-03-29
Table partitions

kayibal:
I was afraid that our sync times and load times would be only getting worse with increasingly more data being ingested. So I updated Tycho to use partitioned tables in the background. 

Fortunately that was possible without introducing any interface changes to the storage crate.

The migration to partitions is a bit more complex though and requires manual intervention to move historical records.

## Partitioning
The strategy implemented here partitions our 3 biggest tables into daily partitions of data. The gateway also has a retention horizon date. Anything older than that we don't even bother versioning - this greatly speeds up syncing old substreams.

All the latest version entries are kept in the default partition with a `valid_to` date set to `262142-12-31T23:59:59.9999Z`. So no more weird `... OR IS NULL` checks. This also means lookups to the latest state should now be very fast since they only touch the latest data and don't have to filter through any historical records.

Also latest versions are now updated which should help postgres keeping those entries cached in memory.

Creating deltas is now likely more expensive and we might have to adjust our queries here to later on. Since this is not on the hot path anymore with in memory reverts, this is a tradeoff I am very willing to take.

### Data Retention
`pg_partman` can automatically detach partitions once they reach a certain age. We can then run a cronjob to archive the table to S3 (this is supported by aws with a postgres extension) from there we can re-import it in case we ever need it again.

This is not implemented here yet but partition enables this use case and that will allow our database to stay at a controllable  disk usage size.

If we ever need historical deltas again keeping history limited will also ensure that these queries run faster.

## Executed tests

- [x] Sync a fresh uniswap-v3 substream and compare historical states against the graph
- [x] Sync ambient and compare historical states against an archive node
- [x] Sync our v0.1.0 uniswap-v3 substream from scratch for a few days to test retention horizon
- [x] Sync ambient from scratch to test retention horizon on contract storage
- [x] Execute the migration on a 700Mb database dump of a sync on v0.3.0 of roughly 1-2 months of data


Readme.md
```diff
@@ -115,6 +115,17 @@ RUST_LOG=info cargo run -- \
     --stop-block +1000
 ```
 
+#### Quickly run a spkg from our repository:
+Somtimes it is useful to quickly debug the messages from a production spkgs. The command
```
tvinagre:
```suggestion
Sometimes it is useful to quickly debug the messages from a production spkgs. The command 
```

tycho-indexer/src/main.rs
```diff
@@ -56,46 +57,11 @@ struct CliArgs {
...
-
-    /// Substreams stop block
-    ///
-    /// Optional. If not provided, the extractor will run until the latest block.
-    /// If prefixed with a `+` the value is interpreted as an increment to the start block.
-    /// Defaults to STOP_BLOCK env var or None.
     #[clap(long, env)]
-    stop_block: Option<String>,
-}
+    retention_horizon: String,
```
tvinagre: Comments svp

tycho-storage/src/postgres/orm.rs
```diff
@@ -1495,71 +1429,115 @@ impl StoredVersionedRow for ContractStorage {
...
+            .zip(slots.iter())
+            .collect::<HashSet<_>>();
+        Ok(contract_storage::table
+            .select(ContractStorage::as_select())
+            .into_boxed()
+            .filter(
+                contract_storage::account_id
+                    .eq_any(&accounts)
+                    .and(contract_storage::slot.eq_any(&slots))
+                    .and(contract_storage::valid_to.is_null()),
```
tvinagre: Why checking if `is_null`

Readme.md
```diff
@@ -12,11 +12,11 @@ state as quickly as possible.
...
 
--   Clients connect to Tycho's real-time delta messages and start buffering them.
--   Next, they will request a historical state.
--   Once the historical state has been fully retrieved, the buffered changes are applied on top of that state.
--   Now the client possess the latest state of all protocols it is interested in.
--   New incoming messages can now be applied directly to the existing state.
+- Clients connect to Tycho's real-time delta messages and start buffering them.
+- Next, they will request a historical state.
+- Once the historical state has been fully retrieved, the buffered changes are applied on top of that state.
+- Now the client possess the latest state of all protocols it is interested in.
```
zizou0x:
```suggestion
- Now the client possesses the latest state of all protocols it is interested in.
```

tycho-storage/src/postgres/versioning.rs
```diff
@@ -250,56 +210,132 @@ where
...
+    fn get_valid_to(&self) -> NaiveDateTime;
+    fn archive(&mut self, next_version: &mut Self);
+    fn delete(&mut self, delete_version: NaiveDateTime);
+    async fn latest_versions_by_ids(
+        ids: Vec<Self::EntityId>,
+        conn: &mut AsyncPgConnection,
+    ) -> Result<Vec<Self>, StorageError>
+    where
+        Self: Sized;
+}
```
zizou0x: Would be nice to have doc here

scripts/compare-uniswap-v3-the-graph.py
```diff
@@ -1,16 +1,36 @@
...
+        } for cid in component_ids
+      ],
+      "version": {
+        "block": {
+          "chain": "ethereum",
+          "number": block
+        }
+      }
+    }
+    uri="http://127.0.0.1:4242/v1/ethereum/protocol_state?balances_flag=false"
```
tvinagre, line 24: would be nice to also check if balances are correct, before we resync usv2 and v3

tycho-storage/migrations/2024-03-16-103329_v0.5.0/up.sql
```diff
@@ -0,0 +1,295 @@
...
+    valid_from,
+    valid_to,
+    inserted_ts,
+    modified_ts,
+    protocol_component_id
+FROM
+    latest_protocol_state;
+
+--- Create historical partitions, maybe this should not be done as part of the migration since it is quite slow....
+--- Unfortunately columns need to match to can't keep id column in old table to migrate data...
```
tvinagre:
```
--- Unfortunately columns need to match so can't keep id column in old table to migrate data...
```

tycho-storage/migrations/2024-03-16-103329_v0.5.0/up.sql
```diff
@@ -0,0 +1,295 @@
...
+DROP TRIGGER IF EXISTS invalidate_previous_entry_protocol_calls_contract ON
+    protocol_calls_contract;
+
+-- PROTOCOL STATE
+ALTER TABLE protocol_state RENAME TO protocol_state_old;
+
+UPDATE
+    protocol_state_old
+SET
+    valid_to = '262142-12-31T23:59:59.9999Z'
```
tvinagre: If we didn't set `Z` in the end, do u know if Postgres would default to UTC or would it try to regionalize according to the OS time?

tycho-indexer/src/cli.rs
```diff
@@ -64,6 +64,12 @@ pub struct IndexArgs {
     /// Extractors configuration file
     #[clap(long, env, default_value = "./extractors.yaml")]
     pub extractors_config: String,
+
+    /// Retention horizon date
+    ///
+    /// Any data before this date is not kept in storage.
+    #[clap(long, env, default_value = "2024-01-01T00:00:00")]
```
tvinagre, line 71: No Z needed?

tvinagre: 2024-01-01T00:00:00Z

tycho-storage/src/postgres/mod.rs
```diff
@@ -157,6 +157,23 @@ mod versioning;
 
 const MIGRATIONS: EmbeddedMigrations = embed_migrations!("./migrations/");
 
+// +262142-12-31T23:59:59.999999999
+const MAX_TS: NaiveDateTime = NaiveDateTime::MAX;
+
+lazy_static! {
+    /// Simplifies querying current and historical versions by introducing a special marker version.
```
tvinagre, line 164: Nicee

tycho-storage/src/postgres/mod.rs
```diff
@@ -388,14 +405,24 @@ async fn maybe_lookup_version_ts(
 pub(crate) struct PostgresGateway {
     protocol_system_id_cache: Arc<ProtocolSystemEnumCache>,
     chain_id_cache: Arc<ChainEnumCache>,
+    /// Any versions dated before this date will be discarded and never be inserted into
```
tvinagre: by dated, I think it's nice to enforce that you mean "valid_to" date and not "created_at" date

tycho-storage/src/postgres/schema.rs
```diff
@@ -302,31 +232,132 @@ diesel::joinable!(protocol_component_holds_contract -> contract_code (contract_c
...
+        modify_tx -> Int8,
+        protocol_component_id -> Int8,
+        inserted_ts -> Timestamptz,
+        valid_from -> Timestamptz,
+        valid_to -> Timestamptz,
+    }
+}
+
+diesel::table! {
+    component_balance_default (protocol_component_id, token_id, modify_tx) {
```
tvinagre, line 279: Not complaining, but any reason why these are called `default` and not `latest` or `active` ?

tycho-storage/src/postgres/versioning.rs
```diff
@@ -250,56 +210,132 @@ where
...
+///
+/// ## Batch Updates
+/// If inserting a lot of rows, as is usually the case, and the update contains multiple version of
+/// the same entity, we directly create the archival version on the application side saving us
+/// multiple round trips to the database. This method will handle this for you.
+///
+/// ## Retention Horizon
+/// Partitioned tables usually have a retention horizon meaning any outdated versions
+/// older than the horizon are not kept in storage. To achieve this, archive versions strictly older
+/// than the horizon are simply dropped before issuing the inserts.
```
tvinagre, line 297: Do you think it's worth manually moving the state previous to Jan 1st 2024 to S3 before we merge this? Otherwise, old data will automatically be purged and will be unrecoverable, no?

tvinagre: And is this purge done real-time or via cronjob? 
Because if we're syncing - it would apply extra pressure on the database

tycho-storage/src/postgres/versioning.rs
```diff
@@ -250,56 +210,132 @@ where
...
+///
+/// ## Deletions
+/// Deletion simply move the row from the default partition to an archive parition by setting the
+/// valid_to column and skipping the update or insert into the current state.
```
tvinagre, line 301: Why skip the update into the current state? Shouldn't we remove it? Or we will keep the latest value without changing the last valid_to ?

tvinagre: From the code I thought we removed it from the default partition

tvinagre: Hm I guess by moving the row you mean that the old row will be removed from default table

tycho-storage/src/postgres/versioning.rs
```diff
@@ -250,56 +210,132 @@ where
...
+///
+/// ## Overview
+///
+/// This function will execute the following steps:
+///
+/// - Retrieve the current state of all entities to be updated or deleted.
+/// - Apply application side versioning, calling either delete or archive on the respective rows.
+/// - Filter any archived rows by the retention horizon.
+///
```
tvinagre, line 310: This doc is amazing. Thank you!

tycho-storage/migrations/2024-03-16-103329_v0.5.0/down.sql
```diff
@@ -0,0 +1 @@
+-- This file should undo anything in `up.sql`
```
zizou0x: I forgot about this, is it fine to merge without it?

tycho-storage/migrations/2024-03-16-103329_v0.5.0/up.sql
kayibal: If a timestamp ends with Z it is in UTC

tycho-indexer/src/cli.rs
kayibal: No I think it is not needed here since on the application side we only deal with naive timestamps where the tz is assumed to be UTC.

tycho-storage/src/postgres/schema.rs
kayibal: I think pg_partman extensions wants these named default I tried changing it but it was too much effort

tycho-storage/src/postgres/versioning.rs
kayibal: We already manually "partitioned" dev + we have wrong state and potentially missing pools so I don't think we need to store that.

This doc does not mention how the current migration works. The migration to this release will rename the current tables by appending `_old` suffix to their name, then it will migrate only the current state. Any old state that you'd like to have in your new partitioned table you have to migrate manually (with another script).

tycho-storage/src/postgres/versioning.rs
kayibal: So a deletion happens by removing the row from the default partition, then reinserting it into archive with valid_to set. This is the normal process for any update actually, just if it is an update, we don't remove the row from the default partition but instead just update it :]

tycho-storage/migrations/2024-03-16-103329_v0.5.0/down.sql
kayibal: I don't think we want to go backwards via a migration on such a big change^^ worst case I'd prefer to restore from snapshot.

PR #217, opened 2024-04-09
Change snapshot format for vm protocols.

kayibal:
Instead of handling vm an native protocols entirely differently, vm protocols will also request native protocol state to retrieve balances first, then add any required vm storage to the snapshot.

This allows us to correctly retrieve balances per component for vm protocols where the contract does not map 1:1 to contract/account (e.g. ambient).

### DIFF
--- a/tycho-client/src/feed/synchronizer.rs
+++ b/tycho-client/src/feed/synchronizer.rs
@@ -49,20 +49,14 @@ struct SharedState {
 }
 
 #[derive(Clone, PartialEq, Debug, Serialize)]
-pub struct VMSnapshot {
-    pub state: HashMap<Bytes, ResponseAccount>,
-    pub component: ProtocolComponent,
-}
-
-#[derive(Clone, PartialEq, Debug, Serialize)]
-pub struct NativeSnapshot {
+pub struct ComponentWithState {
     pub state: ResponseProtocolState,
     pub component: ProtocolComponent,
 }
 
 #[derive(Clone, PartialEq, Debug, Default, Serialize)]
 pub struct Snapshot {
-    states: HashMap<String, NativeSnapshot>,
+    states: HashMap<String, ComponentWithState>,
     vm_storage: HashMap<Bytes, ResponseAccount>,
 }
 
@@ -207,7 +201,7 @@ where
                 if let Some(state) = protocol_states.remove(&component.id) {
                     Some((
                         component.id.clone(),
-                        NativeSnapshot { state, component: component.clone() },
+                        ComponentWithState { state, component: component.clone() },
                     ))
                 } else if ids.contains(&&component.id) {
                     // only emit error event if we requested this component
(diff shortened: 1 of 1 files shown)


tycho-client/src/feed/synchronizer.rs
```diff
@@ -33,6 +33,7 @@ pub type SyncResult<T> = anyhow::Result<T>;
 pub struct ProtocolStateSynchronizer<R: RPCClient, D: DeltasClient> {
     extractor_id: ExtractorIdentity,
     is_native: bool,
+    #[allow(dead_code)]
```
zizou0x, line 36: Should we remove it?

kayibal: Yes we can

tycho-client/src/feed/synchronizer.rs
```diff
@@ -180,14 +189,44 @@ where
             return Ok(StateSyncMessage { header, ..Default::default() });
         }
 
-        if !self.is_native {
-            let contract_ids = tracked_components.get_contracts_by_component(ids);
+        let component_ids = tracked_components.get_tracked_component_ids();
+        let mut protocol_states = self
+            .rpc_client
+            .get_protocol_states_paginated(self.extractor_id.chain, &component_ids, &version, 50, 4)
```
louise-poole, line 189: Could we not use a larger chunk size? How was this chosen?

tycho-client/src/feed/synchronizer.rs
```diff
@@ -180,14 +189,44 @@ where
...
+            .collect();
+
+        let contract_ids = tracked_components.get_contracts_by_component(ids.clone());
+        let vm_storage = if !contract_ids.is_empty() {
             let contract_states = self
                 .rpc_client
                 .get_contract_state(
                     self.extractor_id.chain,
-                    &StateRequestParameters::new(self.retrieve_balances),
+                    &StateRequestParameters::new(false),
```
louise-poole, line 223: Should we remove the balance retrieval logic from the contract state gateway then?

tycho-client/src/feed/synchronizer.rs
```diff
@@ -59,18 +60,25 @@ pub struct NativeSnapshot {
...
 }
 
-#[derive(Clone, PartialEq, Debug, Serialize)]
-pub enum Snapshot {
-    VMSnapshot(VMSnapshot),
-    NativeSnapshot(NativeSnapshot),
+#[derive(Clone, PartialEq, Debug, Default, Serialize)]
+pub struct Snapshot {
+    states: HashMap<String, NativeSnapshot>,
+    vm_storage: HashMap<Bytes, ResponseAccount>,
```
louise-poole, line 60: We could also remove `balances` from the `ResponseAccount` struct no? I won't be used anymore.

tycho-client/src/feed/synchronizer.rs
```diff
@@ -59,18 +60,25 @@ pub struct NativeSnapshot {
     pub component: ProtocolComponent,
 }
 
-#[derive(Clone, PartialEq, Debug, Serialize)]
-pub enum Snapshot {
-    VMSnapshot(VMSnapshot),
```
louise-poole, line 64: We can remove this VMSnapshot struct entirely now, no?

tycho-client/src/feed/synchronizer.rs
kayibal: Random guess but it runs quick enough afaik.

tycho-client/src/feed/synchronizer.rs
kayibal: That is for later

PR #218, opened 2024-04-10
Allow to upsert contracts

kayibal:
One of the balancer pools already existed before we actually synced the substream. Most likely because the LP token is traded on another already synced protocol. In this case the same contract is used as token as well as for pools.

### DIFF
--- a/tycho-storage/src/postgres/cache.rs
+++ b/tycho-storage/src/postgres/cache.rs
@@ -44,7 +44,7 @@ pub(crate) enum WriteOp {
     // Simply keep last
     SaveExtractionState(ExtractionState),
     // Support saving a batch
-    InsertContract(Vec<models::contract::Contract>),
+    UpsertContract(Vec<models::contract::Contract>),
     // Simply merge
     UpdateContracts(Vec<(TxHash, models::contract::ContractDelta)>),
     // Simply merge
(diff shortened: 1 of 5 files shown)


tycho-storage/src/postgres/cache.rs
```diff
@@ -63,7 +63,7 @@ impl WriteOp {
             WriteOp::UpsertBlock(_) => "UpsertBlock",
             WriteOp::UpsertTx(_) => "UpsertTx",
             WriteOp::SaveExtractionState(_) => "SaveExtractionState",
-            WriteOp::InsertContract(_) => "InsertContract",
+            WriteOp::UpsertContract(_) => "InsertContract",
```
zizou0x: Need update

tycho-storage/src/postgres/contract.rs
```diff
@@ -989,12 +997,22 @@ impl PostgresGateway {
...
+        let account_id = {
+            use schema::account::dsl;
+            diesel::insert_into(schema::account::table)
+                .values(new_contract.new_account())
+                .on_conflict(on_constraint("account_chain_id_address_key"))
+                .do_update()
+                .set((
+                    dsl::title.eq(excluded(dsl::title)),
+                    dsl::creation_tx.eq(excluded(dsl::creation_tx)),
+                    dsl::created_at.eq(excluded(dsl::created_at)),
```
zizou0x, line 1009: Update code also?

kayibal: non

PR #219, opened 2024-04-10
Implement token analyzer

zizou0x:


token-analyzer/src/http_client.rs
```diff
@@ -0,0 +1,120 @@
+use anyhow::{anyhow, Result};
+use reqwest::{Client, ClientBuilder, Response};
+use std::{
+    fmt::{self, Display, Formatter},
+    time::Duration,
+};
+
+const USER_AGENT: &str = "cowprotocol-services/2.0.0";
```
kayibal: Let's change this I see this causing confusion later down the line.

substreams/ethereum-ambient/substreams.yaml
```diff
@@ -19,34 +19,33 @@ binaries:
 modules:
   - name: map_pool_changes
     kind: map
-    initialBlock: 17361664
+    initialBlock: 17381533
```
kayibal: Why did you change these?

token-analyzer/src/main.rs
```diff
@@ -33,15 +19,25 @@ async fn main() -> Result<(), ()> {
         "transport".to_owned(),
     ));
     let w3 = Web3::new(transport);
-    let tf = TokenFinder {};
+    let tf = TokenFinder::new(HashMap::from([(
```
kayibal: Let's remove main from token-analyzer or better let's put it under examples?

token-analyzer/src/trace_call.rs
```diff
@@ -28,16 +28,20 @@ pub struct TraceCallDetector {
...
 }
 
 #[async_trait::async_trait]
 impl BadTokenDetecting for TraceCallDetector {
-    async fn detect(&self, token: H160) -> Result<(TokenQuality, Option<U256>, Option<U256>)> {
-        let quality = self.detect_impl(token).await?;
+    async fn detect(
+        &self,
+        token: H160,
+        block: BlockNumber,
```
kayibal, line 42: Can we maybe use our types in public interfaces and cast them to the required types in the implementations?

tycho-indexer/src/extractor/evm/native.rs
```diff
@@ -254,9 +256,27 @@ where
...
         if !new_tokens_addresses.is_empty() {
+            let balance_map: HashMap<H160, (H160, U256)> = balance_changes
+                .iter()
+                .filter_map(|bc| match H160::from_str(bc.component_id.as_str()) {
+                    Ok(address) => Some((
+                        bc.token.clone().into(),
+                        (address, U256::from_big_endian(&bc.new_balance.clone().to_vec())),
+                    )),
+                    Err(e) => {
+                        warn!("Error parsing component_id to H160: {}", e);
```
kayibal, line 267: This should be debug or maybe even trace. If we fail to analyze a token e.g. because after this no balances are left, in that case we can issue a warn.

tycho-indexer/src/extractor/evm/token_pre_processor.rs
```diff
@@ -1,40 +1,64 @@
...
 }
 #[cfg_attr(test, mockall::automock)]
 #[async_trait]
 pub trait TokenPreProcessorTrait: Send + Sync {
-    async fn get_tokens(&self, addresses: Vec<H160>) -> Vec<ERC20Token>;
+    async fn get_tokens(
+        &self,
+        addresses: Vec<H160>,
+        token_finder: Arc<dyn TokenOwnerFinding>,
+        block: BlockNumber,
```
kayibal, line 35: Can we use internal type here as well?

substreams/ethereum-ambient/substreams.yaml
zizou0x: mb

token-analyzer/src/trace_call.rs
zizou0x: I also don't like it but let's keep it for cd week

tycho-indexer/src/extractor/evm/native.rs
zizou0x: I would prefer to keep it a warning now, as we really don't expect this to happen, and change it to debug if it's becomes too much (like if we add a protocol that breaks this basically)

PR #224, opened 2024-04-10
Add token quality filter to RPC

tvinagre:


tycho-storage/src/postgres/protocol.rs
```diff
@@ -724,6 +728,12 @@ impl PostgresGateway {
             query = query.filter(schema::account::address.eq_any(addrs));
         }
 
+        if let Some(min_quality) = min_quality {
+            // let min_quality =
+            // diesel::dsl::sql::<diesel::sql_types::BigInt>(&min_quality.to_string());
```
louise-poole, line 733: We can remove this now no?

PR #232, opened 2024-04-15
Filter tokens by recent activity

tvinagre:


tycho-core/src/storage.rs
```diff
@@ -354,6 +354,7 @@ pub trait ProtocolGateway {
         chain: Chain,
         address: Option<&[&Address]>,
         min_quality: Option<i32>,
+        traded_n_days_ago: Option<NaiveDateTime>,
```
kayibal, line 357: Should be named differently now that it is a datetime, e.g: `traded_after_ts`

tycho-indexer/src/services/rpc.rs
```diff
@@ -387,9 +388,15 @@ where
         let converted_params: PaginationParams = (&request.pagination).into();
         let min_quality = request.min_quality;
 
+        let traded_n_days_ago = request.traded_n_days_ago;
+
+        let n_days_ago = traded_n_days_ago
+            .filter(|&days| days >= 0)
```
kayibal: Symptom of the type not being u64

PR #234, opened 2024-04-16
Impl get token in tycho client

zizou0x:


tycho-client/src/rpc.rs
```diff
@@ -474,4 +526,77 @@ mod tests {
...
+                symbol: "USDC".to_string(),
+                decimals: 6,
+                tax: 0,
+                gas: vec![Some(40652)],
+            },
+        ];
+
+        mocked_server.assert();
+        assert_eq!(response.tokens, expected);
+        assert_eq!(response.pagination, PaginationParams { page: 0, page_size: 2 });
```
louise-poole, line 639: Should this not be page 1? From my understanding the response should hold the next page that you will want to request.

louise-poole: Ah, it's a hardcoded test asset 😆 AG then

tycho-client/src/rpc.rs
```diff
@@ -287,11 +293,57 @@ impl RPCClient for HttpRPCClient {
 
         Ok(states)
     }
+
+    async fn get_tokens(
```
tvinagre, line 339: I think you'll need a `get_all_tokens`  as this function will hardly be used solo. 
`get_all_tokens` should spawn the request with different pages until it gets an empty response. 
Also, would be nice to document this function to be explicit that this just spawns the request and parses the response, but doesn't handle pagination

tycho-client/src/rpc.rs
tvinagre: No, the response is just a copy from the request actually. If it had the next page it'd be smth like a cursor

PR #241, opened 2024-04-24
Handle sorting outside of versioning, add tx index when sorting

zizou0x:
This PR aims to fix a bug in the versioning that could happen when two transactions were changing the same value at the same block. 

**Issue:** We currently rely on transaction id to sort them by execution order when applying the versioning, however even though we insert new transactions ordered by execution order, Postgres doesn't insure that their id will respect our insertion order. 

For example: in a block, `tx 10` and `tx 20` change the same value. When inserting them in the db, it's possible that `tx 20` get a lower id than `tx 10`. Then during the versioning we would consider that the latest state of the value is given by `tx 10` instead of `tx 20`.

**Fix:** To fix this we now consider tx index instead of tx id for when sorting updates. Also we moved the sorting out of the versioning module for simplicity.

### DIFF
--- a/tycho-storage/src/postgres/protocol.rs
+++ b/tycho-storage/src/postgres/protocol.rs
@@ -1260,6 +1260,7 @@ impl PostgresGateway {
                 schema::component_balance::new_balance,
                 schema::component_balance::balance_float,
             ))
+            .order(schema::component_balance::protocol_component_id.asc())
             .get_results::<(i64, i64, Balance, f64)>(conn)
             .await
             .map_err(PostgresError::from)?


tycho-storage/src/postgres/mod.rs
```diff
@@ -280,6 +286,65 @@ impl<T> Deref for WithTxHash<T> {
...
+    ordinal: O,
+}
+
+impl<T, O> Deref for WithOrdinal<T, O> {
+    type Target = T;
+
+    fn deref(&self) -> &Self::Target {
+        &self.entity
+    }
+}
```
kayibal: Unfortunately this is an anti pattern. We shouldn't do this.

tycho-storage/src/postgres/mod.rs
```diff
@@ -280,6 +286,65 @@ impl<T> Deref for WithTxHash<T> {
...
+#[derive(Debug)]
+struct SortedVec<T>(Vec<T>);
+
+impl<T> Deref for SortedVec<T> {
+    type Target = Vec<T>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
```
kayibal: Same here this is an anti pattern: https://rust-unofficial.github.io/patterns/anti_patterns/deref.html#disadvantages

tycho-storage/src/postgres/mod.rs
```diff
@@ -280,6 +286,65 @@ impl<T> Deref for WithTxHash<T> {
...
+    fn from(mut value: Vec<WithOrdinal<T, O>>) -> Self {
+        value.sort_by_cached_key(|v| v.ordinal.clone());
+        SortedVec(
+            value
+                .into_iter()
+                .map(|v| v.entity)
+                .collect(),
+        )
+    }
+}
```
kayibal: It's a obscure type tbh... It will only be sorted if you use the from method... I'd probably prefer just doing things explicitly.

tycho-storage/src/postgres/mod.rs
```diff
@@ -280,6 +286,65 @@ impl<T> Deref for WithTxHash<T> {
     }
 }
 
+// Helper type to wrap entities with their associated ordinal.
+#[derive(Debug)]
+struct WithOrdinal<T, O> {
+    entity: T,
+    ordinal: O,
+}
```
kayibal, line 288: Maybe the versioning trait can rely on this instead. Could be sth like:
```
pub async fn apply_versioning<'a, N, S, O>(
    new_data: &mut SortedVec<WithOrdinal<N, O>>,
    conn: &mut AsyncPgConnection,
) -> Result<(), StorageError>
where
    N: VersionedRow,
    S: StoredVersionedRow<EntityId = N::EntityId, Version = N::Version>,
    <N as VersionedRow>::EntityId: Clone,
    O: PartialEq,
```

But I really don't expect this from you. If you are curious please give it a try but if it does not immediately work, let's just use a sorted `Vec<N>` here. And we document that it needs to be passed in correctly sorted order.

It probably makes sense for you to implement PartialEq on WithOrdinal so you can easily sort vectors with structs of this type.

tycho-storage/src/postgres/protocol.rs
```diff
@@ -738,12 +738,15 @@ impl PostgresGateway {
...
-                            tx_db.2,
+                        WithOrdinal::new(
+                            orm::NewProtocolState::new(
+                                component_db_id,
+                                attribute,
+                                Some(value),
+                                *tx_id,
+                                *tx_ts,
+                            ),
+                            ((component_db_id, attribute), tx_ts, tx_index),
```
kayibal: Why the double nested tuple here?

tycho-storage/src/postgres/protocol.rs
```diff
@@ -764,9 +767,12 @@ impl PostgresGateway {
 
         // insert the prepared protocol state deltas
         if !state_data.is_empty() {
-            apply_delta_versioning::<_, orm::ProtocolState>(&mut state_data, conn).await?;
+            let mut sorted = state_data.into();
```
kayibal: It is very obscure that into sorts the data in the background. It would be much clearer if you just sorted the data here.

tycho-storage/src/postgres/protocol.rs
```diff
@@ -1031,7 +1037,7 @@ impl PostgresGateway {
                     error!(?chain, ?component_balance.token, ?component_balance, "Token not found");
                     StorageError::NotFound("Token".to_string(), component_balance.token.to_string())
                 })?;
-            let (transaction_id, transaction_ts) =
+            let (transaction_id, transaction_index, transaction_ts) =
                 transaction_ids_and_ts[&component_balance.modify_tx];
```
kayibal: This can panic, would you mind making this an error instead like above? Or at least use an expect?

PR #246, opened 2024-05-06
docs(pec): protocol external contracts design doc.

kayibal:


docs/ProtocolExternalContracts.md
```diff
@@ -0,0 +1,390 @@
...
+implementation contract code and whenever the implementation is updated, we have to
+re-execute the analysis.
+
+The static analysis works as follows:
+
+- Check the code for the presence of `STATICCALL`, `CALL` or `DELEGATECALL` opcodes
+    - If not terminate
+    - Note: `DELEGATECALL` likely indicates a proxy
+- Find all PUSH20 - PUSH32 opcodes and collect their parameters
+- Check if the parameter looks like an address (~20 bytes, high entropy), if it does save it
```
zizou0x, line 112: This part seems complex. I'm not sure but I think this would work only if the external contract is hardcoded in the contract bytecodes (`constant` or `immutable` in solidity). But if it's a storage variable it could be a `SLOAD` + `SWAP` no?

docs/ProtocolExternalContracts.md
```diff
@@ -0,0 +1,390 @@
...
+- Next check the first N (e.g. 512) storage slots values for address like values, save them
+- Repeat until you have collected all addresses
+
+The procedure above should give you all possibly called addresses in a high percentage
+of cases.
+
+#### Dynamic contract indexing
+
+Once we get a set of new contracts, we need a way to index them. This includes
+retrieving a one time snapshot and receiving all state updates to these contracts from
```
zizou0x, line 122: How do you see this implemented? It's hard to get a one time snapshot no?

docs/ProtocolExternalContracts.md
zizou0x: Very draft idea, but did you consider doing something like using rEVM with a Tycho cached state **and** RPC reader for missing things, if we have to do any rpc read call, we notify DCI to index this one asap

tvinagre: I thought of a similar idea - but then we come back to the same loop of doing things node-specific.

tvinagre: But, in the end - if we're analyzing OPCODES, we would need to rewrite this for every new VM architecture

kayibal: Yes correct the code analyzer would need to be chain specific, same as the token analyzer atm. I don't think there is a way around that.

I did consider using revm - but this would:
1. Add a lot of latency, since for each storage access we need to do a rpc request
2. Apart from that it would add a lot of complexity, having to import protosim on tycho... apart from this being a cyclical dependency

I think the static code analysis here will do a fine job.

kayibal: > But if it's a storage variable it could be a SLOAD + SWAP no?

This is not meant to detect any addresses that are in storage, we have a special step for this too.

docs/ProtocolExternalContracts.md
kayibal: Ah no, it's just calling `debug_storageRangeAt` on a Erigon or Geth archive node. I tested this against uniswap-v3 USDC/WETH pool and it takes like 1-2s to get all the storage slots.

PR #250, opened 2024-05-10
Fixes realtime tvl & tokens

kayibal:
Several fixes to get realtime token and tvl updates:

1. Cache was initialized once per extractor, this was too slow and used too much memory
2. Unknown shared tokens upon insertion, this was a hard one pls see commit description for details.
3. Tvl calculation was done wrongly, leading to huge values - fixed now
4. Last but not least the client also had a bug when retrieving snapshots..

### DIFF
--- a/tycho-indexer/src/extractor/runner.rs
+++ b/tycho-indexer/src/extractor/runner.rs
@@ -187,7 +187,7 @@ impl ExtractorRunner {
                 }
             }
         }
-            // Additional inner debug span with substreams information
+            // Additional inner info span with substreams information
             // trace_id is set later on in process_substreams_response
         .instrument(tracing::info_span!("loop", trace_id = tracing::field::Empty)))
     }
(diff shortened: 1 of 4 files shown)


tycho-indexer/src/extractor/runner.rs
```diff
@@ -190,7 +189,7 @@ impl ExtractorRunner {
         }
             // Additional inner debug span with substreams information
```
zizou0x:
```suggestion
            // Additional inner info span with substreams information
```

docker-compose.yaml
```diff
@@ -8,7 +8,7 @@ services:
       POSTGRESQL_DATABASE: tycho_indexer_0
       POSTGRESQL_USERNAME: postgres
     ports:
-      - "5432:5432"
+      - "5431:5432"
```
zizou0x, line 11: Please adapt `DATABASE_URL` in the `tycho-indexer` below

kayibal: Shouldn't be necessary since this only changes the port on the host network not on the docker network.

PR #251, opened 2024-05-13
SDK Tycho binary

zizou0x:
This PR aims to provide a way for external teams to test their modules on Tycho (integration test).

For that this PR adds two CLI commands: `run` to run a single extractor and `rpc` to run only Tycho RPC.

Beside that, there are some general CLI improvement (moving some required flags to subcommands, add `chain` flag for multichain, ...)

### DIFF
--- a/tycho-indexer/src/cli.rs
+++ b/tycho-indexer/src/cli.rs
@@ -67,7 +67,7 @@ pub struct GlobalArgs {
 
 #[derive(Args, Debug, Clone, PartialEq)]
 pub struct SubstreamsArgs {
-    /// Ethereum node rpc url
+    /// Node rpc url
     #[clap(env, long)]
     pub rpc_url: String,
 


Readme.md
```diff
@@ -106,13 +106,25 @@ substreams pack substreams/ethereum-ambient/substreams.yaml
 
 7. Run `tycho-indexer` locally using cli:
 
+We currently support 4 commands:
+
+- `index` : Run every extractor set in `./extractors.yaml`
+- `run` : Run a single extractor
+- `analyze-tokens` : Run token analyzer cronjob
+- `rpc` : Run only the server part of Tycho
```
kayibal:
```suggestion
- `rpc` : Run only the http rpc server 
```

tycho-indexer/src/cli.rs
```diff
@@ -57,10 +51,34 @@ pub struct GlobalArgs {
     /// Substreams API endpoint
     #[clap(name = "endpoint", long, default_value = "https://mainnet.eth.streamingfast.io")]
     pub endpoint_url: String,
+
+    /// The blockchain we index
+    #[clap(long, default_value = "ethereum")]
+    pub chain: String,
```
kayibal: What if I want to run multiple chains?

tycho-indexer/src/cli.rs
```diff
@@ -74,6 +92,14 @@ pub struct IndexArgs {
...
 #[derive(Args, Debug, Clone, PartialEq, Eq)]
 pub struct RunSpkgArgs {
+    /// Ethereum node rpc url
+    #[clap(env, long)]
+    pub rpc_url: String,
+
+    /// Substreams API token
+    #[clap(long, env, hide_env_values = true, alias = "api_token")]
+    pub substreams_api_token: String,
+
```
kayibal: Could we make a shared substreams args struct and include it via flatten?

See example here: 
https://rust-cli-recommendations.sunshowers.io/handling-arguments.html

tycho-indexer/src/cli.rs
```diff
@@ -82,6 +108,10 @@ pub struct RunSpkgArgs {
     #[clap(long)]
     pub module: String,
 
+    // The names of the protocol_types to index
+    #[clap(long, value_delimiter = ' ')]
```
kayibal: Hm I don't like using space so much on CLI, maybe we can use `,`?

tycho-indexer/src/main.rs
```diff
@@ -166,6 +164,98 @@ async fn run_indexer(
     res.expect("Extractor- nor ServiceTasks should panic!")
 }
 
+async fn run_spkg(global_args: GlobalArgs, run_args: RunSpkgArgs) -> Result<(), ExtractionError> {
```
kayibal: Seems like we repeat a lot of boilerplate here, maybe we can share with the index command if we factor out a function that takes `extractors` as a parameter?

tycho-storage/src/postgres/builder.rs
```diff
@@ -41,9 +41,13 @@ impl GatewayBuilder {
 
         let inner_gw = PostgresGateway::new(pool.clone(), self.retention_horizon).await?;
         let (tx, rx) = mpsc::channel(10);
+        let chain = self
```
tvinagre, line 44: This is failing if you try the extractor. Not sure but I think the `default` on the chain parameter is not working properly with a vec.

tycho-client/src/feed/synchronizer.rs
```diff
@@ -190,7 +190,7 @@ where
         let request_ids = ids
             .map(|it| {
                 it.into_iter()
-                    .map(|id| ProtocolId { id: id.clone(), chain: Chain::Ethereum })
+                    .map(|id| ProtocolId { id: id.clone(), chain: Chain::Ethereum }) //TODO: remove chain assumption
```
tvinagre, line 193: Not sure if you'll do this in this PR or in a future one but let's not forget about this :)

tycho-core/src/dto.rs
```diff
@@ -38,6 +38,7 @@ pub enum Chain {
     Ethereum,
     Starknet,
     ZkSync,
+    Arbitrum,
```
tvinagre, line 41: 🚀

Readme.md
```diff
@@ -106,13 +106,25 @@ substreams pack substreams/ethereum-ambient/substreams.yaml
...
+- `analyze-tokens` : Run token analyzer cronjob
+- `rpc` : Run only the http rpc server
+
+Each command can be used with the following:
+
+```bash
+RUST_LOG=info cargo run --bin tycho-indexer -- \
+    {your-command}
+```
+
```
tvinagre, line 122: Maybe add instructions to build and run from the build binary?

tycho-indexer/src/cli.rs
```diff
@@ -32,48 +32,79 @@ pub enum Command {
...
+    pub server_port: u16,
+
+    /// The server version prefix
+    #[clap(long, default_value = "v1")]
+    pub server_version_prefix: String,
+}
+
+#[derive(Args, Debug, Clone, PartialEq)]
+pub struct SubstreamsArgs {
+    /// Ethereum node rpc url
```
tvinagre:
```suggestion
    /// Node rpc url
```

Readme.md
tvinagre: You'll also need to export your substreams key and RPC_URL envvar

tycho-indexer/src/extractor/evm/token_analysis_cron.rs
```diff
@@ -172,6 +171,9 @@ mod test {
             concurrency: 10,
             update_batch_size: 100,
             fetch_batch_size: 100,
+            rpc_url:
+                "https://ethereum-mainnet.core.chainstack.com/71bdd37d35f18d55fed5cc5d138a8fac"
```
tvinagre, line 175: Tests will run in a pvt node? Not sure that's nice

zizou0x: Its was already the case as this value was hardcoded. I don't like it either and was surprised, not sure why/how it's used

tycho-client/src/feed/synchronizer.rs
zizou0x: I left it there so we can change this on the multichain PR 👍

Readme.md
zizou0x: > Maybe add instructions to build and run from the build binary?

This is general Rust/cargo knowledge, I'm not sure it needs to be in the readme


> You'll also need to export your substreams key and RPC_URL envvar

Very true, I completely forgot about it because it's in my zsh profile

zizou0x: Actually substreams key is explained above and rpc is part of the cli so I think we're good

tycho-indexer/src/extractor/evm/token_analysis_cron.rs
zizou0x: Ah ok but this test is ignored so all good

PR #255, opened 2024-05-15
Unify Entity and Contract changes on Protobuf

tvinagre:
We want to make our Native and VM protocols hybrid, removing the structural distinction between them. The first step is to refactor our protobuf messages, by removing duplicate entries and keeping all changes inside one single structure, named `TransactionChanges`. 
We'll also remove the distinction between `BlockEntityChanges` and `BlockContractChanges` - by unifying them all in a single `BlockChanges` message.

**NOTES**
`entity.proto` and `vm.proto` won't be used anymore and will be deleted. I just decided to keep them to prevent potentially breaking stuff for now.

### DIFF
--- a/tycho-indexer/src/pb/tycho.evm.v1.rs
+++ b/tycho-indexer/src/pb/tycho.evm.v1.rs
@@ -85,9 +85,6 @@ pub struct ProtocolComponent {
     /// / Represents the functionality of the component.
     #[prost(message, optional, tag = "6")]
     pub protocol_type: ::core::option::Option<ProtocolType>,
-    /// Transaction where this component was created
-    #[prost(message, optional, tag = "7")]
-    pub tx: ::core::option::Option<Transaction>,
 }
 /// A struct for following the changes of Total Value Locked (TVL) of a protocol component.
 /// Note that if the ProtocolComponent contains multiple contracts, the TVL is tracked for the


proto/tycho/evm/v1/common.proto
```diff
@@ -96,3 +96,60 @@ message BalanceChange {
...
+// A component is a set of attributes that are associated with a custom entity.
+message EntityChanges {
+  // A unique identifier of the entity within the protocol.
+  string component_id = 1;
+  // The set of attributes that are associated with the entity.
+  repeated Attribute attributes = 2;
+}
+
+// VM entities
+// A key value entry into contract storage.
```
louise-poole: For my ocd:
```suggestion
// VM entities

// A key value entry into contract storage.
```

proto/tycho/evm/v1/common.proto
```diff
@@ -96,3 +96,60 @@ message BalanceChange {
...
+  // The new code of the contract, empty bytes indicates no change.
+  bytes code = 3;
+  // The changes to this contract's slots, empty sequence indicates no change.
+  repeated ContractSlot slots = 4;
+  // Whether this is an update, a creation or a deletion.
+  ChangeType change = 5;
+}
+
+// A set of changes aggregated by transaction.
+message TransactionChanges {
```
louise-poole, line 137: Would be nice to have a section heading for these messages too... the last heading was 'VM entities' when implies these are still VM specific here. Maybe 'Aggregate messages' or something?

tycho-indexer/src/pb/tycho.evm.v1.rs
```diff
@@ -6,59 +6,59 @@
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct Block {
     /// The blocks hash.
-    #[prost(bytes = "vec", tag = "1")]
+    #[prost(bytes="vec", tag="1")]
```
louise-poole: Looks like someone needs to double check this auto-formatter? 👀 (might not be you... I'm not sure what the agreed formatter is here)

tycho-indexer/src/pb/tycho.evm.v1.rs
```diff
@@ -67,41 +67,121 @@ pub struct ProtocolType {
...
+    #[prost(enumeration="ChangeType", tag="5")]
+    pub change: i32,
+}
+/// A set of changes aggregated by transaction.
+#[allow(clippy::derive_partial_eq_without_eq)]
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct TransactionChanges {
+    /// The transaction instance that results in the changes.
+    #[prost(message, optional, tag="1")]
+    pub tx: ::core::option::Option<Transaction>,
```
louise-poole: Why is it optional?

tycho-indexer/src/pb/tycho.evm.v1.rs
```diff
@@ -67,41 +67,121 @@ pub struct ProtocolType {
...
+    #[prost(message, repeated, tag="5")]
+    pub balance_changes: ::prost::alloc::vec::Vec<BalanceChange>,
+}
+/// A set of transaction changes within a single block.
+#[allow(clippy::derive_partial_eq_without_eq)]
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct BlockChanges {
+    /// The block for which these changes are collectively computed.
+    #[prost(message, optional, tag="1")]
+    pub block: ::core::option::Option<Block>,
```
louise-poole: And this? I know it's like that in the old message structs too... just seems odd though no?

proto/tycho/evm/v1/common.proto
kayibal: Why does it say VM entities? It is contract storage... not sure what vm entities would be?

proto/tycho/evm/v1/common.proto
kayibal: Is this message backwards compatible with the other two transaction changes? It would be nice if it could be, else once we update tycho, all older substreams will stop being supported.

tvinagre: Yes, it should as all the fields are optional on protobuf. They will just have either no `contract_changes` or no `entity_changes` - depending if the protocol is VM or Native

proto/tycho/evm/v1/common.proto
tvinagre: To differentiate VM and Native entities. We expect VM to have ContractSlot and ContractChange, while we expect Native to have EntityChanges. I can happily remove this comment if it bugs u - I'm just trying to be as verbose as possible for the external teams

tycho-indexer/src/pb/tycho.evm.v1.rs
tvinagre: On Substreams or Proto, not sure which one, every field is optional AFAK

tycho-indexer/src/pb/tycho.evm.v1.rs
```diff
@@ -6,59 +6,59 @@
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct Block {
     /// The blocks hash.
-    #[prost(bytes = "vec", tag = "1")]
+    #[prost(bytes="vec", tag="1")]
     pub hash: ::prost::alloc::vec::Vec<u8>,
     /// The parent blocks hash.
-    #[prost(bytes = "vec", tag = "2")]
+    #[prost(bytes="vec", tag="2")]
```
louise-poole: You fixed and then immediately broke it again 😂

tvinagre: lol

PR #256, opened 2024-05-17
ENG 3183: Merge tycho client vm and native structs 

louise-poole:
TODO:
- [x] update struct tests
- [x] update tycho client to emit these new structs

### DIFF
--- a/tycho-core/src/dto.rs
+++ b/tycho-core/src/dto.rs
@@ -1052,7 +1052,7 @@ mod test {
     }
 
     #[test]
-    fn test_parse_block_account_changes() {
+    fn test_parse_block_changes() {
         let json_data = r#"
         {
             "extractor": "vm:ambient",
@@ -1074,6 +1074,13 @@ mod test {
                     "balance": "0x01f4",
                     "code": "",
                     "change": "Update"
+                }
+            },
+            "state_updates": {
+                "component_1": {
+                    "component_id": "component_1",
+                    "updated_attributes": {"attr1": "0x01"},
+                    "deleted_attributes": ["attr2"]
                 }
             },
             "new_protocol_components":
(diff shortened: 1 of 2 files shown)


tycho-core/src/dto.rs
```diff
@@ -489,64 +432,6 @@ pub struct ComponentBalance {
...
-    pub new_tokens: HashMap<Bytes, ResponseToken>,
-    pub state_updates: HashMap<String, ProtocolStateDelta>,
-    pub new_protocol_components: HashMap<String, ProtocolComponent>,
-    pub deleted_protocol_components: HashMap<String, ProtocolComponent>,
-    pub component_balances: HashMap<String, TokenBalances>,
-    pub component_tvl: HashMap<String, f64>,
-}
-
-impl BlockEntityChangesResult {
-    pub fn merge(mut self, other: Self) -> Self {
```
tvinagre, line 514: Do we need this `merge` for `BlockChangesResult` or do we already have it?

louise-poole: The `BlockChangesResult` you see here is the client message struct, not the internal tycho struct. Since we're changing the client to use `BlockChanges` instead, we only need a merge function for that now and can drop these other structs.

tvinagre: Not sure if we're talking about the same thing. We do have a new `BlockChangesResult`, not this `BlockEntityChangesResult` that's deprecated

tvinagre: Anyways, if we have this `merge` function - ag

louise-poole: This `BlockChanges` here is the client representation of our internal `BlockChangesResult`. No point in the word 'result' in the client structs when they have no other version of it (internally we have versions: non-aggregrated and aggregated)

PR #258, opened 2024-05-21
ENG-3181: Merge Native and VM Structs

tvinagre:


tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -1474,6 +1487,506 @@ impl BlockEntityChanges {
     }
 }
 
+/// Changes grouped by their respective transaction.
+#[derive(Debug, Clone, PartialEq, Default)]
+pub struct TxWithChanges {
+    pub protocol_components: HashMap<ComponentId, ProtocolComponent>,
+    pub account_updates: HashMap<H160, AccountUpdate>,
+    pub protocol_states: HashMap<ComponentId, ProtocolStateDelta>,
```
louise-poole: I would prefer this be called `state_updates`. `protocol_states` implies they're full/complete states where they're not, they're just the changes.

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -1474,6 +1487,506 @@ impl BlockEntityChanges {
...
+        Ok(())
+    }
+}
+
+/// A container for updates grouped by protocol component.
+///
+/// Hold a single update per component. This is a condensed form of
+/// [BlockChanges].
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, Default)]
+pub struct BlockChangesResult {
```
louise-poole: Maybe we can take this opportunity can change this name to `AggregatedBlockChanges`? Seems more fitting

tycho-indexer/src/extractor/evm/mod.rs
tvinagre: lollll don't do that to me now pls 🤣

PR #259, opened 2024-05-21
ENG-3182: Add Hybrid extractor

tvinagre:


tycho-indexer/src/extractor/revert_buffer.rs
```diff
@@ -271,18 +271,22 @@ where
 {
     /// Looks up buffered protocol state updates for the provided keys. Returns a map of updates and
     /// a list of keys for which updates were not found in the buffered blocks.
+    ///
+    /// Clippy thinks it is a complex type that is difficult to read
```
zizou0x: `///` makes this comment appear in the function description no?

tycho-indexer/src/extractor/evm/hybrid.rs
louise-poole, line 1: It looks like we're creating a whole new extractor for this. Is there a point then to keeping the old extractors around?  They're now outdated no?

tvinagre: I'll remove them - but wanted to do it in a separate PR, after everything is up and running

louise-poole: Then I think when we do that we remove the word 'hybrid' from the names here. It's weird to have stuff called 'hybrid' where there are no 'non-hybrid' ones in comparison.

tvinagre: Doesn't bother me that much bc it's a hybrid - can process both VM and Native protocols, but I see what you mean - we can change!

PR #260, opened 2024-05-22
Remove abstraction between native and vm states from Revert Buffer

tvinagre:


tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -550,14 +550,28 @@ pub struct BlockContractChanges {
...
 impl StateUpdateBufferEntry for BlockContractChanges {
-    type IdType = H160;
-    type KeyType = U256;
-    type ValueType = U256;
+    type ProtocolStateIdType = ();
+    type ProtocolStateKeyType = ();
+    type ProtocolStateValueType = ();
+    type AccountStateIdType = H160;
+    type AccountStateKeyType = U256;
+    type AccountStateValueType = U256;
```
zizou0x: Now that it's two separate functions we can remove this from the trait and just use "fixed" types no?

PR #261, opened 2024-05-29
[RELEASE] SDK 

tvinagre:


tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -1474,108 +1493,1582 @@ impl BlockEntityChanges {
...
+    fn get_filtered_account_state_update(
+        &self,
+        keys: Vec<(&AccountStateIdType, &AccountStateKeyType)>,
+    ) -> HashMap<(AccountStateIdType, AccountStateKeyType), AccountStateValueType> {
+        let keys_set: HashSet<_> = keys.into_iter().collect();
+        let mut res = HashMap::new();
+
+        for update in self.txs_with_update.iter().rev() {
+            for (address, account_update) in update.account_updates.iter() {
+                for (attr, val) in account_update
```
zizou0x: I'm being picky but I wouldn't call it `attr` but more something like `slot` or `storage_key`

PR #268, opened 2024-06-04
Fix tycho client hanging on initial connection

louise-poole:
There are 2 places in the code where we try establish the ws connection. In the previous fix attempt we improved the handling of errors for the connection within the retry loop. We overlooked the initial connection code (`connect_async(&ws_uri).await?;` - errors are still unhandled here). The fix was to remove the initial connection attempt and allow the retry loop to handle all connection attempts.

### DIFF
--- a/tycho-client/src/deltas.rs
+++ b/tycho-client/src/deltas.rs
@@ -589,8 +589,11 @@ impl DeltasClient for WsDeltasClient {
                 debug!("Connected to WebSocket server");
 
                 let (ws_tx_new, ws_rx_new) = conn.split();
-                let mut guard = this.inner.as_ref().lock().await;
-                *guard = Some(Inner::new(cmd_tx.clone(), ws_tx_new, this.subscription_buffer_size));
+                {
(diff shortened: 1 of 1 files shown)


tycho-client/src/deltas.rs
```diff
@@ -562,12 +565,10 @@ impl DeltasClient for WsDeltasClient {
         info!(?ws_uri, "Starting TychoWebsocketClient");
 
         let (cmd_tx, mut cmd_rx) = mpsc::channel(self.ws_buffer_size);
-        let (conn, _) = connect_async(&ws_uri).await?;
-        let (ws_tx, ws_rx) = conn.split();
-        let mut ws_rx = Some(ws_rx);
+        let mut ws_rx: Option<SplitStream<WebSocketStream<MaybeTlsStream<TcpStream>>>> = None;
```
kayibal: If this is None, do we even still need the if else condition within the retry loop? I think we don't instead the retry loop should simply connect at the start independent of ws_rx.

PR #271, opened 2024-06-07
feat(substreams): add logs only Substreams module for Uniswap V3

zizou0x:
This PR introduces a new Substreams module that indexes Uniswap V3 with not extended blocks. 

I introduced this in a new folder instead of replacing the old one because it feels more like a new module than a new version. Eventually we can make this a complete replacement of the old one.

TODO: check state correctness (latest status on this is Tycho doesn't delete old attributes when this substreams send a ChangeType::Deletion)


substreams/evm-uniswap-v3-logs-only/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "substreams-ethereum-uniswap-v3"
+version = "0.1.0"
+edition = "2021"
+
+[lib]
+name = "substreams_ethereum_uniswap_v3"
```
kayibal, line 7: I think if you want this to be a new module it should have a different name here too.

substreams/evm-uniswap-v3-logs-only/ethereum-uniswap-v3.yaml
```diff
@@ -0,0 +1,124 @@
+specVersion: v0.1.0
+package:
+  name: "substreams_ethereum_uniswap_v3"
```
kayibal, line 3: Change this too.

substreams/evm-uniswap-v3-logs-only/src/pb/mod.rs
```diff
@@ -0,0 +1,17 @@
+// @generated
+pub mod tycho {
+    pub mod evm {
+        // @@protoc_insertion_point(attribute:tycho.evm.v1)
+        pub mod v1 {
+            include!("tycho.evm.v1.rs");
+            // @@protoc_insertion_point(tycho.evm.v1)
+        }
+    }
+}
```
kayibal, line 10: Shouldn't we be importing these from the SDK?

PR #274, opened 2024-06-13
fix: make block ts distinguishable

zizou0x:
I noticed that our ts precision on blocks isn't enough to clearly make a distinction between blocks on chain with subsecond blocktime like Arbitrum. 

Unfortunately, the blockchain itself doesn't give a more precise timestamp: 

> Question asked:
Does Arbitrum work with a more precise timestamp (millis/micro/nanos part) than standard Ethereum which is in seconds? (edited) 
Answer:
No, Arbitrum uses seconds, but consecutive blocks can have the same timestamp

We then decided to append a part of the block number as microseconds, this will help us to maintain an accurate ordering when sorting blocks by timestamps.

Here is how it looks like:

```
number   |ts                           |
---------+-----------------------------+
187765303|2024-03-06 20:20:22.000303+01|
187765304|2024-03-06 20:20:23.000304+01|
187765305|2024-03-06 20:20:23.000305+01|
187765306|2024-03-06 20:20:23.000306+01|
187765307|2024-03-06 20:20:24.000307+01|
187765308|2024-03-06 20:20:24.000308+01|
187765309|2024-03-06 20:20:24.000309+01|
187765310| 2024-03-06 20:20:24.00031+01|
187765311|2024-03-06 20:20:25.000311+01|
187765312|2024-03-06 20:20:25.000312+01|
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4067,6 +4067,37 @@ dependencies = [
 ]
 
 [[package]]
+name = "opentelemetry-otlp"
+version = "0.14.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f24cda83b20ed2433c68241f918d0f6fdec8b1d43b7a9590ab4420c5095ca930"
+dependencies = [
+ "async-trait",
+ "futures-core",
+ "http 0.2.11",
+ "opentelemetry",
+ "opentelemetry-proto",
+ "opentelemetry-semantic-conventions",
+ "opentelemetry_sdk",
+ "prost 0.11.9",
+ "thiserror",
+ "tokio",
+ "tonic 0.9.2",
+]
+
+[[package]]
+name = "opentelemetry-proto"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a2e155ce5cc812ea3d1dffbd1539aed653de4bf4882d60e6e04dcf0901d674e1"
+dependencies = [
+ "opentelemetry",
+ "opentelemetry_sdk",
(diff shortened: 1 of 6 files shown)


tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -618,12 +618,16 @@ impl Block {
...
-            ts: NaiveDateTime::from_timestamp_opt(msg.ts as i64, 0).ok_or_else(|| {
-                ExtractionError::DecodeError(format!(
-                    "Failed to convert timestamp {} to datetime!",
-                    msg.ts
-                ))
-            })?,
+            // For blockchains with subsecond block times, like Arbitrum, timestamps aren't precise
+            // enough to distinguish between two blocks accurately. To maintain accurate ordering,
+            // we adjust timestamps by appending part of the current block number as microseconds.
+            ts: NaiveDateTime::from_timestamp_opt(msg.ts as i64, (msg.number as u32 % 1000) * 1000)
```
zizou0x: Is 1000 enough or should I use something bigger as `%`?

louise-poole: I reckon 1000 is enough - doubt we'll ever deal with a chain that has more than 1000 blocks/sec. Having said that, is there any reason to not use a larger value? If not, why not just go larger anyway? Looks like we can safely go as big as `100 000`?

zizou0x: I doubt 1000 blocks/sec is realistic either so I decided to leave 1000 for now

PR #275, opened 2024-06-13
feat: Add open telemetry tracing.

kayibal:


tycho-indexer/src/extractor/runner.rs
```diff
@@ -157,7 +158,7 @@ impl ExtractorRunner {
                                         trace!("No message to propagate.");
                                     }
                                     Err(err) => {
-                                        error!(error = %err, msg = ?data, "Error while processing tick!");
+                                        error!(error = %err, "Error while processing tick!");
```
kayibal, line 161: The data field can be very huge, this is why I removed it here. Same for some other attributes below..

PR #277, opened 2024-06-14
Add token caching to the RPC

louise-poole:
TODO: 
- test this for multiple simultaneous requests
- adjust cache size (currently 100 items)

### DIFF
--- a/tycho-indexer/src/services/rpc.rs
+++ b/tycho-indexer/src/services/rpc.rs
@@ -386,7 +386,7 @@ where
             let read_lock = self.cache.read().await;
             if let Some(cached_response) = read_lock.get(&cache_key) {
                 trace!("Returning cached response");
-                return Ok(cached_response.clone());
+                return Ok(cached_response);
             }
         }
 
@@ -396,7 +396,7 @@ where
         // Double-check if another thread has already fetched and cached the data
         if let Some(cached_response) = write_lock.get(&cache_key) {
             trace!("Returning cached response after re-check");
-            return Ok(cached_response.clone());
+            return Ok(cached_response);
         }
 
         let response = self


tycho-client/src/rpc.rs
```diff
@@ -164,15 +165,19 @@ pub trait RPCClient {
...
 }
 
 impl HttpRPCClient {
     pub fn new(base_uri: &str) -> Result<Self, RPCError> {
         let uri = base_uri
             .parse::<Uri>()
             .map_err(|e| RPCError::UriParsing(base_uri.to_string(), e.to_string()))?;
+        // cache is limited to 100 entries and each entry lives for a max 24 hours
+        let cache =
+            Arc::new(Mutex::new(TimedSizedCache::with_size_and_lifespan(100, 24 * 60 * 60)));
```
louise-poole: The cache is set to store 100 items. I'm not sure what size to limit it to?

tycho-client/src/rpc.rs
```diff
@@ -341,6 +346,20 @@ impl RPCClient for HttpRPCClient {
...
+        dbg!(&cache_key);
+
+        // Check the cache for a cached response
+        if let Some(cached_response) = self
+            .cache
+            .lock()
+            .unwrap()
+            .cache_get(&cache_key)
+        {
+            trace!(?cached_response, "Returning cached response");
```
kayibal: The response is likely too big for tracing. I would prefer if we had a span on the server with an attribute, `cached=true|false`.

tycho-client/src/rpc.rs
```diff
@@ -164,15 +165,19 @@ pub trait RPCClient {
 pub struct HttpRPCClient {
     http_client: Client<HttpConnector>,
     uri: Uri,
+    cache: Arc<Mutex<TimedSizedCache<String, TokensRequestResponse>>>,
```
kayibal: I think a RWLock would be preferred here, this way we can read concurrently if nobody is writing (which will be most of the time).

Don't forget to use the double check pattern, like ChatGPT recommended, this way concurrent requests will not trigger the same sql query multiple times.

PR #281, opened 2024-06-18
Fix tycho rpc component endpoint bug

louise-poole:
The get_components endpoint would query the db directly for components and ignore those in the revert buffer. Now we first check the revert buffer and only go to the db if the requested components have not all been found.

### DIFF
--- a/tycho-indexer/src/services/deltas_buffer.rs
+++ b/tycho-indexer/src/services/deltas_buffer.rs
@@ -201,9 +201,8 @@ impl PendingDeltas {
         Ok(change_found)
     }
 
-    #[allow(dead_code)]
     pub fn get_new_components(&self, ids: Option<&[&str]>) -> Result<Vec<ProtocolComponent>> {
-        let id_set: Option<HashSet<&str>> = ids.map(|ids| ids.iter().cloned().collect());
+        let requested_ids: Option<HashSet<&str>> = ids.map(|ids| ids.iter().cloned().collect());
         let mut new_components = Vec::new();
         for buffer in self.native.values() {
             let guard = buffer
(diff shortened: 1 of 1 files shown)


tycho-indexer/src/services/deltas_buffer.rs
```diff
@@ -202,7 +202,8 @@ impl PendingDeltas {
     }
 
     #[allow(dead_code)]
```
kayibal: Is this still dead?

tycho-indexer/src/services/deltas_buffer.rs
```diff
@@ -213,7 +214,14 @@ impl PendingDeltas {
                     entry
                         .new_components
                         .clone()
-                        .into_values(),
+                        .into_values()
+                        .filter(|comp| {
+                            if let Some(ref set) = id_set {
```
kayibal:
```suggestion
                            if let Some(ids) = requested_ids.as_ref() {
```

tycho-indexer/src/services/deltas_buffer.rs
```diff
@@ -202,7 +202,8 @@ impl PendingDeltas {
     }
 
     #[allow(dead_code)]
-    pub fn get_new_components(&self) -> Result<Vec<ProtocolComponent>> {
+    pub fn get_new_components(&self, ids: Option<&[&str]>) -> Result<Vec<ProtocolComponent>> {
+        let id_set: Option<HashSet<&str>> = ids.map(|ids| ids.iter().cloned().collect());
```
kayibal: please rename this to requested_ids or similar.

tycho-indexer/src/services/deltas_buffer.rs
```diff
@@ -226,7 +234,14 @@ impl PendingDeltas {
                     entry
                         .new_components
                         .clone()
-                        .into_values(),
+                        .into_values()
+                        .filter(|comp| {
+                            if let Some(ref set) = id_set {
```
kayibal: Same as above

PR #283, opened 2024-06-20
Handle same ts blocks for Arbitrum in hybrid extractor

zizou0x:
This correctly fixes https://github.com/propeller-heads/tycho-indexer/pull/274. The previous fix wasn't working properly at the boundaries - every 1k blocks, we had the timestamp added which changed from 999 to 000, and therefore an incorrect order.

### DIFF
--- a/tycho-indexer/src/extractor/evm/hybrid.rs
+++ b/tycho-indexer/src/extractor/evm/hybrid.rs
@@ -5,7 +5,7 @@ use std::{
 };
 
 use async_trait::async_trait;
-use chrono::NaiveDateTime;
+use chrono::{Duration, NaiveDateTime};
 use ethers::prelude::{H160, H256, U256};
 use mockall::automock;
 use prost::Message;
@@ -14,15 +14,16 @@ use tracing::{debug, error, info, instrument, trace, warn};
 
 use token_analyzer::TokenFinder;
 use tycho_core::{
-    models,
     models::{
+        self,
         contract::Contract,
         protocol::{ComponentBalance, ProtocolComponentState},
         token::CurrencyToken,
         Address, Chain, ChangeType, ExtractionState, ExtractorIdentity, ProtocolType, TxHash,
     },
     storage::{
-        ChainGateway, ContractStateGateway, ExtractionStateGateway, ProtocolGateway, StorageError,
+        BlockIdentifier, ChainGateway, ContractStateGateway, ExtractionStateGateway,
+        ProtocolGateway, StorageError,
     },
     Bytes,
 };
@@ -48,6 +49,9 @@ use crate::{
 pub struct Inner {
     cursor: Vec<u8>,
     last_processed_block: Option<Block>,
+    /// Keeps count of consecutive blocks with the same timestamp. This is then used to add some
+    /// microseconds in the blocks and maintain order.
+    same_ts_blocks_streak: u8,
     /// Used to give more informative logs
     last_report_ts: NaiveDateTime,
     last_report_block_number: u64,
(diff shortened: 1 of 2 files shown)


tycho-indexer/src/extractor/evm/hybrid.rs
```diff
@@ -113,6 +118,13 @@ where
                 }
             }
             Ok(cursor) => {
+                let last_processed_block = gateway
+                    .get_latest_block()
+                    .await
+                    .unwrap_or_else(|err| {
+                        error!(%chain, %err, "Found a cursor but was unable to retreive latest block");
+                        panic!("Unexpected error when fetching latest block {}", err);
+                    });
```
kayibal, line 127: If we are syncing, this might return a future block no? I have the feeling we might have to get the block from the cursor :/

tycho-indexer/src/extractor/evm/hybrid.rs
```diff
@@ -575,6 +597,25 @@ where
         let mut msg =
             if let Some(post_process_f) = self.post_processor { post_process_f(msg) } else { msg };
 
+        if self.chain == Chain::Arbitrum {
```
kayibal, line 600: Can we activate this more explicitly? This way we could avoid forgetting about it in the future (e.g. imagine syncing Fantom or similar and forgetting to set this, then we have to resync everything again).

Also we need to decide if this should happen before or after the post_processor...

tycho-indexer/src/extractor/evm/hybrid.rs
```diff
@@ -600,10 +600,10 @@ where
         if self.chain == Chain::Arbitrum {
             if let Some(last_processed_block) = self.get_last_processed_block().await {
                 let mut inner = self.inner.lock().await;
-                let current_ts =
-                    msg.block.ts + Duration::microseconds(inner.same_ts_blocks_streak.into());
+                let latest_processed_block_ts = last_processed_block.ts -
+                    Duration::microseconds(inner.same_ts_blocks_streak.into());
```
kayibal, line 604: What we substract microseconds now?

tycho-indexer/src/extractor/evm/hybrid.rs
zizou0x: I don't think we can have a divergence here. If a cursor was stored, then it means the latest stored block should be exactly at the cursor. No matter if we're syncing or not, or am I missing something?

I can add an additional security by getting the block number from the cursor and compare it to the latest block number. The only thing is IIRC it's no so straightforward to get the block number from the cursor, there are many cases no?

tycho-indexer/src/extractor/evm/hybrid.rs
zizou0x: I think the simplest would be to activate it for every chains. It shouldn't affect blockchains that have over 1 sec block time. But it would mean a lot of tests to fix I think.

> before or after the post_processor...

I'd expect this to be after the post_processor because most of the time post_processor are based directly on what we expect to receive from the substreams package. But I don't have a big preference here, I doubt that it will ever conflict anyway (but we never know)

tycho-indexer/src/extractor/evm/hybrid.rs
zizou0x: So what we're doing here is we try to see if the `last_processed_block` ts is the same as the new. But as we might already have appended seconds to the `last_processed_block` we need to sub them to compare with the new. 

What was done previously was adding them to the new instead of subing them from the last. I think it was shadier. Another idea would be to save the "raw"/"unmodified" block timestamp in the Inner. But it might be redundant with the `last_processed_block`. What do you think?

tycho-indexer/src/extractor/evm/hybrid.rs
kayibal: If there is other extractors that have already fully synced, retrieving the latest block will give you a block.ts that is very close to the current time.

Now if the extractor is syncing... but other aren't then it might or might no insert a block. It depends on what range of block was already synced... So ideally we'd need to retrieve the last block that this extractor has processed so far - not the latest block...

zizou0x: That is true

PR #287, opened 2024-06-24
feat: Add block_id column to extraction_state table

zizou0x:
The purpose of this change is to establish a relationship between the extraction_state and block tables, allowing for retrieval of the latest indexed block associated with an extraction states.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5950,7 +5950,7 @@ checksum = "1f3ccbac311fea05f86f61904b462b55fb3df8837a366dfc601a0161d0532f20"
 
 [[package]]
 name = "token-analyzer"
-version = "0.13.0"
+version = "0.14.0"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -6440,7 +6440,7 @@ dependencies = [
 
 [[package]]
 name = "tycho-client"
-version = "0.13.0"
+version = "0.14.0"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -6470,7 +6470,7 @@ dependencies = [
 
 [[package]]
 name = "tycho-core"
-version = "0.13.0"
+version = "0.14.0"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -6496,7 +6496,7 @@ dependencies = [
 
 [[package]]
 name = "tycho-indexer"
-version = "0.13.0"
+version = "0.14.0"
 dependencies = [
  "actix",
  "actix-rt",
@@ -6556,7 +6556,7 @@ dependencies = [
 
 [[package]]
 name = "tycho-storage"
-version = "0.13.0"
+version = "0.14.0"
 dependencies = [
  "async-trait",
  "chrono",


tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -38,13 +38,14 @@ use self::utils::TryDecode;
...
+// #[deprecated(since = "0.7.0", note = "Please use the hybrid extractor instead")]
+// pub mod native;
 pub mod protocol_cache;
 pub mod token_analysis_cron;
 pub mod token_pre_processor;
 mod utils;
-pub mod vm;
-
+// #[deprecated(since = "0.7.0", note = "Please use the hybrid extractor instead")]
+// pub mod vm;
```
kayibal: I'd remove the code then as well - we can always get it back via git

tycho-core/src/storage.rs
```diff
@@ -161,7 +161,11 @@ pub trait ExtractionStateGateway {
     /// # Returns
     /// Ok if the corrsponding state was retrieved successfully, Err in
     /// case the state was not found.
-    async fn get_state(&self, name: &str, chain: &Chain) -> Result<ExtractionState, StorageError>;
+    async fn get_state(
+        &self,
+        name: &str,
+        chain: &Chain,
+    ) -> Result<(ExtractionState, Block), StorageError>;
```
kayibal: Would prefer if block was an attribute on `ExtractionState`

README.md
```diff
@@ -2,60 +2,441 @@
 
 ![Asteroid Belt](./assets/belt.webp)
 
-Tycho Indexer is a low-latency solution designed to track and update the state of onchain and offchain financial protocols in real time. With Tycho, clients can subscribe to a stream of state update messages or query historical states at any specific timestamp or block through an RPC interface.
+Low latency indexer for onchain and offchain financial protocols, Tycho allows clients to connect to a stream of delta
```
zizou0x: Is this a Github bug? It's already in `main` no?

louise-poole: Eish - I think this was a bad rebase. I'll fix

tycho-storage/migrations/2024-06-21_v0.7.1/up.sql
kayibal, line 1: Be careful with this migration, since it is very old, it may not be in order with previous ones and idk how diesel handles that.

tycho-storage/migrations/2024-06-21_v0.7.1/up.sql
```diff
@@ -0,0 +1,16 @@
...
+ALTER TABLE extraction_state
+ADD COLUMN "block_id" bigint REFERENCES "block"(id);
+
+-- populate existing extraction_state with oldest block_id
+WITH oldest_block AS (
+    SELECT id
+    FROM "block"
+    ORDER BY id ASC
+    LIMIT 1
+)
```
kayibal, line 11: This would break if we had multiple chains but ok maybe add a check here that all extractors are on the same chain else fail.

tycho-storage/src/postgres/orm.rs
```diff
@@ -65,6 +65,9 @@ pub struct ExtractionState {
     /// Chain identifier that the extractor instance is scoped to.
     pub chain_id: i64,
 
+    /// Block identifier associated with the extraction state.
```
kayibal, line 68: I am sorry but this comment it quite useless, why is the block attached with extraction state, what does it mean? (Afaik last block processed by the extractor)

tycho-storage/src/postgres/orm.rs
```diff
@@ -171,6 +177,14 @@ impl Block {
             .await
     }
 
+    pub async fn by_db_id(id: i64, conn: &mut AsyncPgConnection) -> QueryResult<Block> {
```
kayibal, line 180: Where is this used?

tycho-storage/migrations/2024-06-21_v0.7.1/up.sql
louise-poole: I double checked - they don't clash with any of the migrations marked as newer. After merging the migration was correctly run which is good. Though lesson learnt: be careful with date/order of migrations you're about to merge.

tycho-storage/src/postgres/orm.rs
louise-poole: It's not - I'll remove it

PR #290, opened 2024-06-26
feat: Add pg_cron extension in the migration

zizou0x:
This PR adds a migration to add `pg_cron` extension and a cronjob that will run the `pg_partman` maintenance every day.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -13,13 +13,18 @@ jobs:
     strategy:
       matrix:
         include:
-          - target: aarch64-apple-darwin
+          - package: tycho-client
+            target: aarch64-apple-darwin
             os: macos-latest
-          - target: x86_64-unknown-linux-gnu
+          - package: tycho-client
+            target: x86_64-unknown-linux-gnu
             os: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -13,33 +13,78 @@ env:
...
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build PostgreSQL Docker Image
+        if: steps.cache-docker-image.outputs.cache-hit != 'true'
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: postgres.Dockerfile
+          tags: custom-postgres-image:latest
+          outputs: type=docker,dest=/tmp/custom-postgres-image.tar
```
tvinagre: Can you be more precise on the name? I know it's an internal use, but if a new developer sees this he will have a hard time understanding what's `custom-postgres-image`. So maybe call it `postgres-partman-pgcron`

zizou0x: I think `postgres-partman-pgcron` can easily become outdated. I called it extended-postgres and added some comment to make it less shady

PR #295, opened 2024-06-27
RPC bugfixes

zizou0x:
This PRs fixes two bugs in the RPC for protocol components:

1. Cached (new) protocol components aren't filtered by system, so always returned no matter what the requested system name is
2. Protocol type name isn't correctly returned when fetching protocol components in the database

### DIFF
--- a/tycho-indexer/src/services/deltas_buffer.rs
+++ b/tycho-indexer/src/services/deltas_buffer.rs
@@ -271,10 +271,7 @@ impl PendingDeltas {
         }
 
         if let Some(system) = protocol_system {
-            new_components = new_components
-                .into_iter()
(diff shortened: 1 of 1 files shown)


tycho-indexer/src/services/deltas_buffer.rs
```diff
@@ -202,7 +202,11 @@ impl PendingDeltas {
         Ok(change_found)
     }
 
-    pub fn get_new_components(&self, ids: Option<&[&str]>) -> Result<Vec<ProtocolComponent>> {
+    pub fn get_new_components(
```
tvinagre, line 225: need doc to understand what happens if user provides both `ids` and `protocol_system`

PR #296, opened 2024-07-09
Reduce token cache retention time

tvinagre:


tycho-indexer/src/services/rpc.rs
```diff
@@ -121,20 +121,20 @@ impl From<evm::ProtocolComponent> for dto::ProtocolComponent {
...
 impl<G> RpcHandler<G>
 where
     G: Gateway,
 {
     pub fn new(db_gateway: G, pending_deltas: PendingDeltas) -> Self {
-        let cache = Cache::builder()
+        let token_cache = Cache::builder()
             .max_capacity(50)
-            .time_to_live(std::time::Duration::from_secs(24 * 60 * 60))
+            .time_to_live(std::time::Duration::from_secs(12))
```
kayibal: That is very low, no? Also if we keep this, the value is very Ethereum specific and might not work as expected eg on Arbitrum.

tvinagre: Oh yeah, my bad. I had 10s in mind instead of minutes 🫠

tycho-indexer/src/services/rpc.rs
```diff
@@ -121,20 +121,20 @@ impl From<evm::ProtocolComponent> for dto::ProtocolComponent {
...
 impl<G> RpcHandler<G>
 where
     G: Gateway,
 {
     pub fn new(db_gateway: G, pending_deltas: PendingDeltas) -> Self {
-        let cache = Cache::builder()
+        let token_cache = Cache::builder()
             .max_capacity(50)
-            .time_to_live(std::time::Duration::from_secs(24 * 60 * 60))
+            .time_to_live(std::time::Duration::from_secs(7 * 12))
```
zizou0x:
```suggestion
            .time_to_live(std::time::Duration::from_secs(7 * 60))
```

PR #297, opened 2024-07-11
chore: Add trace logging for tokens queries

zizou0x:


tycho-indexer/src/services/rpc.rs
```diff
@@ -413,6 +413,8 @@ where
             .get_tokens_inner(chain, request)
             .await?;
 
+        trace!(received_tokens=?response.tokens.len(), "Retrieved tokens from DB");
```
louise-poole: Would prefer it be called `n_tokens_received` so that it's clear we're talking about the number of tokens.

PR #298, opened 2024-07-17
Fix new component snapshots returning empty

louise-poole:
Problem:
When snapshots were requested for new components (ie. exists in the buffer only and not in the db yet), we would return an empty snapshot. This is because we were only applying the buffered deltas to existing states returned from the db. For new components, there is no db state and so the buffered states were also ignore.

Fix: 
If we request a component whose state isn't in those returned from the db, create a new empty state and apply the buffered deltas to that.

### DIFF
--- a/tycho-indexer/src/services/deltas_buffer.rs
+++ b/tycho-indexer/src/services/deltas_buffer.rs
@@ -123,14 +123,15 @@ impl PendingDeltas {
     ///   fetched.
     pub fn merge_native_states(
         &self,
-        protocol_ids: Option<Vec<&str>>,
+        protocol_ids: Option<&[&str]>,
         db_states: &mut Vec<ProtocolComponentState>,
         version: Option<BlockNumberOrTimestamp>,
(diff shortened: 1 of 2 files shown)


tycho-indexer/src/services/deltas_buffer.rs
```diff
@@ -112,14 +112,41 @@ impl PendingDeltas {
...
+    /// Arguments:
+    ///
+    /// * `protocol_ids`: A list of the requested protocol ids. Note: `None` is not supported yet.
+    /// * `db_states`: A mutable reference to the states fetched from the db.
+    /// * `version`: The version of the state to be fetched. If `None`, the latest state will be
+    ///   fetched.
+    pub fn merge_native_states(
         &self,
-        db_states: &mut [ProtocolComponentState],
+        protocol_ids: Option<Vec<&str>>,
```
zizou0x: Could this be an array instead of `Vec`?

PR #299, opened 2024-07-25
Implement EVMContractExtractor

tvinagre:
Struct that is able to fetch an Account's data from a node connection

### DIFF
--- a/tycho-indexer/src/extractor/evm/contract.rs
+++ b/tycho-indexer/src/extractor/evm/contract.rs
@@ -1,13 +1,13 @@
 use std::collections::HashMap;
 
 use async_trait::async_trait;
+use chrono::NaiveDateTime;
 use ethers::{
     middleware::Middleware,
     prelude::{BlockId, Http, Provider, H160, H256, U256},
 };
 use serde::{Deserialize, Serialize};
 use tracing::trace;
-
 use tycho_core::{
     models::{Address, Chain, ChangeType},
     Bytes,
@@ -31,6 +31,19 @@ pub trait AccountExtractor {
 pub struct EVMAccountExtractor {
     provider: Provider<Http>,
     chain: Chain,
+}
+
+impl<TX> From<ethers::core::types::Block<TX>> for Block {
+    fn from(value: ethers::core::types::Block<TX>) -> Self {
+        Block {
+            number: value.number.unwrap().as_u64(),
+            hash: value.hash.unwrap(),
+            parent_hash: value.parent_hash,
+            chain: Chain::Ethereum,
+            ts: NaiveDateTime::from_timestamp_opt(value.timestamp.as_u64() as i64, 0)
+                .expect("Failed to convert timestamp"),
+        }
+    }
 }
 
 #[async_trait]
@@ -83,7 +96,7 @@ impl AccountExtractor for EVMAccountExtractor {
 
 impl EVMAccountExtractor {
     #[allow(dead_code)]
-    async fn new(node_url: &str, chain: Chain) -> Result<Self, RPCError>
+    pub async fn new(node_url: &str, chain: Chain) -> Result<Self, RPCError>
     where
         Self: Sized,
     {
@@ -105,7 +118,7 @@ impl EVMAccountExtractor {
         loop {
             let params = serde_json::json!([
                 block, 0, // transaction index, 0 for the state at the end of the block
-                address, start_key, 1024 // limit
+                address, start_key, 2147483647 // limit
             ]);
 
             trace!("Requesting storage range for {:?}, block: {:?}", address, block);
@@ -127,6 +140,15 @@ impl EVMAccountExtractor {
         }
 
         Ok(all_slots)
+    }
+
+    pub async fn get_block_data(&self, block_id: i64) -> Result<Block, RPCError> {
+        let block = self
+            .provider
+            .get_block(BlockId::from(u64::try_from(block_id).expect("Invalid block number")))
+            .await?
+            .expect("Block not found");
+        Ok(Block::from(block))
     }
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5895,9 +5895,9 @@ dependencies = [
 
 [[package]]
 name = "time"
-version = "0.3.34"
+version = "0.3.36"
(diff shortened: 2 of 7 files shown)


tycho-indexer/src/extractor/evm/contract.rs
```diff
@@ -0,0 +1,222 @@
...
+use serde::{Deserialize, Serialize};
+use std::{collections::HashMap, error::Error};
+use tracing::trace;
+use tycho_core::{
+    models::{Chain, ChangeType},
+    Bytes,
+};
+
+pub struct ContractExtractor {
+    addresses_per_block: HashMap<u64, Vec<H160>>,
```
zizou0x: IIRC we have an alias for address `Address`

tycho-indexer/src/extractor/evm/contract.rs
```diff
@@ -0,0 +1,222 @@
...
+}
+
+#[cfg_attr(test, mockall::automock)]
+#[async_trait]
+pub trait EVMContractExtractor {
+    async fn new(
+        node_url: &str,
+        chain: Chain,
+        address_block_pairs: Vec<(H160, u64)>,
+    ) -> Result<Self, Box<dyn std::error::Error>>
```
zizou0x: I don't really like the dynamic dispatch here. I think we can either make it static or introduce a trait type placeholder

tycho-indexer/src/extractor/evm/contract.rs
```diff
@@ -0,0 +1,222 @@
...
+
+        match self
+            .addresses_per_block
+            .get(&block.number)
+        {
+            Some(addresses) => {
+                for address in addresses {
+                    trace!(contract=?address, block_number=?block.number, block_hash=?block.hash, "Extracting contract code and storage" );
+                    let block_id = Some(BlockId::from(block.number));
+                    println!("Block Id: {:?}", block_id);
```
zizou0x: I guess this was for debugging, we need to remove them

tycho-indexer/src/extractor/evm/contract.rs
```diff
@@ -0,0 +1,222 @@
...
+
+pub struct ContractExtractor {
+    addresses_per_block: HashMap<u64, Vec<H160>>,
+    provider: Provider<Http>,
+    chain: Chain,
+}
+
+#[cfg_attr(test, mockall::automock)]
+#[async_trait]
+pub trait EVMContractExtractor {
```
zizou0x: Maybe if we use something more flexible for addresses, we can make this trait not EVM tailored? 

I'm just afraid about `AccountUpdate`. It's kind of EVM tailored too, we might need to use generics here, I don't know if you want to go into that now or leave it for when we will try to implement a non-evm chain

tycho-indexer/src/extractor/evm/contract.rs
```diff
@@ -0,0 +1,222 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::str::FromStr;
+
+    #[tokio::test]
+    #[ignore]
```
zizou0x:
```suggestion
    #[ignore = "require RPC connection"]
```

tycho-indexer/src/extractor/evm/contract.rs
```diff
@@ -0,0 +1,222 @@
...
+    // This test requires a real RPC URL
+    async fn test_contract_extractor() -> Result<(), Box<dyn std::error::Error>> {
```
zizou0x, line 176: Could we have a test with a mocked RPC? So we keep testing this in CI

tycho-indexer/src/extractor/evm/contract.rs
kayibal: I think this interface isn't very good because it is too complex too early. Also since this will be an interface used by the DCI and we want the DCI to not depend on a single chain, we need to avoid chain specific types in the interface.

So I would suggest the following concrete steps:

- Do not dictate how the struct is constructed -> remove the "new" from the interface
- Simplify by avoiding any state on this trait -> remove register_contracts method
- Just have a single method that retrieves state for a set of addresses at a certain block, the addresses should be passed as Bytes - I currently don't understand why you decided for a statesful approach?

tycho-indexer/src/extractor/evm/contract.rs
tvinagre: Ah yeah - tnx

tycho-indexer/src/extractor/evm/contract.rs
tvinagre: Yeah me neither - will polish

tycho-indexer/src/extractor/evm/contract.rs
tvinagre: I've decided for it because I wanted to encapsulate the map between addresses and block_id. This would make this extractor behave similarly to the HybridExtractor, and would allow us to extract contracts on different blocks. 
But this feature can be handled outside the extractor, so I'll make it stateless

tycho-indexer/src/extractor/evm/contract.rs
tvinagre: Yeah I've tried but it became a mess with the Types, so I've decided to do it later, as we really need to ship this

tycho-indexer/src/extractor/evm/contract.rs
```diff
@@ -0,0 +1,190 @@
...
+pub struct EVMAccountExtractor {
+    provider: Provider<Http>,
+    chain: Chain,
+}
+
+#[async_trait]
+impl AccountExtractor for EVMAccountExtractor {
+    async fn get_accounts(
+        &self,
+        block: Block,
```
zizou0x, line 53: Do we need the full block or can we just take a `BlockIdentifier`?

tvinagre: Unfortunately, we need the whole `Block` 
`get_balance` and `get_code` use the block number, while `get_storage_range` only works with block hash

tvinagre: Some providers take both, but our archive node only works this way, so it's the safest

PR #300, opened 2024-07-25
Use EVMContractExtractor on tycho-indexer

tvinagre:


tycho-indexer/src/extractor/evm/contract.rs
```diff
@@ -25,7 +29,7 @@ pub trait EVMContractExtractor {
         node_url: &str,
         chain: Chain,
         address_block_pairs: Vec<(H160, u64)>,
-    ) -> Result<Self, Box<dyn std::error::Error>>
```
zizou0x: Nice

tycho-indexer/src/extractor/evm/contract.rs
```diff
@@ -59,7 +63,7 @@ impl EVMContractExtractor for ContractExtractor {
                     });
                 Ok(Self { addresses_per_block, provider: p, chain })
             }
-            Err(e) => Err(Box::new(e)),
+            Err(e) => Err(ExtractionError::Setup(e.to_string())),
```
zizou0x: I don't think this error type will stay accurate once we have DCI. Maybe we can introduce a new one for RPCs?

zizou0x: Also you can easily clean the code above with `Provider::<Http>::try_from(node_url).map_err(...)?;`

tycho-indexer/src/main.rs
```diff
@@ -287,9 +291,37 @@ async fn build_all_extractors(
...
+
+        if !(extractor_config
+            .initialized_accounts
+            .is_empty())
+        {
+            let register_contracts: Vec<(H160, u64)> = extractor_config
+                .initialized_accounts
+                .clone()
+                .into_iter()
+                .map(|a| (H160::from_str(&a).expect("Unable to convert address into H160"), block))
```
zizou0x: Small detail but I tend to prefer being really explicit on panic. You can probably do that `.unwrap_or_else(|| panic!("Unable to convert {&a} into H160"));` or `.expect(&format!("Unable to convert {} into H160", &a))` but IIRC clippy complains about the later

tycho-indexer/src/main.rs
```diff
@@ -287,9 +291,37 @@ async fn build_all_extractors(
     );
     protocol_cache.populate().await?;
 
+    let mut contract_extractor =
+        ContractExtractor::new(rpc_url, *chains.first().unwrap(), vec![]).await?;
+
```
zizou0x: This looks wrong that we create this outside of the for loop, looks like it'll be shared for every extractor?

tycho-indexer/src/extractor/runner.rs
```diff
@@ -276,11 +277,12 @@ pub struct ExtractorConfig {
     chain: Chain,
     implementation_type: ImplementationType,
     sync_batch_size: usize,
-    start_block: i64,
+    pub start_block: i64,
     stop_block: Option<i64>,
     protocol_types: Vec<ProtocolTypeConfig>,
     spkg: String,
     module_name: String,
+    pub initialized_accounts: Vec<String>,
```
zizou0x: I think this is going to break things... I don't think it should be part of this struct unless we want to add them in the `extractors.yaml` but I don't see any usecase for it

zizou0x: Basically this struct need to match what's in the `extractor.yaml`. I realise that it's not written anywhere so if you could add a comment explaining this, would be perfect

tycho-indexer/src/extractor/evm/contract.rs
```diff
@@ -12,6 +15,7 @@ use tycho_core::{
     Bytes,
 };
 
+#[derive(Clone, Debug)]
```
zizou0x: Should it be clone?

tycho-indexer/src/extractor/evm/hybrid.rs
```diff
@@ -478,6 +483,45 @@ where
             .collect();
         Ok(new_tokens)
     }
+
+    async fn load_related_contracts(&self, msg: BlockChanges) -> BlockChanges {
```
kayibal: Idk what you mean by `related_contracts`? I think this should be `retrieve_pre_initialize_accounts` or similar.

tycho-indexer/src/extractor/evm/hybrid.rs
```diff
@@ -585,6 +629,8 @@ where
             Err(e) => return Err(e),
         };
 
+        let msg = self.load_related_contracts(msg).await;
```
kayibal: Wait you do this every block that we process? I think we should only do this once in the beginning.

tycho-indexer/src/extractor/runner.rs
kayibal: I think this can be useful here. I would keep it. I am not quite happy with the types here though.. why is this a vector of strings - addresses have always been Bytes. Apart from that I think we should allow to also specify at what block you want the account to be initialised at here.

tycho-indexer/src/main.rs
```diff
@@ -287,9 +291,37 @@ async fn build_all_extractors(
...
     for extractor_config in config.extractors.values() {
+        let block: u64 = extractor_config.start_block as u64;
```
kayibal: Hm that is a weird assumption the application makes, no?

tycho-indexer/src/extractor/evm/hybrid.rs
tvinagre: I've tried to make it as agnostic as possible. 
So in this case, it can receive a list of addresses and target blocks - on every block we check if there are any contracts to be extracted, if not, skip - O(1).

tycho-indexer/src/extractor/runner.rs
tvinagre: Ah - the types is a mistake. I was afraid it would be a pain to work with `bytes` so I did with string but forgot to add a comment to remove it - I'll fix

tvinagre: > Apart from that I think we should allow to also specify at what block you want the account to be initialised at here.

I agree - in case a user wants to initialize in a block different from `start_block`

tycho-indexer/src/main.rs
tvinagre: Yeah I'll add another param

tycho-indexer/src/main.rs
tvinagre: No you're right - should be one per-extractor

tycho-indexer/src/extractor/evm/hybrid.rs
kayibal: agnostic to what? I am not sure what case you are optimising for here but I think it is unnecessarily complex to do it this way. Like there is no point in putting this here other than being too "lazy" to insert the corresponding block at the start...

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -85,6 +85,18 @@ pub struct Block {
     pub ts: NaiveDateTime,
 }
 
+impl<TX> From<ethers::core::types::Block<TX>> for Block {
```
kayibal: Can you put this close to the EVMContractExtractor please? This way ethers dependency is further contained.

tycho-indexer/src/main.rs
```diff
@@ -301,6 +321,93 @@ async fn build_all_extractors(
...
+    rpc_url: &str,
+    chain: Chain,
+    cached_gw: &CachedGateway,
+) {
+    let (block, extracted_accounts) = get_accounts_data(accounts, block_id, rpc_url, chain).await;
+
+    info!(block_number = block.number, "Initializing accounts");
+
+    let tx = Transaction {
+        hash: H256::zero(),
```
kayibal: Idk - putting zero here is kind of dangerous, no? Don't we have a unique key constraint on this column?

tycho-indexer/src/main.rs
```diff
@@ -301,6 +321,93 @@ async fn build_all_extractors(
...
+    rpc_url: &str,
+    chain: Chain,
+) -> (Block, HashMap<H160, AccountUpdate>) {
+    let provider = Provider::<Http>::try_from(rpc_url).expect("Failed to create provider");
+    let block_data = provider
+        .get_block(BlockId::from(u64::try_from(block_id).expect("Invalid block number")))
+        .await
+        .expect("Failed to get block")
+        .expect("Block not found");
+
```
kayibal: I'd probably put this behind the `EVMAccountExtractor` as well so ethers-rs is better contained.

tycho-indexer/src/main.rs
```diff
@@ -301,6 +321,93 @@ async fn build_all_extractors(
     Ok(extractor_handles)
 }
 
+async fn initialize_accounts(
```
kayibal, line 324: Could you instrument this please? I think only block id and number of accounts are of interest here.

tycho-indexer/src/main.rs
```diff
@@ -545,6 +545,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[ignore = "requires DB connection"]
```
kayibal: We don't have a db connection on the CI?

tycho-indexer/src/main.rs
```diff
@@ -338,3 +445,125 @@ async fn run_token_analyzer(
...
     Ok(())
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use tycho_storage::postgres::testing::run_against_db;
+
+    #[tokio::test]
+    #[ignore = "require archive node (RPC) and DB connection"]
```
kayibal: Can you remove the DB connection reason here and in the other two below pls?

PR #304, opened 2024-08-02
feat(tycho-client-py): Add a tycho-client python package.

kayibal:
This commit adds support for connecting to a tycho stream of block synchronized protocol updates.

### DIFF
--- a/tycho-client-py/tycho_client/dto.py
+++ b/tycho-client-py/tycho_client/dto.py
@@ -187,6 +187,14 @@ class ResponseAccount(BaseModel):
     class Config:
         allow_population_by_field_name = True
 
+    @property
+    def balance(self) -> HexBytes:
+        return self.native_balance
+
+    @property
+    def change(self) -> ChangeType:
+        return ChangeType.creation
+
 
 class Snapshot(BaseModel):
     states: Dict[str, ComponentWithState] = Field(default_factory=dict)


tycho-client-py/tycho_client/dto.py
```diff
@@ -0,0 +1,223 @@
...
+    creation = "Creation"
+    unspecified = "Unspecified"
+
+
+class ExtractorIdentity(BaseModel):
+    chain: Chain
+    name: str
+
+    @classmethod
+    def new(cls, chain: Chain, name: str) -> "ExtractorIdentity":
```
tvinagre: I'm curious why did you create this if it's already a Pydantic Model?

tycho-client-py/tycho_client/stream.py
```diff
@@ -0,0 +1,148 @@
...
+from pydantic import ValidationError
+
+from .dto import Chain, FeedMessage
+from .exception import TychoStreamException
+
+log = getLogger(__name__)
+
+
+class TychoStream:
+    # we allow to initialise this class with some preset PoolStates for use cases such as DAIPSM that do not need
```
tvinagre: I think this comment is not needed here - or we need to add back the `pool_state` parameter.

tycho-client-py/tycho_client/dto.py
kayibal: Uhps that was chatgpt

tycho-client-py/tycho_client/rpc_client.py
```diff
@@ -0,0 +1,119 @@
...
+
+        res = self._post_request(f"/v1/{self._chain}/tokens", body=body, params={})
+        return [ResponseToken(**t) for t in res["tokens"]]
+
+
+if __name__ == "__main__":
+    client = TychoRPCClient("http://localhost:4242")
+    print(client.get_protocol_components("uniswap_v2"))
+    print(client.get_protocol_state(protocol_system="vm:balancer"))
+    print(client.get_contract_state(include_balances=True))
```
zizou0x: This looks like a test logic, should it stay?

kayibal: Probably useful to quickly check tings

PR #308, opened 2024-08-05
fix: add defaults for optional extractor configs

louise-poole:
Tested - works like a charm ✨

### DIFF
--- a/extractors.yaml
+++ b/extractors.yaml
@@ -32,7 +32,7 @@ extractors:
     protocol_types:
       - name: "sushiswap_v2_pool"
         financial_type: "Swap"
-    spkg: "substreams/ethereum-sushiswap-v2/ethereum-sushiswap-v2-v0.2.0.spkg"
+    spkg: "substreams/ethereum-uniswap-v2/ethereum-sushiswap-v2-v0.2.0.spkg"
     module_name: "map_pool_events"
 
   uniswap_v3:


extractors.yaml
```diff
@@ -32,7 +32,7 @@ extractors:
     protocol_types:
       - name: "sushiswap_v2_pool"
         financial_type: "Swap"
-    spkg: "substreams/ethereum-uniswap-v2/ethereum-sushiswap-v2-v0.2.0.spkg"
+    spkg: "substreams/ethereum-sushiswap-v2/substreams-ethereum-sushiswap-v2-v0.2.0.spkg"
```
zizou0x: We usually name them like this in S3, I'd like to keep the naming consistent

```suggestion
    spkg: "substreams/ethereum-sushiswap-v2/ethereum-sushiswap-v2-v0.2.0.spkg"
```

kayibal: Why do we need a different package here?

tycho-indexer/src/extractor/runner.rs
```diff
@@ -281,10 +281,16 @@ pub struct ExtractorConfig {
     protocol_types: Vec<ProtocolTypeConfig>,
     spkg: String,
     module_name: String,
+    #[serde(default)]
     pub initialized_accounts: Vec<Bytes>,
+    #[serde(default = "default_initialized_accounts_block")]
```
zizou0x: Can't we just `serde(default)` this one too? I'd expect `i64` to default to 0

extractors.yaml
louise-poole: Because the package has sushiswap specific info (factory address etc)

PR #311, opened 2024-08-06
feat: Create remove protocol script

louise-poole:
Added since the last reviews:
- reminder to make a db snapshot before proceeding with the deletion
- generates an audit file so you can manually verify the components and accounts that will be deleted before proceeding
- skips deleting accounts used by other protocol systems

### DIFF
--- a/tycho-storage/src/postgres/mod.rs
+++ b/tycho-storage/src/postgres/mod.rs
@@ -396,10 +396,7 @@ async fn maybe_lookup_version_ts(
     conn: &mut AsyncPgConnection,
 ) -> Result<NaiveDateTime, StorageError> {
     if !matches!(version.1, VersionKind::Last) {
-        return Err(StorageError::Unsupported(format!(
-            "Unsupported version kind: {:?}",
-            version.1
-        )));
+        return Err(StorageError::Unsupported(format!("Unsupported version kind: {:?}", version.1)));
     }
     maybe_lookup_block_ts(&version.0, conn).await
 }


tycho-storage/sql_scripts/remove_protocol_system.sql
```diff
@@ -0,0 +1,29 @@
...
+--- code entries too
+DELETE FROM account
+WHERE id IN (
+    SELECT cc.account_id
+    FROM contract_code cc
+    JOIN protocol_component_holds_contract pchc ON pchc.contract_code_id = cc.id
+    JOIN protocol_component pc ON pchc.protocol_component_id = pc.id
+    JOIN protocol_system ps ON pc.protocol_system_id = ps.id
+    WHERE ps.name = :'protocol_system_name'
+);
```
kayibal: What if an account is used by two components of two different systems?

tycho-storage/sql_scripts/remove_protocol_system.sql
```diff
@@ -0,0 +1,29 @@
+--- Removes a protocol system and all related entries from the database.
```
zizou0x: I would add a big warning in the first line that if an account is shared between many protocol systems, this script should not be used because it would delete this account.

We might need something to link accounts to components that use it

zizou0x: Were you able to test this somehow?

louise-poole: Yes, tested it locally by running tycho for 100 blocks, then deleting a native protocol and verifying all the necessary info was removed from the db. Did the same for a vm protocol.

PR #313, opened 2024-08-07
fix(tycho-client-py): Change tycho client boolean serialization

tvinagre:
Comments in the PR

### DIFF
--- a/tycho-client-py/tycho_client/dto.py
+++ b/tycho-client-py/tycho_client/dto.py
@@ -187,6 +187,14 @@ class ResponseAccount(BaseModel):
     class Config:
         allow_population_by_field_name = True
 
+    @property
+    def balance(self) -> HexBytes:
+        return self.native_balance
+
+    @property
+    def change(self) -> ChangeType:
+        return ChangeType.creation
+
 
 class Snapshot(BaseModel):
     states: Dict[str, ComponentWithState] = Field(default_factory=dict)


tycho-client-py/tycho_client/dto.py
```diff
@@ -30,6 +30,9 @@ def validate(cls, v: str):
 
         return cls(data)
 
+    def __int__(self):
```
tvinagre, line 33: `int(hex_value, 16)` doesn't work if the object is `HexBytes`. With this, user can convert to int by doing `int(hex_value)`. 
The other solution is to require user to do `int(hex_value.hex(), 16)`

tycho-client-py/tycho_client/dto.py
```diff
@@ -241,26 +245,38 @@ class VersionParams(BaseModel):
 
 class ProtocolComponentsParams(BaseModel):
     protocol_system: Optional[str] = None
-    component_addresses: Optional[List[HexBytes]] = None
+    component_addresses: Optional[List[HexBytes]] = Field(
+        default=None, alias="componentAddresses"
```
tvinagre, line 257: This is temporary, to fix lack of coherence between the parameter names.

tycho-client-py/tycho_client/rpc_client.py
```diff
@@ -29,28 +31,34 @@ def __init__(
...
-        self, endpoint: str, params: dict = None, body: dict = None
+            self, endpoint: str, params: dict = None, body: dict = None
     ) -> dict:
         """Sends a POST request to the given endpoint and returns the response."""
+
+        # Convert to JSON strings to cast booleans to strings
+        if body is not None:
+            body = json.dumps(body)
+        if params:
+            params = json.dumps(params)
```
tvinagre, line 42: We had requests failing because, by using Dict, the request would be serialized to `protocol_state?include_balances=True`. This upper case in the boolean value makes the request invalid.

tycho-client-py/tycho_client/rpc_client.py
```diff
@@ -29,28 +31,34 @@ def __init__(
...
+
         response = requests.post(
             self.rpc_url + endpoint,
             headers=self._headers,
-            json=body or {},
+            data=body or "{}",
```
kayibal, line 47: I think you should default earlier to empty dict, ideally directly at the start of the method. It's minor though but I don't like this string json literals in the code.

tycho-client-py/tycho_client/rpc_client.py
```diff
@@ -19,7 +21,7 @@ class TychoRPCClient:
     """
 
     def __init__(
-        self, rpc_url: str = "http://0.0.0.0:4242", chain: Chain = Chain.ethereum
+            self, rpc_url: str = "http://0.0.0.0:4242", chain: Chain = Chain.ethereum
```
zizou0x, line 24: Is this Black formatting?

PR #323, opened 2024-08-13
feat: Specify extractor on rpc requests

louise-poole:
Fixes finality bug when extractors are not on the same block - we used to use a random extractor to determine block finality (very risky when an extractor is syncing)

**Assumption**: extractor name and protocol system name are the same thing.

### DIFF
--- a/tycho-indexer/src/services/deltas_buffer.rs
+++ b/tycho-indexer/src/services/deltas_buffer.rs
@@ -655,7 +655,9 @@ mod test {
             .get_new_components(None, None)
             .unwrap();
 
-        assert_eq!(new_components, exp);
+        for expected in &exp {
+            assert!(new_components.contains(expected));
+        }
 
         let new_components = buffer
             .get_new_components(Some(&["component3"]), None)


tycho-indexer/src/services/deltas_buffer.rs
```diff
@@ -307,29 +307,50 @@ impl PendingDeltas {
...
+                        .map_err(|e| PendingDeltasError::LockError(system, e.to_string()))?;
+                    Ok(guard.get_finality_status(version))
+                } else if let Some(buffer) = self.vm.get(&system) {
+                    let guard = buffer
+                        .lock()
+                        .map_err(|e| PendingDeltasError::LockError(system, e.to_string()))?;
+                    Ok(guard.get_finality_status(version))
+                } else {
+                    Err(PendingDeltasError::UnknownExtractor(system))
+                }
```
kayibal: Interesting seems like we forgot to merge the buffers here? Since we only have a hybrid extractor, can we simplify this maybe?

louise-poole: True!

PR #324, opened 2024-08-13
refactor(rpc): remove unused params and endpoint

zizou0x:
This PR removes deltas endpoints and some unused/old params for our contracts endpoint.

Our motivation for this is to keep Tycho RPC as simple and straightforward as possible.

Also this PR make our RPC return an error message if an unknown field is passed in the body. This is to prevent typo mistakes. 

The returned message looks like this: 

```
Json deserialize error: unknown field `contract_idss`, expected one of `contractIds`, `contract_ids`, `version` at line 2 column 17
```

### DIFF
--- a/tycho-core/src/dto.rs
+++ b/tycho-core/src/dto.rs
@@ -958,7 +958,6 @@ mod test {
                 "timestamp": "2069-01-01T04:20:00",
                 "block": {
                     "hash": "0x24101f9cb26cd09425b52da10e8c2f56ede94089a8bbe0f31f1cda5f4daa52c4",
-                    "parentHash": "0x8d75152454e60413efe758cc424bfd339897062d7e658f302765eb7b50971815",
                     "number": 213,
                     "chain": "ethereum"
                 }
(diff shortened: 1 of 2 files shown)


tycho-core/src/models/mod.rs
```diff
@@ -181,9 +169,8 @@ pub enum ChangeType {
     Creation,
 }
 
-#[derive(Debug, Serialize, Deserialize, PartialEq, Clone, ToSchema)]
+#[derive(Debug, Serialize, Deserialize, PartialEq, Clone)]
 pub struct ContractId {
-    #[schema(value_type=String)]
     pub address: Address,
```
louise-poole, line 174: Why remove this? Isn't it needed? From what I remember, `Address` is `Bytes`?

zizou0x: I noticed that we were still using structs from this file instead of dto in our RPC. So I removed DTO related things from here and added them in the dto.rs file and changed our RPC to use only dto structs

PR #325, opened 2024-08-13
docs(dci): Sketch DCI interfaces

tvinagre:
mermaid-diagram-2024-08-19-011834


tycho-indexer/src/extractor/evm/contract.rs
```diff
@@ -1,23 +1,79 @@
...
+        &self,
+        inp: BlockUndoSignal,
+    ) -> Result<Option<ExtractorMsg>, ExtractionError>;
+}
+
+pub struct DynamicContractExtractorImpl {
+    account_extractor: Box<dyn AccountExtractor>,
+    tracked_contracts: Vec<Address>,
+    // TODO: Make PG Gateway generic and remove "Hybrid" from the name
+    hybrid_pg_gateway: HybridPgGateway,
```
louise-poole: This isn't part of this project right? This looks more like an open-sourcing task?

tycho-indexer/src/extractor/evm/dynamic_account_indexer.rs
```diff
@@ -0,0 +1,63 @@
...
+//         Attribute {
+//             name: "dci_entrypoint_function_signature_0".to_string(),
+//             value: CONTRACT_SIGNATURE.to_vec(),
+//             change: ChangeType::Creation.into(),
+//         }
+//         Attribute {
+//             name: "dci_entrypoint_function_parameter_0_0".to_string(),
+//             value: CONTRACT_PARAMETERS[0].to_vec(),
+//             change: ChangeType::Creation.into(),
+//         }
```
louise-poole: Reminder about what we discussed:
```suggestion
//         Attribute {
//             name: "dci_entrypoint_function_parameter_0_0_0_0".to_string(),
//             value: CONTRACT_PARAMETERS[0][0][0].to_vec(),
//             change: ChangeType::Creation.into(),
//         }
```
`dci_entrypoint_function_parameter_{address}_{function}_{parameter}_{case}` <-- might be nice to add this sort of a breakdown to the docs. Makes it much easier to understand.

louise-poole: I can't shake the feeling that we're forcing things here too much trying to use Attributes... what is the down side to introducing a new struct that will handle these nested values better?

tycho-indexer/src/extractor/evm/contract.rs
tvinagre: yes

tycho-indexer/src/extractor/evm/dynamic_account_indexer.rs
tvinagre: Nice! Will add

tvinagre: Yes, I agree, I'd like to make a new structure, but would be way more effort (would need to touch pgadapters, extractors, new table on the db..

tvinagre: We can have an intermediary structure Substreams, but cast it to attributes before returning the final `BlockChanges` message

tycho-indexer/src/extractor/evm/dynamic_account_indexer.rs
```diff
@@ -0,0 +1,100 @@
...
+// attributes.
+// The goal of the DCI is to identify contracts that are accessed by a contract, so we can index
+// them. The list of contracts should be saved as a relation to our protocol components, and
+// propagated to DynamicContractExtractor, so it can extract the contract data.
+
+// DCI Specific EntityChanges. Each collection of dci_entrypoint_* attributes specify a contract
+// that will be used on analysis.
+//
+// Those attributes are, for example, an Oracle address, its interface and the parameters to call.
+// IDEA: Can't we skip this and use our Adapter contracts for this? If yes, ExternalAccountEntrypoint should be
```
kayibal: Please don't :D

tycho-indexer/src/extractor/evm/dynamic_account_indexer.rs
```diff
@@ -0,0 +1,100 @@
...
+//             change: ChangeType::Creation.into(),
+//         }
+//         NAME FORMAT: dci_entrypoint_function_parameter_{address}_{function}_{parameter}_{idx}
+//         // [0][1][3] -> 0x00000001232cow -> CowSwapCaller
+
+#[derive(Debug, Clone)]
+pub struct ExternalAccountEntrypoint {
+    pub address: Address,       // The address of the contract
+    pub signature: String,      // The name of the function and parameter types
+    pub parameters: Vec<Bytes>, // The parameters of the function
```
kayibal: What are these exactly? Is this set if the function is always called with the same paramters?

tycho-indexer/src/extractor/evm/dynamic_account_indexer.rs
```diff
@@ -0,0 +1,100 @@
...
+//         Attribute {
+//             name: "dci_retrigger_storage_key_0_1".to_string(),
+//             value: CONTRACT_ADDRESS.to_vec(),
+//             change: ChangeType::Creation.into(),
+//         },
+#[derive(Debug, Clone)]
+pub struct EntrypointRetriggers {
+    pub addresses: Vec<Address>,  // Collection of addresses
+    pub storage_keys: Vec<Bytes>, // Collection of storage keys that trigger re-analysis
+}
```
kayibal: This doesn't work... a Retrigger is a pair of address and storage keys, this doesn't work.... because you need to retrigger on any storage slot on any of the addresses.

tycho-indexer/src/extractor/evm/contract.rs
```diff
@@ -1,23 +1,79 @@
...
+    async fn handle_tick_scoped_data(&self, block: BlockScopedData) -> Result<(), RPCError>;
+
+    // In case of a revert in the block, we need to rollback the account updates. Otherwise we risk
+    // having inconsistent state in the db.
+    // Behaviour should be similar to HybridExtractor's implementation for `handle_revert` - but
+    // only care about AccountUpdates.
+    async fn handle_revert(
+        &self,
+        inp: BlockUndoSignal,
+    ) -> Result<Option<ExtractorMsg>, ExtractionError>;
```
kayibal, line 156: It seems like the DCI should also implement Extractor since you have this methods here? Also it will have to process substream Block messages.

tycho-indexer/src/extractor/evm/contract.rs
kayibal: So it seems you decided that you want to communicate between extractors via the database? I think this is a bad pattern, instead the extractors should request/subscribe to the DCI and then handle the data as we used to. But that is just my intuition I think you will have to think about some potential alternatives here.

tycho-indexer/src/extractor/evm/dynamic_account_indexer.rs
kayibal: Why would you need to touch pgadapter? Did we consider anything else here? I guess the main thing that changes if you encode sth into an attribute is that it would be potentially harder to extend it.

tycho-indexer/src/extractor/evm/dynamic_account_indexer.rs
tvinagre: Sorry, I explained in the call but didn't comment properly. 
It's an optional set of fixed parameters that can be added to the fuzz. 
Example, we have a function that, if called with an specific address, will call another contract. Fuzz would miss this address, as it tests random addresses. This allows us to mix the random values with ones that we know need to be tested. 
They are the `fixtures` on Foundry fuzz test

image

tycho-indexer/src/extractor/evm/contract.rs
tvinagre: Yes - I want to separate the traits, `Extractor` trait has too many functions IMO.

tvinagre: > it will have to process substream Block messages.

yes, will do on `handle_tick_scoped_data`

tycho-indexer/src/extractor/evm/contract.rs
tvinagre: Yes, because `Accounts` should be independent of `ProtocolSystems`. So this would be an `Account` specific extractor.
Also, I would like to avoid any dependency of the DCI on the `HybridExtractor` because most likely the DCI will be too slow, so adding this synchonization can make things super slow on the extractor

tycho-indexer/src/extractor/evm/contract.rs
```diff
@@ -1,23 +1,79 @@
...
+}
+
+pub struct DynamicContractExtractorImpl {
+    account_extractor: Box<dyn AccountExtractor>,
+    tracked_contracts: Vec<Address>,
```
kayibal: Might be valuable to know which protocol this contract belongs to... for timing and waiting purposes.

tycho-indexer/src/extractor/evm/dci_dto.rs
```diff
@@ -0,0 +1,81 @@
...
+//     parameters: vec![
+//     (0, vec!["0xa5547190e3d59f2bfeb4174ca3454b2f2acaeed644bc7ad7018014516f73f2bd".to_string()]),
+//     (1, vec!["0x9008D19f58AAbD9eD0D60971565AA8510560ab41".to_bytes()]),
+//     (2, vec!["0xe96a45f66bdDA121B24F0a861372A72E8889523d".to_bytes()]),
+//     ]
+//     .into_iter()
+//     .collect(),
+// };
+// Attribute {
+//     name: "dci_entrypoint_0".to_string(),
```
kayibal, line 49: Should we use a list here? Or maybe this could be a dict with keys sth like `dci_entrypoint_authorizer`

PR #327, opened 2024-08-16
fix: WebSocketMessage deserialisation bug

louise-poole:
The serde arbitrary_precision feature (often required by dependencies we use) breaks some untagged enum deserialisations (more details found [here](https://github.com/serde-rs/json/issues/740)). 
Solution: bypass directly deserialising the message into a WebSocketMessage. Instead deserialise it into a generic `Value` and convert that into a WebSocketMessage.

### DIFF
--- a/tycho-client/src/deltas.rs
+++ b/tycho-client/src/deltas.rs
@@ -453,11 +453,11 @@ impl WsDeltasClient {
                         }
                     },
                     Err(e) => {
-                        error!(error = %e, message=text, "Failed to deserialize Value into WebSocketMessage");
+                        error!(error = %e, message=text, "Failed to deserialize WebSocketMessage: message does not match expected structs");
                     }
                 },
                 Err(e) => {
-                    error!(error = %e, message=text, "Failed to deserialize message as serde_json::Value");
+                    error!(error = %e, message=text, "Failed to deserialize message: invalid json");
                 }
             },
             Ok(tungstenite::protocol::Message::Ping(_)) => {


tycho-client/src/deltas.rs
```diff
@@ -397,49 +397,67 @@ impl WsDeltasClient {
...
-
-                    let inner = guard
-                        .as_mut()
-                        .ok_or_else(|| DeltasError::NotConnected)?;
-                    inner.remove_subscription(subscription_id);
-                }
+                },
                 Err(e) => {
-                    error!(error = %e, message=text, "Failed to deserialize message");
+                    error!(error = %e, message=text, "Failed to deserialize message as serde_json::Value");
```
kayibal: I guess this would mean that the message is invalid json. Maybe the message could be a bit clearer about that?

tycho-client/src/deltas.rs
```diff
@@ -397,49 +397,67 @@ impl WsDeltasClient {
...
+                        }) => {
+                            info!(?subscription_id, "Received a subscription ended");
+                            let inner = guard
+                                .as_mut()
+                                .ok_or_else(|| DeltasError::NotConnected)?;
+                            inner.remove_subscription(subscription_id);
+                        }
+                    },
+                    Err(e) => {
+                        error!(error = %e, message=text, "Failed to deserialize Value into WebSocketMessage");
```
kayibal: Again this message could be clearer this case means it does not fit into the structure properly.

tycho-client/src/deltas.rs
```diff
@@ -397,49 +397,67 @@ impl WsDeltasClient {
...
+                                    .await;
+                                }
+                                _ => { /* Do nothing */ }
+                            }
                         }
-                        Err(_) => {
-                            warn!(?subscription_id, "Receiver for has gone away, unsubscribing!");
-                            let (tx, _) = oneshot::channel();
-                            let _ =
-                                WsDeltasClient::unsubscribe_inner(inner, subscription_id, tx).await;
```
kayibal, line 418: What happened to this case is it still here?

louise-poole: Yes - it's just nested 1 more level inside the match statement. It's the lines directly above here ☝️

PR #328, opened 2024-08-19
feat(tycho-client): Add tvl range as a component filter

louise-poole:
Issue: components whose tvl fluctuates around the tvl threshold we have set cause us to repeatedly add and remove the same component.
Solution: to buffer against this, we have introduced a tvl range threshold. Here you define a lower (removal) threshold and an upper (addition) threshold. The distance between the removal and the addition thresholds means a component has to experience a large movement in order to be considered low enough to remove/high enough to add.

### DIFF
--- a/tycho-client/src/feed/synchronizer.rs
+++ b/tycho-client/src/feed/synchronizer.rs
@@ -28,7 +28,7 @@ use tycho_core::{
 use super::Header;
 use crate::{
     deltas::{DeltasClient, SubscriptionOptions},
-    feed::component_tracker::{ComponentFilter, ComponentFilterVariant, ComponentTracker},
+    feed::component_tracker::{ComponentFilter, ComponentTracker},
     rpc::RPCClient,
 };
 
--- a/tycho-client/src/main.rs
+++ b/tycho-client/src/main.rs
@@ -152,6 +152,7 @@ async fn main() {
     run(exchanges, args).await;
 }
 
+#[allow(deprecated)]
 async fn run(exchanges: Vec<(String, Option<String>)>, args: CliArgs) {
     let tycho_ws_url = format!("ws://{}", &args.tycho_url);
     let tycho_rpc_url = format!("http://{}", &args.tycho_url);


tycho-client/src/feed/component_tracker.rs
```diff
@@ -13,6 +13,11 @@ use tycho_core::{
 pub enum ComponentFilter {
     Ids(Vec<String>),
     MinimumTVL(f64),
+    /// MinimumTVLRange is a tuple of (remove_tvl_threshold, add_tvl_threshold). Components that
+    /// drop below the remove threshold will be removed from tracking, components that exceed the
+    /// add threshold will be added. This helps buffer against components that fluctuate on the
+    /// tvl threshold boundary.
+    MinimumTVLRange((f64, f64)),
```
kayibal, line 19: I would prefer if we had added this to `MinimumTVL` variant, since you get the old behaviour by setting upper and lower to the same number. So if my intuition is not off, this would mean we can only have 2 paths instead of 3 wherever this is used.

louise-poole: I initially did this, but then it's not backwards compatible no?

tycho-client/src/feed/component_tracker.rs
```diff
@@ -10,16 +10,46 @@ use tycho_core::{
 };
 
 #[derive(Clone, Debug)]
-pub enum ComponentFilter {
+pub enum ComponentFilterVariant {
```
kayibal: Can we make this private? Or only `pub(crate)`?

tycho-client/src/feed/component_tracker.rs
```diff
@@ -10,16 +10,46 @@ use tycho_core::{
...
 }
 
+#[derive(Clone, Debug)]
+pub struct ComponentFilter {
+    variant: ComponentFilterVariant,
+}
+
+impl ComponentFilter {
+    #[allow(non_snake_case)] // for backwards compatibility
+    pub fn MinimumTVL(min_tvl: f64) -> ComponentFilter {
```
kayibal, line 30: let's deprecate this

tycho-client/src/feed/component_tracker.rs
```diff
@@ -10,16 +10,46 @@ use tycho_core::{
...
+        ComponentFilter { variant: ComponentFilterVariant::MinimumTVLRange((min_tvl, min_tvl)) }
+    }
+
+    #[allow(non_snake_case)]
+    pub fn MinimumTVLRange(remove_tvl_threshold: f64, add_tvl_threshold: f64) -> ComponentFilter {
```
kayibal: let's rename to `with_tvl_range(...)`

tycho-client/src/feed/component_tracker.rs
```diff
@@ -10,16 +10,46 @@ use tycho_core::{
...
+        ComponentFilter {
+            variant: ComponentFilterVariant::MinimumTVLRange((
+                remove_tvl_threshold,
+                add_tvl_threshold,
+            )),
+        }
+    }
+
+    #[allow(non_snake_case)] // for backwards compatibility
+    pub fn Ids(ids: Vec<String>) -> ComponentFilter {
```
kayibal, line 46: Let's deprecate this with `#[deprecated]`

tycho-client/src/feed/component_tracker.rs
```diff
@@ -10,16 +10,46 @@ use tycho_core::{
...
+        ComponentFilter { variant: ComponentFilterVariant::Ids(ids) }
+    }
+
+    pub fn variant(&self) -> &ComponentFilterVariant {
```
kayibal: Ah damn if this thing is still public it's not very sensible.

EDIT: Maybe if it was only public to the crate with `pub(crate)`, then this would work

louise-poole: Yeah - only way around it that I see is moving the synchroniser's logic to check components tvls, to the component tracker. Would actually make more sense there IMO.

tycho-client/src/feed/component_tracker.rs
```diff
@@ -183,6 +181,20 @@ where
...
+        match &self.filter.variant {
+            ComponentFilterVariant::Ids(_) => (Default::default(), Default::default()),
+            ComponentFilterVariant::MinimumTVLRange((remove_tvl, add_tvl)) => deltas
+                .component_tvl
+                .iter()
+                .filter(|(_, &tvl)| tvl < *remove_tvl || tvl > *add_tvl)
+                .map(|(id, _)| id.clone())
+                .partition(|id| deltas.component_tvl[id] > *add_tvl),
+        }
+    }
```
kayibal, line 197: Very nice!

PR #333, opened 2024-08-21
docs: Update Tycho Client docs

louise-poole:


tycho-client/Readme.md
```diff
@@ -1,52 +1,62 @@
...
 
-To use the client, you will need a connection to the indexer, we usually
-port-forward to our cluster's instance for this. If you are planning to use a local
-instance e.g. running within docker-compose you can skip this step.
-
-```
-kubectl port-forward -n dev-tycho deploy/tycho-indexer 4242
-```
+To use the Tycho Client, you will need a connection to the Tycho Indexer. We usually
+port-forward to our cluster's instance for this. If you are using a local instance, such as one running within Docker Compose, you can skip this step.
```
kayibal: I'd probably remove the cluster thing here since this will be soon open source.

PR #334, opened 2024-08-22
refactor: do not error if the protocol system isn't found in `get_block_finality`

zizou0x:
This PR allow to filter by protocol_system in Tycho RPC even if the corresponding extractor isn't currently running (for example if we use the `rpc` command of Tycho). In that case, we just get the state from the DB and don't apply anything on top of it

### DIFF
--- a/tycho-indexer/src/services/deltas_buffer.rs
+++ b/tycho-indexer/src/services/deltas_buffer.rs
@@ -261,6 +261,7 @@ impl PendingDeltas {
 
     /// Returns finality for any extractor, can error if lock is poisened. Returns None if buffer is
     /// empty.
+    /// Returns an error if the provided protocol system isn't found in the buffer.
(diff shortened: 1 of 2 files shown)


tycho-indexer/src/services/deltas_buffer.rs
```diff
@@ -276,7 +276,8 @@ impl PendingDeltas {
                         .map_err(|e| PendingDeltasError::LockError(system, e.to_string()))?;
                     Ok(guard.get_finality_status(version))
                 } else {
-                    Err(PendingDeltasError::UnknownExtractor(system))
+                    debug!(?system, "Mising requested protocol system in pending deltas");
+                    Ok(None)
```
kayibal: Hmmmm wouldn't it be nicer to catch the error outside and convert it there to None?

PR #337, opened 2024-08-22
chore: remove ethers dependency from `tycho-storage`

zizou0x:
This PR aims to remove the `ethers` dependency from `tycho-storage` without changing its logic.

### DIFF
--- a/tycho-core/src/hex_bytes.rs
+++ b/tycho-core/src/hex_bytes.rs
@@ -48,9 +48,83 @@ impl LowerHex for Bytes {
 }
 
 impl Bytes {
-    /// Return bytes as [`Vec::<u8>`]
+    /// This function converts the internal byte array into a `Vec<u8>`
+    ///
+    /// # Returns
+    ///
+    /// A `Vec<u8>` containing the bytes from the `Bytes` struct.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// let bytes = Bytes::from(vec![0x01, 0x02, 0x03]);
+    /// let vec = bytes.to_vec();
+    /// assert_eq!(vec, vec![0x01, 0x02, 0x03]);
+    /// ```
     pub fn to_vec(&self) -> Vec<u8> {
         self.as_ref().to_vec()
+    }
+
+    /// Left-pads the byte array to the specified length with the given padding byte.
+    ///
+    /// This function creates a new `Bytes` instance by prepending the specified padding byte
+    /// to the current byte array until its total length matches the desired length.
+    ///
+    /// If the current length of the byte array is greater than or equal to the specified length,
+    /// the original byte array is returned unchanged.
+    ///
+    /// # Arguments
+    ///
+    /// * `length` - The desired total length of the resulting byte array.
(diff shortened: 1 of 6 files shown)


tycho-storage/src/postgres/mod.rs
```diff
@@ -401,6 +402,19 @@ async fn maybe_lookup_version_ts(
...
+pub fn keccak256<T: AsRef<[u8]>>(bytes: T) -> [u8; 32] {
+    let mut output = [0u8; 32];
+
+    let mut hasher = Keccak::v256();
+    hasher.update(bytes.as_ref());
+    hasher.finalize(&mut output);
+
+    output
+}
+
```
zizou0x, line 417: This is going to be used by `tycho-indexer` too. Should we introduce a `tycho-utils` crate for these shared utils functions or maybe directly in `tycho-core`?

zizou0x: This was moved in `tycho-core` in my other PR, with an open question about `tycho-utils` still

louise-poole: I think it's fine in tycho-core 👍

tycho-core/src/hex_bytes.rs
```diff
@@ -226,6 +300,24 @@ macro_rules! impl_from_for_ethers_fixed_hash {
 
 impl_from_for_ethers_fixed_hash!(H160, H256);
 
+macro_rules! impl_from_uint_for_bytes {
```
louise-poole, line 303: Made your own macro - aren't you cool 🤓

tycho-storage/src/postgres/chain.rs
```diff
@@ -582,16 +581,14 @@ mod test {
...
         let c0_address =
             Bytes::from_str("6B175474E89094C44Da98b954EedeAC495271d0F").expect("c0 address valid");
-        let exp_slots: HashMap<U256, U256> = vec![
-            (U256::from(0), U256::from(1)),
-            (U256::from(1), U256::from(5)),
-            (U256::from(2), U256::from(1)),
+        let exp_slots: HashMap<Bytes, Bytes> = vec![
+            (Bytes::from(0_u8).lpad(32, 0), Bytes::from(1_u8).lpad(32, 0)),
+            (Bytes::from(1_u8).lpad(32, 0), Bytes::from(5_u8).lpad(32, 0)),
+            (Bytes::from(2_u8).lpad(32, 0), Bytes::from(1_u8).lpad(32, 0)),
```
louise-poole, line 591: I assume we'd need to do this combo very often (create left padded bytes). Maybe we could combine them into a single function? Maybe a new macro for a function like `into_left_padded_bytes` and we could do:
`0_u8.into_left_padded_bytes()`

louise-poole: Hmmmm... going through the rest of this PR I see this isn't actually used anywhere else (I had expected it to be used often). We can ignore this comment above then 👍

zizou0x: Actually it's used a lot, but I introduced a function especially for this case directly on the `Bytes` struct

PR #340, opened 2024-08-23
refactor: remove `ethers` dependencies from all crates except `tycho-ethereum`

zizou0x:
This PR completely removes `ethers` dependency from every Tycho modules except `tycho-ethereum`. 

Removing `ethers` was done with a big refactor of the whole repo. We introduced some generic traits for blockchain tailored functions and moved many generic structs into `tycho-core`.

It also introduces some breaking changes such as bytes not being padded anymore (this has to be handled by Substreams ideally).

### DIFF
--- a/tycho-core/src/dto.rs
+++ b/tycho-core/src/dto.rs
@@ -220,8 +220,9 @@ pub struct BlockChanges {
     pub revert: bool,
     #[serde(with = "hex_hashmap_key", default)]
     pub new_tokens: HashMap<Bytes, ResponseToken>,
-    #[serde(with = "hex_hashmap_key")]
+    #[serde(alias = "account_deltas", with = "hex_hashmap_key")]
     pub account_updates: HashMap<Bytes, AccountUpdate>,
+    #[serde(alias = "state_deltas")]
     pub state_updates: HashMap<String, ProtocolStateDelta>,
     pub new_protocol_components: HashMap<String, ProtocolComponent>,
     pub deleted_protocol_components: HashMap<String, ProtocolComponent>,
@@ -406,6 +407,7 @@ pub struct ProtocolComponent {
     pub chain: Chain,
     #[schema(value_type=Vec<String>)]
     pub tokens: Vec<Bytes>,
+    #[serde(alias = "contract_addresses")]
     #[schema(value_type=Vec<String>)]
     pub contract_ids: Vec<Bytes>,
     #[serde(with = "hex_hashmap_value")]
@@ -1061,6 +1063,8 @@ pub enum Health {
 
 #[cfg(test)]
 mod test {
+    use std::str::FromStr;
+
     use maplit::hashmap;
     use rstest::rstest;
 
@@ -1237,6 +1241,104 @@ mod test {
         };
 
         assert_eq!(result, expected);
+    }
+
+    fn create_models_block_entity_changes() -> crate::models::blockchain::BlockAggregatedChanges {
+        let base_ts = 1694534400; // Example base timestamp for 2023-09-14T00:00:00
+
+        crate::models::blockchain::BlockAggregatedChanges {
+            extractor: "native_name".to_string(),
+            chain: models::Chain::Ethereum,
+            block: models::blockchain::Block::new(
+                3,
+                models::Chain::Ethereum,
+                Bytes::from_str("0x0000000000000000000000000000000000000000000000000000000000000003").unwrap(),
+                Bytes::from_str("0x0000000000000000000000000000000000000000000000000000000000000002").unwrap(),
+                NaiveDateTime::from_timestamp_opt(base_ts + 3000, 0).unwrap(),
+            ),
+            finalized_block_height: 1,
+            revert: true,
+            state_deltas: HashMap::from([
+                ("pc_1".to_string(), models::protocol::ProtocolComponentStateDelta {
+                    component_id: "pc_1".to_string(),
+                    updated_attributes: HashMap::from([
+                        ("attr_2".to_string(), Bytes::from("0x0000000000000002")),
+                        ("attr_1".to_string(), Bytes::from("0x00000000000003e8")),
+                    ]),
+                    deleted_attributes: HashSet::new(),
+                }),
+            ]),
+            new_tokens: HashMap::new(),
+            new_protocol_components: HashMap::from([
+                ("pc_2".to_string(), crate::models::protocol::ProtocolComponent {
+                    id: "pc_2".to_string(),
+                    protocol_system: "native_protocol_system".to_string(),
+                    protocol_type_name: "pt_1".to_string(),
+                    chain: models::Chain::Ethereum,
+                    tokens: vec![
+                        Bytes::from_str("0xdac17f958d2ee523a2206206994597c13d831ec7").unwrap(),
+                        Bytes::from_str("0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48").unwrap(),
+                    ],
+                    contract_addresses: vec![],
+                    static_attributes: HashMap::new(),
+                    change: models::ChangeType::Creation,
+                    creation_tx: Bytes::from_str("0x000000000000000000000000000000000000000000000000000000000000c351").unwrap(),
+                    created_at: NaiveDateTime::from_timestamp_opt(base_ts + 5000, 0).unwrap(),
+                }),
+            ]),
+            deleted_protocol_components: HashMap::from([
+                ("pc_3".to_string(), crate::models::protocol::ProtocolComponent {
+                    id: "pc_3".to_string(),
+                    protocol_system: "native_protocol_system".to_string(),
+                    protocol_type_name: "pt_2".to_string(),
+                    chain: models::Chain::Ethereum,
+                    tokens: vec![
+                        Bytes::from_str("0x6b175474e89094c44da98b954eedeac495271d0f").unwrap(),
+                        Bytes::from_str("0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2").unwrap(),
+                    ],
+                    contract_addresses: vec![],
+                    static_attributes: HashMap::new(),
+                    change: models::ChangeType::Deletion,
+                    creation_tx: Bytes::from_str("0x0000000000000000000000000000000000000000000000000000000000009c41").unwrap(),
+                    created_at: NaiveDateTime::from_timestamp_opt(base_ts + 4000, 0).unwrap(),
+                }),
+            ]),
+            component_balances: HashMap::from([
+                ("pc_1".to_string(), HashMap::from([
+                    (Bytes::from_str("0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48").unwrap(), models::protocol::ComponentBalance {
+                        token: Bytes::from_str("0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48").unwrap(),
+                        balance: Bytes::from("0x00000001"),
+                        balance_float: 1.0,
+                        modify_tx: Bytes::from_str("0x0000000000000000000000000000000000000000000000000000000000000000").unwrap(),
+                        component_id: "pc_1".to_string(),
+                    }),
+                    (Bytes::from_str("0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2").unwrap(), models::protocol::ComponentBalance {
+                        token: Bytes::from_str("0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2").unwrap(),
+                        balance: Bytes::from("0x000003e8"),
+                        balance_float: 1000.0,
+                        modify_tx: Bytes::from_str("0x0000000000000000000000000000000000000000000000000000000000007531").unwrap(),
+                        component_id: "pc_1".to_string(),
+                    }),
+                ])),
+            ]),
+            component_tvl: HashMap::new(),
+            account_deltas: Default::default(),
+        }
+    }
+
+    #[test]
+    fn test_serialize_deserialize_block_changes() {
+        // Test that models::BlockAggregatedChanges serialized as json can be deserialized as
+        // dto::BlockChanges.
+
+        // Create a models::BlockAggregatedChanges instance
+        let block_entity_changes = create_models_block_entity_changes();
+
+        // Serialize the struct into JSON
+        let json_data = serde_json::to_string(&block_entity_changes).expect("Failed to serialize");
+
+        // Deserialize the JSON back into a dto::BlockChanges struct
+        serde_json::from_str::<BlockChanges>(&json_data).expect("parsing failed");
     }
 
     #[test]


tycho-indexer/src/extractor/compat/attributes.rs
```diff
@@ -23,13 +22,13 @@ pub fn add_default_attributes(mut changes: BlockChanges, attributes: &[&str]) ->
                     {
                         state
                             .updated_attributes
-                            .insert(mandatory_attr.to_string(), Bytes::from(U256::zero()));
+                            .insert(mandatory_attr.to_string(), Bytes::from([0u8; 32]));
```
zizou0x: This will most likely be used alot, might be worth to implement it directly on `Bytes`

tycho-core/src/lib.rs
```diff
@@ -4,9 +4,24 @@ pub mod serde_primitives;
...
+/// Note that strings are interpreted as UTF-8 bytes,
+pub fn keccak256<T: AsRef<[u8]>>(bytes: T) -> [u8; 32] {
+    let mut output = [0u8; 32];
+
+    let mut hasher = Keccak::v256();
+    hasher.update(bytes.as_ref());
+    hasher.finalize(&mut output);
+
+    output
+}
```
zizou0x, line 28: As mentioned in my other PR, I'm not sure this really belongs in `tycho-core`. But on the other hand introducing a `tycho-utils` crate only for this seems a bit too much to me.

tycho-indexer/Cargo.toml
```diff
@@ -61,6 +61,8 @@ utoipa-swagger-ui = { version = "6.0.0", features = ["actix-web"] }
 ethrpc = { git = "https://github.com/cowprotocol/services", rev = "f3678428991e055ceb517f184d1b37244f2d8f51" }
 web3 = { version = "0.19", default-features = false }
 mini-moka = "0.10.3"
+num-bigint = "0.4.4"
```
zizou0x, line 62: Currently only used to convert a int string into `Bytes`. We could implement a homemade utils function for this, but it is probably ok to have `bigint` as a dependency anyway?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -82,24 +82,12 @@ pub struct Block {
...
+pub trait TryFromMessage {
+    type Args<'a>;
 
-    // This serves to expose the 'try_decode' method from utils externally
-    pub fn hash_from_bytes(hash: Bytes) -> H256 {
-        H256::try_decode(&hash, "tx hash").expect("Failed to decode tx hash")
-    }
+    fn try_from_message(args: Self::Args<'_>) -> Result<Self, ExtractionError>
+    where
+        Self: Sized;
```
zizou0x: This is still widely open for discussion, the main reason for this is that we can't implement functions on foreign structs without trait. I hope this trait is flexible enough to be implemented on every struct on which we need the `try_from_message`. 

We have others alternatives to this design, like introducing a wrapped struct on which we implement the `try_from_message`.

tycho-indexer/src/extractor/compat/attributes.rs
zizou0x: Addressed on https://github.com/propeller-heads/tycho-indexer/pull/340/commits/6206f7c73d6777dcc7fd5e6855704f41da87757a

tycho-core/src/lib.rs
louise-poole: I think let's keep this on tycho-core 👍

tycho-indexer/src/main.rs
```diff
@@ -329,10 +332,10 @@ async fn initialize_accounts(
 
     info!(block_number = block.number, "Initializing accounts");
 
-    let tx = Transaction {
-        hash: H256::random(),
-        block_hash: block.hash,
-        from: H160::zero(),
+    let tx = CoreTransaction {
```
louise-poole: What is meant by 'core' transaction?

tycho-indexer/Cargo.toml
kayibal: It may be useful in the future and should be rather lighweight so I'd say it is ok.

tycho-core/src/traits.rs
```diff
@@ -0,0 +1,19 @@
...
+
+#[async_trait]
+pub trait AccountExtractor {
+    type Error;
+
+    async fn get_accounts(
+        &self,
+        block: Block,
+        account_addresses: Vec<Address>,
+    ) -> Result<HashMap<Bytes, ContractDelta>, Self::Error>;
```
kayibal: The facto that we return a Delta if we expect a snapshot is somewhat weird no? If it is because of the missing transaction, maybe we move this responsibility to the AccountExtractor as well?

tycho-indexer/src/extractor/compat/balances.rs
```diff
@@ -42,9 +42,9 @@ pub fn transcode_ambient_balances(mut changes: BlockChanges) -> BlockChanges {
                     balance
                         .iter_mut()
                         .for_each(|(_, value)| {
-                            value.balance = transcode_ascii_balance_to_be(&value.balance)
+                            value.new_balance = transcode_ascii_balance_to_be(&value.new_balance)
                                 .expect("Balance transcoding failed");
-                            value.balance_float = bytes_to_f64(value.balance.as_ref())
+                            value.balance_float = bytes_to_f64(value.new_balance.as_ref())
```
louise-poole: When was this changed to `new_balance`? 👀  Isn't it already obvious that when this balance is emitted it's a new balance?
If we want to stick with the `new_balance` wording (though it feels unnecessary) we should also update the float var to `new_balance_float` no?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -650,21 +638,22 @@ impl Block {
...
 }
 
-impl Transaction {
-    /// Parses transaction from tychos protobuf transaction message
-    pub fn try_from_message(
-        msg: substreams::Transaction,
-        block_hash: &H256,
-    ) -> Result<Self, ExtractionError> {
+impl TryFromMessage for Transaction {
+    type Args<'a> = (substreams::Transaction, &'a H256);
```
kayibal: Why are we using H256 still here?

tycho-indexer/src/extractor/evm/hybrid.rs
```diff
@@ -624,10 +625,11 @@ where
                     .await?;
             }
         }
-        self.update_last_processed_block(msg.block)
+        self.update_last_processed_block(msg.block.clone())
             .await;
 
-        self.report_progress(msg.block).await;
+        self.report_progress(msg.block.clone())
+            .await;
```
kayibal: We should probably check if we would be ok passing a reference here instead.

tycho-indexer/src/extractor/evm/hybrid.rs
```diff
@@ -2038,7 +2039,7 @@ mod test_serial_db {
         evm::BlockChanges {
             extractor: "vm:ambient".to_owned(),
             chain: Chain::Ethereum,
-            block: evm::Block::default(),
+            block: Block { hash: Bytes::zero(32), ..Default::default() },
```
kayibal, line 2061: We should probably add a new method for these?

tycho-core/src/models/contract.rs
```diff
@@ -80,7 +80,7 @@ impl Account {
 }
 
 #[derive(Clone, Debug, PartialEq, Default)]
-pub struct ContractDelta {
+pub struct AccountUpdate {
```
kayibal: I think I'd have a slight preference for `AccountDelta` here

tycho-core/src/models/contract.rs
```diff
@@ -55,7 +55,7 @@ impl Account {
         self.balance_modify_tx = modified_at.clone();
     }
 
-    pub fn apply_contract_delta(&mut self, delta: &ContractDelta) -> Result<(), DeltaError> {
+    pub fn apply_contract_delta(&mut self, delta: &AccountUpdate) -> Result<(), DeltaError> {
```
kayibal: Would be good rename this to just `apply_delta`

tycho-core/src/models/contract.rs
```diff
@@ -100,21 +103,14 @@ impl AccountUpdate {
...
-        address: &Address,
-        slots: Option<&HashMap<StoreKey, Option<StoreVal>>>,
-        balance: Option<&Balance>,
-        code: Option<&Code>,
+        chain: Chain,
+        address: Address,
+        slots: HashMap<StoreKey, Option<StoreVal>>,
+        balance: Option<Balance>,
+        code: Option<Code>,
         change: ChangeType,
```
kayibal, line 114: I preferred the version where params are references...

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -509,72 +509,38 @@ impl TryFromMessage for ComponentBalance {
...
-/// e.g., swap, lend, bridge. The component is associated with a specific `Chain` and holds
-/// information about tradable tokens, related contract IDs, and static attributes.
-///
-/// A `ProtocolComponent` can have a one-to-one or one-to-many relationship with contracts.
-/// For example, `UniswapV2` and `UniswapV3` have a one-to-one relationship one component (pool) one
-/// contract, while `Ambient` has a one-to-many relationship with a single component and multiple
-/// contracts.
-///
-/// The `ProtocolComponent` struct is designed to store static attributes related to the associated
-/// smart contract.
```
kayibal, line 734: Did you double check we have these docs on core?

tycho-core/src/traits.rs
```diff
@@ -0,0 +1,29 @@
...
+
+// #[async_trait]
+// pub trait TokenPreProcessor: Send + Sync {
+//     async fn get_tokens(
+//         &self,
+//         addresses: Vec<Bytes>,
+//         token_finder: Arc<dyn TokenOwnerFinding>,
+//         block: BlockNumber,
+//     ) -> Vec<ERC20Token>;
+// }
```
louise-poole: Do we need this?

tycho-core/src/models/blockchain.rs
```diff
@@ -51,58 +53,94 @@ pub struct TransactionDeltaGroup<T> {
     tx: Transaction,
 }
 
-#[derive(Debug, Clone, PartialEq)]
-pub struct BlockAggregatedDeltas {
+#[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq)]
+pub struct AggregatedBlockChanges {
```
kayibal: I preferred the other name since it made it clear that changes are BlockAggregated instead of BlockChanges being aggregated by something unknown.

tycho-core/src/models/blockchain.rs
```diff
@@ -51,58 +53,94 @@ pub struct TransactionDeltaGroup<T> {
...
     pub block: Block,
-    pub finalised_block_height: u64,
+    pub finalized_block_height: u64,
     pub revert: bool,
-    pub state_deltas: HashMap<String, ProtocolComponentStateDelta>,
-    pub account_deltas: HashMap<Bytes, AccountUpdate>,
-    pub new_components: HashMap<String, ProtocolComponent>,
-    pub deleted_components: HashMap<String, ProtocolComponent>,
+    pub state_updates: HashMap<String, ProtocolComponentStateDelta>,
+    pub account_updates: HashMap<Bytes, AccountUpdate>,
```
kayibal: I'd prefer if we sticked to deltas

tycho-indexer/src/extractor/evm/hybrid.rs
```diff
@@ -843,16 +842,16 @@ where
...
                 .fold(HashMap::new(), |mut acc, ((addr, key), value)| {
                     acc.entry(addr)
-                        .or_insert_with(|| evm::AccountUpdate {
-                            address: addr,
+                        .or_insert_with(|| AccountUpdate {
+                            address: addr.into(),
                             chain: self.chain,
                             slots: HashMap::new(),
                             balance: None, //TODO: handle balance changes
                             code: None,    //TODO: handle code changes
```
louise-poole: Is there anything to handle here? We're just setting a default here that can be updated by subsequent updates no? I think we can remove these TODOs.

tycho-indexer/src/extractor/evm/hybrid.rs
```diff
@@ -309,7 +308,7 @@ where
         &self,
         revert_buffer: &RevertBuffer<BlockUpdateWithCursor<evm::BlockChanges>>,
         reverted_balances_keys: &[(&String, &Bytes)],
-    ) -> Result<HashMap<String, HashMap<H160, evm::ComponentBalance>>, ExtractionError> {
+    ) -> Result<HashMap<String, HashMap<Bytes, evm::ComponentBalance>>, ExtractionError> {
```
kayibal: Should this be `evm::ComponentBalance` still here?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -479,6 +382,7 @@ impl StateUpdateBufferEntry for BlockContractChanges {
         todo!()
     }
 
+    #[allow(clippy::mutable_key_type)]
```
louise-poole: Was this not complaining before? Why do we need this now?

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -1989,9 +1897,17 @@ pub mod fixtures {
...
     }
 
-    pub fn evm_slots(data: impl IntoIterator<Item = (u64, u64)>) -> HashMap<U256, U256> {
+    pub fn evm_slots(data: impl IntoIterator<Item = (u64, u64)>) -> HashMap<Bytes, Bytes> {
+        data.into_iter()
+            .map(|(s, v)| (U256::from(s).into(), U256::from(v).into()))
+            .collect()
+    }
+
+    //Utils function that return slots that match `AccountUpdate` slots.
```
louise-poole: ```// Utils function that return slots that match `AccountUpdate` slots.```

tycho-indexer/src/extractor/evm/hybrid.rs
```diff
@@ -423,14 +422,15 @@ where
                         map_vault(&change.protocol_system)
                             .or_else(|| {
                                 change
-                                    .contract_ids
+                                    .contract_addresses
                                     // TODO: Currently, it's assumed that the pool is always the
                                     // first contract in the
                                     // protocol component. This approach is a temporary
                                     // workaround and needs to be revisited for a more robust
                                     // solution.
```
louise-poole, line 427:
```
// TODO: Currently, it's assumed that the balance owner is 
// always the first contract in the protocol component. This 
// approach is a temporary workaround and needs to be 
// revisited for a more robust solution.
```
I think this can already be updated to use the `balance_owner` attribute when applicable no? Not something for this PR, but I can add a task to do it so it's ready for when we resync balancer. Does ambient emit a balance owner attribute? It should.

tycho-indexer/src/extractor/evm/mod.rs
zizou0x: I'm planning to change this once I finished to refactor every `evm::mod` structs (if it's not already, most of the time when I change the structs everything tend to naturally change to `Bytes`)

tycho-core/src/traits.rs
zizou0x: Yes I also think that it's unexpected, I would keep the scope of this PR to remove ethers type only though.

It will be done in another PR later, I'll add a TODO and create a jira task for it

tycho-core/src/models/contract.rs
zizou0x: Hm don't you think that the `new` function shouldn't force us to clone things? My idea was that I prefer being able to give ownership if possible instead of cloning then dropping.

tycho-indexer/src/extractor/evm/mod.rs
zizou0x: I'm planning to do it in my last commit, get all the docs that were remove and merge them with what's currently in tycho-core

tycho-core/src/models/blockchain.rs
zizou0x: True

tycho-indexer/src/extractor/evm/hybrid.rs
zizou0x: Nope, I've tried to remove them when I was noticing them but I might have missed some like this because it doesn't error. The `tycho-core` struct is just reimported from `evm`. I'll try to clean this once I'm done with the big work

tycho-core/src/traits.rs
zizou0x: We will but I decided that I would do it in another PR to keep it reviewable, I'll remove it from this one

tycho-indexer/src/extractor/evm/hybrid.rs
zizou0x: I would say it's out of scope for this PR, we should do a PR where we address/remove TODOs later

tycho-indexer/src/extractor/evm/hybrid.rs
zizou0x: Ideally yes, but unfortunately Ambient doesn't emit it.

tycho-indexer/src/extractor/evm/mod.rs
zizou0x: I've spent a lot of time on this, I think there is no way to fix it correctly. It's a false positive from clippy because we use `Bytes`.

tycho-indexer/src/extractor/compat/balances.rs
zizou0x: Not sure why it's called `new_balance` either. I didn't focus on naming in this PR. 

But this one feels too weird, so I'll change it

tycho-indexer/src/main.rs
zizou0x: This was a temporary fix because we had two `Transaction` structs still. Now that it's normalized I'll just call it `Transaction`

tycho-core/src/models/blockchain.rs
```diff
@@ -44,59 +73,238 @@ pub struct TransactionDeltaGroup<T> {
...
 
-#[derive(Debug, Clone, PartialEq)]
-pub struct BlockAggregatedDeltas {
+#[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq)]
+pub struct BlockAggregatedChanges {
     pub extractor: String,
     pub chain: Chain,
     pub block: Block,
-    pub finalised_block_height: u64,
+    pub finalized_block_height: u64,
```
tvinagre, line 81: We need to decide if we're american or british 🤣

tycho-core/src/models/blockchain.rs
```diff
@@ -44,59 +73,238 @@ pub struct TransactionDeltaGroup<T> {
...
+        deleted_components: HashMap<String, ProtocolComponent>,
+        component_balances: HashMap<ComponentId, HashMap<Bytes, ComponentBalance>>,
+        component_tvl: HashMap<String, f64>,
     ) -> Self {
         Self {
             extractor: extractor.to_string(),
             chain,
             block,
-            finalised_block_height,
+            finalized_block_height: finalised_block_height,
```
tvinagre: Why not rename the local var as well?

tycho-core/src/models/contract.rs
```diff
@@ -95,30 +106,124 @@ impl ContractDelta {
...
+            tx.hash.clone(),
+            Some(tx.hash.clone()),
+        )
+    }
+
+    /// Merge this update (`self`) with another one (`other`)
+    ///
+    /// This function is utilized for aggregating multiple updates into a single
+    /// update. The attribute values of `other` are set on `self`.
+    /// Meanwhile, contract storage maps are merged, in which keys from `other`
```
tvinagre:
```suggestion
    /// Meanwhile, contract storage maps are merged, with keys from `other` taking precedence.
```

tycho-core/src/traits.rs
```diff
@@ -0,0 +1,19 @@
...
+
+#[async_trait]
+pub trait AccountExtractor {
+    type Error;
+
+    async fn get_accounts(
+        &self,
+        block: Block,
+        account_addresses: Vec<Address>,
+    ) -> Result<HashMap<Bytes, AccountDelta>, Self::Error>; //TODO: do not return `AccountUpdate` but `Account`
```
tvinagre, line 24: TODO reminder

zizou0x: As mentioned [here](https://github.com/propeller-heads/tycho-indexer/pull/340#discussion_r1736066395), I consider this to be out of scope for this PR

tycho-core/src/models/blockchain.rs
zizou0x: I guess we're using American english? Only this one was British

tycho-core/src/models/blockchain.rs
```diff
@@ -179,7 +179,7 @@ pub struct TxWithChanges {
 impl TxWithChanges {
     pub fn new(
         protocol_components: HashMap<ComponentId, ProtocolComponent>,
-        account_updates: HashMap<Bytes, AccountUpdate>,
+        account_updates: HashMap<Bytes, AccountDelta>,
```
louise-poole: Should we not rename it to `account_deltas` for TxWithChanges too?

tycho-core/src/models/protocol.rs
```diff
@@ -12,6 +12,16 @@ use super::{
     TxHash,
 };
 
+/// Represents the static parts of a protocol component.
```
louise-poole: This doesn't actually describe what a protocol component is though... just that it is static. In the main docs I added something like:
```
A Protocol Component represents an entity that facilitates the swapping of two tokens. A component consists of static values that remain constant after creation
```

tycho-core/src/models/protocol.rs
```diff
@@ -12,6 +12,16 @@ use super::{
...
+///
+/// `ProtocolComponent` provides detailed descriptions of a component of a protocol,
+/// for example, swap pools that enables the exchange of two tokens.
+///
+/// A `ProtocolComponent` can be associated with an `Account`, and it has an identifier (`id`) that
+/// can be either the on-chain address or a custom one. It belongs to a specific `ProtocolSystem`
+/// and has a `ProtocolTypeID` that associates it with a `ProtocolType` that describes its behaviour
+/// e.g., swap, lend, bridge. The component is associated with a specific `Chain` and holds
+/// information about tradable tokens, related contract IDs, and static attributes.
```
louise-poole, line 22: Might be good to mention that its dynamic counterpart is ProtocolComponentState.

PR #342, opened 2024-08-26
docs: Update rpc swagger docs

louise-poole:


tycho-core/src/dto.rs
```diff
@@ -694,10 +694,13 @@ fn default_include_balances_flag() -> bool {
 
 #[derive(Serialize, Deserialize, Default, Debug, IntoParams)]
 pub struct StateRequestParameters {
+    /// The minimum TVL of the protocol components to return, denoted in ETH.
```
kayibal: Let's keep this chain agnostic and mention that it is denoted in the chains native token.

tycho-core/src/dto.rs
```diff
@@ -694,10 +694,13 @@ fn default_include_balances_flag() -> bool {
...
     #[param(default = 0)]
     pub tvl_gt: Option<u64>,
+    /// The minimum inertia of the protocol components to return.
```
kayibal, line 703: I think we should remove this parameter.

tycho-core/src/dto.rs
```diff
@@ -694,10 +694,13 @@ fn default_include_balances_flag() -> bool {
...
     #[param(default = 0)]
     pub inertia_min_gt: Option<u64>,
+    /// Whether to include balances in the response.
```
kayibal: ERC20 balances

tycho-indexer/src/services/rpc.rs
```diff
@@ -478,9 +478,27 @@ where
     }
 }
 
+/// Retrieve contract states
+///
+/// This endpoint retrieves the state of contracts within a specific execution environment. If no
+/// contract ids are given, all contracts are returned. Note that `protocol_system` is not a filter;
```
kayibal, line 515: Hmm good point I haven't considered that... I feel it should also be a filter though, but ok we can handle this client side for now. Please mention here that if wished filtering by protocol should currently be handled client side. 

The reason for not directly supporting this filter this is that paginating this endpoint may be very difficult with the current design...

tycho-indexer/src/services/rpc.rs
```diff
@@ -509,9 +527,20 @@ pub async fn contract_state<G: Gateway>(
...
 
+/// Retrieve tokens
+///
+/// This endpoint retrieves tokens for a specific execution environment, filtered by various
+/// criteria. The tokens are returned in a paginated format.
+///
+/// Filters:
+/// - `token_addresses`: (optional) A list of token addresses to filter by. If ommitted, will return
+///   all tokens.
+/// - `min_quality`: (optional) The minimum quality of tokens to return. Main quality levels: 100
```
kayibal: Maybe here we can refer to the struct instead? This would avoid us having to constantly update this line.

tycho-indexer/src/services/rpc.rs
```diff
@@ -573,9 +611,27 @@ pub async fn protocol_components<G: Gateway>(
     }
 }
 
+/// Retrieve protocol states
+///
+/// This endpoint retrieves the state of protocols within a specific execution environment.
+/// Currently, the filters are not compounded, meaning that if multiple filters are provided, one
+/// will be prioritised. The proority from highest to lowest is as follows: 'protocol_ids',
```
kayibal: typo

tycho-indexer/src/services/rpc.rs
louise-poole: Referencing the struct means they have to leave the nice swagger interface and go into the code to figure out what this means.

tycho-core/src/dto.rs
louise-poole: It will no longer be backwards compatible then no? I thought it would be fine to leave it for now as we'll be removing the use of these request parameters entirely soon.

tycho-indexer/src/services/rpc.rs
kayibal: Are you sure? AFAIK the dto models docs should be contained within the swagger

PR #344, opened 2024-08-28
feat: Move rpc endpoint params to request body

louise-poole:
Design decision: for backwards compatibility we have kept the old endpoint interfaces and introduced new ones for this change. I marked the old ones as deprecated and we will remove it in a future version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5950,7 +5950,7 @@ checksum = "1f3ccbac311fea05f86f61904b462b55fb3df8837a366dfc601a0161d0532f20"
 
 [[package]]
 name = "token-analyzer"
-version = "0.11.0"
+version = "0.11.1"
 dependencies = [
  "anyhow",
  "async-trait",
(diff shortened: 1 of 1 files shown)


tycho-indexer/src/services/rpc.rs
```diff
@@ -549,19 +612,50 @@ pub async fn tokens<G: Gateway>(
     request_body = ProtocolComponentsRequestBody,
     params(
         ("execution_env" = Chain, description = "Execution environment"),
-        ProtocolComponentRequestParameters
+        ("tvl_gt" = Option<f64>, Query, description = "Filter components by TVL greater than this value")
     ),
```
louise-poole, line 616: Note for anyone interested: Using the `ProtocolComponentRequestParameters` object here causes the swagger docs to behave weirdly: they will generate places for you to enter values for these params, but will not include them in the request it sends. You have to explicitly list the values you expect here for it to actually include them.

PR #345, opened 2024-08-30
feat(rpc): add pagination to all rpc endpoints

tvinagre:


tycho-client/src/feed/synchronizer.rs
```diff
@@ -77,7 +76,7 @@ impl Snapshot {
 
 #[derive(Clone, PartialEq, Debug, Default, Serialize)]
 pub struct StateSyncMessage {
-    /// The block information for this update.
+    /// The bloc information for this update.
```
zizou0x: Why?

tycho-core/src/dto.rs
```diff
@@ -673,13 +680,16 @@ pub struct ResponseAccount {
     #[schema(value_type=HashMap<String, String>, example="0x123456789")]
     #[serde(with = "hex_bytes")]
     pub code_hash: Bytes,
-    #[schema(value_type=HashMap<String, String>, example="0x8f1133bfb054a23aedfe5d25b1d81b96195396d8b88bd5d4bcf865fc1ae2c3f4")]
+    #[schema(value_type=HashMap<String, String>, example="0x8f1133bfb054a23aedfe5d25b1d81b96195396d8b88bd5d4bcf865fc1ae2c3f4"
```
zizou0x: 🫠

tycho-client/src/rpc.rs
```diff
@@ -52,6 +52,52 @@ pub trait RPCClient {
...
+        }
+
+        try_join_all(tasks)
+            .await
+            .map(|responses| StateRequestResponse {
+                accounts: responses
+                    .into_iter()
+                    .flat_map(|r| r.accounts.into_iter())
+                    .collect(),
+                // TODO: Should use the response from the task
```
zizou0x: I didn't get this TODO

tycho-client/src/rpc.rs
```diff
@@ -52,6 +52,52 @@ pub trait RPCClient {
...
+        ids: &[Bytes],
+        protocol_system: &Option<String>,
+        version: &VersionParam,
+        chunk_size: usize,
+        concurrency: usize,
+    ) -> Result<StateRequestResponse, RPCError> {
+        let semaphore = Arc::new(Semaphore::new(concurrency));
+        let chunked_bodies = ids
+            .chunks(chunk_size)
+            .map(|_| StateRequestBody {
```
zizou0x: This looks very inefficient to me, if we don't use the chunk there is probably a better way to do this. Something like 
`len/chuck_size`

tycho-client/src/rpc.rs
```diff
@@ -52,6 +52,52 @@ pub trait RPCClient {
...
+                contract_ids: Some(ids.to_vec()),
+                protocol_system: protocol_system.clone(),
+                chain,
+                version: version.clone(),
+                pagination: PaginationParams { page: 0, page_size: chunk_size as i64 },
```
zizou0x, line 78: Page stays at 0? We're gonna get the same response multiple times no?

tycho-indexer/src/services/rpc.rs
```diff
@@ -303,6 +309,7 @@ where
                 .into_iter()
                 .map(dto::ResponseProtocolState::from)
                 .collect(),
+            request.pagination.clone(),
```
zizou0x: Can't we just give ownership here?

tycho-client/src/feed/synchronizer.rs
tvinagre: lol

tycho-core/src/dto.rs
tvinagre: no comments, undoing

tycho-client/src/rpc.rs
tvinagre: It was because on the next line we're creating a fake pagination instead of using the response. But turns out that it doesn't make sense to use the response so I'm removing this comment:

pagination: PaginationParams { page: 0, page_size: chunk_size as i64 },

tycho-client/src/rpc.rs
```diff
@@ -4,7 +4,7 @@
 //! Call (RPC) endpoints of Tycho. These endpoints are chiefly responsible for facilitating data
 //! queries, especially querying snapshots of data.
 //!
-//! Currently we provide only a HTTP implementation.
+//! Currently, we provide only a HTTP implementation.
```
louise-poole:
```suggestion
//! Currently, we provide only an HTTP implementation.
```

tycho-core/src/dto.rs
```diff
@@ -543,6 +546,7 @@ impl StateRequestBody {
             protocol_system: None,
             version: VersionParam { timestamp: None, block: Some(block.clone()) },
             chain: block.chain.unwrap_or_default(),
+            pagination: PaginationParams::default(),
```
louise-poole, line 548: I'm thinking: should the default pagination params be endpoint-specific? Probably not necessary... but yeah, I would think the default for tokens could be a lot higher than the default for accounts with contract code etc.

tycho-core/src/dto.rs
```diff
@@ -1127,6 +1160,7 @@ mod test {
                 }),
             },
             chain: Chain::Ethereum,
+            pagination: PaginationParams { page: 0, page_size: 20 },
```
louise-poole: Why not:
```suggestion
            pagination: Default::default(),
```

tycho-storage/src/postgres/contract.rs
```diff
@@ -814,6 +815,14 @@ impl PostgresGateway {
             if let Some(contract_ids) = ids {
                 q = q.filter(address.eq_any(contract_ids));
             }
+
+            // Apply pagination if provided
+            if let Some(pagination) = pagination_params {
+                q = q
+                    .limit(pagination.page_size)
+                    .offset(pagination.offset());
+            }
```
louise-poole, line 855: Shouldn't we enforce that pagination is always provided? To protect db query size?

tycho-storage/src/postgres/orm.rs
```diff
@@ -633,22 +633,42 @@ impl ProtocolState {
...
             .filter(protocol_component::chain_id.eq(chain_id))
+            .select(protocol_component::external_id)
+            .distinct()
+            .into_boxed();
+
+        if let Some(pagination) = pagination_params {
+            component_query = component_query
+                .limit(pagination.page_size)
+                .offset(pagination.page * pagination.page_size);
+        }
```
louise-poole: I think we should enforce pagination and not let it be optional.

tycho-storage/src/postgres/orm.rs
```diff
@@ -633,22 +633,42 @@ impl ProtocolState {
...
         conn: &mut AsyncPgConnection,
     ) -> QueryResult<Vec<(Self, ComponentId)>> {
-        let mut query = protocol_state::table
-            .inner_join(protocol_component::table)
+        // Subquery to get distinct component external IDs based on pagination
+        let mut component_query = protocol_component::table
             .filter(protocol_component::external_id.eq_any(component_ids))
             .filter(protocol_component::chain_id.eq(chain_id))
+            .select(protocol_component::external_id)
+            .distinct()
```
louise-poole: Don't think 'distinct' is necessary here - there is a table constraint that external_id has to be unique. Duplicates cannot happen.

tycho-storage/src/postgres/orm.rs
```diff
@@ -633,22 +633,42 @@ impl ProtocolState {
...
+        // Apply additional filtering by timestamp if provided
         if let Some(ts) = version_ts {
             query = query.filter(protocol_state::valid_from.le(ts));
         }
 
+        query = query.order_by(protocol_component::external_id);
+
+        // Fetch the results
         query
-            .order_by(protocol_component::external_id)
```
louise-poole: Why move this?

tycho-storage/src/postgres/orm.rs
```diff
@@ -633,22 +633,42 @@ impl ProtocolState {
...
+        pagination_params: Option<&PaginationParams>,
         conn: &mut AsyncPgConnection,
     ) -> QueryResult<Vec<(Self, ComponentId)>> {
-        let mut query = protocol_state::table
-            .inner_join(protocol_component::table)
+        // Subquery to get distinct component external IDs based on pagination
+        let mut component_query = protocol_component::table
             .filter(protocol_component::external_id.eq_any(component_ids))
             .filter(protocol_component::chain_id.eq(chain_id))
+            .select(protocol_component::external_id)
```
louise-poole: Could simplify the later query by selecting `id` here instead of `external_id`. Then the join in the main query is unnecessary.

tycho-storage/src/postgres/orm.rs
```diff
@@ -667,26 +687,46 @@ impl ProtocolState {
...
+
+        // Main query to get ProtocolStates for the selected components
+        let mut query = protocol_state::table
+            .inner_join(
+                protocol_component::table
+                    .on(protocol_state::protocol_component_id.eq(protocol_component::id)),
+            )
+            .filter(protocol_component::id.eq_any(component_query))
             .filter(protocol_state::valid_to.gt(version_ts.unwrap_or(*MAX_VERSION_TS)))
             .into_boxed();
```
louise-poole, line 741: This is an unnecessary join here:
```suggestion
        let mut query = protocol_state::table
            .filter(protocol_state::protocol_component_id.eq_any(component_query))
            .filter(protocol_state::valid_to.gt(version_ts.unwrap_or(*MAX_VERSION_TS)))
            .into_boxed();
```

tycho-storage/src/postgres/orm.rs
```diff
@@ -667,26 +687,46 @@ impl ProtocolState {
...
+        let mut component_query = protocol_component::table
             .inner_join(
                 protocol_system::table
                     .on(protocol_component::protocol_system_id.eq(protocol_system::id)),
             )
-            .filter(protocol_system::name.eq(system.to_string()))
+            .filter(protocol_system::name.eq(system))
             .filter(protocol_component::chain_id.eq(chain_id))
+            .select(protocol_component::id)
+            .distinct()
```
louise-poole: Distinct isn't necessary here either.

tycho-storage/src/postgres/orm.rs
```diff
@@ -703,21 +743,41 @@ impl ProtocolState {
...
+        pagination_params: Option<&PaginationParams>,
         conn: &mut AsyncPgConnection,
     ) -> QueryResult<Vec<(Self, ComponentId)>> {
-        let mut query = protocol_state::table
-            .inner_join(protocol_component::table)
+        // Subquery to get distinct component IDs based on pagination
+        let mut component_query = protocol_component::table
             .filter(protocol_component::chain_id.eq(chain_id))
+            .select(protocol_component::id)
+            .distinct()
```
louise-poole: and here

tycho-storage/src/postgres/orm.rs
```diff
@@ -703,21 +743,41 @@ impl ProtocolState {
...
+
+        if let Some(pagination) = pagination_params {
+            component_query = component_query
+                .limit(pagination.page_size)
+                .offset(pagination.page * pagination.page_size);
+        }
+
+        // Main query to get ProtocolStates for the selected components
+        let mut query = protocol_state::table
+            .inner_join(
```
louise-poole, line 843: Unnecessary join

tycho-storage/src/postgres/protocol.rs
```diff
@@ -52,6 +52,7 @@ impl PostgresGateway {
         let data_vec = states_result
             .map_err(|err| storage_error_from_diesel(err, "ProtocolStates", context, None))?;
 
+        println!("data_vec: {:?}", data_vec);
```
louise-poole: Can we use logging instead of `println`?

tycho-storage/src/postgres/contract.rs
tvinagre: hmm not really.  All these postgres functions are also used internally - one example is when handling a new block on the extractors. I didn't want to make such a radical change, so decided to make this optional. I don't think it's worth paginating these internal requests tbh

tycho-storage/src/postgres/orm.rs
tvinagre: Same issue as above - should be optional to allow non-paginated requests from internal functions

tycho-client/src/rpc.rs
tvinagre: lol tnx

tycho-core/src/dto.rs
tvinagre: I thought about this but decided to keep it like this. Seemed over-engineering to add one default per-chain. Also, keeping it to a low value prevents system abuse

tycho-core/src/dto.rs
tvinagre: I think it's better to leave at least the type explicit, so `PaginationParams::default()` - what do you think?

tycho-storage/src/postgres/orm.rs
tvinagre: hmm, the order_by should be the last thing in the query to ensure consistency, no?

tycho-storage/src/postgres/orm.rs
tvinagre: true - I've mixed the tables, thanks!

tycho-storage/src/postgres/orm.rs
tvinagre: oh yeah fr, don't know why I did that :p tnx!

tycho-storage/src/postgres/protocol.rs
tvinagre: this was for debugging only and I forgot to remove - will delete, tnx!

tycho-storage/src/postgres/orm.rs
louise-poole: Yes. But where you have moved it hasn't changed anything?

tycho-storage/src/postgres/orm.rs
tvinagre: Yeah, I tried to remove but we still need to join, because we return the `protocol_component::external_id` - not `protocol_component::id` 😕

tycho-storage/src/postgres/orm.rs
tvinagre: Sorry - I don't understand. It's still after the version_ts filtering - what am I missing?

tvinagre: Ah, sorry. Yeah this was moved before, when I did the naive pagination on the end-result. Now it doesn't make a difference, you're right - moving back :)

tycho-storage/src/postgres/orm.rs
tvinagre: hmm u sure? We need to join to return the associated external_id for each protocol_state

tycho-storage/src/postgres/orm.rs
tvinagre: same case as the others - not sure we can get rid of the join

tycho-storage/src/postgres/orm.rs
louise-poole: True true... didn't see that! Okay, I have a new suggestion that might work then. The component query can collect both `id` and `external_id` and then the main query can just do a join with the `component_query` to have all the info it needs (no join necessary to the rather large component table then).

tycho-storage/src/postgres/orm.rs
louise-poole: Same suggestion as above then.

tycho-storage/src/postgres/cache.rs
```diff
@@ -707,7 +707,7 @@ impl ContractStateGateway for CachedGateway {
         version: Option<&Version>,
         include_slots: bool,
         pagination_params: Option<&PaginationParams>,
-    ) -> Result<Vec<Contract>, StorageError> {
+    ) -> Result<(i64, Vec<Contract>), StorageError> {
```
louise-poole: Why put the total first? I would expect the value that most use cases will ignore to be the second in the tuple emphasising it's 'extra information'.

tycho-storage/src/postgres/orm.rs
```diff
@@ -762,8 +796,21 @@ impl ProtocolState {
                     .filter(protocol_state::valid_to.gt(*MAX_VERSION_TS))
                     .filter(protocol_state::valid_from.le(ts)),
             ));
+
+            count_query = count_query.filter(exists(
+                protocol_state::table
+                    .filter(protocol_state::protocol_component_id.eq(protocol_component::id))
+                    .filter(protocol_state::valid_to.gt(*MAX_VERSION_TS))
```
louise-poole: the 'valid_to' filters used here are already applied in step 1. This repeated application of it is unnecessary, this if block should only be about applying a valid_from filter.

tycho-storage/src/postgres/orm.rs
```diff
@@ -774,11 +821,18 @@ impl ProtocolState {
...
             .load::<i64>(conn)
-            .await?;
+            .await;
+
+        let component_ids = match component_ids {
+            Ok(ids) => ids,
+            Err(err) => {
+                return (0, Err(err));
+            }
+        };
```
louise-poole: Simplification suggestion:
```
let component_ids = component_ids_query
    .load::<i64>(conn)
    .await
    .map_err(|err| (0, Err(err)))?;
```

tycho-core/src/dto.rs
```diff
@@ -879,6 +893,19 @@ impl Default for PaginationParams {
     }
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, ToSchema)]
+#[serde(deny_unknown_fields)]
+pub struct PaginationResponse {
+    pub page: i64,
+    pub page_size: i64,
+    pub total: i64,
```
zizou0x, line 899: Can we add a docstring here, this one is not clear from the name only

tycho-core/src/storage.rs
```diff
@@ -270,7 +270,8 @@ pub trait ProtocolGateway {
         system: Option<String>,
         ids: Option<&[&str]>,
         min_tvl: Option<f64>,
-    ) -> Result<Vec<models::protocol::ProtocolComponent>, StorageError>;
+        pagination_params: Option<&PaginationParams>,
+    ) -> Result<(i64, Vec<models::protocol::ProtocolComponent>), StorageError>;
```
zizou0x: Maybe use an alias or document the output? It's not clear from the interface what this `i64` is

tycho-storage/src/postgres/cache.rs
```diff
@@ -706,13 +706,14 @@ impl ContractStateGateway for CachedGateway {
         addresses: Option<&[Address]>,
         version: Option<&Version>,
         include_slots: bool,
-    ) -> Result<Vec<Account>, StorageError> {
+        pagination_params: Option<&PaginationParams>,
+    ) -> Result<(i64, Vec<Account>), StorageError> {
```
zizou0x: Same here, given the number of interfaces with this (i64 + something) tuple, we could eventually even consider introducing a wrapper, something like we did for `WithTxHash`.

tycho-indexer/src/services/rpc.rs
```diff
@@ -45,15 +45,15 @@ impl From<anyhow::Error> for RpcError {
 
 pub struct RpcHandler<G> {
     db_gateway: G,
-    pending_deltas: PendingDeltas,
+    pending_deltas: Arc<dyn PendingDeltasBuffer + Send + Sync>,
```
zizou0x, line 52: FMI, why did we make it an `Arc`?

tycho-indexer/src/extractor/evm/hybrid.rs
```diff
@@ -2162,12 +2164,13 @@ mod test_serial_db {
                 .expect("upsert should succeed");
 
             let cached_gw: CachedGateway = gw.state_gateway;
-            let res = cached_gw
+            let (_, res) = cached_gw
```
zizou0x: I dislike that we added the `total_count` everywhere.

The problem is that we now also do this count query everytime and then ignore the result, that seems wrong to me. We should either introduce new functions or find a way to skip the count query if it's not needed

tycho-indexer/src/services/rpc.rs
```diff
@@ -82,30 +82,63 @@ where
...
+                Some(at),
+            )?;
         }
+
+        let total = match addresses {
+            Some(adrs) => {
+                // If contract addresses are specified, the total count is the number of addresses
+                adrs.len() as i64
+            }
+            None => db_total, // TODO: handle case where contract addresses are not specified
```
zizou0x: What is this todo exactly? We are not taking into account buffered entities is that it?

tycho-indexer/src/extractor/evm/hybrid.rs
louise-poole: I will have a look again, but from what I remember the count is mainly just ignored in tests right now

tycho-indexer/src/services/rpc.rs
louise-poole: Yes. Same for protocol_states if ids are not supplied there.

tycho-indexer/src/services/rpc.rs
louise-poole: There are 2 parts to this explanation: 
1) I had to change this to a trait so that I can mock it nicely in tests. Changing it to a trait I had to either Box it or Arc it. 
2) I had to use Arc since the same pending deltas could be used concurrently by multiple endpoints (see [here](https://github.com/propeller-heads/tycho-indexer/blob/main/tycho-indexer/src/services/mod.rs#L139)).

tycho-indexer/src/extractor/evm/hybrid.rs
louise-poole: Hmmm... it is used a few times without needing the total count. I think if pagination is passed as `None` we can skip the count query.

tycho-indexer/src/services/rpc.rs
louise-poole: This is because our buffer only supports fetching buffered states if an id/address is provided.

tycho-storage/src/postgres/orm.rs
```diff
@@ -628,27 +628,30 @@ impl ProtocolState {
...
+            // Count the total number of matching components
+            protocol_component::table
+                .filter(protocol_component::external_id.eq_any(component_ids))
+                .filter(protocol_component::chain_id.eq(chain_id))
+                .count()
+                .get_result::<i64>(conn)
+                .await
+                .unwrap_or(0)
+        } else {
+            0
```
zizou0x: Why not `None`

tycho-indexer/src/services/rpc.rs
zizou0x: I must say that I'm not a big fan of changing this to dynamic dispatch just for testing purpose. Dynamic dispatch must be used carefully because it's one (if not the biggest) thing that can make rust slow.

Can't we use generics or something instead? I'm not saying we need to do it now but most likely we'll need to refactor this, maybe add a TODO

tycho-indexer/src/services/rpc.rs
zizou0x: Let's add a todo+jira task to not forget about that?

louise-poole: There's already a todo in the buffer's code. I'll add a jira task 👍

tycho-indexer/src/services/rpc.rs
louise-poole: I'll add a TODO 👍

PR #346, opened 2024-08-30
feat: Remove chain from contract id param

louise-poole:
**Issue**: chain is passed in multiple places for a contract_state request: as a top level param, and with each contract id passed. 
**Solution**: remove it from the contract ids struct. We only need the top level param.

**Design decision**: I implemented a custom deserialiser to allow for backwards compatibility with the old contract ids input type. This should be removed in a future version.

### DIFF
--- a/tycho-core/src/dto.rs
+++ b/tycho-core/src/dto.rs
@@ -26,8 +26,7 @@ use crate::{
         token::CurrencyToken,
     },
     serde_primitives::{
-        hex_bytes, hex_bytes_option, hex_bytes_vec, hex_hashmap_key, hex_hashmap_key_value,
-        hex_hashmap_value,
+        hex_bytes, hex_bytes_option, hex_hashmap_key, hex_hashmap_key_value, hex_hashmap_value,
     },
     Bytes,
 };
@@ -408,10 +407,8 @@ pub struct ProtocolComponent {
     pub protocol_system: String,
     pub protocol_type_name: String,
     pub chain: Chain,
-    #[serde(with = "hex_bytes_vec")]
     #[schema(value_type=Vec<String>)]
     pub tokens: Vec<Bytes>,
-    #[serde(with = "hex_bytes_vec")]
     #[schema(value_type=Vec<String>)]
     pub contract_ids: Vec<Bytes>,
     #[serde(with = "hex_hashmap_value")]
--- a/tycho-core/src/serde_primitives.rs
+++ b/tycho-core/src/serde_primitives.rs
@@ -79,44 +79,6 @@ pub mod hex_bytes_option {
     }
 }
 
-/// serde functions for handling Vec of bytes
-pub mod hex_bytes_vec {
-    use super::decode_hex_with_prefix;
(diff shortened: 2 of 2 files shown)


tycho-client/src/feed/synchronizer.rs
```diff
@@ -247,7 +247,6 @@ where
                         contract_ids
                             .clone()
                             .into_iter()
-                            .map(|id| ContractId::new(self.extractor_id.chain, id))
```
zizou0x, line 250: You can remove `.into_iter().collect()` I think

tycho-client-py/tycho_indexer_client/dto.py
```diff
@@ -272,10 +272,29 @@ class Config:
...
     version: Optional[VersionParams] = None
 
+    # Backward compatibility with old ContractId format
+    # To be removed in the future
+    def __init__(
+        self,
+        contract_ids: Optional[Union[List[ContractId], List[str]]] = None,
+        protocol_system: Optional[str] = None,
+        version: Optional[VersionParams] = None,
+        chain: Optional[str] = None,
```
zizou0x: Is it really Optional?

tycho-core/src/dto.rs
```diff
@@ -513,11 +516,12 @@ impl ProtocolStateDelta {
...
 }
 
-#[derive(Serialize, Deserialize, Debug, Default, PartialEq, ToSchema)]
+#[derive(Serialize, Debug, Default, PartialEq, ToSchema)]
 #[serde(deny_unknown_fields)]
 pub struct StateRequestBody {
-    #[serde(alias = "contractIds")]
-    pub contract_ids: Option<Vec<ContractId>>,
+    #[serde(alias = "contractIds", with = "hex_bytes_vec_option")]
+    #[schema(value_type=String)]
```
zizou0x: It doesn't work with `#[schema(value_type=Option<Vec<String>>)]`?

Then I'm not sure you would need `hex_bytes_vec_option` either

tycho-client-py/tycho_indexer_client/dto.py
louise-poole: Actually shouldn't even be here 👀

tycho-client/src/feed/synchronizer.rs
louise-poole: Unfortunately not.. contract_ids is a Hashset here and needs to be a Vec to be passed.

tycho-core/src/dto.rs
louise-poole: Interesting 👀 then why do we have all of these custom hex bytes serialisation functions that we use?

zizou0x: Which ones do you mean? I'm not sure it works but I don't see why it wouldn't

louise-poole: It works for the test so will do it like you suggested. However for all hashmaps etc we have custom logic such as `hex_hashmap_key_value` etc. I assume this is hashmap specific then...

PR #349, opened 2024-09-02
feat: Automate removal of orphaned transactions from the DB

louise-poole:


tycho-storage/migrations/2024-09-02_transaction_cleanup/up.sql
```diff
@@ -0,0 +1,30 @@
...
+    )
+    AND NOT EXISTS (
+        SELECT 1 FROM account_balance ab WHERE ab.modify_tx = t.id 
+    );
+
+END;
+$$ LANGUAGE plpgsql;
+
+-- Schedule the cleanup function to run every day at midnight
+SELECT cron.schedule('clean_transaction_table', '0 0 * * *', 'SELECT clean_transaction_table();');
```
zizou0x: Do you know how long it takes to run? If it's too slow this might affect Tycho

louise-poole: It's super slow - at least on the first run it is. Still waiting for it to finish (started running it this morning).

zizou0x: Then I'm not sure we want to add this cronjob... Maybe just make a script for now, until we find a better long term solution?

louise-poole: I improved it by deleting in batches and only triggering the cronjob once a week.

zizou0x: Ok sounds good, let's hope it works nicely

tycho-storage/migrations/2024-09-02_transaction_cleanup/up.sql
```diff
@@ -0,0 +1,70 @@
...
+    LEFT JOIN contract_storage cs ON cs.modify_tx = t2.id
+    WHERE cc.modify_tx IS NULL
+      AND pc.creation_tx IS NULL
+      AND pc.deletion_tx IS NULL
+      AND a.creation_tx IS NULL
+      AND a.deletion_tx IS NULL
+      AND ab.modify_tx IS NULL
+      AND cb.modify_tx IS NULL
+      AND ps.modify_tx IS NULL
+      AND cs.modify_tx IS NULL;
```
zizou0x: Won't this take forever on a big table?

tycho-storage/migrations/2024-09-02_transaction_cleanup/up.sql
```diff
@@ -0,0 +1,70 @@
...
+            EXIT;
+        END IF;
+
+        -- Remove the deleted rows from the temporary table
+        DELETE FROM temp_transaction_ids
+        WHERE id IN (
+            SELECT id
+            FROM temp_transaction_ids
+            LIMIT batch_size
+        );
```
zizou0x: Wouldn't it be fast enough to just drop the table here?

tycho-storage/migrations/2024-09-02_transaction_cleanup/up.sql
louise-poole: Yes - but much quicker than the previous attempt (using NOT EXISTS)

tycho-storage/migrations/2024-09-02_transaction_cleanup/up.sql
louise-poole: No - here we are only deleting the ids of the batch we've just processed. We don't want to drop the whole table yet

PR #350, opened 2024-09-02
refactor: move token analysis related structs into Ethereum specific crate

zizou0x:
This PR aims to move our token analysis structs into the ethereum tycho crate (currently called token-analyzer). This allowed us to remove many `ethers` dependency from `tycho-indexer`.

The main goal of this PR isn't to make token analysis flexible between chains, but during this refactor we also introduced some traits for token analysis in tycho-core. This is the first step toward making token-analysis fully modulable across blockchains.


token-analyzer/src/abi/erc20.json
tvinagre, line 1: This ABI is still `evm` specific, no? I understood you've moved it to `token-analyzer`, but maybe it's worth keeping the `evm` subfolder for things that are evm-specific?

token-analyzer/src/token_pre_processor.rs
```diff
@@ -1,47 +1,57 @@
...
-        addresses: Vec<H160>,
-        token_finder: Arc<dyn TokenOwnerFinding>,
-        block: BlockNumber,
-    ) -> Vec<CurrencyToken>;
-}
 
 const ABI_STR: &str = include_str!("./abi/erc20.json");
 
-impl TokenPreProcessor {
+impl EthereumTokenPreProcessor {
```
tvinagre: Not sure if you have, but would be nice to document the node requirements for the preprocessor

token-analyzer/src/token_pre_processor.rs
```diff
@@ -51,34 +61,36 @@ impl TokenPreProcessor {
 }
 
 /// Map a protocol system into its vault
-pub fn map_vault(protocol_system: &str) -> Option<H160> {
+/// TODO: This is a hack until we can use the `balance_owner` attribute. Needs to be fixed once we
+/// emit this attribute for every protocol in Substreams
```
louise-poole: From what I understand, we will emit this attribute only for protocols whose balances are not owned by the component itself. So protocols such as usv2 etc will not be expected to emit a balance owner.

tycho-core/src/models/blockchain.rs
```diff
@@ -308,3 +308,19 @@ impl From<ProtocolChangesWithTx> for TxWithChanges {
         }
     }
 }
+
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum BlockTag {
+    /// Finalized block
+    Finalized,
+    /// Safe block
+    Safe,
```
louise-poole, line 317: FMI: what's a 'safe' block?

tycho-core/src/models/token.rs
```diff
@@ -40,3 +50,42 @@ impl CurrencyToken {
         }
     }
 }
+
+/// How well behaved a token is.
```
louise-poole: I'm not that keen on this doc. What makes a token 'well behaved'? I think this could be clearer.

tycho-core/src/traits.rs
```diff
@@ -17,3 +23,44 @@ pub trait AccountExtractor {
         account_addresses: Vec<Address>,
     ) -> Result<HashMap<Bytes, AccountDelta>, Self::Error>; //TODO: do not return `AccountUpdate` but `Account`
 }
+
+/// Analyze how well behaved a token is.
```
louise-poole: Also here. 'Well behaved' makes me think of children, not tokens.

token-analyzer/src/token_pre_processor.rs
zizou0x: Yeah of course, it's not really clear from the comment, good catch

tycho-core/src/models/blockchain.rs
zizou0x: It's a status on Ethereum consensus, not sure if it's applicable to every chain, but it's basically right before finalized (which might also not make sense in the context of single block finality). So basically this enum is currently EVM focused, we might need to adapt it later but I think it's alright for now.

You can read more on this [here](https://docs.tatum.io/docs/evm-block-finality-and-confidence#block-tags-and-their-meanings)

token-analyzer/src/abi/erc20.json
kayibal: There should not be anything evm specific anymore in tycho - we will rename or remove the evm folder

token-analyzer/src/token_pre_processor.rs
kayibal: Do we still need this map_vault thing?

token-analyzer/src/token_pre_processor.rs
```diff
@@ -168,21 +176,27 @@ mod tests {
         ));
         let w3 = Web3::new(transport);
 
-        let processor = TokenPreProcessor::new(client, w3, Chain::Ethereum);
+        let processor = EthereumTokenPreProcessor::new(client, w3, Chain::Ethereum);
 
-        let tf = TokenFinder::new(HashMap::new());
+        let tf = TokenFinderStore::new(HashMap::new());
```
kayibal: This finds owners no? It's currently not clear from the name...

tycho-core/src/hex_bytes.rs
```diff
@@ -149,6 +150,28 @@ impl Bytes {
...
+    ///
+    /// A `Bytes` object of the specified length, filled with random bytes.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// let random_bytes = Bytes::random(5);
+    /// assert_eq!(random_bytes.len(), 5);
+    /// ```
+    pub fn random(length: usize) -> Bytes {
```
kayibal, line 170: I would only add this for test compiles and move rand to test dependencies...

tycho-core/src/models/token.rs
```diff
@@ -1,13 +1,23 @@
...
 use serde::{Deserialize, Serialize};
 
+use super::{Address, Balance};
+
+/// Cost related to a token transfer, for example amount of gas in evm chains.
+pub type TransferCost = u64;
+
+/// Tax related to a token transfer. Should be given in Basis Points (1/100th of a percent)
+pub type TransferTax = u64;
+
```
kayibal, line 13: Idk having type alias for u64 makes something in me unhappy. I understand that it makes the interfaces below nicer but potentially just having very nice docs on the interfaces is more useful. You can move the docs here directly to the CurrencyToken attributes..

token-analyzer/src/token_pre_processor.rs
zizou0x: Unfortunately yes, until we can change Ambient Substreams to make it return a `balance_owner`

tycho-core/src/hex_bytes.rs
zizou0x: Sadly it's not possible because it's used in the `initialize_accounts` to link DCI account with a random transaction hash

tycho-core/src/models/token.rs
zizou0x: Interesting, the main reason why I introduced these aliases wasn't the struct below but [this trait interface](https://github.com/propeller-heads/tycho-indexer/blob/d7ebd0585d8df8de62d08511bd52de3bf08ba879/tycho-core/src/traits.rs#L34-L38) which has always been confusing.

I'm not against having it documented instead but fmi why don't you like aliases for `u64` in particular, is there a particular reason?

zizou0x: Alternatively we could introduce a token analysis results struct instead of a tuple

kayibal: The reason I don't like it is that if I see that in an interface, I expect something fancy so I have to go look it up and then just find u64... So I could have saved myself that lookup. It's not a big deal. I think the best would be to just properly document the return type - either via interface docs (preferred) or a struct (preferred if you expect a lot of data changes or additional behaviour on the type in the future)

PR #351, opened 2024-09-03
refactor: remove bytes padding

zizou0x:
This PR remove utils function that were used for bytes padding. The idea is that Tycho shouldn't do any bytes length assumption, this should be handled by Substreams.

### DIFF
--- a/tycho-indexer/Cargo.toml
+++ b/tycho-indexer/Cargo.toml
@@ -21,7 +21,6 @@ tracing.workspace = true
 async-trait.workspace = true
 tokio.workspace = true
 console-subscriber.workspace = true
-ethers.workspace = true
 diesel-async.workspace = true
 tycho-core = { workspace = true }
 tycho-storage.workspace = true
@@ -65,6 +64,7 @@ num-bigint = "0.4.4"
 num-traits = "0.2.19"
 
 [dev-dependencies]
+ethers.workspace = true
 actix-rt = "2.9.0"
 actix-test = "0.1.2"
 rstest = "0.18.2"
--- a/tycho-indexer/src/extractor/u256_num.rs
+++ b/tycho-indexer/src/extractor/u256_num.rs
@@ -5,7 +5,8 @@
 
 use std::{cmp::max, panic};
 
-use ethers::types::U256;
+use num_bigint::BigInt;
+use num_traits::{One, ToPrimitive, Zero};
 use tracing::warn;
 
 /// Converts a U256 integer into it's closest floating point representation
@@ -36,9 +37,9 @@ pub fn bytes_to_f64(data: &[u8]) -> Option<f64> {
         warn!(?data, "Received invalid balance bytes!");
         return None;
     }
-    let x = U256::from(data);
+    let x = BigInt::from_bytes_be(num_bigint::Sign::Plus, data);
     let res = panic::catch_unwind(|| {
-        if x == U256::zero() {
+        if x == BigInt::zero() {
             return Some(0.0);
         }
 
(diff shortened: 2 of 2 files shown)


tycho-indexer/src/extractor/evm/hybrid.rs
```diff
@@ -184,13 +183,15 @@ where
...
+            let hours = time_remaining.num_hours();
+            let minutes = (time_remaining.num_minutes()) % 60;
             info!(
                 extractor_id = self.name,
                 blocks_per_minute = format!("{blocks_per_minute:.2}"),
                 blocks_processed,
                 height = block.number,
                 current = current_block,
-                time_remaining = format_duration(&time_remaining),
+                time_remaining = format!("{:02}h{:02}m", hours, minutes),
```
louise-poole, line 194: Was this necessary? Wasn't this already formatted well? Looking back on old logs I see it displayed as
```
extractor_id="uniswap_v2" blocks_per_minute="7816.00" blocks_processed=7816 height=10523234 current=19249015 time_remaining="18h36m" name="SyncProgress"
```

tycho-indexer/src/extractor/evm/mod.rs
```diff
@@ -50,26 +48,12 @@ impl TryFromMessage for AccountDelta {
         let change = msg.change().into();
         let update = AccountDelta::new(
             chain,
-            pad_and_parse_h160(&msg.address.into()) //TODO: we shouldn't pad here, this is not blockchain agnostic
-                .map_err(ExtractionError::DecodeError)?
-                .into(),
+            msg.address.into(),
```
louise-poole, line 51: We need to go double check that the substreams are padding correctly.

zizou0x: I think they do, also I would say that ideally it's a shared responsibility between Substreams and Protosim (and other client). Imo they should be done in a way that they don't break if i.e. a slot or a balance (uint256) isn't left padded (revm should know that it needs 32 bytes for them).

But in a 100% ideal world, every Substreams return exactly what's needed by the vm and we don't need to worry about it.

tycho-indexer/src/extractor/evm/hybrid.rs
zizou0x: It won't change anything in the log, here I just removed the `format_duration()` utils function and adapted this code to do exactly what it was doing. It was used only once in the whole codebase so I think a function wasn't necessary.

tycho-indexer/src/extractor/evm/hybrid.rs
```diff
@@ -2633,7 +2633,8 @@ mod test_serial_db {
                 component_tvl: HashMap::new(),
                 state_deltas: Default::default(),
             };
-
+            dbg!(&res);
+            dbg!(&block_account_expected);
```
zizou0x: TODO: remove

tycho-indexer/src/extractor/evm/mod.rs
kayibal: We may later add support for checks on the extractor. E.g. the extractor knows that an address on evm is 20 bytes. It also knows which fields of the message should be addresses and can check for that. 

So I think the "last check" responsibility of this should be on the indexer, not on protosim. But how we actually do this should be just a length check - not actually parsing bytes into a H160 struct. So we could do sth like:




```rust



struct ChainValidationContext {
    address_length: usize
    slot_key_length: usize
    slot_value_length: usize
}


struct BlockChanges {
  // ...

  pub validate_lengths(&self, chain_context: ChainValidationContext) {
     ...
  
  }

}
```

zizou0x: True, we might also need something similar for how to decode them, for example currently it's hex encoded. But on Solana we will most likely need utf-8 (because addresses have letters).

PR #356, opened 2024-09-05
fix: Delete protocol system script to delete tokens as necessary

louise-poole:
**Issue**: some tokens are also pools. If we try delete those pools we get an error because it is still being referenced as a token.
**Solution**: we should first delete all tokens that belong solely to pools of the protocol system we're deleting. Then we can delete the pools.

### DIFF
--- a/tycho-storage/sql_scripts/remove_protocol_system.sh
+++ b/tycho-storage/sql_scripts/remove_protocol_system.sh
@@ -81,8 +81,7 @@ JOIN protocol_component_holds_token pcht ON t.id = pcht.token_id
 JOIN protocol_component pc ON pcht.protocol_component_id = pc.id
 JOIN protocol_system ps ON pc.protocol_system_id = ps.id
 GROUP BY t.symbol, a.address, t.id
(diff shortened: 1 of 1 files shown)


tycho-storage/sql_scripts/remove_protocol_system.sh
```diff
@@ -72,6 +72,17 @@ AND NOT EXISTS (
...
+--- List of tokens to be deleted
+SELECT t.symbol AS "Token Symbol", '0x' || encode(a.address::bytea, 'hex') AS "Token Address", t.id AS "Token ID"
+FROM token t
+JOIN account a ON t.account_id = a.id
+JOIN protocol_component_holds_token pcht ON t.id = pcht.token_id
+JOIN protocol_component pc ON pcht.protocol_component_id = pc.id
+JOIN protocol_system ps ON pc.protocol_system_id = ps.id
+GROUP BY t.symbol, a.address, t.id
+HAVING COUNT(DISTINCT CASE WHEN ps.name <> :'protocol_system_name' THEN ps.id END) = 0
+AND COUNT(DISTINCT ps.id) > 0;
```
zizou0x: Not sure this is needed

PR #357, opened 2024-09-05
docs: Create Tycho rustbook

louise-poole:
To view locally:
- install mdbook
- on the terminal run `mdbook build` and then `mdbook serve`
- on your browser go to `localhost:3000` and ENJOY

(do we need a README with this info? Or is that excessive?)

---

Open to suggestions on renaming/moving things around!

TODO:
- update Usage Options sections [pending actual details on how to connect, use the api key etc]
- update Self-Hosting Tycho [for when tycho is open-sourced]
- add to github pages


docs/src/tycho-as-a-service.md
kayibal, line 1: Seems like we are a bit ahead here of what is actually possible?

PR #370, opened 2024-09-18
fix: fetch contracts from deltas buffer if not in db yet

louise-poole:
Issue: We had a bug where new contracts that exist only in the deltas buffer (not been sent to the db yet) were not returned by the rpc when requested. This is because the rpc would apply the deltas buffer on top of fetched db states only, it did not create a new state for a contract that did not exist in the db yet.

### DIFF
--- a/tycho-indexer/src/services/deltas_buffer.rs
+++ b/tycho-indexer/src/services/deltas_buffer.rs
@@ -170,12 +170,12 @@ impl PendingDeltas {
         Ok(change_found)
     }
 
-    /// Updates the given db states with the buffered deltas. If a requested contract is not in the
+    /// Updates the given db states with the buffered deltas. If a requested account is not in the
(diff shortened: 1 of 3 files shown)


tycho-indexer/src/services/deltas_buffer.rs
```diff
@@ -212,6 +241,34 @@ impl PendingDeltas {
...
+            for entry in guard.get_block_range(None, version)? {
+                if let Some(delta) = entry.account_deltas.get(&address) {
+                    // Update contract state or create a new one if not present
+                    let contract_ref =
+                        contract.get_or_insert_with(|| delta.clone().into_contract());
+                    contract_ref.apply_contract_delta(delta)?;
+                }
+            }
+        }
+        dbg!(&contract);
```
kayibal: I think you missed this

PR #372, opened 2024-09-19
refactor(token-cronjob): make the cronjob retry 5 times for each token

zizou0x:
It now progressively decrease token quality by one on each iteration (from 10 to 5).

### DIFF
--- a/tycho-client/README.md
+++ b/tycho-client/README.md
@@ -111,10 +111,10 @@ This will stream updates for all components whose TVL exceeds the minimum thresh
 
   2) Set a ranged TVL boundary:
 ```bash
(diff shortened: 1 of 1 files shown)


tycho-indexer/src/extractor/evm/token_analysis_cron.rs
```diff
@@ -130,8 +130,8 @@ async fn analyze_batch(
 
         if let TokenQuality::Bad { reason } = token_quality {
             debug!(?t.address, ?reason, "Token quality detected as bad!");
-            // Don't try to analyze again.
-            t.quality = 5;
+            // Remove 1 to the quality for each attempt. If it fails 5 times we won't try again.
+            t.quality -= 1;
```
kayibal, line 134: OK somewhat hacky but whatever

PR #373, opened 2024-09-20
refactor(tycho-indexer): remove chain state hack

zizou0x:
This PR removes the hack that we did previously with getting the syncing state using `ChainState`.

`ChainState` gives us an approximation of the current block header. It makes an assumption on the blocktime which can lead to difference with the blockchain true state. Using this for report progress led to weird logs (we were getting this report progress logs even though the extractor was synced).


tycho-indexer/src/extractor/evm/hybrid.rs
```diff
@@ -612,7 +591,8 @@ where
         self.update_last_processed_block(msg.block.clone())
             .await;
 
-        self.report_progress(&msg.block).await;
+        self.report_progress(&msg.block, is_syncing, inp.final_block_height)
```
louise-poole: Instead of passing is_syncing as a param, we can instead check if it's syncing here already and only call `report_progress` if it is.

PR #374, opened 2024-09-20
fix: Remove deleted attributes from default table

louise-poole:
Issue: deleted attributes were not handled correctly by our versioning logic. We would mark them for archiving with an updated valid_to value (correct behaviour), but not remove the old entry in the default table (so a 'valid' entry still existed for the deleted attribute).

Solution: remove deleted attributes from the default table.

### DIFF
--- a/tycho-storage/src/postgres/protocol.rs
+++ b/tycho-storage/src/postgres/protocol.rs
@@ -818,19 +818,21 @@ impl PostgresGateway {
                 .await
                 .map_err(PostgresError::from)?;
             // remove deleted attributes from the default table
-            let mut delete_query =
-                diesel::delete(schema::protocol_state_default::table).into_boxed();
-            for ((component_id, attr_name), _) in deleted_attributes {
-                delete_query = delete_query.or_filter(
-                    schema::protocol_state_default::protocol_component_id
-                        .eq(component_id)
-                        .and(schema::protocol_state_default::attribute_name.eq(attr_name)),
-                );
+            if !deleted_attributes.is_empty() {
+                let mut delete_query =
+                    diesel::delete(schema::protocol_state_default::table).into_boxed();
+                for ((component_id, attr_name), _) in deleted_attributes {
+                    delete_query = delete_query.or_filter(
+                        schema::protocol_state_default::protocol_component_id
+                            .eq(component_id)
+                            .and(schema::protocol_state_default::attribute_name.eq(attr_name)),
+                    );
+                }
+                delete_query
+                    .execute(conn)
+                    .await
+                    .map_err(PostgresError::from)?;
             }
-            delete_query
-                .execute(conn)
-                .await
-                .map_err(PostgresError::from)?;
         }
         Ok(())
     }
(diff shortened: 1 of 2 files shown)


tycho-storage/src/postgres/protocol.rs
```diff
@@ -2053,7 +2067,7 @@ mod test {
 
         // set up deletable attribute state
         let protocol_component_id = schema::protocol_component::table
-            .filter(schema::protocol_component::external_id.eq("state2"))
+            .filter(schema::protocol_component::external_id.eq("state3"))
```
zizou0x, line 2072: Why was this needed?

tycho-storage/src/postgres/protocol.rs
```diff
@@ -2181,6 +2195,17 @@ mod test {
...
+        let deleted_state = schema::protocol_state::table
+            .inner_join(schema::protocol_component::table)
+            .filter(schema::protocol_component::external_id.eq(new_state2.component_id.as_str()))
+            .filter(schema::protocol_state::attribute_name.eq("deletable"))
+            .select(orm::ProtocolState::as_select())
+            .first::<orm::ProtocolState>(&mut conn)
+            .await
+            .expect("Failed to fetch protocol state");
+        assert_eq!(deleted_state.valid_to, older_state.valid_to);
     }
```
zizou0x, line 2211: How does this work exactly? If here the valid_to is set to older_state it means it was removed from the default table?

tycho-storage/src/postgres/protocol.rs
```diff
@@ -817,6 +817,20 @@ impl PostgresGateway {
...
+            let mut delete_query =
+                diesel::delete(schema::protocol_state_default::table).into_boxed();
+            for ((component_id, attr_name), _) in deleted_attributes {
+                delete_query = delete_query.or_filter(
+                    schema::protocol_state_default::protocol_component_id
+                        .eq(component_id)
+                        .and(schema::protocol_state_default::attribute_name.eq(attr_name)),
+                );
+            }
+            delete_query
```
zizou0x: So we correctly add deleted attributes to the partitions already?

tycho-storage/src/postgres/protocol.rs
louise-poole: So that I can insert a protocol state directly into the test db (don't want to do it through a gateway as I'm not testing this insert functionality here). To insert the state directly I need to first fetch the db id for any foreign keys (such as the component that was added during test db set up).

tycho-storage/src/postgres/protocol.rs
louise-poole: Basically yes. Only currently valid states are added to the default table. Our versioning logic checks all updated states and sends ones that are no longer valid to the archive tables. 
Previously our bug was that it was doing this fine (the newly outdated state was sent to the archive table), but we weren't removing the original from the default table... so a 'valid' version still existed, as well as the outdated version. 
For this test assert: if this attribute was still in the default table we'd have a currently valid version returned here, not an outdated one.

tycho-storage/src/postgres/protocol.rs
louise-poole: Yes. The bug was that it existed in 2 places: the archive table with the correct versioning applied, and the default table as 'still valid'.

PR #378, opened 2024-09-24
feat(rpc): Protocol state and contract storage rpc caching

kayibal:
Generalizes caching and implements it for protocol state and contract storage.

### DIFF
--- a/tycho-indexer/src/services/rpc.rs
+++ b/tycho-indexer/src/services/rpc.rs
@@ -49,6 +49,8 @@ pub struct RpcHandler<G> {
     pending_deltas: PendingDeltas,
     token_cache: RpcCache<dto::TokensRequestBody, dto::TokensRequestResponse>,
     contract_storage_cache: RpcCache<dto::StateRequestBody, dto::StateRequestResponse>,
+    protocol_state_cache:
+        RpcCache<dto::ProtocolStateRequestBody, dto::ProtocolStateRequestResponse>,
 }
 
 impl<G> RpcHandler<G>
@@ -69,7 +71,18 @@ where
                 7 * 60,
             );
 
-        Self { db_gateway, pending_deltas, token_cache, contract_storage_cache }
+        let protocol_state_cache = RpcCache::<
+            dto::ProtocolStateRequestBody,
+            dto::ProtocolStateRequestResponse,
+        >::new("contract_storage", 50, 7 * 60);
+
+        Self {
+            db_gateway,
+            pending_deltas,
+            token_cache,
+            contract_storage_cache,
+            protocol_state_cache,
+        }
     }
 
     #[instrument(skip(self, request))]
@@ -208,13 +221,18 @@ where
         request: &dto::ProtocolStateRequestBody,
     ) -> Result<dto::ProtocolStateRequestResponse, RpcError> {
         debug!(?request, "Getting protocol state.");
-        self.get_protocol_state_inner(request)
+        self.protocol_state_cache
+            .get(request.clone(), |r| async {
+                self.get_protocol_state_inner(r)
+                    .await
+                    .map(|res| (res, true))
+            })
             .await
     }
 
     async fn get_protocol_state_inner(
         &self,
-        request: &dto::ProtocolStateRequestBody,
+        request: dto::ProtocolStateRequestBody,
     ) -> Result<dto::ProtocolStateRequestResponse, RpcError> {
         //TODO: handle when no id is specified with filters
         let at = BlockOrTimestamp::try_from(&request.version)?;
--- a/tycho-core/src/dto.rs
+++ b/tycho-core/src/dto.rs
@@ -1012,7 +1012,7 @@ fn default_include_balances_flag() -> bool {
     true
 }
 
-#[derive(Debug, Serialize, Deserialize, PartialEq, ToSchema, Default, Eq, Hash)]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, ToSchema, Default, Eq, Hash)]
 #[serde(deny_unknown_fields)]
 pub struct ProtocolStateRequestBody {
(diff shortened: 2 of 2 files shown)


tycho-indexer/src/services/cache.rs
```diff
@@ -0,0 +1,121 @@
...
+                .time_to_live(std::time::Duration::from_secs(ttl))
+                .build(),
+        ));
+        Self { name: name.to_string(), cache }
+    }
+    pub async fn get<
+        'a,
+        E: Error,
+        Fut: Future<Output = Result<(V, bool), E>> + Send + 'a,
+        F: Fn(R) -> Fut + Send + Sync,
```
zizou0x, line 30: Isn't it `Send` twice? 

Also can't we simplify this?

```suggestion
        F: Fn(R) -> impl Future<Output = Result<(V, bool), E>> + Send + Sync + 'a,
```

kayibal: You find yours simpler ??? :D

tycho-indexer/src/services/cache.rs
```diff
@@ -0,0 +1,122 @@
...
+    >(
+        &'a self,
+        request: R,
+        fallback: F,
+    ) -> Result<V, E> {
+        // Cache entry count is only used for logging purposes
+        #[allow(unused_assignments)]
```
louise-poole, line 36: Why do we need this here?

tycho-indexer/src/services/rpc.rs
```diff
@@ -61,7 +62,14 @@ where
             7 * 60,
         );
 
-        Self { db_gateway, pending_deltas, token_cache }
+        let contract_storage_cache =
+            RpcCache::<dto::StateRequestBody, dto::StateRequestResponse>::new(
+                "contract_storage",
+                50,
+                7 * 60,
```
louise-poole, line 71: I think we need a shorter ttl for contract storage. 7 mins is a bit long - the db values could have been updated many times during then.

tycho-indexer/src/services/rpc.rs
```diff
@@ -69,7 +71,18 @@ where
                 7 * 60,
             );
 
-        Self { db_gateway, pending_deltas, token_cache, contract_storage_cache }
+        let protocol_state_cache = RpcCache::<
+            dto::ProtocolStateRequestBody,
+            dto::ProtocolStateRequestResponse,
+        >::new("contract_storage", 50, 7 * 60);
```
louise-poole, line 77: This should be `state_storage` no? Also should have a shorter ttl. Ideally the cache should invalidate whenever we write to the db (if the cache holds only db data which I suspect it doesn't)... but obviously we don't have access to that info here. Not sure what the ideal ttl is.

tycho-indexer/src/services/cache.rs
```diff
@@ -0,0 +1,121 @@
...
+        trace!(?request, cache_size = cache_entry_count, cache = self.name, "CacheMiss");
+        // Acquire a write lock before querying the database (prevents concurrent db queries)
+        let write_lock = self.cache.write().await;
+
+        // Double-check if another thread has already fetched and cached the data
+        if let Some(cached_response) = write_lock.get(&request) {
+            return Ok(cached_response);
+        }
+
+        let (response, should_cache) = (fallback)(request.clone()).await?;
```
louise-poole, line 58: I'm a bit concerned with the caching logic... right now we have buffered data included in the response for protocol_states and contract_storage. Ideally we don't want to cache this buffered data at all, we should cache the db data only so that the cache is viable for longer. Otherwise we need to invalidate the cache on every block because the buffer values update every block.

tycho-indexer/src/services/rpc.rs
kayibal: I am assuming that you have to pass a version/block here.

tycho-indexer/src/services/rpc.rs
kayibal: Same answer as above we should probably only use the cache if a version/block was passed. Double checked this and users have to pass a version or a block - so we are safe here.

tycho-indexer/src/services/cache.rs
kayibal: I think for our use case the fact that we store data that might revert is not a big deal since in this case the client will be also receiving the revert event and can construct the correct state from it.

tycho-indexer/src/services/cache.rs
kayibal:
```
warning: value assigned to `cache_entry_count` is never read
  --> tycho-indexer/src/services/cache.rs:36:17
   |
36 |         let mut cache_entry_count: u64 = 0;
   |                 ^^^^^^^^^^^^^^^^^
   |
   = help: maybe it is overwritten before being read?
   = note: `#[warn(unused_assignments)]` on by default

```

kayibal: This was already present before... so I didn't change the logic here.

louise-poole: Weird - and what about removing the assignment entirely if it's never used?

tycho-indexer/src/services/rpc.rs
louise-poole: Ah - yes, true.

tycho-indexer/src/services/cache.rs
kayibal: It is used in a macro

tycho-indexer/src/services/rpc.rs
louise-poole: We're safe here - the version param defaults to now() if none was provided. Our client supplies a block for this param so for the most part the cache will work as intended. The cache will be useless for requests that use the default version (it's a timestamp so requests made even a second apart will never result in a cache hit).

PR #379, opened 2024-09-26
feat(tycho-client): add auth key and support for https

zizou0x:
In this PR we added an optional authorization key that can be passed as either env variable or to the CLI. If given, this key will then be used as "Authorizer" in the request headers for both ws and http. 

We also added some other headers such as a user-agent.

Note: it should still allow us to connect from inside the cluster. It shouldn't break backward compatibility.

Note 2: this PR move our http client from `hyper` to `reqwest`. The later is a wrapper around `hyper`. They use a relatively equivalent interface and `reqwest` comes with some complexity abstraction such as handling both https and http.

### DIFF
--- a/tycho-client-py/tycho_indexer_client/stream.py
+++ b/tycho-client-py/tycho_indexer_client/stream.py
@@ -30,6 +30,7 @@ class TychoStream:
         include_state=True,
         logs_directory: str = None,
         tycho_client_path: str = None,
+        use_tls: bool = True
     ):
         """
         Initializes the TychoStream instance.
@@ -45,6 +46,7 @@ class TychoStream:
             include_state: Whether to include protocol states in the stream. Defaults to True.
             logs_directory: Directory to store log files. If not specified, a default directory based on the OS is used.
             tycho_client_path: Path to the Tycho client binary. If not specified, the binary is searched for in the system's PATH.
+            use_tls: Whether to use TLS connections with `tycho_url` or not. Defaults to `True`.
         """
         self.tycho_url = tycho_url
         self.min_tvl = min_tvl
@@ -55,6 +57,7 @@ class TychoStream:
         self._blockchain = blockchain
         self._logs_directory = logs_directory or get_default_log_directory()
         self._tycho_client_path = tycho_client_path or find_tycho_client()
+        self._use_tls = use_tls
 
     async def start(self):
         """Start the tycho-client Rust binary through subprocess"""
@@ -77,6 +80,9 @@ class TychoStream:
 
         if not self._include_state:
             cmd.append("--no-state")
+            
+        if not self._use_tls:
+            cmd.append("--no-tls")
 
         for exchange in self.exchanges:
             cmd.append("--exchange")


tycho-client/src/main.rs
```diff
@@ -154,9 +159,9 @@ async fn main() {
 
 #[allow(deprecated)]
 async fn run(exchanges: Vec<(String, Option<String>)>, args: CliArgs) {
-    let tycho_ws_url = format!("ws://{}", &args.tycho_url);
-    let tycho_rpc_url = format!("http://{}", &args.tycho_url);
-    let ws_client = WsDeltasClient::new(&tycho_ws_url).unwrap();
+    let tycho_ws_url = format!("wss://{}", &args.tycho_url);
+    let tycho_rpc_url = format!("https://{}", &args.tycho_url);
```
zizou0x: Surprisingly this worked when I port-forwarded Tycho and connected to `https://local...`. 

I initially wanted to add a flag `--unsecure-connection` to still handle it as it was before but this would have been breaking change.

zizou0x: I still need to test if it works from inside the cluster. I have some concerns about it

tycho-client/src/deltas.rs
```diff
@@ -609,11 +612,12 @@ impl DeltasClient for WsDeltasClient {
...
-                    .header("Host", "tycho-beta.propellerheads.xyz")
-                    .header("Connection", "Upgrade")
-                    .header("Upgrade", "websocket")
-                    .header("Sec-WebSocket-Version", 13);
+                    .header(SEC_WEBSOCKET_KEY, generate_key())
+                    .header(SEC_WEBSOCKET_VERSION, 13)
+                    .header(CONNECTION, "Upgrade")
+                    .header(UPGRADE, "websocket")
+                    .header(HOST, "tycho-beta.propellerheads.xyz")
+                    .header(USER_AGENT, format!("tycho-client-{}", env!("CARGO_PKG_VERSION")));
```
louise-poole, line 623: Good idea 👍

tycho-client/src/main.rs
zizou0x: Update: it didn't work so I added a flag https://github.com/propeller-heads/tycho-indexer/pull/379/commits/2f5678025418ee0066d079d531b790d0ea1075d0

tycho-client/README.md
```diff
@@ -63,11 +63,12 @@ tycho-client \
     --exchange uniswap_v3 \
     --exchange vm:ambient \
     --exchange vm:balancer \
-    --min-tvl 100
+    --min-tvl 100 \
     --tycho-url {TYCHO_INDEXER_URL}
 ```
  - TYCHO_INDEXER_URL defaults to `localhost:4242`
 
+ Note: if an authentication is required by the gateway, you can set it via `export TYCHO_AUTH_TOKEN={your_token}`.
```
louise-poole, line 71: And you should use the `--no-tls` flag otherwise (or for non-secured connections)

zizou0x: Not really, you could have a https gateway that doesn't require `Authorization`

PR #382, opened 2024-09-30
refactor(tycho-indexer): Use a dedicated tokio runtime for extraction

zizou0x:
This PR aims to improve Tycho stability by dedicating a runtime (currently single threaded) to extraction. That way we all others tasks (server, cache gateways..) won't affect the extractors' CPU ressources.

### DIFF
--- a/tycho-indexer/src/main.rs
+++ b/tycho-indexer/src/main.rs
@@ -2,7 +2,14 @@
 
 use futures03::future::select_all;
 use serde::Deserialize;
-use std::{collections::HashMap, fs::File, io::Read, str::FromStr, sync::Arc, thread};
+use std::{
+    collections::HashMap,
+    fs::File,
+    io::Read,
+    process,
+    str::FromStr,
+    sync::{mpsc, Arc},
+};
 use tracing_subscriber::EnvFilter;
 use tycho_ethereum::{
     account_extractor::contract::EVMAccountExtractor,
@@ -15,7 +22,7 @@ use actix_web::dev::ServerHandle;
 use chrono::{NaiveDateTime, Utc};
 use clap::Parser;
 use tokio::{runtime::Handle, select, task::JoinHandle};
-use tracing::{info, instrument, warn};
+use tracing::{error, info, instrument, warn};
 use tycho_core::{
     models::{
         blockchain::{Block, Transaction},
@@ -70,9 +77,7 @@ impl ExtractorConfigs {
 
 type ExtractionTasks = Vec<JoinHandle<Result<(), ExtractionError>>>;
 type ServerTasks = Vec<JoinHandle<Result<(), ExtractionError>>>; //TODO: introduce an error type for it
-
-#[tokio::main]
-async fn main() -> Result<(), anyhow::Error> {
+fn main() {
     let cli: Cli = Cli::parse();
     let global_args = cli.args();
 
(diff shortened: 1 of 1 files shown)


tycho-indexer/src/extractor/runner.rs
```diff
@@ -211,7 +225,12 @@ impl ExtractorRunner {
                     Ok(())
                 }.instrument(loop_span).await?
             }
-        })
+        };
+        if let Some(runtime) = runtime {
```
kayibal: Instead of this additional if-else, maybe we can default to the default runtime above in line 146. I think there should be a method to get it sth like `self.runtime_handle.or_else(|| tokio:.default_runtime())`

tycho-indexer/src/main.rs
```diff
@@ -289,12 +299,18 @@ async fn build_all_extractors(
...
         .await;
 
-        let (task, handle) = ExtractorBuilder::new(extractor_config, endpoint_url)
+        let extractor_builder = ExtractorBuilder::new(extractor_config, endpoint_url)
             .build(chain_state, cached_gw, token_pre_processor, &protocol_cache)
-            .await?
-            .run()
             .await?;
 
+        let extractor_builder = if let Some(runtime) = runtime.clone() {
```
kayibal: Same as above here maybe try to default earlier if runtime is None

tycho-indexer/src/main.rs
```diff
@@ -133,11 +141,21 @@ async fn run_indexer(
...
     .await?;
 
-    let (res, _, _) = select_all(tasks).await;
-    res.expect("Extractor- nor ServiceTasks should panic!")
+    let handle_extractor =
+        thread::spawn(move || extraction_runtime.block_on(select_all(extraction_tasks)));
+
+    let (res, _, _) = handle_extractor
+        .join()
+        .expect("Extractor thread shouldn't panic!");
```
kayibal: This statement blocks no? So the statements below are never executed unless the extractors finish? The correct way to work around this is to remove the `#[tokio::main]` macro and also instantiate a runtime for the other tasks. Unfortunately this means some additional overhead to join both threads since this is not so easy to do with Rust...

kayibal: Actually we should probably discuss this in more detail tomorrow - spawning the runtimes in threads is maybe not ideal (since joining threads requires a synchronization primitive anyway), instead we should probably do sth like this:

```rust
let tx, rx = std::sync::mpsc::channel(1);

rt1.spawn(|| async {
     // schedule the extraction tasks
    select_all!(extractor_tasks);
    // if any task finishes signal the exit channel
    tx.send(())
})


rt2.spawn(|| async {
  // schedule other tasks
   ....
  // on exit or failure
  tx.send(())
})

// block until one runtime sends a signal
rx.recv()

// shut down the runtimes
rt1.shutdown_background()
rt2.shutdown_background()
```

zizou0x: Hm you're right, this doesn't properly handle errors. 

I also thought about removing the main macro and code wise it makes more sense (so we can create the runtimes in main and distribute them as we want). The main "issue" that I got with this is that we need to pass down handles so it's not very beautiful in term of interface. Maybe we can use a `lazy_static` to create the runtimes but then it's weird IMO.

Let's discuss the channels thing, I think it can work

PR #392, opened 2024-10-03
fix: avoid concurrent requests for empty pages

louise-poole:


tycho-indexer/src/services/rpc.rs
```diff
@@ -438,7 +438,10 @@ where
             .get(request.clone(), |r| async {
                 self.get_protocol_components_inner(r)
                     .await
-                    .map(|res| (res, true))
+                    .map(|res| {
+                        let last_page = res.pagination.total_pages() - 1;
+                        (res, request.pagination.page >= last_page)
```
kayibal: Are you sure about this condition we cache only if it is the last page?

PR #396, opened 2024-10-03
feat(rpc): Implement per-key sharded locking in RpcCache

kayibal:
This commit introduces a per-key sharded locking mechanism in RpcCache by storing results wrapped in `tokio::sync::Mutex` within the cache. Each cache entry now has its own mutex, ensuring that only one query is executed for a given key while allowing concurrent access to other keys.

This approach reduces lock contention, as the entire cache is no longer locked for writes, and simplifies lock management by leveraging the cache's eviction policy. When a key is evicted, the associated mutex is automatically cleaned up.

By implementing this change, we improve performance under load and ensure that multiple identical requests for the same key wait on the result of a single query, reducing redundant database queries.

### DIFF
--- a/tycho-indexer/src/services/cache.rs
+++ b/tycho-indexer/src/services/cache.rs
@@ -1,12 +1,11 @@
 use futures03::Future;
 use mini_moka::sync::Cache;
 use std::{error::Error, fmt::Debug, hash::Hash, sync::Arc};
-use tokio::sync::RwLock;
 use tracing::{debug, instrument, Level};
 
 pub struct RpcCache<R, V> {
     name: String,
-    cache: Arc<RwLock<Cache<R, V>>>,
+    cache: Arc<Cache<R, Arc<tokio::sync::Mutex<Option<V>>>>>,
 }
 
 impl<R, V> RpcCache<R, V>
@@ -15,12 +14,12 @@ where
     V: Clone + Send + Sync + 'static,
 {
     pub fn new(name: &str, capacity: u64, ttl: u64) -> Self {
-        let cache = Arc::new(RwLock::new(
+        let cache = Arc::new(
             Cache::builder()
                 .max_capacity(capacity)
                 .time_to_live(std::time::Duration::from_secs(ttl))
                 .build(),
-        ));
+        );
         Self { name: name.to_string(), cache }
     }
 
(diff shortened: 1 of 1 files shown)


tycho-indexer/src/services/cache.rs
```diff
@@ -35,34 +34,53 @@ where
...
-        let (response, should_cache) = (fallback)(request.clone()).await?;
-
-        if should_cache {
-            debug!(?request, name = self.name, "IncompleteCacheValue");
-            write_lock.insert(request, response.clone());
-        }
+        debug!(?request, cache_size = self.cache.entry_count(), cache = self.name, "CacheMiss");
+        // Acquire a write lock before querying the database (prevents concurrent db queries)
+        let response = if let Some(inflight_val) = self.cache.get(&request) {
+            // the value is being written to
```
zizou0x: Should we add a debug event here?

tycho-indexer/src/services/cache.rs
```diff
@@ -35,34 +34,53 @@ where
...
+                    self.cache.invalidate(&request);
+                    e
+                })?;
+
+            // PERF: unnecessary lock if we don't cache the value, could be improved
+            //  if `should_cache` value can be determined beforehand.
+            if should_cache {
+                *guard = Some(response.clone())
+            } else {
+                self.cache.invalidate(&request);
```
zizou0x, line 66: What if we have a request waiting [here](https://github.com/propeller-heads/tycho-indexer/pull/396/files#diff-14b8dcf04eec2280d6f5cc5f3a0306b5baaa211169c977d46d693253a81c7b98R55)? 

We would invalidate the cache but then the awaiting request is gonna panic at [line 57](https://github.com/propeller-heads/tycho-indexer/pull/396/files#diff-14b8dcf04eec2280d6f5cc5f3a0306b5baaa211169c977d46d693253a81c7b98R57) because the value in the mutex would be `None`.

tycho-indexer/src/services/cache.rs
```diff
@@ -35,35 +34,41 @@ where
...
         // Check the cache for a cached response
-        {
-            let read_lock = self.cache.read().await;
-            cache_entry_count = read_lock.entry_count();
-            if let Some(cached_response) = read_lock.get(&request) {
-                debug!(cache = self.name, ?request, "CacheHit");
-                return Ok(cached_response);
+        if let Some(inflight_val) = self.cache.get(&request) {
+            // the value is present or is being written to
+            debug!(cache = self.name, ?request, cache_size = self.cache.entry_count(), "CacheHit");
```
zizou0x: It would be more accurate inside the other condition below

PR #408, opened 2024-10-07
refactor(rpc): be more efficient when updating state with buffer

zizou0x:
This PR allows us to be more efficient for during buffer lookups if the protocol system is given in the request.

### DIFF
--- a/tycho-indexer/src/services/deltas_buffer.rs
+++ b/tycho-indexer/src/services/deltas_buffer.rs
@@ -58,7 +58,7 @@ pub trait PendingDeltasBuffer {
         protocol_ids: Option<&[&str]>,
         db_states: &mut Vec<ProtocolComponentState>,
         version: Option<BlockNumberOrTimestamp>,
-        protocol_system: Option<&str>,
+        protocol_system: &str,
     ) -> Result<()>;
 
     fn update_vm_states(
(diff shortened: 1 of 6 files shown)


tycho-indexer/src/services/deltas_buffer.rs
```diff
@@ -138,36 +140,64 @@ impl PendingDeltas {
...
+                PendingDeltasError::UnknownExtractor(ps.to_string())
+            })?;
+            change_found = update_state(ps, buffer, db_state)?;
+        } else {
+            // If no protocol system is provided, iterate through all buffers
+            for (name, buffer) in self.buffers.iter() {
+                change_found = update_state(name, buffer, db_state)?;
+                if change_found {
+                    break; // Stop once a change is found
+                }
```
kayibal: I think we made `protocol_system` required so it should always be available. Doing it this way is very unsafe especially if you have an extractor that is syncing. I think we should error in this case.

tycho-indexer/src/services/deltas_buffer.rs
```diff
@@ -176,27 +206,44 @@ impl PendingDeltas {
...
```
kayibal: Same here, I think we should simply error if we don't get a protocol_system.

tycho-indexer/src/services/deltas_buffer.rs
louise-poole: I assume this is done for backwards compatibility? But I agree - I'd rather error as it's clear that not providing a protocol system has unexpected behaviour...

zizou0x: addressed in https://github.com/propeller-heads/tycho-indexer/pull/408/commits/a2db521bb2a0a30f659764899c794d47c6dde6ef

tycho-indexer/src/services/deltas_buffer.rs
zizou0x: addressed in https://github.com/propeller-heads/tycho-indexer/pull/408/commits/a2db521bb2a0a30f659764899c794d47c6dde6ef

PR #411, opened 2024-10-09
fix(rpc): fix running RPC without extractors

zizou0x:
This is useful when running the `rpc` CLI command. In that case we don't run any extractor so it's expected that no `PendingDeltasBuffer` are registered in the RPC. But due to the recent changes in how we handle buffer lookups, it was failing if the buffer wasn't found.

This commit also refactor the service builder to avoid spawning unnecessary tasks and add docs.

### DIFF
--- a/tycho-indexer/src/services/rpc.rs
+++ b/tycho-indexer/src/services/rpc.rs
@@ -155,14 +155,16 @@ where
 
         // Apply pagination to the contract addresses. This is done so that we can determine which
         // contracts were not returned from the db and get them from the buffer instead.
-        let mut paginated_addrs: Vec<Bytes> = Vec::new();
-        if let Some(adrs) = addresses {
-            paginated_addrs = adrs
-                .iter()
-                .skip(pagination_params.offset() as usize)
-                .take(pagination_params.page_size as usize)
-                .cloned()
-                .collect();
+        let mut paginated_addrs: Option<Vec<Bytes>> = None;
+        if let Some(addrs) = addresses {
+            paginated_addrs = Some(
+                addrs
+                    .iter()
+                    .skip(pagination_params.offset() as usize)
+                    .take(pagination_params.page_size as usize)
+                    .cloned()
+                    .collect(),
+            );
         }
 
         // Get the contract states from the database
@@ -170,7 +172,7 @@ where
             .db_gateway
             .get_contracts(
                 &chain,
-                Some(&paginated_addrs),
+                paginated_addrs.as_deref(),
                 Some(&db_version),
                 true,
                 Some(&pagination_params),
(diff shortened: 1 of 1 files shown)


tycho-indexer/src/services/mod.rs
```diff
@@ -117,27 +124,61 @@ where
         )]
         struct ApiDoc;
 
-        let openapi = ApiDoc::openapi();
+        let open_api = ApiDoc::openapi();
+
+        // If no extractors are registered, run the server without spawning extractor-related tasks.
+        if self.extractor_handles.is_empty() {
+            self.start_server(None, open_api)
```
kayibal, line 132: start_standalone_server

tycho-indexer/src/services/mod.rs
```diff
@@ -117,27 +124,61 @@ where
...
+            Ok(())
+        });
+
+        Ok((server_handle, task))
+    }
+
+    /// Helper to spawn the main server task, optionally enabling WebSocket services.
+    fn start_server(
+        self,
+        ws_data: Option<web::Data<ws::WsData>>,
```
kayibal, line 177: Why do we have this here if we have a method that does it all for us already? Maybe we should have a method that creates WsData and one that starts the server with or without WsData?

zizou0x: > one that starts the server with or without WsData

I didn't get this, isn't it exactly what this method does? The alternative that I considered what to introduce a builder pattern for the server. But that would mean nested builders (ServerBuilder inside ServicesBuilder)  and seemed a bit overkill for that because we only have 2 paths in the run function.

> if we have a method that does it all for us already

Do you mean `start_server_with_deltas`? If yes, it uses this generic `start_server` method internally.

So basically this one is a helper to spawn the server. But it's confusing it seems, maybe we should introduce the `start_standalone_server` function that would just call this one like this then `self.start_server(None, open_api)`. That way it would be the same pattern in both paths - do things and call use `self.start_server` to create the server.

kayibal: Yes I would prefer what you described in the last paragraph.

tycho-indexer/src/services/mod.rs
```diff
@@ -116,27 +123,61 @@ where
         )]
         struct ApiDoc;
 
-        let openapi = ApiDoc::openapi();
+        let open_api = ApiDoc::openapi();
+
+        // If no extractors are registered, run the server without spawning extractor-related tasks.
+        if self.extractor_handles.is_empty() {
+            self.start_server(None, open_api)
```
kayibal, line 132: Should we put an info log here?

tycho-indexer/src/services/mod.rs
```diff
@@ -168,6 +168,17 @@ where
...
+    /// websocket).
+    fn start_standalone_server(
+        self,
+        openapi: utoipa::openapi::OpenApi,
+    ) -> Result<(ServerHandle, JoinHandle<Result<(), ExtractionError>>), ExtractionError> {
+        // This function is shallow but helps to keep the building pattern consistent between paths
+        // in `self.run()`
+        self.start_server(None, openapi)
+    }
+
```
kayibal: I think I'd remove it...

PR #417, opened 2024-10-11
refactor(indexer): clean up outdated naming, modules and tests

louise-poole:
Included in this refactor:
- removal of all references to the 'hybrid' extractor. We used to have multiple extractors that we merged into one hybrid one. Now that the old ones have been removed and we have only a single extractor, the use of the name 'hybrid' doesn't make sense.
- Split overly-large files into smaller, more appropriately grouped files (the old evm/mod.rs file was excessively long)
- move tests that were left behind during the first major repo refactor splitting the core models from the indexing logic
- remove the use of 'evm' vs 'compat' modules. A previous refactor removed the reliance on the ethers crate inside the indexer making the distinction unnecessary.
- remove inconsistencies: some structs implement the TryFromMessage trait while others implemented `try_from_message` directly. Now all structs that require it use the trait.

### DIFF
--- a/tycho-indexer/src/services/cache.rs
+++ b/tycho-indexer/src/services/cache.rs
@@ -64,11 +64,10 @@ where
             .insert(request.clone(), lock.clone());
         let (response, should_cache) = (fallback)(request.clone())
             .await
-            .map_err(|e| {
+            .inspect_err(|_| {
                 // invalidate the cache if the fallback errors
                 self.cache.invalidate(&request);
                 trace!("FallbackFailure");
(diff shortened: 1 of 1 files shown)


tycho-indexer/src/extractor/protobuf_deserialisation.rs
```diff
@@ -0,0 +1,701 @@
+use std::collections::{hash_map::Entry, HashMap, HashSet};
```
tvinagre, line 1: 💯

PR #419, opened 2024-10-21
feat(indexer): remove le to be transcode of usv2 balances

louise-poole:
The new usv2 substream package (0.3.0) encodes balances with big endian encoding, we no longer need to transcode it.

### DIFF
--- a/tycho-indexer/src/extractor/post_processors/attributes.rs
+++ b/tycho-indexer/src/extractor/post_processors/attributes.rs
@@ -5,7 +5,6 @@ use tycho_core::{models::protocol::ProtocolComponentStateDelta, Bytes};
 use crate::extractor::models::BlockChanges;
 
 const USV3_MANDATORY_ATTRIBUTES: [&str; 3] = ["liquidity", "tick", "sqrt_price_x96"];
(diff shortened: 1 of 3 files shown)


tycho-indexer/src/extractor/post_processors/balances.rs
```diff
@@ -86,7 +80,7 @@ pub fn transcode_usv2_balances(mut changes: BlockChanges) -> BlockChanges {
                     balance
                         .iter_mut()
                         .for_each(|(_, value)| {
-                            value.balance = transcode_le_balance_to_be(&value.balance).unwrap();
```
zizou0x, line 89: I think we can remove the whole post processors then? It looks like it's just doing this

louise-poole: It's also calculating the float value no?

zizou0x: You're gonna have to double check but I think it's just updating the balance float after updating the balance encoding. But if it's correctly encoded from the beginning we shouldn't need this either

louise-poole: Ahhhh - okay okay, let me test rn

PR #420, opened 2024-10-21
fix(indexer): correctly handle attributes deletions

zizou0x:
Attributes used to be marked as deleted for the whole update batch. In some scenarios where the attribute was updated after being deleted this would lead to wrong data being sent to the db.

### DIFF
--- a/tycho-storage/src/postgres/versioning.rs
+++ b/tycho-storage/src/postgres/versioning.rs
@@ -259,10 +259,15 @@ pub trait PartitionedVersionedRow: Clone + Send + Sync {
         Self: Sized;
 }
 
+type LatestRows<N> = Vec<N>;
+type ArchivedRows<N> = Vec<N>;
+type DeletedIds<N> = Vec<<N as PartitionedVersionedRow>::EntityId>;
+type RowsChanges<N> = (LatestRows<N>, ArchivedRows<N>, DeletedIds<N>);
+
 fn set_partitioned_versioning_attributes<N: PartitionedVersionedRow>(
     current_latest_data: &[N],
     new_data: &[VersioningEntry<N>],
-) -> Result<(Vec<N>, Vec<N>, Vec<N::EntityId>), StorageError> {
+) -> Result<RowsChanges<N>, StorageError> {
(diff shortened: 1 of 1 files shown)


tycho-storage/src/postgres/versioning.rs
```diff
@@ -241,31 +245,41 @@ pub trait PartitionedVersionedRow: Clone + Send + Sync {
...
-            continue;
+        match item {
+            VersioningEntry::Update(r) => {
+                // Handle updated rows
+                let mut row = r.clone();
+                if let Some(mut prev) = latest.remove(&id) {
+                    prev.archive(&mut row);
+                    archived.push(prev);
+                }
+                // If it's updated after being deleted, don't need to mark it as deleted
```
zizou0x:
```suggestion
                // If it's updated after being deleted, it doesn't need to be marked as deleted
```

tycho-storage/src/postgres/versioning.rs
```diff
@@ -241,31 +260,41 @@ pub trait PartitionedVersionedRow: Clone + Send + Sync {
...
+                }
+                // If it's updated after being deleted, it doesn't need to be marked as deleted
+                deleted.remove(&id);
+                latest.insert(id, row);
+            }
+            VersioningEntry::Deletion((id, delete_version)) => {
+                // Handle deleted rows
+                let mut delete_row = latest
+                    .remove(id)
+                    .expect("missing state to delete"); //TODO: This is strict to avoid corrupting the data, would it be fine to just skip?
```
zizou0x: I decided to be strict here and crash if it happens. This would mean that we try to delete a row that isn't initialized

kayibal: It's ok to be strict, but maybe try to crash the application gracefully (return an error)?

zizou0x: Yes true, addressed in https://github.com/propeller-heads/tycho-indexer/pull/420/commits/46d59d105a5d735bc6d1093080f4eeac1b2b7dc3

PR #423, opened 2024-10-22
feat(indexer): introduce configurable post processors

louise-poole:
Make the post processors configurable. This gives us more control on where and when post processors are used. We now do not need to worry about prod deployments if prod is still using an outdated substream that requires a specific post processor still.

Note: in order to not waste too much time on this I kept the solution as simple as possible. This means only 1 post processor can be set per extractor. If we need to stack them (like with the old usv2 substreams), they must be combined within the code itself and registered as a single post processor. I started looking into making this fully configurable and it gets complicated with types etc... since we're deprecating the use of the usv2 stacked post processors I decided to not waste too much more time on this.

### DIFF
--- a/tycho-indexer/src/extractor/runner.rs
+++ b/tycho-indexer/src/extractor/runner.rs
@@ -448,7 +448,7 @@ impl ExtractorBuilder {
             .config
             .post_processor
             .as_ref()
-            .and_then(|name| {
+            .map(|name| {
                 POST_PROCESSOR_REGISTRY
                     .get(name)
                     .cloned()
@@ -458,8 +458,8 @@ impl ExtractorBuilder {
                             name
                         ))
                     })
-                    .ok()
-            });
+            })
+            .transpose()?;
 
         self.extractor = Some(Arc::new(
             ProtocolExtractor::new(


tycho-indexer/src/extractor/runner.rs
```diff
@@ -444,6 +444,23 @@ impl ExtractorBuilder {
...
+                POST_PROCESSOR_REGISTRY
+                    .get(name)
+                    .cloned()
+                    .ok_or_else(|| {
+                        ExtractionError::Setup(format!(
+                            "Post processor '{}' not found in registry",
+                            name
+                        ))
+                    })
+                    .ok()
```
zizou0x: We should propagate the error or even panic here, it would most likely mean a typo in the name that we want to know about before running without it

tycho-indexer/src/extractor/post_processors/mod.rs
```diff
@@ -1,13 +1,48 @@
...
+        transcode_ambient_balances as PostProcessorFn,
+    );
+    registry.insert(
+        "add_default_attributes_uniswapv3".to_string(),
+        add_default_attributes_uniswapv3 as PostProcessorFn,
+    );
+    registry.insert("ignore_self_balances".to_string(), ignore_self_balances as PostProcessorFn);
+    registry.insert(
+        "trim_curve_component_token".to_string(),
+        trim_curve_component_token as PostProcessorFn,
```
zizou0x, line 41: fmi why is `as PostProcessorFn` needed here

louise-poole: Because without it I get:
```
mismatched types
expected fn item `fn(extractor::models::BlockChanges) -> extractor::models::BlockChanges {extractor::post_processors::balances::transcode_ambient_balances}`
   found fn item `fn(extractor::models::BlockChanges) -> extractor::models::BlockChanges {extractor::post_processors::attributes::add_default_attributes_uniswapv3}`
different fn items have unique types, even if their signatures are the same
```

zizou0x: Makes sense

PR #429, opened 2024-10-24
feat(hex_bytes): change hex bytes conversions to big endian

louise-poole:
All numerical encoding is now done with big endian, therefore these Bytes utils needed updating.

NOTE: these conversion utils are not used within the tycho indexing logic itself. They are used in tests only and are publicly exposed to help simplify the handling of tycho messages.

### DIFF
--- a/tycho-ethereum/Cargo.toml
+++ b/tycho-ethereum/Cargo.toml
@@ -24,11 +24,11 @@ mockall = "0.11"
 unicode-segmentation = "1.12.0"
 
 # Optional dependencies
-contracts = { git = "https://github.com/cowprotocol/services", rev = "f3678428991e055ceb517f184d1b37244f2d8f51", optional = true}
-ethrpc = { git = "https://github.com/cowprotocol/services", rev = "f3678428991e055ceb517f184d1b37244f2d8f51", optional = true}
+contracts = { git = "https://github.com/cowprotocol/services", rev = "f3678428991e055ceb517f184d1b37244f2d8f51", optional = true }
+ethrpc = { git = "https://github.com/cowprotocol/services", rev = "f3678428991e055ceb517f184d1b37244f2d8f51", optional = true }
 ethcontract = { version = "0.25.4", default-features = false, features = [
     "aws-kms",
-] }
+], optional = true }
 clap = { version = "4", features = ["derive", "env"], optional = true }
 humantime = { version = "2.1.0", optional = true }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6749,6 +6749,7 @@ dependencies = [
  "ethers",
  "ethrpc",
  "humantime",
+ "mockall",
  "primitive-types",
  "reqwest 0.11.24",
  "serde",


tycho-ethereum/Cargo.toml
```diff
@@ -15,18 +13,24 @@ tycho-core.workspace = true
...
-reqwest.workspace = true
-url.workspace = true
+], optional = true}
+clap = { version = "4", features = ["derive", "env"], optional = true}
+humantime = { version = "2.1.0", optional = true }
+
+
+[features]
+default = []
+contracts = ["ethrpc", "ethcontract", "dep:contracts", "humantime", "clap"]
```
louise-poole: I don't like this feature name... suggestions welcome

tycho-ethereum/src/lib.rs
```diff
@@ -115,9 +119,9 @@ impl BytesConvertible for U256 {
         let mut u256_bytes: [u8; 32] = [0; 32];
 
         // Copy bytes from `bytes_slice` to `u256_bytes`.
-        u256_bytes[..bytes_slice.len()].copy_from_slice(bytes_slice);
+        u256_bytes[32 - bytes_slice.len()..].copy_from_slice(bytes_slice);
 
-        // Convert the byte array to `U256` using little-endian.
-        U256::from_little_endian(&u256_bytes)
+        // Convert the byte array to `U256` using big-endian.
+        U256::from_big_endian(&u256_bytes)
```
zizou0x, line 125: Please add it in the docs as this is really specific. I would even rename the function to `from_big_endian`

tycho-ethereum/src/lib.rs
```diff
@@ -44,7 +48,7 @@ impl From<BlockTagWrapper> for BlockNumber {
 /// let bytes: Bytes = h160_value.to_bytes(); // Converts H160 to Bytes
 /// let new_h160 = H160::from_bytes(bytes);   // Converts Bytes back to H160
 /// ```
-pub trait BytesConvertible {
+pub trait BytesCodec {
```
zizou0x, line 51: I think I prefer `BytesConvertible` but it's subjective

tycho-ethereum/Cargo.toml
zizou0x: Could be something like "on_chain_data" or "rpc_calls" or something in that direction?

tycho-ethereum/src/lib.rs
louise-poole: I changed it to make it more explicit that it's dealing with encoding and decoding... 'convertible' felt more broad/general. But I don't feel too strongly about this so can change it back?

tycho-ethereum/src/lib.rs
louise-poole: Since this is tycho conversions and tycho will from now on only have big endian, I don't think the function name needs to be so explicit. But I'll defs add it to the docs 👍

zizou0x: IMO in general any function that converts Bytes into Int needs to precise endianness. But that's just my own feeling, if it's well documented and people don't miss it it's probably ok

tycho-ethereum/src/lib.rs
zizou0x: Same, it's really subjective. I summon @tvinagre to give a decisive opinion here

tycho-ethereum/src/lib.rs
tvinagre: I tend to agree with Flo. Worst case, `from_be`

tycho-ethereum/src/lib.rs
tvinagre: I vote for `Codec`, as it stands for `encoder & decoder`

PR #435, opened 2024-10-29
fix(rpc): add buffer lookup for version given as block hash

zizou0x:
This PR fixes request that specify the version using a block hash. 

In the current code this would fail if the requested block was recent and still stored in the buffer.

Proposed solution: if the hash isn't found in the db, also search for it in the buffer. Then if it's not found in the buffer either, return the block not found error.

### DIFF
--- a/tycho-indexer/src/services/rpc.rs
+++ b/tycho-indexer/src/services/rpc.rs
@@ -236,26 +236,26 @@ where
                 BlockNumberOrTimestamp::Number(*no as u64)
             }
             BlockOrTimestamp::Block(BlockIdentifier::Hash(hash)) => {
-                let block_number = match self
-                    .db_gateway
-                    .get_block(&BlockIdentifier::Hash(hash.clone()))
-                    .await
+                let block_number = if let Some(block_number) = self
+                    .pending_deltas
+                    .as_ref()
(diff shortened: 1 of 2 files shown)


tycho-indexer/src/services/rpc.rs
```diff
@@ -235,6 +235,37 @@ where
...
             }
+            BlockOrTimestamp::Block(BlockIdentifier::Hash(hash)) => {
+                let block_number = match self
+                    .db_gateway
+                    .get_block(&BlockIdentifier::Hash(hash.clone()))
+                    .await
+                {
+                    Ok(b) => Some(b.number), // Successfully found the block in the DB
+                    Err(_) => {
+                        // Try to find the block in pending deltas if the DB lookup failed
```
kayibal: I don't like the order of things here, DB access is expensive, while deltas buffer access is cheap. Why do we look for it first on the DB?

zizou0x: I made the assumption that we have more blocks in the db so it's gonna be there most of the time

But given how cheap it is to look into the buffer I think you're right and we should search first in the buffer. I'll change it

tvinagre (approved):

PR #437, opened 2024-10-29
feat(rpc): remove chain param from individual protocol ids

louise-poole:


tycho-client/src/feed/component_tracker.rs
```diff
@@ -1,11 +1,13 @@
...
 use std::collections::{HashMap, HashSet};
 use tracing::{debug, instrument, warn};
+
 use tycho_core::{
-    dto::{BlockChanges, Chain, ProtocolComponent, ProtocolComponentsRequestBody, ProtocolId},
+    dto::{BlockChanges, Chain, ProtocolComponent, ProtocolComponentsRequestBody},
     Bytes,
 };
 
+use crate::{rpc::RPCClient, RPCError};
```
zizou0x, line 9: Why so many imports changes, is it RustRover?

tycho-indexer/src/services/mod.rs
```diff
@@ -1,5 +1,5 @@
 //! This module contains Tycho web services implementation
-
+#![allow(deprecated)]
```
zizou0x, line 3: Please add a TODO to remove else it's gonna stay there forever

tycho-core/src/dto.rs
```diff
@@ -1036,8 +1056,122 @@ pub struct ProtocolStateRequestBody {
 }
 
 impl ProtocolStateRequestBody {
-    pub fn id_filtered(ids: Vec<ProtocolId>) -> Self {
-        Self { protocol_ids: Some(ids), ..Default::default() }
+    pub fn id_filtered<I, T>(ids: I) -> Self
+    where
+        I: IntoIterator<Item = T>,
+        T: Into<String>,
```
zizou0x, line 1062: Would it work with just one generic here? 

I: IntoIterator<Item = Into<String>>

tycho-client/src/feed/component_tracker.rs
louise-poole: No - it was me. When I refactored the indexer earlier a cleaned up a lot of imports and made them consistent. Couldn't resist sneaking in doing that here too.

tycho-core/src/dto.rs
louise-poole: I previously tried to sombine it into one generic and got lots of errors. This was the cleanest solution I could come up with.

tycho-client/src/feed/component_tracker.rs
```diff
@@ -178,10 +180,10 @@ where
             .collect()
     }
 
-    pub fn get_tracked_component_ids(&self) -> Vec<ProtocolId> {
+    pub fn get_tracked_component_ids(&self) -> Vec<String> {
         self.components
             .keys()
-            .map(|k| ProtocolId { chain: self.chain, id: k.clone() })
+            .cloned()
```
tvinagre, line 186: Is `ProtocolId` used with chain + id in more places? If not, we should refactor it and keep it as a String Type

louise-poole: We can do that when we actually remove it. For now it needs to remain as is for backwards compatibility.

PR #438, opened 2024-10-30
DO NOT MERGE fix(versioning): correctly retrieve old state

zizou0x:
`latest_versions_by_ids` where wrongly filtering all state with `valid_to.eq(MAX_TS)` which is not working because of timestamp precision.

We now use the alternative `valid_to.gt(*MAX_VERSION_TS)` everywhere.

Note: for efficiency reason, when diesel allows it, we query the default table directly instead of filtering by ts.

**Edit: this was fixed on the db side, but ideally we want to query the default table instead of filtering by timestamps**

### DIFF
--- a/tycho-storage/src/postgres/protocol.rs
+++ b/tycho-storage/src/postgres/protocol.rs
@@ -373,16 +373,17 @@ impl PostgresGateway {
         let mut res: HashMap<Address, (String, Bytes)> = HashMap::new();
         schema::protocol_component::table
             .inner_join(schema::protocol_component_holds_token::table)
-            .inner_join(schema::component_balance::table)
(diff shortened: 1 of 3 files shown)


tycho-storage/src/postgres/protocol.rs
```diff
@@ -373,16 +373,17 @@ impl PostgresGateway {
         let mut res: HashMap<Address, (String, Bytes)> = HashMap::new();
         schema::protocol_component::table
             .inner_join(schema::protocol_component_holds_token::table)
-            .inner_join(schema::component_balance::table)
+            .inner_join(schema::component_balance_default::table)
```
zizou0x, line 376: Should be equivalent to filtering by ts and faster

PR #439, opened 2024-10-30
feat(tycho-client): implement a rust client stream builder

louise-poole:


tycho-client/src/runner.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    let (sync_jh, mut rx) = block_sync
+        .run()
+        .await
+        .expect("block sync start error");
+
+    let msg_printer = tokio::spawn(async move {
+        while let Some(msg) = rx.recv().await {
+            if let Ok(msg_json) = serde_json::to_string(&msg) {
+                println!("{}", msg_json);
```
tvinagre: Why is it printing a msg?

tycho-client/src/runner.rs
```diff
@@ -0,0 +1,122 @@
...
+///
+/// # Returns
+///
+/// Returns a `Receiver<FeedMessage>` channel that can be used to process `FeedMessage` events generated by
+/// the Tycho client in real-time, allowing external handling and processing of these messages.
+#[allow(clippy::too_many_arguments)]
+pub async fn start_tycho_client(
+    tycho_url: &str,
+    chain: Chain,
+    exchanges: Vec<ExchangeFilter>,
```
kayibal: Maybe this should be a hashmap to avoid specifying the same exchange twice

tycho-client/src/runner.rs
```diff
@@ -0,0 +1,122 @@
...
+/// * `no_tls`: If `true`, disables TLS/SSL for the connections, using `http` and `ws` protocols instead of `https` and `wss`.
+///
+/// # Returns
+///
+/// Returns a `Receiver<FeedMessage>` channel that can be used to process `FeedMessage` events generated by
+/// the Tycho client in real-time, allowing external handling and processing of these messages.
+#[allow(clippy::too_many_arguments)]
+pub async fn start_tycho_client(
+    tycho_url: &str,
+    chain: Chain,
```
kayibal: Would be cool if setting chain also sets defaults for the timing parameters

tycho-client/README.md
```diff
@@ -44,25 +46,57 @@ tycho-client \
...
+
+## Rust Builder
+
+You can also integrate Tycho Client directly in your Rust projects using the rust client builder:
+
+```rust
+use tycho_core::dto::Chain;
+use tycho_client::{client::TychoClientBuilder, feed::component_tracker::ComponentFilter};
+
+let receiver = TychoClientBuilder::new("localhost:4242", Chain::Ethereum)
```
kayibal: Could we call this `TychoStreamBuilder` please?  It's a bit clearer since basically what you get is a stream of messages.

tycho-client/src/client.rs
```diff
@@ -0,0 +1,196 @@
...
+        } else {
+            self.auth_key
+        };
+
+        tracing_subscriber::fmt()
+            .with_env_filter(
+                tracing_subscriber::EnvFilter::try_from_default_env()
+                    .unwrap_or_else(|_| tracing_subscriber::EnvFilter::new("info")),
+            )
+            .init();
```
kayibal: This shouldn't be here... the user usually will want to set up this themselves.

tycho-client/src/client.rs
```diff
@@ -0,0 +1,196 @@
...
+            let tycho_rpc_url = format!("https://{}", self.tycho_url);
+            (tycho_ws_url, tycho_rpc_url)
+        };
+
+        // Initialize the WebSocket client
+        let ws_client = WsDeltasClient::new(&tycho_ws_url, auth_key.as_deref()).unwrap();
+        let ws_jh = ws_client
+            .connect()
+            .await
+            .expect("ws client connection error");
```
kayibal: Since this is library code now, we really should avoid panics.

tycho-client/src/client.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    /// Disables TLS/SSL for the connection, using `http` and `ws` protocols.
+    pub fn no_tls(mut self, no_tls: bool) -> Self {
+        self.no_tls = no_tls;
+        self
+    }
+
+    /// Builds and starts the Tycho client, connecting to the Tycho server and
+    /// setting up the synchronization of exchange components.
+    pub async fn build(self) -> Result<Receiver<FeedMessage>, &'static str> {
```
kayibal: We also need an error type here sorry...

tycho-client/README.md
```diff
@@ -44,25 +46,57 @@ tycho-client \
...
 
 #### Authentication
 
-If the gateway requires authentication you can provide it one of two ways: 
-- Set it via an enviroment variable:
- ```bash
- export TYCHO_AUTH_TOKEN={your_token}
- ```
-- Use the `--auth-key {your_token}` flag directly in the command.
+If your Tycho Indexer requires authentication, you can provide your token in one of two ways:
```
tvinagre, line 54:
```suggestion
Our hosted Tycho Indexer requires client authentication. You can provide your token in one of two ways:
```

tycho-client/README.md
```diff
@@ -44,25 +46,57 @@ tycho-client \
...
+
+Environment Variable:
```
tvinagre, line 56:
```suggestion
**Environment Variable (preferred)**:
```

tycho-client/README.md
tvinagre: Because we don't offer authentication options for the self-hosted version. So that should be handled by the user.

tycho-client/README.md
kayibal: Why is it preferred?

tycho-client/src/stream.rs
```diff
@@ -13,6 +14,18 @@ use crate::{
     HttpRPCClient, WsDeltasClient,
 };
 
+#[derive(Error, Debug)]
+pub enum StreamError {
```
kayibal, line 18: This name is too generic

tycho-client/src/stream.rs
```diff
@@ -94,17 +107,19 @@ impl TychoStreamBuilder {
...
+            ));
         }
 
         // If no auth_key is set and no_tls is false, try to read from the TYCHO_AUTH_TOKEN
         // environment variable
         let auth_key = if self.auth_key.is_none() && !self.no_tls {
             match env::var("TYCHO_AUTH_TOKEN") {
                 Ok(token) => Some(token),
-                Err(_) => return Err("Authentication key is required when `no_tls` is false. Set the TYCHO_AUTH_TOKEN environment variable or use the `auth_key` method."),
+                Err(_) => return Err(StreamError::SetUpError("Authentication key is required when `no_tls` is false. Set the TYCHO_AUTH_TOKEN environment variable or use the `auth_key` method.".to_string())),
```
kayibal: Idk this seems overly restrictive, what is someone hosts an internal endpoint but needs to use tls, or what if someone just has a public endpoint without authentication?

tycho-client/src/stream.rs
louise-poole: Any suggestions on what to change it to? 👀

tycho-client/README.md
zizou0x: Bash history I guess. I also think passing secrets to a cli is a weird thing to do

PR #440, opened 2024-10-31
feat(indexer): parse s3 bucket from env variable

zizou0x:


tycho-indexer/src/extractor/runner.rs
```diff
@@ -397,18 +397,21 @@ impl ExtractorBuilder {
...
         // Pull spkg from s3 and copy it at `spkg_path`
         if !Path::new(&self.config.spkg).exists() {
-            download_file_from_s3(
-                "repo.propellerheads",
-                &self.config.spkg,
-                Path::new(&self.config.spkg),
-            )
-            .await
-            .map_err(|e| {
+            let bucket = std::env::var("TYCHO_S3_BUCKET").map_err(|e| {
```
kayibal: There error message here is not good

kayibal: Would be nice to have a default so this doesn't have to be an error case? Eventually I guess we should expose this to the CLI too

zizou0x: I think it is correct? It would say something like 

> Failed to download uniswap_v2 from s3. missing env variable

Also I'm not sure what the default for this should be for an open source software? This is very tied to who runs it

Ideally yes we should expose it to the CLI as an `Option<String>` and pass it all the way down

zizou0x: I exposed it to clap, still no default as I don't know what a correct default value would be

PR #444, opened 2024-11-07
feat: make Snapshot attributes public

tamaralipows:
Otherwise, they cannot be accessed from the token quoter.

### DIFF
--- a/tycho-client/src/feed/synchronizer.rs
+++ b/tycho-client/src/feed/synchronizer.rs
@@ -62,8 +62,8 @@ pub struct ComponentWithState {
 
 #[derive(Clone, PartialEq, Debug, Default, Serialize)]
 pub struct Snapshot {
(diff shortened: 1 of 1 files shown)


tycho-client/src/feed/synchronizer.rs
```diff
@@ -62,8 +62,8 @@ pub struct ComponentWithState {
 
 #[derive(Clone, PartialEq, Debug, Default, Serialize)]
 pub struct Snapshot {
-    states: HashMap<String, ComponentWithState>,
-    vm_storage: HashMap<Bytes, ResponseAccount>,
+    pub states: HashMap<String, ComponentWithState>,
```
dianacarvalho1, line 65: this is not needed @tamaralipows! there are getter methods below!

tamaralipows: how did I miss this ehehehe sorry

PR #452, opened 2024-12-03
fix: add openssl dependency to tycho-client

louise-poole:
to avoid a user needing to install it, or libssl at runtime

### DIFF
--- a/tycho-client/Cargo.toml
+++ b/tycho-client/Cargo.toml
@@ -33,6 +33,8 @@ tracing-subscriber = { version = "0.3.17", default-features = false, features = 
 ] }
 clap = { workspace = true, features = ["derive", "env"] }
 tracing-appender = { workspace = true }
+# link openssl statically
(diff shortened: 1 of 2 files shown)


tycho-client/Cargo.toml
```diff
@@ -33,6 +33,8 @@ tracing-subscriber = { version = "0.3.17", default-features = false, features =
 ] }
 clap = { workspace = true, features = ["derive", "env"] }
 tracing-appender = { workspace = true }
+# link openssl statically
+openssl = { version = "0.10", features = ["vendored"] }
```
kayibal, line 39: Maybe feature gate this?

kayibal (approved):

tycho-client/Cargo.toml
louise-poole: I found a way to conditionally include it depending on the target os. Thought that was cleaner than feature gating it - let me know what you think.

PR #454, opened 2024-12-06
feat: add tycho-indexer metrics 

louise-poole:


tycho-indexer/src/substreams/stream.rs
```diff
@@ -98,6 +98,14 @@ fn stream_blocks(
                     for await response in stream {
                         match process_substreams_response(response).await {
                             BlockProcessedResult::BlockScopedData(block_scoped_data) => {
+                                let time_ms = SystemTime::now()
+                                    .duration_since(UNIX_EPOCH)
+                                    .expect("Time went backwards")
+                                    .as_millis();
+
+                                gauge!("substreams_block_number").set(block_scoped_data.final_block_height as f64);
```
kayibal: This is the finalized block number, not the current blocks number.

tycho-indexer/src/substreams/stream.rs
```diff
@@ -98,6 +98,14 @@ fn stream_blocks(
...
+                                gauge!("substreams_block_timestamp").set(time_ms as f64);
```
kayibal: Confusing name imo, this should be `substreams_block_witnessed_ts` or similar

tycho-indexer/src/substreams/stream.rs
```diff
@@ -103,6 +104,14 @@ fn stream_blocks(
...
                         match process_substreams_response(response).await {
                             BlockProcessedResult::BlockScopedData(block_scoped_data) => {
+                                if let Some(block) = block_scoped_data.clock.clone() {
+                                    if let Some(block_ts) = block.timestamp {
+                                        let now = SystemTime::now().duration_since(UNIX_EPOCH).expect("Time went backwards!?").as_secs();
+                                        let lag = max(now - block_ts.seconds as u64, 0);
+                                        gauge!("substreams_lag_seconds", "extractor" => extractor_id.clone()).set(lag as f64);
+                                    }
+                                };
+
```
kayibal, line 114: Yeah much simpler! A bit surprised that these fields are optional though. Any idea why?

tycho-core/src/dto.rs
```diff
@@ -208,7 +208,6 @@ pub struct Transaction {
 }
 
 impl Transaction {
-    #[allow(clippy::too_many_arguments)]
```
tvinagre, line 211: Unrelated to the PR but ok

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -190,6 +192,11 @@ where
             );
             state.last_report_ts = now;
             state.last_report_block_number = block.number;
+            gauge!(
+                "extractor_sync_time_remaining",
```
tvinagre: In this case I think it's nice to add the unit
```suggestion
                "extractor_sync_remaining_minutes",
```

tycho-indexer/src/extractor/runner.rs
```diff
@@ -173,6 +175,11 @@ impl ExtractorRunner {
                             Some(Ok(BlockResponse::New(data))) => {
                                 let block_number = data.clock.as_ref().map(|v| v.number).unwrap_or(0);
                                 tracing::Span::current().record("block_number", block_number);
+                                gauge!("extractor_current_block", "extractor_id" => id.to_string()).set(block_number as f64);
```
tvinagre:
```suggestion
                                gauge!("extractor_current_block_number", "extractor_id" => id.to_string()).set(block_number as f64);
```

tycho-indexer/src/extractor/runner.rs
```diff
@@ -188,6 +195,9 @@ impl ExtractorRunner {
                                         return Err(err);
                                     }
                                 }
+
+                                let duration = start_time.elapsed();
+                                histogram!("block_processing_time").record(duration.as_millis() as f64);
```
tvinagre:
```suggestion
                                histogram!("block_processing_time_ms").record(duration.as_millis() as f64);
```

tycho-indexer/src/main.rs
```diff
@@ -118,6 +119,11 @@ fn create_tracing_subscriber() {
 
 /// Creates and runs the Prometheus metrics exporter using Actix Web.
 pub fn create_metrics_exporter() -> tokio::task::JoinHandle<()> {
+    describe_gauge!(
+        "extractor_sync_time_remaining",
+        "Time remaining for the extractor to sync, measured in minutes."
```
tvinagre: Ah, you did describe them here :)
Does this appear on Grafana when selecting VictoriaMetrics metrics?

tycho-indexer/src/services/rpc.rs
```diff
@@ -6,10 +6,18 @@ use actix_web::{web, HttpResponse, ResponseError};
 use anyhow::Error;
 use chrono::{Duration, Utc};
 use diesel_async::pooled_connection::deadpool;
+use metrics::counter;
 use reqwest::StatusCode;
 use thiserror::Error;
 use tracing::{debug, error, info, instrument, trace, warn};
 
+use tycho_core::{
```
tvinagre, line 14: A lot of formatting changes - I'm assuming this isn't being covered by CI?

tycho-indexer/src/substreams/stream.rs
```diff
@@ -144,7 +163,7 @@ fn stream_blocks(
                     // We failed to connect and will try again; this is another
                     // case where we actually _want_ to back off in case we keep
                     // having connection errors.
-
+                    counter!("substreams_failure", "module" => output_module_name.clone(), "cause" => "connection_error").increment(1);
```
tvinagre, line 166: Some metrics you add a "total" in the end, like (rpc_requests_total) - and other you leave it out - is there a reason? I'd like to have a pattern here

tvinagre (approved):
Looking good! Smol comments on metric names

tycho-indexer/src/substreams/stream.rs
louise-poole: No clue - I ran this a few times and had a test metric to log whenever it was None and it was never logged. It seems like in general we can expect a value here.

tycho-core/src/dto.rs
louise-poole: Got distracted while we were all debating what path to take for the lag metric 😆  I added this skip somewhere else within the PR and noticed it was used a lot where it wasn't needed.

tycho-indexer/src/main.rs
louise-poole: According to the docs, this should appear. In my local test I haven't seen it though. I think I will rather rename the metrics as you've suggested to be 100% clear.

tycho-indexer/src/substreams/stream.rs
louise-poole: I actually started to remove them from the metric names as it was making them long without adding any info. I missed this one I suppose.

tycho-indexer/src/services/rpc.rs
louise-poole: No - for the most part our code follows the pattern of: std imports, external imports, repo imports, crate imports. If I worked on a file that broke this pattern I fixed it, but these are inconsequential changes only done for consistency.

PR #460, opened 2024-12-17
fix: improve websocket metrics with extended metadata

louise-poole:
With this we will be able to track the following:
- the active websocket connections as well as which tycho-client version they use and what api_key they use (to group connections by user)
- the active exchange subscriptions so that we can see which protocols are actively being streamed

Design decision: we use the **gauge** metric to track which connections/subscriptions are active. We use the id to label the connection/subscription as the id is known at the time of disconnect and can be used to properly update the metric to reflect which connection/subscription ended.
We use the **counter** metric to link metadata to the id labeled in the 'active' connections gauge. This allows us to report info that is not known at the time of disconnect (only id is known there) and on our dashboard it lets us link the metadata to the active connections.

### DIFF
--- a/tycho-indexer/src/services/ws.rs
+++ b/tycho-indexer/src/services/ws.rs
@@ -120,17 +120,7 @@ impl WsActor {
                     .to_string()
             })
             .unwrap_or_default();
-        let api_key = req
-            .headers()
-            .get("authorization")
-            .map(|value| {
-                value
-                    .to_str()
-                    .unwrap_or_default()
-                    .to_string()
-            })
-            .unwrap_or_default();
-        counter!("websocket_connections_metadata", "id" => ws_actor.id.to_string(), "client_version"=> user_agent.clone(), "api_key" => api_key.clone()).increment(1);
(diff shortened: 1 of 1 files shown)


tycho-indexer/src/services/ws.rs
```diff
@@ -107,7 +107,32 @@ impl WsActor {
...
+            .map(|value| {
+                value
+                    .to_str()
+                    .unwrap_or_default()
+                    .to_string()
+            })
+            .unwrap_or_default();
+        let api_key = req
+            .headers()
+            .get("authorization")
```
tvinagre: It's not a good practice to log API Keys - as they are personalized and sensitive. It's better to inject the username on the authentication layer and collect it here

louise-poole: Okay, that's a bigger change than I was wanting to do so I will just remove the api key part entirely. We don't easily have access to the username when creating this connection request header.

tvinagre (approved):
LGTM!

PR #469, opened 2024-12-28
fix: add last_finalized_block and optimize index calculation

zach030:
### Summary
In the current `drain_new_finalized_blocks` function, the index of the finalized_block is determined by:
```rust
let target_index = self.find_index(|b| b.block().number == final_block_height);
```
However, the subsequent use of `split_off`:

```rust
let mut temp = self.block_messages.split_off(idx);
std::mem::swap(&mut self.block_messages, &mut temp);
```
excludes the finalized_block from the return value because `split_off` uses an open interval. This causes the finalized_block to be skipped when writing to the database.

### Details
To fix this, I adjusted the index to include the finalized_block. However, this introduced a new issue: subsequent calls to `find_index` would return None, triggering a `StorageError::NotFound` error in strict mode (default).

To resolve this, I added `last_finalized_block` to cache the last finalized block height. When the finalized_block changes, the function uses `find_index` to locate the old block's index and increments it by 1. Otherwise, it returns 0.

### Key Changes
- Fixed split_off excluding the finalized_block.
- Added last_finalized_block to handle index calculation and avoid Block NotFound errors.

### DIFF
--- a/tycho-indexer/src/extractor/reorg_buffer.rs
+++ b/tycho-indexer/src/extractor/reorg_buffer.rs
@@ -58,6 +58,7 @@ impl TryFrom<BlockOrTimestamp> for BlockNumberOrTimestamp {
 pub(crate) struct ReorgBuffer<B: BlockScoped> {
     block_messages: VecDeque<B>,
     strict: bool,
+    last_finalized_block: Option<u64>,
 }
 
 /// The finality status of a block or block-scoped data.
@@ -76,7 +77,7 @@ where
     B: BlockScoped + std::fmt::Debug,
 {
     pub(crate) fn new() -> Self {
-        Self { block_messages: VecDeque::new(), strict: false }
+        Self { block_messages: VecDeque::new(), strict: false, last_finalized_block: None }
     }
 
     /// Inserts a new block into the buffer. Ensures the new block is the expected next block,
@@ -105,7 +106,12 @@ where
         &mut self,
         final_block_height: u64,
     ) -> Result<Vec<B>, StorageError> {
-        let target_index = self.find_index(|b| b.block().number == final_block_height);
+        let target_index = if self.last_finalized_block.replace(final_block_height) != Some(final_block_height) {
+            self.find_index(|b| b.block().number == final_block_height).map(|idx| idx + 1)
+        } else {
+            Some(0)
+        };
+
         let first = self
             .get_block_range(None, None)?
             .next()


zach030:
@zizou0x Please review. Thanks!

royvardhan:
I was about to open PR for the same issue. No worries. 

@zach030 Have you taken a look at this test? https://github.com/zach030/tycho-indexer/blob/95a713e023463b9e48190d78048b59e0c9cbb537/tycho-indexer/src/extractor/reorg_buffer.rs#L680

royvardhan:
Also @zizou0x my question is that the latest finalized block stayed in the buffer, was that intentional?

zach030:
> latest finalized block stayed in the buffer

I missed this test, and I’m not sure if the latest finalized block should stay in the buffer.

royvardhan:
> > latest finalized block stayed in the buffer
> 
> 
> 
> I missed this test, and I’m not sure if the latest finalized block should stay in the buffer.
> 
> 

If the latest finalized block in the buffer was intentional, then I can share my workaround in here so that you can implement it without having to change any test.

tycho-indexer/src/extractor/reorg_buffer.rs
```diff
@@ -105,7 +106,12 @@ where
         &mut self,
         final_block_height: u64,
     ) -> Result<Vec<B>, StorageError> {
-        let target_index = self.find_index(|b| b.block().number == final_block_height);
+        let target_index = if self.last_finalized_block.replace(final_block_height) != Some(final_block_height) {
+            self.find_index(|b| b.block().number == final_block_height).map(|idx| idx + 1)
```
kayibal, line 110: Isn't this equivalent and more concise? (under the assumption that blocks are ordered and without gaps)
```suggestion
            self.find_index(|b| b.block().number > final_block_height)
```

tycho-indexer/src/extractor/reorg_buffer.rs
```diff
@@ -105,7 +106,12 @@ where
         &mut self,
         final_block_height: u64,
     ) -> Result<Vec<B>, StorageError> {
-        let target_index = self.find_index(|b| b.block().number == final_block_height);
+        let target_index = if self.last_finalized_block.replace(final_block_height) != Some(final_block_height) {
```
kayibal, line 109: So this breaks down to:
```rust
if let Some(prev_final_block_height) = self.last_finalized_block {
   if prev_final_block_height == final_block_height {
       self.find_index(|b| b.block().number == final_block_height).map(|idx| idx + 1)
   } else {
       // Inner case: here may have mixed up buffers, should we throw an error?
       Some(0)
   }
} else {
    // Outer case: this case happens if finalized block height is not yet set
    Some(0)
}
```

So if we have no previously recorded final block height we simply drain the entire buffer and we do the same in case there is a mismatch in finalized block heights. The outer case may be handled by requiring last finalized block height upon initialisation instead of setting it to `None` but I am not too concerned about this one. 

I wonder if in the inner case we shouldn't throw an error instead to be more on the cautious side.

kayibal (commented):
Thanks @zach030 for taking this on. I think things are generally looking very good. I have some minor concerns we should quickly discuss.

tycho-indexer/src/extractor/reorg_buffer.rs
zach030: Thanks for the breakdown, but there are a few issues with it.

Currently, `split_off(idx)` excludes the block matching `final_block_height`. For example, if the buffer is [20, 21, 22, 23] and final_block_height = 23, `find_index` returns idx = 3, but `split_off(3)` removes 23, leaving [20, 21, 22]. This causes 23 to remain in the buffer.

Regarding the "Inner case" you mentioned, I guess it refers to when `prev_final_block_height == final_block_height`, meaning the finalized_block is unchanged. In this case, idx = 0 applies, and no return blocks is needed. The buffer only needs to be drained when `prev_final_block_height != final_block_height`.

If we expand the details, it should be equivalent to:
```rust
let target_index = if let Some(prev_final_block_height) = self.last_finalized_block {
    if prev_final_block_height == final_block_height {
        Some(0);
    } else {
        self.find_index(|b| b.block().number > final_block_height);
    }
} else{
    Some(0);
};
self.last_finalized_block = Some(final_block_height);
```

kayibal: Thanks for the clarification it makes sense now. If `target_index = Some(0)` nothing is drained, correct? 

I think I might have encountered another issue here: during the first call to this method, `self.last_finalized_block = None` So we don't drain anything and on the second call we would then drain twice the usual size?

zach030: Yes, if target_index = Some(0), nothing is drained.

About the issue you mentioned, you're right that it exists, but to clarify: the first call refers to the first block received after indexer starts. However, since `last_finalized_block` is updated with each call, it will not be `None` from the second call onwards.

Here's an example:
1. First call: block = 100, final_block_height = 90. last_finalized_block is None and the returned index will be 0, nothing be drained. last_finalized_block will be updated to 90.
2. Second call: block = 101, final_block_height = 90. last_finalized_block is now 90, matches the current final_block_height, the returned index will still be 0, and still nothing be drained.
3. Subsequent calls: For instance, when block = 120 and final_block_height = 110, last_finalized_block is 90, and the buffer will return data range from first block 100 to final_block 110

In short, the buffer only drains blocks when `final_block_height` exceeds `last_finalized_block`. Since `last_finalized_block` is updated with each call, it won’t result in double-draining on the second call.

kayibal: Thanks makes perfect sense now.

kayibal:
@zach030 Sorry for the delay here but I think we should be good to merge this once CI passes.

Issue #467, opened 2024-12-27
Getting `Internal Server Error` when no `Version` is specified


zizou0x:
### Description
When no version is specified for state queries, Tycho should default to the latest version.
This behaviour was implemented and functioning correctly in the past (it was fixed in https://github.com/propeller-heads/tycho-indexer/pull/352) but appears to be broken now.

#### Example
This query should return the latest state:

```
curl -X 'POST' \
  'http://0.0.0.0:4242/v1/protocol_state' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "chain": "ethereum",
  "include_balances": true,
  "pagination": {
    "page": 0,
    "page_size": 1
  },
  "protocol_system": "uniswap_v2"
}'
```

`Version` defaults to the current timestamp if not specified. When trying the above query we're getting this `ReorgBufferError`:

```
Error while getting protocol states. error=Failed to apply pending deltas: ReorgBufferError: Could not find Block with id `Timestamp(2024-12-27T12:37:06.659388194)`! body=Json(ProtocolStateRequestBody { protocol_ids: None, protocol_system: "uniswap_v2", chain: Ethereum, include_balances: true, version: VersionParam { timestamp: Some(2024-12-27T12:37:06.659388194), block: None }, pagination: PaginationParams { page: 0, page_size: 1 } })
```

### Expected Behavior   
Tycho should default to the latest version when no version is provided for state queries.

### Actual Behavior   
Tycho returns `Internal Server Error`.

PR #470, opened 2024-12-28
feat: predefined behaviour for latest ts

royvardhan:
Closes: #467 

The fix modifies timestamp handling in get_block_range to:
- Return the latest block when current timestamp is provided (default case)
- Find appropriate block range for historical timestamps
- Maintain existing behavior for block number queries

This makes protocol_state behavior consistent with contract_state for timestamp-based queries.

### DIFF
--- a/tycho-indexer/src/extractor/reorg_buffer.rs
+++ b/tycho-indexer/src/extractor/reorg_buffer.rs
@@ -192,27 +192,60 @@ where
         tracing::Span::current().record("buffered_range", buffered_range.as_str());
 
         let start_index = if let Some(version) = start_version {
-            self.find_index(|b| !version.greater_than(&b.block()))
-                .ok_or_else(|| {
-                    StorageError::NotFound("Block".to_string(), format!("{:?}", version))
-                })?
+            match version {
+                BlockNumberOrTimestamp::Timestamp(ts) => {
+                    if let Some(last) = self.block_messages.back() {
+                        if ts > last.block().ts {
+                            // Return the latest block when timestamp is greater than last block
+                            self.block_messages.len() - 1
+                        } else {
+                            self.find_index(|b| b.block().ts >= ts)
+                                .unwrap_or(0)
+                        }
+                    } else {
+                        0
+                    }
+                }
+                _ => self
+                    .find_index(|b| !version.greater_than(&b.block()))
+                    .ok_or_else(|| {
+                        StorageError::NotFound("Block".to_string(), format!("{:?}", version))
+                    })?,
+            }
         } else {
             0
         };
 
         let end_index = if let Some(version) = end_version {
-            let end_idx = self
-                .find_index(|b| !version.greater_than(&b.block()))
-                .ok_or_else(|| {
-                    StorageError::NotFound("Block".to_string(), format!("{:?}", version))
-                })?;
+            match version {
+                BlockNumberOrTimestamp::Timestamp(ts) => {
+                    if let Some(last) = self.block_messages.back() {
+                        if ts > last.block().ts {
+                            // Include the latest block when timestamp is greater than last block
+                            self.block_messages.len()
+                        } else {
+                            self.find_index(|b| b.block().ts > ts)
+                                .unwrap_or(self.block_messages.len())
+                        }
+                    } else {
+                        self.block_messages.len()
+                    }
+                }
+                _ => {
+                    let end_idx = self
+                        .find_index(|b| !version.greater_than(&b.block()))
+                        .ok_or_else(|| {
+                            StorageError::NotFound("Block".to_string(), format!("{:?}", version))
+                        })?;
 
-            if end_idx < start_index {
-                return Err(StorageError::Unexpected(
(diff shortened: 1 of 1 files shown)


royvardhan:
While this fixes the issue, I am thinking of a more robust fix.

royvardhan:
> While this fixes the issue, I am thinking of a more robust fix.

Problem: Latest timestamp comes from Version default, which results in storage error because when the lookup happens, it cannot find block for that timestamp as latest ts is greater than any block. 

Proposed solution: Instead of relying on Version default, we should make the version param optional, so that when None is provided, it gives the latest block data, and when out of bound Block or TS is provided, it errors as expected.

royvardhan:
@kayibal My previous changes were limiting the upper bound of the get_block_range so that when a future timestamp (Now and greater) is provided, it will always return the latest block data without returning storage error. That fixed the issue, but the edge case was that anyone querying with a future block (not existing one) will still get a response and may mistake it with the data of that block. Instead, I made the version param optional so that the following scenarios satisfies: 

1. Explicit latest data retrieval via version=None
2. Predictable errors for invalid block/timestamp queries
3. No ambiguity between requested vs returned block data

Let me know your thoughts.

zizou0x:
Hey @royvardhan, thanks for your contribution! 

I like your changes, they make sense. However, this introduces a breaking change on our dto struct which can become a big headache for our users. Would there be a way to do it without breaking the interfaces?

@tvinagre @kayibal any input here?

royvardhan:
> Hey @royvardhan, thanks for your contribution! 
> 
> 
> 
> I like your changes, they make sense. However, this introduces a breaking change on our dto struct which can become a big headache for our users. Would there be a way to do it without breaking the interfaces?
> 
> 
> 
> @tvinagre @kayibal any input here?

Thanks! There is also a third option without any breaking changes where we verify that the version is generated by default, if yes then:

1. `None` goes [here](https://github.com/propeller-heads/tycho-indexer/blob/9b8014544bb31b57a198235bcfc3ae0c48cafae7/tycho-indexer/src/services/rpc.rs#L440)

2. Which then triggers `update_native_states` with `None` and `get_block_range` gets both `start` and `end` as `None`.

[Here](https://github.com/propeller-heads/tycho-indexer/blob/9b8014544bb31b57a198235bcfc3ae0c48cafae7/tycho-indexer/src/services/deltas_buffer.rs#L165)

Allowing it to return full buffer range [here](https://github.com/propeller-heads/tycho-indexer/blob/9b8014544bb31b57a198235bcfc3ae0c48cafae7/tycho-indexer/src/extractor/reorg_buffer.rs#L217)

zizou0x:
Hey @royvardhan, I'm having a deeper look now, can you explain in more detail what exactly causes the issue, point me to the line that is bugged please?

royvardhan:
> Hey @royvardhan, I'm having a deeper look now, can you explain in more detail what exactly causes the issue, point me to the line that is bugged please?

This [error](https://github.com/royvardhan/tycho-indexer/blob/31d721cc03a6c79b20de9c7b7bb34062b4390756/tycho-indexer/src/extractor/reorg_buffer.rs#L207) is triggered because of latest timestamp.

royvardhan:
@zizou0x I think we are good now, no more a breaking change.

tycho-indexer/src/services/rpc.rs
```diff
@@ -141,7 +141,15 @@ where
         &self,
         request: dto::StateRequestBody,
     ) -> Result<dto::StateRequestResponse, RpcError> {
-        let at = BlockOrTimestamp::try_from(&request.version)?;
+        let at: BlockOrTimestamp = match &request.version {
+            // Match the version to the latest block if the timestamp is the current time
+            v if v.timestamp.unwrap().trunc_subsecs(0) ==
+                Utc::now().naive_utc().trunc_subsecs(0) =>
+            {
+                BlockOrTimestamp::Block(BlockIdentifier::Latest(request.chain.into()))
```
zizou0x: I think this doesn't work as expected. 
Here is why:

If we use this `BlockIdentifier::Latest` here, in `calculate_versions` it goes into this branch of the match https://github.com/propeller-heads/tycho-indexer/blob/e19acf6ae2a88bbfacfabe6c8b41c45ce53a6fd4/tycho-indexer/src/services/rpc.rs#L271-L276

So we're getting the block number of the most recent block **in the database** and we're completely skipping the revert buffer.

zizou0x (commented):
Thanks @royvardhan for your efforts. Please correct me if I'm wrong but I think the current behaviour isn't the one we expect. 

We should explore updating `get_block_range` to return the whole buffer if the version is given as a timestamp and it is greater than every block version in the buffer. If that works we need to make sure this is explained in the function's docstring and it is not breaking somewhere else.

royvardhan:
@zizou0x So three solutions for this bug so far, either I am dumb or these were overkill haha!

zizou0x (commented):
LGTM, thanks @royvardhan

zizou0x (approved):

PR #473, opened 2025-01-03
perf: move database commits off critical path

### LINKED ISSUE #472: Reduce latency by avoiding database commit blocking
Database commits currently occur on [the critical path](https://github.com/propeller-heads/tycho-indexer/blob/f15f2e2ffa17cd4f34a4a8493bdb6e8892a7789e/tycho-indexer/src/extractor/protocol_extractor.rs#L609), which slows down the process. 

Since this operation is typically time-consuming, we should avoid waiting for it when the extractor isn’t syncing. Instead, we can submit and store the future and await with the next message if it is set. This approach allows database commits to run in the background, ideally completing before the next block arrives. 

Taking this idea further, we should support storing multiple futures to effectively handle short bursts of larger data volumes (which are likely to happen on L2s).

### Outline of required changes

The following outline serves as a conceptual guide to illustrate the proposed changes. It’s not intended to be prescriptive or definitive.

```rust
pub struct ProtocolExtractor<G, T> {
    ...
    reorg_buffer: Mutex<ReorgBuffer<BlockUpdateWithCursor<BlockChanges>>>,
    // A queue of pending database tasks
    db_futures: Mutex<VecDeque<JoinHandle<Result<(), StorageError>>>>
}
```

We can then store the tasks using:

```rust

// todo: remove all finished tasks from the db_futures

while let Some(msg) = msgs.next() {
    // Force a database commit if we're not syncing and this is the last block to be
    // sent. Otherwise, wait to accumulate a full batch before
    // committing.
    let force_db_commit = if is_syncing { false } else { msgs.peek().is_none() };

    let fut = tokio::spawn(self.gateway.advance(
        msg.block_update(),
        msg.cursor(),
        force_db_commit,
    ));
    // todo: ensure this doesn't grow to big.
    self.db_futures.push_back(fut)

}
```

royvardhan:
Closes #472

### DIFF
--- a/tycho-indexer/src/extractor/protocol_extractor.rs
+++ b/tycho-indexer/src/extractor/protocol_extractor.rs
@@ -6,7 +6,6 @@ use std::{
 
 use async_trait::async_trait;
 use chrono::NaiveDateTime;
-use futures03::FutureExt;
 use metrics::gauge;
 use mockall::automock;
 use prost::Message;
@@ -46,10 +45,6 @@ use crate::{
     pb,
     pb::sf::substreams::rpc::v2::{BlockScopedData, BlockUndoSignal, ModulesProgress},
 };
-
-// Todo: Change in PR evaluation
-const MAX_PENDING_COMMITS: usize = 100;
-const TARGET_PENDING_COMMITS: usize = 50;
 
 pub struct Inner {
     cursor: Vec<u8>,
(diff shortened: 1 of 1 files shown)


royvardhan:
For anyone reviewing this WIP pr, there are two things I think I will be adding:

1. Safeguard from potential race conditions, particularly at achieving lock at `db_futures`, and inside the `start_transaction` method. 

2. At the time of pushing futures into `db_futures`, pushing the commit task too:

`struct DbCommitTask {
changes: BlockChanges,
cursor: String,
force_commit: bool,
block_number: u64
}`

So that for every future we will have the commit task data, and at the time of evaluating the stored futures (and clearing them), we can retry the ones which failed (because if block n fails, all the blocks n+1..n+j will also fail because of parent child relation?)

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -604,9 +687,17 @@ where
...
+                    gateway
+                        .advance(msg.block_update(), msg.cursor(), force_db_commit)
+                        .await
+                });
+
+                self.db_futures
+                    .lock()
+                    .await
+                    .push_back(fut);
             }
```
louise-poole: It's essential that we execute these `advance` futures in order. I see you mentioned in a PR comment that you expect commits for future blocks to fail if a previous block commit didn't complete yet (due to parent block constraints in the db). It would be preferable if we did not use db resources to attempt commits we know will fail - we should instead ensure we only commit in order.

louise-poole: Another comment here with regards to maintaining order: we have a few read functions on this extractor... these read functions must also be executed in relation to when they were called (any advance futures created before the read was called much first execute).
A work-around for this would be to not remove the block messages from the reorg_buffer until they are successfully committed (so that reads can access them from the buffer and not need to wait for the db commits to complete). However this introduces a new problem where we'll be cloning the messages and run the risk of committing the same message twice if we get a new finalised block before the previous ones were finished committing (I'm thinking about potential chains with fast block times and 1 block finalisation).

louise-poole (commented):
Thanks for tackling this! It was on my list to look into next - really appreciate you taking it on! I have a few concerns about maintaining the order of commits + handling reads: please see my comments below.

tycho-indexer/src/extractor/protocol_extractor.rs
royvardhan: > It's essential that we execute these `advance` futures in order. I see you mentioned in a PR comment that you expect commits for future blocks to fail if a previous block commit didn't complete yet (due to parent block constraints in the db). It would be preferable if we did not use db resources to attempt commits we know will fail - we should instead ensure we only commit in order.

Thanks for the review! In the current PR code we are running advance fut in order. 

So here is the flow I was talking about: 

1. Run advance like before but in fut, in a separate thread so that we don't block the critical path. 

2. Check the accumulated futures if any advance returned any errors. 

3. Retry them?

louise-poole: Yes, you are invoking them in order, but since we are no longer blocking on the `advance` call, subsequent `advance` calls could be invoked before the first one finishes executing. If the first message has a large amount of data and the second one contains a message with no protocol changes, the second call will likely complete before the first has.
Ideally we should only start executing the second `advance` after the first has fully completed.

royvardhan: Yes that makes sense now. Thank you for clarifying. Will update the code to always check if previous future completed and then only start a new one. 🫡

royvardhan: Hi, I have updated the code to await previous fut.

PR #477, opened 2025-01-07
fix: fix protocol system deletion script account balances bug

louise-poole:
Context: in the tycho database schema, we use `accounts` for both token contracts and vm pool contracts. When we use the account as a token, we do not track balances of the account. When we use it as a pool, we do track balances (specifically eth balances) and store them in account_balance. 

Issue: In some special cases the same contract acts as both a pool and a token. Currently, when we delete a protocol where the account is used as a pool, we cascade delete the balances associated with it. However if the account is also used as a token outside of the protocol, we cannot delete the account and so the balances remain as well. This causes issues later if we resync the protocol that was deleted as the account balances already exist and cannot be re-created (only updated).

Solution: The account balances should be deleted as they're related to component accounts and not token accounts.

Identification of issue: This issue was identified when deleting vm:curve from tycho db: some curve pools are used as tokens in other systems and their account balances were left in the db. This caused errors when we tried to resync curve as you cannot create a new balance entry for an account balance that already exists (uniqueness is enforced on the account_id value... only 1 eth balance is allowed per account).

### DIFF
--- a/tycho-storage/sql_scripts/remove_protocol_system.sh
+++ b/tycho-storage/sql_scripts/remove_protocol_system.sh
@@ -141,13 +141,12 @@ WHERE account_id IN (
     JOIN protocol_component pc ON pchc.protocol_component_id = pc.id
     JOIN protocol_system ps ON pc.protocol_system_id = ps.id
     WHERE ps.name = :'protocol_system_name'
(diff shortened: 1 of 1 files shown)


zizou0x (approved):
LGTM

PR #479, opened 2025-01-08
refactor(indexer): remove hardcoded vault addresses in `map_vault`

zizou0x:
We now have the `balance_owner` attribute that dynamically gives us the owner. This commit remove the use of these hardcoded addresses and use `balance_owner` instead.


tycho-core/src/storage.rs
```diff
@@ -335,14 +335,14 @@ pub trait ProtocolGateway {
     /// # Parameters
     /// - `chain` The chain of the component
     /// - `system` The protocol system this component belongs to
-    /// - `id` The external id of the component e.g. address, or the pair
+    /// - `ids` The external ids of the components e.g. addresses, or the pairs
```
louise-poole, line 338: Why change this to singular? the param expects a slice which can contain multiple ids, no?

louise-poole (commented):
Looks good! Just unsure about the changes you made to the ProtocolGateway? The actual balance owner changes look 100s.

tycho-core/src/storage.rs
zizou0x: I changed it from singular to plural

louise-poole: Let's pretend this never happened 😇

tycho-core/src/storage.rs
```diff
@@ -335,14 +335,14 @@ pub trait ProtocolGateway {
...
     /// - `at` The version at which the state is valid at.
     async fn get_protocol_states(
         &self,
         chain: &Chain,
         at: Option<Version>,
         system: Option<String>,
-        id: Option<&[&str]>,
+        ids: Option<&[&str]>,
```
louise-poole, line 345: I think there are a few places that implement this trait - they'll need to be updated too no? Like on the CachedGateway...

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -409,7 +408,14 @@ where
                     // Filtering to keep only components with ChangeType::Creation
                     .filter(|(_, c_change)| c_change.change == ChangeType::Creation)
                     .filter_map(|(c_id, change)| {
-                        map_vault(&change.protocol_system)
+                        tx.state_updates
```
tvinagre, line 411: would be nice to have a test for this

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -1520,7 +1520,7 @@ mod test {
         ])
     }
 
-    #[test_log::test(tokio::test)]
+    #[test_log::test(tokio::test(flavor = "multi_thread", worker_threads = 2))]
```
louise-poole, line 1523: My favourite ice-cream flavor: multi_thread 😂

louise-poole (approved):
LGTM!

tycho-indexer/src/extractor/protocol_extractor.rs
zizou0x: I also found it fun that they called it flavor

PR #480, opened 2025-01-08
feat: add user identity to ws connection metrics

louise-poole:
Added user identity (a new header added during connection authentication) to ws connection metrics. This will allow us to monitor:
- how many connections each user has active
- what version client they're using
- which protocols they have active subscriptions to

This is to give us insight as to how tycho is currently being used as well as allow us to provide smoother easier support to our clients.

### DIFF
--- a/tycho-indexer/src/services/ws.rs
+++ b/tycho-indexer/src/services/ws.rs
@@ -115,7 +115,7 @@ impl WsActor {
             .map(|value| {
                 value
                     .to_str()
-                    .unwrap_or_default()
+                    .unwrap_or("unknown")
                     .to_string()
             });
         let ws_actor = WsActor::new(data, user_identity);
@@ -135,7 +135,7 @@ impl WsActor {
             "websocket_connections_metadata",
             "id" => ws_actor.id.to_string(),
             "client_version" => user_agent,
-            "user_identity" => ws_actor.user_identity.clone().unwrap_or_default()
+            "user_identity" => ws_actor.user_identity.clone().unwrap_or("unknown".to_string()),
         )
         .increment(1);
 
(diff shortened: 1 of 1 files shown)


tycho-indexer/src/services/ws.rs
```diff
@@ -120,7 +131,13 @@ impl WsActor {
                     .to_string()
             })
             .unwrap_or_default();
-        counter!("websocket_connections_metadata", "id" => ws_actor.id.to_string(), "client_version"=> user_agent.clone()).increment(1);
+        counter!(
+            "websocket_connections_metadata",
+            "id" => ws_actor.id.to_string(),
+            "client_version" => user_agent,
+            "user_identity" => ws_actor.user_identity.clone().unwrap_or_default()
```
zizou0x: What is the default in that case? Empty string? Maybe we should `unwrap_or("unknown")`?

zizou0x: Because internal connections (us) won't have this field because we don't go through nginx

tvinagre: Fine with either - as long as it doesn't fail

louise-poole: Will default it to "internal" then, all external access will defs have a user identity. Or do you think this is a bad approach as someone running their own instance of tycho might not use the same auth set up as us so all of their metrics will report as 'internal' here?

zizou0x: Yes I wouldn't call it internal because we're open source and we don't know how others will set this up with their gateways. I have a preference for something more generic like "unknown"

zizou0x (approved):

tvinagre (approved):

PR #482, opened 2025-01-10
feat: extend substreams block message metrics

louise-poole:
Added metrics for:
- substreams block message size
- chain reorg messages

### DIFF
--- a/tycho-indexer/src/substreams/stream.rs
+++ b/tycho-indexer/src/substreams/stream.rs
@@ -75,6 +75,7 @@ fn stream_blocks(
     extractor_id: String,
 ) -> impl Stream<Item = Result<BlockResponse, Error>> {
     let mut latest_cursor = cursor.unwrap_or_default();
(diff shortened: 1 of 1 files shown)


tycho-indexer/src/substreams/stream.rs
```diff
@@ -123,6 +126,8 @@ fn stream_blocks(
                                 // Reset backoff because we got a good value from the stream
                                 backoff = DEFAULT_BACKOFF.clone();
 
+                                counter!("chain_reorg", "extractor" => extractor_id.clone()).increment(1);
```
zizou0x: Does `counter!` also give us the timestamps? Also would be interesting to know the from and to block numbers of the reorg, but not mandatory

louise-poole: Yes - all metrics automatically have timestamp data. I'll look into adding the to and from block info too.

zizou0x (approved):

tvinagre (approved):

PR #483, opened 2025-01-10
docs: add tycho-indexer binary build instructions

louise-poole:


tycho-indexer/README.md
```diff
@@ -111,3 +111,25 @@ For more information about the flags, you can run:
...
+To build the tycho-indexer binary and make it globally accessible on your system, follow these steps:
+
+1. **Build the binary in release mode:**
+```bash
+cargo build --release
+```
+
+2. **Link the binary to a directory in your system's PATH:**
+```bash
+sudo ln -s $(pwd)/target/release/tycho-indexer /usr/local/bin/tycho-indexer
```
zizou0x, line 126: This works only if `/usr/local/bin/` is in the `PATH`. 
Should be the case most of the time but I'm not sure about some minimalist Linux or things like this. Maybe we should precise this requirement or offer an alternative?

zizou0x (approved):

PR #484, opened 2025-01-12
feat: add protocol system rpc

### LINKED ISSUE #478: Add `protocol_systems` RPC endpoint
### Context
Currently, we lack a straightforward way to identify which protocols are actively supported by the Tycho indexing service. This information is critical for developers and users relying on the service. While the documentation is manually maintained, it's prone to becoming outdated. A simple solution is to add a protocol_systems endpoint to the RPC, allowing real-time retrieval of supported protocols.

### Outline of required changes:
1) Add a `get_protocol_systems` function to the [protocol implementation](https://github.com/propeller-heads/tycho-indexer/blob/main/tycho-storage/src/postgres/protocol.rs#L31) of the PostgresGateway. 
```
    pub async fn get_protocol_systems(
        &self,
        chain: Chain,
    ) -> Result<WithTotal<Vec<String>>, StorageError> {
```
This function should utilise the existing [protocol system cache](https://github.com/propeller-heads/tycho-indexer/blob/main/tycho-storage/src/postgres/mod.rs#L422) and should not access the database directly. This cache is populated on start-up and contains all protocol_systems in the database.

2) Add `get_protocol_systems` to the [ProtocolGateway](https://github.com/propeller-heads/tycho-indexer/blob/main/tycho-core/src/storage.rs#L263) trait and implement it on the [CachedGateway](https://github.com/propeller-heads/tycho-indexer/blob/main/tycho-storage/src/postgres/cache.rs#L777).

3) Add a `protocol_systems` GET endpoint to the RPC. The endpoint should use the `get_protocol_systems` function defined above to return a paginated list of supported protocol_systems.

### Deliverables
- The new function in PostgresGateway.
- The updated RPC with the new endpoint.
- Unit tests for both the function and the endpoint.
- Documentation updates for the new endpoint, including API usage and response examples.

zach030:
This PR adds a new `GET /v1/protocol_systems` endpoint to fetch supported protocol systems in real time, using the existing cache and supports pagination with page and page_size.

The `chain` parameter is currently unused because protocol systems are not linked to specific chains in the cache or database. The association between protocol_systems and chain can be added in a follow-up PR when needed.

Unit tests and documentation updates are included.

Close #478

### DIFF
--- a/tycho-client/src/rpc.rs
+++ b/tycho-client/src/rpc.rs
@@ -21,8 +21,9 @@ use tycho_core::{
     dto::{
         Chain, PaginationParams, PaginationResponse, ProtocolComponentRequestResponse,
         ProtocolComponentsRequestBody, ProtocolStateRequestBody, ProtocolStateRequestResponse,
-        ResponseToken, StateRequestBody, StateRequestResponse, TokensRequestBody,
-        TokensRequestResponse, VersionParam,
+        ProtocolSystemsRequestBody, ProtocolSystemsRequestResponse, ResponseToken,
+        StateRequestBody, StateRequestResponse, TokensRequestBody, TokensRequestResponse,
+        VersionParam,
     },
     Bytes,
 };
@@ -373,6 +374,11 @@ pub trait RPCClient: Send + Sync {
         }
         Ok(all_tokens)
     }
+
+    async fn get_protocol_systems(
+        &self,
+        request: &ProtocolSystemsRequestBody,
+    ) -> Result<ProtocolSystemsRequestResponse, RPCError>;
 }
 
 #[derive(Debug, Clone)]
@@ -603,6 +609,39 @@ impl RPCClient for HttpRPCClient {
         })?;
 
         Ok(tokens)
+    }
+
+    async fn get_protocol_systems(
+        &self,
+        request: &ProtocolSystemsRequestBody,
+    ) -> Result<ProtocolSystemsRequestResponse, RPCError> {
+        let uri = format!(
+            "{}/{}/protocol_systems",
+            self.url
+                .to_string()
+                .trim_end_matches('/'),
+            TYCHO_SERVER_VERSION
+        );
+        debug!(%uri, "Sending protocol_systems request to Tycho server");
+        let response = self
+            .http_client
+            .post(&uri)
+            .json(request)
+            .send()
+            .await
+            .map_err(|e| RPCError::HttpClient(e.to_string()))?;
+
+        let body = response
+            .text()
+            .await
+            .map_err(|e| RPCError::ParseResponse(e.to_string()))?;
+        let protocol_systems = serde_json::from_str::<ProtocolSystemsRequestResponse>(&body)
+            .map_err(|e| {
+                error!("Failed to parse protocol systems response {:?}", &body);
+                RPCError::ParseResponse(e.to_string())
+            })?;
+
+        Ok(protocol_systems)
     }
 }
 
@@ -953,5 +992,42 @@ mod tests {
         mocked_server.assert();
         assert_eq!(response.tokens, expected);
         assert_eq!(response.pagination, PaginationResponse { page: 0, page_size: 20, total: 10 });
+    }
+
+    #[tokio::test]
+    async fn test_get_protocol_systems() {
+        let mut server = Server::new_async().await;
+        let server_resp = r#"
+        {
+            "protocol_systems": [
+                "system1",
+                "system2"
+            ],
+            "pagination": {
+                "page": 0,
+                "page_size": 20,
+                "total": 10
+            }
+        }
+        "#;
+        // test that the response is deserialized correctly
+        serde_json::from_str::<ProtocolSystemsRequestResponse>(server_resp).expect("deserialize");
+
+        let mocked_server = server
+            .mock("POST", "/v1/protocol_systems")
+            .expect(1)
+            .with_body(server_resp)
+            .create_async()
+            .await;
+        let client = HttpRPCClient::new(server.url().as_str(), None).expect("create client");
+
+        let response = client
+            .get_protocol_systems(&Default::default())
+            .await
+            .expect("get protocol systems");
+        let protocol_systems = response.protocol_systems;
+
+        mocked_server.assert();
+        assert_eq!(protocol_systems, vec!["system1", "system2"]);
     }
 }
--- a/tycho-client/src/rpc.rs
+++ b/tycho-client/src/rpc.rs
@@ -625,7 +625,7 @@ impl RPCClient for HttpRPCClient {
         debug!(%uri, "Sending protocol_systems request to Tycho server");
         let response = self
             .http_client
-            .post(&uri)
+            .get(&uri)
             .json(request)
             .send()
             .await
@@ -1014,7 +1014,7 @@ mod tests {
         serde_json::from_str::<ProtocolSystemsRequestResponse>(server_resp).expect("deserialize");
 
         let mocked_server = server
-            .mock("POST", "/v1/protocol_systems")
+            .mock("GET", "/v1/protocol_systems")
             .expect(1)
             .with_body(server_resp)
             .create_async()
--- a/tycho-client/src/rpc.rs
+++ b/tycho-client/src/rpc.rs
@@ -623,6 +623,7 @@ impl RPCClient for HttpRPCClient {
             TYCHO_SERVER_VERSION
         );
         debug!(%uri, "Sending protocol_systems request to Tycho server");
+        trace!(?request, "Sending request to Tycho server");
         let response = self
             .http_client
             .get(&uri)
@@ -630,7 +631,7 @@ impl RPCClient for HttpRPCClient {
             .send()
             .await
             .map_err(|e| RPCError::HttpClient(e.to_string()))?;
-
+        trace!(?response, "Received response from Tycho server");
         let body = response
             .text()
             .await
@@ -640,7 +641,7 @@ impl RPCClient for HttpRPCClient {
                 error!("Failed to parse protocol systems response {:?}", &body);
                 RPCError::ParseResponse(e.to_string())
             })?;
-
+        trace!(?protocol_systems, "Received protocol_systems response from Tycho server");
         Ok(protocol_systems)
     }
 }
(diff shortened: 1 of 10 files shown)


tycho-client/src/rpc.rs
```diff
@@ -604,6 +610,39 @@ impl RPCClient for HttpRPCClient {
...
+        request: &ProtocolSystemsRequestBody,
+    ) -> Result<ProtocolSystemsRequestResponse, RPCError> {
+        let uri = format!(
+            "{}/{}/protocol_systems",
+            self.url
+                .to_string()
+                .trim_end_matches('/'),
+            TYCHO_SERVER_VERSION
+        );
+        debug!(%uri, "Sending protocol_systems request to Tycho server");
```
louise-poole, line 625: Please add a trace log here too:
```suggestion
        debug!(%uri, "Sending protocol_systems request to Tycho server");
        trace!(?request, "Sending request to Tycho server");

```

tycho-client/src/rpc.rs
```diff
@@ -604,6 +610,39 @@ impl RPCClient for HttpRPCClient {
...
+        let response = self
+            .http_client
+            .get(&uri)
+            .json(request)
+            .send()
+            .await
+            .map_err(|e| RPCError::HttpClient(e.to_string()))?;
+
```
louise-poole: And a trace log here too:
```suggestion
        trace!(?response, "Received response from Tycho server");

```

tycho-client/src/rpc.rs
```diff
@@ -604,6 +610,39 @@ impl RPCClient for HttpRPCClient {
...
+        let body = response
+            .text()
+            .await
+            .map_err(|e| RPCError::ParseResponse(e.to_string()))?;
+        let protocol_systems = serde_json::from_str::<ProtocolSystemsRequestResponse>(&body)
+            .map_err(|e| {
+                error!("Failed to parse protocol systems response {:?}", &body);
+                RPCError::ParseResponse(e.to_string())
+            })?;
+
```
louise-poole: And a final trace log here please:
```suggestion
        trace!(?protocol_systems, "Received protocol_systems response from Tycho server");

```

tycho-indexer/src/services/rpc.rs
```diff
@@ -347,6 +347,39 @@ where
...
+                        .unwrap_or_default(),
+                ),
+            )),
+            Err(err) => {
+                error!(error = %err, "Error while getting protocol systems.");
+                Err(err.into())
+            }
+        }
+    }
+
```
louise-poole: Please move this to below `get_protocol_state_inner`. I would like to not split up the `get_protocol_state` and `get_protocol_state_inner` functions.

tycho-indexer/src/services/rpc.rs
```diff
@@ -861,6 +894,50 @@ pub async fn protocol_state<G: Gateway>(
...
+    // Call the handler to get protocol systems
+    let response = handler
+        .into_inner()
+        .get_protocol_systems(&body)
+        .await;
+
+    match response {
+        Ok(systems) => HttpResponse::Ok().json(systems),
+        Err(err) => {
+            error!(error = %err, "Error while getting protocol systems.");
```
louise-poole: Should add the request body here:
```suggestion
            error!(error = %err, ?body, "Error while getting protocol systems.");
```

tycho-storage/src/postgres/protocol.rs
```diff
@@ -1708,6 +1708,32 @@ impl PostgresGateway {
             .map_err(PostgresError::from)?;
         Ok(())
     }
+
+    pub async fn get_protocol_systems(
+        &self,
+        _chain: &Chain, // TODO: Add chain filter
```
louise-poole: Currently we plan to run a separate indexing service per chain, so will use this chain param at a higher level to redirect the requests to the correct rpc. Therefore internally we do not need to filter by chain. We do however need a validation check of sorts here... need to make sure the chain in the request matches the one the gateway is servicing. If I remember correctly there is a method called `get_chain_id` on our gateway that uses a chain cache similar to the protocol systems cache... we can use this to validate the chain and error if no chain exists for the chain requested.

louise-poole (commented):
Just a couple small requests, otherwise looks great!

tycho-storage/src/postgres/protocol.rs
zach030: done! I've added validation for the chain param

tycho-storage/src/postgres/protocol.rs
```diff
@@ -1711,9 +1711,10 @@ impl PostgresGateway {
 
     pub async fn get_protocol_systems(
         &self,
-        _chain: &Chain, // TODO: Add chain filter
+        chain: &Chain,
         pagination_params: Option<&PaginationParams>,
     ) -> Result<WithTotal<Vec<String>>, StorageError> {
+        let _ = self.get_chain_id(chain);
```
louise-poole: Hmmm... I see that `get_chain_id` panics if the entry is not found. While this is what we want for the usual use-case of this cache (writing to storage), it is not what we want for this rpc request validation use-case. I think it's safest to add a `exists_in_cache` fn to the `ValueIdTableCache` that returns a bool indicating if an entry exists. This way we can validate the chain in the request without crashing the rpc service.

zach030: Got it! I’ve added a cache existence check to avoid the panic.

tycho-storage/src/postgres/protocol.rs
```diff
@@ -1708,6 +1708,35 @@ impl PostgresGateway {
...
     }
+
+    pub async fn get_protocol_systems(
+        &self,
+        chain: &Chain,
+        pagination_params: Option<&PaginationParams>,
+    ) -> Result<WithTotal<Vec<String>>, StorageError> {
+        if !self.chain_id_cache.value_exist(chain) {
+            return Err(StorageError::NotFound("Chain".to_string(), chain.to_string()));
+        }
```
louise-poole, line 1719: Looks great! Lastly: please add a test for this error case.

louise-poole (commented):
I'm very happy with this now! Just one more test needed and then I think we're good.

tycho-storage/src/postgres/mod.rs
```diff
@@ -228,6 +228,26 @@ where
...
+
+    /// Checks if a database ID exists in the cache. Returns `true` if the ID is found,
+    /// otherwise returns `false`.
+    ///
+    /// # Arguments
+    ///
+    /// * `id` - The database ID to check for existence.
+    fn id_exist(&self, id: &i64) -> bool {
+        self.map_enum.contains_key(id)
+    }
```
louise-poole: This is unused. Let's remove it for now. It's easy to implement if someone ever needs it later.

tycho-storage/src/postgres/protocol.rs
zach030: done

tycho-storage/src/postgres/mod.rs
zach030: done

tycho-storage/src/postgres/protocol.rs
```diff
@@ -3743,6 +3743,29 @@ mod test {
...
 
+    #[tokio::test]
+    async fn test_get_protocol_systems_chain_not_exist() {
+        let mut conn = setup_db().await;
+        let _ = setup_data(&mut conn).await;
+        let gw = EVMGateway::from_connection(&mut conn).await;
+        let exp = ["ambient", "zigzag"]
+            .iter()
+            .map(|s| s.to_string())
+            .collect::<HashSet<_>>();
```
louise-poole: This looks unused?

zach030: Sorry, I deleted it

tycho-storage/src/postgres/protocol.rs
```diff
@@ -3663,6 +3692,76 @@ mod test {
...
+        assert_eq!(
+            res.entity
+                .into_iter()
+                .collect::<HashSet<_>>(),
+            exp
+        );
+    }
+
+    #[tokio::test]
+    async fn test_get_protocol_systems_with_pagination() {
```
louise-poole, line 3721: Please check CI - this test is currently failing.

louise-poole: The expected 'total' in the pagination response should still be 2 (there are 2 protocol systems available in total even though only 1 is returned).

zach030: Tested it locally, works fine. Setting up the test environment locally is a bit tricky 🥲

tvinagre: What's are the hard parts on setting it up? I'd like to see how can we make this easier

tycho-storage/src/postgres/protocol.rs
```diff
@@ -3721,26 +3721,18 @@ mod test {
...
             .expect("retrieving protocol systems failed!");
 
         assert_eq!(res.total, Some(exp.len() as i64));
-        assert_eq!(
-            res.entity
-                .into_iter()
-                .collect::<HashSet<_>>(),
-            exp
-        );
+        assert_eq!(res.entity.iter().collect_vec(), exp);
```
louise-poole: What I was suggesting was that we keep the test as it was (to ensure pagination cuts the data where expected instead of returning all) and just update this line:
```assert_eq!(res.total, Some(exp.len() as i64));```
to be:
```assert_eq!(res.total, Some(2));```

zach030: I see. The data order isn't fixed because it's from a HashMap cache. I plan to sort them first, then paginate to keep the order consistent.

tycho-storage/src/postgres/protocol.rs
zach030: Oh, I just checked the docs in tycho-storage. Previously, I relied on the GitHub CI yaml file to set up and run local tests.

tycho-storage/src/postgres/protocol.rs
zach030: Thanks for your patience! I added sorting and fixed the tests.

louise-poole (approved):
Nice! Very happy with this - thank you!

PR #492, opened 2025-01-22
feat: add AccountBalances to the protobuf messages

louise-poole:
Not sure why there are all those formatting changes to the generated protobuf file. I found the PR where the protobuf formatting check was added to CI and followed the instructions there to lint the file.

### DIFF
--- a/proto/README.md
+++ b/proto/README.md
@@ -1,0 +1,21 @@
+# Protobuf
+We use protobuf mainly to communicate with substreams. You can use the [language support extension](https://marketplace.visualstudio.com/items?itemName=pbkit.vscode-pbkit) and for formatting use the [buf extension for VS Code](https://marketplace.visualstudio.com/items?itemName=bufbuild.vscode-buf).
+1. To communicate with substreams, we always want to use their latest protobuf images. Usually the sf module is already committed but in case there was an update we can pull the latest version into the repo with:
+```
+buf export buf.build/streamingfast/substreams -o ./proto
+```
+2. Now that we have all our protobuf modules together we can generate the respective Rust code for them:
+```
+buf generate 
+```
+
+## Protobuf Linting
+1. To install the linter execute:
+```
+brew install bufbuild/buf/buf@1.28.1
+```
+2. Linting the files works like:
+```
+buf lint
+```
+It will return a list of errors that you need to fix. If not, then all the changes you made conform to the linter.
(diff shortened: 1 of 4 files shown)


kayibal (approved):

proto/README.md
```diff
@@ -0,0 +1,21 @@
+# Protobuf
```
tvinagre, line 1: nice nice but let's add this to the Gitbook please

tycho-indexer/src/pb/testing.rs
```diff
@@ -53,25 +53,13 @@ pub mod fixtures {
...
-                Bytes::from_str("0x6B175474E89094C44Da98b954EedeAC495271d0F")
-                    .unwrap()
-                    .0
-                    .to_vec(),
-                Bytes::from_str("0x6B175474E89094C44Da98b954EedeAC495271d0F")
-                    .unwrap()
-                    .0
-                    .to_vec(),
-            ],
+            tokens: vec![address_from_str(DAI_ADDRESS), address_from_str(DAI_ADDRESS)],
```
tvinagre, line 56: I like this

tycho-indexer/src/pb/tycho.evm.v1.rs
```diff
@@ -6,59 +6,59 @@
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct Block {
     /// The blocks hash.
-    #[prost(bytes = "vec", tag = "1")]
+    #[prost(bytes="vec", tag="1")]
```
tvinagre, line 9: Same as the other PR - do you know what happened that it wants to change the formatting? Please make sure it's consistent on the other files, if you haven't yet

tvinagre (approved):

proto/README.md
louise-poole: I think it's good to have here still too. It's likely you're working deep in the code when you realise you need to do these commands so it's nice to have it close to the code. Also these instructions are not easily outdated. 
When we move all repo readmes we can copy this info too (but I'd vote for having it duplicated).

tycho-indexer/src/pb/tycho.evm.v1.rs
louise-poole: This is the only proto messages generated file in this repo so there's nothing to be consistent with. However I have added a cooldown task to remove these files and rather import from protocol-sdk repo so that we avoid this duplication.

PR #493, opened 2025-01-23
feat: Add AccountBalance to core models

louise-poole:


tycho-core/src/models/blockchain.rs
```diff
@@ -170,25 +172,28 @@ impl BlockScoped for BlockAggregatedChanges {
 #[derive(Debug, Clone, PartialEq, Default)]
 pub struct TxWithChanges {
     pub protocol_components: HashMap<ComponentId, ProtocolComponent>,
-    pub account_deltas: HashMap<Bytes, AccountDelta>,
+    pub account_deltas: HashMap<Address, AccountDelta>,
     pub state_updates: HashMap<ComponentId, ProtocolComponentStateDelta>,
-    pub balance_changes: HashMap<ComponentId, HashMap<Bytes, ComponentBalance>>,
+    pub balance_changes: HashMap<ComponentId, HashMap<Address, ComponentBalance>>,
```
zizou0x, line 177: Maybe we can rename this to `component_balance_changes` now that we have this distinction?

tycho-indexer/src/extractor/models.rs
```diff
@@ -78,6 +80,7 @@ impl BlockScoped for BlockContractChanges {
 /// Hold the detailed state changes for a block alongside with protocol
 /// component changes.
 #[derive(Debug, PartialEq, Default, Clone)]
+#[deprecated(note = "Use BlockChanges instead")]
```
zizou0x, line 83: Does it make sense to mark it as deprecated here? It's more relevant for the producer no? (lib repo)

Also do we still have some Substreams that use this? if no we can completely remove it from both repo

zizou0x (commented):
Looks nice!

tycho-indexer/src/extractor/models.rs
louise-poole: I want to mark it as depreciated in case someone adds a new feature or gateway and tries to use this struct without reading the comment in the protobuf file that marks it as depreciated. Like while I was working on this PR I forgot it was depreciated and had to go down a rabbit hole to find out where it's used and why. Eventually found the protobuf comment that it's depreciated...

louise-poole: I agree - would be nice to move it entirely. I'm adding a cooldown task for it though as it's out of scope.

tycho-core/src/models/blockchain.rs
tvinagre: It would make sense, but then we might lose backwards compatibility, so I'm not sure it's worth the effort rn

tycho-indexer/src/extractor/models.rs
tvinagre: yeah let's do a cleanup when we have some time

tvinagre (approved):

PR #494, opened 2025-01-23
feat: Handle blocks with same ts

tvinagre:
Fixes and closes (https://github.com/propeller-heads/tycho-indexer/pull/283) 

Reason: EVM nodes gives us timestamps in seconds resolution. On sub-second blocks, we will have a conflict of multiple blocks with the same timestamp - causing bugs in the versioning system.
This solution appends n+1 microsecond for each block in a row (sequence size n) that is found with the same timestamp.

### DIFF
--- a/tycho-indexer/src/extractor/protocol_extractor.rs
+++ b/tycho-indexer/src/extractor/protocol_extractor.rs
@@ -25,7 +25,8 @@ use tycho_core::{
         ProtocolType, TxHash,
     },
     storage::{
-        ChainGateway, ContractStateGateway, ExtractionStateGateway, ProtocolGateway, StorageError,
+        BlockIdentifier, ChainGateway, ContractStateGateway, ExtractionStateGateway,
+        ProtocolGateway, StorageError,
     },
     traits::TokenPreProcessor,
     Bytes,
@@ -46,9 +47,13 @@ use crate::{
     pb::sf::substreams::rpc::v2::{BlockScopedData, BlockUndoSignal, ModulesProgress},
 };
 
+#[allow(dead_code)]
 pub struct Inner {
     cursor: Vec<u8>,
     last_processed_block: Option<Block>,
+    /// Keeps count of consecutive blocks with the same timestamp. This is then used to add some
+    /// microseconds in the blocks and maintain order.
+    same_ts_blocks_streak: u8,
     /// Used to give more informative logs
     last_report_ts: NaiveDateTime,
     last_report_block_number: u64,
@@ -102,6 +107,7 @@ where
                     inner: Arc::new(Mutex::new(Inner {
                         cursor: vec![],
                         last_processed_block: None,
+                        same_ts_blocks_streak: 0,
                         last_report_ts: chrono::Utc::now().naive_utc(),
                         last_report_block_number: 0,
                         first_message_processed: false,
@@ -111,7 +117,15 @@ where
                     reorg_buffer: Mutex::new(ReorgBuffer::new()),
                 }
             }
-            Ok(cursor) => {
+            Ok((cursor, block_id)) => {
+                let last_processed_block = gateway
+                    .get_block(block_id)
+                    .await
+                    .unwrap_or_else(|err| {
+                        error!(%chain, %err, "Found a cursor but was unable to retreive latest block");
+                        panic!("Unexpected error when fetching latest block {}", err);
+                    });
+
                 let cursor_hex = hex::encode(&cursor);
                 info!(
                     ?name,
@@ -126,8 +140,11 @@ where
                     chain_state,
                     inner: Arc::new(Mutex::new(Inner {
                         cursor,
-                        last_processed_block: None,
+                        last_processed_block: Some(last_processed_block),
                         last_report_ts: chrono::Local::now().naive_utc(),
+                        // TODO: What if the last block has the same timestamp as the previously
+                        // saved one?
+                        same_ts_blocks_streak: 0,
                         last_report_block_number: 0,
                         first_message_processed: false,
                     })),
@@ -1152,7 +1169,7 @@ pub struct ExtractorPgGateway {
 #[automock]
 #[async_trait]
 pub trait ExtractorGateway: Send + Sync {
-    async fn get_cursor(&self) -> Result<Vec<u8>, StorageError>;
+    async fn get_cursor(&self) -> Result<(Vec<u8>, i64), StorageError>;
 
     async fn ensure_protocol_types(&self, new_protocol_types: &[ProtocolType]);
 
@@ -1174,6 +1191,8 @@ pub trait ExtractorGateway: Send + Sync {
         &self,
         component_ids: &[&'a str],
     ) -> Result<HashMap<String, HashMap<Bytes, ComponentBalance>>, StorageError>;
+
+    async fn get_block(&self, block_number: i64) -> Result<Block, StorageError>;
 
     async fn get_account_balances(
         &self,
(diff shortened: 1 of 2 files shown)


tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -110,7 +111,15 @@ where
...
             }
-            Ok(cursor) => {
+            Ok((cursor, block_hash)) => {
+                let last_processed_block = gateway
+                    .get_block(block_hash)
+                    .await
+                    .unwrap_or_else(|err| {
+                        error!(%chain, %err, "Found a cursor but was unable to retreive latest block");
+                        panic!("Unexpected error when fetching latest block {}", err);
+                    });
```
louise-poole: Seems strange to both error and panic, no? I feel like the error is telling us the same info as the panic, but with more detail so let's make that the panic instead?
```suggestion
                    .unwrap_or_else(|err| {
                        panic!(%chain, %err, "Found a cursor but was unable to retrieve latest block");
                    });
```

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -201,7 +210,8 @@ where
         }
     }
 
-    #[instrument(skip_all, fields(chain = % self.chain, name = % self.name, block_number = % msg.block.number))]
+    #[instrument(skip_all, fields(chain = % self.chain, name = % self.name, block_number = % msg.block.number
+    ))]
```
louise-poole: This formatting looks quite odd to me?

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -576,6 +586,21 @@ where
...
+                // Blockchains with fast block times (e.g., Arbitrum) may produce blocks with
+                // identical timestamps. To ensure accurate ordering, we adjust each
+                // block's timestamp by adding a microsecond offset based on the
+                // number of blocks with the same timestamp encountered so far.
+                msg.block.ts += chrono::Duration::microseconds(
+                    last_processed_block
+                        .ts
+                        .timestamp_subsec_micros() as i64 +
+                        1,
+                );
```
louise-poole: This could be simplified a lot no?
```suggestion
                msg.block.ts = last_processed_block.ts + Duration::microseconds(1);
```

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -576,6 +586,21 @@ where
         let mut msg =
             if let Some(post_process_f) = self.post_processor { post_process_f(msg) } else { msg };
 
+        if let Some(last_processed_block) = self.get_last_processed_block().await {
+            if msg.block.ts.timestamp() == last_processed_block.ts.timestamp() {
+                // Blockchains with fast block times (e.g., Arbitrum) may produce blocks with
+                // identical timestamps. To ensure accurate ordering, we adjust each
+                // block's timestamp by adding a microsecond offset based on the
+                // number of blocks with the same timestamp encountered so far.
```
louise-poole: Can we emphasise in the comment that the comparison above is comparing seconds only? Which is why we can add microseconds to last_processed_block and still expect it to equal a future block with no microseconds yet.

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -1075,19 +1106,28 @@ impl ExtractorPgGateway {
...
+        self.state_gateway
+            .get_block(&BlockIdentifier::Hash(block_hash))
+            .await
+    }
+    async fn get_cursor(&self) -> Result<(Vec<u8>, Bytes), StorageError> {
+        let extraction_state = self.get_last_extraction_state().await;
+        match extraction_state {
+            Ok(state) => Ok((state.cursor, state.block_hash)),
+            Err(e) => Err(e),
+        }
```
louise-poole, line 1264: I don't really like that we have to go to the db twice here.... first to get the cursor with the block hash, then the get the block, given that hash. If fetching this is always coupled why not just straight up change `get_cursor` to `get_extraction_state` and make that return (cursor, block). Then on the PgGateway we just need to update the `get_state` method accordingly.

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -1075,19 +1106,28 @@ impl ExtractorPgGateway {
...
 #[async_trait]
 impl ExtractorGateway for ExtractorPgGateway {
-    async fn get_cursor(&self) -> Result<Vec<u8>, StorageError> {
-        self.get_last_cursor().await
+    async fn get_block(&self, block_hash: Bytes) -> Result<Block, StorageError> {
+        self.state_gateway
+            .get_block(&BlockIdentifier::Hash(block_hash))
+            .await
+    }
+    async fn get_cursor(&self) -> Result<(Vec<u8>, Bytes), StorageError> {
```
zizou0x, line 1259: Please add docs

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -576,6 +586,21 @@ where
         let mut msg =
             if let Some(post_process_f) = self.post_processor { post_process_f(msg) } else { msg };
 
+        if let Some(last_processed_block) = self.get_last_processed_block().await {
+            if msg.block.ts.timestamp() == last_processed_block.ts.timestamp() {
+                // Blockchains with fast block times (e.g., Arbitrum) may produce blocks with
+                // identical timestamps. To ensure accurate ordering, we adjust each
+                // block's timestamp by adding a microsecond offset based on the
+                // number of blocks with the same timestamp encountered so far.
+                msg.block.ts += chrono::Duration::microseconds(
```
zizou0x: Isn't it `=` instead of `+=`?

Edit: okay I think I get it, we use `.timestamp_subsec_micros()` so that looks correct

louise-poole (commented):
Logic looks solid 👍  Just have a few minor implementation detail comments

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -576,6 +586,21 @@ where
         let mut msg =
             if let Some(post_process_f) = self.post_processor { post_process_f(msg) } else { msg };
 
+        if let Some(last_processed_block) = self.get_last_processed_block().await {
+            if msg.block.ts.timestamp() == last_processed_block.ts.timestamp() {
```
zizou0x, line 684: Is this buggy? 

What if we receive 3 blocks with a ts of 1000? 

First will be false here (999 == 1000)
Second will be true (1000 == 1000) therefore we will add one to the ts
Third with be false (1000 == 1001) so we will have 1 and 3 with the same ts

Or am I missing something?

tycho-indexer/src/extractor/protocol_extractor.rs
zizou0x: Why?

tycho-indexer/src/extractor/protocol_extractor.rs
louise-poole: No - here he's fetching the microseconds of the current latest block, adding 1 and then adding it to msg.block.ts. I left a comment for a more readable way to do this though.

tycho-indexer/src/extractor/protocol_extractor.rs
zizou0x: I also find this more natural

tycho-indexer/src/extractor/protocol_extractor.rs
louise-poole: msg.block.ts.timestamp() returns the block time in seconds. So when we alter the block time by adding 1 microsecond, it does not affect this evaluation. 
First will be false here (999 == 1000)
Second will be true (1000 == 1000) therefore we will add one microsecond to the ts (1000.000001)
Third with be true (1000 == 1000) as we're comparing seconds. Then will add another microsecond (1000.000002)

louise-poole: (I already left a comment asking to make the fact that this compares seconds clearer)

zizou0x: True, I missed that "msg.block.ts.timestamp() returns the block time in seconds"

tycho-indexer/src/extractor/protocol_extractor.rs
louise-poole: Another question here though... why microseconds instead of milliseconds? What granularity do we store in the db? I thought it was only to milliseconds?

tycho-indexer/src/extractor/protocol_extractor.rs
tvinagre: I have no idea, since we're using `stable` for Rust toolchain, I'm assuming maybe cargo fmt rules have changed? 
This was done by cargo fmt

tycho-indexer/src/extractor/protocol_extractor.rs
tvinagre: oops - I copied this from the other PR and overlooked this line - thanks!

tycho-indexer/src/extractor/protocol_extractor.rs
tvinagre: ohh - nice one, tnx!

tycho-indexer/src/extractor/protocol_extractor.rs
tvinagre: This is done only once, on startup so I don't see this as a big problem. But, since this method is only used here and in other tests, I'll update it to do the join in the db

louise-poole (approved):
Looks great!

PR #495, opened 2025-01-24
feat: add migration for token_id in account_balance table

louise-poole:
Adds the following columns to the account_balance table:
- token_id (defaults existing entries to the native token id)
- balance_float (calculates this from the balance bytes value)

Tested to correctly handle the following scenarios:
- if ETH already exists: it will not insert a new account and token for ETH, but instead use the existing one to link to in the account_balance table. On the down migration it will not remove the ETH token (as it's referenced by a table other than account_balance)
- if ETH does not exist, but the DB is already populated for either ethereum or base: it will insert a zero address account and a new ETH token linked to it. On the down migration this account and token will be removed.
- if the DB is empty: no native token is added. It is expected the code will do this when a new chain is added.

Assumptions: 
- We only have Ethereum and Base DBs populated so can assume the native token is ETH in these migrations.
- Only one chain exists per DB migrated.


tycho-storage/migrations/2025-01-23_account_balances/up.sql
```diff
@@ -0,0 +1,63 @@
...
+-- Assumes each db only has 1 chain
+INSERT INTO account (chain_id, title, address)
+VALUES (
+    (SELECT id FROM chain LIMIT 1),
+    'ETH_0x0000000000000000000000000000000000000000',
+    '\x0000000000000000000000000000000000000000'
+)
+ON CONFLICT (chain_id, address) DO NOTHING;
+
+-- Step 2: add ETH native token (currently only ethereum and base chains are used in our dbs
```
tvinagre: I think this belongs on the code, not in the migration. Somewhere on the startup.
Reason is that, if we're on a chain where ETH is an ERC20, or it doesn't exist at all, we're still hardcoding this in the db - it's a wrong assumption if we're designing Tycho to be chain-agnostic

louise-poole: Hm... okay, but then how do we migrate existing DBs? This column needs to be NOT NULL so I need to populate it with something upon migration.

louise-poole: I should anyway skip the INSERT steps if the chain table is empty (there is no data in the DB yet).

louise-poole: What I have added so that we can migrate existing DBs (ethereum and base) and also not mistakenly add ETH to new DBs (when we batch run all migrations on an empty/new DB) is to add a chain condition check before the INSERTs.

tycho-storage/migrations/2025-01-23_account_balances/up.sql
```diff
@@ -0,0 +1,80 @@
...
+
+
+-- ADD balance_float to account_balance table
+
+-- Step 1: Add balance_float column
+ALTER TABLE account_balance ADD COLUMN balance_float FLOAT8;
+
+-- Step 2: Update the new column by converting 'balance' bytea to float8
+UPDATE account_balance
+SET balance_float = ('x' || encode(balance, 'hex'))::bit(64)::bigint::float8;
```
kayibal: Two questions:

1. Why do we need balance_float, we use this for tvl queries mainly not sure we need this here right away... but I see how it could be eventually useful.
2. Are you sure encode works for all possible cases? Where is this function specified? I don't think that uint256 encoding is the same as float8 encoding especially because this operation is lossy.

kayibal (changes requested):

tycho-storage/migrations/2025-01-23_account_balances/up.sql
louise-poole: You're right... we do not need balance_float. I tested this on 10 values I had in my local DB and all converted exactly... but you are right, if the sample size was larger we might encounter variations due to the operation being lossy. I think I'll remove balance_float for now and we can always add it when we feel it is actually needed. Don't want to waste mission time on this now.

tvinagre (approved):

tycho-storage/migrations/2025-01-23_account_balances/up.sql
```diff
@@ -0,0 +1,76 @@
...
+        FROM chain 
+        WHERE name IN ('ethereum', 'base')
+    ) = 1 THEN
+        INSERT INTO token (account_id, symbol, decimals, tax, gas, quality)
+        VALUES (
+            (SELECT id FROM account WHERE address = '\x0000000000000000000000000000000000000000'),
+            'ETH',
+            18,
+            0,
+            ARRAY[0]::BIGINT[],
```
kayibal: Is this the transfer gas? Can't we set a proper value here like `[2300] * 4`

kayibal (approved):
Smol neg, other than that LGTM! Thanks!

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -2680,25 +2837,50 @@ mod test_serial_db {
...
+                        (Bytes::from_str(WETH_ADDRESS).unwrap(), ComponentBalance {
+                            token: Bytes::from_str(WETH_ADDRESS).unwrap(),
                             balance: Bytes::from("0x00000001"),
                             balance_float: 1.0,
                             modify_tx: Bytes::from_str("0x0000000000000000000000000000000000000000000000000000000000000000").unwrap(),
                             component_id: "pc_1".to_string(),
                         }),
                     ])),
                 ]),
+                account_balances: HashMap::from([
```
tvinagre, line 2856: Can we also add a test that verifies that our "merge" is correct?
I'm worried about the case where we get 2 AccountUpdates for the same account on different transactions.

PR #497, opened 2025-01-29
feat: add AccountBalance to postgres gateway

louise-poole:
I recommend review this commit-by-commit.

This PR covers:
- removes balance_float from AccountBalance models (was decided [here](https://github.com/propeller-heads/tycho-indexer/pull/495#discussion_r1929032199) it wasn't necessary right now)
- adds hardcoded mapping of chain->native token
- inserts native token on new chain inserts
- adds native token DB id cache to gateway
- updates existing postgres gateway methods to include token_id field on AccountBalance inserts
- updates existing postgres gateway methods to filter for native token on account balance queries for native balance
- adds a postgres gateway method for inserting account balances
- adds a postgres gateway method for fetching account balances
- updates the extractor to handle account balances on forward and revert changes


tycho-core/src/models/mod.rs
```diff
@@ -74,6 +76,42 @@ impl From<dto::Chain> for Chain {
     }
 }
 
+fn native_eth(chain: Chain) -> CurrencyToken {
+    CurrencyToken::new(
+        &Bytes::from_str("0x0000000000000000000000000000000000000000").unwrap(),
+        "ETH",
+        18,
+        0,
+        &[None],
```
kayibal: Gas unset?

tycho-storage/src/postgres/mod.rs
```diff
@@ -504,16 +517,65 @@ impl PostgresGateway {
...
             retention_horizon,
         );
 
         Ok(gw)
     }
+
+    // Could not use FromConnection trait as it is already implemented for a Chain->id map type.
+    // Also this custom 'from connection' fn signature allows us to reuse the already fetched
+    // chain cache.
+    async fn native_cache_from_connection(
```
kayibal: weird naming... shouldn't it be `native_token_cache...`

tycho-storage/src/postgres/contract.rs
```diff
@@ -1518,6 +1518,132 @@ impl PostgresGateway {
...
     }
+
+    pub async fn get_account_balances(
+        &self,
+        chain: &Chain,
+        accounts: Option<&[Address]>,
+        at: Option<&Version>,
+        conn: &mut AsyncPgConnection,
+    ) -> Result<HashMap<Address, HashMap<Address, AccountBalance>>, StorageError> {
+        // NOTE: the returned AccountBalances have a default value for tx_hash as it is assumed
```
kayibal, line 1526: Why did you decide to do this? Speed? Maybe add this here

tycho-storage/src/postgres/contract.rs
```diff
@@ -1518,6 +1518,132 @@ impl PostgresGateway {
...
+                schema::account_balance::account_id,
+                schema::account_balance::token_id,
+                schema::account_balance::balance,
+            ))
+            .order(schema::account_balance::account_id.asc())
+            .get_results::<(i64, i64, Balance)>(conn)
+            .await
+            .map_err(PostgresError::from)?
+            .into_iter()
+            .group_by(|e| e.0)
```
kayibal, line 1577: Uh not sure we should use a groupby here. Is this really necessary? This query looks super expensive... Honestly this looks a bit off we should only need to filter by `valid_from < ts < valid_to`. Why is this not enough?

tycho-storage/src/postgres/contract.rs
```diff
@@ -786,14 +769,16 @@ impl PostgresGateway {
...
             account_orm.address,
             account_orm.title,
             HashMap::new(),
-            balance_orm.balance,
+            native_balance.balance,
+            account_balances.clone(),
             code_orm.code,
             code_orm.hash,
-            balance_tx,
+            // TODO: remove balance_modify_tx from Account
```
kayibal, line 780: Good point ...

tycho-storage/src/postgres/contract.rs
louise-poole: Yes - it's the slowest part of the query and the value is never used after being returned here. We omit the modify tx for component_balances too.

tycho-storage/src/postgres/contract.rs
louise-poole: I followed the same steps as we do for fetching component balances. Not 100% sure why we went with doing this, but I'll play around and see if it's necessary.

kayibal (approved):

tycho-core/src/models/mod.rs
```diff
@@ -74,6 +76,33 @@ impl From<dto::Chain> for Chain {
...
+}
+
+impl Chain {
+    /// Returns the native token symbol for the chain.
+    pub fn native_token(&self) -> CurrencyToken {
+        match self {
+            Chain::Ethereum => native_eth(Chain::Ethereum),
+            // It was decided that STRK token will be tracked as a dedicated AccountBalance on
+            // Starknet accounts and ETH balances will be tracked as a native balance.
+            Chain::Starknet => native_eth(Chain::Ethereum),
```
tvinagre:
```suggestion
            Chain::Starknet => native_eth(Chain::Starknet),
```
?

tvinagre (approved):

tycho-core/src/models/mod.rs
louise-poole: Was just testing you 👀 hehe

PR #504, opened 2025-02-05
fix: improve efficiency of activity filter on tokens query

louise-poole:
Performance improvement: using EXISTS instead of `eq_any` shows significant performance improvements during tests.

### DIFF
--- a/tycho-storage/src/postgres/protocol.rs
+++ b/tycho-storage/src/postgres/protocol.rs
@@ -934,12 +934,16 @@ impl PostgresGateway {
         }
 
         if let Some(last_traded_ts_threshold) = last_traded_ts_threshold {
-            let active_tokens_subquery = schema::component_balance_default::table
-                .select(schema::component_balance_default::token_id)
-                .filter(schema::component_balance_default::valid_from.gt(last_traded_ts_threshold))
-                .distinct();
-            query = query.filter(schema::token::id.eq_any(active_tokens_subquery));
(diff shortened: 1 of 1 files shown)


zizou0x (approved):
lgtm

tycho-storage/src/postgres/protocol.rs
```diff
@@ -934,12 +934,16 @@ impl PostgresGateway {
...
-            count_query = count_query.filter(schema::token::id.eq_any(active_tokens_subquery));
+            let active_tokens_exists = diesel::dsl::exists(
+                schema::component_balance_default::table
+                    .filter(
+                        schema::component_balance_default::valid_from.gt(last_traded_ts_threshold),
+                    )
+                    .filter(schema::component_balance_default::token_id.eq(schema::token::id)),
+            );
+
+            query = query.filter(active_tokens_exists);
```
kayibal, line 945: How does is this able to correlate this via the id column?

louise-poole: If you have a look at the `active_tokens_exists` subquery you'll see I use `schema::token::id` in one of the filters. Since the token table isn't joined in the subquery itself (only component_balance_default is introduced in the subquery), it then uses the referenced column from the outer query instead. This is how I pass the token id column from the outer to the inner.
Not very intuitive, I know. But it's how diesel's `exists` handles this and it's working as expected in the test.

PR #506, opened 2025-02-07
feat: Update Tycho Python client Account DTO

tvinagre:


zizou0x (approved):

tycho-client-py/python/tycho_indexer_client/dto.py
```diff
@@ -183,6 +183,7 @@ class ResponseAccount(BaseModel):
     title: str
     slots: Dict[HexBytes, HexBytes]
     native_balance: HexBytes = Field(alias="balance")
+    token_balances: Dict[HexBytes, HexBytes]
```
louise-poole:
```suggestion
    token_balances: Dict[HexBytes, Dict[HexBytes, HexBytes]]
```

louise-poole: We also need to update BlockChanges too

PR #507, opened 2025-02-07
feat(indexer): pull Substreams spkgs from Substreams registry

zizou0x:
Context: S3 is very specific to our infra, this PR aims to make pulling our Substreams module permisionless. We decided to use Substreams registry for this.

TODO: publish our modules to the registry

### DIFF
--- a/tycho-indexer/src/extractor/runner.rs
+++ b/tycho-indexer/src/extractor/runner.rs
@@ -1,11 +1,13 @@
-use std::{collections::HashMap, env, path::Path, sync::Arc};
+use std::{collections::HashMap, env, ffi::OsStr, fs, path::Path, sync::Arc};
 
 use anyhow::{format_err, Context, Result};
 use async_trait::async_trait;
 use aws_config::meta::region::RegionProviderChain;
-use aws_sdk_s3::Client;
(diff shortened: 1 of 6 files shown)


tycho-indexer/src/extractor/runner.rs
```diff
@@ -413,26 +422,67 @@ impl ExtractorBuilder {
...
+            let s3_bucket = self.s3_bucket.as_ref().ok_or_else(|| {
                 ExtractionError::Setup(format!(
-                    "Failed to download {} from s3. {}",
-                    &self.config.spkg, e
+                    "Missing SPKG file and no S3 bucket configured for {}",
+                    &self.config.spkg
                 ))
             })?;
+
+            download_file_from_s3(s3_bucket, &self.config.spkg, spkg_path)
```
zizou0x, line 476: Should we completely remove this logic?

Currently it's used as a fallback if the call to the registry fails

louise-poole: I'd keep it until we've got all existing spkgs up on the substreams registry and confirmed they're loading correctly.

louise-poole (approved):
Nice!!!! LGTM!

PR #509, opened 2025-02-07
feat(client): log a list of available protocols on startup

zizou0x:


tycho-client/src/cli.rs
```diff
@@ -197,13 +201,34 @@ async fn run(exchanges: Vec<(String, Option<String>)>, args: CliArgs) {
...
+        .map(|(name, _)| name.clone())
+        .collect::<HashSet<_>>();
+
+    let not_requested_protocols = available_protocols_set
+        .difference(&requested_protocol_set)
+        .cloned()
+        .collect::<Vec<_>>();
+
+    if !not_requested_protocols.is_empty() {
+        tracing::info!("Other available protocols: {}", not_requested_protocols.join(", "));
```
zizou0x, line 226: Maybe we can just log everything even the requested?

louise-poole: Nah, I like it filtered like this and it makes it very clear what you missed. When we start supporting 20+ protocols having to manually filter the list to see what you haven't used yet will not be fun and defeat the purpose of this log.

louise-poole (commented):
Nice! Now we need to do the same on the rust interface too.

tycho-client/src/stream.rs
```diff
@@ -135,6 +140,7 @@ impl TychoStreamBuilder {
 
         // Initialize the WebSocket client
         let ws_client = WsDeltasClient::new(&tycho_ws_url, auth_key.as_deref()).unwrap();
+        let rpc_client = HttpRPCClient::new(&tycho_rpc_url, auth_key.as_deref()).unwrap();
```
louise-poole, line 143: You can probably pass a clone of this to ProtocolStateSynchronizer?

louise-poole (approved):

PR #511, opened 2025-02-12
fix(rpc): correctly apply TVL filtering on `/protocol_component` requests

zizou0x:
This fixes a bug that was occuring when someone was querying components with a TVL filter. In that case, any component in the buffer would be returned, even if it wasn't passing the tvl threshold requested.


louise-poole (approved):
LGTM!

tvinagre (approved):

PR #513, opened 2025-02-14
refactor(indexer): better error handling in `maybe_report_progress`

zizou0x:
Currently it panics if the conversion to minute fails. This commit makes it just skip the report and log a warning instead.

### DIFF
--- a/tycho-indexer/src/extractor/protocol_extractor.rs
+++ b/tycho-indexer/src/extractor/protocol_extractor.rs
@@ -177,28 +177,43 @@ where
             let distance_to_current = current_block - block.number;
             let blocks_processed = block.number - state.last_report_block_number;
             let blocks_per_minute = blocks_processed as f64 * 60.0 / time_passed as f64;
-            let time_remaining =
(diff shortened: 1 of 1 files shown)


tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -177,28 +177,35 @@ where
...
+                    "chain" => extractor_id.chain.to_string(),
+                    "extractor" => extractor_id.name.to_string(),
+                )
+                .set(time_remaining.num_minutes() as f64);
+            } else {
+                warn!(
+                    "Failed to convert {} to a duration",
+                    (distance_to_current as f64 / blocks_per_minute) as i64,
+                )
+            }
```
louise-poole: Can we still log all the other values? Like still have an info log, but without the 'time remaining' data.

zizou0x: Yes would be nicer indeed

louise-poole (approved):

PR #515, opened 2025-02-18
fix: propagate more meaningful RPC errors on the client

louise-poole:
Currently when an RPC call within the tycho-client fails, the client panics with:
`Fatal error: No response received` 
and the error that is logged is:
`ParseResponse("expected value at line 1 column 1")`.

This doesn't give us much info on what the problem was. With the changes in this PR we instead propagate a more informative error message. The client will now panic with (as an example):
`Failed to parse response: Unknown extractor: uniswap_v5` 
and log the error:
`ParseResponse("Unknown extractor: uniswap_v5")`.

### DIFF
--- a/tycho-client/src/rpc.rs
+++ b/tycho-client/src/rpc.rs
@@ -423,8 +423,8 @@ impl RPCClient for HttpRPCClient {
         request: &StateRequestBody,
     ) -> Result<StateRequestResponse, RPCError> {
         // Check if contract ids are specified
-        if request.contract_ids.is_none()
-            || request
+        if request.contract_ids.is_none() ||
+            request
                 .contract_ids
                 .as_ref()
                 .unwrap()
@@ -521,8 +521,8 @@ impl RPCClient for HttpRPCClient {
         request: &ProtocolStateRequestBody,
     ) -> Result<ProtocolStateRequestResponse, RPCError> {
         // Check if contract ids are specified
-        if request.protocol_ids.is_none()
-            || request
+        if request.protocol_ids.is_none() ||
+            request
                 .protocol_ids
                 .as_ref()
                 .unwrap()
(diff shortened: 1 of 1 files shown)


tvinagre (commented):
Is there a more elegant way of doing this? Maybe decoding the errors that come from the RPC. It's usually not a good practice to just propagate the errors without filtering them - this could lead to secrets exposure, or logic exposure (which we don't care much in this case as it's OS)

tycho-client/src/rpc.rs
```diff
@@ -468,9 +468,9 @@ impl RPCClient for HttpRPCClient {
             });
         }
 
-        let accounts = serde_json::from_str::<StateRequestResponse>(&body).map_err(|e| {
-            error!("Failed to parse contract state response {:?}", &body);
-            RPCError::ParseResponse(e.to_string())
+        let accounts = serde_json::from_str::<StateRequestResponse>(&body).map_err(|_| {
+            error!("Failed to parse contract state response: {:?}", &body);
+            RPCError::ParseResponse(body)
```
zizou0x: From your experience, would it be relevant to also have the `e` here? `RPCError::ParseResponse(e,body)`

louise-poole: So far `e` is always the same: `"expected value at line 1 column 1"`. This is because when an error occurs we get a string in the `body` and not a structured json. If `e` were to be anything else here (as in it's decoding a json but getting unexpected keys or something) then we have fundamentally broken our RPC types and will notice the issue in tests and on CI before it ever reaches the point that the user should see this.

zizou0x: I see, I'm still worried about if one day we introduce a breaking change on the RPC but someone still tries to use a old client. In that case this person would get a parse error with a very unclear message no? It would just be the body basically

louise-poole: True true. For now we've found it easy enough to have all changes to the RPC structs be backwards compatible. But rather safe than sorry I guess 👍 I'll add the error message here

tvinagre (approved):

PR #516, opened 2025-02-24
feat: update postgres gateway get_tokens quality filter to be ranged

louise-poole:
ISSUE: Currently our postgres gateway’s get_tokens fn optionally filters tokens by a min quality. For the most part this is all we need as all client interactions are concerned about highest quality/most relevant tokens only. However, our token analysis cronjob is interested in tokens with quality 6-10 only. A min quality filter alone is not sufficient here and results in the cronjob needlessly fetching excessive numbers of tokens from the DB only to skip analysing them.

### DIFF
--- a/tycho-core/src/models/protocol.rs
+++ b/tycho-core/src/models/protocol.rs
@@ -199,7 +199,7 @@ impl ComponentBalance {
 /// Token quality range filter
 ///
 /// The quality range is considered inclusive and used as a filter, will be applied as such.
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Clone)]
 pub struct QualityRange {
     pub min: Option<i32>,
     pub max: Option<i32>,
@@ -212,6 +212,11 @@ impl QualityRange {
 
     pub fn min_only(min: i32) -> Self {
         Self { min: Some(min), max: None }
+    }
+
+    #[allow(non_snake_case)]
+    pub fn None() -> Self {
+        Self { min: None, max: None }
     }
 }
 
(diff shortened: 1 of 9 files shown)


tycho-indexer/src/extractor/protocol_cache.rs
```diff
@@ -82,7 +86,7 @@ impl ProtocolMemoryCache {
         {
             let mut cached_tokens = self.tokens.write().await;
             self.gateway
-                .get_tokens(self.chain, None, None, None, None)
+                .get_tokens(self.chain, None, QualityRange::default(), None, None)
```
zizou0x: I think using default for "no filter" is not 100% obvious from the interface. Maybe we shouldn't derive the struct with default and add a none function to make it clearer?

`QualityRange::none()`

zizou0x (commented):
Looks good, just a small comment about the `QualityRange` interface

tycho-core/src/models/protocol.rs
```diff
@@ -196,6 +196,29 @@ impl ComponentBalance {
...
+impl QualityRange {
+    pub fn new(min: i32, max: i32) -> Self {
+        Self { min: Some(min), max: Some(max) }
+    }
+
+    pub fn min_only(min: i32) -> Self {
+        Self { min: Some(min), max: None }
+    }
+
+    pub fn none() -> Self {
```
zizou0x: Tiny detail but I think it would be cleaner with a caps "N"

zizou0x (approved):
Looks good to me! 👍 

I just added a tiny comment but feel free to merge without addressing if you think it's not necessary

tvinagre (approved):
LGTM, just +1 on @zizou0x's suggestion

PR #523, opened 2025-03-05
refactor: Import tycho protobuf messages from protocol sdk

louise-poole:
Currently the protobuf message definitions are duplicated between tycho repos. This PR make the tycho-protocol-sdk repo the main source of truth for our custom tycho protobuf message definitions and we now import those message structs from there.


tycho-indexer/src/extractor/protobuf_deserialisation.rs
```diff
@@ -607,18 +608,1834 @@ impl TryFromMessage for BlockChanges {
     }
 }
 
+#[cfg(test)]
+pub mod fixtures {
```
zizou0x: Are these used? I don't understand this change in the context of the PR

tycho-indexer/Cargo.toml
```diff
@@ -64,6 +64,7 @@ mini-moka = "0.10.3"
 num-bigint = "0.4.4"
 num-traits = "0.2.19"
 num_cpus = "1.16.0"
+tycho-substreams = { git = "https://github.com/propeller-heads/tycho-protocol-sdk.git", rev = "04d14db" }
```
zizou0x: I think it would be nicer to create a tag but not mandatory

tycho-indexer/src/extractor/protobuf_deserialisation.rs
louise-poole: These fixtures used to be defined inside the `pb` module and then imported here in `protobuf_deserialisation` as test assets to deserialise. If we're removing the `pb` module then it made sense to move the fixtures here to the file that uses them/deserialises them.

tycho-indexer/Cargo.toml
louise-poole: Agreed 👍  Let me actually do that RN. Any suggestions on what to call it? Maybe just `0.1.0`?

tycho-indexer/src/extractor/protobuf_deserialisation.rs
zizou0x: I see, these were in the `testing.rs` file? 

I think I would prefer if we kept them in a separate file, even if they are only used here, I think 1k8 lines of fixtures clutter this file too much

tycho-indexer/Cargo.toml
zizou0x: Yes if we don't have any tag yet I think that's good

zizou0x (approved):

PR #525, opened 2025-03-06
feat: update substreams proto messages

louise-poole:
And clean up dependencies not used now that protocol substream packages are developed in a separate repo.
Also clean up workspace dependencies.


tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -2206,7 +2206,10 @@ mod test_serial_db {
     const VM_TX_HASH_1: &str = "0x0d9e0da36cf9f305a189965b248fc79c923619801e8ab5ef158d4fd528a291ad";
 
     // Ambient Contract
-    const VM_CONTRACT: [u8; 20] = hex_literal::hex!("aaaaaaaaa24eeeb8d57d431224f73832bc34f688");
+    const VM_CONTRACT: [u8; 20] = [
```
zizou0x, line 2207: Wasn't it better with the macro?

zizou0x (approved):

tycho-indexer/src/extractor/protocol_extractor.rs
louise-poole: This was the only place we used the `hex_literal` crate... felt weird to have a whole project dependency just for this one usecase. I can put it back if we want - I don't feel strongly about this.

zizou0x: Hm okay if it removes a dependency then it's better

PR #530, opened 2025-03-19
fix(rpc-client): log error with response on failed response parse

louise-poole:


tycho-client/src/rpc.rs
```diff
@@ -468,6 +468,7 @@ impl RPCClient for HttpRPCClient {
         }
 
         let accounts = serde_json::from_str::<StateRequestResponse>(&body).map_err(|err| {
+            error!("Error: {}", err);
```
zizou0x, line 471: Why not add it in the error log below?

louise-poole: We error here in 2 scenarios:
1) the RPC returns an error: here the error itself will just be 'ParseError: unexpected token at line 1 column 1' and the body gives us info such as 'Unrecognised protocol system'.
2) We actually failed to decode the body. Here the error would be interesting: 'missing field account_balance', but the body could be a super long response body full of contract data etc. In this case, having them reported on the same line makes it hard to see what the error is so I opted to log them on separate lines.

PR #531, opened 2025-03-19
feat: Exit gracefully on substreams stream end

tvinagre:


tycho-indexer/src/main.rs
```diff
@@ -290,7 +290,28 @@ async fn run_spkg(global_args: GlobalArgs, run_args: RunSpkgArgs) -> Result<(),
...
-    res.expect("Extractor- nor ServiceTasks should panic!")
+    match res {
+        Ok(_) => Ok(()),
+        Err(e) => {
+            let err = e.try_into_panic();
+            match err {
+                Ok(err) => {
+                    if let Ok(extraction_err) = err.downcast::<ExtractionError>() {
+                        if let ExtractionError::SubstreamsError(msg) = *extraction_err {
+                            if msg.contains("stream ended") {
```
kayibal, line 301: Maybe introduce a new error type for this instead?

PR #536, opened 2025-03-26
feat: Rename tycho-core to tycho-common.

kayibal:
This change is necessary to push to creates.io. There is a conflict with another crate.


tvinagre (approved):

zizou0x (approved):
LGTM bravo!

PR #539, opened 2025-03-26
feat: Handle delayed exchanges on tycho-client

louise-poole:


tycho-client/src/feed/mod.rs
```diff
@@ -75,20 +75,20 @@ type BlockSyncResult<T> = anyhow::Result<T>;
 /// ## Synchronization Logic
 ///
 /// To classify a synchronizer as delayed or advanced, we need to first define the current
-/// block. Currently, we simply expect all synchronizer to be at the same block at startup. The
+/// block. Currently, we simply expect all synchronizers to be at the same block at startup. The
```
kayibal: This comment seems outdated

tycho-client/src/feed/mod.rs
```diff
@@ -212,34 +212,51 @@ impl SynchronizerStream {
...
+            {
+                Ok(Some(msg)) => {
+                    debug!(%extractor_id, block_num=?msg.header.number, "Received new message during catch-up");
+                    let block_pos = block_history
+                        .determine_block_position(&msg.header)
+                        .unwrap();
+                    results.push(msg);
+                    if matches!(block_pos, BlockPosition::NextExpected) {
+                        break;
+                    }
```
kayibal, line 251: Shouldn't we error in this case? I mean something must be really wrong if we got here.

tycho-client/src/feed/mod.rs
```diff
@@ -269,11 +287,10 @@ impl SynchronizerStream {
...
             BlockPosition::Latest | BlockPosition::Delayed => {
                 let now = Local::now().naive_utc();
-                if self
+                let wait_duration = self
                     .modify_ts
-                    .signed_duration_since(now) >
-                    chrono::Duration::seconds(60)
-                {
+                    .signed_duration_since(now);
+                if wait_duration > chrono::Duration::seconds(60) {
```
kayibal: Should we keep these 60s hardcoded?

tycho-client/src/feed/mod.rs
```diff
@@ -252,14 +269,15 @@ impl SynchronizerStream {
         }
     }
 
-    /// Logic for transition a state synchronizer.
+    /// Logic to transition a state synchronizer.
     ///
-    /// Given a newly received block from a state synchronizer, this method transitions it's state
+    /// Given a newly received block from a state synchronizer, this method transitions its state
```
kayibal: Can we summarize the strategy employed below here?

tycho-client/src/feed/mod.rs
```diff
@@ -376,55 +393,76 @@ where
...
                 }
-            });
+                Ok(None) => {
+                    warn!(%extractor_id, "Dead synchronizer at startup will be purged!");
+                    synchronizer.state = SynchronizerState::Ended;
+                    synchronizer.modify_ts = Local::now().naive_utc();
+                    None
+                }
+                Err(_) => {
+                    warn!(%extractor_id, "Stale synchronizer at startup will be purged!");
```
kayibal: So the error here is if we timed out waiting for a message, right? Maybe this can be made a bit clearer in the warning message?

tycho-client/src/feed/mod.rs
```diff
@@ -376,55 +393,76 @@ where
...
+            n_healthy=?ready_sync_msgs.len(),
+            "Block synchronization started successfully!"
+        );
 
-        // Purge any stale synchronizers, require rest to be ready on the same block, else fail.
-        // It's probably worth doing more complex things here if this is problematic e.g. setting
-        // some as delayed and waiting for those to catch up, but I want to go with the simplest
-        // solution first.
+        // Purge any stale synchronizers
         sync_streams.retain(|_, v| matches!(v.state, SynchronizerState::Ready(_)));
```
kayibal: Are we only purging stale ones with this? Aren't we also purging delayed ones here? Ah we set all to ready on the first iteration including delayed ones right? Maybe worth a little comment. 

EDIT: Or remove this line and merge the logic with the next block.

tycho-client/src/feed/mod.rs
```diff
@@ -376,55 +393,76 @@ where
...
-        let start_header = if let Some(first) = ready_sync_msgs.values().next() {
-            if !ready_sync_msgs
-                .values()
-                .all(|v| first.header == v.header)
-            {
-                anyhow::bail!("not all synchronizers on same block!")
+
+        // Determine correct state for each remaining synchronizer, based on their header vs the latest one
+        for (_, stream) in sync_streams.iter_mut() {
+            if let SynchronizerState::Ready(header) = &stream.state.clone() {
```
kayibal, line 477: Is this check necessary? You purged all the other ones previously

tycho-client/src/feed/synchronizer.rs
```diff
@@ -341,15 +341,11 @@ where
             .await?;
 
         info!("Waiting for deltas...");
-        // we need to wait 2 messages because of cache gateways insertion delay.
-        let first_msg = timeout(Duration::from_secs(360), msg_rx.recv())
+        // wait for first deltas message
+        let mut first_msg = timeout(Duration::from_secs(360), msg_rx.recv())
```
kayibal: Should we keep this 360s hardcoded value?

tycho-client/src/feed/mod.rs
louise-poole: WDYM? This statement breaks the while loop once we reach the next expected block. This isn't an error, but is the ideal conclusion of the loop.

tycho-client/src/feed/mod.rs
louise-poole: Sure. Though this transition method is really simple no? 
Lol - all I changed in this method is introduced a var for wait_duration so the if statement would fit on one line 😂

tycho-client/src/feed/mod.rs
louise-poole: Sure. Should we set it to a multiple of the block time? Any suggestion on what multiple to use?

tycho-client/src/feed/mod.rs
louise-poole: I can't find a clean way to merge these without the code getting more complicated to read. Any suggestions?
For now I'll just improve the comment.

tycho-client/src/feed/mod.rs
louise-poole: Thing is, not all SynchroniserStates have a header. Rust requires me to ensure the state I'm dealing with is one with a header before I can proceed with using the header... hence the `if let` binding.

tycho-client/src/feed/synchronizer.rs
louise-poole: Which should we use instead: `block_time+timeout` or a multiple of the block time?

kayibal: yeah

tycho-client/src/feed/mod.rs
```diff
@@ -290,7 +304,9 @@ impl SynchronizerStream {
                 let wait_duration = self
                     .modify_ts
                     .signed_duration_since(now);
-                if wait_duration > chrono::Duration::seconds(60) {
+                let stale_threshold_chrono = ChronoDuration::from_std(stale_threshold)
+                    .expect("Staleness threshold could not convert to chrono Duration");
```
kayibal, line 308: Is this expect safe to have here?

louise-poole: I would say so, yes. 'stale_threshold' should be measured in minutes and this conversion should only panic with very large values (hundreds of years). We can address better error handling in a separate PR?

kayibal (approved):

PR #546, opened 2025-04-03
feat(versioning): drop new entries where newer versions already exist in the DB

louise-poole:
Important decisions made:
- The versioning logic assumes data is inserted as block-scoped data. Therefore if the db has an update for a specific entity id at a given block, it is assumed all updates for that entity have been inserted at that block
- Due to the above, db data inserts/updates much be batched by block for dynamic/versioned tables. For the most part this is already how things are done on the indexer with one exception: new contracts.
- `upsert_contract` used to both insert a new contract and update the dynamic tables related to it. This makes it difficult to batch all changes done to the dynamic tables (slots, balance, code etc) into one db interaction as these same tables are also updated by `update_contracts`. Therefore `upsert_contract` was changed to `insert_contract` and only inserts the static contract values. It is expected that `update_contract` be subsequently called with the dynamic values for them to be inserted for new contracts. 

NOTE - it is expected that for a well designed substream package we will only get a single message for a given contract: either a creation with the latest state values, or an update with the latest state values. However if for any reason a substream doesn't abide by this and has both a creation and update event for the same contract on the same block, our decisions made above will allow us to predictably handle the scenario.

### DIFF
--- a/tycho-indexer/src/extractor/protocol_extractor.rs
+++ b/tycho-indexer/src/extractor/protocol_extractor.rs
@@ -1342,10 +1342,14 @@ impl ExtractorGateway for ExtractorPgGateway {
                     let new: Account = account_update.ref_into_account(&tx_update.tx);
                     info!(block_number = ?changes.block.number, contract_address = ?new.address, "NewContract");
 
-                    // Insert new accounts
+                    // Insert new account static values
                     self.state_gateway
-                        .upsert_contract(&new)
+                        .insert_contract(&new)
                         .await?;
+
+                    // Collect new account dynamic values for block-scoped batch insert (necessary
+                    // for correct versioning)
+                    account_changes.push((tx_update.tx.hash.clone(), account_update.clone()));
                 } else if account_update.is_update() {
                     account_changes.push((tx_update.tx.hash.clone(), account_update.clone()));
                 } else {
@@ -1410,6 +1414,7 @@ impl ExtractorGateway for ExtractorPgGateway {
         }
 
         // Insert component balance changes
+        dbg!(&component_balance_changes);
         if !component_balance_changes.is_empty() {
             self.state_gateway
                 .add_component_balances(component_balance_changes.as_slice())
@@ -2172,10 +2177,7 @@ mod test_serial_db {
         storage::BlockOrTimestamp,
         traits::TokenOwnerFinding,
     };
-    use tycho_storage::postgres::{
-        builder::GatewayBuilder, db_fixtures, db_fixtures::yesterday_midnight,
-        testing::run_against_db,
-    };
+    use tycho_storage::postgres::{builder::GatewayBuilder, db_fixtures, testing::run_against_db};
 
     use super::*;
     use crate::{
@@ -2340,7 +2342,7 @@ mod test_serial_db {
                     )
                     .unwrap(),
                     parent_hash: Bytes::default(),
-                    ts: db_fixtures::yesterday_one_am(),
+                    ts: db_fixtures::yesterday_half_past_midnight(),
                 }])
                 .await
                 .expect("block insertion succeeded");
@@ -2581,8 +2583,6 @@ mod test_serial_db {
     }
 
     // Tests a forward call with a native contract creation and an account update
-    // TODO: Fix this test. It was already disabled for native extractors, because of
-    // protocol_type_name mismatch
     #[ignore]
     #[tokio::test]
     async fn test_forward_native_protocol() {
@@ -2590,7 +2590,7 @@ mod test_serial_db {
             let (gw, _) = setup_gw(pool, ImplementationType::Custom).await;
             let msg = native_pool_creation();
 
-            let _exp = [ProtocolComponent {
+            let exp = [ProtocolComponent {
                 id: NATIVE_CREATED_CONTRACT.to_string(),
                 protocol_system: "test".to_string(),
                 protocol_type_name: "pool".to_string(),
@@ -2627,8 +2627,7 @@ mod test_serial_db {
                 .entity;
             println!("{res:?}");
 
-            // TODO: This is failing because protocol_type_name is wrong in the gateway - waiting
-            // assert_eq!(res, exp);
+            assert_eq!(res, exp);
         })
         .await;
     }
@@ -2681,6 +2680,7 @@ mod test_serial_db {
                 .unwrap();
 
             // TODO: improve asserts
+            dbg!(&component_balances);
             assert_eq!(component_balances.len(), 1);
             assert_eq!(component_balances[0].component_id, "ambient_USDC_ETH");
         })
@@ -3137,7 +3137,7 @@ mod test_serial_db {
             .expect("Failed to create extractor");
 
             // Send a sequence of block scoped data with the same timestamp.
-            let base_ts = yesterday_midnight().timestamp() as u64;
+            let base_ts = db_fixtures::yesterday_midnight().timestamp() as u64;
             let versions = [1, 2, 3, 4];
 
             let inp_sequence = versions
--- a/tycho-indexer/src/extractor/protocol_extractor.rs
+++ b/tycho-indexer/src/extractor/protocol_extractor.rs
@@ -1414,7 +1414,6 @@ impl ExtractorGateway for ExtractorPgGateway {
         }
 
         // Insert component balance changes
-        dbg!(&component_balance_changes);
         if !component_balance_changes.is_empty() {
             self.state_gateway
                 .add_component_balances(component_balance_changes.as_slice())
--- a/tycho-storage/src/postgres/contract.rs
+++ b/tycho-storage/src/postgres/contract.rs
@@ -994,12 +994,11 @@ impl PostgresGateway {
         Ok(WithTotal { entity: res, total: Some(total_count) })
     }
 
-    /// Upsert contract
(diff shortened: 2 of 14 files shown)


tycho-storage/src/postgres/versioning.rs
```diff
@@ -302,6 +377,18 @@ fn set_partitioned_versioning_attributes<N: PartitionedVersionedRow>(
...
+                if let Some(latest_row) = latest.get(id) {
+                    if latest_row.get_valid_from() > *delete_version {
+                        trace!(
+                            "Skipping delete for {:?} since it's older than the latest version",
+                            id
+                        );
+                        latest.remove(id);
+                        continue;
+                    }
+                }
```
kayibal: Did you think this case properly through? E.g. with attributes that oscillate between being deleted and being present...

kayibal (approved):
Thanks this looks good, just one q

tycho-storage/src/postgres/versioning.rs
```diff
@@ -282,6 +355,19 @@ fn set_partitioned_versioning_attributes<N: PartitionedVersionedRow>(
...
                 let mut row = r.clone();
+
+                // Skip if new version is older than the latest version in the db
+                if let Some(latest_row) = latest.get(&id) {
+                    if latest_row.get_valid_from() > row.get_valid_from() {
+                        trace!(
+                            "Skipping update for {:?} since it's older than the latest version",
+                            id
+                        );
+                        latest.remove(&id);
```
zizou0x: I think this breaks if we have two old updates for the same ids. We would ignore the first one and remove it from `latest` but the next one wouldn't be ignored because `latest.get(&id)` would be `None`

zizou0x (commented):
Looks good but I think there are some edge cases that aren't properly handled yet

I think we should add more extreme unit test cases to make sure we can handle them

tycho-storage/src/postgres/versioning.rs
louise-poole: Fixed it [here](https://github.com/propeller-heads/tycho-indexer/pull/546/commits/735928d9b2737683aaa9b0d403c131367f37b3d1)

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -1410,6 +1414,7 @@ impl ExtractorGateway for ExtractorPgGateway {
         }
 
         // Insert component balance changes
+        dbg!(&component_balance_changes);
```
zizou0x: Please remove this

tycho-storage/src/postgres/contract.rs
```diff
@@ -995,12 +995,11 @@ impl PostgresGateway {
...
 
-    /// Upsert contract
+    /// Insert contract
     ///
-    /// Inserts a contract or updates it if it already exists. It will not update
-    /// contract balance or contract code if they already exist though. Since a separate
+    /// Inserts a contract. It will not insert contract code, slots or balance since a separate
     /// method exists for updating these related components.
-    pub async fn upsert_contract(
+    pub async fn insert_contract(
```
zizou0x, line 1001: Isn't this still an upsert if we `do_update()`?

I would maybe rename to `upsert_account` instead?

tycho-storage/src/postgres/orm.rs
```diff
@@ -1620,13 +1612,17 @@ impl PartitionedVersionedRow for NewSlot {
...
-                    .and(contract_storage::slot.eq_any(&slots))
-                    .and(contract_storage::valid_to.eq(MAX_TS)),
+                    .and(contract_storage::slot.eq_any(&slots)),
             )
+            .distinct_on((contract_storage::account_id, contract_storage::slot))
+            .order_by((
+                contract_storage::account_id,
+                contract_storage::slot,
+                contract_storage::valid_to.desc(),
+            ))
```
zizou0x, line 1639: Isn't this completely equivalent to what was there before?

tycho-storage/src/postgres/versioning.rs
```diff
@@ -218,6 +231,37 @@ where
...
+        .map(|row| row.get_entity_id())
+        .collect();
+    let latest_db_versions = S::latest_versions_by_ids(entity_ids.clone(), conn)
+        .await
+        .map_err(PostgresError::from)?
+        .iter()
+        .map(|row| (row.get_entity_id(), row.get_valid_from()))
+        .collect::<HashMap<N::EntityId, N::Version>>();
+
+    // Filter out entries that are older than the latest version in the db
```
zizou0x: Older or same ts

tycho-storage/src/postgres/versioning.rs
```diff
@@ -266,31 +319,92 @@ type ArchivedRows<N> = Vec<N>;
...
+/// This function handles the versioning logic for partitioned tables by:
+/// 1. Setting up archive versions for entities that are being updated
+/// 2. Marking rows as deleted when deletion entries are encountered
+/// 3. Maintaining the latest version of each entity
+/// 4. Skipping updates or deletions that are older than versions already in the database
+///
+/// The function maintains a collection of latest rows, archived rows, and deleted entity IDs.
+/// For each entity:
+/// - If a newer version arrives, the current version is archived and the new one becomes the latest
+/// - If a deletion arrives, the current version is archived, marked as deleted, and added to the
```
zizou0x:
```suggestion
/// - If a new deletion arrives, the current version is archived, marked as deleted, and added to the
```

tycho-storage/src/postgres/versioning.rs
```diff
@@ -266,31 +319,92 @@ type ArchivedRows<N> = Vec<N>;
...
+///   deleted IDs
+/// - If an update or deletion is older than the current version in the database, it's skipped
```
zizou0x:
```suggestion
/// - If an update or deletion is older or equal than the current version in the database, it's skipped
```

tycho-storage/src/postgres/versioning.rs
```diff
@@ -266,31 +319,92 @@ type ArchivedRows<N> = Vec<N>;
...
+                }
+
                 if let Some(mut prev) = latest.remove(&id) {
                     prev.archive(&mut row);
                     archived.push(prev);
                 }
                 // If it's updated after being deleted, it doesn't need to be marked as deleted
                 deleted.remove(&id);
+                // If it is updated after being skipped, it doesn't need to be skipped
+                to_skip.remove(&id);
```
zizou0x: Can this happen? I like it like this, it's easy and robust, but I just wonder

zizou0x (approved):
Looks good to me, just some small comments. It would be nice to get one more approval just to be sure

tycho-storage/src/postgres/orm.rs
zizou0x: I think this is to handle deleted rows right?

tycho-storage/src/postgres/orm.rs
```diff
@@ -455,13 +455,17 @@ impl PartitionedVersionedRow for NewComponentBalance {
...
-                    .and(component_balance::token_id.eq_any(&token_ids))
-                    .and(component_balance::valid_to.eq(MAX_TS)),
+                    .and(component_balance::token_id.eq_any(&token_ids)),
             )
+            .distinct_on((component_balance::protocol_component_id, component_balance::token_id))
+            .order_by((
+                component_balance::protocol_component_id,
+                component_balance::token_id,
+                component_balance::valid_to.desc(),
+            ))
```
kayibal, line 473: Very worried about the performance here... I guess we had an index on valid_to and now this needs to query all partitions, aggregate the data in a temp table, sort it and then remove dupes...

After reading the all changes I understand you want to also get the latest deleted row but I still don't understand why.

tycho-storage/src/postgres/contract.rs
```diff
@@ -1034,68 +1033,27 @@ impl PostgresGateway {
...
-                .map_err(|err| {
-                    error!("Failed inserting account");
-                    storage_error_from_diesel(err, "Account", &hex_addr, None)
-                })?
-        };
+        use schema::account::dsl;
+        diesel::insert_into(schema::account::table)
+            .values(new_contract.new_account())
+            .on_conflict(on_constraint("account_chain_id_address_key"))
+            .do_update()
```
kayibal: Since it is called insert, I think this should error if the contract already exists.

tycho-storage/src/postgres/contract.rs
louise-poole: Ah true! Will change this back. Technically the stuff it updates is static (title, creation_tx etc). I'm not sure why we consider this updatable. But won't get into that in this PR - I'll just revert the rename.
Edit: a later comment by Alan suggests removing the update part so I'll do that instead.

tycho-storage/src/postgres/orm.rs
louise-poole: Yes - this is to handle deleted entities. Previously this would fetch only currently valid versions, but for our versioning logic to work we need to fetch the latest versions (irrespective of if they're considered current or not).

tycho-storage/src/postgres/orm.rs
louise-poole: I assume after the slack thread it's clear?
I've added a PERF comment here so that it's clear where we need to return to for the performance improvements.

Issue #532, opened 2025-03-21
Expose TVL information to tycho-client users


louise-poole:
# Context

Knowing the TVL of a pool can be very useful for many consumers of Tycho data: it could be used to determine the stability of the spot price of the pool, to prioritise pools by liquidity and much more. 
The Tycho indexer itself calculates component TVLs and stores them be used to filter component queries. Since they’re currently used for filtering only, the tvl values are not returned by any of our existing RPC endpoints. This means when a new tycho-client starts up and fetches initial snapshots via the RPC, it is not able to fetch tvl data for the user. This needs to be addressed.
Note: the streamed block updates already include tvl data. It is only snapshots that miss this information.

# Outline of required changes:

- [tycho-indexer] Add a GET `component_tvl` endpoint to the RPC. See [this](https://github.com/propeller-heads/tycho-indexer/issues/478) old issue as a starting point for what needs to be done for this.
- [tycho-client] Add a corresponding `get_component_tvls` fn to the RPCClient.
- [tycho-client] Add an `include_tvl` flag to the CliArgs (to allow for a consumer to opt in for tvl retrieval). Document this flag nicely. Be sure to indicate that skipping tvl retrieval reduces the number of requests used to start up the client which improves client efficiency.
- [tycho-client] Add an `include_tvl` bool field to `ProtocolStateSynchronizer` struct. 
- [tycho-client] Add a component_tvl field to `ComponentWithState` struct.
- [tycho-client] In  `ProtocolStateSynchronizer`’s `get_snapshots`, check the `self.include_tvl` property and if it is true, query for component tvl and add the data to the snapshots returned (using the updated ComponentWithState struct).

# Deliverables
- all changes outlined above
- unit tests for any new functions and endpoints
- documentation updates for the new endpoint and cli args

adpthegreat:
Hi please can you assign this to me? @louise-poole

PR #547, opened 2025-04-03
feat: add component tvl

zach030:
Close https://github.com/propeller-heads/tycho-indexer/issues/532

### DIFF
--- a/tycho-client/src/cli.rs
+++ b/tycho-client/src/cli.rs
@@ -93,6 +93,8 @@ struct CliArgs {
     max_missed_blocks: u64,
 
     /// If set, the synchronizer will include TVL in the messages.
+    /// Enabling this option will increase the number of network requests made during start-up,
+    //  which may result in increased start-up latency.
     #[clap(long)]
     include_tvl: bool,
 }
@@ -253,10 +255,10 @@ async fn run(exchanges: Vec<(String, Option<String>)>, args: CliArgs) {
             filter,
             3,
             !args.no_state,
+            args.include_tvl,
             rpc_client.clone(),
             ws_client.clone(),
             args.block_time + args.timeout,
-            args.include_tvl,
         );
         block_sync = block_sync.register_synchronizer(id, sync);
     }
--- a/tycho-client/src/cli.rs
+++ b/tycho-client/src/cli.rs
@@ -91,6 +91,10 @@ struct CliArgs {
     /// in conjunction with block_time to calculate a timeout: block_time * max_missed_blocks.
     #[clap(long, default_value = "10")]
     max_missed_blocks: u64,
+
+    /// If set, the synchronizer will include TVL in the messages.
+    #[clap(long)]
+    include_tvl: bool,
 }
 
 impl CliArgs {
@@ -252,6 +256,7 @@ async fn run(exchanges: Vec<(String, Option<String>)>, args: CliArgs) {
             rpc_client.clone(),
             ws_client.clone(),
             args.block_time + args.timeout,
+            args.include_tvl,
         );
         block_sync = block_sync.register_synchronizer(id, sync);
     }
--- a/tycho-client/src/feed/synchronizer.rs
+++ b/tycho-client/src/feed/synchronizer.rs
@@ -156,10 +156,10 @@ where
         component_filter: ComponentFilter,
         max_retries: u64,
         include_snapshots: bool,
+        include_tvl: bool,
         rpc_client: R,
         deltas_client: D,
         timeout: u64,
-        include_tvl: bool,
     ) -> Self {
         Self {
             extractor_id: extractor_id.clone(),
@@ -225,11 +225,12 @@ where
         }
 
         let component_tvl = if self.include_tvl {
+            let body = ProtocolComponentTvlRequestBody::id_filtered(
+                request_ids.clone(),
+                self.extractor_id.chain,
+            );
             self.rpc_client
-                .get_component_tvl(&ProtocolComponentTvlRequestBody {
-                    chain: self.extractor_id.chain,
-                    component_ids: Some(request_ids.clone()),
-                })
+                .get_component_tvl_paginated(&body, 100, 4)
                 .await?
                 .tvl
         } else {
@@ -668,6 +669,7 @@ mod test {
 
     fn with_mocked_clients(
         native: bool,
+        include_tvl: bool,
         rpc_client: Option<MockRPCClient>,
         deltas_client: Option<MockDeltasClient>,
     ) -> ProtocolStateSynchronizer<ArcRPCClient<MockRPCClient>, ArcDeltasClient<MockDeltasClient>>
@@ -681,10 +683,10 @@ mod test {
             ComponentFilter::with_tvl_range(50.0, 50.0),
             1,
             true,
+            include_tvl,
             rpc_client,
             deltas_client,
             10_u64,
-            false,
         )
     }
 
@@ -698,13 +700,22 @@ mod test {
         }
     }
 
+    fn component_tvl_snapshot() -> ProtocolComponentTvlRequestResponse {
+        let tvl = HashMap::from([("Component1".to_string(), 100.0)]);
+
+        ProtocolComponentTvlRequestResponse {
+            tvl,
+            pagination: PaginationResponse { page: 0, page_size: 20, total: 1 },
+        }
+    }
+
     #[test(tokio::test)]
     async fn test_get_snapshots_native() {
         let header = Header::default();
         let mut rpc = MockRPCClient::new();
         rpc.expect_get_protocol_states()
             .returning(|_| Ok(state_snapshot_native()));
-        let state_sync = with_mocked_clients(true, Some(rpc), None);
+        let state_sync = with_mocked_clients(true, false, Some(rpc), None);
         let mut tracker = ComponentTracker::new(
             Chain::Ethereum,
             "uniswap-v2",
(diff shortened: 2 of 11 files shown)


tycho-client/src/cli.rs
```diff
@@ -91,6 +91,10 @@ struct CliArgs {
     /// in conjunction with block_time to calculate a timeout: block_time * max_missed_blocks.
     #[clap(long, default_value = "10")]
     max_missed_blocks: u64,
+
+    /// If set, the synchronizer will include TVL in the messages.
+    #[clap(long)]
+    include_tvl: bool,
```
louise-poole: Please extend the comment to make it clear to a user that including tvl increases the number of requests used at start-up and therefore increases start-up latency.

tycho-client/src/stream.rs
```diff
@@ -201,6 +201,7 @@ impl TychoStreamBuilder {
                 rpc_client.clone(),
                 ws_client.clone(),
                 self.block_time + self.timeout,
+                false,
```
louise-poole: Apologies for missing this in the plan: but please extend the TychoStreamBuilder to allow setting `include_tvl` to true? Make sure to document that this will increase start-up latency.

tycho-client/src/feed/synchronizer.rs
```diff
@@ -157,6 +159,7 @@ where
         rpc_client: R,
         deltas_client: D,
         timeout: u64,
+        include_tvl: bool,
```
louise-poole: Hate to be picky here: but please could you move this 'include_tvl' parameter to just under the 'include_snapshots' one? I'd like to logically group them.

tycho-common/src/dto.rs
```diff
@@ -1322,6 +1322,27 @@ impl ProtocolSystemsRequestResponse {
...
 }
 
+#[derive(Serialize, Deserialize, Debug, Default, PartialEq, ToSchema, Eq, Hash, Clone)]
+#[serde(deny_unknown_fields)]
+pub struct ProtocolComponentTvlRequestBody {
+    #[serde(default)]
+    pub chain: Chain,
+    #[serde(default)]
+    pub component_ids: Option<Vec<String>>,
+}
```
louise-poole: Please add pagination to the request body? We're enforcing that all endpoints use pagination... even lightweight ones like fetching supported protocol systems (see `ProtocolSystemsRequestBody`).

tycho-indexer/src/services/rpc.rs
```diff
@@ -489,6 +489,33 @@ where
...
+        let ids_slice = ids_strs.as_deref();
+
+        match self
+            .db_gateway
+            .get_component_tvls(&chain, ids_slice)
+            .await
+        {
+            Ok(tvl) => Ok(dto::ProtocolComponentTvlRequestResponse::new(tvl)),
+            Err(err) => {
+                error!(error = %err, "Error while getting protocol systems.");
```
```suggestion
                error!(error = %err, "Error while getting component tvls.");
```

tycho-common/src/storage.rs
```diff
@@ -478,6 +478,23 @@ pub trait ProtocolGateway {
         chain: &Chain,
         pagination_params: Option<&PaginationParams>,
     ) -> Result<WithTotal<Vec<String>>, StorageError>;
+
+    /// Retrieve the total value locked for a component
+    ///
+    /// Fetches the total value locked for a component
```
louise-poole: Considering it's getting multiple tvls we should adjust this doc string. Also the second sentence doesn't add new info so we could skip it.
```suggestion
    /// Retrieve the components total value locked (TVL).
```

tycho-common/src/storage.rs
```diff
@@ -478,6 +478,23 @@ pub trait ProtocolGateway {
...
+    ///
+    /// # Parameters
+    /// - `chain` The chain for which to retrieve the total value locked
+    /// - `system` The protocol system for which to retrieve the total value locked
```
louise-poole, line 486: You mention 'system' here but don't have a parameter for it? Might actually be a good idea to add the parameter since the ids list is optional. Therefore we could fetch all component tvl for a given protocol system.
Please add it to `ProtocolComponentTvlRequestBody` too then?

tycho-common/src/storage.rs
```diff
@@ -478,6 +478,23 @@ pub trait ProtocolGateway {
...
+    /// - `ids` The ids of the components to retrieve the total value locked for
+    ///
+    /// # Return
+    /// The total value locked for the component
```
louise-poole: More accurate would be something more like:
```suggestion
    /// A result with a map of component ids to their TVL. Err if storage access failed.
```

louise-poole (commented):
Hi @zach030 
Great start, thank you for this! I left some comments on some adjustments we'd need. Looking forward to having this feature merged, thank you very much for taking it on!

tycho-storage/src/postgres/protocol.rs
```diff
@@ -1741,6 +1741,77 @@ impl PostgresGateway {
...
+        use schema::{component_tvl::dsl as ct, protocol_component::dsl as pc};
+
+        if !self.chain_id_cache.value_exists(chain) {
+            return Err(StorageError::NotFound("Chain".to_string(), chain.to_string()));
+        }
+
+        let system_id = match system {
+            Some(ref s) => Some(self.get_protocol_system_id(s)?),
+            None => None,
+        };
```
louise-poole: To clean this up a little, we could just move `self.get_protocol_system_id(s)` inside of the block lower down where you apply the system filter?

tycho-storage/src/postgres/protocol.rs
```diff
@@ -3888,4 +3959,152 @@ mod test {
...
+    async fn test_get_component_tvls() {
+        let mut conn = setup_db().await;
+        setup_data(&mut conn).await;
+        let gw = EVMGateway::from_connection(&mut conn).await;
+
+        let protocol_system = "ambient".to_string();
+        let new_tvl = [("state1".to_owned(), 100.0), ("no_tvl".to_owned(), 1.0)]
+            .into_iter()
+            .collect::<HashMap<_, _>>();
+        gw.upsert_component_tvl(&Chain::Ethereum, &new_tvl, &mut conn)
```
louise-poole: We tend to try isolate testing one feature only per unit test. Here you're making use of the `upsert_component_tvl` gateway function as well as your new `get_component_tvl` function - so are technically covering 2 things in this test. I would prefer you ensure the database is set up with the data you want to test within the `setup_data` function (I think it already is) and skip using `upsert_component_tvl` entirely.

tycho-storage/src/postgres/protocol.rs
```diff
@@ -3888,4 +3959,152 @@ mod test {
...
+                &mut conn,
+            )
+            .await
+            .expect("failed retrieving component tvl");
+
+        assert_eq!(res.total.unwrap(), 2);
+        assert_eq!(res.entity.get("state1"), Some(&100.0));
+        assert_eq!(res.entity.get("no_tvl"), Some(&1.0));
+        assert!(!res.entity.contains_key("not_exist"));
+    }
```
louise-poole: This test can be merged with the one above it.

tycho-storage/src/postgres/protocol.rs
```diff
@@ -3888,4 +3959,152 @@ mod test {
...
+                Some(protocol_system),
+                None,
+                Some(&PaginationParams { page: 0, page_size: 10 }),
+                &mut conn,
+            )
+            .await
+            .expect("failed retrieving component tvl");
+
+        assert_eq!(tvls.entity.get("state1"), Some(&100.0));
+        assert_eq!(tvls.entity.get("no_tvl"), Some(&1.0));
```
louise-poole: I would expect `state3` too no?

tycho-storage/src/postgres/protocol.rs
```diff
@@ -3888,4 +3959,152 @@ mod test {
...
+            .collect::<HashMap<_, _>>();
+        gw.upsert_component_tvl(&Chain::Ethereum, &new_tvl, &mut conn)
+            .await
+            .expect("upsert failed!");
+
+        let tvls = gw
+            .get_component_tvls(
+                &Chain::Ethereum,
+                Some(protocol_system),
+                Some(&["state1", "no_tvl"]),
```
louise-poole: Would also be nice to add a component id that doesn't belong to 'ambient' in our test data. Looks like `state2` belongs to 'zigzag' protocol. Please add it here so that we can ensure both the protocol system and the component ids are used in the filter correctly (should not be in our result as it should get filtered by protocol component).

louise-poole (commented):
Perfect! Just some test updates/improvements and then I think we're good to go! 👍

tycho-indexer/src/services/mod.rs
```diff
@@ -252,6 +252,10 @@ where
                     web::resource(format!("/{}/protocol_systems", self.prefix))
                         .route(web::post().to(rpc::protocol_systems::<G>)),
                 )
+                .service(
+                    web::resource(format!("/{}/component_tvl", self.prefix))
+                        .route(web::post().to(rpc::component_tvl::<G>)),
+                )
```
louise-poole, line 262: In this file, you also need to add `ProtocolComponentTvlRequestBody` and `ProtocolComponentTvlRequestResponse` to openapi the schemas.

tycho-common/src/dto.rs
```diff
@@ -1323,6 +1323,53 @@ impl ProtocolSystemsRequestResponse {
     }
 }
 
+#[derive(Serialize, Deserialize, Debug, Default, PartialEq, ToSchema, Eq, Hash, Clone)]
+#[serde(deny_unknown_fields)]
+pub struct ProtocolComponentTvlRequestBody {
```
louise-poole: Would be easier to read if we just called this `ComponentTvlRequestBody`. Same for the response.

tycho-indexer/src/services/mod.rs
louise-poole: And add the endpoint to the `paths`.

louise-poole (commented):
Ah - a few small extra things I almost missed!

tycho-storage/src/postgres/protocol.rs
```diff
@@ -3965,76 +3961,31 @@ mod test {
...
-
-        assert_eq!(res.total.unwrap(), 2);
-        assert_eq!(res.entity.get("state1"), Some(&100.0));
-        assert_eq!(res.entity.get("no_tvl"), Some(&1.0));
-        assert!(!res.entity.contains_key("not_exist"));
+        assert_eq!(tvls.total.unwrap(), 1);
+        assert_eq!(tvls.entity.get("state1"), Some(&2.0));
+        assert!(!tvls.entity.contains_key("no_tvl")); // component tvl is 0
+        assert!(!tvls.entity.contains_key("not_exist")); // component does not exist
+        assert!(!tvls.entity.contains_key("state2")); // component is not in the system
```
louise-poole:
```suggestion
        assert!(!tvls.entity.contains_key("no_tvl")); // component not requested
        assert!(!tvls.entity.contains_key("not_exist")); // component does not exist
        assert!(!tvls.entity.contains_key("state2")); // component not in the requested system
```

louise-poole (approved):
Just left a comment for a small doc improvement, otherwise this is ready to merge 👌

PR #548, opened 2025-04-07
feat(dci): add entry point tracing functionality

zizou0x:
This commit introduces a new `EntryPointTracer` trait for tracing smart contract entry points, along with related types such as `EntryPoint` and `TracedEntryPoint`. The `trace` method allows for analyzing execution behavior at a specific block. Additionally, the `tycho_common::blockchain` module has been updated to include necessary structures for this functionality.

This is the first version of the entry point tracing, we have plans to add more features in the next iterations, such as symbolic execution, fuzzing, ect..

### DIFF
--- a/tycho-common/src/models/blockchain.rs
+++ b/tycho-common/src/models/blockchain.rs
@@ -1,6 +1,6 @@
 use std::{
     any::Any,
-    collections::{hash_map::Entry, HashMap},
+    collections::{hash_map::Entry, HashMap, HashSet},
     sync::Arc,
 };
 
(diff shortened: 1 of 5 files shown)


tycho-common/src/models/blockchain.rs
```diff
@@ -341,6 +342,31 @@ pub enum BlockTag {
     /// Block by number
     Number(u64),
 }
+#[derive(Debug, Clone)]
+/// An entry point to trace. Different types of entry points tracing will be supported in the
+/// future. Like RPC debug tracing, symbolic execution, etc.
+pub enum EntryPoint {
```
kayibal, line 348: Can you please add this:

https://doc.rust-lang.org/reference/attributes/type_system.html#r-attributes.type-system.non_exhaustive

PR #549, opened 2025-04-08
feat: Implement retry logic for deadlock detection in database transactions

zizou0x:


tycho-storage/src/postgres/cache.rs
```diff
@@ -310,30 +310,61 @@ impl DBCacheWriteExecutor {
...
                         }
+                        Result::<(), PostgresError>::Ok(())
+                    }
+                    .scope_boxed()
+                })
+                .await;
+
+            match res {
+                Ok(_) => break,
+                Err(PostgresError(StorageError::Unexpected(ref e)))
```
zizou0x, line 349: Should we retry for any error?

tycho-storage/src/postgres/cache.rs
```diff
@@ -310,30 +310,61 @@ impl DBCacheWriteExecutor {
...
+                    if e.contains("deadlock detected") =>
+                {
+                    retry_count += 1;
+                    if retry_count < max_retries {
+                        let delay = std::time::Duration::from_secs(retry_count);
+                        debug!(
+                            "Deadlock detected, retrying in {:?} (attempt {}/{})",
+                            delay,
+                            retry_count + 1,
```
tvinagre, line 358: you've already added 1 above, so I think this should be `retry_count`
```suggestion
                            retry_count,
```

tvinagre (approved):

PR #551, opened 2025-04-10
feat: Postgres and Gateway functions for EntryPoints

zizou0x:
This PR adds the core elements for DCI entry points in `tycho-storage`. This includes both DB migrations and gateway functions. It also adds the `tycho-ethereum` tracing module.

This implementation follows the design agreed on by the team: the main links are with tracing data to keep things as granular as possible and be able to trace back a tracing data/result to the related component.

A bit more context: EntryPoints are new entities defined by Substreams modules that want to use DCI. Inside the Substreams module we most of the time only know about the first level of dependency (Pool X relies on contract A). But we don't know if contract A itself relies on another contract. We decided to implement a tracing feature that will be used to detect these subdependencies. EntryPoints are used by the tracing module to perform the tracing operation.

### DIFF
--- a/tycho-common/src/models/mod.rs
+++ b/tycho-common/src/models/mod.rs
@@ -54,6 +54,9 @@ pub type ComponentId = String;
 /// Protocol system literal type to uniquely identify a protocol system.
 pub type ProtocolSystem = String;
 
+/// Entry point id literal type to uniquely identify an entry point.
+pub type EntryPointId = String;
+
 #[derive(
     Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize, EnumString, Display, Default,
 )]
--- a/tycho-common/src/storage.rs
+++ b/tycho-common/src/storage.rs
@@ -1,5 +1,8 @@
 //! Storage traits used by Tycho
-use std::{collections::HashMap, fmt::Display};
+use std::{
+    collections::{HashMap, HashSet},
+    fmt::Display,
+};
 
 use async_trait::async_trait;
 use chrono::NaiveDateTime;
@@ -8,17 +11,15 @@ use thiserror::Error;
 use crate::{
     dto,
     models::{
-        blockchain::{
-            Block, EntryPoint, EntryPointWithData, TracedEntryPoint, TracingResult, Transaction,
-        },
+        blockchain::{Block, EntryPointWithData, TracedEntryPoint, TracingResult, Transaction},
         contract::{Account, AccountBalance, AccountDelta},
         protocol::{
             ComponentBalance, ProtocolComponent, ProtocolComponentState,
             ProtocolComponentStateDelta, QualityRange,
         },
         token::CurrencyToken,
-        Address, BlockHash, Chain, ComponentId, ContractId, ExtractionState, PaginationParams,
-        ProtocolSystem, ProtocolType, TxHash,
+        Address, BlockHash, Chain, ComponentId, ContractId, EntryPointId, ExtractionState,
+        PaginationParams, ProtocolSystem, ProtocolType, TxHash,
     },
     Bytes,
 };
@@ -515,11 +516,11 @@ pub trait EntryPointGateway {
         traced_entry_points: &[TracedEntryPoint],
     ) -> Result<(), StorageError>;
 
-    /// Retrieves the tracing results for an entry point from the database.
-    async fn get_traced_entry_point(
+    /// Retrieves all tracing results for a set of entry points from the database.
+    async fn get_traced_entry_points(
         &self,
-        entry_point: EntryPoint,
-    ) -> Result<Vec<TracingResult>, StorageError>;
+        entry_points: &HashSet<EntryPointId>,
+    ) -> Result<HashMap<EntryPointId, Vec<TracingResult>>, StorageError>;
 }
 
 /// Manage contracts and their state in storage.
--- a/tycho-storage/src/postgres/cache.rs
+++ b/tycho-storage/src/postgres/cache.rs
@@ -1,4 +1,8 @@
-use std::{collections::HashMap, num::NonZeroUsize, sync::Arc};
+use std::{
+    collections::{HashMap, HashSet},
+    num::NonZeroUsize,
+    sync::Arc,
+};
 
 use async_trait::async_trait;
 use chrono::NaiveDateTime;
@@ -15,17 +19,15 @@ use tracing::{debug, info, info_span, instrument, trace, warn, Instrument};
 use tycho_common::{
     models::{
         self,
-        blockchain::{
-            Block, EntryPoint, EntryPointWithData, TracedEntryPoint, TracingResult, Transaction,
-        },
+        blockchain::{Block, EntryPointWithData, TracedEntryPoint, TracingResult, Transaction},
         contract::{Account, AccountBalance, AccountDelta},
         protocol::{
             ComponentBalance, ProtocolComponent, ProtocolComponentState,
             ProtocolComponentStateDelta, QualityRange,
         },
         token::CurrencyToken,
-        Address, Chain, ComponentId, ContractId, ExtractionState, PaginationParams, ProtocolType,
-        TxHash,
+        Address, Chain, ComponentId, ContractId, EntryPointId, ExtractionState, PaginationParams,
+        ProtocolType, TxHash,
     },
     storage::{
         BlockIdentifier, BlockOrTimestamp, ChainGateway, ContractStateGateway, EntryPointFilter,
@@ -497,13 +499,12 @@ impl DBCacheWriteExecutor {
             WriteOp::UpsertEntryPoints(new_entry_points) => {
                 self.state_gateway
                     .insert_entry_points(
-                        new_entry_points
+                        &new_entry_points
                             .iter()
                             .map(|(component_id, entry_points)| {
                                 (component_id.as_str(), entry_points)
                             })
-                            .collect::<Vec<_>>()
-                            .as_slice(),
+                            .collect::<HashMap<_, _>>(),
                         &self.chain,
                         conn,
                     )
@@ -1174,16 +1175,16 @@ impl EntryPointGateway for CachedGateway {
     }
 
     #[instrument(skip_all)]
-    async fn get_traced_entry_point(
+    async fn get_traced_entry_points(
         &self,
-        entry_point: EntryPoint,
-    ) -> Result<Vec<TracingResult>, StorageError> {
+        entry_points: &HashSet<EntryPointId>,
+    ) -> Result<HashMap<EntryPointId, Vec<TracingResult>>, StorageError> {
         let mut conn =
             self.pool.get().await.map_err(|e| {
                 StorageError::Unexpected(format!("Failed to retrieve connection: {e}"))
             })?;
         self.state_gateway
-            .get_traced_entry_point(entry_point, &mut conn)
+            .get_traced_entry_points(entry_points, &mut conn)
             .await
     }
 }
--- a/tycho-storage/src/postgres/entry_point.rs
+++ b/tycho-storage/src/postgres/entry_point.rs
@@ -10,7 +10,7 @@ use tycho_common::{
         blockchain::{
             EntryPoint, EntryPointTracingData, EntryPointWithData, TracedEntryPoint, TracingResult,
         },
-        Chain,
+        Chain, EntryPointId,
     },
     storage::{EntryPointFilter, StorageError},
     Bytes,
@@ -27,9 +27,16 @@ use super::{
 };
 
 impl PostgresGateway {
+    /// Insert entry points into the database.
+    ///
+    /// # Arguments
+    ///
+    /// * `new_data` - A map of protocol component external ids to a list of entry points.
+    /// * `chain` - The chain to insert the entry points for.
+    /// * `conn` - The database connection to use.
     pub(crate) async fn insert_entry_points(
         &self,
-        new: &[(&str, &Vec<EntryPointWithData>)],
+        new_data: &HashMap<&str, &Vec<EntryPointWithData>>,
         chain: &Chain,
         conn: &mut AsyncPgConnection,
     ) -> Result<(), StorageError> {
@@ -41,9 +48,10 @@ impl PostgresGateway {
         let chain_id = self.get_chain_id(chain);
 
         let pc_ids = orm::ProtocolComponent::ids_by_external_ids(
-            &new.iter()
-                .map(|(id_, _)| *id_)
-                .collect::<Vec<&str>>(),
+            &new_data
+                .keys()
+                .map(Clone::clone)
+                .collect::<Vec<_>>(),
             chain_id,
             conn,
         )
@@ -53,7 +61,7 @@ impl PostgresGateway {
         .map(|(id_, ext_id)| (ext_id, id_))
         .collect::<HashMap<_, _>>();
 
-        let new_entry_points = new
+        let new_entry_points = new_data
             .iter()
             .flat_map(|(_, ep)| {
                 ep.iter()
@@ -76,7 +84,7 @@ impl PostgresGateway {
 
         // Fetch entry points by their external_ids, we can't use .returning() on the insert above
         // because it doesn't return the ids on conflicts.
-        let input_external_ids: Vec<String> = new
+        let input_external_ids: Vec<String> = new_data
             .iter()
             .flat_map(|(_, ep)| {
                 ep.iter()
@@ -87,7 +95,7 @@ impl PostgresGateway {
         let entry_point_ids =
             orm::EntryPoint::ids_by_external_ids(&input_external_ids, conn).await?;
 
-        let new_tracing_data = new
+        let new_tracing_data = new_data
             .iter()
             .flat_map(|(_, ep)| {
                 ep.iter().map(|ep| {
@@ -125,7 +133,8 @@ impl PostgresGateway {
             .map_err(|e| storage_error_from_diesel(e, "EntryPointData", "Batch upsert", None))?;
 
         let data_ids = orm::EntryPointTracingData::ids_by_entry_point_with_data(
-            &new.iter()
+            &new_data
+                .iter()
                 .flat_map(|(_, ep)| ep.iter())
                 .cloned()
                 .collect::<Vec<_>>(),
@@ -133,7 +142,7 @@ impl PostgresGateway {
         )
         .await?;
 
-        let pc_links = new
+        let pc_links = new_data
             .iter()
             .flat_map(|(pc_ext_id, ep)| {
                 ep.iter().map(|ep| {
@@ -174,6 +183,12 @@ impl PostgresGateway {
         Ok(())
     }
 
+    /// Get entry points with data from the database.
+    ///
+    /// # Arguments
+    ///
+    /// * `filter` - The filter to apply to the query.
+    /// * `conn` - The database connection to use.
     pub(crate) async fn get_entry_points_with_data(
         &self,
         filter: EntryPointFilter,
@@ -208,6 +223,12 @@ impl PostgresGateway {
             .collect())
     }
 
+    /// Upsert traced entry points into the database.
+    ///
+    /// # Arguments
+    ///
+    /// * `traced_entry_points` - The traced entry points to upsert.
+    /// * `conn` - The database connection to use.
     pub(crate) async fn upsert_traced_entry_points(
         &self,
         traced_entry_points: &[TracedEntryPoint],
@@ -339,32 +360,66 @@ impl PostgresGateway {
         Ok(())
     }
 
-    pub(crate) async fn get_traced_entry_point(
+    /// Get all tracing results for a set of entry points from the database.
+    ///
+    /// # Arguments
+    ///
+    /// * `entry_points` - The entry point ids to get tracing results for.
+    /// * `conn` - The database connection to use.
+    pub(crate) async fn get_traced_entry_points(
         &self,
-        entry_point: EntryPoint,
+        entry_points: &HashSet<EntryPointId>,
         conn: &mut AsyncPgConnection,
-    ) -> Result<Vec<TracingResult>, StorageError> {
+    ) -> Result<HashMap<EntryPointId, Vec<TracingResult>>, StorageError> {
         use schema::entry_point_tracing_result::dsl::*;
-        let entry_point_id = orm::EntryPoint::id_by_target_and_signature(
-            &entry_point.target,
-            &entry_point.signature,
+        let entry_point_ids = orm::EntryPoint::ids_by_external_ids(
+            &entry_points
+                .iter()
+                .map(Clone::clone)
+                .collect::<Vec<_>>(),
             conn,
         )
         .await?;
 
+        // Reverse the map, this is safe because we know it's a 1:1 mapping.
+        // This makes retrieving the external id attached to a tracing result faster.
+        let reverse: HashMap<i64, String> = entry_point_ids
+            .iter()
+            .map(|(k, &v)| (v, k.clone()))
+            .collect();
+
         let results = entry_point_tracing_result
-            .filter(entry_point_tracing_data_id.eq(entry_point_id))
-            .select(detection_data)
-            .load::<serde_json::Value>(conn)
+            .filter(entry_point_tracing_data_id.eq_any(entry_point_ids.values().cloned()))
+            .select((entry_point_tracing_data_id, detection_data))
+            .load::<(i64, serde_json::Value)>(conn)
             .await
             .map_err(|e| storage_error_from_diesel(e, "TracingResult", "Query", None))?;
 
-        results
+        let mut results_by_entry_point = HashMap::new();
+        for (ep_id, tracing_result) in results {
+            let ep_ext_id = reverse.get(&ep_id).ok_or_else(|| {
+                StorageError::NotFound("EntryPoint".to_string(), ep_id.to_string())
+            })?;
+            results_by_entry_point
+                .entry(ep_ext_id.clone())
+                .or_insert_with(Vec::new)
+                .push(tracing_result);
+        }
+
+        results_by_entry_point
             .into_iter()
-            .map(|(data)| {
-                serde_json::from_value(data).map_err(|e| {
-                    StorageError::Unexpected(format!("Failed to deserialize TracingResult: {e}"))
-                })
+            .map(|(ep_ext_id, data)| {
+                let converted_data = data
+                    .into_iter()
+                    .map(|d| {
+                        serde_json::from_value(d).map_err(|e| {
+                            StorageError::Unexpected(format!(
+                                "Failed to deserialize TracingResult: {e}"
+                            ))
+                        })
+                    })
+                    .collect::<Result<Vec<_>, StorageError>>()?;
+                Ok((ep_ext_id, converted_data))
             })
             .collect()
     }
@@ -499,7 +554,7 @@ mod test {
 
         let entry_point = rpc_tracer_entry_point();
         gw.insert_entry_points(
-            &[("pc_0", &vec![entry_point.clone()])],
+            &HashMap::from([("pc_0", &vec![entry_point.clone()])]),
             &Chain::Ethereum,
             &mut conn,
         )
@@ -523,7 +578,7 @@ mod test {
 
         let entry_point = rpc_tracer_entry_point();
         gw.insert_entry_points(
-            &[("pc_0", &vec![entry_point.clone()])],
+            &HashMap::from([("pc_0", &vec![entry_point.clone()])]),
             &Chain::Ethereum,
             &mut conn,
         )
@@ -556,7 +611,7 @@ mod test {
         let traced_entry_point = traced_entry_point();
 
         gw.insert_entry_points(
-            &[("pc_0", &vec![entry_point.clone()])],
+            &HashMap::from([("pc_0", &vec![entry_point.clone()])]),
             &Chain::Ethereum,
             &mut conn,
         )
@@ -568,10 +623,25 @@ mod test {
             .unwrap();
 
         let retrieved_traced_entry_points = gw
-            .get_traced_entry_point(entry_point.entry_point, &mut conn)
+            .get_traced_entry_points(
+                &HashSet::from([entry_point
+                    .entry_point
+                    .external_id
+                    .clone()]),
+                &mut conn,
+            )
             .await
             .unwrap();
 
-        assert_eq!(retrieved_traced_entry_points, vec![traced_entry_point.tracing_result]);
+        assert_eq!(
+            retrieved_traced_entry_points,
+            HashMap::from([(
+                entry_point
+                    .entry_point
+                    .external_id
+                    .clone(),
+                vec![traced_entry_point.tracing_result]
+            )])
+        );
     }
 }
--- a/tycho-storage/src/postgres/orm.rs
+++ b/tycho-storage/src/postgres/orm.rs
@@ -1817,6 +1817,7 @@ impl EntryPoint {
     }
 
     /// Retrieves the database id of an entry point from a target address and function signature.
+    #[allow(dead_code)]
     pub(crate) async fn id_by_target_and_signature(
         target_: &Bytes,
         signature_: &String,


check.sh
```diff
@@ -1,6 +1,6 @@
 set -e 
 
-cargo +nightly-2024-04-03 fmt -- --check
-cargo +nightly-2024-04-03 clippy --locked --all --all-features --all-targets -- -D warnings
+cargo +nightly fmt -- --check
+cargo +nightly clippy --locked --all --all-features --all-targets -- -D warnings
```
tvinagre: was about to do that in my PR 😄 Tnx!

tycho-common/src/models/blockchain.rs
```diff
@@ -342,32 +342,89 @@ pub enum BlockTag {
...
+
+impl EntryPoint {
+    pub fn new(target: Address, signature: String) -> Self {
+        Self { target, signature }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct EntryPointWithData {
+    pub entry_point: EntryPoint,
```
tvinagre, line 365: Can you document this struct pls?

tycho-common/src/models/blockchain.rs
```diff
@@ -342,32 +342,89 @@ pub enum BlockTag {
...
+    pub data: Vec<EntryPointTracingData>,
```
tvinagre: If I have 2 components that have the same `EntryPoint` (address and signature) but are called with different calldata, will they be grouped on the same `EntryPointWithData`? 
OR is `EntryPointWithData` per-component, and allows for one component have a set of calldata to be traced?

tycho-storage/migrations/2025-04-09_dci_entry_points/up.sql
```diff
@@ -0,0 +1,37 @@
...
+CREATE TABLE IF NOT EXISTS "protocol_component_holds_entry_point"(
+    "protocol_component_id" bigint REFERENCES "protocol_component"(id) NOT NULL,
+    "entry_point_id" bigint REFERENCES "entry_point"(id) NOT NULL,
+    PRIMARY KEY ("protocol_component_id", "entry_point_id")
+);
+
+CREATE TABLE IF NOT EXISTS "traced_entry_point"(
+    "entry_point_id" bigint NOT NULL REFERENCES "entry_point"(id) PRIMARY KEY,
+    "detection_block" bigint NOT NULL REFERENCES "block"(id),
+    "detection_data" JSONB NOT NULL
```
tvinagre: JSONB to keep it flexible for the future?

tvinagre: What I'm not sure about this design is the following issue:
Component A calls EntrypointA with params X
Component B calls EntrypointA with params Y.

Since we create EntryPoint <-> Component relation, ComponentA and ComponentB would be tied to the same entrypoint.

Assuming that different call parameters return differnet accounts & slots, we can't differentiate the accounts that are relevant to ComponentA from the ones that are used by ComponentB.

So, if we have a Protocol with many Components with one EntryPoint and the client only requests only one Component, we would still be returning ALL the accounts, most of which are irrelevant to him.

tvinagre: If we want to relate Component -> traced accounts, we'd need to use 
``` sql
CREATE TABLE IF NOT EXISTS "protocol_component_holds_entry_point_tracing_data"(
    "protocol_component_id" bigint REFERENCES "protocol_component"(id) NOT NULL,
    "entry_point_id" bigint REFERENCES "entry_point_tracing_data"(id) NOT NULL,
    PRIMARY KEY ("protocol_component_id", "entry_point_id")
);
```

tycho-storage/src/postgres/entry_point.rs
```diff
@@ -0,0 +1,536 @@
...
+use crate::postgres::{
+    orm::{
+        EntryPointTracingType, NewEntryPointCallsAccount, NewEntryPointTracingData,
+        NewProtocolComponentHoldsEntryPoint,
+    },
+    PostgresError,
+};
+
+pub struct EntryPointFilter {
+    protocol_name: Option<String>,
```
tvinagre: I'm assuming this is `ProtocolSystem`, not `ProtocolComponent` - please document

tycho-storage/src/postgres/entry_point.rs
```diff
@@ -0,0 +1,536 @@
...
+
+impl PostgresGateway {
+    pub async fn upsert_entry_points(
+        &self,
+        entry_points: &[EntryPointWithData],
+        component_id: &str,
+        chain: &Chain,
+        conn: &mut AsyncPgConnection,
+    ) -> Result<(), StorageError> {
+        use schema::{
```
tvinagre, line 43: noob question but why the choice of importing inside the function?

tycho-common/src/models/blockchain.rs
zizou0x: Yes, one entrypoint is defined only by `target` and `signature`, if two components have the exact same entrypoint, they will point to the same `EntryPointWithData`, with one `EntryPointTracingData` for each unless they use exactly the same tracing method + args

tycho-storage/migrations/2025-04-09_dci_entry_points/up.sql
zizou0x: > What I'm not sure about this design is the following issue: Component A calls EntrypointA with params X Component B calls EntrypointA with params Y.
> 
> Since we create EntryPoint <-> Component relation, ComponentA and ComponentB would be tied to the same entrypoint.
> 
> Assuming that different call parameters return differnet accounts & slots, we can't differentiate the accounts that are relevant to ComponentA from the ones that are used by ComponentB.
> 
> So, if we have a Protocol with many Components with one EntryPoint and the client only requests only one Component, we would still be returning ALL the accounts, most of which are irrelevant to him.

Yes this is true but I don't see this as an issue, I feel like each `entry_point_tracing_data` added to an entry point will just make it better because it means that this `entry_point` can call all the accounts somehow, we just haven't detected it with the first `entry_point_tracing_data` and the second one helped. You're correct that it might lead to some unnecessary data being attached to a component but I expect it to be quite small given that we will only index a few slots, we're already sending a ton of data that isn't actually used during simulation.

We can for sure decide to attach one `traced_entry_point` per `entry_point_tracing_data`, and link `entry_point_tracing_data` to `protocol_component`, that way each component would have its trace. I think this would make the design easier, it's really a design decision, let's discuss this today

tycho-common/src/models/blockchain.rs
louise-poole: If `EntryPoint` and `EntryPointWithData` are 1:1 then why not merge them into one struct? What's the benefit of keeping them separate?
EDIT: chatted to Flo directly. This is done so that EntryPoint can be returned with `TracedEntryPoint` without repeating the args data there unnecessarily.

tycho-common/src/models/blockchain.rs
```diff
@@ -342,32 +342,89 @@ pub enum BlockTag {
...
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct RPCTracerEntryPoint {
-    /// The address of the contract to trace.
-    pub target: Address,
-    // The args are a list of tuples, where the first element is the address of the caller, and the
-    // second element is the data of the call.
-    pub args: Vec<(Option<Address>, Bytes)>,
+    pub caller: Option<Address>,
+    pub data: Bytes,
 }
```
louise-poole: Also I'm not sure EntryPointTracingData needs to be a whole struct... it isn't adding functionality from what I see. We could just define it as a type and use a HashMap or Vec of tuples?
EDIT: I guess having a designated struct makes docs a bit nicer though...

tycho-storage/migrations/2025-04-09_dci_entry_points/up.sql
kayibal: I think in the DCI we still want to maintain a merged version of the entrypoint traces? 

The only use case that I see for figuring out what belongs to what, is for debugging mainly. E.g. if some new component all of a sudden needs a lot of data. In this case, we'd probably delete the EP and retrace it again manually with the non problematic data. This could def. cause some downtime but adding what is suggested here by Thales wouldn't avoid the downtime it would as best make it quicker to resolve.

tycho-storage/src/postgres/entry_point.rs
zizou0x: Good question, I actually found it out while implementing this PR: when we import these things it makes it impossible to use any of the column names as variable (because they are taken by a structs). So usually import in the function scope only to make it easier, else we would most likely have a ton of conflicts with generic names like `id`

tycho-storage/src/postgres/schema.patch
```diff
@@ -1,6 +1,6 @@
---- schema_old.rs	2025-04-14 18:43:16
-+++ schema.rs	2025-04-14 18:43:03
-@@ -65,6 +65,34 @@
+--- schema_old.rs	2025-04-17 23:16:57
```
zizou0x: We should probably group partitioned table changes together to avoid conflicts with future updates (probably at the beginning of the file?). Currently it's sorted by alphabetical order so it makes it easy to create new conflicts

tycho-common/src/models/blockchain.rs
```diff
@@ -342,32 +342,89 @@ pub enum BlockTag {
...
     Number(u64),
 }
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
+pub struct EntryPoint {
+    /// The address of the contract to trace.
+    pub target: Address,
+    /// The signature of the function to trace.
+    pub signature: String,
+}
```
louise-poole: Would like an EntryPoint id of sorts here. This will make it easier on the substreams and client side to link components to entrypoints. Idea is to either use a hash or just combine {target}:{signature} (more readable).

tycho-storage/src/postgres/schema.patch
zizou0x: Addressed

tycho-common/src/models/blockchain.rs
zizou0x: I didn't get this, maybe it's outdated?

tycho-common/src/models/blockchain.rs
```diff
@@ -341,6 +342,114 @@ pub enum BlockTag {
...
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct TracingResult {
+    /// A set of (address, storage slot) pairs representing state that contain a called address.
+    /// If any of these storage slots change, the execution path might change.
+    pub retriggers: HashSet<(Address, StoreKey)>,
+    /// A set of all addresses that were called during the trace.
+    pub called_addresses: HashSet<Address>,
```
tvinagre, line 418: As discussed, would be nice to return an Option<HashSet<StoreKey>> with each Address - let's do on another PR

tycho-common/src/storage.rs
```diff
@@ -479,6 +481,44 @@ pub trait ProtocolGateway {
...
+    ) -> Result<Vec<EntryPointWithData>, StorageError>;
+
+    /// Upserts a list of traced entry points into the database.
+    async fn upsert_traced_entry_points(
+        &self,
+        traced_entry_points: &[TracedEntryPoint],
+    ) -> Result<(), StorageError>;
+
+    /// Retrieves the tracing results for an entry point from the database.
+    async fn get_traced_entry_point(
```
tvinagre: We also need:

async fn get_traced_entry_point_with_data (
    &self,
    entry_point: EntryPointWithData,
) -> Result<TracingResult>, StorageError>;

tycho-common/src/models/blockchain.rs
louise-poole: Can you explain this? Why does it need to be option? Can't we just have an empty HashSet if no addresses are detected?

tycho-common/src/models/blockchain.rs
```diff
@@ -341,6 +342,114 @@ pub enum BlockTag {
...
+impl EntryPoint {
+    pub fn new(target: Address, signature: String) -> Self {
+        Self { target, signature }
+    }
+
+    /// Returns the external id of the entry point, which is a combination of the target address and
+    /// the function signature.
+    pub fn external_id(&self) -> String {
+        format!("{}:{}", self.target, self.signature)
+    }
```
louise-poole: So the entrypoint id is going to be used on the substreams level too... it will be used to link the entrypoint to the component + args and this link/id needs to persist beyond the block the entrypoint was first added as new components could use the same entrypoint created at the start. I would prefer we only define the rules to generate this ID once to ensure constancy and avoid information leakage (hehe... red flag coming in handy!). So the substreams side should have a util to generate this ID and inside the indexing we just consume and use the already created ID.

tycho-common/src/models/blockchain.rs
```diff
@@ -341,6 +342,114 @@ pub enum BlockTag {
...
+/// An entry point to trace. Different types of entry points tracing will be supported in the
+/// future. Like RPC debug tracing, symbolic execution, etc.
+pub enum EntryPointTracingData {
+    /// Uses RPC calls to retrieve the called addresses and retriggers
+    RPCTracer(RPCTracerEntryPoint),
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Eq, Hash)]
+pub struct RPCTracerEntryPoint {
+    /// The caller address of the transaction, will use the vm default if not provided
```
louise-poole: What is the vm default?

tycho-common/src/models/blockchain.rs
```diff
@@ -341,6 +342,114 @@ pub enum BlockTag {
...
+        let mut state = serializer.serialize_struct("RPCTracerEntryPoint", 2)?;
+        state.serialize_field("caller", &self.caller)?;
+        state.serialize_field("data", &self.data)?;
+        state.end()
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct TracingResult {
+    /// A set of (address, storage slot) pairs representing state that contain a called address.
```
louise-poole, line 414: It's not limited to a state that contains a called address though no? What happens in the case where my contract has a flag that indicates if I'm using the primary or secondary oracle... the primary and secondary oracle addresses never change, but the flag updating would change which one is used and could impact our tracing results.

tycho-common/src/models/mod.rs
```diff
@@ -51,6 +51,9 @@ pub type AccountToContractStore = HashMap<Address, ContractStore>;
 /// Component id literal type to uniquely identify a component.
 pub type ComponentId = String;
 
+/// Protocol system literal type to uniquely identify a protocol system.
+pub type ProtocolSystem = String;
```
louise-poole, line 55: ❤️

tycho-common/src/storage.rs
```diff
@@ -479,6 +481,44 @@ pub trait ProtocolGateway {
...
 }
+pub struct EntryPointFilter {
+    pub protocol_system: Option<ProtocolSystem>,
+}
+
+impl EntryPointFilter {
+    pub fn new(protocol: Option<String>) -> Self {
+        Self { protocol_system: protocol }
+    }
+}
```
louise-poole:
```suggestion
impl EntryPointFilter {
    pub fn new(protocol: Option<ProtocolSystem>) -> Self {
        Self { protocol_system: protocol }
    }
}
```

tycho-common/src/storage.rs
```diff
@@ -479,6 +481,44 @@ pub trait ProtocolGateway {
...
+
+// Trait for entry point gateway operations.
+#[async_trait]
+pub trait EntryPointGateway {
+    /// Upserts a list of entry points into the database.
+    async fn upsert_entry_points(
+        &self,
+        entry_points: &[EntryPointWithData],
+        component_id: &str,
+    ) -> Result<(), StorageError>;
```
louise-poole: This feels mismatched... the name and doc string suggest it's upserting EntryPoints, but then takes EntryPointWithData?

tycho-common/src/traits.rs
```diff
@@ -93,3 +93,28 @@ pub trait TokenPreProcessor: Send + Sync {
         block: BlockTag,
     ) -> Vec<CurrencyToken>;
 }
+
+/// Traces smart contract entry points to analyze their execution behavior and state changes.
```
louise-poole: What specifically do you mean by 'state changes'? The tracer just analyses what contracts are touched and their retriggers no? 'state changes' indicates to me that it monitors contract changes which is misleading here.

tycho-common/src/traits.rs
```diff
@@ -93,3 +93,28 @@ pub trait TokenPreProcessor: Send + Sync {
...
+    /// # Parameters
+    /// * `block_hash` - The hash of the block at which to perform the trace. The trace will use the
+    ///   state of the blockchain at this block.
+    /// * `entry_points` - A list of entry points to trace with their data.
+    ///
+    /// # Returns
+    /// Returns a vector of `TracedEntryPoint`, where each element contains:
+    /// * `retriggers` - A set of (address, storage slot) pairs representing state that could affect
+    ///   future executions. If any of these storage slots change, the set of called contract might
+    ///   be outdated.
```
louise-poole: Not a fan of the use of the word 'state' here (it has other connotations throughout the indexer).
```suggestion
    /// * `retriggers` - A set of (address, storage slot) pairs representing storage that could alter
    ///   tracing results. If any of these storage slots change, the set of called contract might
    ///   be outdated.
```

tycho-ethereum/src/entrypoint_tracer/tracer.rs
```diff
@@ -0,0 +1,277 @@
...
+use crate::{BytesCodec, RPCError};
+
+struct EVMEntrypointService {
+    provider: Provider<Http>,
+}
+
+impl EVMEntrypointService {
+    pub fn new_from_url(rpc_url: &str) -> Self {
+        Self {
+            provider: Provider::<Http>::try_from(rpc_url).expect("Error creating HTTP provider"),
```
louise-poole: Would prefer not to panic here. Maybe we can error with `RPCError::SetupError`?

tycho-ethereum/src/entrypoint_tracer/tracer.rs
```diff
@@ -0,0 +1,277 @@
...
+                            &block_hash,
+                            GethDebugBuiltInTracerType::CallTracer,
+                        )
+                        .await?;
+
+                    let called_addresses =
+                        if let GethTrace::Known(GethTraceFrame::CallTracer(frame)) = call_trace {
+                            flatten_calls(&frame)
+                        } else {
+                            return Err(RPCError::UnknownError("CallTracer failed".to_string()));
```
louise-poole: This feels super generic... is there a way we can add more detail to this error to make future debugging easier?

tycho-ethereum/src/entrypoint_tracer/tracer.rs
```diff
@@ -0,0 +1,277 @@
...
+                                                (*slot).to_bytes(),
+                                            ));
+                                        }
+                                    }
+                                }
+                            }
+                        }
+                        retriggers
+                    } else {
+                        return Err(RPCError::UnknownError("PreStateTracer failed".to_string()));
```
louise-poole: Same here... more info would be better (if possible)

tycho-common/src/models/blockchain.rs
zizou0x: He means something like this:
```
pub accessed_storage_slots: HashMap<Address,HashSet<SlotKey>>,
```

Instead of just Address, we will return a HashSet of slot keys that were accessed during the tracing

zizou0x: Then we can get the set of addresses with `accessed_storage_slots.keys()`

tycho-storage/migrations/2025-04-09_dci_entry_points/up.sql
```diff
@@ -0,0 +1,45 @@
...
+);
+
+CREATE TABLE IF NOT EXISTS "protocol_component_holds_entry_point_tracing_data"(
+    "protocol_component_id" bigint REFERENCES "protocol_component"(id) ON DELETE CASCADE NOT NULL,
+    "entry_point_tracing_data_id" bigint REFERENCES "entry_point_tracing_data"(id) ON DELETE CASCADE NOT NULL,
+    PRIMARY KEY ("protocol_component_id", "entry_point_tracing_data_id")
+);
+
+CREATE TABLE IF NOT EXISTS "entry_point_tracing_result"(
+    "entry_point_tracing_data_id" bigint NOT NULL REFERENCES "entry_point_tracing_data"(id) ON DELETE CASCADE PRIMARY KEY,
```
louise-poole: Hmmm... I think this design here might shoot us in the foot if we ever decide to add versioning. Because then `entry_point_tracing_data` and `entry_point_tracing_result` would not be strictly 1:1. Also what happens with retriggers? Do we just overwrite the result entry on a retriggered result? Or store a new entry that references the new block it was analysed on?
Most robust approach would be to just have an "id" column for the primary key.

tycho-storage/src/postgres/cache.rs
```diff
@@ -26,14 +28,14 @@ use tycho_common::{
...
-        ExtractionStateGateway, Gateway, ProtocolGateway, StorageError, Version, WithTotal,
+        BlockIdentifier, BlockOrTimestamp, ChainGateway, ContractStateGateway, EntryPointFilter,
+        EntryPointGateway, ExtractionStateGateway, Gateway, ProtocolGateway, StorageError, Version,
+        WithTotal,
     },
     Bytes,
 };
 
 use super::{PostgresError, PostgresGateway};
-
```
louise-poole: This line is vital to my sanity. I demand its return 😂

tycho-storage/src/postgres/cache.rs
```diff
@@ -60,6 +62,12 @@ pub(crate) enum WriteOp {
     InsertComponentBalances(Vec<models::protocol::ComponentBalance>),
     // Simply merge
     UpsertProtocolState(Vec<(TxHash, models::protocol::ProtocolComponentStateDelta)>),
+    // Simply merge
+    #[allow(unused)] //TODO: Remove this once we have usage in extractors
+    UpsertEntryPoints((Vec<models::blockchain::EntryPointWithData>, models::ComponentId)),
+    // Simply merge
+    #[allow(unused)] //TODO: Remove this once we have usage in extractors
+    UpsertTracedEntryPoints(Vec<models::blockchain::TracedEntryPoint>),
```
louise-poole: There are used no?

louise-poole: You also need to add these to `add_operation` no?

tycho-storage/src/postgres/entry_point.rs
```diff
@@ -0,0 +1,546 @@
...
+            .map_err(PostgresError::from)?
+            .into_iter()
+            .map(|(pc_id, ext_id)| (ext_id, pc_id))
+            .collect::<HashMap<_, _>>();
+
+        let pc_id = pc_ids
+            .get(component_id)
+            .ok_or_else(|| {
+                StorageError::NotFound("ProtocolComponent".to_string(), component_id.to_string())
+            })?;
```
louise-poole: The `pc_ids` will only have 1 id in it (if the component exists) so you don't need all this overhead of making and using a hashmap. 
Unless you have it like this in preparation for handling multiple components + entrypoints at once? Which I don't think is necessary.

tycho-storage/src/postgres/entry_point.rs
```diff
@@ -0,0 +1,546 @@
...
+
+        let new_entry_points = entry_points
+            .iter()
+            .map(|ep| NewEntryPoint {
+                external_id: ep.entry_point.external_id(),
+                target: ep.entry_point.target.clone(),
+                signature: ep.entry_point.signature.clone(),
+            })
```
louise-poole: We should collect only unique entrypoints here, no? We'll often have the case where 1 entry point will be passed multiple times with different data so we'll reduce conflicts/overhead if we collect a HashSet first.

tycho-storage/src/postgres/entry_point.rs
```diff
@@ -0,0 +1,546 @@
...
+    },
+    storage::{EntryPointFilter, StorageError},
+    Bytes,
+};
+
+use super::{
+    orm::{
+        EntryPoint as ORMEntryPoint, EntryPointTracingData as ORMEntryPointTracingData,
+        EntryPointTracingResult, NewEntryPoint, ProtocolComponent, ProtocolSystem,
+    },
```
louise-poole: IIRC on the other gateway files we simple import just `orm` here and do `orm::EntryPoint` to access the various structs inside orm. This makes it obvious that you're using the orm version of the struct without you needing to give the imports aliases.

tycho-storage/src/postgres/entry_point.rs
```diff
@@ -0,0 +1,546 @@
...
+            .map_err(|e| storage_error_from_diesel(e, "EntryPoint", "Batch upsert", None))?;
+
+        // Fetch entry points by their external_ids, we can't use .returning() on the insert above
+        // because it doesn't return the ids on conflicts.
+        let input_external_ids: Vec<String> = entry_points
+            .iter()
+            .map(|ep| ep.entry_point.external_id())
+            .collect();
+
+        let entry_point_ids = ORMEntryPoint::ids_by_external_ids(&input_external_ids, conn).await?;
```
louise-poole: I did some thinking here and I think it's much better to use the 'on conflict do update' with a redundant update just so that we can get it to return the ids we need. This has a lot less overhead than using 2 separate queries.

tycho-storage/src/postgres/entry_point.rs
```diff
@@ -0,0 +1,546 @@
...
+        // because it doesn't return the ids on conflicts.
+        diesel::insert_into(entry_point_tracing_data)
+            .values(&new_tracing_data)
+            .on_conflict_do_nothing()
+            .execute(conn)
+            .await
+            .map_err(|e| storage_error_from_diesel(e, "EntryPointData", "Batch upsert", None))?;
+
+        let data_ids =
+            ORMEntryPointTracingData::ids_by_entry_point_with_data(entry_points, conn).await?;
```
louise-poole: Same here

tycho-storage/src/postgres/entry_point.rs
```diff
@@ -0,0 +1,546 @@
...
+            .into_values()
+            .map(|data_id| NewProtocolComponentHoldsEntryPointTracingData {
+                protocol_component_id: *pc_id,
+                entry_point_tracing_data_id: data_id,
+            })
+            .collect::<Vec<_>>();
+
+        diesel::insert_into(protocol_component_holds_entry_point_tracing_data)
+            .values(&pc_links)
+            .on_conflict_do_nothing() //TODO: Do we need to delete every previously inserted links?
```
louise-poole: I assume for now we're treating this as accumulative. Removed links is enough of an edge case to not need to worry about it yet (and should probably be handled in a 'delete' gateway fn).

tycho-storage/src/postgres/entry_point.rs
```diff
@@ -0,0 +1,546 @@
...
+use crate::postgres::{
+    orm::{
+        EntryPointTracingType, NewEntryPointTracingData, NewEntryPointTracingDataCallsAccount,
+        NewEntryPointTracingResult, NewProtocolComponentHoldsEntryPointTracingData,
+    },
+    PostgresError,
+};
+
+impl PostgresGateway {
+    pub(crate) async fn upsert_entry_points(
```
louise-poole: This is more of an `insert` fn than an `upsert` fn no? We insert new entrypoints + data and skip those that already exist (no updates of any entries are actually done).

tycho-common/src/storage.rs
louise-poole: Also is this struct necessary? It seems a bit shallow right now?

tycho-storage/src/postgres/entry_point.rs
```diff
@@ -0,0 +1,546 @@
...
+                    None,
+                )
+            })?;
+
+        Ok(())
+    }
+
+    pub(crate) async fn get_entry_points_with_data(
+        &self,
+        filter: EntryPointFilter,
```
louise-poole, line 194: I think we should just have protocol_system as a non-optional param here.

tycho-storage/src/postgres/entry_point.rs
```diff
@@ -0,0 +1,546 @@
...
+        conn: &mut AsyncPgConnection,
+    ) -> Result<Vec<EntryPointWithData>, StorageError> {
+        use schema::{
+            entry_point::dsl::*, entry_point_tracing_data as eptd, protocol_component as pc,
+            protocol_component_holds_entry_point_tracing_data as pchep,
+        };
+
+        let orm_ep = if let Some(ref protocol) = filter.protocol_system {
+            let ps_id = ProtocolSystem::id_by_name(protocol, conn).await?;
```
louise-poole: There is a PostgresGateway fn called `get_protocol_system_id` that uses the `protocol_system_id_cache` to get the db id of the protocol system. Rather use that cache here and reduce the need for another db query.

tycho-storage/src/postgres/entry_point.rs
```diff
@@ -0,0 +1,546 @@
...
+            .iter()
+            .map(|ep| ep.id)
+            .collect::<Vec<_>>();
+
+        let orm_ep_data = schema::entry_point_tracing_data::table
+            .filter(schema::entry_point_tracing_data::entry_point_id.eq_any(ep_ids))
+            .select(ORMEntryPointTracingData::as_select())
+            .load::<ORMEntryPointTracingData>(conn)
+            .await
+            .map_err(|err| storage_error_from_diesel(err, "EntryPointData", "None", None))?;
```
louise-poole: What the reasoning behind using 2 queries to fetch everything? Have you played around with if it's more performant to just add 1 more join to the big query above or if it really is better to use the 2 separate queries? If the table is small and well indexed we likely will do better with 1 big query.

tycho-storage/src/postgres/entry_point.rs
```diff
@@ -0,0 +1,546 @@
...
+            .await
+            .map_err(PostgresError::from)?;
+        let block_id_map: HashMap<_, _> = blocks.into_iter().collect();
+
+        let mut values = Vec::with_capacity(traced_entry_points.len());
+        let mut data_ids = Vec::with_capacity(traced_entry_points.len());
+        for tep in traced_entry_points {
+            let data_id =
+                ORMEntryPointTracingData::id_from_entry_point_with_data(&tep.entry_point, conn)
+                    .await?;
```
louise-poole: This is a lot of individual db queries... why not use `ids_by_entry_point_with_data`?

tycho-storage/src/postgres/orm.rs
```diff
@@ -1740,3 +1760,304 @@ impl ComponentTVL {
...
+}
+
+#[derive(Insertable, AsChangeset)]
+#[diesel(table_name = entry_point_tracing_result)]
+#[diesel(check_for_backend(diesel::pg::Pg))]
+pub struct NewEntryPointTracingResult {
+    pub entry_point_tracing_data_id: i64,
+    pub detection_block: i64,
+    pub detection_data: serde_json::Value,
+    pub modified_ts: Option<NaiveDateTime>,
```
louise-poole: It's unnecessary to have this here and manually need to set it. We have a db function that updates this column for us called `update_modified_column` that can be set with a trigger for the tables that have the `modified_ts` column like so:
```
CREATE TRIGGER update_modtime_entrypoint_tracing_result
    BEFORE UPDATE ON entrypoint_tracing_result
    FOR EACH ROW
    EXECUTE PROCEDURE update_modified_column();
```

tycho-storage/src/postgres/entry_point.rs
```diff
@@ -0,0 +1,546 @@
...
+                new_entry_point_calls_account.push(NewEntryPointTracingDataCallsAccount {
+                    entry_point_tracing_data_id: data_id,
+                    account_id: *acc_id,
+                });
+            }
+        }
+
+        diesel::insert_into(entry_point_tracing_data_calls_account)
+            .values(&new_entry_point_calls_account)
+            .on_conflict_do_nothing() // TODO: Do we need to delete every previously inserted links?
```
louise-poole: No, for now let's leave deletes. This is only necessary on a retrigger of an entrypoint analysis and those should be rare enough to start. We'll need a separate gateway fn to remove/mark invalid the outdated links and not handle it here.

tycho-storage/src/postgres/orm.rs
```diff
@@ -290,6 +296,20 @@ pub struct ProtocolSystem {
...
+    ) -> Result<i64, StorageError> {
+        protocol_system::table
+            .filter(protocol_system::name.eq(name))
+            .select(protocol_system::id)
+            .first::<i64>(conn)
+            .await
+            .map_err(|e| StorageError::from(PostgresError::from(e)))
+    }
+}
+
```
louise-poole: This is not needed - the PostgresGateway populates a cache with this data and that cache should be used. I would remove this to discourage the temptation of querying the db for already cached data.

tycho-storage/src/postgres/orm.rs
```diff
@@ -1740,3 +1760,304 @@ impl ComponentTVL {
...
+            EntryPointTracingDataCommon::RPCTracer(rpc_tracer) => serde_json::to_value(rpc_tracer)
+                .map_err(|e| {
+                    StorageError::Unexpected(format!(
+                        "Failed to serialize RPCTracerEntryPoint: {}",
+                        e
+                    ))
+                })?,
+        };
+
+        let id_ = super::schema::entry_point_tracing_data::table
```
louise-poole, line 1889: You imported this schema on the file level no? The `super::schema::` qualifier on every line makes it harder to read.

tycho-storage/src/postgres/orm.rs
```diff
@@ -1740,3 +1760,304 @@ impl ComponentTVL {
...
+            .first::<i64>(conn)
+            .await
+            .map_err(PostgresError::from)?;
+
+        Ok(id_)
+    }
+
+    // Retrieves the database ids of entry point tracing data from a list of entry points with data.
+    pub(crate) async fn ids_by_entry_point_with_data(
+        entry_point_with_data: &[EntryPointWithDataCommon],
```
louise-poole: This should be plural no?

tycho-storage/migrations/2025-04-09_dci_entry_points/up.sql
```diff
@@ -0,0 +1,45 @@
...
+);
+
+CREATE TABLE IF NOT EXISTS "entry_point"(
+    "id" bigserial PRIMARY KEY,
+    "external_id" text UNIQUE NOT NULL,
+    "target" bytea NOT NULL,
+    "signature" text NOT NULL,
+    "inserted_ts" timestamptz NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "modified_ts" timestamptz NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    UNIQUE ("target", "signature")
```
louise-poole, line 13: "external_id" should be unique too?

tycho-storage/src/postgres/orm.rs
```diff
@@ -1740,3 +1760,304 @@ impl ComponentTVL {
...
+                let tracing_data = match tracing_type {
+                    EntryPointTracingType::RpcTracer => {
+                        let data = serde_json::from_value(data.clone().unwrap()).unwrap();
+                        EntryPointTracingDataCommon::RPCTracer(data)
+                    }
+                };
+                tuple_ids.contains(&(ext_id, tracing_type, &tracing_data))
+            })
+            .map(|(_, target, signature, tracing_type, data, id)| {
+                dbg!(&data);
```
louise-poole: Don't forget to remove this

tycho-storage/src/postgres/orm.rs
```diff
@@ -1740,3 +1760,304 @@ impl ComponentTVL {
...
+        if entry_point_with_data.is_empty() {
+            return Ok(HashMap::new());
+        }
+
+        let (external_ids, data): (HashSet<_>, HashSet<_>) = entry_point_with_data
+            .iter()
+            .map(|ep| {
+                let data = match &ep.data {
+                    EntryPointTracingDataCommon::RPCTracer(rpc_tracer) => {
+                        serde_json::to_value(rpc_tracer).unwrap()
```
louise-poole: There are quite a few unwraps in this fn... can we raise an error instead please?

tycho-storage/src/postgres/orm.rs
```diff
@@ -1740,3 +1760,304 @@ impl ComponentTVL {
...
+            tuple_data
+                .iter()
+                .map(|(ext_id, tracing_type, data)| (ext_id, tracing_type, data))
+                .collect::<Vec<_>>();
+
+        let res = super::schema::entry_point_tracing_data::table
+            .inner_join(super::schema::entry_point::table)
+            .filter(
+                // Not filtered by tracing type eq_any because of a diesel bug with OID, shouldn't
+                // affect results much because of data filtering and the application side filterin
```
louise-poole:
```suggestion
                // affect results much because of data filtering and the application side filtering
```

tycho-storage/src/postgres/orm.rs
```diff
@@ -1740,3 +1760,304 @@ impl ComponentTVL {
...
+            .into_iter()
+            .filter(|(ext_id, _, _, tracing_type, data, _)| {
+                let tracing_data = match tracing_type {
+                    EntryPointTracingType::RpcTracer => {
+                        let data = serde_json::from_value(data.clone().unwrap()).unwrap();
+                        EntryPointTracingDataCommon::RPCTracer(data)
+                    }
+                };
+                tuple_ids.contains(&(ext_id, tracing_type, &tracing_data))
+            })
```
louise-poole: I think we can filter within the map no? Then we don't need to do 2 passes of the iterator.

tycho-storage/src/postgres/orm.rs
```diff
@@ -1740,3 +1760,304 @@ impl ComponentTVL {
...
+                        ep.data.clone(),
+                    )
+                })
+                .collect::<Vec<_>>();
+
+        let tuple_ids: Vec<(&String, &EntryPointTracingType, &EntryPointTracingDataCommon)> =
+            tuple_data
+                .iter()
+                .map(|(ext_id, tracing_type, data)| (ext_id, tracing_type, data))
+                .collect::<Vec<_>>();
```
louise-poole: Idea: we could build this as a map where the key is this tuple and the value is the `EntryPointWithDataCommon` object. Then when we filter things out after the query we can just do a map lookup: if the key doesn't exist we skip, if it does then we already have the `EntryPointWithDataCommon` and don't need to make a new one. With that we can also drop returning target and signature.

tycho-common/src/storage.rs
zizou0x: As discussed on slack, this is not needed in the end

tycho-common/src/models/blockchain.rs
zizou0x: I agree with the idea of having it defined it once but I didn't get the suggestion

tycho-common/src/models/blockchain.rs
zizou0x: If your contract has a flag, it's most likely pointing to a storage slot that has been read during the execution. If it's a flag pointing to hardcoded addresses in the bytecode, it's static by definition so we know it will never change.

The only case we don't support now is a calldata lookup, but this is out of scope for retriggers because its defined by the caller

tycho-common/src/storage.rs
zizou0x: This struct allows us to add more filters capabilities without breaking backward compatibility later. Ideally I need to rename the new function to `new_with_protocol_system`

tycho-common/src/traits.rs
zizou0x: What you described is the definition for the `trace()` function of this trait. The trait itself is a more generic entity (we can decide later that we want more function). It doesn't make sense to repeat the same doc twice, I will remove the misleading part though

tycho-common/src/traits.rs
zizou0x: A tuple of contract+storage slots is more a chain state than storage IMO so I've kept "state" but I applied the second part of your suggestion

tycho-storage/src/postgres/entry_point.rs
zizou0x: I don't think it's a good practice, we can still decide to do it but first read [this](https://stackoverflow.com/a/42217872). I think we might fall into the "not safe" case because we do have many conflict target and we will most likely have triggers for the `modified_ts`.

We have a way out of this which would be to remove the `modified_ts` column (because they should be static) and say that all conflict targets are the same (basically a sort of id). But I'm not sure at this point if it's worth the risk. I agree it would be less overhead but most likely the performance benefit would be tiny in the whole scope, and entrypoint inserts will probably not happen so much (like protocol component inserts now)

tycho-storage/src/postgres/entry_point.rs
zizou0x: Answered above

tycho-storage/src/postgres/entry_point.rs
zizou0x: Answered above

tycho-storage/src/postgres/entry_point.rs
zizou0x: Nice

tycho-common/src/models/blockchain.rs
louise-poole: The suggestion is to remove the logic here that creates the external id and just have a struct field that holds the external id. The substream should set the id, not the indexer 👀

tycho-common/src/traits.rs
louise-poole: Throughout the code 'state' represents a set of versioned values (emphasis on the fact that it's representing the values themselves). However, here we're using 'state' to represent the location of a value and not the value itself.

tycho-storage/src/postgres/entry_point.rs
louise-poole: Ah yes - currently there are no triggers but indeed once you add them for the `modified_ts` column this will cause an issue. Cool cool, let's keep it as is.

tycho-storage/migrations/2025-04-09_dci_entry_points/up.sql
zizou0x: Is is unique, see the line where it's defined

tycho-storage/src/postgres/entry_point.rs
louise-poole: I like their suggestion of the nested insert. We could use that to make this one query?

tycho-common/src/storage.rs
louise-poole: Hmmm... these gateway fns are used internally for the most part so I wouldn't worry too much about backward compatibility. Though I do like this approach and wish we took it for all gateway fns. I'm bugged by the inconsistency, but accept the reasoning 👍

tycho-common/src/models/blockchain.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/eac7b4467de4354c4835c8050bb6e485e3a116e9

tycho-common/src/storage.rs
zizou0x: I've decided that ProtocolSystem would be a mandatory filter, I don't see any usecase where we won't want to filter by protocol system. I kept the same approach though, this stuct is shallow because we want to be able to adapt it without breaking backward compatibility

See https://github.com/propeller-heads/tycho-indexer/pull/551/commits/049a0d22b9f7c382351fa129fe827beb999ba67a

tycho-common/src/storage.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/eac7b4467de4354c4835c8050bb6e485e3a116e9

tycho-common/src/traits.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/eac7b4467de4354c4835c8050bb6e485e3a116e9

tycho-common/src/traits.rs
zizou0x: Fair, but "storage" doesn't help with this IMO. Maybe it should be "state slots" or "keys"?

tycho-ethereum/src/entrypoint_tracer/tracer.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/eac7b4467de4354c4835c8050bb6e485e3a116e9

We don't know much more about what caused the error, it just means that we received something unexpected from the node

tycho-ethereum/src/entrypoint_tracer/tracer.rs
zizou0x: Same reply as above

tycho-storage/src/postgres/cache.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/eac7b4467de4354c4835c8050bb6e485e3a116e9

tycho-storage/src/postgres/orm.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/34ad7749deec05090cbfe5b74eddb3498488e4ab

tycho-storage/src/postgres/orm.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/049a0d22b9f7c382351fa129fe827beb999ba67a

tycho-storage/src/postgres/orm.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/049a0d22b9f7c382351fa129fe827beb999ba67a

tycho-storage/src/postgres/orm.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/049a0d22b9f7c382351fa129fe827beb999ba67a

tycho-storage/src/postgres/orm.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/049a0d22b9f7c382351fa129fe827beb999ba67a

tycho-storage/src/postgres/orm.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/049a0d22b9f7c382351fa129fe827beb999ba67a

tycho-storage/src/postgres/orm.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/049a0d22b9f7c382351fa129fe827beb999ba67a

tycho-storage/src/postgres/orm.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/049a0d22b9f7c382351fa129fe827beb999ba67a

tycho-storage/src/postgres/orm.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/049a0d22b9f7c382351fa129fe827beb999ba67a

tycho-storage/src/postgres/entry_point.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/049a0d22b9f7c382351fa129fe827beb999ba67a

tycho-storage/src/postgres/entry_point.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/049a0d22b9f7c382351fa129fe827beb999ba67a

tycho-storage/src/postgres/entry_point.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/049a0d22b9f7c382351fa129fe827beb999ba67a

tycho-storage/src/postgres/entry_point.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/049a0d22b9f7c382351fa129fe827beb999ba67a

tycho-storage/src/postgres/entry_point.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/049a0d22b9f7c382351fa129fe827beb999ba67a

tycho-storage/src/postgres/entry_point.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/049a0d22b9f7c382351fa129fe827beb999ba67a

tycho-storage/src/postgres/entry_point.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/049a0d22b9f7c382351fa129fe827beb999ba67a

tycho-storage/src/postgres/entry_point.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/049a0d22b9f7c382351fa129fe827beb999ba67a

tycho-ethereum/src/entrypoint_tracer/tracer.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/049a0d22b9f7c382351fa129fe827beb999ba67a

tycho-storage/src/postgres/entry_point.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/049a0d22b9f7c382351fa129fe827beb999ba67a

tycho-common/src/traits.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/2e16f4ef5c2f45b4a3fb671989572fbb505feed2

tycho-common/src/models/blockchain.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/6f7aef24b04adbedee3192294c93fb7136f1171d

tycho-storage/migrations/2025-04-09_dci_entry_points/up.sql
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/8e9f54493c3c7ab923ca106e909bd38b9758a126

tycho-storage/src/postgres/cache.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-indexer/pull/551/commits/0c49fc93f8b19e2c23be9b14958d81199023c4e6

tycho-storage/src/postgres/entry_point.rs
zizou0x: I don't mind looking into it later but I'm scared it will make the code overly complex for such a small improvement. Do you we have an example of nested inserts with Diesel in our codebase?

tycho-storage/src/postgres/entry_point.rs
```diff
@@ -0,0 +1,577 @@
...
+            .execute(conn)
+            .await
+            .map_err(|err| {
+                storage_error_from_diesel(err, "EntryPointPointsToAccount", "Batch upsert", None)
+            })?;
+
+        Ok(())
+    }
+
+    pub(crate) async fn get_traced_entry_point(
```
louise-poole: Ideally we should have a gateway function where we can specify the components/protocol system and get back all those components linked to their relevant detected contract addresses. The client needs this on the RPC and ideally we have a function that can do this all at once instead of needing the RPC to use and loop through multiple gateway calls. That gateway function should support optional pagination (needed by the RPC).

tycho-storage/src/postgres/entry_point.rs
```diff
@@ -0,0 +1,577 @@
...
+        NewProtocolComponentHoldsEntryPointTracingData,
+    },
+    schema::{self},
+    storage_error_from_diesel, PostgresError, PostgresGateway,
+};
+
+impl PostgresGateway {
+    pub(crate) async fn insert_entry_points(
+        &self,
+        new: &[(&str, &Vec<EntryPointWithData>)],
```
louise-poole: Can you document this? What is `new`? I assume it is tuples of components with their entrypoint datas... why tuples and not a HashMap? It's highly likely whatever calls this gateway function will already have this data as a Hashmap.

tycho-storage/src/postgres/entry_point.rs
zizou0x: I've improved this to be able to retrieve batches of result instead of 1 by 1.

But currently this needs to be done in two queries by the RPC. First we get the entrypoints and then we get all the results for them. 

Imo, later if retrieving entrypoint results is too slow we can decide to explore and see if we can improve this, like make a single more optimized function for the RPC. But I think this needs to be done in a later iteration, let's keep it simple and minimal for now. This design makes sense given how the db is structured

tycho-storage/src/postgres/entry_point.rs
```diff
@@ -339,32 +360,66 @@ impl PostgresGateway {
...
-                    StorageError::Unexpected(format!("Failed to deserialize TracingResult: {e}"))
-                })
+            .map(|(ep_ext_id, data)| {
+                let converted_data = data
+                    .into_iter()
+                    .map(|d| {
+                        serde_json::from_value(d).map_err(|e| {
+                            StorageError::Unexpected(format!(
+                                "Failed to deserialize TracingResult: {e}"
+                            ))
```
louise-poole, line 418: This would be a StorageError::DecodeError no?

louise-poole (approved):
Looks good to me! Of course needs to be updated to reflect our latest design decisions. But I'm happy for this to be merged so long 👍

tycho-storage/src/postgres/entry_point.rs
zizou0x: Will be addressed in the next PR 👁️

PR #552, opened 2025-04-11
fix: Make Snapshot attributes pub

carloszanella:
It is impossible to create test fixtures for FeedMessage atm because the attributes of Snapshot are not public


zizou0x (approved):

pistomat (approved):
Very good PR, this core contributor @carloszanella definitely deserves some retroactive funding for their effort!

PR #553, opened 2025-04-14
chore: Update PostgreSQL schema patch to avoid conflicts on `diesel print-schema`

zizou0x:
This PR fixes a bug that was causing diesel print-schema to fail with `error applying hunk #1`

This bug was due to an outdated patch that was leading to conflicts when it was applied on top of the current migrations.

### DIFF
--- a/tycho-storage/src/postgres/schema.patch
+++ b/tycho-storage/src/postgres/schema.patch
@@ -1,38 +1,10 @@
-diff --git a/tycho-storage/src/postgres/schema.rs b/tycho-storage/src/postgres/schema.rs
-index d71a8300..6178d82a 100644
---- a/tycho-storage/src/postgres/schema.rs
-+++ b/tycho-storage/src/postgres/schema.rs
-@@ -236,21 +236,128 @@ diesel::joinable!(token -> account (account_id));
- diesel::joinable!(token_price -> token (token_id));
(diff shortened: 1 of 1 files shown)


tycho-storage/src/postgres/schema.patch
```diff
@@ -1,38 +1,10 @@
-diff --git a/tycho-storage/src/postgres/schema.rs b/tycho-storage/src/postgres/schema.rs
-index d71a8300..6178d82a 100644
---- a/tycho-storage/src/postgres/schema.rs
-+++ b/tycho-storage/src/postgres/schema.rs
-@@ -236,21 +236,128 @@ diesel::joinable!(token -> account (account_id));
- diesel::joinable!(token_price -> token (token_id));
- diesel::joinable!(transaction -> block (block_id));
+--- schema_old.rs	2025-04-14 18:43:16
++++ schema.rs	2025-04-14 18:43:03
```
louise-poole, line 2: Previously the header used a git style format... are we sure changing it is okay? I would assume so, but just double checking.

zizou0x: Yes, the previous diff was made between two commits and this one is between two files. But this metadata doesn't really matter in our case because diesel will ignore it and always tries to apply this on the output of the `print-schema` command

PR #555, opened 2025-04-17
fix(tycho-client): relax default timeout values

louise-poole:


zizou0x (approved):

tycho-client/src/stream.rs
```diff
@@ -65,12 +65,12 @@ impl TychoStreamBuilder {
...
-            Chain::ZkSync => (3, 2, 50),
-            Chain::Arbitrum => (1, 0, 100), // Typically closer to 0.25s
-            Chain::Base => (2, 2, 50),
-            Chain::Unichain => (1, 1, 100),
+            Chain::Ethereum => (12, 36, 10),
+            Chain::Starknet => (2, 8, 50),
+            Chain::ZkSync => (3, 12, 50),
+            Chain::Arbitrum => (1, 2, 100), // Typically closer to 0.25s
+            Chain::Base => (2, 12, 50),
+            Chain::Unichain => (1, 10, 100),
```
zizou0x, line 73: Why so much for Unichain?

louise-poole: Because the current start-up logic uses blocktime + max wait as a timeout for how long it waits for snapshots. So for fast chains we need to exaggerate the timeout here to give it a bit more space to finish fetching snapshots

PR #557, opened 2025-04-23
WIP: DCI interface

tvinagre:


tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -0,0 +1,348 @@
...
+
+        for entrypoint in entrypoints {
+            self.protocol_entrypoints
+                .insert(entrypoint.clone());
+
+            let traced_entrypoints = self
+                .entrypoint_gw
+                .get_traced_entry_point(entrypoint)
+                .await
+                .unwrap();
```
kayibal: This will be a very slow startup if we have one query per EP

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -0,0 +1,348 @@
...
+                self.protocol_entrypoints
+                    .get(&component_id)
+                    .get_or_insert_default()
+                    .push(entrypoint.clone());
+            }
+        }
+
+        // Select for analysis the newly detected EntryPointsWithData that haven't been analyzed
+        // yet. This filter prevents us from re-analyzing entrypoints that have already been
+        // analyzed, which can be a case if all the components have the same entrypoint.
```
kayibal: Hmm it's not clear to me why this is necessary? In this first iteration I assumed that we will:

1. Get new entrypoint
2. Analyze
3. Finish

So not sure why we have two collections? Are we already implementing some kind of retry mechanism here?

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -0,0 +1,348 @@
...
+        debug!("Will analyze {:?} entrypoints", entrypoints_to_analyze.len());
+
+        // Trace analyze all the entrypoints
+        let entrypoints_to_analyze: Vec<EntryPointWithData> = entrypoints_to_analyze.into();
+
+        let traced_entry_points = self
+            .tracer
+            .trace(block.hash.clone(), entrypoints_to_analyze)
+            .await
+            .expect("Failed to trace entrypoints");
```
kayibal: So this panics if one EP fails to analyze? So then by design I think that `protocol_entrypoints` should be always empty?

tycho-common/src/traits.rs
```diff
@@ -13,14 +13,39 @@ use crate::{
...
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// ```
+    async fn get_accounts_at_block(
         &self,
-        block: Block,
-        account_addresses: Vec<Address>,
+        block: &Block,
```
kayibal, line 48: Can we move the block into the request please? I think it will be way more flexible this way.

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -0,0 +1,348 @@
...
+        let mut aux_struct: HashMap<Address, Vec<EntryPoint>> = HashMap::new();
+        for (entrypoint_with_data, trace) in self.entrypoint_results {
+            for address in trace.tracing_result.called_addresses {
+                let entrypoint = entrypoint_with_data.entry_point.clone();
+                aux_struct
+                    .entry(address.clone())
+                    .or_insert_with(Vec::new)
+                    .push(entrypoint);
+            }
+        }
```
kayibal: This looks like we should keep this around between blocks? Else we always have a kind of huge memory allocation here on each block?

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -0,0 +1,348 @@
...
+}
+
+struct DCIBlockUpdate {
+    // A Hashmap mapping Transactions to the relevant AccountDeltas (either Create or Update)
+    // detected on the block.
+    result_deltas: HashMap<Transaction, Vec<AccountDelta>>,
+    // A Hashmap mapping component IDs to the new Accounts that were detected by the DCI on the
+    // block
+    result_components: HashMap<ComponentId, Vec<Address>>,
+}
```
kayibal: I wonder if we could just emit a regular block update here instead that is merged with the current BlockUpdate of the Extractor?

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -0,0 +1,519 @@
...
+};
+
+/// A unique identifier for a storage location, consisting of an address and a storage key.
+type StorageLocation = (Address, StoreKey);
+
+/// The index of a transaction in the transaction vector
+#[allow(dead_code)] // Clippy thinks this is dead code, but it's used below for clarity
+type TxVecIndex = usize;
+
+#[allow(unused)]
```
tvinagre: Should we add a TODO so we can remember to remove this allows?

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -0,0 +1,519 @@
...
+    #[instrument(skip_all, fields(chain = % self.chain, protocol = % self.protocol, block_number = % block_changes.block.number))]
+    pub(super) async fn process_block_update(
+        &mut self,
+        block_changes: &mut BlockChanges,
+    ) -> Result<(), ExtractionError> {
+        // Note: in the end we need to link DCI indexed accounts to a transaction, therefore we need
+        // to keep track of the transaction index in the txs_with_update vector related to any
+        // possible newly indexed accounts. This index will be represented by the TxVecIndex type.
+
+        let mut new_entrypoint_params: HashMap<String, Vec<(TxVecIndex, TracingParams)>> =
```
tvinagre: Being picky but maybe we should create a Type for `EntrypointId` and use it instead of `String` ? Just eases reading and comprehension.

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -0,0 +1,519 @@
...
+            HashMap::new();
+        for (tx_vec_idx, tx) in block_changes
```
tvinagre: I'd add a simple comment explaining this step: ""First, we extract all new entrypoints detected in a block"

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -0,0 +1,519 @@
...
+                    .contains_key(&(entrypoint_id.clone(), param.clone()))
+                {
+                    continue;
+                }
+
+                let entrypoint = self
+                    .cache
+                    .ep_id_to_entrypoint
+                    .get(entrypoint_id)
+                    .ok_or(ExtractionError::Storage(StorageError::NotFound(
```
tvinagre: Won't this fail on a cold start? It will not have the entrypointId in the cache whenever it detects a new entrypoint

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -0,0 +1,519 @@
...
+            )
+            .await
+            .map_err(ExtractionError::from)?;
+
+        self.cache.ep_id_to_entrypoint.extend(
+            entrypoints_with_params
+                .values()
+                .flat_map(|e| {
+                    e.iter()
+                        .map(|ep| (ep.entry_point.external_id.clone(), ep.entry_point.clone()))
```
tvinagre, line 119: I'm confused, external_id should be the id of the protocol component that the entry point belongs to - not an unique identifier of the entrypoint (which I think it's what we aimed to store here)

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -0,0 +1,519 @@
...
+        for (tx_vec_idx, traced_entry_point) in tx_to_traced_entry_point {
+            for account in traced_entry_point
+                .tracing_result
+                .called_addresses
+                .clone()
+            {
+                account_to_tx_idx
+                    .entry(account)
+                    .and_modify(|existing_idx| {
+                        if tx_vec_idx < *existing_idx {
```
tvinagre: we want to keep the Tx with smallest idx?

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -0,0 +1,519 @@
...
+            .iter()
+            .flat_map(|traced_entry_point| {
+                traced_entry_point
+                    .tracing_result
+                    .called_addresses
+                    .iter()
+                    .map(|address| {
+                        StorageSnapshotRequest {
+                            address: address.clone(),
+                            // TODO: Set the slots once it's available from the Tracer
```
tvinagre: We cannot forget this one

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
tvinagre: saw later the retrigger implementation - lgtm

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -0,0 +1,519 @@
...
+                        .clone(),
+                ),
+                traced_entry_point
+                    .tracing_result
+                    .clone(),
+            );
+        }
+    }
+
+    fn extract_tracked_updates(
```
tvinagre, line 507: this method could use some comments / docs

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -0,0 +1,519 @@
...
+
+                let tx_with_changes = TxWithChanges::new(
+                    tx.tx.clone(),
+                    Default::default(),
+                    account_delta,
+                    Default::default(),
+                    Default::default(),
+                    Default::default(),
+                    Default::default(),
+                    Default::default(),
```
tvinagre: ugly lol

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
zizou0x: No, `ProtocolComponent.external_id` is what you described, here we use `EntryPoint.external_id` which is a unique identifier for entry points

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
zizou0x: We already have it, I'll use it here

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
zizou0x: The cache should be loaded from the db before we run

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
zizou0x: Yes there is a thread about this on Slack, the decision was: if it's a new entrypoint, we link the update to the first transaction that pushed it. If it's a retrigger, we link the update to the last transaction that retriggered it

tycho-common/src/traits.rs
zizou0x: Currently the function `get_accounts_at_block` needs all accounts at the same block, this helps to make the interface simpler because there is no confusion about the block, we know it's always the same. If we were allowing for the block to be passed with each requests we might have some cases where a single account is requested twice at different blocks, in that case we would need the returned struct to link blocks to the account delta.

It's doable but makes the interfaces and logic more complex so I'm not sure it's worth if we don't need it

tycho-common/src/models/blockchain.rs
```diff
@@ -214,8 +214,8 @@ impl TxWithChanges {
 
     /// Merges this update with another one.
     ///
-    /// The method combines two [`TxWithChanges`] instances if they for different transactions on
-    /// the same block.
+    /// The method combines two [`TxWithChanges`] instances if they are for different transactions
+    /// on the same block.
```
louise-poole: You removed the check that required them to be for different transactions, no?
```suggestion
    /// The method combines two [`TxWithChanges`] instances if they are on the same block.
```

tycho-common/src/storage.rs
```diff
@@ -546,9 +552,13 @@ pub trait EntryPointGateway {
     /// Retrieves a map of component ids to a set of entry points with their tracing data from the
     /// database.
     ///
+    /// # Arguments
     /// * `filter` - The EntryPointFilter to apply to the query.
     /// * `pagination_params` - The pagination parameters to apply to the query, if None, all
     ///   results are returned.
+    ///
+    /// # Returns
+    /// A map of component ids to a set of entry points with their tracing data.
```
louise-poole:
```suggestion
    /// A map of component ids to a set of entry points with their tracing params.
```

tycho-common/src/traits.rs
```diff
@@ -13,14 +13,36 @@ use crate::{
     Bytes,
 };
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct StorageSnapshotRequest {
+    pub address: Address,
+    pub slots: Option<Vec<Bytes>>,
+}
```
louise-poole: Can you document this? What does a None for `slots` mean? All slots will be fetched?

tycho-ethereum/src/account_extractor/contract.rs
```diff
@@ -1,54 +1,96 @@
...
+            .iter()
+            .map(|&address| {
+                self.provider
+                    .get_code(address, block_id)
+            })
+            .collect();
+
+        // Execute all balance and code requests concurrently
+        let balances = try_join_all(balance_futures).await?;
+        let codes = try_join_all(code_futures).await?;
```
louise-poole: Hmmm... this will do all the balance futures concurrently, then only once that's complete it'll do the code futures concurrently. But not all of them at once. From your comment it sounds like you want all futures executed together. Maybe you could try:
```
let (balances, codes) = tokio::join!(
    try_join_all(balance_futures),
    try_join_all(code_futures)
);
```

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
tvinagre: I just copied the definition from the type, that said:
`    /// The id of the protocol component that the entry point belongs to. `

This just got addressed on Tam's PR (https://github.com/propeller-heads/tycho-indexer/pull/581/files#diff-0d9538699a9923ff5203214cdd9c79778aaf7030740f978ec26ec3b59ecfca40L372) 

So this matter seems resolved now :)

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -199,107 +207,172 @@ where
...
-                let entrypoint = self
-                    .cache
-                    .ep_id_to_entrypoint
+                let entrypoint = new_entrypoints
                     .get(entrypoint_id)
-                    .ok_or(ExtractionError::Storage(StorageError::NotFound(
-                        "Entrypoint".to_string(),
-                        entrypoint_id.to_string(),
-                    )))?;
+                    .or_else(|| {
```
tvinagre, line 212: Looks neater now :)

tycho-ethereum/src/account_extractor/contract.rs
```diff
@@ -138,6 +181,292 @@ impl EVMAccountExtractor {
...
+                .as_mut()
+                .await
+                .map_err(|e| {
+                    RPCError::RequestError(ProviderError::CustomError(format!(
+                        "Failed to collect balance request data: {e}",
+                    )))
+                })?;
+
+            // TODO: Check if this should be big-endian or little-endian
+            balances.insert(address.clone(), Bytes::from(balance_result.to_be_bytes::<32>()));
```
louise-poole: It's big-endian. We removed all use of little-endian encoding in the entirety of tycho.

tycho-ethereum/src/account_extractor/contract.rs
```diff
@@ -138,6 +181,292 @@ impl EVMAccountExtractor {
...
+                            storage_batch
+                                .add_call(
+                                    "eth_getStorageAt",
+                                    &(&request.address, slot, BlockNumberOrTag::from(block.number)),
+                                )
+                                .map_err(|_| {
+                                    RPCError::RequestError(ProviderError::CustomError(
+                                        "Failed to get storage".to_string(),
+                                    ))
+                                })?
```
louise-poole: Might be good to add the actual error to the message?

tycho-ethereum/src/account_extractor/contract.rs
```diff
@@ -138,6 +181,292 @@ impl EVMAccountExtractor {
...
+                        } else {
+                            Some(Bytes::from(storage_result))
+                        };
+
+                        result.insert(slot.clone(), value);
+                    }
+                }
+            }
+            None => {
+                // TODO: Implement this -> Call get_storage_range
```
louise-poole: Is this still TODO?

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -0,0 +1,1214 @@
...
+        }
+    }
+
+    /// Scan the block storage changes and extract the updates for the tracked contracts.
+    ///
+    /// # Returns
+    /// A map of tx_hash to tx_with_changes, where the tx_with_changes contains the account deltas
+    /// for the tracked contracts.
+    ///
+    /// Note: the tx_with_changes are only containing the account deltas for the tracked contracts,
```
tvinagre, line 502: are only containing -> only contain

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -431,10 +493,18 @@ where
         }
     }
 
+    /// Scan the block storage changes and extract the updates for the tracked contracts.
```
tvinagre:
```suggestion
    /// Scans the block storage changes and extracts the updates for the tracked contracts.
```

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -431,10 +493,18 @@ where
...
+    ///
+    /// # Returns
+    /// A map of tx_hash to tx_with_changes, where the tx_with_changes contains the account deltas
+    /// for the tracked contracts.
+    ///
+    /// Note: the tx_with_changes are only containing the account deltas for the tracked contracts,
```
tvinagre:
```suggestion
    /// Note: the tx_with_changes only contain the account deltas for the tracked contracts;
```

tycho-ethereum/src/account_extractor/contract.rs
```diff
@@ -138,6 +181,292 @@ impl EVMAccountExtractor {
...
+                .provider
+                .request(
+                    "debug_storageRangeAt",
+                    &(
+                        block.hash.to_string(),
+                        0, // transaction index, 0 for the state at the end of the block)
+                        address,
+                        start_key,
+                        100000,
+                    ), // limit
```
louise-poole:
```suggestion
                        100000, // limit
                    ),
```

tycho-ethereum/src/account_extractor/contract.rs
```diff
@@ -138,6 +181,292 @@ impl EVMAccountExtractor {
...
+        warn!("Requesting all storage slots for address: {:?}. This request can consume a lot of data, and the method might not be available on the requested chain / node.", address);
+
+        let mut all_slots = HashMap::new();
+        let mut start_key = H256::zero();
+        // let block = format!("0x{:x}", block);
+        loop {
+            // let params = serde_json::json!([
+            //     block, 0, // transaction index, 0 for the state at the end of the block
+            //     address, start_key, 100000 // limit
+            // ]);
```
louise-poole: Do we still need these comments?

tycho-ethereum/src/account_extractor/contract.rs
```diff
@@ -138,6 +181,292 @@ impl EVMAccountExtractor {
...
+
+            trace!("Requesting storage range for {:?}, block: {:?}", address.clone(), block);
+            let result: StorageRange = self
+                .provider
+                .request(
+                    "debug_storageRangeAt",
+                    &(
+                        block.hash.to_string(),
+                        0, // transaction index, 0 for the state at the end of the block)
```
louise-poole:
```suggestion
                        0, // transaction index, 0 for the state at the end of the block
```

tycho-ethereum/src/account_extractor/contract.rs
```diff
@@ -138,6 +181,292 @@ impl EVMAccountExtractor {
...
+                        start_key,
+                        100000,
+                    ), // limit
+                )
+                .await
+                .map_err(|_| {
+                    RPCError::RequestError(ProviderError::CustomError(
+                        "Failed to get storage".to_string(),
+                    ))
+                })?;
```
louise-poole: Can we include details from the original error here?

tycho-ethereum/src/account_extractor/contract.rs
```diff
@@ -138,6 +181,292 @@ impl EVMAccountExtractor {
...
+        // Remove duplicates to avoid making more requests than necessary.
+        let unique_requests: Vec<StorageSnapshotRequest> = requests
+            .iter()
+            .cloned()
+            .collect::<HashSet<_>>()
+            .into_iter()
+            .collect();
+
+        // TODO: Make these configurable and optimize for preventing rate limiting.
+        // TODO: Handle rate limiting / individual connection failures & retries
```
louise-poole, line 414: FMI: when do we plan to address these? Part of v1 DCI?

tycho-ethereum/src/account_extractor/contract.rs
```diff
@@ -138,6 +181,292 @@ impl EVMAccountExtractor {
...
+            let storage_results = try_join_all(storage_futures).await?;
+
+            for (idx, request) in chunk.iter().enumerate() {
+                let address = &request.address;
+                let code = codes.get(address).cloned();
+                let balance = balances.get(address).cloned();
+                let storage = storage_results
+                    .get(idx)
+                    .cloned()
+                    .expect("Failed to get account storage");
```
louise-poole: Can we return an error instead of panic?

tycho-ethereum/src/account_extractor/contract.rs
```diff
@@ -155,44 +484,476 @@ struct StorageRange {
...
-            )
+            .get(&Bytes::from_str(BALANCER_VAULT_STR).expect("valid address"))
             .expect("update exists");
 
         assert_eq!(update.slots.len(), 47690);
 
         Ok(())
     }
+
+    #[traced_test]
```
louise-poole, line 595: What's this? 👀

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -0,0 +1,1214 @@
...
+
+        for (entrypoint_id, params_results_map) in entrypoint_results.into_iter() {
+            for (param, result) in params_results_map.into_iter() {
+                for location in result.retriggers.clone() {
+                    let entrypoint_with_params = EntryPointWithTracingParams::new(
+                        self.cache
+                            .ep_id_to_entrypoint
+                            .get(&entrypoint_id)
+                            .ok_or(ExtractionError::Setup(format!(
+                                "Got a tracing result for a unknown entrypoint: {entrypoint_id:?}"
```
louise-poole: `entrypoint_id` is already a string... no need to try get a debug string for it.
```suggestion
                                "Got a tracing result for a unknown entrypoint: {entrypoint_id}"
```

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -0,0 +1,1214 @@
...
+        trace!("DCI: Entrypoints to analyze: {:?}", entrypoints_to_analyze);
+
+        if !entrypoints_to_analyze.is_empty() {
+            let traced_entry_points = self
+                .tracer
+                .trace(
+                    block_changes.block.hash.clone(),
+                    entrypoints_to_analyze
+                        .keys()
+                        .map(Clone::clone)
```
louise-poole: Personal preference (feel free to ignore):
```suggestion
                        .cloned()
```

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -0,0 +1,1214 @@
...
+
+            let mut tx_to_traced_entry_point: HashMap<&Transaction, &TracedEntryPoint> =
+                HashMap::new();
+            for traced_entry_point in traced_entry_points.iter() {
+                let tx = entrypoints_to_analyze
+                    .get(&traced_entry_point.entry_point_with_params)
+                    // Safe to expect because we constructed the request to trace the entrypoints
+                    // with the entrypoints_to_analyze map. So every traced entrypoint should be in
+                    // the map.
+                    .expect("Every traced entrypoint should be in the entrypoints_to_analyze map");
```
louise-poole: I know you say the expect is safe here, but I'd still feel better if we raised an error instead 🥺

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -0,0 +1,1214 @@
...
+                        .and_modify(|existing_tx| {
+                            if existing_tx.index > tx.index {
+                                *existing_tx = tx;
+                            }
+                        })
+                        .or_insert(tx);
+                }
+            }
+
+            // Get the code, balance and storage changes for the new traced entrypoints
```
louise-poole:
```suggestion
            // Get the code, balance and storage changes for the new traced entrypoints.
```

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -0,0 +1,1214 @@
...
+                .map_err(|e| ExtractionError::AccountExtractionError(format!("{e:?}")))?;
+
+            // Update the block changes
+            for (account, tx) in new_account_addr_to_tx.into_iter() {
+                let account_delta = new_accounts
+                    .remove(&account)
+                    // Safe to expect because we constructed the request to get
+                    // the accounts with the new_account_addr_to_tx map. So every account
+                    // in the map should have a result.
+                    .expect("All accounts in request should have a result");
```
louise-poole: Same here - prefer an error even if we consider this impossible. Would prefer 0 risk of the indexer ever panicking, even if the panic is from a bug

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -0,0 +1,1214 @@
...
+
+                if let Some(tracked_keys) = self
+                    .cache
+                    .tracked_contracts
+                    .get(account)
+                    // Safe to expect because we checked that the contract is tracked above
+                    .expect("Contract should be tracked")
+                {
+                    slot_updates.retain(|slot, _| tracked_keys.contains(slot));
+                }
```
louise-poole: Would rather avoid the `expect` by design.
```suggestion
                // Early skip if the contract is not tracked, else get its tracked keys
                let tracked_keys = match self
                    .cache
                    .tracked_contracts
                    .get(account)
                {
                    None => continue,
                    Some(keys) => keys,
                };

                let mut slot_updates = contract_store
                    .iter()
                    .map(|(slot, value)| {
                        if value.is_zero() {
                            (slot.clone(), None)
                        } else {
                            (slot.clone(), Some(value.clone()))
                        }
                    })
                    .collect::<ContractStoreDeltas>();

                if let Some(keys) = tracked_keys {
                    slot_updates.retain(|slot, _| keys.contains(slot));
                }
```

louise-poole (commented):
Wow - tons of work here, looks great! Mainly have small comments, but for the most part this looks solid 👌

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
zizou0x: This will be addressed in a separate PR to keep decent sized PRs

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
zizou0x: Ups, that's a very dirty one indeed

tycho-ethereum/src/account_extractor/contract.rs
zizou0x: No sure, I think this should have been in a separate PR from the beginning

I think we will need to handle retries, not sure if the logic should be inside or outside this function

tycho-ethereum/src/account_extractor/contract.rs
zizou0x: It seems this is to enable tracing logs during a test, I didn't know it existed tbh

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
zizou0x: I like it a lot indeed!

tycho-ethereum/src/account_extractor/contract.rs
zizou0x: Good question, I will remove it because it seems it's done below

@tvinagre if you see this can you confirm please?

tycho-common/src/models/blockchain.rs
zizou0x: Good catch

tycho-common/src/traits.rs
```diff
@@ -13,9 +13,13 @@ use crate::{
...
 };
 
+/// A struct representing a request to get an account state.
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct StorageSnapshotRequest {
+    // The address of the account to get the state of.
     pub address: Address,
+    // The specific slots to get the state of. If `None`, the entire account state will be
+    // returned.
     pub slots: Option<Vec<Bytes>>,
```
louise-poole:
```suggestion
    pub slots: Option<Vec<StoreKey>>,
```

louise-poole (approved):
Nice! Looks great 👌

PR #561, opened 2025-04-29
feat(tycho-client): Remove panics from client

louise-poole:
Improve error handling on the tycho-client. This is done by:
- removing all panics
- replacing them with warnings, or by raising appropriate errors

In a few cases it was decided to handle an error internally instead of propagate it. Such cases are:
- **Component Tracker**
Old behaviour: panics if you try fetch linked contracts for components that are not tracking
New behaviour: logs a warning and skips that component. Extended the docs to emphasise that it is expected for `start_tracking` or `initialise_components` to be called before fetching contracts for tracked components.

Bug fix: I also found a bug where if the ws client connection dropped and could not be re-established during the first half of the `state_sync` fn on the ProtocolStateSynchronizer (before the main loop), then we'd retry starting the synchroniser on a broken connection and freeze. This has been fixed here: [22b77b3](https://github.com/propeller-heads/tycho-indexer/pull/561/commits/22b77b3eec28de7403fe98919addf3204c181d6c).

### DIFF
--- a/tycho-client-py/src/main.rs
+++ b/tycho-client-py/src/main.rs
@@ -1,4 +1,7 @@
 #[tokio::main]
 async fn main() {
-    tycho_client::cli::run_cli().await;
+    if let Err(e) = tycho_client::cli::run_cli().await {
+        eprintln!("Error: {}", e);
+        std::process::exit(1);
+    }
 }
--- a/tycho-client-py/src/main.rs
+++ b/tycho-client-py/src/main.rs
@@ -1,7 +1,9 @@
+use tycho_client::cli::run_cli;
+
 #[tokio::main]
 async fn main() {
-    if let Err(e) = tycho_client::cli::run_cli().await {
+    run_cli().await.unwrap_or_else(|e| {
         eprintln!("Error: {e}");
         std::process::exit(1);
-    }
+    });
 }
--- a/tycho-client-py/src/main.rs
+++ b/tycho-client-py/src/main.rs
@@ -1,7 +1,7 @@
 #[tokio::main]
 async fn main() {
     if let Err(e) = tycho_client::cli::run_cli().await {
-        eprintln!("Error: {}", e);
+        eprintln!("Error: {e}");
         std::process::exit(1);
     }
 }
--- a/tycho-client/src/cli.rs
+++ b/tycho-client/src/cli.rs
@@ -1,7 +1,7 @@
 use std::{collections::HashSet, str::FromStr, time::Duration};
 
 use clap::Parser;
-use tracing::{debug, info, warn};
+use tracing::{debug, error, info, warn};
 use tracing_appender::rolling;
 use tycho_common::dto::{Chain, ExtractorIdentity, PaginationParams, ProtocolSystemsRequestBody};
 
@@ -164,7 +164,7 @@ pub async fn run_cli() -> Result<(), String> {
         })
         .collect();
 
-    tracing::info!("Running with exchanges: {:?}", exchanges);
+    info!("Running with exchanges: {:?}", exchanges);
 
     run(exchanges, args).await?;
     Ok(())
@@ -227,7 +227,7 @@ async fn run(exchanges: Vec<(String, Option<String>)>, args: CliArgs) -> Result<
         .collect::<Vec<_>>();
 
     if !not_requested_protocols.is_empty() {
-        tracing::info!("Other available protocols: {}", not_requested_protocols.join(", "));
+        info!("Other available protocols: {}", not_requested_protocols.join(", "));
     }
 
     for (name, address) in exchanges {
@@ -265,7 +265,7 @@ async fn run(exchanges: Vec<(String, Option<String>)>, args: CliArgs) -> Result<
             if let Ok(msg_json) = serde_json::to_string(&msg) {
                 println!("{msg_json}");
             } else {
-                tracing::error!("Failed to serialize FeedMessage");
+                error!("Failed to serialize FeedMessage");
             }
         }
     });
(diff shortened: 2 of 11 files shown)


tycho-client/src/feed/component_tracker.rs
```diff
@@ -192,18 +192,30 @@ where
...
-                let comp = self
-                    .components
-                    .get(cid)
-                    .unwrap_or_else(|| panic!("requested component that is not present: {cid}"));
-                comp.contract_ids.iter().cloned()
+            .filter_map(|cid| {
+                if let Some(comp) = self.components.get(cid) {
+                    Some(comp.contract_ids.clone())
+                } else {
+                    warn!("Requested component is not tracked: {cid}.");
```
tvinagre: Requested component is not tracked. Skipping fetching related contracts for {cid} 

Or something related - to make sure for the user that this component will be skipped.

tycho-client/src/feed/synchronizer.rs
```diff
@@ -478,21 +532,24 @@ where
...
-                        {
+                        match res {
                             Err(e) => {
                                 error!(
                                     extractor_id=%&this.extractor_id,
                                     retry_count,
                                     error=%e,
                                     "State synchronization errored!"
                                 );
+                                if let SynchronizerError::ConnectionClosed = e {
```
tvinagre, line 543: Good catch!

tycho-client/src/deltas.rs
```diff
@@ -208,7 +213,7 @@ struct Inner {
 
 /// Shared state betweeen all client instances.
 ///
-/// This state is behind a mutex and requires synchronisation to be read of modified.
+/// This state is behind a mutex and requires synchronization to be read of modified.
```
tvinagre, line 216: Did I just witness Louise correcting british english in favor of US english? <3

tvinagre (approved):
Pure class! Thank you - small suggestion only

tycho-client/src/deltas.rs
louise-poole: I do this under great duress 😂

tycho-client-py/src/main.rs
```diff
@@ -1,4 +1,7 @@
 #[tokio::main]
 async fn main() {
-    tycho_client::cli::run_cli().await;
+    if let Err(e) = tycho_client::cli::run_cli().await {
+        eprintln!("Error: {e}");
+        std::process::exit(1);
```
zizou0x, line 7: This is equivalent to a `unwrap_or_else(panic)` or `expect()` in the end no?

tycho-client/src/cli.rs
```diff
@@ -274,17 +273,24 @@ async fn run(exchanges: Vec<(String, Option<String>)>, args: CliArgs) {
     // Monitor the WebSocket, BlockSynchronizer and message printer futures.
     tokio::select! {
         res = ws_jh => {
-            let _ = res.expect("WebSocket connection dropped unexpectedly");
+            if let Err(e) = res {
+                tracing::error!("WebSocket connection dropped unexpectedly: {}", e);
```
zizou0x: Can we import `error!` please?

tycho-client/src/feed/component_tracker.rs
```diff
@@ -192,18 +192,32 @@ where
...
-                comp.contract_ids.iter().cloned()
+            .filter_map(|cid| {
+                if let Some(comp) = self.components.get(cid) {
+                    Some(comp.contract_ids.clone())
+                } else {
+                    warn!(
+                        "Requested component is not tracked: {cid}. Skipping fetching contracts..."
+                    );
+                    None
+                }
```
zizou0x, line 218: Is it safe to change the behaviour here?

Shouldn't we forward an error if one id isn't found?

I find it weird that we try to get contracts and some of them will just sneakily be ignored?

tycho-client/src/main.rs
```diff
@@ -2,5 +2,8 @@ use tycho_client::cli::run_cli;
 
 #[tokio::main]
 async fn main() {
-    run_cli().await;
+    if let Err(e) = run_cli().await {
+        eprintln!("Error: {e}");
+        std::process::exit(1);
```
zizou0x, line 7: Same comment as above here

tycho-client/src/stream.rs
```diff
@@ -132,6 +132,7 @@ impl TychoStreamBuilder {
         // Attempt to read the authentication key from the environment variable if not provided
         let auth_key = self
             .auth_key
+            .clone()
```
zizou0x, line 135: Do we really need to clone here? Can't we just take a ref?

zizou0x (approved):
Overall looks good to me, just some small comments

tycho-client-py/src/main.rs
louise-poole: Not quite - from my understanding a panic will abruptly terminate the process and only show the error in the stack trace (not printed to the terminal in this case) while this will print it to the terminal and then gracefully try exit. It felt like a cleaner/more elegant way to end a CLI program if it errors.

tycho-client/src/stream.rs
louise-poole: Yes - since the `or_else` used here takes ownership of it and later I want to use `self` again to call the new `display_available_protocols` function. I guess I could move the `display_available_protocols` off the builder though? But this is such a small clone I think it's fine no?

tycho-client/src/main.rs
louise-poole: Same response here

tycho-client/src/feed/component_tracker.rs
louise-poole: This approach is consistant with what is done for native components too... with native components if we request snapshots for components not in the tracker we just quietly skip them. See [here](https://github.com/propeller-heads/tycho-indexer/blob/main/tycho-client/src/feed/synchronizer.rs#L256).
Note - this fn is only ever called in the `get_snapshots` linked in this comment.

tycho-client-py/src/main.rs
zizou0x: I think it doesn't make any difference in the end, I just find unwrap_or_else cleaner but it's just a personal preference I guess so all good

tycho-client/src/feed/component_tracker.rs
zizou0x: I find it weird with this name but as far as it's well documented it's fine I would say

But we should think about tweaking the function name to make it obvious from the name too? Maybe get_available_...? What do you think?

tycho-client/src/stream.rs
zizou0x: I was hoping that calling `.as_deref` before the `or_else` would work but I guess we can't because we `env ::var` returns a string and not a ref

Yes yes a small clone doesn't justify a big refactor

tycho-client-py/src/main.rs
louise-poole: Ah - you mean an unwrap_or_else without the panic. Yeah, that does feel more rusty - will do that 👍

tycho-client/src/feed/component_tracker.rs
louise-poole: Yeah! I really like that suggestion! 👌

PR #562, opened 2025-04-30
fix: add backoff for retries on connections error with Substreams

zizou0x:
This is needed so we don't flood them on error. We recently experienced this.

### DIFF
--- a/tycho-indexer/src/substreams/stream.rs
+++ b/tycho-indexer/src/substreams/stream.rs
@@ -63,6 +63,22 @@ impl SubstreamsStream {
 static DEFAULT_BACKOFF: Lazy<ExponentialBackoff> =
     Lazy::new(|| ExponentialBackoff::from_millis(500).max_delay(Duration::from_secs(45)));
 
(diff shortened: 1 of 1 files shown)


tycho-indexer/src/substreams/stream.rs
```diff
@@ -175,11 +175,18 @@ fn stream_blocks(
...
                     error!("Unable to connect to endpoint: {:#}", e);
+
+                    // If we reach this point, we must wait a bit before retrying
+                    if let Some(duration) = backoff.next() {
+                        info!("Will try to reconnect after {:?}", duration);
+                        sleep(duration).await;
+                        retry_count += 1;
+                    } else {
+                        counter!("substreams_failure", "extractor" => extractor_id.clone(), "cause" => "max_retries_exceeded").increment(1);
+                        return Err(anyhow!("Backoff requested to stop retrying, quitting"))?;
```
kayibal: This code is just a copy of the above, no? Maybe we should put it into a function?

louise-poole (approved):

PR #565, opened 2025-05-01
refactor: propagate error on entry not found in the id/value cache

zizou0x:
Currently if the entry isn't found we panic. This leads to missleading error message in the RPC if, for example, we can't find the given "chain". This commit should improve this by propagating of the error in this case.


louise-poole (approved):
LGTM!

PR #568, opened 2025-05-02
feat: enhance shutdown handling with SIGTERM support

zizou0x:
This update improves the shutdown process by adding support for SIGTERM signals.

This is needed to properly terminate the pod in Kubernetes. Currently it only ends when Kubernetes sends the `SIGKILL` after waiting 30 secs.


louise-poole (approved):
Nice find!

PR #570, opened 2025-05-13
fix(tycho-client): fix block position bug for delayed extractors on start up

louise-poole:
Delayed extractors on start up were erroring with `Could not determine history` when they tried to catch up to the later blocks. This was due to the fact that on start-up, the BlockHistory created by the client has a gap for the unwitnessed blocks between the delayed extractor's block and the current block the other extractors are on. The `determine_block_position` function was then unable to determine where the catch-up blocks belonged within that gap.

We would also get a similar error if an extractor was delayed longer than the block history cache limit (currently 15 blocks).

Solution: 
1) improve the BlockHistory instantiation logic to skip detached blocks and ensure a fully connected history.
2) update `determine_block_position` to mark blocks older than the oldest in history as delayed.

### DIFF
--- a/tycho-client/src/feed/block_history.rs
+++ b/tycho-client/src/feed/block_history.rs
@@ -149,21 +149,36 @@ impl BlockHistory {
         let latest = self
             .latest()
             .ok_or(BlockHistoryError::EmptyHistory)?;
+
         Ok(if block.parent_hash == latest.hash {
+            // if the block is the next expected block.
             BlockPosition::NextExpected
         } else if (block.hash == latest.hash) & !block.revert {
+            // if the block is the latest block and it is not a revert.
             BlockPosition::Latest
         } else if self.reverts.contains(&block.hash) {
-            // in this case the block is still on an already reverted branch.
+            // if the block is still on an already reverted branch.
             BlockPosition::Delayed
         } else if block.number <= latest.number {
-            // if this block is potentially delayed we have seen it's hash before.
-            if block.revert & self.hash_in_history(&block.hash) {
-                // if it is a revert, that is a expected forward update.
-                BlockPosition::NextExpected
-            } else if self.hash_in_history(&block.hash) {
-                // if this is not a revert it means this block is delayed.
+            // block is potentially delayed or reverted.
+
+            let oldest = self
+                .oldest()
+                .ok_or(BlockHistoryError::EmptyHistory)?;
+
+            if block.number < oldest.number {
+                // if this block is older than the oldest block in our history it means it is
+                // delayed.
                 BlockPosition::Delayed
+            } else if self.hash_in_history(&block.hash) {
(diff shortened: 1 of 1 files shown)


tycho-client/src/feed/block_history.rs
```diff
@@ -114,21 +149,36 @@ impl BlockHistory {
...
+                // delayed.
                 BlockPosition::Delayed
+            } else if self.hash_in_history(&block.hash) {
+                // if this block is in our history
+                if block.revert {
+                    // if it is a revert, that is a expected forward update.
+                    BlockPosition::NextExpected
+                } else {
+                    // if this is not a revert it means this block is delayed.
+                    BlockPosition::Delayed
```
kayibal, line 180: Is it still possible to get into this case?

tycho-client/src/feed/block_history.rs
```diff
@@ -260,15 +310,48 @@ mod test {
         assert!(res.is_err());
     }
 
+    #[test]
+    fn test_new_block_history() {
```
kayibal, line 314: weird test name

tycho-client/src/feed/block_history.rs
```diff
@@ -114,21 +149,36 @@ impl BlockHistory {
...
-                // if this is not a revert it means this block is delayed.
+            // block is potentially delayed or reverted.
+
+            let oldest = self
+                .oldest()
+                .ok_or(BlockHistoryError::EmptyHistory)?;
+
+            if block.number < oldest.number {
+                // if this block is older than the oldest block in our history it means it is
+                // delayed.
```
kayibal, line 171: This should only happen at startup or if the delay is > block_history. Should we check for this here? Like set a flag once the history is full and also check that the distance between the numbers is > history?

tycho-client/src/feed/block_history.rs
louise-poole: Yes. We'll hit this else block if the block we're looking at is in the history, but isn't the latest block or a revert.

tycho-client/src/feed/block_history.rs
louise-poole: Do you feel it's necessary? Either way the block is a delayed block and that's all we're trying to determine here...

tycho-client/src/feed/block_history.rs
louise-poole: What do you prefer? `test_block_history_new`?

kayibal: Idk what you mean by "new" here

tycho-client/src/feed/block_history.rs
kayibal: Could be useful to catch unexpected cases of this but it's not required I think

kayibal (approved):

PR #571, opened 2025-05-14
fix(tycho-client): include message text in websocket error logging

louise-poole:


tycho-client/src/deltas.rs
```diff
@@ -482,11 +482,14 @@ impl WsDeltasClient {
...
+                        error!(
+                            "Failed to deserialize WebSocketMessage: {}. \nMessage: {}",
+                            e, text
+                        );
                     }
                 },
-                Err(e) => {
-                    error!(error = %e, message=text, "Failed to deserialize message: invalid json");
+                Err(_) => {
+                    error!("Failed to deserialize message: invalid JSON. \nMessage: {}", text);
```
zizou0x: The error doesn't matter here?

louise-poole: I suppose the more detail the better. Though for most cases just seeing 'invalid json' with the invalid text is enough to know what's up.

PR #572, opened 2025-05-14
refactor(indexer): log error code for substream error metrics

louise-poole:
"Tonic error" does not give us much information. Instead, the error code tells us if it is an internal error, unavailable connection etc.


zizou0x (approved):
LGTM, IMO it's a refactor or feat but not a fix

PR #575, opened 2025-05-16
refactor: improve Entrypoints methods on PostgresGateway

zizou0x:
This PR improves some logic related to `Entrypoints`, specifically:

- The main link with protocol components has been moved from the `EntrypointTracingParams` to the `Entrypoint`. The old link stays for debugging purposes but becomes optional
- The methods to insert and get `Entrypoint` and `EntrypointTracingParams` are now separated. This granularity was needed to be able to insert new data more with more flexibility from the extractor.
- Improved namings, change from `TracingData` to `TracingParams` for clarity. 

All these changes have previously been discussed [here](https://propellerswap.slack.com/archives/C05MPCJUQPM/p1747236824357709)

### DIFF
--- a/tycho-common/src/models/blockchain.rs
+++ b/tycho-common/src/models/blockchain.rs
@@ -364,11 +364,11 @@ pub struct EntryPointWithTracingParams {
     /// The entry point to trace, containing the target contract address and function signature
     pub entry_point: EntryPoint,
     /// The tracing parameters for this entry point
-    pub params: EntryPointTracingParams,
+    pub params: TracingParams,
 }
 
 impl EntryPointWithTracingParams {
-    pub fn new(entry_point: EntryPoint, params: EntryPointTracingParams) -> Self {
+    pub fn new(entry_point: EntryPoint, params: TracingParams) -> Self {
         Self { entry_point, params }
     }
 }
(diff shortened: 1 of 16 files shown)


tycho-common/src/storage.rs
```diff
@@ -498,17 +501,29 @@ impl EntryPointFilter {
...
+        &self,
+        entry_points: &[(ComponentId, Vec<EntryPoint>)],
+    ) -> Result<(), StorageError>;
+
     /// Upserts a list of entry points with their tracing data into the database.
-    async fn upsert_entry_points_with_data(
+    async fn upsert_entry_point_tracing_data(
         &self,
-        entry_points: &[(ComponentId, Vec<EntryPointWithData>)],
+        entry_points_data: &[(EntryPointId, Vec<(EntryPointTracingParams, Option<ComponentId>)>)],
```
zizou0x: Does this interface make sense to be called from the extractor?

tycho-common/src/models/blockchain.rs
```diff
@@ -358,25 +358,25 @@ impl EntryPoint {
...
+    pub fn new(entry_point: EntryPoint, params: EntryPointTracingParams) -> Self {
+        Self { entry_point, params }
     }
 }
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Eq, Hash)]
 /// An entry point to trace. Different types of entry points tracing will be supported in the
 /// future. Like RPC debug tracing, symbolic execution, etc.
-pub enum EntryPointTracingData {
+pub enum EntryPointTracingParams {
```
louise-poole: Is 'entrypoint' actually needed in the struct name here? Just thinking, `EntryPointWithTracingParams` and `EntryPointTracingParams` are so similar which isn't so nice. Maybe this could rather just be:
```suggestion
pub enum TracingParams {
```

louise-poole: Then to follow up: `RPCTracerEntryPoint` isn't an entrypoint per-se... It's just tracing configs/data so might be better to be `RPCTracerParams`?

tycho-common/src/storage.rs
```diff
@@ -498,17 +501,32 @@ impl EntryPointFilter {
 // Trait for entry point gateway operations.
 #[async_trait]
 pub trait EntryPointGateway {
-    /// Upserts a list of entry points with their tracing data into the database.
-    async fn upsert_entry_points_with_data(
+    /// Upserts a list of entry points into the database.
+    async fn upsert_entry_points(
+        &self,
+        entry_points: &[(ComponentId, Vec<EntryPoint>)],
```
zizou0x: We should make this a hashmap directly

louise-poole (approved):

PR #577, opened 2025-05-16
feat: extend extractor models to support DCI requirements

louise-poole:


tycho-indexer/Cargo.toml
```diff
@@ -72,7 +72,7 @@ mini-moka = "0.10.3"
 num-bigint = "0.4.4"
 num-traits = "0.2.19"
 num_cpus = "1.16.0"
-tycho-substreams = "0.2.0"
+tycho-substreams = { git = "https://github.com/propeller-heads/tycho-protocol-sdk.git", tag = "0.3.1" }
```
louise-poole: I will publish tycho-substreams on crates.io and remove the github link once we're settled with our SDK changes

zizou0x: Let's add a todo to not forget about this?

kayibal: I think we should move the protobuf models to tycho-common, depending on substreams is wrong imo

louise-poole: If we do that then we have duplicate code no? We used to have it defined on both sdk and here and found inconsistencies easily appeared between them. Or do you want our sdk repo to depend on tycho-common and the only place we define it be here instead?

zizou0x (approved):
LGTM thanks

PR #580, opened 2025-05-20
chore: Optimized DB queries

dianacarvalho1:
Still use the old (very fast) query to get most of the entries. For the entries that are missing, look for them in the archived tables (this query is much slower)

I used SQL `EXPLAIN ANALYSE` to compare the queries and:
- the query that uses the `valid_to.eq(MAX_TS)` takes:
````
Planning Time: 0.142 ms     
Execution Time: 0.027 ms    
````
- the one that uses the `distinct` takes:
````
Planning Time: 10.915 ms  
Execution Time: 0.738 ms  
````

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -2,6 +2,7 @@ version: '3.1'
 services:
   db:
     build:
+      context: .
       dockerfile: postgres.Dockerfile
     restart: "always"
     environment:
@@ -35,6 +36,6 @@ services:
       - ./extractors.yaml:/opt/tycho-indexer/extractors.yaml
       - ./substreams/:/opt/tycho-indexer/substreams/
     entrypoint: [ "/usr/wait-for-postgres.sh", "db" ]
-    command: [ "/opt/tycho-indexer/tycho-indexer", "--endpoint", "https://mainnet.eth.streamingfast.io:443", "index"]
+    command: [ "/opt/tycho-indexer/tycho-indexer", "--endpoint", "https://mainnet.eth.streamingfast.io:443", "index" ]
 volumes:
   postgres_data:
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: docker-compose.yaml [hunk 1]
```diff
@@ -2,6 +2,7 @@ version: '3.1'
 services:
   db:
     build:
+      context: .
```
dianacarvalho1, line 5: without this the docker compose didn't work for me

tycho-storage/src/postgres/orm.rs
```diff
@@ -460,27 +460,56 @@ impl PartitionedVersionedRow for NewComponentBalance {
...
-                component_balance::token_id,
-                component_balance::valid_to.desc(),
-            ))
             .get_results(conn)
             .await
-            .map_err(PostgresError::from)?
+            .map_err(PostgresError::from)?;
+        let found_ids: HashSet<_> = results
+            .iter()
+            .map(|cs| (&cs.protocol_component_id, &cs.token_id))
```
louise-poole: What does `cs` stand for? 👀

tycho-storage/src/postgres/orm.rs
```diff
@@ -460,27 +460,56 @@ impl PartitionedVersionedRow for NewComponentBalance {
...
-            .order_by((
-                component_balance::protocol_component_id,
-                component_balance::token_id,
-                component_balance::valid_to.desc(),
-            ))
             .get_results(conn)
             .await
-            .map_err(PostgresError::from)?
+            .map_err(PostgresError::from)?;
+        let found_ids: HashSet<_> = results
```
louise-poole: Just to nit pick 👀 😂 
```suggestion
            .map_err(PostgresError::from)?;

        let found_ids: HashSet<_> = results
```

tycho-storage/src/postgres/orm.rs
dianacarvalho1: AHAHAH I thought the same.. 👀 but it was like that already here.. I guess it is `contract_storage` and it was a bad copy pasta 🍝 i'll make it better

PR #581, opened 2025-05-20
feat: (WIP) Add get_traced_entrypoints endpoint.

tamaralipows:


tycho-indexer/src/services/rpc.rs
```diff
@@ -114,13 +123,20 @@ where
             dto::ProtocolComponentRequestResponse,
         >::new("protocol_components", 500, 7 * 60);
 
+        let traced_entry_point_cache = RpcCache::<
+            dto::TracedEntryPointRequestBody,
+            // TODO is this a good capacity?
```
louise-poole: Yeah, more than enough I reckon 👍

tycho-indexer/src/services/rpc.rs
```diff
@@ -175,6 +191,8 @@ where
             .db_gateway
             .get_contracts(
                 &chain,
+                // TODO question: Do we need to manually apply the pagination to the filter params
+                //  like it's done here?
```
louise-poole: Good question. We added this because our rpc caches whole response pages. This means if I query for components A, B, C and D with pagination size of 2, but only A, C and D have data in the db then the response from the gateway query will be:
page 1: A and C
page 2: D
We only cache full pages so will cache page 1. Then someone repeats this query later when B was added to the DB. The response they will get is:
page 1 (from cache): A and C
page 2: C and D (the DB query's 'page 1' would now technically contain A and B)
This in unexpected for a user to get multiple pages with the same component repeated. For consistency we pre-apply pagination to ensure it's predetermined which components will appear on which page.

tycho-indexer/src/services/rpc.rs
```diff
@@ -718,6 +736,110 @@ where
...
     }
+
+    async fn get_traced_entry_points(
+        &self,
+        request: &dto::TracedEntryPointRequestBody,
+    ) -> Result<dto::TracedEntryPointRequestResponse, RpcError> {
+        info!(?request, "Getting traced entry points.");
+
+        // TODO AFAIU this cache is automatically written to by the fallback method during a cache
+        //  miss. Is this correct?
```
louise-poole: Yes, depending on the condition you provide below. You gave the condition `request.pagination.page < last_page` so it will write to cache anytime a response meets that condition (is not the last page).

tycho-indexer/src/services/rpc.rs
```diff
@@ -718,6 +736,110 @@ where
...
+        self.traced_entry_point_cache
+            .get(request.clone(), |r| async {
+                self.get_traced_entry_points_inner(r)
+                    .await
+                    .map(|res| {
+                        // TODO should I check that all requested components are in the response?
```
louise-poole: No need - someone might request a bogus component and we don't care that it's not in the response. We can trust the gateway will diligently return all that it has data for.

tycho-indexer/src/services/rpc.rs
```diff
@@ -718,6 +736,110 @@ where
...
+                        // TODO why do we not cache if it's the last page?
```
louise-poole: Because that page is incomplete and will change as soon as more data is entered into the DB. If we cached the last page then we would not re-query it and any newly added data will not appear in any responses until that cached last page is invalidated (cache ttl is 7 mins).

tycho-indexer/src/services/rpc.rs
tamaralipows: Ahhhhh good explanation - thank you! It's missing this implementation in my code then

tycho-indexer/src/services/rpc.rs
```diff
@@ -1233,6 +1426,134 @@ mod tests {
...
+            .unwrap();
+
+        assert_eq!(
+            traced_entry_points
+                .traced_entry_points
+                .get(&component_id)
+                .unwrap()
+                .len(),
+            2
+        );
```
tamaralipows: Didn't test for actual contents here because I saw a few other tests don't do it, so I wasn't sure if we just test for this more thoroughly at a different level? 

I can add a more thorough assert tomorrow (ignoring the ordering inside the returned vec) though if you think it's important!

tycho-common/src/dto.rs
```diff
@@ -1322,6 +1322,103 @@ impl ProtocolSystemsRequestResponse {
...
+#[derive(Serialize, PartialEq, ToSchema, Eq, Clone)]
+pub struct TracedEntryPointRequestResponse {
+    pub traced_entry_points: HashMap<String, Vec<(EntryPointWithTracingParams, TracingResult)>>,
+    pub pagination: PaginationResponse,
+}
+
+pub struct AddEntrypointRequestBody {
+    pub entrypoints_with_tracing_data: Vec<(String, Vec<EntryPointWithTracingParams>)>,
+    pub chain: Chain,
+}
```
zizou0x, line 1434: This is not needed for this task right?

tycho-common/src/dto.rs
```diff
@@ -1322,6 +1322,103 @@ impl ProtocolSystemsRequestResponse {
...
+    pub chain: Chain,
+    /// Filters by protocol, required to correctly apply unconfirmed state from
+    /// ReorgBuffers
+    pub protocol_system: String,
+    /// Filter by component ids
+    pub component_ids: Option<Vec<String>>,
+    /// Max page size supported is 500
+    #[serde(default)]
+    pub pagination: PaginationParams,
+}
```
zizou0x:
```suggestion
pub struct TracedEntryPointRequestBody {
    #[serde(default)]
    pub chain: Chain,
    /// Filters by protocol, required to correctly apply unconfirmed state from
    /// ReorgBuffers
    pub protocol_system: String,
    /// Filter by component ids
    pub component_ids: Option<Vec<String>>,
    /// Max page size supported is 500
    #[serde(default)]
    pub pagination: PaginationParams,
}
```

tycho-common/src/dto.rs
```diff
@@ -1322,6 +1322,103 @@ impl ProtocolSystemsRequestResponse {
...
+    #[serde(with = "hex_bytes")]
+    pub target: Bytes,
+    pub signature: String,
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Clone, ToSchema, Eq, Hash)]
+pub struct RPCTracerParams {
+    pub caller: Option<Bytes>,
+    pub calldata: Bytes,
+}
```
zizou0x: We need to add some macros here to keep our OpenAPI clean (see how it's done above)


```suggestion
#[derive(Debug, Serialize, Deserialize, PartialEq, Clone, ToSchema, Eq, Hash)]
pub struct EntryPoint {
    #[schema(example="0x123456789:getRate(address,bool)")]
    pub external_id: String,
    #[schema(value_type=String, example="0x123456789")]
    #[serde(with = "hex_bytes")]
    pub target: Bytes,
    #[schema(example="getRate(address,bool)")]
    pub signature: String,
}

#[derive(Debug, Serialize, Deserialize, PartialEq, Clone, ToSchema, Eq, Hash)]
pub struct RPCTracerParams {
    #[schema(value_type=Option<String>)]
    #[serde(with = "hex_bytes_option", default)]
    pub caller: Option<Bytes>,
    #[schema(value_type=String, example="0x123456780000000000")]
    #[serde(with = "hex_bytes")]
    pub calldata: Bytes,
}
```

tycho-common/src/dto.rs
```diff
@@ -1322,6 +1322,103 @@ impl ProtocolSystemsRequestResponse {
...
+            params: value.params.into(),
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, Default, PartialEq, ToSchema, Eq, Clone)]
+pub struct TracingResult {
+    pub retriggers: HashSet<(Bytes, Bytes)>,
+    pub called_addresses: HashSet<Bytes>,
+}
```
zizou0x:
```suggestion
#[derive(Serialize, Deserialize, Debug, Default, PartialEq, ToSchema, Eq, Clone)]
    #[schema(value_type=HashSet<(String, String)>)]
    pub retriggers: HashSet<(Bytes, Bytes)>,
    #[schema(value_type=HashSet<String>)]
    pub called_addresses: HashSet<Bytes>,
}
```

tycho-indexer/src/services/rpc.rs
```diff
@@ -718,6 +732,134 @@ where
             }
         }
     }
+
+    async fn get_traced_entry_points(
```
zizou0x, line 737: We need this to get metrics for this rpc method

```suggestion
    #[instrument(skip(self, request))]
    async fn get_traced_entry_points(
```

tycho-indexer/src/services/rpc.rs
```diff
@@ -718,6 +732,134 @@ where
...
+
+        let entry_points_tracing_params_data = self
+            .db_gateway
+            .get_entry_points_tracing_params(filter, Some(&pagination_params))
+            .await
+            .map_err(|err| {
+                error!(error = %err, "Error while getting entry points with tracing params.");
+                err
+            })?;
+
```
zizou0x, line 788: Let's add a trace! log here with the returned data, this can be useful for debugging

tycho-indexer/src/services/rpc.rs
```diff
@@ -718,6 +732,134 @@ where
...
+        let traced_entry_points = self
+            .db_gateway
+            .get_traced_entry_points(&entry_point_ids)
+            .await
+            .map_err(|err| {
+                error!(error = %err, "Error while getting traced entry points.");
+                err
+            })?;
+
```
zizou0x, line 819: Same here with the `trace!` log

tycho-indexer/src/services/rpc.rs
```diff
@@ -967,6 +1109,53 @@ pub async fn protocol_systems<G: Gateway>(
...
+pub async fn traced_entry_points<G: Gateway>(
+    body: web::Json<dto::TracedEntryPointRequestBody>,
+    handler: web::Data<RpcHandler<G>>,
+) -> HttpResponse {
+    // Tracing and metrics
+    tracing::Span::current().record("page", body.pagination.page);
+    tracing::Span::current().record("page.size", body.pagination.page_size);
+    counter!("rpc_requests", "endpoint" => "traced_entry_points").increment(1);
+
+    if body.pagination.page_size > 100 {
```
zizou0x, line 1141: It says 500 in the dto docstrings, we can set it to 100 for now but we need to keep the docstring updated

```
    /// Max page size supported is 500
    #[serde(default)]
    pub pagination: PaginationParams,
```

tycho-indexer/src/services/rpc.rs
```diff
@@ -718,6 +732,134 @@ where
...
+
+        for (component_id, entry_points_set) in entry_points_tracing_params_data.entity {
+            let mut pairs = Vec::new();
+
+            for entry_point_with_params in entry_points_set {
+                let entry_point_id = entry_point_with_params
+                    .entry_point
+                    .external_id
+                    .clone();
+                let tracing_param = entry_point_with_params.params.clone();
```
zizou0x: Do we need these two clones? It seems we're only using a reference below

tycho-indexer/src/services/rpc.rs
zizou0x: I think it could help to find a bug in the result aggregation logic in that case no? Like if I'm not wrong we could be mixing up the component_id -> result map without noticing

tycho-indexer/src/services/rpc.rs
```diff
@@ -1233,6 +1426,134 @@ mod tests {
...
+                .len(),
+            2
+        );
+        assert_eq!(traced_entry_points.pagination.total, 3);
+        assert_eq!(
+            traced_entry_points
+                .pagination
+                .total_pages(),
+            2
+        );
```
zizou0x, line 1625: Would be nice to add a case for pagination, like add another component and make sure we only get one of them with a page size of 1

zizou0x (commented):
Left a few comments to polish things but the main logic is there and it looks solid

tycho-common/src/dto.rs
zizou0x: It's actually just missing `#[serde(default)]` on chain

tycho-indexer/src/testing.rs
```diff
@@ -30,6 +37,116 @@ mock! {
         async fn save_state(&self, state: &ExtractionState) -> Result<(), StorageError>;
     }
 
+    impl EntryPointGateway for Gateway {
+        #[allow(clippy::type_complexity, clippy::type_repetition_in_bounds)]
+        fn insert_entry_points<'life0, 'life1, 'async_trait>(
```
dianacarvalho1, line 42: 😰

tycho-common/src/dto.rs
tamaralipows: Nope - will leave it there for the next task though

tycho-indexer/src/services/rpc.rs
tamaralipows: I assume you mean if it returns successfully, not if it errors. Adding this!

tycho-indexer/src/services/rpc.rs
tamaralipows: Unfortunately it seems so, for both. :(
```
(base) [16:36:37] [~/Code/tycho-indexer] >>> cargo clippy --workspace --lib --all-targets --all-features -- 
    Checking tycho-indexer v0.66.4 (/Users/tamaralipowski/Code/tycho-indexer/tycho-indexer)
error[E0382]: borrow of partially moved value: `entry_point_with_params`
   --> tycho-indexer/src/services/rpc.rs:852:29
    |
832 |                 let tracing_param = entry_point_with_params.params;
    |                                     ------------------------------ value partially moved here
...
852 |                 pairs.push((entry_point_with_params.clone().into(), tracing_result.clone().into()));
    |                             ^^^^^^^^^^^^^^^^^^^^^^^ value borrowed here after partial move
    |
    = note: partial move occurs because `entry_point_with_params.params` has type `tycho_common::models::blockchain::TracingParams`, which does not implement the `Copy` trait

For more information about this error, try `rustc --explain E0382`.
```

tycho-indexer/src/services/rpc.rs
tamaralipows: Haha this test will be a thousand lines long. That's fine though, LFG

tamaralipows: Just realized that we paginate before the db queries so this won't make much difference at all. The mocked db responses will be exactly the same. I added an attr anyway

tycho-indexer/src/services/rpc.rs
zizou0x: Yes you will need to use references instead of taking partial ownership

```suggestion
                let entry_point_id = entry_point_with_params
                    .entry_point
                    .external_id
                    .as_str();
                let tracing_param = &entry_point_with_params.params;
```

tamaralipows: Yayyyy thanks  a lot for this tip

tycho-indexer/src/services/rpc.rs
```diff
@@ -718,6 +732,146 @@ where
...
+            for entry_point_with_params in entry_points_set {
+                let entry_point_id = entry_point_with_params
+                    .entry_point
+                    .external_id
+                    .as_str()
+                    .to_string();
+                let tracing_param = &entry_point_with_params.params;
+
+                let results_by_param = traced_entry_points
+                    .get(&entry_point_id)
```
zizou0x: FYI currently this still clones `entry_point_id` (the `to_string` methods is equivalent to a clone here), and then we only use the reference to `get()` in the hashmap. Instead we can completely avoid writing to the heap like this:

```suggestion
                let entry_point_id = entry_point_with_params
                    .entry_point
                    .external_id
                    .as_str();
                let tracing_param = &entry_point_with_params.params;

                let results_by_param = traced_entry_points
                    .get(entry_point_id)
```

tycho-indexer/src/services/rpc.rs
```diff
@@ -1233,6 +1438,198 @@ mod tests {
...
+        );
+        assert_eq!(traced_entry_points.pagination.total, 2);
+        assert_eq!(
+            traced_entry_points
+                .pagination
+                .total_pages(),
+            2
+        );
+
+        // Second request (should hit cache and not increase gateway access count)
```
zizou0x, line 1581: FMI how do you know if it hits the cache? Would the Mock panic if it was called twice?

zizou0x (approved):
Thanks! Looks good to me, just added a question and a picky 🦀 comment

tycho-indexer/src/services/rpc.rs
tamaralipows: Yeah! It panics, I've seen it haha. I was also confused how this works, but I guess because we do `return_once`.

tycho-indexer/src/services/rpc.rs
tamaralipows: Oh, gosh. Thanks for pointing this out. Noted and fixed!

tycho-common/src/dto.rs
```diff
@@ -1322,6 +1322,121 @@ impl ProtocolSystemsRequestResponse {
...
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Clone, ToSchema, Eq, Hash)]
+pub struct RPCTracerParams {
+    #[schema(value_type=Option<String>)]
+    #[serde(with = "hex_bytes_option", default)]
+    pub caller: Option<Bytes>,
+    #[schema(value_type=String, example="0x679aefce")]
+    #[serde(with = "hex_bytes")]
+    pub calldata: Bytes,
```
tvinagre, line 1363: Picky comment, but since those are user-facing interfaces, I'd appreciate a little more detail like you did with the structs above, with a description of what is expected on those parameters

tycho-common/src/dto.rs
```diff
@@ -1322,6 +1322,121 @@ impl ProtocolSystemsRequestResponse {
...
+            models::blockchain::TracingParams::RPCTracer(params) => {
+                TracingParams::RPCTracer(params.into())
+            }
+        }
+    }
+}
+
+impl From<models::blockchain::EntryPoint> for EntryPoint {
+    fn from(value: models::blockchain::EntryPoint) -> Self {
+        EntryPoint {
```
tvinagre: Rusty picky but I think it's nicer to use `Self` instead of EntryPoint for the return type. Avoids repeating the type name and if the name changes is one less update :)
(I've seen both patterns on the code, we don't enforce it, it's just a suggestion)

tycho-indexer/src/services/rpc.rs
```diff
@@ -718,6 +732,144 @@ where
...
+
+    async fn get_traced_entry_points_inner(
+        &self,
+        request: dto::TracedEntryPointRequestBody,
+    ) -> Result<dto::TracedEntryPointRequestResponse, RpcError> {
+        let pagination_params: PaginationParams = (&request.pagination).into();
+
+        // For consistency, we pre-apply pagination to ensure it's predetermined which components
+        // will appear on which page.
+        // Apply pagination to the contract addresses. This is done so that we can determine which
```
tvinagre: component ids?

tycho-indexer/src/services/rpc.rs
```diff
@@ -718,6 +732,144 @@ where
...
+            .await
+    }
+
+    async fn get_traced_entry_points_inner(
+        &self,
+        request: dto::TracedEntryPointRequestBody,
+    ) -> Result<dto::TracedEntryPointRequestResponse, RpcError> {
+        let pagination_params: PaginationParams = (&request.pagination).into();
+
+        // For consistency, we pre-apply pagination to ensure it's predetermined which components
```
tvinagre: components -> entrypoints

tvinagre (approved):
Pure class :)

tycho-common/src/storage.rs
```diff
@@ -546,9 +552,13 @@ pub trait EntryPointGateway {
     /// Retrieves a map of component ids to a set of entry points with their tracing data from the
     /// database.
     ///
+    /// # Arguments
     /// * `filter` - The EntryPointFilter to apply to the query.
     /// * `pagination_params` - The pagination parameters to apply to the query, if None, all
     ///   results are returned.
+    ///
+    /// # Returns
+    /// A map of component ids to a set of entry points with their tracing data.
```
louise-poole, line 561: 'data' is pretty generic... is it the result of the trace? Or the input data? Or more than just input data?. 
```suggestion
    /// A map of component ids to a set of entry points with their tracing params.
```

tycho-indexer/src/services/rpc.rs
louise-poole: No she's right - we make this request using component ids. The RPC gets a request for specific components and then it translates that into multiple gateway calls (first one used components to get entrypoints + params, second one uses the entrypoints to fetch the trace results).

tycho-indexer/src/services/rpc.rs
```diff
@@ -718,6 +732,144 @@ where
...
+            .await
+            .map_err(|err| {
+                error!(error = %err, "Error while getting traced entry points.");
+                err
+            })?;
+
+        trace!(
+            traced_entry_points = ?traced_entry_points,
+            "Retrieved traced entry points entry points from database."
+        );
```
```suggestion
        trace!(
            traced_entry_points = ?traced_entry_points,
            "Retrieved traced entry points from database."
        );
```

tycho-indexer/src/services/rpc.rs
```diff
@@ -718,6 +732,144 @@ where
...
+                    })?;
+
+                let tracing_result = results_by_param
+                    .get(tracing_param)
+                    .ok_or_else(|| {
+                        RpcError::Storage(StorageError::NotFound(
+                            "TracedEntryPoint".to_string(),
+                            format!("{entry_point_id:?} with params {tracing_param:?}"),
+                        ))
+                    })?;
```
louise-poole: Not fully sure what to do in this case. If we have entrypoint params in the DB, but no trace result for it, then the trace failed. I don't think we should make the entire RPC request error due to this (it is bound to occasionally happen), but I'm not sure what to suggest otherwise. We should at least log a `warn` and skip it?
@zizou0x what does DCI do with failed traces?

tycho-indexer/src/services/rpc.rs
```diff
@@ -718,6 +732,144 @@ where
...
+                            format!("{entry_point_id:?}"),
+                        ))
+                    })?;
+
+                let tracing_result = results_by_param
+                    .get(tracing_param)
+                    .ok_or_else(|| {
+                        RpcError::Storage(StorageError::NotFound(
+                            "TracedEntryPoint".to_string(),
+                            format!("{entry_point_id:?} with params {tracing_param:?}"),
```
louise-poole: `entry_point_id` is a str so there is no 'debug' display for it - you can just use it directly in the string.
```suggestion
                            format!("{entry_point_id} with params {tracing_param:?}"),
```

tycho-indexer/src/services/rpc.rs
```diff
@@ -718,6 +732,144 @@ where
...
+                    .entry_point
+                    .external_id
+                    .as_str();
+                let tracing_param = &entry_point_with_params.params;
+                let results_by_param = traced_entry_points
+                    .get(entry_point_id)
+                    .ok_or_else(|| {
+                        RpcError::Storage(StorageError::NotFound(
+                            "TracedEntryPoint".to_string(),
+                            format!("{entry_point_id:?}"),
```
```suggestion
                            entry_point_id.to_string(),
```

tycho-indexer/src/services/rpc.rs
```diff
@@ -718,6 +732,144 @@ where
...
+                let tracing_result = results_by_param
+                    .get(tracing_param)
+                    .ok_or_else(|| {
+                        RpcError::Storage(StorageError::NotFound(
+                            "TracedEntryPoint".to_string(),
+                            format!("{entry_point_id:?} with params {tracing_param:?}"),
+                        ))
+                    })?;
+
+                pairs.push((entry_point_with_params.clone().into(), tracing_result.clone().into()));
```
louise-poole: We should be able to avoid at least 1 clone here:
```suggestion
                pairs.push((entry_point_with_params.into(), tracing_result.clone().into()));
```

tycho-indexer/src/services/rpc.rs
```diff
@@ -927,11 +1079,11 @@ pub async fn protocol_state<G: Gateway>(
     post,
     path = "/v1/protocol_systems",
     responses(
-        (status = 200, description = "OK", body = ProtocolSystemsRequestResponse),
+    (status = 200, description = "OK", body = ProtocolSystemsRequestResponse),
     ),
     request_body = ProtocolSystemsRequestBody,
     security(
-         ("apiKey" = [])
+    ("apiKey" = [])
```
louise-poole: Can you remove this format change? All other endpoints are formatted with the tab there, let's stay consistant.

tycho-indexer/src/services/rpc.rs
```diff
@@ -967,6 +1119,53 @@ pub async fn protocol_systems<G: Gateway>(
...
+#[utoipa::path(
+    post,
+    path = "/v1/traced_entry_points",
+    responses(
+    (status = 200, description = "OK", body = TracedEntryPointRequestResponse),
+    ),
+    request_body = TracedEntryPointRequestBody,
+    security(
+    ("apiKey" = [])
+    ),
```
louise-poole: And add the tabs here? It's easier to read with them.

tycho-indexer/src/services/rpc.rs
```diff
@@ -718,6 +732,144 @@ where
...
+
+        trace!(
+            traced_entry_points = ?traced_entry_points,
+            "Retrieved traced entry points entry points from database."
+        );
+
+        let mut traced_entry_points_by_component = HashMap::new();
+
+        for (component_id, entry_points_set) in entry_points_tracing_params_data.entity {
+            let mut pairs = Vec::new();
```
louise-poole: Small optimisation suggestion: if you know the length of a Vec before you populate it, you can already set the Vec capacity when you instantiate it. This allows the compiler to know the Vec size from the start and it doesn't need to reallocate the vec as it grows. 
```suggestion
            let mut pairs = Vec::with_capacity(entry_points_set.len());
```
Optimisation isn't crazy important here so this small improvement is just me finding a teaching moment and isn't actually necessary.

tycho-indexer/src/services/rpc.rs
zizou0x: I don't think this is equivalent, I think one uses `Debug` and the other uses `Display`

tycho-indexer/src/services/rpc.rs
zizou0x: We don't really have a similar case in the DCI component, we just load everything that the gateway give us, no assumption about this

So I think here we should follow the same behavior, we should just return the result that we have and ignore if it's missing for a params

louise-poole: Hmmm... considering Alan is planning some features for tycho-simulation that is endpoint/endpoint param aware maybe we should still return the param, but with empty results?

tycho-common/src/storage.rs
louise-poole: Ahhh - this is from Flo's PR no? You cherry picked the commit I assume? A bit confusing to see changes here that you didn't do 👀 😂 I'll leave this comment on Flo's PR instead then.

tycho-indexer/src/services/rpc.rs
tvinagre: Ah yeah, true true

tvinagre: You're right, thanks!

tycho-indexer/src/services/rpc.rs
tamaralipows: Oh, cool! Good to know!

tycho-indexer/src/services/rpc.rs
tamaralipows: Yikes, idk why my nightly did this? Will manually remove + double check the others

tycho-indexer/src/services/rpc.rs
tamaralipows: Ahhhh but in the end, after the changes discussed this morning, we won't know the capacity (in case tracing results are not found for specific entry points). I'll keep it in mind for the future though.

louise-poole: You could actually still use it here. The capacity can be a best-guess, does not need to be 100% accurate. And the likelihood of there being a failed trace is hopefully low enough that your best guess is accurate.

tycho-indexer/src/services/rpc.rs
```diff
@@ -758,10 +758,8 @@ where
     ) -> Result<dto::TracedEntryPointRequestResponse, RpcError> {
         let pagination_params: PaginationParams = (&request.pagination).into();
 
-        // For consistency, we pre-apply pagination to ensure it's predetermined which components
-        // will appear on which page.
-        // Apply pagination to the contract addresses. This is done so that we can determine which
-        // contracts were not returned from the db and get them from the buffer instead.
+        // For consistency, we pre-apply pagination to the component ids to ensure it's
+        // predetermined which components ill appear on which page.
```
```suggestion
        // predetermined which components will appear on which page.
```

tycho-indexer/src/services/rpc.rs
```diff
@@ -835,25 +833,21 @@ where
...
-                    .get(tracing_param)
-                    .ok_or_else(|| {
-                        RpcError::Storage(StorageError::NotFound(
-                            "TracedEntryPoint".to_string(),
-                            format!("{entry_point_id:?} with params {tracing_param:?}"),
-                        ))
-                    })?;
 
-                pairs.push((entry_point_with_params.clone().into(), tracing_result.clone().into()));
+                if let Some(results_for_entry_point) = traced_entry_points.get(entry_point_id) {
```
louise-poole, line 837: Can we also warn on the `else` for this? It should technically never happen (have an entrypoint with no params) but if it does it would be nice to have it logged/warned.

louise-poole (commented):
Looking goooood! 🔥

PR #583, opened 2025-05-21
feat: add pagination to EntryPointGateway

zizou0x:
This commit updates the EntryPointGateway trait to include pagination parameters for the get_entry_points and get_entry_points_tracing_params methods, returning results wrapped in a WithTotal struct. This is needed by the RPC methods.

### DIFF
--- a/tycho-common/src/models/blockchain.rs
+++ b/tycho-common/src/models/blockchain.rs
@@ -700,13 +700,15 @@ pub mod fixtures {
                 .len(),
             2
         );
+        let mut expected_entry_points = changes1
+            .entrypoints
+            .values()
+            .flat_map(|ep| ep.iter())
+            .map(|ep| ep.signature.clone())
+            .collect::<Vec<_>>();
+        expected_entry_points.sort();
         assert_eq!(
-            changes1
-                .entrypoints
-                .values()
-                .flat_map(|ep| ep.iter())
-                .map(|ep| ep.signature.clone())
-                .collect::<Vec<_>>(),
+            expected_entry_points,
             vec!["function()".to_string(), "function_2()".to_string()],
         );
     }
--- a/tycho-common/src/storage.rs
+++ b/tycho-common/src/storage.rs
@@ -512,37 +512,56 @@ impl EntryPointFilter {
 #[async_trait]
 pub trait EntryPointGateway {
     /// Upserts a list of entry points into the database.
+    ///
+    /// * `entry_points` - The map of component ids to their entry points to upsert.
     async fn upsert_entry_points(
         &self,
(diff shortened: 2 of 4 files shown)


tycho-common/src/models/blockchain.rs
```diff
@@ -700,13 +700,15 @@ pub mod fixtures {
...
+            .collect::<Vec<_>>();
+        expected_entry_points.sort();
         assert_eq!(
-            changes1
-                .entrypoints
-                .values()
-                .flat_map(|ep| ep.iter())
-                .map(|ep| ep.signature.clone())
-                .collect::<Vec<_>>(),
+            expected_entry_points,
```
zizou0x, line 711: This is not related to this PR, this test was sometimes failing due to the order not being deterministic in collect()

tycho-common/src/storage.rs
```diff
@@ -512,37 +512,56 @@ impl EntryPointFilter {
...
+    /// * `traced_entry_points` - The list of traced entry points to upsert.
     async fn upsert_traced_entry_points(
         &self,
         traced_entry_points: &[TracedEntryPoint],
     ) -> Result<(), StorageError>;
 
     /// Retrieves all tracing results for a set of entry points from the database.
+    ///
+    /// * `entry_points` - The set of entry points to retrieve tracing results for.
     async fn get_traced_entry_points(
```
louise-poole, line 570: Does this one not need pagination too? Or is it not used on the RPC? I thought it was...

tycho-storage/src/postgres/entry_point.rs
```diff
@@ -24,14 +24,14 @@ use super::{
 };
 
 impl PostgresGateway {
-    /// Insert entry points into the database.
+    /// Upsert entry points into the database.
```
louise-poole: Considering we use 'on conflict do nothing' and will never change the target/signature (which is the entrypoint), I would keep this as 'insert'. I know you're probably using 'update' to refer to linking a new component to an existing entrypoint, but I think it's fine to keep it as 'insert' for that (we're inserting a new link and not updating existing rows). The function name `upsert_entry_points` is misleading because the entrypoints themselves cannot be updated.

tycho-common/src/storage.rs
zizou0x: Not really, this one takes a `HashSet<EntryPointId>`, no filter involved. This one is meant to be called after calling the one with pagination

tycho-storage/src/postgres/entry_point.rs
zizou0x: Hum ok makes sense

louise-poole (approved):

PR #584, opened 2025-05-23
Lp/eng 4439 dci on extractor

louise-poole:


tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -87,7 +90,15 @@ where
...
         token_pre_processor: T,
         post_processor: Option<fn(BlockChanges) -> BlockChanges>,
+        dci_plugin: Option<DCI>,
     ) -> Result<Self, ExtractionError> {
+        let dci_plugin = if let Some(mut plugin) = dci_plugin {
+            plugin.initialize().await?;
+            Some(plugin)
+        } else {
+            None
+        };
```
zizou0x: Can't we do the same thing as below here? 

```
        if let Some(dci_plugin) = &mut self.dci_plugin {
            dci_plugin
                .process_block_update(&mut msg)
                .await?;
        }

File: tycho-indexer/src/extractor/runner.rs [hunk 2]
```diff
@@ -405,8 +418,13 @@ impl ExtractorBuilder {
         self
     }
 
+    pub fn set_dci_plugin(mut self, dci_type: DCIType) -> Self {
```
zizou0x: We need to extend the config to be able to set this

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -614,7 +628,7 @@ where
     #[allow(deprecated)]
     #[instrument(skip_all, fields(block_number))]
     async fn handle_tick_scoped_data(
-        &self,
+        &mut self,
```
zizou0x: This is a huge change that involves a lot, can't we wrap DCI itself inside a `Arc<Mutex<>>` instead?

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -26,8 +28,19 @@ type StorageLocation = (Address, StoreKey);
 #[allow(dead_code)] // Clippy thinks this is dead code, but it's used below for clarity
 type TxVecIndex = usize;
 
+#[automock]
+#[async_trait]
+pub trait DynamicContractIndexerTrait: Send + Sync {
```
zizou0x: Do you like this name? I'm not a big fan honestly

tycho-indexer/src/extractor/protocol_extractor.rs
louise-poole: Unfortunately not - if I don't make the `if let` return the plugin I get errors later because it was partially moved here inside the `if let` statement.

tycho-indexer/src/extractor/runner.rs
louise-poole: Ah - you think it should be part of `ExtractorConfig`?

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
louise-poole: Nope, but was not feeling very creative today 😂 Super open to suggestions

tycho-indexer/src/extractor/protocol_extractor.rs
louise-poole: Yay - I like this idea. Will give it a try

tycho-indexer/src/extractor/runner.rs
zizou0x: Yes most likely

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
kayibal: Please don't mention trait in the name. I think we don't have to restrict this behaviour to dynamic contract indexing at all. I don't have a specific name yet in mind but maybe: just `AdditionalDataPlugin` or similar. Maybe talk with chatgpt about it to get some good suggestions on commonly used names here.

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -26,8 +28,19 @@ type StorageLocation = (Address, StoreKey);
...
+    async fn initialize(&mut self) -> Result<(), ExtractionError>;
```
kayibal: Can't this be done within the constructor? Is it required on the trait?

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
```diff
@@ -26,8 +28,19 @@ type StorageLocation = (Address, StoreKey);
...
+    async fn process_block_update(
+        &mut self,
+        block_changes: &mut BlockChanges,
+    ) -> Result<(), ExtractionError>;
+    async fn process_revert(&mut self, target_block: u64) -> Result<(), ExtractionError>;
```
kayibal: Do you think number is better than hash? Not sure just asking

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -87,7 +90,15 @@ where
         protocol_types: HashMap<String, ProtocolType>,
         token_pre_processor: T,
         post_processor: Option<fn(BlockChanges) -> BlockChanges>,
+        dci_plugin: Option<DCI>,
     ) -> Result<Self, ExtractionError> {
+        let dci_plugin = if let Some(mut plugin) = dci_plugin {
+            plugin.initialize().await?;
```
kayibal: If we don't pass any extractor specific info here I don't think we need this..

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -702,6 +716,15 @@ where
             }
         }
 
+        // Send message to DCI plugin
+        if let Some(dci_plugin) = &self.dci_plugin {
+            dci_plugin
+                .lock()
+                .await
+                .process_block_update(&mut msg)
+                .await?;
```
kayibal, line 721: I'd prefer if we didn't await for this here but instead used a task to process things async

tycho-indexer/src/extractor/protocol_extractor.rs
kayibal: I think initially I intended to pass like a protocol cache or something in here?

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
tvinagre: Good point. Since we're handling reverts, hash seems more adequate

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
tvinagre: It's an async function that loads the data to the cache - can be removed from the trait but I don't think it should be done within the constructor

tycho-indexer/src/extractor/runner.rs
```diff
@@ -482,8 +500,45 @@ impl ExtractorBuilder {
...
+                                ExtractionError::Setup(format!(
+                                    "Failed to create account extractor for {rpc_url}: {err}"
+                                ))
+                            })?;
+                    let tracer = EVMEntrypointService::try_from_url(rpc_url).map_err(|err| {
+                        ExtractionError::Setup(format!(
+                            "Failed to create entrypoint tracer for {rpc_url}: {err}"
+                        ))
+                    })?;
+                    DynamicContractIndexer::new(
```
tvinagre: Could we initialize it here instead of inside the Extractor?

tycho-indexer/src/extractor/protocol_extractor.rs
louise-poole: This comment is better suited on the [DCI PR](https://github.com/propeller-heads/tycho-indexer/pull/557) then no? That's where this interface is created.

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
louise-poole: This comment is also better suited for the [DCI PR](https://github.com/propeller-heads/tycho-indexer/pull/557) where the interface is defined.

tycho-indexer/src/extractor/protocol_extractor.rs
louise-poole: We decided on this first iteration of DCI we will keep it simple and handle everything sequentially on the extractor. The other block processing that happens here (creating tokens etc) is very quick (takes a few ms) while the DCI processing is slow (can take multiple minutes). The relative speedup of handling these in parallel is almost negligible here. I can leave a comment that there is room for potential speed up here, but don't think it's worth implementing just yet.

tycho-indexer/src/extractor/protocol_extractor.rs
zizou0x: What would be the alternative? We could pass the `dci_type` here and do this initialization in DynamicContractIndexer::new() but I think I like it like this honestly

this way we keep `DynamicContractIndexer::new()` sync and fast

tycho-indexer/src/extractor/dynamic_contract_indexer.rs
zizou0x: I can change in my PR but I think it's easier if you change it here directly to avoid having to rebase? 

This function isn't implemented anyway so you shouldn't have any conflict

tycho-indexer/src/cli.rs
```diff
@@ -141,6 +141,12 @@ pub struct RunSpkgArgs {
     /// Block number to initialize the accounts at
     #[clap(long, default_value = "0")]
     pub initialization_block: i64,
+
+    /// DCI plugin to use
+    ///
+    /// Optional. If not provided, the extractor will not use DCI.
```
zizou0x: Should we maintain a list for possibilities in the docs?

zizou0x (approved):
LGTM thanks!

PR #585, opened 2025-05-27
chore: add empty tick detection script

tvinagre:


scripts/identify_empty_ticks.py
```diff
@@ -0,0 +1,175 @@
...
+
+def hex_to_signed_int(hex_str):
+    # Remove '0x' prefix and convert to bytes
+    hex_val = hex_str[2:]
+    bytes_val = bytes.fromhex(hex_val)
+    # Convert bytes to signed integer (big-endian)
+    return int.from_bytes(bytes_val, byteorder="big", signed=True)
+
+
+def has_empty_ticks(attributes: dict[str, str]):
```
louise-poole: The logic looks good, but the naming is very misleading - we're not looking for empty ticks at all. We're looking for ticks that should have been deleted but weren't leading to a non-zero net tick liquidity.

scripts/identify_empty_ticks.py
```diff
@@ -0,0 +1,175 @@
...
+log.basicConfig(
+    format="{asctime} - {levelname} - {message}",
+    style="{",
+    datefmt="%Y-%m-%d %H:%M",
+    level=log.INFO)
+
+"""
+This script identifies protocol components with empty ticks in various exchanges. 
+Used to find components that have no net liquidity in their ticks, which can indicate issues with the protocol's state.
+"""
```
zizou0x: This is not quite true, we identify if the sum of all ticks is 0. By definition it should be 0, else it means we have a wrong state (could be a missing tick or a dangling tick)

scripts/identify_empty_ticks.py
```diff
@@ -0,0 +1,175 @@
...
+        except requests.RequestException as e:
+            log.error(f"Failed to fetch data for {protocol_system}: {e}")
+            continue
+
+        log.info(f"Fetched {len(protocol_states)} protocol systems")
+        empty_tick_components = identify_empty_ticks(protocol_states)
+
+        log.info(f"Identified {len(empty_tick_components)} components with empty ticks:")
+        for protocol in empty_tick_components:
+            log.info(protocol)
```
louise-poole: Any way we can trigger an alert on slack or something? Or what is the plan with this script? Will it be added as a temp cronjob or are we to manually run it daily? If we plan to do this manually then just logging this is good enough.

scripts/identify_empty_ticks.py
```diff
@@ -0,0 +1,175 @@
+import os
```
zizou0x: I'd call the file "validate_tick_liquidities_sum.py" or "validate_ticks_state.py" something like this

zizou0x (approved):

scripts/identify_empty_ticks.py
```diff
@@ -0,0 +1,175 @@
...
+            empty_ticks.append(component["component_id"])
+
+    return empty_ticks
+
+
+@click.command()
+@click.option("--tvl_gt", type=int, default=1,)
+@click.option("--exchange", type=StringListParamType(','), default="uniswap_v3,uniswap_v4,ekubo_v2,pancakeswap_v3",
+              help="Comma-separated list of exchanges to filter by")
+def main(tvl_gt, exchange: list[str]):
```
louise-poole: We should also be able to run this on other chains no? Base and Unichain too?

tvinagre: tru tru

scripts/identify_empty_ticks.py
tvinagre: I didn't want to spend more time triggering slack alerts and adding to the cluster, as we have other priorities right now. Since we'll run this only for a couple of weeks, I'll leave it as it is and maybe add it as a cronjob once we have more time

PR #586, opened 2025-05-27
feat: extend tycho-client to handle dci data

louise-poole:
This PR covers the following:
- extends RPCClient to include `get_traced_entry_points` endpoint and support auto-pagination
- extends the emitted message models to include all entrypoint data (entrypoints, params and traces)
- extends the component_tracker to track entrypoints and monitor which contracts to filter accordingly
- extends the client's protocol synchronizer to use the updated component_tracker and entrypoint data where necessary

### DIFF
--- a/tycho-client/src/feed/component_tracker.rs
+++ b/tycho-client/src/feed/component_tracker.rs
@@ -2,7 +2,7 @@ use std::collections::{HashMap, HashSet};
 
 use tracing::{debug, instrument, warn};
 use tycho_common::{
-    dto::{BlockChanges, Chain, ProtocolComponent, ProtocolComponentsRequestBody},
+    dto::{BlockChanges, Chain, DCIData, ProtocolComponent, ProtocolComponentsRequestBody},
     Bytes,
 };
 
@@ -74,7 +74,7 @@ impl ComponentFilter {
     }
 }
 
-/// Helper struct to store which components are being tracked atm.
+/// Helper struct to determine which components and contracts are being tracked atm.
 pub struct ComponentTracker<R: RPCClient> {
     chain: Chain,
     protocol_system: String,
@@ -82,6 +82,9 @@ pub struct ComponentTracker<R: RPCClient> {
     // We will need to request a snapshot for components/contracts that we did not emit as
     // snapshot for yet but are relevant now, e.g. because min tvl threshold exceeded.
     pub components: HashMap<String, ProtocolComponent>,
+    /// Map of entrypoint id to a tuple of (set of component ids for components that have this
+    /// entrypoint, set of detected contracts for the entrypoint)
+    pub entrypoints: HashMap<String, (HashSet<String>, HashSet<Bytes>)>,
     /// Derived from tracked components. We need this if subscribed to a vm extractor because
     /// updates are emitted on a contract level instead of a component level.
     pub contracts: HashSet<Bytes>,
@@ -101,9 +104,12 @@ where
             components: Default::default(),
             contracts: Default::default(),
             rpc_client: rpc,
+            entrypoints: Default::default(),
         }
     }
-    /// Retrieve all components that belong to the system we are extracing and have sufficient tvl.
+
+    /// Retrieves all components that belong to the system we are streaming that have sufficient
+    /// tvl. Also detects which contracts are relevant for simulating on those components.
     pub async fn initialise_components(&mut self) -> Result<(), RPCError> {
         let body = match &self.filter.variant {
             ComponentFilterVariant::Ids(ids) => ProtocolComponentsRequestBody::id_filtered(
@@ -119,7 +125,6 @@ where
                 )
             }
         };
-
         self.components = self
             .rpc_client
             .get_protocol_components_paginated(&body, 500, 4)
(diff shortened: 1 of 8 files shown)


tycho-client/src/feed/component_tracker.rs
```diff
@@ -74,14 +74,17 @@ impl ComponentFilter {
...
 pub struct ComponentTracker<R: RPCClient> {
     chain: Chain,
     protocol_system: String,
     filter: ComponentFilter,
     // We will need to request a snapshot for components/contracts that we did not emit as
     // snapshot for yet but are relevant now, e.g. because min tvl threshold exceeded.
     pub components: HashMap<String, ProtocolComponent>,
+    /// Map of entrypoint id to a tuple of (set of component ids for components that have this
+    /// entrypoint, set of detected contracts for the entrypoint)
+    pub entrypoints: HashMap<String, (HashSet<String>, HashSet<Bytes>)>,
```
tvinagre: If the key is the ComponentId, I like the idea of using a custom type for it instead of String - we've been doing that on the DCI implementation and I think it eases code comprehension, what do u think?

tvinagre: I also don't like the `value` type much.. Maybe we could introduce a new struct ?
```rust
#[derive(Debug, Default, Clone)]
pub struct EntrypointRelation {
    /// Component IDs that have this entrypoint
    pub component_ids: HashSet<String>,
    /// Contract addresses detected for this entrypoint
    pub contract_addresses: HashSet<Bytes>,
}
```

tycho-client/src/feed/component_tracker.rs
```diff
@@ -128,16 +133,60 @@ where
...
+        // Identify entrypoints linked to the given components
+        for (_, (components, contracts)) in self.entrypoints.iter() {
+            if !components.is_disjoint(&tracked_component_ids) {
+                self.contracts
+                    .extend(contracts.iter().cloned());
+            }
+        }
     }
 
     /// Add a new component to be tracked
```
tvinagre: New components

tycho-client/src/feed/component_tracker.rs
```diff
@@ -174,22 +230,50 @@ where
...
+        // Refresh the tracked contracts list. This is more reliable and efficient than directly
+        // removing contracts from the list because some contracts are shared between components.
+        self.initialise_contracts();
+
+        debug!(n_components = removed_components.len(), "StoppedTracking");
+        removed_components
+    }
+
+    /// Updates the tracked entrypoints and contracts based on the given DCI data.
+    pub fn process_entrypoints(&mut self, dci_data: &DCIData) -> Result<(), RPCError> {
```
tvinagre: Maybe DCIUpdate instead of DCIData?

tycho-client/src/feed/synchronizer.rs
```diff
@@ -204,27 +195,34 @@ where
...
-                it.into_iter()
-                    .cloned()
-                    .collect::<Vec<_>>()
-            })
-            .unwrap_or_else(|| tracked_components.get_tracked_component_ids());
-
-        let component_ids = request_ids
-            .iter()
-            .collect::<HashSet<_>>();
+        let component_ids: Vec<_> = match ids {
```
tvinagre, line 198: way more elegant :)

tycho-client/src/feed/component_tracker.rs
zizou0x: I think we already have a type for each of these three, I agree it would be clearer to use them

tycho-client/src/feed/component_tracker.rs
```diff
@@ -174,22 +230,50 @@ where
...
+        // Update detected contracts for entrypoints
+        for (entrypoint, traces) in &dci_data.trace_results {
+            self.entrypoints
+                .entry(entrypoint.clone())
+                .or_insert_with(|| (HashSet::new(), HashSet::new()))
+                .1
```
zizou0x: I agree with Thales' comment about introducing a struct instead of using a tuple, would probably make this clearer

tycho-common/src/models/blockchain.rs
```diff
@@ -175,6 +175,13 @@ impl BlockScoped for BlockAggregatedChanges {
     }
 }
 
+#[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq)]
+pub struct DCIData {
+    pub new_entrypoints: HashMap<ComponentId, HashSet<EntryPoint>>,
+    pub new_entrypoint_params: HashMap<EntryPointId, HashSet<(TracingParams, Option<ComponentId>)>>,
+    pub trace_results: HashMap<EntryPointId, TracingResult>,
+}
```
zizou0x: We most likely need a way to notify any deletion of these too for reorgs

tycho-client/src/feed/component_tracker.rs
```diff
@@ -128,16 +133,60 @@ where
...
     }
 
-    fn update_contracts(&mut self) {
-        self.contracts.extend(
-            self.components
-                .values()
-                .flat_map(|comp| comp.contract_ids.iter().cloned()),
-        );
+    /// Initialise the tracked contracts list from tracked components and their entrypoints
+    fn initialise_contracts(&mut self) {
```
zizou0x: I think this name is misleading, this is used in `stop_tracking` to refresh the list, this should be reflected in the name

Also small comment but it will probably not be needed if we rename it, this should be written `initialize` for consistency with everything else

tycho-client/src/feed/component_tracker.rs
```diff
@@ -128,16 +133,60 @@ where
...
+        // efficient lookup.
+        let tracked_component_ids = components
+            .iter()
+            .filter(|&id| self.components.contains_key(id))
+            .map(|s| s.to_string())
+            .collect::<HashSet<_>>();
+
+        // Add contracts from the components
+        for comp in &tracked_component_ids {
+            if let Some(component) = self.components.get(comp) {
```
zizou0x: I think it's better to `.expect` here, we know that it's here because we filtered by `self.components.contains_key` above. If it's not here it means something really unexpected happened and I feel it's unsafe to ignore and continue in this case

tycho-client/src/feed/component_tracker.rs
```diff
@@ -128,16 +133,60 @@ where
...
+
+    /// Update the tracked contracts list with contracts associated with the given components
+    fn update_contracts(&mut self, components: Vec<String>) {
+        // Only process components that are actually being tracked. Convert to HashSet for
+        // efficient lookup.
+        let tracked_component_ids = components
+            .iter()
+            .filter(|&id| self.components.contains_key(id))
+            .map(|s| s.to_string())
+            .collect::<HashSet<_>>();
```
zizou0x: We also can avoid these clones no? With something like this (might need to slightly adapt the filter)

```suggestion
        // Only process components that are actually being tracked. Convert to HashSet for
        // efficient lookup.
        let tracked_component_ids = components
            .into_iter()
            .filter(|&id| self.components.contains_key(id))
            .collect::<HashSet<_>>();
```

tycho-client/src/feed/component_tracker.rs
zizou0x: I wonder if it wouldn't be clearer with two HashMaps instead of a single complex one

```
entrypoints_to_components: HashMap<EntrypointId,HashSet<ComponentId>>
entrypoints_to_contracts: HashMap<EntrypointId,HashSet<Address>>
```

zizou0x (commented):
Thanks!

Overall looks good, added a few small comments but nothing huge

tycho-client/src/feed/component_tracker.rs
louise-poole: Yeah - the custom types are all defined in tycho-common so it's easy to use them here too. We didn't use them at all in tycho-client which is why I didn't use them here right now. But I agree - much easier to read/understand type definitions with the custom types 👍

louise-poole: > I wonder if it wouldn't be clearer with two HashMaps instead of a single complex one

The complex type avoids the need for 2 look-ups which is why I opted for it. But this isn't a big performance save so if it's easier to understand separately then I'd rather do it.

louise-poole: Value? You referring to the tuple? It felt excessive to add a new struct for a simple tuple-like relation that is used internally only, but it's for sure easier to understand. Will implement this 👍

tycho-client/src/feed/component_tracker.rs
louise-poole: Yes I like that better 💃  'DCIData' wasn't sitting well with me...

tycho-common/src/models/blockchain.rs
louise-poole: Yup - I started a thread for this before, but we came to the conclusion that for now it wasn't vital. The info is only used to filter out contract updates on the client side... as long as that filter covers contracts actually needed then we're fine, even if it lets more contracts through than are strictly needed (the data is already client-side so doesn't have a big impact by increasing network load etc).

tycho-client/src/feed/component_tracker.rs
louise-poole: I'll name it `reinitialize` because it completely rebuilds the list from scratch (initializes it). 'refresh' sounds too lightweight compared to that...

zizou0x (approved):
Looks perfect! Thanks

PR #587, opened 2025-05-28
feat: traced entrypoints RPC setter

tamaralipows:
**Current status**

# Biggest changes
- **`add_entry_points` method**
- **Access Control**
- **`DirectGatway` implementation in order to fix `Unexpected storage error: Usage error: No transaction started`**

If reviewing by commit, beware that some future commits may contradict earlier commits. Tried my best here while also not messing with commit history in between reviews - sorry.

It would have been nice to add some mock protocol components into the DB for running the RPC locally for testing, but I ran out of time here while attempting to do so, so I won't.

### Proof that access control works:
Request with proper key (the empty reply is expected):
```(base) [14:05:11] [~/Code/tycho-indexer] >>> curl -v POST \
  http://localhost:4242/v1/add_entry_points \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -H 'authorization: test-key' \
  -d '{
  "chain": "ethereum",
  "block_hash": "0x354c90a0a98912aff15b044bdff6ce3d4ace63a6fc5ac006ce53c8737d425ab2",
  "protocol_system": "Balancer",
  "entry_points_with_tracing_data": [
    [
      "0x0000000000000000000000000000000000000001",
      [
        {
          "entry_point": {
            "external_id": "0xEdf63cce4bA70cbE74064b7687882E71ebB0e988:getRate()",
            "target": "0xEdf63cce4bA70cbE74064b7687882E71ebB0e988",
            "signature": "getRate()"
          },
          "params": {
            "method": "rpctracer",
            "caller": null,
            "calldata": "0x29ae8114"
          }
        },
        {
          "entry_point": {
            "external_id": "0x8f4E8439b970363648421C692dd897Fb9c0Bd1D9:getRate()",
            "target": "0x8f4E8439b970363648421C692dd897Fb9c0Bd1D9",
            "signature": "getRate()"
          },
          "params": {
            "method": "rpctracer",
            "caller": null,
            "calldata": "0x29ae8114"
          }
        }
      ]
    ]
  ]
}'
* Could not resolve host: POST
* Closing connection 0
curl: (6) Could not resolve host: POST
*   Trying 127.0.0.1:4242...
* Connected to localhost (127.0.0.1) port 4242 (#1)
> POST /v1/add_entry_points HTTP/1.1
> Host: localhost:4242
> User-Agent: curl/8.1.2
> accept: application/json
> Content-Type: application/json
> authorization: test-key
> Content-Length: 1032
> 
< HTTP/1.1 404 Not Found
< content-length: 86
< vary: Origin, Access-Control-Request-Method, Access-Control-Request-Headers
< date: Tue, 03 Jun 2025 18:17:23 GMT
< 
* Connection #1 to host localhost left intact
Could not find ProtocolComponent with id `0x0000000000000000000000000000000000000001`!% 
```

Request with improper key:
```
(base) [1:38:40] [~/Code/tycho-simulation] >>> curl -v POST \
  http://localhost:4242/v1/add_entry_points \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -H 'authorization: sampletoken' \
  -H 'X-API-Key: wrong-key' \
  -d '{
  "chain": "ethereum",
  "block_hash": "0x354c90a0a98912aff15b044bdff6ce3d4ace63a6fc5ac006ce53c8737d425ab2",
  "protocol_system": "Balancer",
  "entry_points_with_tracing_data": [
    [
      "0x0000000000000000000000000000000000000001",
      [
        {
          "entry_point": {
            "external_id": "0xEdf63cce4bA70cbE74064b7687882E71ebB0e988:getRate()",
            "target": "0xEdf63cce4bA70cbE74064b7687882E71ebB0e988",
            "signature": "getRate()"
          },
          "params": {
            "method": "rpctracer",
            "caller": null,
            "calldata": "0x29ae8114"
          }
        },
        {
          "entry_point": {
            "external_id": "0x8f4E8439b970363648421C692dd897Fb9c0Bd1D9:getRate()",
            "target": "0x8f4E8439b970363648421C692dd897Fb9c0Bd1D9",
            "signature": "getRate()"
          },
          "params": {
            "method": "rpctracer",
            "caller": null,
            "calldata": "0x29ae8114"
          }
        }
      ]
    ]
  ]
}'
* Could not resolve host: POST
* Closing connection 0
curl: (6) Could not resolve host: POST
*   Trying 127.0.0.1:4242...
* Connected to localhost (127.0.0.1) port 4242 (#1)
> POST /v1/add_entry_points HTTP/1.1
> Host: localhost:4242
> User-Agent: curl/8.1.2
> accept: application/json
> Content-Type: application/json
> authorization: sampletoken
> X-API-Key: wrong-key
> Content-Length: 1032
> 
< HTTP/1.1 401 Unauthorized
< content-length: 13
< vary: Origin, Access-Control-Request-Method, Access-Control-Request-Headers
< date: Tue, 03 Jun 2025 05:39:06 GMT
< 
* Connection #1 to host localhost left intact
Access denied%    
```

### DIFF
--- a/tycho-common/src/dto.rs
+++ b/tycho-common/src/dto.rs
@@ -19,6 +19,7 @@ use uuid::Uuid;
 
 use crate::{
     models,
+    models::ComponentId,
     serde_primitives::{
         hex_bytes, hex_bytes_option, hex_hashmap_key, hex_hashmap_key_value, hex_hashmap_value,
     },
@@ -959,7 +960,7 @@ pub struct ProtocolComponentsRequestBody {
     pub protocol_system: String,
     /// Filter by component ids
     #[serde(alias = "componentAddresses")]
-    pub component_ids: Option<Vec<String>>,
+    pub component_ids: Option<Vec<ComponentId>>,
     /// The minimum TVL of the protocol components to return, denoted in the chain's
     /// native token.
     #[serde(default)]
@@ -1328,7 +1329,7 @@ impl ProtocolSystemsRequestResponse {
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize, Default)]
 pub struct DCIUpdate {
     /// Map of component id to the new entrypoints associated with the component
-    pub new_entrypoints: HashMap<String, HashSet<EntryPoint>>,
+    pub new_entrypoints: HashMap<ComponentId, HashSet<EntryPoint>>,
     /// Map of entrypoint id to the new entrypoint params associtated with it (and optionally the
     /// component linked to those params)
     pub new_entrypoint_params: HashMap<String, HashSet<(TracingParams, Option<String>)>>,
@@ -1344,7 +1345,7 @@ pub struct TracedEntryPointRequestBody {
     /// ReorgBuffers
     pub protocol_system: String,
     /// Filter by component ids
-    pub component_ids: Option<Vec<String>>,
+    pub component_ids: Option<Vec<ComponentId>>,
     /// Max page size supported is 100
     #[serde(default)]
     pub pagination: PaginationParams,
@@ -1438,7 +1439,8 @@ impl From<models::blockchain::TracingResult> for TracingResult {
 pub struct TracedEntryPointRequestResponse {
     /// Map of protocol component id to a list of a tuple containing each entry point with its
     /// tracing parameters and its corresponding tracing results.
-    pub traced_entry_points: HashMap<String, Vec<(EntryPointWithTracingParams, TracingResult)>>,
+    pub traced_entry_points:
+        HashMap<ComponentId, Vec<(EntryPointWithTracingParams, TracingResult)>>,
     pub pagination: PaginationResponse,
 }
 
@@ -1498,14 +1500,15 @@ pub struct AddEntryPointRequestBody {
     #[serde(default)]
     pub block_hash: Bytes,
     /// The map of component ids to their tracing params to insert
-    pub entry_points_with_tracing_data: Vec<(String, Vec<EntryPointWithTracingParams>)>,
+    pub entry_points_with_tracing_data: Vec<(ComponentId, Vec<EntryPointWithTracingParams>)>,
 }
 
 #[derive(Serialize, PartialEq, ToSchema, Eq, Clone, Debug, Deserialize)]
 pub struct AddEntryPointRequestResponse {
     /// Map of protocol component id to a list of a tuple containing each entry point with its
     /// tracing parameters and its corresponding tracing results.
-    pub traced_entry_points: HashMap<String, Vec<(EntryPointWithTracingParams, TracingResult)>>,
+    pub traced_entry_points:
+        HashMap<ComponentId, Vec<(EntryPointWithTracingParams, TracingResult)>>,
 }
 
 #[cfg(test)]


tycho-common/src/models/blockchain.rs
```diff
@@ -383,6 +384,12 @@ impl EntryPoint {
     }
 }
 
+impl From<dto::EntryPoint> for EntryPoint {
+    fn from(value: dto::EntryPoint) -> Self {
+        Self { external_id: value.external_id, target: value.target, signature: value.signature }
+    }
+}
+
```
tamaralipows, line 395: Today I learnt that rust allows circular imports. I loooooooooovvveeee rust

tycho-common/src/dto.rs
```diff
@@ -1428,9 +1428,16 @@ pub struct TracedEntryPointRequestResponse {
...
 pub struct AddEntrypointRequestBody {
-    pub entrypoints_with_tracing_data: Vec<(String, Vec<EntryPointWithTracingParams>)>,
+    #[serde(default)]
     pub chain: Chain,
+    /// The map of entry point ids to their tracing params to insert and optionally a component id
+    /// used for debugging only.
+    // TODO fix this warning
+    #[allow(clippy::type_complexity)]
+    pub entry_points_with_tracing_data:
+        Vec<(String, Vec<(EntryPointWithTracingParams, Option<String>)>)>,
```
tamaralipows: I can't see Jira right now because someone forgot to pay the bill lol - was the first param meant to be the component id and not the entry point ids? That would make a lot more sense haha, since the entry point id right now is kindof useless. Then I could remove the innermost Option<String> which should represent the component id

zizou0x: Yes it is component id for sure

I think it was correct previously, not sure why it was changed

`pub entrypoints_with_tracing_data: Vec<(ComponentId, Vec<EntryPointWithTracingParams>)>,`

tycho-common/src/models/blockchain.rs
```diff
@@ -392,6 +399,24 @@ pub struct EntryPointWithTracingParams {
...
+                    signature: value.entry_point.signature.clone(),
+                },
+                params: TracingParams::RPCTracer(RPCTracerParams {
+                    caller: tracer_params.caller.clone(),
+                    calldata: tracer_params.calldata.clone(),
+                }),
+            },
+        }
+    }
+}
```
zizou0x: It's very likely we can remove most of the clones here, we have ownership of `value` so we should just consume it instead of cloning

tycho-indexer/src/services/rpc.rs
```diff
@@ -89,6 +92,8 @@ pub struct RpcHandler<G> {
         RpcCache<dto::ProtocolComponentsRequestBody, dto::ProtocolComponentRequestResponse>,
     traced_entry_point_cache:
         RpcCache<dto::TracedEntryPointRequestBody, dto::TracedEntryPointRequestResponse>,
+    #[allow(dead_code)]
+    tracer: Option<EVMEntrypointService>,
```
zizou0x: You should make it generic over the `EntryPointTracer` trait instead

tycho-indexer/src/services/rpc.rs
```diff
@@ -98,6 +103,9 @@ where
     pub fn new(
         db_gateway: G,
         pending_deltas: Option<Arc<dyn PendingDeltasBuffer + Send + Sync>>,
+        // TODO leaving this as option for testing purposes. Not sure how to mock this in
+        //  tests otherwise.
+        tracer: Option<EVMEntrypointService>,
```
zizou0x: If you make it generic over the trait I managed to allow for mocking it in my latest PR, it's not merged yet but will hopefully be by EOD

tycho-indexer/src/services/rpc.rs
```diff
@@ -866,6 +875,78 @@ where
             ),
         })
     }
+
+    #[allow(dead_code)]
+    async fn add_entry_points(
+        &self,
+        request: &dto::AddEntrypointRequestBody,
+        // TODO how do I get the block hash?
```
zizou0x: Hm that's a good question 🤔

We could get the latest from RPC but I think it's simpler and more flexible if we add it to `AddEntrypointRequestBody`

tycho-indexer/src/services/rpc.rs
```diff
@@ -866,6 +875,78 @@ where
...
+                        .insert(params.entry_point.clone().into());
+                }
+                // Extend the entry point parameters for the given entry point id
+                entry_points_params
+                    .entry(params.entry_point.external_id.clone())
+                    .or_default()
+                    .insert((params.params.clone().into(), component_id_option.clone()));
+            }
+        }
+        // TODO is this sequence correct? Should I be calling all three?
```
zizou0x: Yes this is correct

tycho-common/src/models/blockchain.rs
zizou0x: Yaaay 🦀🔪🐍

zizou0x (commented):
Looooook good!

tycho-indexer/src/main.rs
```diff
@@ -302,6 +302,15 @@ async fn run_rpc(global_args: GlobalArgs) -> Result<(), ExtractionError> {
...
 
+    // TODO without this line I get Unexpected storage error: Usage error: No transaction started%
+    //  Calling this line solves the issue.
+    //  I can't put this inside the RPCHandler since it takes a Gateway - not a CachedGateway
+    //  generic - is it fine I change that?
+    //  Note to self: I guess I also have to commit after?
+    // cached_gw
+    //     .start_transaction(&Block::default(), None)
+    //     .await;
+
```
tamaralipows: Need an opinion here - is it fine if I change the generic trait G in `rpc.rs` to `CachedGateway` instead of `Gateway`? `start_transaction` doesn't exist for just `Gateway`. Or should I move the `start_transaction` to here?

tycho-common/src/dto.rs
```diff
@@ -1490,9 +1490,14 @@ impl From<TracedEntryPointRequestResponse> for DCIUpdate {
...
 }
 
-pub struct AddEntrypointRequestBody {
-    pub entrypoints_with_tracing_data: Vec<(String, Vec<EntryPointWithTracingParams>)>,
+#[derive(Serialize, Deserialize, Debug, Default, PartialEq, ToSchema, Eq, Clone)]
+pub struct AddEntryPointRequestBody {
+    #[serde(default)]
     pub chain: Chain,
+    #[serde(default)]
+    pub block_hash: Bytes,
```
zizou0x, line 1501:
```suggestion
    #[schema(value_type=String)]
    #[serde(default)]
    pub block_hash: Bytes,
```

tycho-indexer/src/services/access_control.rs
```diff
@@ -0,0 +1,72 @@
...
+    type Response = ServiceResponse<BoxBody>;
+    type Error = Error;
+    type Future = Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>>>>;
+
+    forward_ready!(service);
+
+    fn call(&self, req: ServiceRequest) -> Self::Future {
+        let required_key = self.required_key.clone();
+
+        if let Some(header_value) = req.headers().get("X-API-Key") {
```
zizou0x: It shoulnd't make a huge difference but we currently use `Authorization` instead of `X-API-Key`

tycho-indexer/src/services/mod.rs
```diff
@@ -108,6 +110,7 @@ where
                 rpc::tokens,
                 rpc::protocol_components,
                 rpc::traced_entry_points,
+                rpc::add_entry_points,
```
zizou0x: As mentioned in the plan this shouldn't be added to the swagger UI.

Ideally we should try to add a new OpenAPI and SwaggerUI for this method, but this wasn't in the scope of this task initally so I'll leave it up to you if you wanna include it or leave it for later

tycho-indexer/src/services/rpc.rs
```diff
@@ -866,6 +873,61 @@ where
...
+        self.db_gateway
+            .insert_entry_points(&entry_points)
+            .await?;
+        self.db_gateway
+            .insert_entry_point_tracing_params(&entry_points_params)
+            .await?;
+        self.db_gateway
+            .upsert_traced_entry_points(&tracing_result)
+            .await?;
+        Ok(())
```
zizou0x: Just thought about it, I think it would be helpful to return the tracing result to the caller if possible

tycho-indexer/src/services/mod.rs
```diff
@@ -233,31 +242,36 @@ where
...
+                        .route(web::post().to(rpc::protocol_components::<G, EVMEntrypointService>)),
                 )
                 .service(
                     web::resource(format!("/{}/traced_entry_points", self.prefix))
-                        .route(web::post().to(rpc::traced_entry_points::<G>)),
+                        .route(web::post().to(rpc::traced_entry_points::<G, EVMEntrypointService>)),
+                )
+                .service(
+                    web::resource(format!("/{}/add_entry_points", self.prefix))
+                        .wrap(access_control::AccessControl::new("test-key"))
```
zizou0x: Beautiful!

tycho-indexer/src/services/mod.rs
```diff
@@ -214,7 +218,12 @@ where
         openapi: utoipa::openapi::OpenApi,
         pending_deltas: Option<Arc<dyn PendingDeltasBuffer + Send + Sync>>,
     ) -> Result<(ServerHandle, JoinHandle<Result<(), ExtractionError>>), ExtractionError> {
-        let rpc_data = web::Data::new(rpc::RpcHandler::new(self.db_gateway, pending_deltas));
+        // TODO remove expect
+        let url = env::var("RPC_URL").expect("RPC_URL is not set");
+        let tracer = EVMEntrypointService::try_from_url(&url).unwrap();
```
zizou0x: Small detail but I have a feeling it would be cleaner to pass this to the Service Builder, maybe even add this to the `new` since it's mandatory? That way we can catch errors at compile time instead of run time (edit: it's actually the same if we get it from env, we would just get the error earlier)

tycho-indexer/src/main.rs
zizou0x: We need to do this inside the add function and commit at the end so we definitely need to have access to this from inside the rpc. 

I think the easiest would be to remove the generics and pass the CacheGateway directly as you suggested?

For long term I think we should have a direct gateway (so a siblings of `CacheGateway`) that implement `Gateway` and goes to the DB directly, this way we would be able to have the Gateway trait without needing to start and commit transaction. But this is very out of scope for now.

tycho-indexer/src/services/access_control.rs
```diff
@@ -0,0 +1,72 @@
...
+};
+
+use actix_web::{
+    body::BoxBody,
+    dev::{forward_ready, Service, ServiceRequest, ServiceResponse, Transform},
+    Error, HttpResponse,
+};
+
+pub struct AccessControl {
+    required_key: Rc<String>,
```
zizou0x: FMI why do we need it as a `Rc<_>` here?

zizou0x (commented):
Looks very good, we're close!

tycho-indexer/src/services/access_control.rs
tamaralipows: Good catch - I was following a tutorial. No need for our case - removed.

tycho-indexer/src/services/rpc.rs
```diff
@@ -866,6 +873,96 @@ where
...
+            .entry_points_with_tracing_data
+            .iter()
+            .flat_map(|(_, params)| params.iter().cloned().map(Into::into))
+            .collect();
+        let trace_results = self
+            .tracer
+            .trace(request.block_hash.clone(), entry_points_with_params)
+            .await
+            .map_err(|_e| {
+                // TODO proper error handling.
```
tamaralipows: Come to think of it, I'm not sure what error to raise here.

We have these possibilities:
```
#[derive(Error, Debug)]
pub enum RpcError {
    #[error("Failed to parse JSON: {0}")]
    Parse(String),

    #[error("Failed to get storage: {0}")]
    Storage(#[from] StorageError),

    #[error("Failed to get database connection: {0}")]
    Connection(#[from] deadpool::PoolError),

    #[error("Failed to apply pending deltas: {0}")]
    DeltasError(#[from] PendingDeltasError),
}
```

And this is what's raised in the tracer:
```
#[derive(Error, Debug)]
pub enum RPCError {
    #[error("RPC setup error: {0}")]
    SetupError(String),
    #[error("RPC error: {0}")]
    RequestError(#[from] ProviderError),
    #[error("Unknown error: {0}")]
    UnknownError(String),
}
```

It doesn't seem that any of those errors quite match, I guess the Parse error serves as the catch-all here?

tycho-indexer/src/services/mod.rs
tamaralipows: Will get it from env inside the `new` then

EDIT: Actually it's better to pass this since then I don't need to change `new` to return an error

tycho-indexer/src/services/mod.rs
```diff
@@ -233,31 +241,36 @@ where
...
+                        .route(web::post().to(rpc::protocol_components::<G, EVMEntrypointService>)),
                 )
                 .service(
                     web::resource(format!("/{}/traced_entry_points", self.prefix))
-                        .route(web::post().to(rpc::traced_entry_points::<G>)),
+                        .route(web::post().to(rpc::traced_entry_points::<G, EVMEntrypointService>)),
+                )
+                .service(
+                    web::resource(format!("/{}/add_entry_points", self.prefix))
+                        .wrap(access_control::AccessControl::new("test-key"))
```
tamaralipows: Oops - forgot that I left this hardcoded. Will change this tomorrow to read from env.

tycho-indexer/src/services/mod.rs
```diff
@@ -233,31 +241,36 @@ where
...
                     web::resource(format!("/{}/health", self.prefix))
                         .route(web::get().to(rpc::health)),
                 )
                 .service(
                     web::resource(format!("/{}/protocol_systems", self.prefix))
-                        .route(web::post().to(rpc::protocol_systems::<G>)),
+                        .route(web::post().to(rpc::protocol_systems::<G, EVMEntrypointService>)),
                 )
                 .wrap(RequestTracing::new())
                 .service(
```
zizou0x, line 279: Can you add a "TODO: add swagger service for internal endpoints" please

tycho-indexer/src/main.rs
```diff
@@ -361,7 +368,10 @@ async fn create_indexing_tasks(
             .unzip();
 
     let server_url = format!("http://{}:{}", global_args.server_ip, global_args.server_port);
-    let (server_handle, server_task) = ServicesBuilder::new(cached_gw.clone())
+    let rpc_url = env::var("RPC_URL").map_err(|_| {
```
zizou0x: This is already passed as input in the function, no need to extract it again

tycho-indexer/src/main.rs
```diff
@@ -298,13 +299,19 @@ async fn run_spkg(global_args: GlobalArgs, run_args: RunSpkgArgs) -> Result<(),
...
+    let direct_gw = GatewayBuilder::new(&global_args.database_url)
+        .set_chains(&[Chain::Ethereum]) // TODO: handle multichain
+        .build_direct_gw()
         .await?;
 
     info!("Starting Tycho RPC");
     let server_url = format!("http://{}:{}", global_args.server_ip, global_args.server_port);
-    let (server_handle, server_task) = ServicesBuilder::new(cached_gw)
+
+    let rpc_url = env::var("RPC_URL").map_err(|_| {
```
zizou0x: Would it make sense to move the `rpc_url` to `GlobalArgs` in the clap struct?

tycho-indexer/src/services/rpc.rs
zizou0x: Hm I don't like that there are both named the same but they have different meaning. I'll add this in our cooldown tasks. 

So this is a new case, I would just add a catch all error in RpcError

```
    #[error("Unknown error: {0}")]  
    UnknownError(String), 
```

Another approach would be to add a custom error that convert from tracing rpc error to internal rpc error but with the current naming it looks awful

```
    #[error("Node RPC error: {0}")]  
    NodeError(#[from] RPCError),  
```

tycho-indexer/src/services/rpc.rs
```diff
@@ -1432,6 +1570,228 @@ mod tests {
...
+            .called_addresses
+            .iter()
+            .cloned()
+            .collect();
+        called_addresses.sort();
+        (retriggers, called_addresses)
+    }
+
+    #[test]
+    #[ignore = "requires a RPC connection"]
```
zizou0x: You can use a `MockEntryPointTracer` to not require a connection, you can find examples of this already mocked in `dynamic_contract_indexer.rs`

tycho-indexer/src/main.rs
louise-poole: Agreed - looks like it's used by all sub commands so is appropriate as a global arg.

tycho-indexer/src/services/rpc.rs
```diff
@@ -866,6 +873,96 @@ where
             ),
         })
     }
+
+    #[allow(dead_code)]
+    async fn add_entry_points(
+        &self,
+        request: &dto::AddEntryPointRequestBody,
+    ) -> Result<dto::AddEntryPointRequestResponse, RpcError> {
+        let tracing_result = self.trace_entry_points(request).await?;
```
louise-poole, line 887: Design question here: do we really want to conduct the trace here? This means when someone uses this endpoint they potentially have to wait quite a while for a response. Benefit is input validation: we can error if the trace fails and the provided entrypoint/params are bad. So the forced wait might be good.
But then how does this get propagated to users connected to our streams? If I understand correctly, as it is the Extractor will never be aware of entrypoints added with this endpoint (unless we restart the indexer and DCI fetches it from the DB on start-up).
@zizou0x

tycho-indexer/src/services/rpc.rs
```diff
@@ -1162,6 +1259,43 @@ pub async fn traced_entry_points<G: Gateway>(
     }
 }
 
+/// Trace given entry points and add results to db, along with entry points and tracing params.
```
```suggestion
/// Add Entry Point
///
/// Trace the given entry points and add the entry point, tracing params and results to DB.
```

tycho-indexer/src/main.rs
```diff
@@ -298,13 +299,19 @@ async fn run_spkg(global_args: GlobalArgs, run_args: RunSpkgArgs) -> Result<(),
 #[tokio::main]
 async fn run_rpc(global_args: GlobalArgs) -> Result<(), ExtractionError> {
     create_tracing_subscriber();
-    let cached_gw = GatewayBuilder::new(&global_args.database_url)
-        .build_gw()
+
+    let direct_gw = GatewayBuilder::new(&global_args.database_url)
+        .set_chains(&[Chain::Ethereum]) // TODO: handle multichain
```
louise-poole, line 302: We could also probably set chain as a global arg?

tycho-storage/src/postgres/builder.rs
```diff
@@ -69,4 +69,20 @@ impl GatewayBuilder {
...
+        let pool = postgres::connect(&self.database_url).await?;
+        postgres::ensure_chains(&self.chains, pool.clone()).await;
+        postgres::ensure_protocol_systems(&self.protocol_systems, pool.clone()).await;
+
+        let inner_gw = PostgresGateway::new(pool.clone(), self.retention_horizon).await?;
+
+        let chain = self
+            .chains
+            .first()
+            .expect("No chains provided"); //TODO: handle multichain?
```
louise-poole, line 83: Out of scope here... but I think chain should be an enforced parameter... we should not be allowed to try build a gateway with no chain provided. (will need to add it as a cooldown task I guess)

tycho-storage/src/postgres/direct.rs
```diff
@@ -0,0 +1,687 @@
...
+        BlockIdentifier, BlockOrTimestamp, ChainGateway, ContractStateGateway, EntryPointFilter,
+        EntryPointGateway, ExtractionStateGateway, Gateway, ProtocolGateway, StorageError, Version,
+        WithTotal,
+    },
+    Bytes,
+};
+
+use super::{PostgresError, PostgresGateway};
+
+pub struct DirectGateway {
```
louise-poole, line 37: Can you add docs explaining what the direct gateway is for? I assume it's to allow us to bypass the gateway cache no?

tycho-storage/src/postgres/direct.rs
```diff
@@ -0,0 +1,687 @@
...
+
+impl Clone for DirectGateway {
+    fn clone(&self) -> Self {
+        Self {
+            pool: self.pool.clone(),
+            state_gateway: self.state_gateway.clone(),
+            chain: self.chain,
+        }
+    }
+}
```
louise-poole: This looks like a very standard clone... why do it manually instead of just using `#[derive(Clone)]` on the struct?

tycho-indexer/src/services/rpc.rs
zizou0x: This endpoint in meant to be used for maintenance only (it's like changing things in the db but in a more convenient way). 

It would be nice to cover these runtime things but it's not possible and out of scope for now, we have 0 context about extractors things here

tycho-indexer/src/services/rpc.rs
tamaralipows: Hmmm to be honest I prefer to have a real tracer for this to be a proper integration test. It felt safer that way since we can't easily test the RPC locally

tycho-storage/src/postgres/direct.rs
tamaralipows: For some reason when I simply derive it I get
> Trait `Clone` is not implemented for `Pool<AsyncPgConnection>` [E0277]

No idea why it works this way... it's copypasta'd from the cache gw

tycho-indexer/src/main.rs
tamaralipows: I think this ties in with the multichain support / making it an enforced parameter... could this be a part of cooldown too?

tycho-indexer/src/main.rs
```diff
@@ -307,16 +305,17 @@ async fn run_rpc(global_args: GlobalArgs) -> Result<(), ExtractionError> {
...
-        .port(global_args.server_port)
-        .register_extractors(vec![])
-        .run()?;
+
+    let (server_handle, server_task) =
+        ServicesBuilder::new(direct_gw.clone(), global_args.rpc_url.clone(), api_key)
+            .prefix(&global_args.server_version_prefix)
+            .bind(&global_args.server_ip)
+            .port(global_args.server_port)
+            .register_extractors(vec![])
```
louise-poole: This is an no-op if you pass it an empty vec
```suggestion
```

tycho-indexer/src/services/mod.rs
```diff
@@ -93,6 +95,12 @@ where
         self
     }
 
+    /// Sets the api key for access control
+    pub fn api_key(mut self, v: String) -> Self {
+        self.api_key = v;
+        self
+    }
+
```
louise-poole: This is already a required parameter for `new`... we don't need a setter for it too then.

tycho-storage/src/postgres/direct.rs
louise-poole: Curious 👀  I'll play around with this locally for a bit to see if I can find why.

louise-poole: I didn't get this error locally. Does your IDE do this warning? Or clippy? I ran both clippy and rustfmt and couldn't reproduce

tycho-indexer/src/services/mod.rs
tamaralipows: oops. kept going back and forth and back and forth on this and then i forgot to remove it finally. thanks!

tycho-storage/src/postgres/direct.rs
tamaralipows: My IDE did it - you're right that clippy doesn't do it. Wow... I wonder what else it lies to me about. Committing this - sorry!

tycho-indexer/src/main.rs
tamaralipows: Not sure what the intention here was... I haven't touched it 🤔 Maybe it was meant to make it more explicit to read? I'll delete it anyway

tycho-indexer/src/services/rpc.rs
zizou0x: Yes I see your point and it's valid, but if we make this one an integration test we would lack an unit test no?
The biggest thing that bothers me here is this will not run in CI, so we won't have any CI test for this endpoint at all, it feel it's kinda risky

tycho-common/src/dto.rs
```diff
@@ -1490,9 +1490,22 @@ impl From<TracedEntryPointRequestResponse> for DCIUpdate {
...
-    pub entrypoints_with_tracing_data: Vec<(String, Vec<EntryPointWithTracingParams>)>,
+#[derive(Serialize, Deserialize, Debug, Default, PartialEq, ToSchema, Eq, Clone)]
+pub struct AddEntryPointRequestBody {
+    #[serde(default)]
     pub chain: Chain,
+    #[schema(value_type=String)]
+    #[serde(default)]
+    pub block_hash: Bytes,
+    /// The map of component ids to their tracing params to insert
+    pub entry_points_with_tracing_data: Vec<(String, Vec<EntryPointWithTracingParams>)>,
```
tvinagre: Picky comment, but we have a type in `tycho-common` that I'd rather use for `ComponentId`;
```rust 
pub type ComponentId = String;
```

Makes it easier to identify that it's not any string, but a `ComponentId` one - so I like to use them, specially on external interfaces

PR #588, opened 2025-06-02
feat: add specific slots tracing

zizou0x:
This is needed by DCI to avoid indexing the whole contract, instead we will just index some slots that were used during the tracing (read or write)

### DIFF
--- a/tycho-storage/src/postgres/orm.rs
+++ b/tycho-storage/src/postgres/orm.rs
@@ -1620,7 +1620,7 @@ pub struct NewContract {
 }
 
 impl NewContract {
-    pub fn new_account(&self) -> NewAccount {
+    pub fn new_account(&self) -> NewAccount<'_> {
         NewAccount {
             title: &self.title,
             address: &self.address,


tycho-common/src/dto.rs
```diff
@@ -1423,14 +1423,14 @@ impl From<models::blockchain::EntryPointWithTracingParams> for EntryPointWithTra
 #[derive(Serialize, Deserialize, Debug, Default, PartialEq, ToSchema, Eq, Clone)]
 pub struct TracingResult {
     #[schema(value_type=HashSet<(String, String)>)]
-    pub retriggers: HashSet<(Bytes, Bytes)>,
-    #[schema(value_type=HashSet<String>)]
-    pub called_addresses: HashSet<Bytes>,
+    pub retriggers: HashSet<(StoreKey, StoreVal)>,
+    #[schema(value_type=HashMap<String,HashSet<String>>)]
+    pub accessed_slots: HashMap<Address, HashSet<StoreKey>>,
```
louise-poole, line 1428: ❤️

tycho-ethereum/src/entrypoint_tracer/tracer.rs
```diff
@@ -82,6 +85,8 @@ impl EntryPointTracer for EVMEntrypointService {
                 TracingParams::RPCTracer(ref rpc_entry_point) => {
                     // First call to get the list of called addresses
                     // TODO: Can we only use one call to get the retriggers and called addresses?
```
tvinagre: I've seen us start using `Perf` instead of TODO for possible performance optimizations that don't change the outcome of the feature. What do you think?

tvinagre (approved):

tycho-ethereum/src/entrypoint_tracer/tracer.rs
zizou0x: Yes I like using perf, will apply this in the PR

PR #590, opened 2025-06-04
feat(dci): add revert logic

zizou0x:
This PR implements the logic needed by DCI to properly deal with chain reorgs. 

It moves all dci related files into a "folder module".

We added a helper structs which is essentially a wrapper around HashMap. This struct helps to efficiently keep track of the permanent and pending state and provides functions to revert the cache state to a previous state.

**Review commit by commit suggested**

### DIFF
--- a/tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
+++ b/tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
@@ -152,7 +152,7 @@ where
     }
 
     /// Gets a mutable entry to a value in the permanent layer.
-    pub(super) fn permanent_entry(&mut self, k: &K) -> Entry<K, V> {
+    pub(super) fn permanent_entry<'a>(&'a mut self, k: &K) -> Entry<'a, K, V> {
         self.permanent.entry(k.clone())
     }
 
@@ -160,11 +160,11 @@ where
     ///
     /// If the block is not found, we check if the block is the child of the latest pending
     /// block. If it is, a new layer is created. Otherwise, an error is returned.
-    pub(super) fn pending_entry(
-        &mut self,
+    pub(super) fn pending_entry<'a>(
+        &'a mut self,
         block: &Block,
         k: &K,
-    ) -> Result<Entry<K, V>, DCICacheError> {
+    ) -> Result<Entry<'a, K, V>, DCICacheError> {
         let layer = self.validate_and_ensure_block_layer(block)?;
         Ok(layer.data.entry(k.clone()))
     }
--- a/tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
+++ b/tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
@@ -1,0 +1,27 @@
+use std::collections::{HashMap, HashSet};
+
+use tycho_common::models::{
+    blockchain::{EntryPoint, EntryPointWithTracingParams, TracingParams, TracingResult},
+    Address, EntryPointId, StoreKey,
+};
+
+/// A unique identifier for a storage location, consisting of an address and a storage key.
+type StorageLocation = (Address, StoreKey);
+
+pub(super) struct DCICache {
+    pub(super) ep_id_to_entrypoint: HashMap<EntryPointId, EntryPoint>,
+    pub(super) entrypoint_results: HashMap<(EntryPointId, TracingParams), TracingResult>,
+    pub(super) retriggers: HashMap<StorageLocation, HashSet<EntryPointWithTracingParams>>,
+    pub(super) tracked_contracts: HashMap<Address, Option<HashSet<StoreKey>>>,
+}
+
+impl DCICache {
+    pub(super) fn new_empty() -> Self {
+        Self {
+            ep_id_to_entrypoint: HashMap::new(),
+            entrypoint_results: HashMap::new(),
+            retriggers: HashMap::new(),
+            tracked_contracts: HashMap::new(),
+        }
+    }
+}
--- a/tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
+++ b/tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
@@ -27,7 +27,7 @@ pub(super) struct DCICache {
 }
 
 impl DCICache {
-    pub(super) fn new_empty() -> Self {
+    pub(super) fn new() -> Self {
         Self {
             ep_id_to_entrypoint: VersionedCache::new(),
             entrypoint_results: VersionedCache::new(),
@@ -487,7 +487,7 @@ mod tests {
 
     #[test]
     fn test_dci_cache_revert() {
-        let mut cache = DCICache::new_empty();
+        let mut cache = DCICache::new();
         let block1 = create_test_block(1, "0x01", "0x00");
         let block2 = create_test_block(2, "0x02", "0x01");
 
--- a/tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
+++ b/tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
@@ -43,12 +43,36 @@ impl DCICache {
     ///
     /// # Arguments
     /// * `block` - Target block hash to revert to.
-    pub(super) fn revert_to(&mut self, block: &BlockHash) {
+    pub(super) fn revert_to(&mut self, block: &BlockHash) -> Result<(), DCICacheError> {
         self.ep_id_to_entrypoint
-            .revert_to(block);
-        self.entrypoint_results.revert_to(block);
-        self.retriggers.revert_to(block);
-        self.tracked_contracts.revert_to(block);
+            .revert_to(block)?;
+        self.entrypoint_results
+            .revert_to(block)?;
+        self.retriggers.revert_to(block)?;
+        self.tracked_contracts
+            .revert_to(block)?;
+
+        Ok(())
+    }
+
+    /// Move new finalized blocks state to the permanent layer.
+    ///
+    /// # Arguments
+    /// * `finalized_block_height` - The height of the finalized block.
+    pub(super) fn handle_finality(
+        &mut self,
+        finalized_block_height: u64,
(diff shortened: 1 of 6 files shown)


tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
```diff
@@ -0,0 +1,556 @@
...
+                    last_layer.block.number,
+                ));
+            }
+        }
+
+        Ok(self
+            .pending
+            .back_mut()
+            .expect("Layer should now exist"))
+    }
```
zizou0x, line 335: Should we allow to update previously pending data? 

Currently this errors if we try to update a block that isn't either the latest in the cache or a new block that follows the correct sequence (its parent_hash should match the latest stored block hash).

We could allow for a full cache lookup instead of only the latest one.

It shouldn't be needed because DCI never need to update data for a previous block, I've made it like this to provide a security. But I'm opened for suggestions

tycho-indexer/src/extractor/dynamic_contract_indexer/dci.rs
```diff
@@ -323,10 +304,9 @@ where
         Ok(())
     }
 
-    #[allow(unused)]
     async fn process_revert(&mut self, target_block: &BlockHash) -> Result<(), ExtractionError> {
-        // TODO: Handle reverts, need to cleanup reverted internal state
-        todo!()
+        self.cache.revert_to(target_block);
```
zizou0x: This can't error right now, I think we should change the interface accordingly?

Waiting for your opinion here

tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
```diff
@@ -0,0 +1,556 @@
...
+        self.retriggers.revert_to(block);
+        self.tracked_contracts.revert_to(block);
+    }
+}
+
+#[derive(Error, Debug, PartialEq)]
+pub enum DCICacheError {
+    #[error("Unexpected block order: new block number {0} is not a child of {1}")]
+    UnexpectedBlockOrder(u64, u64),
+}
```
zizou0x: Might be a bit overkill, I initially used `Result<(), String>` but decided to introduce this for consistency

tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
```diff
@@ -0,0 +1,706 @@
...
+            entrypoint_results: VersionedCache::new(),
+            retriggers: VersionedCache::new(),
+            tracked_contracts: VersionedCache::new(),
+        }
+    }
+
+    /// Reverts the cache to the state at a specific block hash.
+    ///
+    /// This operation will discard all changes made in blocks after the specified block.
+    /// If the block hash is not found in the pending cache, all pending layers will be dropped.
```
louise-poole: Does this mean the whole cache will be dropped?

tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
```diff
@@ -0,0 +1,706 @@
...
+    /// # Returns
+    /// * `Ok(())` - On successful processing
+    /// * `Err(DCICacheError::FinalityBlockNotFound)` - If the finalized block is not found in
+    ///   pending layers
+    ///
+    /// Note: to make sure the finalized height is always found in normal conditions, we keep the
+    /// finalized block in the pending layers. For example:
+    /// If pending layers contain blocks [100, 101, 102, 103] and block 102 is finalized:
+    /// - Blocks 100 and 101 are moved to permanent storage
+    /// - Block 102 and 103 remain in pending storage
```
louise-poole: Ahhh - okay, is this for reverting? If we revert to the last finalized block it needs to be in the pending values and not in the permanent values? Or did I misunderstand?
And if this is true - can't we rather stick to the more expected behaviour of putting all finalized blocks in the permanent storage and make sure our revert logic caters for that?

EDIT: looks like the revert logic already takes this into account (checks the parent of the last block in the pending blocks). So why do we need to keep the last finalized block in the pending blocks?

tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
```diff
@@ -0,0 +1,706 @@
...
+            return Ok(());
+        }
+
+        let finalized_index = self
+            .pending
+            .iter()
+            .position(|layer| layer.block.number == finalized_block_height)
+            .ok_or_else(|| DCICacheError::FinalityNotFound(finalized_block_height))?;
+
+        // Move all finalized layers to permanent storage
```
louise-poole: Technically this isn't 'all' finalized layers 👀

tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
```diff
@@ -0,0 +1,706 @@
...
+            if layer.block.hash == *block {
+                let _ = self.pending.split_off(index);
+                found = true;
+                break;
+            }
+        }
+
+        if !found {
+            // On startup, the pending cache could not contain the latest finalized block but only
+            // its child. In this case, we clear the pending cache and return.
```
louise-poole, line 271: What happens if more than one block is reverted? Then we could have the child of the child in the pending cache no?

tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
louise-poole: What happens when we're syncing a protocol? Does DCI not run until it has fully synced?
Also what happens if a protocol's substream falls behind for a few blocks (as we have seen it occasionally do)?

tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
```diff
@@ -0,0 +1,706 @@
...
+
+impl DCICache {
+    pub(super) fn new_empty() -> Self {
+        Self {
+            ep_id_to_entrypoint: VersionedCache::new(),
+            entrypoint_results: VersionedCache::new(),
+            retriggers: VersionedCache::new(),
+            tracked_contracts: VersionedCache::new(),
+        }
+    }
```
louise-poole: Can't this just be `new`? It's pretty obvious that a new cache instantiated without passing it any data would be empty.

tycho-indexer/src/extractor/dynamic_contract_indexer/dci.rs
```diff
@@ -621,12 +614,6 @@ where
 
         Ok(tracked_updates)
     }
-
-    // TODO: Handle reverts, need to cleanup reverted internal state
-    #[allow(unused)]
-    pub(super) fn process_revert(&mut self, target_block: u64) -> Result<(), ExtractionError> {
-        todo!()
-    }
```
louise-poole, line 660: Eish.... thought I removed this. Thanks!

tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
zizou0x: This is outdated, the real documentation is up to date in the inner `revert_to` function, I'll update this one

tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
zizou0x: Yes, I wasn't sure when I implemented it but it can be just `new`

tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
zizou0x: It's explained above 

>      /// Note: to make sure the finalized height is always found in normal conditions, we keep the
>     /// finalized block in the pending layers. For example:
 
 Not sure what isn't clear but basically below we do 
 
 ```
         let finalized_index = self
            .pending
            .iter()
            .position(|layer| layer.block.number == finalized_block_height)
            .ok_or_else(|| DCICacheError::FinalityNotFound(finalized_block_height))?;
```
 
We are getting 32 block with the same `finalized_block_height`, if we don't keep the last block in the pending it would fail. A solution would be to allow for the last block - 1 and skip, I'm not against it but I decided to go for this because this is consistent with the revert buffer

tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
zizou0x: Then we would be good, if it reverts to the child we would find it and else if it reverts to the last finalized block it should be the parent_hash of `self.pending.front()`

tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
zizou0x: It does run while syncing with the pluging version. But that's not a problem, we still process blocks sequentially.

There is probably room for improvement here though, if we know that we're syncing we can push directly to the permanent state. I'll leave a TODO for this

tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
louise-poole: Ahhh - okay I get it. The 'normal conditions' part was ambiguous to me. But essentially you mean this fn could be repeatedly called with the same finalized block and we need that block to be in pending deltas for the current logic to work.

tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
louise-poole: No - I mean it's reverting 2 blocks on start-up. So say we startup at block 5. Block 5 is put in the pending deltas with block 4 listed as it's parent. Then we get a revert message to revert to block 3... it is neither the in the pending deltas or the parent of an existing block?

zizou0x: This is not possible because if we startup at block 5 it means that block 4 is finalized

tvinagre (approved):

PR #591, opened 2025-06-05
fix(tycho-client): update cli docs, validation and address handling

louise-poole:
No longer quietly emits empty messages if the pool address supplied is not lower case. 
Also added a --verbose flag to toggle on debug logs.


zizou0x (approved):
LGTM thanks

PR #592, opened 2025-06-11
feat: Add user identity to RPC request metrics

louise-poole:


tycho-indexer/src/services/rpc.rs
```diff
@@ -784,18 +785,22 @@ where
 pub async fn contract_state<G: Gateway>(
     body: web::Json<dto::StateRequestBody>,
     handler: web::Data<RpcHandler<G>>,
+    req: HttpRequest,
```
zizou0x, line 788: I don't understand this, shouldn't this interface change affect something? What calls this?

Also I think we should make these function `pub(crate)`

louise-poole: Nope. Because we register these endpoints with actix-web (see [here](https://github.com/propeller-heads/tycho-indexer/blob/main/tycho-indexer/src/services/mod.rs#L236)) and according to what I found online:

> When you register a handler with .to(), actix-web will automatically extract the parameters it needs from the request context. The HttpRequest parameter is one of the standard extractors that actix-web provides, and it's safe to add it to any handler function. The framework will automatically provide the request object when calling the handler.

zizou0x: I see, I didn't know it worked like this 👀 

I still think we shouldn't have them as `pub` though

PR #593, opened 2025-06-12
fix: miscellaneous fixes after DCI testing with Balancer V2

zizou0x:
This PR fixes many bugs that appeared when testing DCI release with Balancer V2 integration.

For more details on each bugs please check commits descriptions.


louise-poole (approved):
These all look good to me! Thanks for finding and fixing these all 🙏

PR #602, opened 2025-06-25
fix(storage): fix bug that filters out contracts missing a creation_tx

louise-poole:
When an account is added as a token, no creation_tx is assigned to the account entry. Since the `get_contracts` DB gateway function uses the linked `creation_tx` to apply versioning to, if it is set to NULL we will always filter it out when we apply versioning. Therefore we should only apply versioning to account entries that have a creation_tx specified.

Done in this PR:
- do not filter out accounts that have no `creation_tx`
- optimise the queries: 
    - remove transaction table join (creation tx isn't always available and is misleading for DCI added contracts as they are linked to the transaction where DCI detected them and not where the contract was created)
    - improve count query (and skip if no pagination was applied)

### DIFF
--- a/tycho-common/src/dto.rs
+++ b/tycho-common/src/dto.rs
@@ -661,6 +661,7 @@ pub struct ResponseAccount {
     #[serde(with = "hex_bytes")]
     pub code_modify_tx: Bytes,
     /// Transaction hash which created the account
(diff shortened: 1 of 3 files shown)


zizou0x (approved):
LGTM

PR #605, opened 2025-06-26
fix: correctly fill the reorg buffer

zizou0x:
This PR fixes a bug causing the block to not be found when calling `handle_finality`. This is because `handle_finality` was called before the new block was inserted. Also the block insertion and finality handling was happening only if `entrypoints_to_analyze` wasn't empty.

We are now always inserting a buffer layer for each block and always calling handle finality after processing everything. This should ensure that the latest finalized block is always found in the buffer.


tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
```diff
@@ -89,6 +114,8 @@ pub enum DCICacheError {
     RevertToBlockNotFound(BlockHash),
     #[error("Finality block {0} not found in pending cache.")]
     FinalityNotFound(u64),
+    #[error("Block {0} not found in pending cache.")]
+    BlockNotFound(u64),
```
louise-poole: It's unclear to me how these two errors are different? Both of them are technically 'block not found' errors no?

PR #606, opened 2025-06-26
feat(client-py): update python RPC client

louise-poole:
Updated the TychoRPCClient to include:
- all new/missing endpoints: 
    - health
    - component_tvl
    - protocol_systems
    - traced_entry_points
- add pagination to all requests
- a simple test script

### DIFF
--- a/tycho-client-py/python/test_rpc_client.py
+++ b/tycho-client-py/python/test_rpc_client.py
@@ -1,0 +1,181 @@
+#!/usr/bin/env python3
+"""
+Test script for the updated Tycho RPC client.
+This script tests all the available endpoints to ensure they work correctly.
+"""
+
+import sys
+import os
+
+# Add the current directory to the Python path
+sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
+
+from tycho_indexer_client import (
+    TychoRPCClient,
+    TychoRPCError,
+    Chain,
+    ProtocolComponentsParams,
+    ProtocolStateParams,
+    ContractStateParams,
+    TokensParams,
+    ProtocolSystemsParams,
+    ComponentTvlParams,
+    TracedEntryPointParams,
+    PaginationParams,
+)
+
+
+def test_client():
+    """Test the updated Tycho RPC client."""
+
+    # Initialize client
+    client = TychoRPCClient(
+        rpc_url="https://tycho-beta.propellerheads.xyz",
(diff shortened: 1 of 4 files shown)


tycho-client-py/python/test_rpc_client.py
```diff
@@ -0,0 +1,181 @@
...
+    ContractStateParams,
+    TokensParams,
+    ProtocolSystemsParams,
+    ComponentTvlParams,
+    TracedEntryPointParams,
+    PaginationParams,
+)
+
+
+def test_client():
```
tvinagre: I'm curious why you didn't use a testing framework like pytest

tvinagre: But yeah, if it's bc you want to run this as a script I'm fine with it

tycho-client-py/python/tycho_indexer_client/dto.py
```diff
@@ -34,14 +34,16 @@ def __int__(self):
         if self.hex() != "0x":
             return int(self.hex(), 16)
         else:
-          return 0
+            return 0
```
tvinagre, line 37: I think we should add black formatting check to CI

tvinagre (approved):

tycho-client-py/python/test_rpc_client.py
zizou0x: Also curious here, this seems to be redundant with the main of `rpc_client.py`

zizou0x (approved):
🚀

tycho-client-py/python/test_rpc_client.py
louise-poole: The test script covered a lot more than the example in the main of `rpc_client` - but I agree there is too much overlap. I have instead just extended the main of `rpc_client` and removed this script.

tycho-client-py/python/tycho_indexer_client/dto.py
louise-poole: Will add a cooldown task

tycho-client-py/python/test_rpc_client.py
louise-poole: > I'm curious why you didn't use a testing framework like pytest

Didn't want to spend too much time on this as I wanted to focus on DCI related tasks (which these changes were needed for). So I made a quick test script by copying the example in main instead of setting up an official test suite from scratch.

PR #610, opened 2025-07-04
feat: add a cache for tokens

zizou0x:
Tokens have been a big issue recently on Base. We have 4M tokens on it and the queries are taking too much time. This commit aims to make this better by storing everything in memory.

Still TODO: 

- [ ] use this in the protocol cache to avoid storing all tokens twice
- [ ] avoid unlimited memory usage in the cache -> need to remove old entries on inserts
- [ ] test

### DIFF
--- a/tycho-storage/src/postgres/token_cache.rs
+++ b/tycho-storage/src/postgres/token_cache.rs
@@ -1,0 +1,325 @@
+use std::{
+    collections::{BTreeMap, HashMap, HashSet},
+    ops::Bound,
+    str::FromStr,
+    sync::Arc,
+};
+
+use chrono::NaiveDateTime;
+use diesel::prelude::*;
+use diesel_async::{pooled_connection::deadpool::Pool, AsyncPgConnection, RunQueryDsl};
+use tokio::sync::RwLock;
+use tycho_common::{
+    models::{protocol::QualityRange, token::CurrencyToken, Address, Chain, PaginationParams},
+    storage::{StorageError, WithTotal},
+};
+
+use crate::postgres::{orm, schema, PostgresError};
+
+/// Index of tokens by address.
+type TokenIndex = Arc<RwLock<HashMap<Address, CurrencyToken>>>;
+/// Index of tokens by quality.
+type QualityIndex = Arc<RwLock<BTreeMap<i32, HashSet<Address>>>>;
+/// Index of tokens by last traded timestamp.
+type TradedTsIndex = Arc<RwLock<BTreeMap<NaiveDateTime, HashSet<Address>>>>;
+
+#[derive(Debug, Clone)]
+pub struct TokenCache {
+    tokens: HashMap<Chain, TokenIndex>,
+    quality_index: HashMap<Chain, QualityIndex>,
+    traded_ts_index: HashMap<Chain, TradedTsIndex>,
+}
+
+impl TokenCache {
+    pub async fn from_pool(pool: Pool<AsyncPgConnection>) -> Result<Self, StorageError> {
+        let mut conn = pool
+            .get()
+            .await
+            .map_err(|err| StorageError::Unexpected(err.to_string()))?;
+
+        Self::from_connection(&mut conn).await
+    }
+
+    pub async fn from_connection(mut conn: &mut AsyncPgConnection) -> Result<Self, StorageError> {
+        use super::schema::token as dsl_token;
+
+        //TODO: Would be safer to get the chain from the CLI args, but it requires some interface
+        // changes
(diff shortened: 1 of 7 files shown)


tycho-storage/src/postgres/token_cache.rs
```diff
@@ -0,0 +1,325 @@
...
+/// Index of tokens by address.
+type TokenIndex = Arc<RwLock<HashMap<Address, CurrencyToken>>>;
+/// Index of tokens by quality.
+type QualityIndex = Arc<RwLock<BTreeMap<i32, HashSet<Address>>>>;
+/// Index of tokens by last traded timestamp.
+type TradedTsIndex = Arc<RwLock<BTreeMap<NaiveDateTime, HashSet<Address>>>>;
+
+#[derive(Debug, Clone)]
+pub struct TokenCache {
+    tokens: HashMap<Chain, TokenIndex>,
```
kayibal, line 41: Maybe we should use DashMap here to avoid the RWLock around the whole cache...

tycho-storage/src/postgres/token_cache.rs
```diff
@@ -0,0 +1,325 @@
...
+    models::{protocol::QualityRange, token::CurrencyToken, Address, Chain, PaginationParams},
+    storage::{StorageError, WithTotal},
+};
+
+use crate::postgres::{orm, schema, PostgresError};
+
+/// Index of tokens by address.
+type TokenIndex = Arc<RwLock<HashMap<Address, CurrencyToken>>>;
+/// Index of tokens by quality.
+type QualityIndex = Arc<RwLock<BTreeMap<i32, HashSet<Address>>>>;
```
kayibal: Can we move the lock inside here? `Arc<BTreeMap<i32, RwLock<HashSet<Address>>>>;`

tycho-storage/src/postgres/token_cache.rs
```diff
@@ -0,0 +1,325 @@
...
+
+    pub async fn update_last_traded_ts(
+        &self,
+        chain: &Chain,
+        entries: HashMap<NaiveDateTime, Vec<Address>>,
+    ) {
+        // TODO: This function doesn't remove old entries. Is this a problem?
+        // For filtering it is not because we only filter by >= target ts, so worst case we will
+        // encounter the same token twice. But what about memory usage? It will keep growing
+        // forever.
```
kayibal, line 217: Under what condition would we remove something?

tycho-storage/src/postgres/token_cache.rs
```diff
@@ -0,0 +1,325 @@
...
+    ) -> Result<WithTotal<Vec<CurrencyToken>>, StorageError> {
+        let tokens_lock = self
+            .tokens
+            .get(chain)
+            .unwrap()
+            .read()
+            .await;
+        let mut candidate_addrs: HashSet<&Address> = match addresses {
+            Some(addrs) => addrs.iter().copied().collect(),
+            None => tokens_lock.keys().collect(),
```
kayibal: Even though these are only pointers, that latter case is a big allocation on each query e.g. 4M tokens, each address 20 bytes. So this is a 80 Megabyte allocation. Probably can speed this up by using `Vec::with_capacity()`, but ideally it is avoided and you try to use an iterator?

Another option would be to keep a pre allocated vector for this on the struct, but that needs to be protected through a lock too...

tycho-storage/src/postgres/token_cache.rs
```diff
@@ -0,0 +1,325 @@
...
+                .flat_map(|(_, addrs)| addrs)
+                .collect();
+
+            candidate_addrs.retain(|addr| ts_ids.contains(addr));
+        }
+
+        let mut candidate_addrs = candidate_addrs
+            .into_iter()
+            .collect::<Vec<_>>();
+        candidate_addrs.sort_unstable_by_key(|addr| addr.to_string());
```
kayibal: This is also a quite espensive operation here but I guess assuming candidates is small it's alright?

tycho-storage/src/postgres/token_cache.rs
```diff
@@ -0,0 +1,325 @@
...
+
+        let paginated_addrs: Vec<_> = if let Some(pagination) = pagination {
+            let offset = pagination.offset() as usize;
+            let limit = pagination.page_size as usize;
+            candidate_addrs
+                .iter()
+                .skip(offset)
+                .take(limit)
+                .copied()
+                .collect()
```
kayibal: So if we go into this case, we need to redo all of the above just to get the next subset - that might be slower than doing the DB query ....

PR #613, opened 2025-07-08
feat(hook-support): implement metadata generator for Euler Swap

zizou0x:
This PR adds metadata generator for Euler Swap

TODOs:
- [x] Add a way to get the tx hash in `MetadataRequestGenerator`
- [ ] Add real RPC calls tests

### DIFF
--- a/tycho-indexer/src/extractor/dynamic_contract_indexer/euler/metadata_generator.rs
+++ b/tycho-indexer/src/extractor/dynamic_contract_indexer/euler/metadata_generator.rs
@@ -1,0 +1,258 @@
+use tycho_common::models::{blockchain::Block, protocol::ProtocolComponent, TxHash};
+
+use crate::extractor::dynamic_contract_indexer::component_metadata::{
+    MetadataError, MetadataRequest, MetadataRequestGenerator, MetadataRequestType, RpcTransport,
+};
+
+pub struct EulerMetadataGenerator {
+    rpc_url: String,
+}
+
+impl EulerMetadataGenerator {
+    pub fn new(rpc_url: String) -> Self {
+        Self { rpc_url }
+    }
+}
+
+impl MetadataRequestGenerator for EulerMetadataGenerator {
+    fn generate_requests(
+        &self,
+        component: &ProtocolComponent,
+        block: &Block,
+    ) -> Result<Vec<MetadataRequest>, MetadataError> {
+        let mut requests = vec![];
+
+        let balance_transport = RpcTransport::new(
+            self.rpc_url.clone(),
+            "eth_call".to_string(),
+            vec![serde_json::json!([
+                  {
+                    "data": "0x0902f1ac", // getReserves()
+                    "to": component.id
+                  },
+                  block.number.to_string() //TODO: should it be formatted as a hex string?
+                ]
+            )],
+        );
+        requests.push(MetadataRequest::new(
+            format!("euler_balance_{}", component.id),
+            component.id.clone(),
+            MetadataRequestType::ComponentBalance { token_addresses: component.tokens.clone() },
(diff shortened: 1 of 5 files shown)


tycho-indexer/src/extractor/dynamic_contract_indexer/euler/metadata_generator.rs
```diff
@@ -0,0 +1,258 @@
...
+impl MetadataRequestGenerator for EulerMetadataGenerator {
+    fn generate_requests(
+        &self,
+        component: &ProtocolComponent,
+        block: &Block,
+    ) -> Result<Vec<MetadataRequest>, MetadataError> {
+        let mut requests = vec![];
+
+        let balance_transport = RpcTransport::new(
+            self.rpc_url.clone(),
```
tvinagre: I'm wondering if it makes sense to have the RPC URL tied to the RpcTransport or if it should be controlled by the RPC Provider.
I'm fine to have it like this for now but I'd like to hear your opinion on the design here.

tycho-indexer/src/extractor/dynamic_contract_indexer/euler/metadata_generator.rs
```diff
@@ -0,0 +1,258 @@
...
+            "eth_call".to_string(),
+            vec![serde_json::json!([
+                  {
+                    "data": "0x0902f1ac", // getReserves()
+                    "to": component.id
+                  },
+                  block.number.to_string() //TODO: should it be formatted as a hex string?
+                ]
```
tvinagre: Yes, the pattern is hex string for block number: https://www.quicknode.com/docs/ethereum/eth_getStorageAt

tycho-indexer/src/extractor/dynamic_contract_indexer/euler/metadata_generator.rs
```diff
@@ -0,0 +1,253 @@
...
+            vec![serde_json::json!([
+              {
+                "data": format!("0xaaed87a3000000000000000000000000{}000000000000000000000000{}", component.tokens[0], component.tokens[1]),
+                "to": component.id
+              },
+              block.number.to_string()
+            ])],
+        );
+        requests.push(MetadataRequest::new(
+            format!("euler_limits_{}", component.id),
```
tvinagre: request_id will most likely be used for deduplication, so we need to be able to differentiate this request and the next one.

tycho-indexer/src/extractor/dynamic_contract_indexer/euler/metadata_generator.rs
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    fn generate_balance_only_requests(
+        &self,
+        component: &ProtocolComponent,
+        block: &Block,
+    ) -> Result<Vec<MetadataRequest>, MetadataError> {
+        let mut requests = vec![];
+
+        let balance_transport = RpcTransport::new(
```
tvinagre, line 92: can we deduplicate the code here?

tycho-indexer/src/extractor/dynamic_contract_indexer/euler/metadata_generator.rs
zizou0x: Hm, good question, I'd say we could tie it to the RPC provider. I would leave this as a potential interface refactor at the end though as this is not mandatory for the feature

tycho-indexer/src/extractor/dynamic_contract_indexer/euler/metadata_generator.rs
zizou0x: I see, I'll add the pool and direction then

PR #614, opened 2025-07-09
feat: implement metadata registries

zizou0x:
This PR polishes the implementation for the `component -> metadata` generator and `routing key -> request provider` registries. Also add some unit tests for them.


tycho-indexer/src/extractor/dynamic_contract_indexer/component_metadata.rs
```diff
@@ -120,23 +123,45 @@ pub struct MetadataGeneratorRegistry {
...
-                .default_generator
+        if let Some(hook_address) = component.static_attributes.get("hook") {
+            self.hook_generators
+                .get(hook_address)
+                .map(|boxed_generator| boxed_generator.as_ref())
+                .or(self
+                    .default_generator
+                    .as_ref()
+                    .map(|boxed_generator| boxed_generator.as_ref()))
+        } else {
```
tvinagre, line 353: If the component doesn't have a hook (which is already odd for Univ4) we shouldn't use the default, as it expects a hook address

PR #618, opened 2025-07-10
feat: Remove duplicated models

dianacarvalho1:
Done:
- Rename CurrencyToken to Token and implement methods needed for simulation
- Add id and wrapped_native_token to Chain (necessary for simulation)
- Add timestamp to Header (necessary for simulation)

Corresponding PRs:
- https://github.com/propeller-heads/tycho-simulation/pull/255
- https://github.com/propeller-heads/tycho-execution/pull/239

See [task](https://propeller-heads.atlassian.net/jira/software/projects/ENG/boards/24?selectedIssue=ENG-4705) for a summary of how models connect in all three repos


tycho-common/src/models/mod.rs
```diff
@@ -113,6 +128,31 @@ impl Chain {
...
+
+    /// Returns the wrapped native token for the chain.
+    pub fn wrapped_native_token(&self) -> Token {
+        match self {
+            Chain::Ethereum => {
+                wrapped_native_eth(Chain::Ethereum, "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2")
+            }
+            // Starknet does not have a wrapped native token
+            Chain::Starknet => {
+                wrapped_native_eth(Chain::Starknet, "0x0000000000000000000000000000000000000000")
```
dianacarvalho1, line 140: I don't like this much but I couldn't come up with a better solution 😕

tamaralipows: Shouldn't this be in some constants file? Or is this easy enough for people/us to find

dianacarvalho1: meh I think it's fine.. what I don't like is that starknet does not have a wrapped native token. I could change the signature of the method to return a result and throw an error for starknet?

PR #622, opened 2025-07-14
feat: add `RPCMetadataProvider` for batch processing of RPC requests.

zizou0x:
This PR:
- Introduces `RPCMetadataProvider` to handle batch execution of RPC requests.
- Updates `RequestProvider` trait to return results with deduplication IDs. Will be used by the orchestrator later to match requests to responses.
- Adds grouping logic for requests based on a specified batch size.
- Includes tests for the new provider and its functionality.
- Updates dependencies to include `mockito` for testing purposes.

Design refactor decision:
We decided to refactor the `RequestProvider.execute_batch` interface because it previously was reponsible for both handling the requests and decoding the responses. The decoding part can't be generically implemented and need to have custom implementation per protocol. For this reason we are now only handling requests in this function, and the decoding part will be implemented separately through a trait in a later PR.


tycho-indexer/src/extractor/dynamic_contract_indexer/rpc_metadata_provider.rs
```diff
@@ -0,0 +1,664 @@
...
+            }
+
+            let endpoint = rpc_requests[0].endpoint.clone();
+            let all_same_endpoint = rpc_requests
+                .iter()
+                .all(|rpc| rpc.endpoint == endpoint);
+
+            // If the requests are not all to the same endpoint, we can't batch them
+            // and we need to insert errors for each request
+            // TODO: if the final design still include endpoint inside RpcTransport, we need to
```
tvinagre, line 68: Yeah I want to keep the possibility of having different RPC entrypoints but I struggle to see an use case rn. So, for now, I'm fine with raising this error, makes it easier to spot that we have a configuration mistake

tycho-indexer/src/extractor/dynamic_contract_indexer/rpc_metadata_provider.rs
```diff
@@ -0,0 +1,664 @@
...
+use crate::extractor::dynamic_contract_indexer::component_metadata::{
+    DeduplicationId, MetadataError, RequestProvider, RequestTransport, RpcTransport,
+};
+
+pub struct RPCMetadataProvider {
+    client: Arc<Client>,
+    batch_size_limit: usize,
+}
+
+impl RPCMetadataProvider {
```
tvinagre, line 20: This RPC follows ETH interface so let's name it EVMRPCMetadataProvider?

tvinagre (approved):

tycho-indexer/src/extractor/dynamic_contract_indexer/rpc_metadata_provider.rs
zizou0x: It seems like a generic RPC provider to me, what do you think will only work for ETH?

tvinagre: No - you're right, it's a standard RPC call so I think it's all good

PR #623, opened 2025-07-14
feat: Implement default SwapAmountEstimator

tvinagre:
I'll move it to tycho-ethereum once the whole feature is done

### DIFF
--- a/tycho-indexer/src/extractor/dynamic_contract_indexer/entrypoint_generator.rs
+++ b/tycho-indexer/src/extractor/dynamic_contract_indexer/entrypoint_generator.rs
@@ -252,7 +252,11 @@ impl SwapAmountEstimator for DefaultSwapAmountEstimator {
                     if !balance.is_zero() {
                         let one_hundred = BigInt::from(100);
                         let amounts = vec![
-                            Bytes::from((&balance / &one_hundred).to_bytes_be().1), // 1%
+                            Bytes::from(
+                                (&balance / &one_hundred)
+                                    .to_bytes_be()
+                                    .1,
+                            ), // 1%
                             Bytes::from(
                                 (&balance / BigInt::from(10))
                                     .to_bytes_be()


tycho-indexer/src/extractor/dynamic_contract_indexer/entrypoint_generator.rs
```diff
@@ -97,8 +100,8 @@ pub trait SwapAmountEstimator {
     async fn estimate_swap_amounts(
         &self,
         metadata: &ComponentTracingMetadata,
-        pool_key: &PoolKey,
-    ) -> Result<HashMap<Address, Vec<U256>>, EntrypointGenerationError>;
+        tokens: &[Address],
+    ) -> Result<HashMap<(Address, Address), Vec<Bytes>>, EntrypointGenerationError>;
```
zizou0x, line 109: This is a complex type, can we document what this map corresponds to please?

tycho-indexer/src/extractor/dynamic_contract_indexer/entrypoint_generator.rs
```diff
@@ -182,3 +185,201 @@ impl ERC6909Overwrites {
...
+        metadata: &ComponentTracingMetadata,
+        tokens: &[Address],
+    ) -> Result<HashMap<(Address, Address), Vec<Bytes>>, EntrypointGenerationError> {
+        let mut result = HashMap::new();
+
+        // Try to use limits first (preferred)
+        if let Some(Ok((_, limits))) = &metadata.limits {
+            if !limits.is_empty() {
+                for ((token0, token1), (limit0, _limit1, _)) in limits {
+                    // limit0 is the max amount of token0 that can be sold
```
zizou0x: This should be documented on the metadata struct instead

tycho-indexer/src/extractor/dynamic_contract_indexer/entrypoint_generator.rs
```diff
@@ -182,3 +185,201 @@ impl ERC6909Overwrites {
...
+                    let limit_amount = U256::from_be_slice(limit0.as_ref());
```
zizou0x: This can panic, let's use `try_from_be_slice` instead

tycho-indexer/src/extractor/dynamic_contract_indexer/entrypoint_generator.rs
```diff
@@ -182,3 +185,201 @@ impl ERC6909Overwrites {
...
+                    if limit_amount > U256::ZERO {
+                        let amounts = vec![
+                            Bytes::from((limit_amount / U256::from(100)).to_be_bytes_vec()), // 1%
```
zizou0x: If the only reason why we have alloy is because we're doing this, we can directly use `num_bigint::BigInt` instead, see `u256_num.rs`, we used to have ethers type here and replaced them with this one

zizou0x (commented):
Overall looks good, just have a few comments for small things

PR #625, opened 2025-07-15
feat: add metadata response parser

zizou0x:
This PR adds a customizable way to decode values returned by RPC/API calls. This will need to be implemented for each hook protocol. 

It also adds the first implementation of this trait for `EulerSwap`

### DIFF
--- a/tycho-indexer/src/extractor/dynamic_contract_indexer/euler/metadata_generator.rs
+++ b/tycho-indexer/src/extractor/dynamic_contract_indexer/euler/metadata_generator.rs
@@ -1,7 +1,14 @@
-use tycho_common::models::{blockchain::Block, protocol::ProtocolComponent};
+use std::collections::HashMap;
+
+use serde_json::{json, Value};
+use tycho_common::{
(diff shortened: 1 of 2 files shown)


tycho-indexer/src/extractor/dynamic_contract_indexer/euler/metadata_generator.rs
```diff
@@ -118,12 +131,84 @@ impl MetadataRequestGenerator for EulerMetadataGenerator {
...
+
+                let mut balances = HashMap::new();
+                balances.insert(token_0.clone(), balance_0);
+                balances.insert(token_1.clone(), balance_1);
+
+                let res = MetadataValue::Balances(balances);
+                Ok(res)
+            }
+            MetadataRequestType::Tvl => {
+                let res = MetadataValue::Tvl(0.0);
```
tvinagre: Shouldn't we raise a Not Implemented Error?

zizou0x: Good catch, that was a hack to make the compiler happy but raising an error will be better here, will change this

Issue #468, opened 2024-12-27
Optimize queries for latest state by directly targeting the `default` partition


zizou0x:
### Description  

In many places, we use a filter on `valid_to` in queries to retrieve the latest state. However, most of our Postgres tables with updatable states are versioned, and all currently valid states are stored in the `default` partition of the table. 

This means that instead of querying all partitions and filtering by `valid_to`, we can directly query the `default` partition for currently valid states. 

For example, this condition in a Diesel query:  

```rust
.and(contract_storage::valid_to.gt(*MAX_VERSION_TS))
```
is equivalent and can be done much more efficiently by querying directly in the `contract_storage_default` partition. Versioning ensures that the default partition always contains valid states.

PR #628, opened 2025-07-16
perf: optimize queries by targeting `default` partition

itzlambda:
fix: #468

### DIFF
--- a/tycho-storage/src/postgres/orm.rs
+++ b/tycho-storage/src/postgres/orm.rs
@@ -370,6 +370,39 @@ pub struct ComponentBalance {
     pub valid_to: NaiveDateTime,
 }
 
+#[derive(Identifiable, Queryable, Selectable, Debug)]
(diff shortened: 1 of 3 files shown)


tycho-storage/src/postgres/orm.rs
```diff
@@ -460,19 +493,23 @@ impl PartitionedVersionedRow for NewComponentBalance {
             .zip(token_ids.iter())
             .collect::<HashSet<_>>();
 
-        let mut results: Vec<ComponentBalance> = component_balance::table
-            .select(ComponentBalance::as_select())
+        let results: Vec<ComponentBalanceDefault> = component_balance_default::table
+            .select(ComponentBalanceDefault::as_select())
```
itzlambda, line 497: Had to duplicate the `ComponentBalance` type to use with `componenet_balance_default` table. Any better way to handle this?

PR #630, opened 2025-07-17
feat: implement metadata orchestrator

zizou0x:
This PR implements the metadata orchestrator which is the main component for metadata extraction. This component gets a block and a list of components and returns the metadata for each component. For that, it uses the generator/parser and transport handler. It gets all this from a registry that needs to be filled before using it.

### DIFF
--- a/tycho-indexer/src/extractor/dynamic_contract_indexer/metadata_orchestrator.rs
+++ b/tycho-indexer/src/extractor/dynamic_contract_indexer/metadata_orchestrator.rs
@@ -373,7 +373,7 @@ mod tests {
         generator_registry.set_default_generator(Some(Box::new(mock_generator)));
 
         let mut parser_registry = MetadataResponseParserRegistry::new();
-        parser_registry.register_parser("test".to_string(), Box::new(mock_parser));
+        parser_registry.register_parser("test_protocol".to_string(), Box::new(mock_parser));
 
         let mut provider_registry = ProviderRegistry::new();
         provider_registry.register_provider("rpc_mainnet".to_string(), Arc::new(mock_provider));
@@ -493,8 +493,10 @@ mod tests {
         generator_registry.set_default_generator(Some(Box::new(mock_generator)));
 
         let mut parser_registry = MetadataResponseParserRegistry::new();
-        parser_registry
-            .register_parser("test".to_string(), Box::new(MockMetadataResponseParser::new()));
+        parser_registry.register_parser(
+            "test_protocol".to_string(),
+            Box::new(MockMetadataResponseParser::new()),
+        );
 
         let mut provider_registry = ProviderRegistry::new();
         provider_registry.register_provider("rpc_mainnet".to_string(), Arc::new(mock_provider));


tycho-indexer/src/extractor/dynamic_contract_indexer/metadata_orchestrator.rs
```diff
@@ -0,0 +1,763 @@
...
+                                all_results.push(MetadataResult {
+                                    request_id: request.request_id.clone(),
+                                    component_id: component.1.id.clone(),
+                                    request_type: request.request_type.clone(),
+                                    result: metadata_value,
+                                });
+                            }
+                        }
+                    } else {
+                        warn!("No parser found for protocol {}", request.get_protocol_name());
```
zizou0x: What should we do in that case? Is this worth panicking?

tycho-indexer/src/extractor/dynamic_contract_indexer/metadata_orchestrator.rs
```diff
@@ -0,0 +1,763 @@
...
+            let results = provider
+                .execute_batch(&transports)
+                .await;
+            for (dedup_id, result) in results {
+                if let Some(request) = ids_to_requests.get(&dedup_id) {
+                    if let Some(parser) = self
+                        .response_parser_registry
+                        .get_parser(request.get_protocol_name())
+                    {
+                        if let Some(component_ids) = component_ids_by_dedup_id.get(&dedup_id) {
```
zizou0x: A lot of if conditions here, most of them should never happen, but not sure how we should handle the bad path. We can either stay like this and simply skip, or we can panic. Another alternative would be to forward an error for the whole batch but I suppose that would lead to a panic anyway?

tycho-indexer/src/extractor/dynamic_contract_indexer/component_metadata.rs
```diff
@@ -24,12 +20,13 @@ type Balances = HashMap<Address, Bytes>;
 // Here we link the limits to the entrypoint that triggered the limit fetching. This is necessary
 // so that Tycho Simulation can use this information to calculate the limits for the component on
 // every block.
-type Limits = Vec<((Address, Address), (Bytes, Bytes, Option<EntryPointWithTracingParams>))>;
```
tvinagre: Why did we remove the EntryPointWithTracingParams  here?

tycho-indexer/src/extractor/dynamic_contract_indexer/component_metadata.rs
```diff
@@ -64,12 +91,26 @@ impl Clone for MetadataRequest {
...
-        request_id: RequestId,
+        protocol_name: String,
+        request_name: String,
         component_id: ComponentId,
         request_type: MetadataRequestType,
         transport: Box<dyn RequestTransport>,
     ) -> Self {
-        Self { request_id, component_id, request_type, transport }
+        Self {
+            request_id: format!("{protocol_name}_{request_name}"),
```
tvinagre: Why require a protocol_name?

tycho-indexer/src/extractor/dynamic_contract_indexer/component_metadata.rs
```diff
@@ -269,6 +311,31 @@ pub trait MetadataResponseParser: Send + Sync {
...
+}
+
+impl MetadataResponseParserRegistry {
+    pub fn new() -> Self {
+        Self { parsers: HashMap::new() }
+    }
+
+    pub fn register_parser(
+        &mut self,
+        protocol_name: String,
```
tvinagre, line 323: Shouldn't the parser be linked 1-1 to a request generator / provider and not to the protocol?

tycho-indexer/src/extractor/dynamic_contract_indexer/metadata_orchestrator.rs
tvinagre: Idk if we should panic but, at least emit a warning so we know there is a logic error. But I don't see a problem in panicking if it's for a condition that really shouldn't happen. 
I just don't like to simply ignore, as these conditionals can serve as a safety net for us

tycho-indexer/src/extractor/dynamic_contract_indexer/metadata_orchestrator.rs
tvinagre: Hm yes I think so. If we have registered a provider without a corresponding parser then we should panic so we can fix the issue IMO.

tvinagre: Or, since it's bad to panic on a real-time application, we can error and handle this on the caller. We'd mark all the affected components as paused - makes it easy for us to spot a failure as well.

tycho-indexer/src/extractor/dynamic_contract_indexer/component_metadata.rs
zizou0x: Yeah the naming is bad here, it not really a protocol_name, it's the generator/parser name

It's needed so we know what parser to use for what request. The alternative would be to make this separate from the `request_id` and add a new field but I thought this was simpler for now

tycho-indexer/src/extractor/dynamic_contract_indexer/component_metadata.rs
zizou0x: I found it weird, `EntryPoint` and `TracingParams` aren't there in any other interface so I'm not sure how to get this so deep into the metadata logic. Also the `ComponentTracingMetadata` would be weird no? I was hoping we could link them further up in the logic. Not sure if this will be realistic or not

tycho-indexer/src/extractor/dynamic_contract_indexer/component_metadata.rs
zizou0x: Answered above

tycho-indexer/src/extractor/dynamic_contract_indexer/component_metadata.rs
zizou0x: Changed the interface to have a separate field, please have another look

PR #634, opened 2025-07-22
refactor: make extractor return a specific type

zizou0x:
This commit improves how we deal with our types and communication with tycho-client.

The main objective here was to be able to detect any breaking change at compile time instead of runtime. We now deal with specific types to ensure serialization and deserialization are on the same Rust struct `dto::BlockChanges` and equivalent dto for `Command` and `Response`. We also removed the `NormalisedMessage` trait because it was not needed anymore.


tycho-indexer/src/services/ws.rs
```diff
@@ -296,41 +299,20 @@ impl Actor for WsActor {
...
         ctx: &mut Self::Context,
     ) {
         trace!("Message received from extractor");
         match msg {
-            Ok((subscription_id, deltas)) => {
+            Ok((subscription_id, block_delta)) => {
                 trace!("Forwarding message to client");
-                let msg = DeltasMessage { subscription_id, deltas };
+                let msg =
+                    WebSocketMessage::BlockChanges { deltas: block_delta.into(), subscription_id };
```
zizou0x: This is where we convert from `BlockAggregatedChanges` into dto object.

Implementation is in tycho-common, this will help highlighting breaking changes, we just have to look into the `From` implementation to see anything that changed

tycho-indexer/src/services/ws.rs
```diff
@@ -197,7 +200,7 @@ impl WsActor {
                                     let light = item.drop_state();
                                     yield Ok((subscription_id, light));
                                 } else {
-                                    yield Ok((subscription_id, item));
+                                    yield Ok((subscription_id, (*item).clone()));
```
kayibal: What is going on here? Dereferencing Arc?

tycho-indexer/src/services/ws.rs
kayibal: Does this mean the aggregation of changes happens once per subscription?

tycho-indexer/src/services/ws.rs
zizou0x: Yes exactly

tycho-indexer/src/services/ws.rs
zizou0x: Yes, I've changed it to only happen once in the stream above

PR #636, opened 2025-07-24
fix: Remove missing state query on versioning

tvinagre:
The query was too slow and it's not necessary for ProtocolStates, since only one extractor should be able to mutate a ProtocolState.


tycho-storage/src/postgres/orm.rs
```diff
@@ -1732,8 +1710,6 @@ impl PartitionedVersionedRow for NewSlot {
             .filter(|id| !found_ids.contains(id))
             .collect();
 
-        // If we have missing ids, we need to query the archived tables as well. This is necessary
-        // when entries are deleted
```
zizou0x, line 1736: Why did you remove it from here and not above?

zizou0x (approved):
LGTM

PR #639, opened 2025-07-24
feat: Add default to FeedMessage

dianacarvalho1:


tycho-client/src/feed/mod.rs
```diff
@@ -375,7 +375,7 @@ impl SynchronizerStream {
 }
 
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
-pub struct FeedMessage<H>
+pub struct FeedMessage<H = BlockHeader>
```
tamaralipows, line 378: nice trick that I just learnt last week in the book meetings lol

tamaralipows (approved):
Thank you!

tycho-client/src/feed/mod.rs
dianacarvalho1: I don't think this fully worked though 😅 Will try to look into it more next week though

PR #642, opened 2025-07-25
fix: Websocket deadlocks

kayibal:
**Problem**: WebSocket subscriptions were deadlocking when multiple clients subscribed simultaneously, causing timeouts and blocking the entire runtime.

**Root Cause**: block_on() called while holding a mutex in Actix actor context, which blocks the async runtime thread and prevents other actors from processing messages.

**Solution**:
1. Replace blocking with async spawning - Use ctx.spawn() instead of block_on() to prevent runtime blocking
2. Eliminate mutex entirely - Since subscribers map is read-only after startup, replace Arc<Mutex<HashMap>> with Arc<HashMap> for direct access

Test Coverage: Added comprehensive deadlock reproduction test with SlowMessageSender that simulates the original issue.

This fixes the production deadlock while providing significant performance improvements for concurrent WebSocket operations.

### DIFF
--- a/tycho-indexer/src/services/ws.rs
+++ b/tycho-indexer/src/services/ws.rs
@@ -2,7 +2,7 @@
 use std::{
     collections::HashMap,
     fmt::Debug,
-    sync::{Arc, Mutex},
+    sync::Arc,
     time::{Duration, Instant},
 };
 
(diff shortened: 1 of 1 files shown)


tycho-indexer/src/services/ws.rs
```diff
@@ -242,8 +226,91 @@ impl WsActor {
...
+        let actor_id = self.id;
+        let user_identity = self.user_identity.clone();
+        let extractor_id_for_future = extractor_id.clone();
+        let extractor_id_for_error = extractor_id.clone();
+        
+        // Step 3: Create async future for subscription setup
+        // This future will run independently without blocking the actor's message processing
+        // Use async operation instead of block_on to prevent runtime deadlocks
+        let fut = async move {
+            match message_sender.subscribe().await {
```
tvinagre, line 262: I think this is fine bc the current ExtractorHandle has a timeout, but wouldn't it be safe to implement a timeout in here as well instead of relying on the inner implementation?

PR #648, opened 2025-08-08
fix: WsDeltas client forever blocking subscribe.

kayibal:
The WsDeltasClient assumed an infinite reconnection loop although later a max_reconnect_attempts was introduced to avoid spamming the server with reconnects. Unfortunately subscribe and unsubscribe would wrongly wait indefinitely until the connection is reestablished.

This fix introduced a simple flag to mark a client as dead and return an error if we try to subscribe on a client that maxed out it's reconnections attempts.

### DIFF
--- a/tycho-client/src/deltas.rs
+++ b/tycho-client/src/deltas.rs
@@ -6,7 +6,7 @@
 //!
 //! ## Websocket Implementation
 //!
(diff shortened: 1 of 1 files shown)


zizou0x (approved):
LGTM

PR #655, opened 2025-08-13
feat: Track all storage slots on DCI, excluding tokens or blacklist addr

tvinagre:


tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
```diff
@@ -27,6 +27,11 @@ pub(super) struct DCICache {
     pub(super) retriggers: VersionedCache<StorageLocation, HashSet<EntryPointWithTracingParams>>,
     /// Stores tracked contract addresses and their associated storage keys.
     pub(super) tracked_contracts: VersionedCache<Address, Option<HashSet<StoreKey>>>,
+    /// Stores addresses identified as ERC-20 tokens to skip full indexing.
+    pub(super) erc20_addresses: VersionedCache<Address, bool>,
```
zizou0x, line 32: I'd add a "perf: implement a versioned wrapper around HashSet, similar to `VersionedCache` for `HashMap`"

Because here we don't actually need this to be a key/value store. Just a key set would be enough

PR #657, opened 2025-08-14
fix: detect account dependencies from non-call opcodes in EVM tracer

zizou0x:
The tracer was only detecting account dependencies from call opcodes (CALL, DELEGATECALL, etc.) but missing dependencies created by other opcodes like BALANCE and EXTCODESIZE that also access external accounts.

This fix switches to using eth_createAccessList, which comprehensively detects all account dependencies regardless of the opcode that creates them.

Todo:

- [ ] Remove ethers+alloy usage in the Ethereum tracer. Only keep alloy


tycho-ethereum/examples/trace.rs
```diff
@@ -0,0 +1,485 @@
+use std::{
```
zizou0x: I'll remove this, it was just for me to test but it's very dirty. Don't spend time reviewing this

tycho-ethereum/src/entrypoint_tracer/tracer.rs
```diff
@@ -102,6 +111,45 @@ impl EVMEntrypointService {
...
             .map_err(RPCError::RequestError)
     }
+
+    async fn get_access_list(
+        &self,
+        block_hash: &BlockHash,
+        target: &Address,
+        params: &RPCTracerParams,
+    ) -> Result<HashMap<Address, HashSet<Bytes>>, RPCError> {
+        let state_overrides = build_state_overrides(
```
zizou0x: I can move that in the `else {` branch

kayibal (approved):

zizou0x:
The todo is going to be addressed in a separate PR

PR #658, opened 2025-08-15
fix: Update Entrypoint DB Schema to include external_id on uniqueness validation

tvinagre:
Since we've upgraded the Hooks Entrypoint_ids to be `{hook}:execute` - but all of them call the same `{mock_contract}:execute` target, signature we were getting issues on the DB not accepting different Entrypoints with the same target,signature combination. 
This PR updates the Uniqueness validation on Entrypoint table to allow one Entrypoint per-hook.


tycho-indexer/src/extractor/dynamic_contract_indexer/hook_dci.rs
```diff
@@ -639,10 +639,26 @@ where
...
+            let dci_span = span!(Level::INFO, "inner_dci_processing");
+            let _dci_guard = dci_span.enter();
+
+            self.inner_dci
                 .process_block_update(block_changes)
-                .await;
+                .await?;
+
+            drop(_dci_guard);
+            info!("Inner DCI processing completed");
```
zizou0x: I find this complex just to spawn a span. Did you check the `.instrument`?

For example we have this in tycho-indexer

```
let res = async move {
            trace!("tx");
            let res: R = self.inner.request(method, params).await.map_err(Into::into)?;
            trace!(rx = ?serde_json::to_string(&res)?);
            Ok::<_, ProviderError>(res)
        }
        .instrument(span)
```

kayibal: Yeah this span here is not async aware..

tycho-storage/migrations/2025-08-14-fix-entry-point-unique-constraint/up.sql
```diff
@@ -0,0 +1,15 @@
...
+
+-- Add the correct unique constraint on (external_id, target, signature)
+-- This ensures complete uniqueness while allowing multiple hooks to point to the same target+signature
+-- as long as they have different external_ids
+ALTER TABLE entry_point ADD CONSTRAINT entry_point_external_id_target_signature_key 
+    UNIQUE (external_id, target, signature);
+
+-- Add an index on (target, signature) for query performance
+CREATE INDEX IF NOT EXISTS idx_entry_point_target_signature ON entry_point(target, signature);
+
```
kayibal, line 13: This is a huge change to the model. Not sure we should do that right now...

PR #664, opened 2025-08-19
fix: Fetch new storage slots detected after a new tracing

tvinagre:


tycho-indexer/src/extractor/dynamic_contract_indexer/dci.rs
```diff
@@ -1890,4 +1918,280 @@ mod tests {
...
+
+    // TODO: Add test for the case where no new slots are found
+    // This test currently fails because there's a bug where account extraction
+    // is still called even when there are no new slots. The bug is that our
+    // new slot filtering logic isn't being applied correctly in all cases.
+    //
+    // The test would verify that when a tracked token accesses only slots that
+    // are already in the cache, no account extraction should happen.
+    //
+    // Uncomment and fix the implementation to make this test pass:
```
zizou0x: What does it mean?

tycho-indexer/src/extractor/dynamic_contract_indexer/dci.rs
```diff
@@ -222,20 +222,38 @@ where
...
+                                    .filter(|slot| !tracked_slots.contains(*slot))
+                                    .cloned()
+                                    .collect()
+                            } else {
+                                // None means all slots are tracked, so no new slots
+                                HashSet::new()
+                            }
+                        } else {
+                            // Account is not tracked at all, all slots are new
+                            slots.iter().cloned().collect()
```
zizou0x, line 245: I'm confused, shouldn't we do this only if it's a blacklisted contract?

tycho-indexer/src/extractor/dynamic_contract_indexer/dci.rs
tvinagre: I've fixed it and forgot to remove - removing

tycho-indexer/src/extractor/dynamic_contract_indexer/dci.rs
tvinagre: No - we do this filtering later, when doing the proper storage_request. The cache always contains the "accessed_slots" - should never be empty

tvinagre: Filtering is done here:
```rust 
                    if !self.should_skip_full_indexing(address) {
                        // Process all slots for non-token or non-blacklisted contracts
                        Ok(StorageSnapshotRequest { address: address.clone(), slots: None })
```

zizou0x (approved):
LGTM

PR #665, opened 2025-08-19
fix: Fix postgres dockerfile after upstream update

kayibal:


zizou0x (approved):

zizou0x:
Fixed in https://github.com/propeller-heads/tycho-indexer/pull/669

We can come back to this bigger update when needed but for now we're safer with a frozen version of the base image

PR #667, opened 2025-08-20
feat: Add v1 interfaces

kayibal:


tycho-common/examples/chain/mod.rs
```diff
@@ -0,0 +1,41 @@
+//! Action chaining system for composing multi-step DeFi operations.
+//!
+//! This module provides a type-safe builder pattern for creating chains of actions
```
dianacarvalho1, line 3: I find that using the name "Chain" for this might be very confusing.. chain already means something in this context 😬 maybe sequence?

tycho-common/examples/action_chaining/chain/builder.rs
dianacarvalho1, line 1: these files are all repeated in tycho-common/examples/chain

tycho-common/examples/action_chaining/uniswap_v2.rs
```diff
@@ -0,0 +1,263 @@
+//! UniswapV2 implementation for the action chaining playground.
```
dianacarvalho1, line 1: would we like to port all the current protocols to this?

PR #672, opened 2025-09-02
feat: Add slot offset to tracing result

kayibal:
Adds a slot offset to each retrigger in TracingResults. This forms the basis to later only fire retriggers in case the actual address changed and not some other coincidentally "packed" data.


tycho-common/src/dto.rs
```diff
@@ -1689,17 +1689,42 @@ impl From<models::blockchain::EntryPointWithTracingParams> for EntryPointWithTra
...
+    fn from(value: models::blockchain::AddressStorageLocation) -> Self {
+        Self { key: value.key, offset: value.offset }
+    }
+}
+
+impl From<Bytes> for AddressStorageLocation {
+    fn from(value: Bytes) -> Self {
+        Self { key: value, offset: 0 }
+    }
+}
```
zizou0x: Weren't you supposed to get rid of this? Have you looked into what breaks if you remove it? I'm afraid it could cause bugs

tycho-ethereum/src/entrypoint_tracer/tracer.rs
```diff
@@ -325,8 +326,12 @@ impl EntryPointTracer for EVMEntrypointService {
                                         .any(|window| window == address_bytes)
                                     {
                                         retriggers.insert((
-                                            Bytes::from(address.as_slice()),
-                                            Bytes::from(slot.as_slice()),
+                                            tycho_common::Bytes::from(address.as_ref() as &[u8]),
+                                            // TODO: add offset here
```
zizou0x: This will be done in another PR? It would make sense to do it here

tycho-common/src/dto.rs
zizou0x: I see that it's used for some tests, if it's only this then please gate it behind `cfg(test)` but honestly I think we should just be explicit

tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
```diff
@@ -23,8 +23,9 @@ pub(super) struct DCICache {
     /// Stores tracing results for entry points paired with specific tracing parameters.
     pub(super) entrypoint_results: VersionedCache<(EntryPointId, TracingParams), TracingResult>,
     /// Maps a storage location to entry points that should be retriggered when that location
-    /// changes.
-    pub(super) retriggers: VersionedCache<StorageLocation, HashSet<EntryPointWithTracingParams>>,
+    /// changes and the address storage offset for packed slots.
+    pub(super) retriggers:
+        VersionedCache<StorageLocation, (HashSet<EntryPointWithTracingParams>, u8)>,
```
zizou0x, line 44: This is safe because only one address can fit in a EVM slot. This can become problematic if we export this to a vm in which two addresses can fit in a single slot (doesn't exist AFAIK but who knows)

I think we shoul add a small comments here in case one day we want to use Tycho on another VM

zizou0x (commented):
Looks good, just a few comments/questions

tycho-ethereum/src/entrypoint_tracer/tracer.rs
kayibal: Yes it's done here, I am avoiding touching any logic on this PR:

https://github.com/propeller-heads/tycho-indexer/pull/676

tycho-common/src/dto.rs
kayibal: It's already removed

tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
kayibal: I think it is unlikely for a VM to use more than 256 bits.... if a comment is required maybe it should be on AddressStorageLocation? Changing the type there will cause a compilation error here.

tycho-common/src/dto.rs
kayibal: Oh I see I forgot to remove it from the dto module...

kayibal: Fixed now

tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
zizou0x: This comment wasn't about using `u8`, it's about the fact that this strictly maps one slot to one offset. 

This works fine for EVM because only one address fit into a slot. But I was wondering what would happen if we go on a VM in which two addresses could be packed in a single slot. 

This would be very very edge case and it will most likely never happen but I'd just add a small comment about this. Feel free to ignore it if you think it's not necessary

PR #673, opened 2025-09-03
fix: correctly get all values from cache when needed.

zizou0x:
In some cases we need a full cache lookup. For example when getting retriggers or tracked slots we want to return all the elements and not only the last one.

This commit adds a `get_all` function on the `VersionedCache` to allow for this. We now use it instead of a simple `.get()` when needed inside DCI implementation

### DIFF
--- a/tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
+++ b/tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
@@ -14,6 +14,7 @@ use tycho_common::models::{
 type StorageLocation = (Address, StoreKey);
 
 /// Central data cache used by the Dynamic Contract Indexer (DCI).
+#[derive(Debug)]
 pub(super) struct DCICache {
     /// Maps entry point IDs to entry point definitions.
     pub(super) ep_id_to_entrypoint: VersionedCache<EntryPointId, EntryPoint>,
@@ -117,7 +118,7 @@ pub enum DCICacheError {
 /// A versioned data container scoped to a specific block.
 ///
 /// Stores key-value pairs for a block, used in the pending portion of a cache.
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 struct BlockScopedMap<K, V> {
     /// Block metadata for this layer.
     block: Block,
@@ -130,6 +131,7 @@ struct BlockScopedMap<K, V> {
 /// It contains:
 /// - Permanent data that is not revertable.
 /// - Pending data organized in layers per block, supporting reverts.
+#[derive(Debug)]
 pub(super) struct VersionedCache<K, V> {
     /// Entries that are permanent and not affected by block reverts.
     permanent: HashMap<K, V>,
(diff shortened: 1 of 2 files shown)


tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
```diff
@@ -204,6 +204,27 @@ where
...
+        if let Some(v) = self.permanent.get(k) {
+            res.push(v);
+        }
+
+        if res.is_empty() {
+            None
+        } else {
+            Some(res)
+        }
+    }
```
kayibal: I think this would be cleaner if you just returned an iterator here. 

Or alternatively the return collection could be generic too:

```rust
pub(super) fn get_all<C>(&self, k: &K) -> Option<C>
where
    C: FromIterator<&V>,
{
    let mut iter = self
        .pending
        .iter()
        .rev()
        .filter_map(|layer| layer.data.get(k))
        .chain(self.permanent.get(k));

    let res: C = iter.collect();

    if res.into_iter().next().is_none() {
        None
    } else {
        Some(res)
    }
}
```

tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
```diff
@@ -807,4 +840,112 @@ mod tests {
...
+        assert_eq!(result.collect::<Vec<_>>(), vec![&603, &601]); // Latest to oldest pending only
+
+        // Test 8: After revert, get_all should reflect the new state
+        cache.revert_to(&block2.hash).unwrap();
+        let result = cache
+            .get_all("multi".to_string())
+            .unwrap();
+        assert_eq!(result.collect::<Vec<_>>(), vec![&402, &401, &400]); // Block3 data removed after revert
+
+        // Test 9: After finality, get_all should include finalized data in permanent
```
kayibal, line 940: Not going to block here, but I don't like this scheme of packing multiple cases into a single test

tycho-indexer/src/extractor/dynamic_contract_indexer/dci.rs
```diff
@@ -604,13 +604,37 @@ where
...
+                        let mut result = HashSet::new();
+                        let mut whole_contract_tracked = false;
+
+                        for key in keys {
+                            match key {
+                                None => {
+                                    // None is winning over specific keys, it means the whole
+                                    // contract is tracked
+                                    whole_contract_tracked = true;
+                                    break;
```
kayibal, line 624: Why not return here directly?

kayibal (approved):

tycho-indexer/src/extractor/dynamic_contract_indexer/dci.rs
zizou0x: This is just for the inner loop but we still want outer loops to run

zizou0x: So the idea here is that we consume the iterator only once. This is gonna change when merging with my other PR anyway. But it should work exactly like it was before, just we are lazily allocating memory

tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
zizou0x: I generally agree with this. Here I tried to use rstest but it was too complex, the alternative was to make one test for each case but it was a lot of code overhead so I've decided to keep it like this for now

PR #674, opened 2025-09-03
fix: properly order tokens when parsing euler response

zizou0x:
Tycho doesn't ensure that ProtocolComponent.tokens order is maintained. To properly assign token balances we need to assign token_0 and token_1 to tokens sorted in ascending order. Euler confirmed that they will always ensure this order is respected.


tvinagre (approved):

tvinagre (commented):
Looks good, just pls double check if we're not having the same risk anywhere else on HooksDCI

PR #675, opened 2025-09-03
fix: DCI cache fixes

zizou0x:
Resolves two caching issues:

1. **Tracked contract cache semantics**: Fixed `None` value interpretation mismatch that caused unintended full contract indexing instead of skipping storage indexing (more details [here](https://propeller-heads.atlassian.net/browse/ENG-4986))
2. **VersionedCache finality**: Added custom finality logic to properly merge set values instead of replacing them (more details [here](https://propeller-heads.atlassian.net/browse/ENG-4984))

### DIFF
--- a/tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
+++ b/tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
@@ -18,6 +18,8 @@ type StorageLocation = (Address, StoreKey);
 
 /// A function used for merging values when the same key exists in multiple layers during finality
 /// handling.
+// Perf: all the merge functions are defined at compile time, so we can avoid the overhead of
+// dynamic dispatch.
 type MergeFunction<V> = Box<dyn Fn(&V, V) -> V>;
 
 /// Strategy for merging values when the same key exists in multiple layers during finality


tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
```diff
@@ -16,6 +16,19 @@ use super::hook_dci::ComponentProcessingState;
...
+
+/// Strategy for merging values when the same key exists in multiple layers during finality
+/// handling.
+pub(super) enum MergeStrategy<V> {
+    /// Replace existing values with new ones (default behavior).
+    Replace,
+    /// Use a custom function to merge values.
+    Custom(MergeFunction<V>),
+}
+
```
kayibal: I think this is overkill because you are allowing the cache to change merge strategy at runtime. I don't think we need that. For a given cache the MergeStrategy should always be the same (decided at compile time).

I think it is too late now but I would have done it like this:

```rust
pub trait Merge<V> {
    fn merge(permanent: &mut V, value: V);
}

pub struct Replace;
impl<V> Merge<V> for Replace {
    fn merge(permanent: &mut V, value: V) {
        *permanent = value;
    }
}

pub struct ExtendMerge;
impl<T, V> Merge<V> for ExtendMerge
where
    V: Extend<T>,
{
    fn merge(permanent: &mut V, value: V) {
        permanent.extend(value);
    }
}

pub fn handle_finality<M: Merge<V> = Replace>(
    &mut self,
    finalized_block_height: u64,
) -> Result<(), DCICacheError> {
    // …
    for layer in finalized_layers {
        M::merge(&mut self.permanent, layer.data);
    }
    Ok(())
}
```

kayibal (approved):

tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
zizou0x: I completely agree, we can implement this with static dispatch. For now I decided not to for simplicity. I'll add a `perf` todo to come back to this when we have more time allocated.

PR #676, opened 2025-09-04
feat: DCI correctly handle packed slot retriggers

kayibal:
- Correctly sets offset when while creating retriggers during tracing
- Handle retriggers correctly if a contract uses packed slots:
     - Processes contract storage change diffs and uses offset to detect if the actual address value changed 
- Improves logging and traces on the DCI


tycho-indexer/src/extractor/dynamic_contract_indexer/dci.rs
```diff
@@ -696,29 +731,37 @@ where
...
-        if !retriggered_entrypoints.is_empty() {
-            let retrigger_log: Vec<String> = retriggered_entrypoints
-                .keys()
-                .map(|e| e.entry_point.external_id.clone())
-                .collect();
-            tracing::info!("DCI: Retriggered entrypoints: {:?}", retrigger_log);
+        Ok(retriggered_entrypoints)
+    }
+
+    fn retrigger_address_changed(
```
zizou0x, line 872: Please add small documentation here

tycho-common/src/models/blockchain.rs
```diff
@@ -633,6 +642,18 @@ impl TracedEntryPoint {
     }
 }
 
+impl std::fmt::Display for TracedEntryPoint {
```
tvinagre, line 673: Thank you! 🙏🏽

PR #677, opened 2025-09-05
feat: implement pausing component in DCI

zizou0x:
This PR resolves [ENG-4707](https://propeller-heads.atlassian.net/jira/software/projects/ENG/boards/24/backlog?label=Cairoline&selectedIssue=ENG-4707) and [ENG-4844](https://propeller-heads.atlassian.net/jira/software/projects/ENG/boards/24/backlog?label=Cairoline&selectedIssue=ENG-4844).

We've added a configurable retry logic inside the ethereum tracer and added the logic in DCI and its cache to emit the "paused" state attribute for all the related components when a tracing fails after all the retries. 

[ENG-4707]: https://propeller-heads.atlassian.net/browse/ENG-4707?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ENG-4844]: https://propeller-heads.atlassian.net/browse/ENG-4844?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tycho-common/src/models/blockchain.rs
+++ b/tycho-common/src/models/blockchain.rs
@@ -534,13 +534,13 @@ impl std::fmt::Display for RPCTracerParams {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         let caller_str = match &self.caller {
             Some(addr) => format!("caller={}", addr),
-            None => "caller=default".to_string(),
+            None => String::new(),
         };
 
-        let calldata_str = if self.calldata.len() >= 4 {
+        let calldata_str = if self.calldata.len() >= 8 {
             format!(
                 "calldata=0x{}..({} bytes)",
-                hex::encode(&self.calldata[..4]),
+                hex::encode(&self.calldata[..8]),
                 self.calldata.len()
             )
         } else {
--- a/tycho-common/src/models/blockchain.rs
+++ b/tycho-common/src/models/blockchain.rs
@@ -530,6 +530,34 @@ impl RPCTracerParams {
     }
 }
 
+impl std::fmt::Display for RPCTracerParams {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let caller_str = match &self.caller {
+            Some(addr) => format!("caller={}", addr),
+            None => "caller=default".to_string(),
+        };
+
+        let calldata_str = if self.calldata.len() >= 4 {
+            format!(
+                "calldata=0x{}..({} bytes)",
+                hex::encode(&self.calldata[..4]),
+                self.calldata.len()
+            )
+        } else {
+            format!("calldata={}", self.calldata)
+        };
+
+        let overrides_str = match &self.state_overrides {
+            Some(overrides) if !overrides.is_empty() => {
+                format!(", {} state override(s)", overrides.len())
+            }
+            _ => String::new(),
+        };
+
+        write!(f, "{}, {}{}", caller_str, calldata_str, overrides_str)
+    }
+}
+
 // Ensure serialization order, required by the storage layer
 impl Serialize for RPCTracerParams {
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
--- a/tycho-ethereum/src/entrypoint_tracer/tracer.rs
+++ b/tycho-ethereum/src/entrypoint_tracer/tracer.rs
@@ -18,6 +18,7 @@ use alloy_rpc_types_trace::geth::{
 };
 use async_trait::async_trait;
 use serde_json::{json, value::to_raw_value, Value};
+use tracing::error;
 use tycho_common::{
     keccak256,
     models::{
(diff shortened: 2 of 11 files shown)


tycho-ethereum/src/entrypoint_tracer/tracer.rs
```diff
@@ -296,72 +340,123 @@ impl EntryPointTracer for EVMEntrypointService {
...
+            to_retry = failed_retryable;
+            retry_count += 1;
+
+            // If no retryable failures, break the loop
+            if to_retry.is_empty() {
+                break;
+            }
+        }
+
+        // This should never happen, if it does, we log an error and return the results that we have
```
zizou0x, line 459: I'm wondering if this is the best approach. Alternatively we could panic or change the interface to be `Result<Vec<Result<TracedEntryPoint, Self::Error>>>`

tycho-common/src/models/blockchain.rs
```diff
@@ -530,6 +530,34 @@ impl RPCTracerParams {
     }
 }
 
+impl std::fmt::Display for RPCTracerParams {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let caller_str = match &self.caller {
+            Some(addr) => format!("caller={}", addr),
+            None => "caller=default".to_string(),
```
kayibal: Just omit if default I'd say, keep things short.

tycho-common/src/models/blockchain.rs
```diff
@@ -530,6 +530,34 @@ impl RPCTracerParams {
...
+        };
+
+        let calldata_str = if self.calldata.len() >= 4 {
```
kayibal: Probably the most important data, I would allow this to be a bit longer..

tycho-ethereum/src/entrypoint_tracer/tracer.rs
```diff
@@ -216,7 +227,10 @@ impl EVMEntrypointService {
         ]);
 
         let batch_params = to_raw_value(&batch_request).map_err(|e| {
-            RPCError::UnknownError(format!("Failed to serialize batch params: {}", e))
+            RPCError::UnknownError(format!(
+                "Failed to serialize batch params for {} (block: {}, params: {}): {}",
+                target, block_hash, params, e
+            ))
```
kayibal: Since you are doing a lot of work in error handling here, can you please store the original error in all of these enums please?

```rust

#[derive(Error, Debug)]
pub struct MyError {
    msg: String,
    #[source]  // optional if field name is `source`
    source: anyhow::Error,
}

```

tycho-ethereum/src/entrypoint_tracer/tracer.rs
```diff
@@ -296,72 +340,123 @@ impl EntryPointTracer for EVMEntrypointService {
...
+                            .batch_trace_and_access_list(
+                                &entry_point.entry_point.target,
+                                rpc_entry_point,
+                                &block_hash,
+                            )
+                            .await
+                        {
+                            Ok(trace) => trace,
+                            Err(e) => {
+                                if e.should_retry() && retry_count < self.max_retries {
```
kayibal: I wouldn't check the retry_count here again since you already do this outside. Imo just makes this more complex to reason about.

tycho-ethereum/src/entrypoint_tracer/tracer.rs
```diff
@@ -296,72 +340,123 @@ impl EntryPointTracer for EVMEntrypointService {
...
+            }
+
+            // Update retry list and increment counter
+            to_retry = failed_retryable;
+            retry_count += 1;
+
+            // If no retryable failures, break the loop
+            if to_retry.is_empty() {
+                break;
+            }
```
kayibal: So many conditions to exit this loop. This one seems superfluous or if not should be moved to the while condition.

tycho-ethereum/src/entrypoint_tracer/tracer.rs
zizou0x: This one is actually necessary to properly insert the latest error in the final result. I've added an intermediate bool to make it easier to understand

tycho-common/src/models/blockchain.rs
zizou0x: Honestly I hesitated but couldn't come up with something that makes sense as a size beside just the signature. 

Opened for suggestion on the right balance between too floody and not enough data. I will also add `warn` logs into the tracer. The idea is that if a tracing fails we should be able to investigate

tycho-ethereum/src/entrypoint_tracer/tracer.rs
kayibal: We usually have between 1 and 10? params.. so I think it would be nice if 2 or max 1 fits a standard terminal line: 80-120 chars. Means that you can comfortable view a whole entrypoint in a single page.

tycho-ethereum/src/lib.rs
```diff
@@ -25,10 +53,20 @@ pub enum RPCError {
     RequestError(String),
     #[error("Tracing failure: {0}")]
     TracingFailure(String),
+    #[error("Serialize error: {0}")]
+    SerializeError(SerdeJsonError),
+    #[error("Reqwest error: {0}")]
+    ReqwestError(ReqwestError),
```
kayibal: How ist this conceptually different from RequestError? I think I'd prefer if they were the same

kayibal (approved):

PR #679, opened 2025-09-05
feat: implement Balance slot detector

tvinagre:


tycho-common/src/traits.rs
```diff
@@ -146,3 +146,32 @@ pub trait EntryPointTracer: Sync {
...
+        components: Vec<(ComponentId, Vec<Address>)>,
+        holder: Address,
+        block_hash: BlockHash,
+    ) -> HashMap<ComponentId, Result<HashMap<Address, Bytes>, Self::Error>>;
+
+    /// Set the maximum number of components to process concurrently
+    fn set_max_concurrent(&mut self, max: usize);
+
+    /// Get the current max concurrent setting
+    fn max_concurrent(&self) -> usize;
```
kayibal: Does this have to be changed at runtime? If not I think the trait doesn't need it, it can just be part of the constructor.

tycho-ethereum/src/entrypoint_tracer/mod.rs
```diff
@@ -16,6 +16,8 @@ use tycho_common::{
 
 use crate::RPCError;
 
+#[cfg(feature = "onchain_data")]
+pub mod evm_balance_slot_detector;
```
kayibal: Since we are in a ethereum scoped package, I think the evm part of the module could be omitted.

tycho-indexer/src/extractor/dynamic_contract_indexer/entrypoint_generator.rs
```diff
@@ -499,13 +501,22 @@ impl DefaultSwapAmountEstimator {
...
 /// Generates entrypoints using V4MiniRouter for tracing hook interactions
-pub struct UniswapV4DefaultHookEntrypointGenerator<E: SwapAmountEstimator> {
+pub struct UniswapV4DefaultHookEntrypointGenerator<E, B>
+where
+    E: SwapAmountEstimator,
+    B: BalanceSlotDetector,
+{
     config: HookEntrypointConfig,
     estimator: E,
+    balance_slot_detector: Option<B>,
```
kayibal: Surprised this is optional...

tycho-common/src/traits.rs
```diff
@@ -146,3 +146,32 @@ pub trait EntryPointTracer: Sync {
...
+    /// * `block_hash` - Block at which to detect slots
+    ///
+    /// # Returns
+    /// HashMap mapping component_id -> Result containing (token_address -> storage_slot) or error
+    async fn detect_slots_for_components(
+        &self,
+        components: Vec<(ComponentId, Vec<Address>)>,
+        holder: Address,
+        block_hash: BlockHash,
+    ) -> HashMap<ComponentId, Result<HashMap<Address, Bytes>, Self::Error>>;
```
kayibal: Why does this take components instead of tokens? I think instead of components this should take a list of tokens so we can later ideally move this to the token analyzer. The main reason I'd like to move this to the analyzer is so we can persist the slot information and update it manually if required.

tycho-indexer/src/extractor/dynamic_contract_indexer/entrypoint_generator.rs
```diff
@@ -583,6 +598,54 @@ impl<E: SwapAmountEstimator + Send + Sync> HookEntrypointGenerator
 
         info!(swap_pair_count = swap_amounts.len(), "Estimated swap amounts for token pairs");
 
+        // Detect balance slots for tokens if balance slot detector is available
+        // perf: If we move the extraction to the Hooks DCI, we can extract all the token slots at
+        // once. This way we can leverage the Slot Detector's parallelization and improve
+        // speed. For now, this was included on the EntrypointGenerator for quick
+        // integration and to reduce the number of interface changes.
+        let detected_balance_slots = if let Some(ref balance_slot_detector) =
```
kayibal: Ok I think this step should be optional, but not via configuration. I think we should allow this step to fail / skip it:

- If fail and the hook uses external liquidity, only if the pool manager has enough balance available
- Skip if the hook does not use external liquidity
- Else error

tycho-ethereum/src/entrypoint_tracer/evm_balance_slot_detector.rs
```diff
@@ -0,0 +1,808 @@
...
+        components: Vec<(ComponentId, Vec<Address>)>,
+        holder: Address,
+        block_hash: BlockHash,
+    ) -> HashMap<ComponentId, Result<HashMap<Address, Bytes>, Self::Error>> {
+        info!("Starting balance slot detection for {} components", components.len());
+        let mut all_results = HashMap::new();
+
+        // Process components in chunks to optimize, while avoiding being rate limited
+        for (chunk_idx, chunk) in components
+            .chunks(self.max_concurrent_components)
```
kayibal: As said previously I find that using components here makes the whole thing a lot more complex than actually needed. Most likely if you do this concurrently you will be extracting the same thing twice. Also components can have more than 2 tokens so the control over your concurrency is somewhat limited.

tycho-ethereum/src/entrypoint_tracer/tracer.rs
```diff
@@ -48,7 +49,11 @@ impl EVMEntrypointService {
         Ok(Self { rpc_url: url })
     }
 
-    fn create_access_list_params(
+    pub fn rpc_url(&self) -> &url::Url {
+        &self.rpc_url
+    }
+
+    pub fn create_access_list_params(
```
zizou0x: This one doesn't need to be public

tycho-ethereum/src/entrypoint_tracer/tracer.rs
```diff
@@ -81,7 +86,7 @@ impl EVMEntrypointService {
         }
     }
 
-    fn create_trace_call_params(
+    pub fn create_trace_call_params(
```
zizou0x: We should move it outside of this module if it's also used by the balance slot detector. Also would be better to make it `pub(crate)`

tycho-indexer/src/extractor/dynamic_contract_indexer/entrypoint_generator.rs
```diff
@@ -835,12 +924,61 @@ mod tests {
...
     use tycho_common::{
-        models::{protocol::ProtocolComponent, TxHash},
+        models::{protocol::ProtocolComponent, BlockHash, ComponentId, TxHash},
         Bytes,
     };
 
     use super::*;
 
+    // Mock implementation of BalanceSlotDetector for tests
+    struct MockBalanceSlotDetector {
```
zizou0x: Can't we automock the trait instead?

tycho-ethereum/src/entrypoint_tracer/evm_balance_slot_detector.rs
```diff
@@ -0,0 +1,808 @@
...
+                        "Found storage slot for token"
+                    );
+                    token_slots.insert(token.clone(), slot);
+                }
+                Ok(None) => {
+                    debug!(token = %token, "No suitable slot found for token");
+                }
+                Err(e) => {
+                    error!(token = %token, error = %e, "Failed to extract slot for token");
+                    return Err(e);
```
zizou0x: Should we really error for the whole batch if only one detection fails?

tycho-ethereum/src/entrypoint_tracer/evm_balance_slot_detector.rs
```diff
@@ -0,0 +1,808 @@
...
+            _ => {
+                // Find the slot with minimum difference without allocating a new vector
+                let (best_slot, best_value, best_diff) = slot_values
+                    .into_iter()
+                    .map(|(slot, value)| {
+                        let diff = value.abs_diff(expected_balance);
+                        (slot, value, diff)
+                    })
+                    .min_by_key(|(_, _, diff)| *diff)
+                    .unwrap(); // Safe because we know len > 0
```
zizou0x: If we expect this to be safe, it's better to use `.expect()`, so you can put why you expect this to be safe

tycho-ethereum/src/entrypoint_tracer/evm_balance_slot_detector.rs
```diff
@@ -0,0 +1,808 @@
...
+        self.max_concurrent_components
+    }
+}
+
+impl EVMBalanceSlotDetector {
+    /// Direct access to the underlying tracer for advanced use cases
+    pub fn tracer(&self) -> &EVMEntrypointService {
+        &self.tracer
+    }
+}
```
zizou0x: Where is this used? Only for testing?

tycho-ethereum/src/entrypoint_tracer/evm_balance_slot_detector.rs
```diff
@@ -0,0 +1,808 @@
...
+    }
+
+    /// Send a batched JSON-RPC request and return the responses
+    async fn send_batched_request(
+        &self,
+        batch_request: Value,
+    ) -> Result<Vec<Value>, BalanceSlotError> {
+        let client = reqwest::Client::new();
+        let response = client
+            .post(self.tracer.rpc_url().as_str())
```
zizou0x: If I searched correctly, this is the only place where we use the tracer. If so, we don't need it, we can just store the rpc_url

tycho-ethereum/src/entrypoint_tracer/evm_balance_slot_detector.rs
tvinagre: In this case, whole batch is the component's 2 tokens, so yes I'd fail

tycho-ethereum/src/entrypoint_tracer/evm_balance_slot_detector.rs
tvinagre: I thought this was safer than potentially having 2 different RPC_URLS

tycho-common/src/traits.rs
```diff
@@ -148,30 +148,25 @@ pub trait EntryPointTracer: Sync {
...
     ///
     /// # Returns
-    /// HashMap mapping component_id -> Result containing (token_address -> storage_slot) or error
-    async fn detect_slots_for_components(
+    /// HashMap mapping Token -> Result containing (contract_address -> storage_slot) or error.
+    /// The storage slot is the one that controls the token's holder balance.
+    async fn detect_balance_slots(
         &self,
-        components: Vec<(ComponentId, Vec<Address>)>,
+        tokens: &[Address],
```
kayibal, line 185: In my experience the most flexible has been to use sth like `IntoIterator<&Address>` for these use cases.

tycho-ethereum/src/entrypoint_tracer/balance_slot_detector.rs
```diff
@@ -59,45 +67,96 @@ impl EVMBalanceSlotDetector {
...
+            Ok(requests) => requests,
+            Err(e) => {
+                for token in request_tokens.iter() {
+                    // Failed to create the batched request - return with only the cached values.
+                    cached_tokens
+                        .insert(token.clone(), Err(BalanceSlotError::RequestError(e.to_string())));
+                }
+                return cached_tokens;
+            }
+        };
```
kayibal: I was going to criticize here that we skip all tokens if this fails for just one but I noticed that your function `create_balance_request` seems to be infallible. So this error handling here is not needed at all.

tycho-ethereum/src/entrypoint_tracer/balance_slot_detector.rs
```diff
@@ -59,45 +67,96 @@ impl EVMBalanceSlotDetector {
...
+            .await
+        {
+            Ok(responses) => responses,
+            Err(e) => {
+                for token in request_tokens.iter() {
+                    // Failed to send the batched request - return with only the cached values.
+                    cached_tokens
+                        .insert(*token, Err(BalanceSlotError::RequestError(e.to_string())));
+                }
+                return cached_tokens;
```
kayibal: Ok here things can obviously fail, e.g. network issues, rpc node not having the block yet, etc. We for sure need a retry mechnism and then I wouldn't fail everything but just the requests that actually failed.

tycho-ethereum/src/entrypoint_tracer/balance_slot_detector.rs
```diff
@@ -0,0 +1,853 @@
...
+
+        Ok(Value::Array(batch))
+    }
+
+    /// Send a batched JSON-RPC request and return the responses
+    async fn send_batched_request(
+        &self,
+        batch_request: Value,
+    ) -> Result<Vec<Value>, BalanceSlotError> {
+        let client = reqwest::Client::new();
```
kayibal: I think this is an anti-pattern. You need to persist the client, ideally one per application or module. Probably fine for DCI to have it's own. Else the requests will be really slow (creating this client actually expensice and it has internal connection pool that you want to reuse)

tycho-ethereum/src/entrypoint_tracer/balance_slot_detector.rs
```diff
@@ -0,0 +1,853 @@
...
+        let response_json: serde_json::Value = response
+            .json()
+            .await
+            .map_err(|e| BalanceSlotError::InvalidResponse(e.to_string()))?;
+
+        // Handle batched response array
+        match response_json {
+            Value::Array(responses) => Ok(responses),
+            _ => Err(BalanceSlotError::InvalidResponse(
+                "Expected array response for batched request".into(),
```
kayibal: Might be good to log what the actual response was here. And maybe do some error handling before based on the status code?

Also I think based on the status code we should potentially retry the request

tycho-indexer/src/extractor/dynamic_contract_indexer/entrypoint_generator.rs
tvinagre: The first one "if the pool manager has enough balance available" it's tricky because we only know the balance once we actually do the eth call inside the balance_slot_detector. 
I will change the interface to return the value if we manage to extract it, but will do it once I address the other comments

tycho-ethereum/src/entrypoint_tracer/balance_slot_detector.rs
tvinagre: I've decided to not base on status code and just retry.. My life made me not trust status codes

tycho-common/src/traits.rs
tvinagre: I didn't know IntoIterator - will check!

tycho-indexer/src/extractor/dynamic_contract_indexer/entrypoint_generator.rs
tvinagre: Hmm this would invalidate the cache, so I'd rather do this in a later optimization.

tycho-ethereum/src/entrypoint_tracer/balance_slot_detector.rs
```diff
@@ -0,0 +1,1218 @@
...
+
+        // Send batched request
+        let responses = match self
+            .send_batched_request(requests)
+            .await
+        {
+            Ok(responses) => responses,
+            Err(e) => {
+                // If request fails, mark all as failed.
+                // TODO: Add retry mechanism
```
tvinagre: remove this, as there's now retry to send_batched_request

tycho-ethereum/src/entrypoint_tracer/balance_slot_detector.rs
```diff
@@ -0,0 +1,1218 @@
...
+        }
+
+        info!("Balance slot detection completed. Found results for {} tokens", all_results.len());
+        all_results
+    }
+}
+
+impl EVMBalanceSlotDetector {
+    /// Direct access to the underlying tracer for advanced use cases
+    pub fn tracer(&self) -> &EVMEntrypointService {
```
tvinagre: will remove this, unused

tycho-indexer/src/extractor/dynamic_contract_indexer/entrypoint_generator.rs
```diff
@@ -583,6 +600,58 @@ impl<E: SwapAmountEstimator + Send + Sync> HookEntrypointGenerator
...
+            // If balance slot detector was provided but no slots were successfully detected, fail
+            if token_slots.is_empty() && !tokens.is_empty() {
+                return Err(EntrypointGenerationError::NoDataAvailable(
+                    "Balance slot detection failed for all tokens".to_string(),
+                ));
+            }
+
+            token_slots
+        } else {
+            info!("No balance slot detector available, skipping balance overwrites");
```
tvinagre: Update this log as it's now due to a configuration and not lack of detector

tycho-ethereum/src/entrypoint_tracer/balance_slot_detector.rs
```diff
@@ -0,0 +1,1218 @@
...
+        }
+    }
+}
+
+/// Cache type for balance slot detection results
+type BalanceSlotCache = Arc<RwLock<HashMap<(Address, Address), (Address, Bytes)>>>;
+
+/// EVM-specific implementation of BalanceSlotDetector using debug_traceCall
+pub struct EVMBalanceSlotDetector {
+    tracer: EVMEntrypointService,
```
zizou0x: Where is this used?

It seems to me that: you take a `rpc_url` from the `BalanceSlotDetectorConfig`, then create a `EVMEntrypointService` then only use it to get the `rpc_url`.

Why not just storing the `rpc_url` directly?

tycho-ethereum/src/entrypoint_tracer/balance_slot_detector.rs
```diff
@@ -0,0 +1,1218 @@
...
+    }
+
+    /// Send a batched JSON-RPC request and return the responses
+    async fn send_batched_request(
+        &self,
+        batch_request: Value,
+    ) -> Result<Vec<Value>, BalanceSlotError> {
+        self.send_batched_request_with_retry(batch_request)
+            .await
+    }
```
zizou0x: Very shallow interface

tycho-ethereum/src/entrypoint_tracer/tracer.rs
tvinagre: not going to move this right now, as we'll refactor all the client / RPC modules

tycho-ethereum/src/entrypoint_tracer/balance_slot_detector.rs
tvinagre: Well noted! This was being used before but now it's not after the refactor - I forgot to remove. Thanks!

tycho-indexer/src/extractor/dynamic_contract_indexer/entrypoint_generator.rs
```diff
@@ -429,28 +437,26 @@ impl DefaultSwapAmountEstimator {
...
-                                    .to_bytes_be()
-                                    .1,
-                            ), // 0.05%
-                            Bytes::from(
-                                ((&balance * BigInt::from(10)) / &ten_thousand)
+                                ((&balance * BigInt::from(2)) / &hundred)
                                     .to_bytes_be()
                                     .1,
-                            ), // 0.1%
+                            ), // 20%
```
zizou0x:
```suggestion
                            ), // 2%
```

PR #683, opened 2025-09-09
feat: add component pausing in hook_dci 

zizou0x:
This PR adds the logic to pause protocol components when anything goes wrong during hook dci process.

Also fixes a bug that was occuring when inserting "paused" attribute in DCI. We were replacing the current update in case of conflict. We are now correctly merging them.


tycho-indexer/src/extractor/dynamic_contract_indexer/hook_dci.rs
```diff
@@ -280,28 +284,65 @@ where
...
+                .get(tx_hash)
+                .cloned()
+                .unwrap_or_else(|| {
+                    error!("No transaction found for hash {tx_hash} for component {component_id}");
+                    Transaction::new(tx_hash.clone(), block_hash.clone(), Bytes::zero(20), None, 0)
+                }),
+            None => {
+                error!("No transactions available for component {component_id}");
+                Transaction::new(Bytes::zero(32), block_hash.clone(), Bytes::zero(20), None, 0)
+            }
```
zizou0x: I'm really unsure about this, can you foresee any issue? Hopefully it should never happen but not sure how this will evolve in the future. 

The alternatives are: 

- Directly error (which probably means panic)
- Ignore pausing the component

We can also try to insert this into a random transaction of the block.

tvinagre: If this should never happen, I'd really like to avoid creating a transaction right now. This needs proper planning to ensure it won't break things, which I think it will, as we'll have repeated TXs with the same hash in different blocks.

Right now everything is designed to have a matching transaction so I think it's fine to ensure that we find the TX in the block - if we don't, there's a bug and we need to fix.

tycho-indexer/src/extractor/dynamic_contract_indexer/dci.rs
```diff
@@ -452,39 +453,13 @@ where
...
-                            .txs_with_update
-                            .push(tx_with_changes);
-                    }
-                }
+                insert_state_attribute_update(
+                    &mut block_changes.txs_with_update,
+                    component_id,
+                    tx,
+                    &"paused".to_string(),
+                    &vec![PausingReason::TracingError.get_reason_index()].into(),
```
kayibal: This looks weird to me :D should we just use into?

tycho-indexer/src/extractor/dynamic_contract_indexer/hook_dci.rs
```diff
@@ -245,12 +248,13 @@ where
     }
 
     /// Handles component failures by updating the state and logging errors
-    #[instrument(skip(self, block), fields(component_id = %component_id, block_number = block.number))]
+    #[instrument(skip(self, block_changes), fields(component_id = %component_id, block_number = block_changes.block.number))]
```
kayibal: can you also skip the tx here please?

tycho-indexer/src/extractor/dynamic_contract_indexer/hook_dci.rs
```diff
@@ -280,28 +284,65 @@ where
...
+        // This indicates the component processing has been paused due to failures
+        insert_state_attribute_update(
+            &mut block_changes.txs_with_update,
+            &component_id,
+            tx,
+            &"paused".to_string(),
+            &vec![PausingReason::MetadataError.get_reason_index()].into(),
+        )?;
+
+        debug!("Component state updated to Failed with paused attribute");
```
kayibal: Can you add some error context here?

tycho-indexer/src/extractor/dynamic_contract_indexer/hook_dci.rs
```diff
@@ -333,7 +374,14 @@ where
                     "Component has metadata errors"
```
kayibal, line 353: Why do we not log the errors here? Is this already logged previously?

tycho-indexer/src/extractor/dynamic_contract_indexer/hook_dci.rs
```diff
@@ -745,6 +793,12 @@ where
             "Processing components"
         );
 
+        let tx_map = block_changes
+            .txs_with_update
+            .iter()
+            .map(|tx_with_changes| (tx_with_changes.tx.hash.clone(), tx_with_changes.tx.clone()))
+            .collect::<HashMap<_, _>>();
```
kayibal: This piece of code is repeated it seems?

tycho-indexer/src/extractor/dynamic_contract_indexer/hook_dci.rs
```diff
@@ -793,11 +847,48 @@ where
                             "Hook orchestrator failed (full processing)"
                         );
 
+                        // Get the tx that updated the component, default to a dummy tx if not
```
kayibal: What does this mean? How does a tx qualify for "it updated the component"?

tycho-indexer/src/extractor/dynamic_contract_indexer/hook_dci.rs
```diff
@@ -793,11 +847,48 @@ where
...
+                        // found.
+                        let tx_hash = component_metadata_map
```
kayibal: From the code above it seems this map only has a single entry. Is that true?

I find this piece of ode quite cryptic (maybe also the variable name):
```
                let component_metadata_map: HashMap<String, _> = [component]
                    .iter()
                    .filter_map(|comp| {
                        metadata_by_component_id
                            .get(&comp.id)
                            .map(|meta| (comp.id.clone(), meta.clone()))
                    })
                    .collect();
```

How is this map different from `metadata_by_component_id`?

tycho-indexer/src/extractor/dynamic_contract_indexer/hook_dci.rs
```diff
@@ -793,11 +847,48 @@ where
...
+                                .unwrap_or_else(|| {
+                                    error!("No tx hash found for component {}", component.id);
+                                    Transaction::new(
+                                        Bytes::zero(32),
+                                        block_changes.block.hash.clone(),
+                                        Bytes::zero(20),
+                                        None,
+                                        0,
+                                    )
+                                }),
```
kayibal: Isn't this logic duplicated?

tycho-indexer/src/extractor/dynamic_contract_indexer/hook_dci.rs
```diff
@@ -793,11 +847,48 @@ where
...
+                                .get(&tx_hash)
+                                .cloned()
+                                .unwrap_or_else(|| {
+                                    error!("No tx hash found for component {}", component.id);
+                                    Transaction::new(
+                                        tx_hash.clone(),
+                                        block_changes.block.hash.clone(),
+                                        Bytes::zero(20),
+                                        None,
+                                        0,
```
kayibal: These seems like a terrible idea? Why not just use the. actual transaction here?

tycho-indexer/src/extractor/dynamic_contract_indexer/hook_dci.rs
```diff
@@ -807,11 +898,46 @@ where
...
+                        .unwrap_or_else(|| {
+                            error!("No tx hash found for component {}", component.id);
+                            Transaction::new(
+                                Bytes::zero(32),
+                                block_changes.block.hash.clone(),
+                                Bytes::zero(20),
+                                None,
+                                0,
+                            )
+                        }),
```
kayibal: Ok the same code duplicated again.

tycho-indexer/src/extractor/dynamic_contract_indexer/mod.rs
```diff
@@ -9,3 +9,20 @@ mod hook_permissions_detector;
...
+
+impl PausingReason {
+    fn get_reason_index(&self) -> u8 {
+        match self {
+            PausingReason::Substreams => 1,
+            PausingReason::TracingError => 2,
+            PausingReason::MetadataError => 3,
+        }
+    }
+}
```
kayibal: Can you just use impl `From<PausingReason> for u8 {}` I wonder if there is another way to do this...

tycho-indexer/src/extractor/dynamic_contract_indexer/hook_dci.rs
zizou0x: It seems that it's logged inside `self.handle_component_failure`

tycho-indexer/src/extractor/dynamic_contract_indexer/hook_dci.rs
zizou0x: Unfortunately this is a hack that we had to do a long time ago to quickly implement the euler orchestrator without breaking things in the interfaces. This needs to be redesigned. 

It's different from  `metadata_by_component_id` because this one only has one component.

It's terrible but we have to do a full code refactor here. I'll leave a TODO for this

tycho-indexer/src/extractor/dynamic_contract_indexer/hook_dci.rs
zizou0x: I agree, I thought we really wanted to avoid returning error at all cost but I was wrong. 

I've changed this to avoid this transaction hack.

tycho-indexer/src/extractor/dynamic_contract_indexer/hook_dci.rs
```diff
@@ -280,28 +285,43 @@ where
...
             .component_states
-            .insert_pending(block.clone(), component_id.clone(), new_state)
+            .insert_pending(block_changes.block.clone(), component_id.clone(), new_state)
             .map_err(|e| {
                 error!("Failed to update component state: {e}");
                 ExtractionError::Unknown(format!("Failed to update component state: {e}"))
             })?;
 
-        debug!("Component state updated to Failed");
+        // Mark component as paused by setting the "paused" attribute to [3]
```
tvinagre: minor comment but I don't see why adding the number of the Enum here

tycho-indexer/src/extractor/dynamic_contract_indexer/hook_dci.rs
```diff
@@ -280,28 +285,43 @@ where
...
+        // This indicates the component processing has been paused due to failures
```
tvinagre:
```suggestion
        // This indicates the component processing has been paused due to Metadata failures
```

tycho-indexer/src/extractor/dynamic_contract_indexer/hook_dci.rs
```diff
@@ -762,12 +808,16 @@ where
                     "Found hook orchestrator, processing components needing full processing"
                 );
 
-                let component_metadata_map: HashMap<String, _> = [component]
+                // TODO: This map currently contains only one entry. This will change when we
```
tvinagre: This map is ambiguous - I'm assuming you refer to component_metadata_map

tvinagre (approved):

PR #686, opened 2025-09-09
fix: handle buffer overflow in client

kayibal:
Automatically unsubscribes from a protocol if client can't keep up with message flow. This will force the client to reconnect which clears the buffer and thus gives the client a chance to recover. This behaviour is better compared to dropping the message with a warning which most certainly leads to corrupted state.

### DIFF
--- a/tycho-client/src/deltas.rs
+++ b/tycho-client/src/deltas.rs
@@ -403,6 +403,30 @@ impl WsDeltasClient {
         })
     }
 
+    // Construct a new client with custom buffer sizes (for testing)
+    #[cfg(test)]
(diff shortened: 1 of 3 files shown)


tycho-client/src/deltas.rs
```diff
@@ -579,6 +595,7 @@ impl WsDeltasClient {
         subscription_id: Uuid,
         ready_tx: oneshot::Sender<()>,
     ) -> Result<(), DeltasError> {
+        debug!(?subscription_id, "Unsubscribing");
```
tvinagre, line 598: Can we make this info?
Just to ensure this always appears in the log, specially when it's sent to us

PR #690, opened 2025-09-12
feat: Use custom contract to fetch Euler limits with the correct values

tvinagre:
**Summary:**

  Updated EulerMetadataGenerator to use a new lens contract for extracting Euler protocol limits data. The implementation uses eth_call with state overrides to deploy lens contract bytecode at runtime and manipulate storage slot 0 to inject the hook address. This allows us to get more realistic limits

  **Key Changes:**
  - Added lens contract integration: Uses EulerLimitsLens.evm.runtime bytecode and ABI from assets
  - State override mechanism: Deploys lens contract at deterministic address 0x0000000000000000000000000000000000001337 during eth_call execution
  - Storage slot manipulation: Injects hook address into storage slot 0 using vm.store() pattern for lens contract to read
  - Maintained function interface: Uses getLimits(address,address) with function selector 0xaaed87a3
  - Fixed entry point targeting: Entry points correctly reference hooks contract address for tracing (not lens address)

  **Technical Implementation:**
  - Lens contract bytecode loaded from ../assets/EulerLimitsLens.evm.runtime
  - State overrides deploy bytecode and set storage["0x0"] = hookAddress
  - RPC calls use "state" field for storage overrides (not "storage")
  - Entry point targets hooks address while lens contract handles limits calculation

  Testing:
  - Added test_euler_metadata_with_lens_contract() for RPC integration testing with real hook 0x55dcf9455EEe8Fd3f5EEd17606291272cDe428a8
  - Added test_entry_point_target_uses_hooks_address() to validate entry point targeting
  - All existing tests maintained compatibility

### DIFF
--- a/tycho-indexer/src/extractor/dynamic_contract_indexer/euler/metadata_generator.rs
+++ b/tycho-indexer/src/extractor/dynamic_contract_indexer/euler/metadata_generator.rs
@@ -368,12 +368,24 @@ mod tests {
             "euler_limits_0x000000000000000000000000000000000000beef_0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48_to_0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2".to_string()
         );
         assert_eq!(requests[1].transport.routing_key(), "rpc_default".to_string());
-        // Note: The deduplication_id will now include the state override parameter
-        // We just check that it starts with the expected prefix
+
+        // The deduplication_id will now include the state & code override parameters
         assert!(requests[1]
             .transport
             .deduplication_id()
             .starts_with("eth_call_"));
+
+        dbg!(&requests[1].transport.deduplication_id());
+
+        assert!(requests[1]
+            .transport
+            .deduplication_id()
+            .contains("code"));
+
+        assert!(requests[1]
+            .transport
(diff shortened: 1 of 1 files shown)


tycho-indexer/src/extractor/dynamic_contract_indexer/euler/metadata_generator.rs
```diff
@@ -332,13 +365,15 @@ mod tests {
...
-        assert_eq!(
-            requests[1].transport.deduplication_id(),
-            "eth_call_[{\"data\":\"0xaaed87a3000000000000000000000000a0b86991c6218b36c1d19d4a2e9eb0ce3606eb48000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc2\",\"to\":\"0xbeef\"},\"0x3039\"]".to_string()
-        );
+        // Note: The deduplication_id will now include the state override parameter
+        // We just check that it starts with the expected prefix
+        assert!(requests[1]
+            .transport
+            .deduplication_id()
+            .starts_with("eth_call_"));
```
zizou0x, line 376: I think this assertion is too flexible, balances calls are also starting with eth_call

zizou0x (approved):

PR #693, opened 2025-09-16
feat: Add retry on Metadata RPC calls

tvinagre:
Adds retry on failed Metadata RPC calls.
Doesn't differentiate the error by RPC error code to be in the safe spot, since some RPCs don't implement those correctly.
**Note**: This implementation is temporary as we aim to have a default implementation for all RPC requests in this codebase, which will share the same client for improved performance

### DIFF
--- a/tycho-indexer/src/extractor/dynamic_contract_indexer/rpc_metadata_provider.rs
+++ b/tycho-indexer/src/extractor/dynamic_contract_indexer/rpc_metadata_provider.rs
@@ -6,6 +6,7 @@ use std::{
 
 use reqwest::Client;
 use serde_json::{json, Value};
+use thiserror::Error;
 use tonic::async_trait;
 use tracing::{debug, error, warn};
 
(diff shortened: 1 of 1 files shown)


tycho-indexer/src/extractor/dynamic_contract_indexer/rpc_metadata_provider.rs
```diff
@@ -217,6 +220,190 @@ impl RequestProvider for RPCMetadataProvider {
...
+                                endpoint = endpoint,
+                                "RPC batch request returned server error, will retry"
+                            );
+                            // Create a synthetic error to represent server failure
+                            let synthetic_error = response
+                                .error_for_status_ref()
+                                .unwrap_err();
+                            last_error = Some(synthetic_error);
+                        } else {
+                            // Client error (4xx) - not retryable, but still return the response
```
zizou0x: What is the idea behind this? Why not returning an error?

zizou0x (approved):

PR #695, opened 2025-09-17
feat: Expose retry configurations on TychoStreamBuilder

kayibal:
Allows users more fine-grained control over their retry preferences and how the stream should deal with delays and intermediary failures.


tycho-client/src/cli.rs
```diff
@@ -270,6 +270,7 @@ async fn run(exchanges: Vec<(String, Option<String>)>, args: CliArgs) -> Result<
             true,
             filter,
             3,
+            Duration::from_secs(args.block_time),
```
zizou0x: So this is the retry cooldown on the synchronizer? 

Why so much?

kayibal: I think the default configuration for a synchronizer should survive a tycho indexer restart which takes about 1-2m. I actually forgot to update this location and it is a bit too short.

PR #696, opened 2025-09-18
feat: handle end of stream gracefully

adrianbenavides:
Currently, any error happening when running tycho-indexer is returned as a panic. This is problematic when using the binary from other parts (i.e. the tycho-protocol-sdk).

image

-------

To return proper errors, the tokio entry-point has been moved up to the main function so we can await async calls and handle their errors.

Also, in the substreams logic, adds a new `BlockResponse` variant to handle end-of-stream as a valid state, instead of as an error.

image

### DIFF
--- a/tycho-indexer/src/main.rs
+++ b/tycho-indexer/src/main.rs
@@ -10,6 +10,7 @@ use std::{
 };
 
 use actix_web::{dev::ServerHandle, web, App, HttpResponse, HttpServer, Responder};
+use anyhow::anyhow;
 use chrono::{NaiveDateTime, Utc};
 use clap::Parser;
 use futures03::future::select_all;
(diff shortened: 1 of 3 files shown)


tamaralipows (approved):
LGTM, let's just wait for someone from cairoline to review since they know better

tycho-indexer/src/main.rs
```diff
@@ -81,19 +81,32 @@ impl ExtractorConfigs {
...
+    match cli.command() {
+        Command::Run(run_args) => run_spkg(global_args, run_args)
+            .await
+            .map_err(|e| e.into()),
+        Command::AnalyzeTokens(analyze_args) => run_tycho_ethereum(global_args, analyze_args).await,
+        Command::Rpc => run_rpc(global_args)
+            .await
+            .map_err(|e| e.into()),
+        _ => {
+            process::exit(1);
```
zizou0x: Can't we leave the current design as it was here (with the single sync command handling)? I prefer to have a single match statement on `cli.command()` with no default, it makes it more future proof.

Do you think something like this would work?

```
fn main() -> Result<(), anyhow::Error> {
    let cli: Cli = Cli::parse();
    let global_args = cli.args();

    let res: Result<(), anyhow::Error> = match cli.command() {
        Command::Run(run_args) => run_spkg(global_args, run_args).map_err(|e| e.into()),
        Command::Index(indexer_args) => {
            run_indexer(global_args, indexer_args).map_err(|e| e.into())
        }
        Command::AnalyzeTokens(analyze_args) => run_tycho_ethereum(global_args, analyze_args),
        Command::Rpc => run_rpc(global_args).map_err(|e| e.into()),
    };
    res
}
```

adrianbenavides: I think I tried that and got some errors, but I'm not sure now. Will check again!

adrianbenavides: I was able to do that, should be good now 👍

PR #697, opened 2025-09-18
feat: Retry on transient rpc error.

kayibal:
This avoids crashing if one rpc requests fails or gets rate limited. Once we correctly return a retry-after header from the server, the client will respect it and requests will wait until then.

### DIFF
--- a/tycho-client/src/rpc.rs
+++ b/tycho-client/src/rpc.rs
@@ -3,15 +3,25 @@
 //! The objective of this module is to provide swift and simplified access to the Remote Procedure
 //! Call (RPC) endpoints of Tycho. These endpoints are chiefly responsible for facilitating data
 //! queries, especially querying snapshots of data.
-use std::{collections::HashMap, sync::Arc};
+use std::{
+    collections::HashMap,
+    sync::Arc,
+    time::{Duration, SystemTime},
+};
 
 use async_trait::async_trait;
+use backoff::{exponential::ExponentialBackoffBuilder, ExponentialBackoff};
 use futures03::future::try_join_all;
 #[cfg(test)]
 use mockall::automock;
-use reqwest::{header, Client, ClientBuilder, Url};
+use reqwest::{header, Client, ClientBuilder, Response, StatusCode, Url};
+use serde::Serialize;
 use thiserror::Error;
-use tokio::sync::Semaphore;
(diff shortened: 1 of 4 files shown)


tycho-client/src/rpc.rs
```diff
@@ -603,11 +622,140 @@ impl HttpRPCClient {
...
+
+    async fn error_for_response(
+        &self,
+        response: reqwest::Response,
+    ) -> Result<reqwest::Response, RPCError> {
+        match response.status() {
+            StatusCode::TOO_MANY_REQUESTS => {
+                let retry_after_raw = response
+                    .headers()
+                    .get(reqwest::header::RETRY_AFTER)
```
zizou0x, line 669: Did you check if we're emitting this one?

tycho-client/src/rpc.rs
```diff
@@ -603,11 +622,140 @@ impl HttpRPCClient {
...
+            .with_initial_interval(Duration::from_millis(1))
+            .with_multiplier(1.1)
+            .with_max_interval(Duration::from_millis(5))
+            .with_max_elapsed_time(Some(Duration::from_millis(50)))
+            .build();
+        self.server_restart_duration = Duration::from_millis(50);
+        self
+    }
+
+    async fn error_for_response(
```
zizou0x, line 661: Even if they are internals, would be nice to have some docs for all these new funtions

tycho-client/src/rpc.rs
```diff
@@ -603,11 +622,140 @@ impl HttpRPCClient {
...
+    async fn wait_until_retry_after(&self) {
+        if let Some(&until) = self.retry_after.read().await.as_ref() {
+            let now = SystemTime::now();
+            if until > now {
+                if let Ok(duration) = until.duration_since(now) {
+                    sleep(duration).await
+                }
+            }
+        }
+    }
```
zizou0x, line 729: Oh so `retry_after` is a timestamp and not a duration right?

tycho-client/src/rpc.rs
kayibal: Afaik we are not but it should be easy to add.

tycho-client/src/rpc.rs
kayibal: Yes that is correct

PR #698, opened 2025-09-18
feat: Add AllowanceSlotDetector

dianacarvalho1:
Also had to upgrade alloy, contracts, ethrpc, ethcontract

### DIFF
--- a/tycho-ethereum/src/entrypoint_tracer/allowance_slot_detector.rs
+++ b/tycho-ethereum/src/entrypoint_tracer/allowance_slot_detector.rs
@@ -325,7 +325,6 @@ impl EVMAllowanceSlotDetector {
         }))
     }
 
-    /// Send a single request without retry
     async fn send_single_request(
         &self,
         batch_request: &Value,
@@ -441,7 +440,6 @@ impl EVMAllowanceSlotDetector {
         self.find_best_slot_by_value_comparison(slot_values, allowance)
     }
 
-    /// Extract allowance from eth_call response
     fn extract_allowance_from_call_response(
         &self,
         response: &Value,
@@ -467,7 +465,6 @@ impl EVMAllowanceSlotDetector {
             .map_err(|e| AllowanceSlotError::AllowanceExtractionError(e.to_string()))
     }
 
-    /// Extract accessed slots with their values from debug_traceCall response
     fn extract_slot_values_from_trace_response(
         &self,
         response: &Value,
@@ -526,7 +523,6 @@ impl EVMAllowanceSlotDetector {
         Ok(slot_values)
     }
 
-    /// Find the best slot by comparing storage values to the expected allowance
     fn find_best_slot_by_value_comparison(
         &self,
         slot_values: SlotValues,
@@ -572,7 +568,6 @@ impl EVMAllowanceSlotDetector {
         }
     }
 
-    /// Validates if the detected storage slots are correct.
     async fn validate_best_slots(
         &self,
         token_slots: TokenSlotResults,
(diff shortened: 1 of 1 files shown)


tycho-ethereum/src/entrypoint_tracer/allowance_slot_detector.rs
```diff
@@ -0,0 +1,904 @@
...
+                hex_str.len()
+            )));
+        }
+
+        U256::from_str_radix(hex_str, 16)
+            .map_err(|e| AllowanceSlotError::AllowanceExtractionError(e.to_string()))
+    }
+
+    /// Extract accessed slots with their values from debug_traceCall response
+    fn extract_slot_values_from_trace_response(
```
tamaralipows: Oh interesting. So we basically make a dummy request and check which slots are accessed? That's not really brute forcing, that's pretty smart? I guess the brute forcing was something else entirely

tycho-ethereum/src/entrypoint_tracer/allowance_slot_detector.rs
```diff
@@ -0,0 +1,904 @@
...
+        let allowance = self.extract_allowance_from_call_response(eth_call_resp)?;
+
+        // Extract slot values from debug_traceCall response
+        let slot_values = self.extract_slot_values_from_trace_response(debug_resp)?;
+
+        // Find the best slot by comparing values to the expected allowance
+        self.find_best_slot_by_value_comparison(slot_values, allowance)
+    }
+
+    /// Extract allowance from eth_call response
```
tamaralipows: I don't think these comments add much relevant information... they are just repeating the fn name

tycho-ethereum/src/entrypoint_tracer/allowance_slot_detector.rs
dianacarvalho1: yeees exactly!

tycho-ethereum/src/entrypoint_tracer/allowance_slot_detector.rs
dianacarvalho1: I removed the most irrelevant ones, but there are others that I do think add a bit more info

tycho-ethereum/src/entrypoint_tracer/allowance_slot_detector.rs
```diff
@@ -0,0 +1,898 @@
...
+    initial_backoff_ms: u64,
+    max_backoff_ms: u64,
+}
+
+impl EVMAllowanceSlotDetector {
+    pub fn new(config: AllowanceSlotDetectorConfig) -> Result<Self, AllowanceSlotError> {
+        let rpc_url = url::Url::parse(&config.rpc_url)
+            .map_err(|_| AllowanceSlotError::SetupError("Invalid URL".to_string()))?;
+
+        let http_client = reqwest::Client::builder()
```
tvinagre, line 111: Both here and on the Balance detector we could benefit of having a shared client. It will require some refactor so it's not necessary we do this now. I'll add a task for it under Maintenance

tycho-ethereum/src/entrypoint_tracer/allowance_slot_detector.rs
```diff
@@ -0,0 +1,898 @@
...
+            all_results.extend(chunk_results);
+        }
+
+        info!("Allowance slot detection completed. Found results for {} tokens", all_results.len());
+        all_results
+    }
+}
+
+/// Encode allowance(owner, spender) calldata
+pub fn encode_allowance_calldata(owner: &Address, spender: &Address) -> Bytes {
```
tvinagre, line 815: This is the only custom part of the code, so would be wayy cleaner if we just reused the balance slot code and just allow the caller to set the calldata

PR #700, opened 2025-09-22
feat: enhance RpcCache with memory size weighing

zizou0x:
This PR adds a rough memory size limit to contract_state RPC cache. Also contains a small improvement on the client side to maximize cache hits.

### DIFF
--- a/tycho-indexer/src/services/cache.rs
+++ b/tycho-indexer/src/services/cache.rs
@@ -11,12 +11,14 @@ pub struct RpcCache<R, V> {
     cache: Arc<Cache<R, Arc<tokio::sync::Mutex<Option<V>>>>>,
 }
 
+type Weigher<R, V> =
+    Box<dyn Fn(&R, &Arc<tokio::sync::Mutex<Option<V>>>) -> u32 + Send + Sync + 'static>;
+
 pub struct RpcCacheBuilder<R, V> {
     name: String,
     capacity: u64,
     ttl: u64,
-    weigher:
-        Option<Box<dyn Fn(&R, &Arc<tokio::sync::Mutex<Option<V>>>) -> u32 + Send + Sync + 'static>>,
+    weigher: Option<Weigher<R, V>>,
 }
 
 impl<R, V> RpcCacheBuilder<R, V>


tycho-common/src/dto.rs
```diff
@@ -776,6 +777,53 @@ impl StateRequestResponse {
...
+        size += std::mem::size_of::<PaginationResponse>();
+
+        // Account data size
+        for account in &self.accounts {
+            // Base account struct overhead (rough estimate for all fixed fields)
+            size += 200; // Conservative estimate for struct overhead + enum + fixed Bytes
+
+            // Variable-length byte fields
+            size += account.address.len();
+            size += account.title.capacity(); // String allocates capacity, not just len
```
kayibal, line 795: Shouldn't we use capacity everywhere then?

tycho-common/src/dto.rs
```diff
@@ -776,6 +777,53 @@ impl StateRequestResponse {
...
+            size += account.balance_modify_tx.len();
+            size += account.code_modify_tx.len();
+
+            // Creation tx (optional)
+            if let Some(ref creation_tx) = account.creation_tx {
+                size += creation_tx.len();
+            }
+
+            // Storage slots HashMap - this is likely the largest contributor
+            size += account.slots.capacity() * 64; // Rough HashMap overhead per slot
```
kayibal: 64 is ethereum specific...

tycho-common/src/dto.rs
```diff
@@ -776,6 +777,53 @@ impl StateRequestResponse {
...
+            for (key, value) in &account.slots {
+                size += key.len();
+                size += value.len();
+            }
```
kayibal, line 813: Why is the account size duplicated here?

tycho-common/src/dto.rs
```diff
@@ -776,6 +777,53 @@ impl StateRequestResponse {
...
+
+            // Token balances HashMap
+            size += account.token_balances.capacity() * 64; // HashMap overhead
+            for (key, value) in &account.token_balances {
+                size += key.len();
+                size += value.len();
+            }
```
kayibal: I think it is either the first line here, or the for loop, not both...

tycho-common/src/dto.rs
zizou0x: No, this is the two `Bytes` structs (they are 4 usize big, so 32)

tycho-common/src/dto.rs
zizou0x: This is the heap allocation for the `Bytes`

PR #704, opened 2025-09-23
feat(client): handle filtering paused components

zizou0x:


tycho-client/src/feed/component_tracker.rs
```diff
@@ -360,14 +360,38 @@ where
...
+            if s.deleted_attributes.contains("paused") {
+                to_add.push(id.clone());
+            }
+            // If the component is paused, we need to remove it
+            else if s
+                .updated_attributes
+                .contains_key("paused")
+            {
+                to_remove.push(id.clone());
+            }
```
kayibal, line 378: I would prefer if we used the actual value of the attribute. So if not present or =0 -> unpaused if present and > 0 then paused.

tycho-common/src/dto.rs
```diff
@@ -1290,6 +1290,12 @@ impl From<models::protocol::ProtocolComponentState> for ResponseProtocolState {
     }
 }
 
+impl ResponseProtocolState {
+    pub fn is_paused(&self) -> bool {
+        self.attributes.contains_key("paused")
```
kayibal, line 1295: Again would prefer to rely on the attribute value here.

tycho-client/src/feed/component_tracker.rs
zizou0x: According to the plans, on the indexer/sdk side we are deleting the attribute to unpause it, we're never setting this to 0. See [here](https://github.com/propeller-heads/tycho-protocol-sdk/pull/276/files)

Do we still want this, just in case?

PR #707, opened 2025-09-24
fix: make `TracingResult` deserialization backward compatible

zizou0x:
We added a custom deserialization that defaults offset to 0 when it's missing. This is fine because it's not used by tycho-simulation.

This is needed to smoothly update prod with the new release without having to sync with all clients. They can now update their tycho-client version before and it will automatically support both old and new release.

We can remove this once everything is more stable.

### DIFF
--- a/tycho-common/src/dto.rs
+++ b/tycho-common/src/dto.rs
@@ -1824,7 +1824,7 @@ fn deserialize_retriggers_from_value(
                             let storage_key = StoreKey::deserialize(&pair[1]).map_err(|e| {
                                 format!("Failed to deserialize old format storage key: {}", e)
                             })?;
-                            AddressStorageLocation::new(storage_key, 0)
+                            AddressStorageLocation::new(storage_key, 12)
                         }
                         Value::Object(_) => {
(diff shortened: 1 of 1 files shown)


tycho-common/src/dto.rs
```diff
@@ -1802,14 +1802,91 @@ impl From<models::blockchain::AddressStorageLocation> for AddressStorageLocation
...
+                        .map_err(|e| format!("Failed to deserialize key: {}", e))?;
+
+                    // Handle both old format (string) and new format (AddressStorageLocation)
+                    let addr_storage = match &pair[1] {
+                        Value::String(_) => {
+                            // Old format: just a string key with offset defaulted to 0
+                            let storage_key = StoreKey::deserialize(&pair[1]).map_err(|e| {
+                                format!("Failed to deserialize old format storage key: {}", e)
+                            })?;
+                            AddressStorageLocation::new(storage_key, 0)
```
kayibal: I would suggest to make the default offset 12, even though it's ethereum specific.

kayibal (approved):

tvinagre (approved):

PR #710, opened 2025-09-25
feat: decouple `ReorgBuffer` db commit tracking from finality

ElusAegis:
- introduce `committed_upto_block_height` on `BlockAggregatedChanges` (and propagate through DTO fixtures/tests/comments) so we can describe the last block guaranteed in storage independently of finality.
  - `ReorgBuffer` drains by committed height and return a new `CommitStatus`, replacing the former
  finality-based helpers
  - update `PendingDeltas`, RPC handling, websocket, and extractor to consume the new
  commit semantics while still feeding the finalized height as the committed height for now, until the futher rework on the Extractors to introduce batch commit logic.


tycho-common/src/models/blockchain.rs
```diff
@@ -107,10 +109,15 @@ impl BlockAggregatedChanges {
         component_tvl: HashMap<String, f64>,
         dci_update: DCIUpdate,
     ) -> Self {
+        if committed_upto_block_height > finalized_block_height {
+            panic!("committed_upto_block_height cannot be greater than finalized_block_height");
+        }
```
zizou0x: As we're maintaining a service for the indexer, we're trying to avoid panics (also through `.unwrap()`) as much as possible. 

In this particular case, the `new()` pattern is usually safe so I wouldn't even change this for a `Result` here.

Instead, you can handle this specifically in the extractor logic where we create this `BlockAggregatedChanges`. For example we can add this inside the `aggregate_updates` logic -> return an error if we're in that case and add a unit test.

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -749,7 +749,7 @@ where
                 .map_err(ExtractionError::Storage)?;
 
             let mut msgs = reorg_buffer
-                .drain_new_finalized_blocks(inp.final_block_height)
+                .drain_new_committed_blocks(inp.final_block_height)
```
zizou0x: I find this naming missleading, from it it seems like these blocks are already committed at this point while they are not.

I'd suggest something more like `drain_new_committable_blocks` or even it could just be something more generic like `drain_block_older_than`, what do you think?

tycho-common/src/models/blockchain.rs
ElusAegis: Makes sense, thank you for pointing this out. And also, from the looks of it, `new` is never actually used, so this check would not yeild anything anyway.

tycho-indexer/src/extractor/protocol_extractor.rs
ElusAegis: yes, that is a great point, thank you! The awkward part is that then in the context of the `PendingDeltas` it now becomes the already committed blocks, so we use the same function in two quite different contexts. I will have a look around this

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -1189,15 +1197,34 @@ where
...
-            .expect("Couldn't find most recent block in buffer during revert");
+            .ok_or(ExtractionError::ReorgBufferError("Reorg buffer is empty after purge".into()))?;
+
+        // The latest finalized block height is the one of the last block in the reverted_state
+        // (i.e. the most recent block that is reverted)
+        let finalized_block_height = reverted_state
+            .last()
+            .ok_or(ExtractionError::ReorgBufferError("Reorg buffer is empty after purge".into()))?
+            .block_update
+            .finalized_block_height;
```
zizou0x, line 1208: True, nice

zizou0x (approved):
LGTM, thanks 🚀

PR #711, opened 2025-09-26
feat: Client stability improvements

kayibal:
- All protocol's states are kept, previously ended and stale states would be removed
- We do not crash if a protocol ends or goes stale
- Only if all protocols are stale or ended we end the stream
- Final errors are propagated through the Receiver to the user. This makes it easier to understand what caused the crash
- Advanced synchronizer which happens on retries are handled gracefully now and do not cause a crash
- If a protocols recovers quickly it does not retrieve a full snapshot (e.g. short websocket disconnect)

### DIFF
--- a/tycho-client/src/feed/mod.rs
+++ b/tycho-client/src/feed/mod.rs
@@ -197,7 +197,6 @@ impl Display for SynchronizerState {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         match self {
             SynchronizerState::Started => write!(f, "Started"),
-
             SynchronizerState::Ready(b) => write!(f, "Started({})", b.number),
             SynchronizerState::Delayed(b) => write!(f, "Delayed({})", b.number),
             SynchronizerState::Stale(b) => write!(f, "Stale({})", b.number),
@@ -286,8 +285,11 @@ impl SynchronizerStream {
 
     /// Standard way to advance a well-behaved state synchronizer.
     ///
-    /// Will wait for a new block on the synchronizer within a timeout. And modify its state based
-    /// on the outcome.
+    /// Will wait for a new block on the synchronizer within a timeout. And modify its
+    /// state based on the outcome.
+    ///
+    /// ## Note
+    /// This method assumes that the current state is `Ready`.
     async fn try_recv_next_expected(
         &mut self,
         max_wait: std::time::Duration,
@@ -319,34 +321,10 @@ impl SynchronizerStream {
             Err(_) => {
                 // trying to advance a block timed out
                 debug!(%extractor_id, %previous_block, "No block received within time limit.");
-
-                // TODO: as per prev condition check, the state will always be always ready here -
-                // simplify
-                match &self.state {
-                    SynchronizerState::Ready(_) => {
-                        // First timeout: always transition to Delayed
-                        self.state = SynchronizerState::Delayed(previous_block.clone());
-                        self.modify_ts = Local::now().naive_utc();
-                    }
-                    SynchronizerState::Delayed(_) => {
-                        // Already delayed, check if we should go stale
-                        // DON'T update modify_ts here - we want to track time since first delay
-                        self.check_and_transition_to_stale_if_needed(
-                            stale_threshold,
-                            Some(previous_block.clone()),
-                        )?;
-                    }
-                    _ => {
-                        // For other states, use the stale check
-                        if !self.check_and_transition_to_stale_if_needed(
-                            stale_threshold,
-                            Some(previous_block.clone()),
-                        )? {
-                            self.state = SynchronizerState::Delayed(previous_block.clone());
-                            self.modify_ts = Local::now().naive_utc();
-                        }
-                    }
-                }
+                // No need to consider state since we only call this method if we are
+                // in the Ready state.
+                self.state = SynchronizerState::Delayed(previous_block.clone());
+                self.modify_ts = Local::now().naive_utc();
                 Ok(None)
             }
         }
@@ -912,13 +890,17 @@ where
     ///
     /// We call this if we detect a future detached block. This usually only happens if
     /// a synchronizer has a restart.
+    ///
+    /// ## Note
+    /// This method assumes that at least one synchronizer is in Advanced, Ready or
+    /// Delayed state, it will panic in case this is not the case.
     fn reinit_block_history(
         sync_streams: &mut HashMap<ExtractorIdentity, SynchronizerStream>,
         block_history: &mut BlockHistory,
     ) -> Result<BlockHistory, BlockSynchronizerError> {
-        let previous = block_history
-            .latest()
-            .expect("Old block history is not empty, startup finished at this point");
+        let previous = block_history.latest().expect(
+            "Old block history should not be empty, startup should have populated it at this point",
+        );
         let blocks = sync_streams
             .values()
             .filter_map(SynchronizerStream::get_current_header)
@@ -927,7 +909,7 @@ where
         let new_block_history = BlockHistory::new(blocks, 10)?;
         let latest = block_history
             .latest()
-            .expect("block history is not empty");
+            .expect("Block history should not be empty, we just populated it.");
         info!(
              %previous,
             %latest,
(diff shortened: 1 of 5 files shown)


tycho-client/src/feed/mod.rs
```diff
@@ -159,36 +173,73 @@ pub struct BlockSynchronizer<S> {
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 #[serde(tag = "status", rename_all = "lowercase")]
 pub enum SynchronizerState {
+    /// Initial state, assigned before trying to receive any message
```
tvinagre, line 176: way better

tycho-client/src/feed/mod.rs
```diff
@@ -159,36 +173,73 @@ pub struct BlockSynchronizer<S> {
...
+    /// The synchronizer emitted a message for the block as expected
     Ready(BlockHeader),
-    // no progress, we consider it stale at that point we should purge it.
-    Stale(BlockHeader),
+    /// The synchronizer is on a previous block compared to others and is expected to
+    /// catch up soon.
     Delayed(BlockHeader),
+    /// The synchronizer hasn't emitted messages for > `max_missed_blocks` or has
+    /// fallen far behind. At this point we do not wait for it anymore but it can
+    /// still eventually recover.
```
tvinagre, line 185: can it? I thought we purged stale synchronizers

tycho-client/src/feed/mod.rs
```diff
@@ -159,36 +173,73 @@ pub struct BlockSynchronizer<S> {
...
-    Ended,
+    /// The synchronizer ended with an error.
+    Ended(String),
+}
+
+impl Display for SynchronizerState {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            SynchronizerState::Started => write!(f, "Started"),
+
```
tvinagre: empty line?

tycho-client/src/feed/mod.rs
```diff
@@ -308,19 +376,29 @@ impl SynchronizerStream {
...
+                    // Synchronizer errored during catch up
+                    self.mark_errored(e);
                     return Ok(None);
                 }
+                Ok(None) => {
+                    // This case should not happen, as we shouldn't poll the synchronizer after we
+                    // closed it or after it errored.
+                    warn!(
+                        %extractor_id,
+                        "Tried to poll from closed synchronizer during catch up.",
```
tvinagre, line 375: If it shouldn't happen, I'd add this information to the warn msg

tycho-client/src/feed/mod.rs
```diff
@@ -609,147 +748,252 @@ where
...
+            .expect("block history is not empty");
+        info!(
+             %previous,
+            %latest,
+            "Advanced synchronizer detected. Reinitialized block history."
+        );
+        sync_streams
+            .values_mut()
+            .for_each(|stream| {
+                // we only get headers from, advanced, ready and delayed so stale
```
tvinagre: from advanced,

tycho-client/src/feed/mod.rs
```diff
@@ -236,24 +297,31 @@ impl SynchronizerStream {
...
-                self.modify_ts = Local::now().naive_utc();
+                self.mark_closed();
                 Ok(None)
             }
             Err(_) => {
                 // trying to advance a block timed out
-                debug!(%extractor_id, ?previous_block, "No block received within time limit.");
+                debug!(%extractor_id, %previous_block, "No block received within time limit.");
 
+                // TODO: as per prev condition check, the state will always be always ready here -
```
tvinagre: will you still address this?

tycho-common/src/display.rs
```diff
@@ -0,0 +1,20 @@
...
+impl<'a, T: Display> Display for DisplayOption<'a, T> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self.0 {
+            Some(inner) => write!(f, "{}", inner),
+            None => write!(f, "None"),
+        }
+    }
+}
+
+/// Convenience function so you can write `%opt!(value)` in `tracing` logs.
```
zizou0x: It's not a macro so I think it's this instead?

```suggestion
/// Convenience function so you can write `%opt(value)` in `tracing` logs.
```

tycho-client/src/feed/mod.rs
```diff
@@ -397,8 +475,8 @@ impl SynchronizerStream {
             BlockPosition::NextExpected => {
                 self.state = SynchronizerState::Ready(latest_retrieved.clone());
                 trace!(
-                    next = ?latest_retrieved,
-                    extractor = ?extractor_id,
+                    next = %latest_retrieved,
+                    extractor = %extractor_id,
```
zizou0x, line 457: Does this remove some info or it's just nicer? As this is tracing level we should keep full data here

tycho-client/src/feed/mod.rs
```diff
@@ -609,147 +748,252 @@ where
...
-                    .filter_map(|v| match &v.state {
-                        SynchronizerState::Ready(b) | SynchronizerState::Delayed(b) => Some(b),
-                        _ => None,
-                    })
-                    .max_by_key(|b| b.number)
-                {
-                    block_history.push(header.clone())?;
+    /// Checks if we still have at least one active stream else errors.
+    ///
+    /// If there are not active streams meaning all  areended or stale, it returns a
```
zizou0x:
```suggestion
    /// If there are not active streams meaning all are ended or stale, it returns a
```

tycho-client/src/feed/mod.rs
```diff
@@ -609,147 +748,252 @@ where
...
+            {
+                if let Some(err) = &last_errored_stream.error {
+                    // All synchronizers were errored/stale and the last one errored
+                    reason.push(format!("Synchronizer for {last_errored_id} errored with: {err}"))
                 } else {
-                    // No Ready or Delayed synchronizers, but we still have some synchronizers
-                    // we can probably recover here but since this is unlikely we error for now
-                    error!("Only advanced SynchronizerStreams remain");
-                    return Err(BlockSynchronizerError::NoReadySynchronizers.into());
+                    // All synchronizer were errored/stale and the last one also becomae stale
```
zizou0x:
```suggestion
                    // All synchronizer were errored/stale and the last one also became stale
```

tycho-client/src/feed/mod.rs
```diff
@@ -609,147 +748,252 @@ where
...
+    ///
+    /// We call this if we detect a future detached block. This usually only happens if
+    /// a synchronizer has a restart.
+    fn reinit_block_history(
+        sync_streams: &mut HashMap<ExtractorIdentity, SynchronizerStream>,
+        block_history: &mut BlockHistory,
+    ) -> Result<BlockHistory, BlockSynchronizerError> {
+        let previous = block_history
+            .latest()
+            .expect("Old block history is not empty, startup finished at this point");
```
zizou0x: I find this message unclear, do you mean "Old block history should not be empty"?

tycho-client/src/feed/mod.rs
```diff
@@ -609,147 +748,252 @@ where
...
+        let blocks = sync_streams
+            .values()
+            .filter_map(SynchronizerStream::get_current_header)
+            .cloned()
+            .collect();
+        let new_block_history = BlockHistory::new(blocks, 10)?;
+        let latest = block_history
+            .latest()
+            .expect("block history is not empty");
```
zizou0x: Same here

zizou0x (approved):
LGTM, just a few small comments

tycho-client/src/feed/mod.rs
kayibal: It was changed in this commit it can now recover but it does not cause any additional latency anymore since we do not wait for it to recover.

tycho-client/src/feed/mod.rs
kayibal: It omits chain and instead just shows the protocol name

tycho-client/src/feed/mod.rs
kayibal: Mh no I'd prefer to not be overly dramatic in the logs. I think a comment here is good enough if this happens it's hopefully not a big deal.

PR #712, opened 2025-09-26
feat: add configurable batch commit threshold for extractors

ElusAegis:
- add a CLI config controlling how many finalized blocks must accumulate before we flush to Postgres (default 128, matching Ethereum finality)
  - refactor `ProtocolExtractor` to buffer messages via a helper that drains/commits only once the batch size is met, and capture the committed watermark consistently
  - extend the reorg buffer with count helpers plus targeted unit tests to cover the new batching
  behaviour

### DIFF
--- a/tycho-indexer/src/cli.rs
+++ b/tycho-indexer/src/cli.rs
@@ -48,6 +48,11 @@ pub struct GlobalArgs {
     )]
     pub database_url: String,
 
+    /// Batch size for the database inserts
+    /// Defaults to Ethereum finality of 128 blocks
+    #[clap(long, default_value = "128")]
+    pub database_insert_batch_size: u64,
+
     /// Name of the s3 bucket used to retrieve spkgs
     #[clap(env = "TYCHO_S3_BUCKET", long, default_value = "repo.propellerheads-propellerheads")]
     //Default is for backward compatibility but needs to be removed later
@@ -205,6 +210,8 @@ mod cli_tests {
             "http://example.com",
             "--database-url",
             "my_db",
+            "--database-insert-batch-size",
+            "256",
             "--rpc-url",
             "http://example.com",
             "run",
@@ -227,6 +234,7 @@ mod cli_tests {
             global_args: GlobalArgs {
                 endpoint_url: "http://example.com".to_string(),
                 database_url: "my_db".to_string(),
+                database_insert_batch_size: 256,
                 rpc_url: "http://example.com".to_string(),
                 s3_bucket: Some("repo.propellerheads-propellerheads".to_string()),
                 server_ip: "0.0.0.0".to_string(),
@@ -275,6 +283,7 @@ mod cli_tests {
             global_args: GlobalArgs {
                 endpoint_url: "http://example.com".to_string(),
                 database_url: "my_db".to_string(),
+                database_insert_batch_size: 128,
                 rpc_url: "http://example.com".to_string(),
                 s3_bucket: Some("repo.propellerheads-propellerheads".to_string()),
                 server_ip: "0.0.0.0".to_string(),
--- a/tycho-indexer/src/cli.rs
+++ b/tycho-indexer/src/cli.rs
@@ -49,8 +49,8 @@ pub struct GlobalArgs {
     pub database_url: String,
 
     /// Batch size for the database inserts
-    /// Defaults to Ethereum finality of 128 blocks
-    #[clap(long, default_value = "128")]
+    /// Defaults to Ethereum finality of 64 blocks
+    #[clap(long, default_value = "64")]
     pub database_insert_batch_size: usize,
 
     /// Name of the s3 bucket used to retrieve spkgs
@@ -283,7 +283,7 @@ mod cli_tests {
             global_args: GlobalArgs {
                 endpoint_url: "http://example.com".to_string(),
                 database_url: "my_db".to_string(),
-                database_insert_batch_size: 128,
+                database_insert_batch_size: 64,
                 rpc_url: "http://example.com".to_string(),
                 s3_bucket: Some("repo.propellerheads-propellerheads".to_string()),
                 server_ip: "0.0.0.0".to_string(),
--- a/tycho-indexer/src/cli.rs
+++ b/tycho-indexer/src/cli.rs
@@ -51,7 +51,7 @@ pub struct GlobalArgs {
     /// Batch size for the database inserts
     /// Defaults to Ethereum finality of 128 blocks
     #[clap(long, default_value = "128")]
-    pub database_insert_batch_size: u64,
+    pub database_insert_batch_size: usize,
 
     /// Name of the s3 bucket used to retrieve spkgs
     #[clap(env = "TYCHO_S3_BUCKET", long, default_value = "repo.propellerheads-propellerheads")]
--- a/tycho-indexer/src/cli.rs
+++ b/tycho-indexer/src/cli.rs
@@ -49,8 +49,7 @@ pub struct GlobalArgs {
     pub database_url: String,
 
     /// Batch size for the database inserts
-    /// Defaults to Ethereum finality of 64 blocks
-    #[clap(long, default_value = "64")]
+    #[clap(long, default_value = "0")]
     pub database_insert_batch_size: usize,
 
     /// Name of the s3 bucket used to retrieve spkgs
@@ -283,7 +282,7 @@ mod cli_tests {
             global_args: GlobalArgs {
                 endpoint_url: "http://example.com".to_string(),
                 database_url: "my_db".to_string(),
-                database_insert_batch_size: 64,
+                database_insert_batch_size: 0,
                 rpc_url: "http://example.com".to_string(),
                 s3_bucket: Some("repo.propellerheads-propellerheads".to_string()),
                 server_ip: "0.0.0.0".to_string(),
--- a/tycho-indexer/src/extractor/protocol_extractor.rs
+++ b/tycho-indexer/src/extractor/protocol_extractor.rs
@@ -60,6 +60,8 @@ pub struct Inner {
 
 pub struct ProtocolExtractor<G, T, E> {
     gateway: G,
+    #[allow(dead_code)]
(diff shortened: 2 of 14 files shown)


tycho-indexer/src/cli.rs
```diff
@@ -48,6 +48,11 @@ pub struct GlobalArgs {
     )]
     pub database_url: String,
 
+    /// Batch size for the database inserts
+    /// Defaults to Ethereum finality of 128 blocks
+    #[clap(long, default_value = "128")]
```
zizou0x: We should default to 0 to maintain logic stability with the current version. 

Then we can manually set this specifically for each chain and slowly rollout this

tycho-indexer/src/extractor/reorg_buffer.rs
```diff
@@ -187,6 +187,15 @@ where
...
     }
 
+    /// Returns the number of blocks in the buffer with a block number less than the specified
+    /// target block number. Assumes blocks are stored in ascending order.
+    pub fn count_blocks_before(&self, target_block: u64) -> usize {
+        self.block_messages
+            .iter()
+            .take_while(|msg| msg.block().number < target_block)
+            .count()
+    }
```
zizou0x, line 200: I guess this returns `block_messages.len()` if the target_block isn't found?

It's better to return a `None` in that case, it would give us more information. So the caller knows the block wasn't found and it can decide what to do accordingly.

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -584,6 +588,54 @@ where
             .collect();
         Ok(new_tokens)
     }
+
+    async fn insert_and_maybe_flush_buffer(
+        &self,
+        msg: &BlockChanges,
+        inp: &BlockScopedData,
+        is_syncing: bool,
+    ) -> Result<u64, ExtractionError> {
```
zizou0x: I'm not a big fan of this function. The reasons are: it's only used once (so it doesn't remove any code duplication) and it's responsible for several things (so quite complex to understand without looking into it's code). I think it makes the `handle_tick_scoped` function logic harder to follow.

I'd suggest two alternatives:

- Keep everything in line in the `handle_tick_scoped` function
- Make this function only handle reorg buffer things -> it would take the buffer lock as input then insert the new block, check if we want to flush and return `Result<Vec<msgs>, ExtractionError>`. Then in the extractor we would handle the actual db pushs and get the `committed_upto_height_from_buffer`. But need to make sure we can do this with only one `self.reorg_buffer.lock().await;`, else it wouldn't be efficient.

tycho-indexer/src/extractor/runner.rs
```diff
@@ -655,9 +663,14 @@ impl ExtractorBuilder {
             None
         };
 
+        let database_insert_batch_size = self
+            .database_insert_batch_size
+            .ok_or(ExtractionError::Setup("Database insert batch size not set".to_string()))?;
+
```
zizou0x, line 674: It's better to just `unwrap_or(0)` here so we're sure we don't break anything unexpectedly - there could be some places in the code where this is not set, for example `run_spkg`? It's fine to set this to 0 because 0 basically means we keep the current behavior.

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -1732,17 +1772,93 @@ mod test {
     }
 
     #[tokio::test]
-    async fn test_handle_tick_scoped_data_old_native_msg() {
+    async fn test_reorg_buffer_flush_respects_batch_size() {
```
zizou0x, line 1774: Nice test, it's simple and it works

tycho-indexer/src/cli.rs
ElusAegis: Makes sense, thank you for the point!

tycho-indexer/src/extractor/reorg_buffer.rs
ElusAegis: Do we need to know if the target block exists in the buffer for this function? As sematically it hust counts the blocks in the buffer that are older than a given block number

tycho-indexer/src/extractor/protocol_extractor.rs
ElusAegis: Done! However, I have a tradeoff between the performance and clarity of code, which I was wondering about.

We only drop the `reorg_buffer` lock after we insert to the db, though we could drop it earlier. However, if we do so, we need to add an early exit from the scope, which would slightly complicate the code. Is it useful to add the early exit, or we do not care about performance here that much?

```rust
 if reorg_buffer.count_blocks_before(inp.final_block_height) >=
    self.database_insert_batch_size
{
    let mut msgs = reorg_buffer
        .drain_blocks(inp.final_block_height)
        .map_err(ExtractionError::Storage)?
        .into_iter()
        .peekable();

    while let Some(msg) = msgs.next() {
        // Force a database commit if we're not syncing and this is the last block to be
        // sent. Otherwise, wait to accumulate a full batch before
        // committing.
        let force_db_commit = if is_syncing { false } else { msgs.peek().is_none() };

        self.gateway
            .advance(msg.block_update(), msg.cursor(), force_db_commit)
            .await?;
    }
}

reorg_buffer
    .get_oldest_block()
    .map(|block| block.number)
    .ok_or_else(|| ExtractionError::ReorgBufferError("Reorg buffer is empty".into()))?
```

tycho-indexer/src/extractor/protocol_extractor.rs
ElusAegis: 🙏

tycho-indexer/src/extractor/reorg_buffer.rs
zizou0x: Ok that's fair, let's not leak too much information if we don't need it 👍

tycho-indexer/src/extractor/protocol_extractor.rs
zizou0x: We should definitely not keep the lock unnecessarily here.

I think we can simply return `(msgs, ...get_oldest_block().number)` from the scope, or you could even get the `commited_upto_height` from the the last entity in `msgs` (+1 to keep the same logic)

ElusAegis: Resolved this using some heuristics about what is the oldest uncommitted block height

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -738,41 +742,63 @@ where
...
-                .into_iter()
-                .peekable();
+            let commit_upto_block_height = inp.final_block_height;
 
-            while let Some(msg) = msgs.next() {
-                // Force a database commit if we're not syncing and this is the last block to be
-                // sent. Otherwise, wait to accumulate a full batch before
-                // committing.
-                let force_db_commit = if is_syncing { false } else { msgs.peek().is_none() };
+            if reorg_buffer.count_blocks_before(commit_upto_block_height) >=
```
zizou0x: I know it's equivalent but it would make more sense to use `inp.final_block_height` here

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -738,41 +742,63 @@ where
...
+                self.database_insert_batch_size
+            {
+                let mut msgs = reorg_buffer
+                    .drain_blocks_until(commit_upto_block_height)
```
zizou0x: Same here

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -738,41 +742,63 @@ where
...
+                while let Some(msg) = msgs.next() {
+                    // Force a database commit if we're not syncing and this is the last block to be
+                    // sent. Otherwise, wait to accumulate a full batch before
+                    // committing.
+                    let force_db_commit = if is_syncing { false } else { msgs.peek().is_none() };
+
+                    self.gateway
+                        .advance(msg.block_update(), msg.cursor(), force_db_commit)
+                        .await?;
+                }
```
zizou0x: I'm opened for discussion here but IMO: Your logic works but it can be quite hard to follow for someone not used to what this does. I think it would be easier to understand if the design was: gather blocks to commit (if any) and loop through all of them.

Meaning the if condition would only affect the gathering part and the loop always run (obviously it would be a no-op if the blocks to commit are empty). Something like this: 

```
        let (oldest_uncommitted_block_height, blocks_to_commit) = {
            let mut reorg_buffer = self.reorg_buffer.lock().await;

            // Insert the new block into the buffer
            reorg_buffer
                .insert_block(BlockUpdateWithCursor::new(msg.clone(), inp.cursor.clone()))
                .map_err(ExtractionError::Storage)?;

            // Drain blocks if we've accumulated enough finalized blocks
            let should_drain = reorg_buffer.count_blocks_before(inp.final_block_height) >=
                self.database_insert_batch_size;

            let blocks_to_commit = if should_drain {
                reorg_buffer
                    .drain_blocks_until(inp.final_block_height)
                    .map_err(ExtractionError::Storage)?
            } else {
                Vec::new()
            };

            // Track the oldest block still in the buffer (not yet committed)
            let oldest_uncommitted = reorg_buffer
                .get_oldest_block()
                .map(|block| block.number)
                .ok_or_else(|| ExtractionError::ReorgBufferError("Reorg buffer is empty".into()))?;

            (oldest_uncommitted, blocks_to_commit)
        };

        // Commit the drained blocks if any
        let mut blocks_iter = blocks_to_commit.into_iter().peekable();
        while let Some(block) = blocks_iter.next() {
            // Force a database commit on the last block when not syncing
            let is_last_block = blocks_iter.peek().is_none();
            let force_db_commit = !is_syncing && is_last_block;

            self.gateway
                .advance(block.block_update(), block.cursor(), force_db_commit)
                .await?;
        }

        // At this point, we have committed all blocks up to the oldest uncommitted block height.
        let db_committed_upto_block_height = oldest_uncommitted_block_height;
```

zizou0x (approved):
LGTM thanks!

PR #715, opened 2025-09-30
fix: correctly parse token balances in token-analyzer

zizou0x:
Alloy `U256::from_be_bytes` is expecting a [u8; 32] but we were not padding the balances before so it was failing in the conversion.

### DIFF
--- a/tycho-ethereum/src/token_analyzer/trace_call.rs
+++ b/tycho-ethereum/src/token_analyzer/trace_call.rs
@@ -96,9 +96,10 @@ impl TraceCallDetector {
                 let amount = cmp::max(
                     U256::from_be_bytes::<32>(
                         balance
+                            .lpad(32, 0)
                             .as_ref()
                             .try_into()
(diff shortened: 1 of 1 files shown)


tycho-ethereum/src/token_analyzer/trace_call.rs
```diff
@@ -96,9 +96,12 @@ impl TraceCallDetector {
                 let amount = cmp::max(
                     U256::from_be_bytes::<32>(
                         balance
+                            .lpad(32, 0)
                             .as_ref()
                             .try_into()
-                            .expect("slice with incorrect length"),
+                            .unwrap_or_else(|e| {
+                                panic!("slice with incorrect length: {e}. balance: {balance:?}")
```
kayibal: can we use expect here and say it should never happen due to your lpad?

dianacarvalho1 (approved):
Thank you @zizou0x 🙏🏼

PR #717, opened 2025-09-30
feat: make RPC retry config more permissive

tvinagre:


tycho-indexer/src/extractor/dynamic_contract_indexer/hooks_dci_setup.rs
```diff
@@ -52,7 +52,7 @@ pub fn setup_metadata_registries(
 
     // Register RPC provider with default routing key and retry configuration
     let retry_config =
-        RPCRetryConfig { max_retries: 3, initial_backoff_ms: 100, max_backoff_ms: 5000 };
+        RPCRetryConfig { max_retries: 5, initial_backoff_ms: 150, max_backoff_ms: 5000 };
```
zizou0x, line 55: Can we make this configurable?

PR #719, opened 2025-10-02
refactor: move database commit handling to an asynchronous tasks in ws

ElusAegis:
## Decouple protocol extractor DB commits from the hot path

 - `ProtocolExtractor` does DB writes in a single `tokio::spawned` task, joining the previous handle before scheduling a new batch and capturing the committed height from each task’s return value
  - poll the outstanding commit task on every `handle_tick_scoped_data` call so we get completed heights as soon as it is committed
  - represent commit progress as `Option<u64>` (highest committed block) instead of inferring from the oldest buffered block, updating both extractor output and `PendingDeltas` flushing logic (drain_blocks_until) to respect the new semantics


tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -72,12 +72,13 @@ pub struct ProtocolExtractor<G, T, E> {
     /// Allows to attach some custom logic, e.g. to fix encoding bugs without resync.
     post_processor: Option<fn(BlockChanges) -> BlockChanges>,
     reorg_buffer: Mutex<ReorgBuffer<BlockUpdateWithCursor<BlockChanges>>>,
+    database_commit_task: Mutex<Option<JoinHandle<Result<u64, ExtractionError>>>>,
```
zizou0x: Would be nice to store the `db_committed_block_height` here too, this would allow us to always return it as `Some()` unless nothing has been committed yet (i.e. a new protocol just started to sync)

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -756,46 +759,97 @@ where
...
+                                format!("Failed to join database commit task: {join_err}"),
+                            )));
+                        }
+                    }
+                } else {
+                    None
+                }
+            } else {
+                None
+            };
```
zizou0x: If we store the `db_committed_block_height` between blocks in the Mutex we can handle this better

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -756,46 +759,97 @@ where
...
+        // See if we need to join the previous database commit task based on whether it has finished
+        // or if we have new blocks to commit and launch a new database commit task if needed.
+        let db_committed_block_height = {
+            let mut guard = self.database_commit_task.lock().await;
+
+            // We need to join the previous task if it has finished or if we have new blocks to
+            // commit.
+            let need_join = guard
+                .as_ref()
+                .is_some_and(|h| h.is_finished()) ||
```
zizou0x: This line is redundant with the `if let Some()` below, do you think we can handle this directly in the condition below to only keep one of them?

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -756,46 +759,97 @@ where
...
+                let new_handle = tokio::spawn(async move {
+                    // Compute the height we’ll reach in this batch
+                    let most_recent_committed_block_height = blocks_to_commit
+                        .last()
+                        .map(|b| b.block_update.block.number)
+                        .ok_or_else(|| {
+                            ExtractionError::Storage(StorageError::Unexpected(
+                                "No blocks to commit".into(),
+                            ))
+                        })?;
```
zizou0x: Can we do this outside of the task please? 

Also the `last()` should always be `Some()` here because we checked `!blocks_to_commit.is_empty()` above.

We can probably do something like this instead (once again to avoid having the same condition twice).

```
// If we have blocks to commit, spawn a new task to commit them to the database.
if let Some(last_block_height) = blocks_to_commit.last().map(|b| b.block_update.block.number){
...
```

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -756,46 +759,97 @@ where
...
+                            .map_err(ExtractionError::Storage)?;
+                    }
+
+                    Ok(most_recent_committed_block_height)
+                });
 
-        // At this point, we have committed all blocks up to the oldest uncommitted block height.
-        let db_committed_upto_block_height = oldest_uncommitted_block_height;
+                // Store the new handle (short lock)
+                *guard = Some(new_handle);
```
zizou0x: As a safeguard for our data, we should double check that we're not replacing anything here and that it's always empty before we insert this.

tycho-indexer/src/services/deltas_buffer.rs
```diff
@@ -736,6 +738,20 @@ mod test {
         }
     }
 
+    fn simple_block_changes(
+        block_number: u64,
+        db_committed_upto_block_height: Option<u64>,
```
zizou0x: Don't forget to adapt the name here please

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -1771,8 +1817,9 @@ mod test {
     }
 
     #[tokio::test]
-    async fn test_reorg_buffer_flush_respects_batch_size() {
+    async fn test_handle_tick_scoped_respects_batch_async_commit() {
```
zizou0x, line 1826: :fire:

tycho-indexer/src/extractor/reorg_buffer.rs
```diff
@@ -157,6 +157,7 @@ where
         {
             if block_message.block().hash == target_hash {
                 target_index = Some(self.block_messages.len() - index);
+                break;
```
zizou0x, line 160: Nice catch!

tycho-indexer/src/services/deltas_buffer.rs
ElusAegis: Nice, thank you for pointiing out!

tycho-indexer/src/extractor/protocol_extractor.rs
ElusAegis: Sounds good. It should be impossible for it not to be None due to us still holding the lock, but I will do a double check and a clear failor to make it clearer

tycho-indexer/src/extractor/protocol_extractor.rs
ElusAegis: Perfect, just did. Indeed makes the code a lot cleaner

tycho-indexer/src/extractor/protocol_extractor.rs
ElusAegis: The way the logic works right now is that on every call to the `handle_tick_scoped_data`, even if there is no commit to be done, we still check to see if the previous commit task finished, and if so, report it immediatelly to the Deltas buffer so that we purged the committed blocks. So that is why I checked for either if we can already join the task with `h.is_finished())` or if we need to mandatory do it as there is another commit we need to do. We can simplify this ofcource to only join when we are doing another commit, but this will loose a fraction of optimisation (though not anything significant it seems).

tycho-indexer/src/extractor/protocol_extractor.rs
ElusAegis: I have not thought about this, to be honest. However, is there a reason why returning None would be problematic, and we are better off rebroadcasting the same committed height? It seemed to me originally that there was no way a subscriber could miss a single message from the stream, as otherwise they will immediatelly be unsubscribed. And if they connected late after the purge message, they would not have the blocks to purge anyway in the their deltas buffer

ElusAegis:
@zizou0x I made a significant adjustment in the logic:

1. Introduced the `GatewayInner` to contain all of the information about the Gateway / Commit Management 
2. `committed_block_height` is now stored in the `Extractor`, and is set from the async commit thread. This allows for much cleaner code and unifies the logic that the `handle_tick_scoped_data` always reads from one place
3. If we are in the non committing iteration of `handle_tick_scoped_data`, we no longer attempt to join the future, as this unnecessarily complicated the code.

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -58,9 +59,23 @@ pub struct Inner {
...
+    committed_block_height: Arc<Mutex<Option<u64>>>,
+    commit_batch_size: usize,
+}
+
+impl<G> Deref for GatewayInner<G> {
+    type Target = G;
+    fn deref(&self) -> &Self::Target {
+        &self.gw
+    }
+}
```
zizou0x: There are a lot of discussion about this pattern but it's mostly considered anti-pattern by the rust community, so we avoid using it in our code. You can read more about it [here](https://rust-unofficial.github.io/patterns/anti_patterns/deref.html#deref-polymorphism). The reason is because this is using the `Deref` beyond its purpose, so it's unexpected by programmers.

tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -777,78 +801,55 @@ where
...
-                        }
-                        Err(join_err) => {
-                            // the task panicked/cancelled
-                            return Err(ExtractionError::Storage(StorageError::Unexpected(
-                                format!("Failed to join database commit task: {join_err}"),
-                            )));
-                        }
+        // If we have blocks to commit, wait for the previous async commit task to finish and then
+        // spawn a new task that will commit the new blocks and update the committed block height.
+        if let Some(last_block) = blocks_to_commit.last() {
```
zizou0x, line 799: I have a preference for the previous version where we were checking each block the reasons is:

Here we're updating `db_commit_handle_to_join` only when there is a new block to commit. This means if we have a huge batch (let's imagine 100 blocks) we will keep the previous 100 blocks in the buffers until the new batch is ready. In practice, db commits take a few seconds at most, so we're keeping these blocks buffered for minutes instead of seconds. This issue becomes even bigger if we increase batch size.

tycho-indexer/src/extractor/protocol_extractor.rs
ElusAegis: Thank you for pointing it out. And what do you think about the `GatewayInner` structure in general for the purpose?

tycho-indexer/src/extractor/protocol_extractor.rs
ElusAegis: Just to confirm, you are referring to the `blocks_to_commit` that were moved to the async task? As far as I understand, they will be dropped as soon as the task has finished, so it will not stick around for the remainder of the time. And the Join handle itself is tiny, as it does not return anything. So storing it until the next cleanup should be negligible

tycho-indexer/src/extractor/protocol_extractor.rs
zizou0x: I like it!

tycho-indexer/src/extractor/protocol_extractor.rs
zizou0x: I re-read this code and it looks good in the end 👍

zizou0x (approved):
Looks perfect! Thanks for this one, looking forward to see the perf improvements!

PR #818, opened 2025-12-23
feat: minor adjustments to `tycho-ethereum`

ElusAegis:
1. Renamed the `retry_request` to `call_with_retry` for clarity
2. Added a custom RPC error code `3` from Chainstack that should not be retried (execution reverted)
3. Added `simulate_txs_with_trace` method with tests to replace currently custom implementation in the [`integration tests`](https://github.com/propeller-heads/tycho-simulation/blob/47f1e95e6aa45732f6e9129e0c0b20e5fca27ed5/tycho-test/src/execution/execution_simulator.rs#L205)

### DIFF
--- a/crates/tycho-ethereum/src/rpc/mod.rs
+++ b/crates/tycho-ethereum/src/rpc/mod.rs
@@ -979,6 +979,10 @@ impl EthereumRpcClient {
                     }
                 });
 
+                if all_failed {
+                    debug!("All slot detector test requests in batch failed, will retry the entire batch");
+                }
+
                 all_failed || some_retryable_failed
             };
 
--- a/crates/tycho-ethereum/src/rpc/mod.rs
+++ b/crates/tycho-ethereum/src/rpc/mod.rs
@@ -6,15 +6,18 @@ use std::{
 };
 
 use alloy::{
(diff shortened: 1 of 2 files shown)


tycho-ethereum/src/rpc/retry.rs
```diff
@@ -88,6 +88,9 @@ pub(super) fn has_custom_retry_code<T>(e: &ErrorPayload<T>) -> bool {
         -32602 => false, // "invalid params" - wrong parameters
         -32604 => false, // "method not supported" - not supported by this node
 
+        // Other non EIP-1474 errors
+        3 => false, // "execution reverted" - non-retryable EVM execution error
```
zizou0x: Have you done some research here? Is this specific to Chainstack or generic to all nodes?

zizou0x (commented):
Just one question that concerns me, code-wise it's perfect

tycho-ethereum/src/rpc/retry.rs
ElusAegis: Just did some more digging, and the error code 3 is a [specific error code](https://ethereum.github.io/execution-apis/api-documentation/) for the `eth_call` and `eth_estimateGas` and implies the `execution reverted` case. Will update the docs to clarify this

zizou0x (approved):

PR #826, opened 2026-01-19
feat: Universal JIT Liquidity hook integration for tycho

wirew0lf:
- Add Universal JIT Liquidity hook integration for querying pool liquidity and trading limits
  - Implement UniversalMetadataGenerator using a lens contract with state overrides to fetch liquidity data via getLiquidityForPool and limits via getLimitsForPool
  - Register hook under universal_v1 identifier


tycho-indexer/src/extractor/dynamic_contract_indexer/hooks/integrations/universal/metadata_generator.rs
```diff
@@ -0,0 +1,620 @@
...
+                    "data": format!("0x{}000000000000000000000000{}000000000000000000000000{}", GET_LIMITS_SELECTOR, token_0_str, token_1_str),
+                    "to": lens_address
+                }),
+                json!(format!("0x{:x}", block.number)),
+                json!({
+                    lens_address: {
+                        "code": format!("0x{}", lens_bytecode_hex),
+                        "state": {
+                            // Slot 0: HOOK address
+                            "0x0000000000000000000000000000000000000000000000000000000000000000": format!("0x000000000000000000000000{}", target_str)
```
zach030: I didn’t see the `UniversalJITLens` source code, do you know if this override is actually useful?

zach030: Euler is a bit of a special case. We wrote an adapter for getLimit, so we call it via a temporary address with state overrides. If `UniversalJITLens` is deployed, you can just use the actual address.

wirew0lf: Actually you're right, we've just deployed the contract and set everything as constants. So no need for state overrides

tycho-indexer/src/extractor/dynamic_contract_indexer/hooks/integrations/universal/metadata_generator.rs
```diff
@@ -0,0 +1,601 @@
...
+                    let target = component
+                        .static_attributes
+                        .get("hooks")
+                        .expect("Hooks attribute not found");
+
+                    let calldata = params["data"]
+                        .as_str()
+                        .ok_or(MetadataError::UnknownError("calldata not found".to_string()))?;
+
+                    Ok(EntryPointWithTracingParams {
```
zach030, line 224: The target of this entrypoint should be the lens address, not the hook address.

tycho-indexer/src/extractor/dynamic_contract_indexer/hooks/integrations/universal/metadata_generator.rs
```diff
@@ -0,0 +1,601 @@
...
+                }
+
+                // getLimitsForPool(token0, token1) returns:
+                // - First uint256 (0..64): output limit for token1 → token0 swap
+                // - Second uint256 (64..128): output limit for token0 → token1 swap
+                let limit_1_to_0 = Bytes::from(&res_str[0..64]);
+                let limit_0_to_1 = Bytes::from(&res_str[64..128]);
+
+                // Return both directions (limit_1 slot is unused, set to default)
+                Ok(MetadataValue::Limits(vec![
```
zach030, line 259: Can the lens provide maxAmountOut? If not, it’s probably better to default it to `U256::MAX`.

zach030 (commented):
Thanks! Looks good. I noticed a few minor issues during testing, you can take a look.

tycho-indexer/src/extractor/dynamic_contract_indexer/hooks/integrations/universal/mod.rs
```diff
@@ -0,0 +1,24 @@
...
+
+pub(super) mod metadata_generator;
+
+pub(super) fn register_universal_integrations(
+    generator_registry: &mut MetadataGeneratorRegistry,
+    parser_registry: &mut MetadataResponseParserRegistry,
+    _provider_registry: &mut ProviderRegistry,
+    rpc_url: String,
+) {
+    generator_registry.register_hook_identifier(
```
zach030, line 18: I think we should use `generator_registry.register_hook_generator` instead of registering by identifier, since regular Uniswap V4 pools don’t expose this attribute. See [Step 4: Register Components](https://docs.propellerheads.xyz/tycho/for-dexs/protocol-integration/indexing/common-problems-and-patterns/dynamic-contract-indexing-dci/hook-integration-guide) in the doc.

tycho-indexer/src/extractor/dynamic_contract_indexer/hooks/integrations/universal/metadata_generator.rs
```diff
@@ -0,0 +1,601 @@
...
+        let limits_transport = RpcTransport::new(
+            self.rpc_url.clone(),
+            "eth_call".to_string(),
+            vec![
+                json!({
+                    "data": format!("0x{}000000000000000000000000{}000000000000000000000000{}", GET_LIMITS_SELECTOR, token_0_str, token_1_str),
+                    "to": lens_address
+                }),
+                json!(format!("0x{:x}", block.number)),
+            ],
```
zach030, line 81: I realized there’s still an issue here, `tycho-indexer` back process historical blocks, but the lens contract didn’t exist back then, so traces will fail. We probably need to pass bytecode before the quoter’s deployment block, and stop doing that afterward.

zach030 (commented):
I also found a few other issues while testing.

PR #855, opened 2026-02-17
fix(tycho-client): fetch snapshots on full blocks only

louise-poole:
Current start-up behaviour when partials are enabled:
- client starts stream and receives block n partials
- client receives block n+1 partial 1 (indicates block n was completed)
- client buffers the deltas and fetches snapshots at block n+1 -> RPC error as block n+1 is not complete yet and so is not available as a snapshot

The fix: 
Once the client sees the first partial of block n+1, it should buffer the deltas and fetch the snapshots at block n.

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,3 +1,7 @@
 [advisories]
 # rustls-webpki 0.101.7 is locked by tonic 0.9; fix requires tonic 0.11+ (prost 0.11 -> 0.12).
-ignore = ["RUSTSEC-2026-0049"]
+ignore = [
+    "RUSTSEC-2026-0049",
+    "RUSTSEC-2026-0098",
+    "RUSTSEC-2026-0099",
+]
--- a/tycho-client/src/feed/synchronizer.rs
+++ b/tycho-client/src/feed/synchronizer.rs
@@ -480,12 +480,19 @@ where
             // If possible skip retrieving snapshots
             let msg = if !self.is_next_expected(&header) {
                 info!("Retrieving snapshot");
-                 let snapshot_header = BlockHeader { revert: false, ..header.clone() };
+                // With partial blocks, the server only has full blocks in its buffer; pass the
+                // previous block's header so we request state at N-1, then merge with deltas.
+                let snapshot_header = if self.partial_blocks && header.number > 0 {
+                    BlockHeader {
+                        number: header.number - 1,
+                        hash: header.parent_hash.clone(),
+                        ..Default::default()
+                    }
+                } else {
+                    BlockHeader { revert: false, ..header.clone() }
+                };
                 let snapshot = self
-                    .get_snapshots::<Vec<&String>>(
-                        snapshot_header,
-                        None,
-                    )
+                    .get_snapshots::<Vec<&String>>(snapshot_header, None)
                     .await?
                     .merge(deltas_msg);
                 let n_components = self.component_tracker.components.len();
--- a/tycho-client/src/feed/synchronizer.rs
+++ b/tycho-client/src/feed/synchronizer.rs
@@ -94,6 +94,9 @@ pub struct ProtocolStateSynchronizer<R: RPCClient, D: DeltasClient> {
     compression: bool,
     partial_blocks: bool,
     uses_dci: bool,
+    /// Brand-new components (in new_protocol_components) whose snapshot we deferred until the
+    /// first message of the next block. Only used when partial_blocks is true.
+    deferred_snapshot_components: Vec<String>,
 }
 
 #[derive(Clone, PartialEq, Debug, Serialize, Deserialize)]
@@ -256,6 +259,7 @@ where
             compression,
             partial_blocks: false,
             uses_dci: false,
+            deferred_snapshot_components: Vec::new(),
         }
     }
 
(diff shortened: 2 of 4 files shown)


tycho-client/src/feed/synchronizer.rs
```diff
@@ -480,12 +480,19 @@ where
             // If possible skip retrieving snapshots
             let msg = if !self.is_next_expected(&header) {
                 info!("Retrieving snapshot");
-                 let snapshot_header = BlockHeader { revert: false, ..header.clone() };
+                // With partial blocks, the server only has full blocks in its buffer; pass the
+                // previous block's header so we request state at N-1, then merge with deltas.
+                let snapshot_header = if self.partial_blocks && header.number > 0 {
```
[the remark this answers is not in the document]
louise-poole: True. It's actually 2-fold: we need to get snapshots on the previous block, but we also need to defer getting snapshots of newly created components (as they do not exist on the previous block).

tycho-client/src/feed/synchronizer.rs
```diff
@@ -504,29 +515,88 @@ where
...
+                            let flushed_snapshots = if self.partial_blocks
+                                && !self.deferred_snapshot_components.is_empty()
+                                && self
+                                    .last_synced_block
+                                    .as_ref()
+                                    .is_some_and(|prev| header.number > prev.number)
+                            {
+                                let deferred =
+                                    std::mem::take(&mut self.deferred_snapshot_components);
+                                let prev = self.last_synced_block.as_ref().unwrap();
```
ElusAegis: A bit wary of the unwrap here

tycho-client/src/feed/synchronizer.rs
```diff
@@ -504,29 +515,88 @@ where
                             let header: BlockHeader = (&deltas).into();
                             debug!(block_number=?header.number, "Received delta message");
 
+                            // Flush deferred brand-new component snapshots on first message of next block.
+                            let flushed_snapshots = if self.partial_blocks
+                                && !self.deferred_snapshot_components.is_empty()
+                                && self
```
ElusAegis: Would be nice to simplify the last condition to say explicitly what this check verifies (if this is the same block or not). Also, perhaps this could be executed even for full blocks? As then the deferred_snapshot_components would be empty and the flushed_snapshots would be empty. And then we could get rid of the None case, and instead have it as empty.

tycho-client/src/feed/synchronizer.rs
```diff
@@ -504,29 +515,88 @@ where
...
+                                    let snapshot_header = if self.partial_blocks && header.number > 0
+                                    {
+                                        // get snapshot at the previous block (current block is only partial and not available for querying)
+                                        BlockHeader {
+                                            number: header.number - 1,
+                                            hash: header.parent_hash.clone(),
+                                            ..Default::default()
+                                        }
+                                    } else {
+                                        BlockHeader { revert: false, ..header.clone() }
```
ElusAegis, line 597: Looks like this might break if we are on block 0 and partial, as the indexer will not have the full block 0 as it is building partials. Perhaps instead it should be handled in the if case, and panic / error if we have a partial block with a header.number = 0? Same for the initial snapshot

tycho-client/src/feed/synchronizer.rs
```diff
@@ -504,29 +515,88 @@ where
...
+                                    self.deferred_snapshot_components.extend(brand_new);
+                                    preexisting
+                                } else {
+                                    requiring_snapshot
+                                };
+
+                                let mut snapshots = if request_now.is_empty() {
+                                    Snapshot::default()
+                                } else {
+                                    let snapshot_header = if self.partial_blocks && header.number > 0
```
ElusAegis: Could this be merged with the logic that fetches the `flushed_snapshots`? Or at least colocate it, such that it would be easier to follow

tycho-client/src/feed/synchronizer.rs
```diff
@@ -3100,4 +3170,196 @@ mod test {
         let _ = close_tx.send(());
         jh.await.expect("Task should not panic");
     }
+
+    /// Test that brand-new components (in new_protocol_components) have their snapshot deferred
+    /// to the first message of the next block when partial_blocks is enabled, and that the
+    /// snapshot is then requested at the completed previous block and included in that message.
+    #[test_log::test(tokio::test)]
```
ElusAegis, line 3292: Should the preexisting path be also validated? That it requests blocks immediately

ElusAegis (approved):
Approving to unblock, but to be honest the loop logic was pretty complicated for me to parse (especially the handling of preexisting and new components). Perhaps a helper method or some restrcuture could help? Or even more detailed comments that summarise the logic (what is deffered and when)

tycho-client/src/feed/synchronizer.rs
louise-poole: I've refactored it a bit now so that it's easier to follow. If prev block is not set yet (which is the case when streaming block 0) it will continue to defer the snapshots until it is set.
Not that I feel this is an issue - I highly doubt we'll ever index a chain and be live with it before block 0.

.cargo/audit.toml
```diff
@@ -1,3 +1,7 @@
 [advisories]
 # rustls-webpki 0.101.7 is locked by tonic 0.9; fix requires tonic 0.11+ (prost 0.11 -> 0.12).
-ignore = ["RUSTSEC-2026-0049"]
+ignore = [
+    "RUSTSEC-2026-0049",
+    "RUSTSEC-2026-0098",
+    "RUSTSEC-2026-0099",
+]
```
zizou0x, line 7: Are these also related to tonic? If not would be nice to comment what they are for

zizou0x (approved):
Approving in the behalf of Artem 👀

.cargo/audit.toml
louise-poole: Yes they are

PR #908, opened 2026-04-14
fix: update cli and python clients

louise-poole:
The CLI and python clients were outdated and did not reflect recent fixes or changes. This PR includes:
- refactoring the CLI client to use the (actually maintained) stream builder internally
- updating the python client

### DIFF
--- a/crates/tycho-client/src/cli.rs
+++ b/crates/tycho-client/src/cli.rs
@@ -1,20 +1,11 @@
-use std::{collections::HashSet, path::Path, str::FromStr, time::Duration};
+use std::{path::Path, str::FromStr};
 
 use clap::Parser;
-use tracing::{debug, error, info, warn};
+use tracing::{error, info};
 use tracing_appender::rolling;
-use tycho_common::dto::{Chain, ExtractorIdentity};
+use tycho_common::dto::Chain;
 
-use crate::{
-    deltas::DeltasClient,
-    feed::{
-        component_tracker::ComponentFilter, synchronizer::ProtocolStateSynchronizer,
-        BlockSynchronizer,
-    },
-    rpc::HttpRPCClientOptions,
-    stream::ProtocolSystemsInfo,
-    HttpRPCClient, WsDeltasClient,
-};
+use crate::{feed::component_tracker::ComponentFilter, stream::TychoStreamBuilder};
 
 /// Tycho Client CLI - A tool for indexing and tracking blockchain protocol data
 ///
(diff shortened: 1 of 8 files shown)


crates/tycho-client/src/cli.rs
```diff
@@ -231,91 +224,66 @@ async fn run(exchanges: Vec<(String, Option<String>)>, args: CliArgs) -> Result<
         None => Vec::new(),
     };
 
-    info!("Running with version: {}", option_env!("CARGO_PKG_VERSION").unwrap_or("unknown"));
-    //TODO: remove "or args.auth_key.is_none()" when our internal client use the no_tls flag
```
zizou0x, line 235: Was this addressed in the end?

tycho-client/src/cli.rs
```diff
@@ -231,91 +224,66 @@ async fn run(exchanges: Vec<(String, Option<String>)>, args: CliArgs) -> Result<
...
+        None => builder,
+    };
+    let builder = match args.timeout {
+        Some(to) => builder.timeout(to),
+        None => builder,
+    };
+    let builder = match args.max_missed_blocks {
+        Some(mmb) => builder.max_missed_blocks(mmb),
+        None => builder,
+    };
```
zizou0x: Not strong opinion but I find this better

```
if let Some(bt) = args.block_time {
    builder = builder.block_time(bt);
}
if let Some(to) = args.timeout {
    builder = builder.timeout(to);
}
if let Some(mmb) = args.max_missed_blocks {
    builder = builder.max_missed_blocks(mmb);
}
```

crates/tycho-client/src/cli.rs
louise-poole: Yeah, I changed it to be independent of each other: no_tls is independent of if you set an auth key or not. Defaults to false unless you explicitly set it

tycho-client/src/cli.rs
```diff
@@ -231,91 +224,64 @@ async fn run(exchanges: Vec<(String, Option<String>)>, args: CliArgs) -> Result<
...
-        .await
-        .map_err(|e| format!("WebSocket client connection error: {e}"))?;
 
-    let mut block_sync = BlockSynchronizer::new(
-        Duration::from_secs(args.block_time),
-        Duration::from_secs(args.timeout),
-        args.max_missed_blocks,
-    );
+    // Start with chain-appropriate defaults; override only what the user explicitly provided.
+    let builder = TychoStreamBuilder::new(&args.tycho_url, chain);
```
zizou0x: I still find the builder logic weird, we could apply all the unconditional things here and then all the conditional ones below one by one no? I see now that it's separated by sections, not sure this adds value

zizou0x (approved):
Still a tiny comment but not blocking

PR #909, opened 2026-04-16
feat: add self-hosted Firehose profile to docker-compose

kayibal:
## Summary

- Adds a `substreams-endpoint` opt-in profile to `docker-compose.yaml`: a single-container Firehose stack (reader-node-stdin, merger, relayer, firehose, substreams-tier1, substreams-tier2) backed by any EVM JSON-RPC node via `fireeth tools poller generic-evm`
- Auto-resume on restart: poller detects the highest stored one-block file and continues from there, avoiding re-fetching already-fetched blocks
- Moves lgtm/promtail observability to an opt-in `observability` profile (was always-on)
- Fixes tycho-indexer healthcheck to use `bash /dev/tcp` (curl is not installed in the image)
- Fixes OTLP exporter default to empty string (disables exporter when lgtm is not running)
- Makes `SUBSTREAMS_ENDPOINT`, `CHAINS`, and `RETENTION_HORIZON` configurable via env vars; removes hardcoded Ethereum/date defaults
- Removes chain-specific `extractors-tempo.yaml` volume mount
- Deletes `docker-compose.firehose.yaml` (superseded by the new profile)

Tracked in ENG-5744.

## Test plan

- [ ] `docker compose --profile substreams-endpoint up` starts all services healthy
- [ ] Restarting `substreams-endpoint` resumes poller from last stored block (not START_BLOCK)
- [ ] Restarting `tycho-indexer` resumes from cursor
- [ ] Full `docker compose down` + `up` (without `-v`) resumes both correctly
- [ ] `docker compose up` (without profile) starts only db + tycho-indexer (lgtm/promtail not started)

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -16,8 +16,10 @@ services:
       - postgres_data:/bitnami/postgresql
     shm_size: '2gb'
   
-  # Grafana LGTM stack for observability (Loki, Grafana, Tempo, Mimir)
+  # Grafana LGTM stack for observability (Loki, Grafana, Tempo, Mimir) — opt-in, resource-heavy
+  # Enable with: docker compose --profile observability up
   lgtm:
+    profiles: [observability]
     image: grafana/otel-lgtm:latest
     restart: "always"
     ports:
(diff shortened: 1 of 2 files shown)


zizou0x (approved):
LGTM

PR #918, opened 2026-04-20
fix: clone and other chain protocol tests CI running

tamaralipows:
migrated PR: https://github.com/propeller-heads/tycho-protocol-sdk/pull/423 + some other fixes


protocols/testing/entrypoint.sh
```diff
@@ -28,14 +28,44 @@ if [ "${#errors[@]}" -ne 0 ]; then
...
+infer_chain() {
+    local protocol="$1"
+    case "$protocol" in
+        base-*)     echo "base" ;;
+        arbitrum-*) echo "arbitrum" ;;
+        unichain-*) echo "unichain" ;;
+        bsc-*)      echo "bsc" ;;
+        *)          echo "ethereum" ;;
+    esac
+}
```
louise-poole, line 42: Should we add polygon? There should be some protocol config files for polygon already

louise-poole (approved):

protocols/testing/entrypoint.sh
tamaralipows: by protocol config do you mean integration test files @louise-poole? can't find any - but I added support for the future anyway

louise-poole: No, I meant substream config files. True, I don't think any test files were added.... 🤔 
Okay, cool. Thanks! 🙏

PR #919, opened 2026-04-20
feat: Add Fluid V1 DEX indexing and simulation

kayibal:
## Summary

Ports [propeller-heads/tycho-protocol-sdk#412](https://github.com/propeller-heads/tycho-protocol-sdk/pull/412) into the monorepo.

- Adds `protocols/substreams/ethereum-fluid`: Substreams indexer for the Fluid V1 DEX, including ABI bindings, event handling, and integration test config
- Adds `protocols/adapter-integration/evm/src/fluid/FluidSwapExecutor.sol`: EVM executor contract with Fluid's DEX callback swap interface
- Registers the `fluid_v1` native decoder in `protocols/testing` and adds the FluidSwapExecutor bytecode fixture

## Test plan

- [ ] Substreams integration test: `integration_test.tycho.yaml` in `protocols/substreams/ethereum-fluid/`
- [ ] EVM executor test: `protocols/adapter-integration/evm/test/FluidSwapExecutor.t.sol`
- [ ] Protocol testing: `fluid_v1` decoder registration compiles cleanly

Closes propeller-heads/tycho-protocol-sdk#412

PR #923, opened 2026-04-20
fix(ci): correct substreams package detection path prefix

kayibal:
## Summary

- Fixes `substreams-check` action grep pattern from `^substreams/` to `^protocols/substreams/` — `tj-actions/changed-files` returns repo-root-relative paths, so the old pattern never matched anything
- Fixes the directory-walk loop termination condition from `"$dir" != "substreams"` to `"$dir" != "protocols/substreams"` for the same reason

This bug caused the Substreams CI (lint + test) to silently output "No packages to check" / "No packages to test" on every PR, including #919 which added `ethereum-fluid`.

## Test plan

- [ ] Push a change under `protocols/substreams/<any-package>/` and confirm the Substreams CI lint and test steps detect it and run against the correct package name

### DIFF
--- a/.github/actions/substreams-check/action.yml
+++ b/.github/actions/substreams-check/action.yml
@@ -19,7 +19,7 @@ runs:
       shell: bash
       run: |
         # Convert space-separated list to newline-separated list and process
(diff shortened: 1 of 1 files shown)


kayibal:
Duplicate of #914 which is already merged. PR #919 needs a rebase onto main to pick up the fix.

PR #925, opened 2026-04-20
fix: build tycho-indexer from monorepo source instead of GitHub clone 

tamaralipows:


.github/workflows/ci-substreams-integration.yaml
```diff
@@ -137,6 +138,7 @@ jobs:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
         with:
           persist-credentials: false
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
```
tamaralipows, line 141: This was already here in the tycho-protocol-sdk repo: https://github.com/propeller-heads/tycho-protocol-sdk/blob/46c63b915de52449b92030ce868dfd0eaeafbbb9/.github/workflows/substreams.tests.yaml#L153

PR #931, opened 2026-04-21
fix(integration-test): Fix block processing latency metric

louise-poole:
The tycho latency metric was firing false alerts (~12.4s median, threshold 10s) due to two issues in how protocol updates were processed.                                     
                                                                                                                                                                                                                      
 ### Root causes                                                                                                                                                                                                         
                                                                                                                                                                                                                      
  1. Wrong block used for latency measurement. When Tycho delivered an update for block N+1 before the RPC had confirmed it, the update was skipped — but the latency was already recorded against block N's timestamp (~12s older), inflating every reading by roughly one Ethereum block time.                                                                                                                                          
  2. RFQ processing starved protocol update consumption. Slow external API calls (Bebop/Hashflow, up to 15s each) shared the parallel_updates semaphore with protocol updates. When all permits were held by RFQ tasks, protocol updates queued up with stale received_at timestamps, further skewing the metric.                                                                                                                    
   
 ### Changes                                                                                                                                                                                                             
                                                            
  - RPC polling: when Tycho is ahead of the RPC, poll getBlockByNumber up to 10 times (500ms interval) until the RPC catches up, then proceed normally. The latency metric is now only recorded once both sides are on the same block.
  - Channel separation: split the single shared channel and semaphore into separate pairs for protocol and RFQ updates, so slow RFQ API calls cannot block protocol update consumption.

### DIFF
--- a/.github/workflows/cd-deploy-dev.yaml
+++ b/.github/workflows/cd-deploy-dev.yaml
@@ -40,7 +40,7 @@ permissions:
 jobs:
   # ── tycho-indexer ──────────────────────────────────────────────────────────
   build-and-push-tycho-indexer:
-    if: inputs.application == 'tycho-indexer'
+    if: ${{ inputs.application == 'tycho-indexer' }}
     uses: propeller-heads/ci-cd-templates/.github/workflows/build-and-push-docker-image.yaml@main
     permissions:
       id-token: write
@@ -59,7 +59,7 @@ jobs:
 
   deploy-dev-tycho-indexer:
     needs: build-and-push-tycho-indexer
-    if: inputs.application == 'tycho-indexer'
+    if: ${{ inputs.application == 'tycho-indexer' }}
     uses: propeller-heads/ci-cd-templates/.github/workflows/deploy-any-branch-dev.yaml@main
     permissions:
       id-token: write
@@ -77,7 +77,7 @@ jobs:
 
   # ── tycho-integration-test ─────────────────────────────────────────────────
   build-and-push-tycho-integration-test:
-    if: inputs.application == 'tycho-integration-test'
+    if: ${{ inputs.application == 'tycho-integration-test' }}
     uses: propeller-heads/ci-cd-templates/.github/workflows/build-and-push-docker-image.yaml@main
     permissions:
       id-token: write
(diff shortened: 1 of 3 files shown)


louise-poole:
@claude review once

.github/workflows/cd-deploy-dev.yaml
```diff
@@ -40,7 +40,7 @@ permissions:
 jobs:
   # ── tycho-indexer ──────────────────────────────────────────────────────────
   build-and-push-tycho-indexer:
-    if: inputs.application == 'tycho-indexer'
+    if: ${{ inputs.application == 'tycho-indexer' }}
```
zizou0x, line 43: Was this change needed? I think it's minor, but I wonder if there is a reason to do it

crates/tycho-integration-test/src/main.rs
```diff
@@ -490,30 +580,32 @@ async fn process_update(
...
         }
-        // Record block processing latency
-        let latency_seconds =
-            (update.received_at.as_secs_f64() - block.header.timestamp as f64).abs();
-        metrics::record_block_processing_duration(latency_seconds);
 
-        let rpc_block_number = block.header.number;
+        // Poll RPC until it reaches the update's block number
+        const MAX_POLL_ATTEMPTS: u32 = 10;
+        const POLL_INTERVAL: Duration = Duration::from_millis(500);
```
zizou0x: Would be nice to have them configurable, the values will probably depend on the chain we're running on

zizou0x (commented):
Looks good! Just one small comment

zizou0x (approved):

dianacarvalho1 (approved):
Thank you for having a look at this @louise-poole ! 🙏🏼 I agree with Flo's comments. I wonder why only now we started having issues with this though..

.github/workflows/cd-deploy-dev.yaml
louise-poole: Yes - I was trying to trigger the build and deploy workflow and if I selected only the integration test, it was ignoring our checks and deploying everything on every chain. After I added these it would target only the deployment I selected. Apparently it's a caveat of it being a reusable workflow? Honestly I'm not sure why this worked before and not now... Try run build-and-deploy on main for only the integration test on unichain and see what I mean.

PR #937, opened 2026-04-23
feat: add pipeline to run cargo update periodically

Troshchk:
- Add a scheduled GitHub Actions workflow that runs `cargo update` across all workspace lockfiles biweekly (1st and 15th of each month)                                                                    
  - Include a safety check: a script queries crates.io to verify every updated crate was published at least 3 days ago. Too recent crates are pinned to the latest safe version instead of being reverted to the old lockfile version                                                                                                                                                                      
  - The workflow auto-creates a PR scoped to only `Cargo.lock` file changes                                                                                                                                
  - Exclude `Cargo.lock` from the substreams CI trigger path to avoid unnecessary CI runs on dependency-only updates

### DIFF
--- a/.github/scripts/check-crate-ages.sh
+++ b/.github/scripts/check-crate-ages.sh
@@ -1,0 +1,108 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Compares old and new Cargo.lock files, extracts updated crates,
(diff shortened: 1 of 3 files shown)


dianacarvalho1 (approved):
ooohhh this is reaally niceee! Thank you 🙏🏼 🙏🏼

tamaralipows (approved):
Thank you!

PR #943, opened 2026-04-23
fix: Don't allow Curve tokens to be zero address

tamaralipows:
In the CurveExecutor , tokenIn is turned into address(0) if it matches nativeToken. Same for tokenOut. However, both tokenIn and tokenOut can also be set to address(0) directly. This could have produced the situation where tokenIn = address(0) and transferType = ProtocolWillDebit.


Troshchk (approved):
Looks good, thank you!

PR #946, opened 2026-04-23
feat: add limit on number of tries while searching token slots

Troshchk:


crates/tycho-ethereum/src/services/entrypoint_tracer/slot_detector.rs
```diff
@@ -33,12 +33,16 @@ type SlotValues = Vec<((Address, Bytes), U256)>;
...
+/// Maximum number of slot test retries per token before giving up.
+const MAX_SLOT_TEST_RETRIES: u8 = 5;
+
 #[derive(Debug, Clone)]
 pub(crate) struct SlotMetadata {
     token: Address,
     original_value: U256,
     test_value: U256,
     all_slots: SlotValues,
+    retries: u8,
```
dianacarvalho1: do we need to add this to the metada? 🤔

dianacarvalho1: I'm not super familiar with this code but I'm wondering if this is really needed

dianacarvalho1 (commented):
Thank you! Tiny comment

crates/tycho-ethereum/src/services/entrypoint_tracer/slot_detector.rs
Troshchk: As an alternative, `metadata.all_slots` can be truncated to only contain `MAX_SLOT_TEST_RETRIES` number of slots. In this case there is no need to add retries to metadata. Do you think that would be better?

tamaralipows: I also don't remember super well - this can't just live in the code?

Edit: sorry didn't see your reply when I sent this

I think that would be fine - we aren't logging all the slots anywhere else right? There is no reason we would need all?

Troshchk: Ok, will make a change 👍

dianacarvalho1 (approved):

tamaralipows (approved):
Thank you!

PR #953, opened 2026-04-27
feat(indexer): add granular DCI cache metrics

zizou0x:
Break down the single dci_cache_size metric into per-sub-cache gauges to enable quick identification of which cache component and which tracked contract is responsible for memory spikes.

New metrics (emitted every 60s alongside existing dci_cache_size):
- dci_cache_subcache_size{subcache} — deep_size_of per sub-cache
- dci_cache_subcache_key_count{subcache} — unique key count per sub-cache
- dci_cache_pending_layers — pending block layer depth
- dci_cache_tracked_contracts_max_slots — slot count of largest contract

New structured info log with top 10 tracked contracts by slot count, including addresses for direct investigation and blacklisting.

### DIFF
--- a/crates/tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
+++ b/crates/tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
@@ -303,10 +303,10 @@ impl DCICache {
                 .pending_layer_count() as f64,
         );
 
-        let snapshot = self.tracked_contracts.iter_latest();
-        let mut by_slots: Vec<_> = snapshot
-            .iter()
-            .map(|(addr, slots)| (*addr, slots.len()))
+        let mut by_slots: Vec<_> = self
+            .tracked_contracts
+            .iter_latest()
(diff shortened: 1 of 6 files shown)


tvinagre (approved):

crates/tycho-indexer/src/extractor/dynamic_contract_indexer/cache.rs
```diff
@@ -435,6 +582,23 @@ where
...
     }
 
+    /// Returns the number of pending block layers.
+    pub(super) fn pending_layer_count(&self) -> usize {
+        self.pending.len()
+    }
+
+    /// Returns a snapshot of all unique keys with their latest value.
+    /// Pending layers take precedence over the permanent layer.
+    pub(super) fn iter_latest(&self) -> HashMap<&K, &V> {
```
louise-poole: If it returns a snapshot and not an iterator, why is it called 'iter_'? 👀

louise-poole (approved):
One small comment, otherwise approved!

PR #957, opened 2026-04-28
fix: handle short byte slices in ResponseAccount B256 conversion

tvinagre:
## Summary
- Remove dead fields (`balance_modify_tx`, `code_modify_tx`, `creation_tx`, `code_hash`) from simulation-side `ResponseAccount` — none were read by any consumer
- Convert `From` impls to `TryFrom` with `ConversionError` for `ResponseAccount` and `AccountUpdate`, eliminating all panicking `from_slice` calls
- Remove the now-unused `bytes_to_b256` helper from `u256_num.rs`

## Root cause

When a new contract is discovered by the indexer (e.g. via DCI) but its block hasn't finalized yet, the account only exists in the `PendingDeltasBuffer` — not in the DB. If a client queries `/contract_state` for that address during this window, `PendingDeltasBuffer::get_account()` builds the account entirely from pending deltas via `AccountDelta::into_account_without_tx()`. That method sets `balance_modify_tx` and `code_modify_tx` to `Bytes::from("0x00")` (1 byte) because no real tx hash is available — the tx context is dropped during the `BlockChanges → BlockAggregatedChanges` aggregation step.

When the client deserializes this response and converts `dto::ResponseAccount` into `tycho_simulation::ResponseAccount`, `B256::from_slice` expected exactly 32 bytes and panicked on the 1-byte sentinel.

The fix removes these unused fields entirely from the simulation struct and replaces all `From` impls with `TryFrom` to prevent panics from short byte slices.

## Test plan
- [x] Unit test: valid 20-byte address converts successfully
- [x] Unit test: short address returns `ConversionError`
- [x] All existing simulation tests pass (27 failures are pre-existing env-dependent tests)

### DIFF
--- a/crates/tycho-simulation/src/evm/decoder.rs
+++ b/crates/tycho-simulation/src/evm/decoder.rs
@@ -336,7 +336,10 @@ where
                     .get_vm_storage()
                     .iter()
                 {
-                    let account: ResponseAccount = value.clone().into();
+                    let account: ResponseAccount = value
+                        .clone()
+                        .try_into()
+                        .map_err(|e| StreamDecodeError::Fatal(format!("{e}")))?;
 
                     if state_guard.tokens.contains_key(key) {
                         let original_address = account.address;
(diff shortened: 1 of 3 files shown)


zizou0x (approved):
Nice, thanks! LGTM 🚀

PR #958, opened 2026-04-28
refactor: change getTransferData modifiers

Troshchk:


crates/tycho-execution/contracts/interfaces/IExecutor.sol
```diff
@@ -40,7 +40,7 @@ interface IExecutor {
      */
     function getTransferData(bytes calldata data)
         external
-        payable
+        view
```
dianacarvalho1, line 43: why not pure?

dianacarvalho1 (commented):
Thank you @Troshchk ! Can you just explain why are some pure and others view please?

crates/tycho-execution/contracts/interfaces/IExecutor.sol
Troshchk: `pure` only decodes calldata, while `view` need to read contract variables.

Some executors need contract variables access(`view`), so IExecutor.getTransferData is `view` for compatibility.
`pure` implementations are still valid since they’re a subset of view.

tamaralipows (approved):
Thank you!

dianacarvalho1 (approved):
Thank you

PR #962, opened 2026-04-29
fix(indexer): DCI cache unbounded growth from retry params

zizou0x:
## Summary

- **Retry counter fix**: `pending_entry()` only accessed the current block's layer, so `or_insert(0)` fired every block and the counter never incremented across blocks. Retries never stopped. Now reads the current count via `get()` across all layers before writing the incremented value.
- **Retry re-broadcast fix**: The single processing loop wrote retry params into `component_id_to_entrypoint_params` on every block, creating duplicate entries in new pending layers. Split into two passes: new params update cache + build trace list, retry params only build the trace list.

## Metrics before

Capture d’écran 2026-04-30 à 11 00 19
Capture d’écran 2026-04-30 à 11 01 38

## Metrics after
Capture d’écran 2026-04-30 à 10 59 53
Capture d’écran 2026-04-30 à 11 01 28


tvinagre (approved):
Amazing!

PR #964, opened 2026-04-29
fix(RouterV3 model): Curve executor token decoding

tamaralipows:
We were reassigning token_in from CurveNative to Zero, then in the next line checking the already-overwritten value - so TransferNativeInExecutor was never selected.

The warning is gone, and no new suspicious states found.


Troshchk (approved):
Nice catch! Thank you @tamaralipows!

dianacarvalho1 (approved):
🙏🏼

PR #967, opened 2026-04-30
fix(tycho-client): improve WS connection handling

louise-poole:
Improvements include:
  - **WS subscribe/unsubscribe timeouts**: Without timeouts, a server that stops confirming WS commands caused cleanup to block forever, leaving stale processes alive and accumulating WS connections.                                                                                                                                                                                                        
  - **ensure_connection TOCTOU race**: Used Notify::enable() to close the window where a reconnect notification could fire before notified().await, and added a post-wake is_connected() check against spurious wakes.
  - **Stale snapshot blocks**: Added RPCError::StaleBlock so the 'init loop can discard a too-old delta and wait for a fresher block on the same subscription rather than tearing it down and retrying from scratch.      
  - **Unknown extractor no longer crashes**: A misconfigured protocol name now logs a warning and skips that extractor, leaving valid protocols running.                                                                  
  - **Retry count resets after a healthy run**: After processing at least one block, a subsequent failure resets the retry counter so transient errors don't consume quota earned by prior failures.  
  - **Detect stalled connections**: trigger reconnect attempts after a period of silence on the ws connection.

### DIFF
--- a/crates/tycho-client/src/deltas.rs
+++ b/crates/tycho-client/src/deltas.rs
@@ -914,16 +914,31 @@ impl DeltasClient for WsDeltasClient {
                 this.conn_notify.notify_waiters();
                 result = Ok(());
 
+                // If no WS frame arrives within this window the TCP connection is
+                // considered stalled (e.g. network cut without a TCP RST/FIN). The OS
+                // default keepalive fires after ~2 hours; this gives us an
+                // application-level detection that is orders of magnitude faster.
+                const IDLE_TIMEOUT: Duration = Duration::from_secs(60);
                 loop {
                     let res = tokio::select! {
-                        msg = msg_rx.next() => match msg {
-                            Some(msg) => this.handle_msg(msg).await,
-                            None => {
-                                // This code should not be reachable since the stream
-                                // should return ConnectionClosed in the case above
-                                // before it returns None here.
-                                warn!("Websocket connection silently closed, giving up!");
-                                break 'retry
+                        msg_result = tokio::time::timeout(IDLE_TIMEOUT, msg_rx.next()) => {
+                            match msg_result {
+                                Err(_elapsed) => {
+                                    warn!("No WS frame received for {IDLE_TIMEOUT:?}, \
+                                           treating connection as stalled; Reconnecting...");
+                                    retry_count += 1;
+                                    let mut guard = this.inner.as_ref().lock().await;
+                                    *guard = None;
+                                    break; // break inner loop → reconnect
+                                }
+                                Ok(Some(msg)) => this.handle_msg(msg).await,
+                                Ok(None) => {
+                                    // This code should not be reachable since the stream
+                                    // should return ConnectionClosed in the case above
+                                    // before it returns None here.
+                                    warn!("Websocket connection silently closed, giving up!");
+                                    break 'retry
+                                }
                             }
                         },
                         _ = cmd_rx.recv() => {break 'retry},
--- a/crates/tycho-client/src/deltas.rs
+++ b/crates/tycho-client/src/deltas.rs
@@ -504,6 +504,9 @@ impl WsDeltasClient {
         if self.dead.load(Ordering::SeqCst) {
             return Err(DeltasError::NotConnected);
         }
+        if !self.is_connected().await {
+            return Err(DeltasError::NotConnected);
+        }
         Ok(())
     }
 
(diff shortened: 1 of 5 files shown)


crates/tycho-client/src/deltas.rs
```diff
@@ -486,11 +486,24 @@ impl WsDeltasClient {
...
+        tokio::pin!(notified);
+        notified.as_mut().enable();
         if !self.is_connected().await {
-            self.conn_notify.notified().await;
-        };
+            notified.await;
+        }
+        if self.dead.load(Ordering::SeqCst) {
+            return Err(DeltasError::NotConnected);
+        }
```
kayibal, line 506: Do you properly understand this change? It seems kind of weird to me that we need to check twice basically.

crates/tycho-client/src/rpc.rs
```diff
@@ -139,6 +143,23 @@ pub enum RPCError {
...
+/// Converts an HTTP response body parse failure into the correct `RPCError`.
+///
+/// The tycho server returns plain-text error messages (not JSON) when a requested block falls
+/// outside its retention window. Detecting these here gives callers a typed signal to retry
+/// with a more recent block rather than treating it as an unrecoverable parse failure.
+///
+/// NOTE: The string matching below is coupled to the server's error message text. If those
+/// messages change server-side this silently regresses to `ParseResponse`. Replace with a
+/// structured error code if the server ever returns typed error responses.
+fn parse_error(err: serde_json::Error, body: &str) -> RPCError {
```
kayibal: Shouldn't this be moved as a method on the error enum instead? Seems unnecessary toi be a free function.

crates/tycho-client/src/deltas.rs
```diff
@@ -776,9 +776,18 @@ impl DeltasClient for WsDeltasClient {
                 .await?;
         }
         trace!("Waiting for subscription response");
-        let res = ready_rx.await.map_err(|_| {
-            DeltasError::TransportError("Subscription channel closed unexpectedly".to_string())
-        })??;
+        let res = tokio::time::timeout(Duration::from_secs(30), ready_rx)
```
zizou0x, line 799: In the description it says: 

> if the server accepts the WS command but never sends back
NewSubscription or SubscriptionEnded

Do you know how that can happen? Seems like this something we should also treat better on the indexer side, like make sure we return `SubscriptionEnded`

crates/tycho-client/src/rpc.rs
zizou0x: We should introduce error codes on the server side to make this more future proof

crates/tycho-client/src/feed/synchronizer.rs
```diff
@@ -800,7 +803,13 @@ where
                     }
                 }
                 sleep(self.retry_cooldown).await;
-                retry_count += 1;
+                // A run that processed blocks is a healthy run — reset the counter so
+                // transient failures after a long successful period get a fresh retry budget.
+                if made_progress {
+                    retry_count = 0;
```
zizou0x, line 815: Can you add a debug log when reseting the retry count please

zizou0x (commented):
Nice! Just have a few comments/questions but thanks for this one!

crates/tycho-client/src/deltas.rs
louise-poole: The changes here are 2-fold: 
- `notified.as_mut().enable();` was added to avoid any chance of missing a notify signal between creating the future and awaiting it (tokio's recommended best practice).
- For the two `is_connected` checks - this is an optimisation. We early exit if it's already connected and avoid notify futures entirely. The second is_connected check is to ensure we did not quietly connect while setting up the notify future because if we did we would await indefinitely here.

It's a minor optimisation... I'm happy to either better comment it for future understanding, or remove it if you feel it's not beneficial enough?

crates/tycho-client/src/feed/synchronizer.rs
louise-poole: Think it's necessary? Each retry already logs the count - it's pretty obvious when it resets or not from those logs alone.

crates/tycho-client/src/deltas.rs
louise-poole: This is mainly a fallback catch (should never really occur). It would happen if the WS actor ends while the subscribe request is in flight. We should hit a 'channel closed unexpectedly' error before the timeout, but if we don't then the timeout prevents us from waiting for a potentially hanging connection to complete.

crates/tycho-client/src/feed/synchronizer.rs
zizou0x: Ok if we log it for each retry it's all good 👍

zizou0x (approved):
LGTM

PR #968, opened 2026-04-30
feat: add gas_error_ratio metric histogram for integration tests

Troshchk:
- add gas_error_ratio prometheus hist to tycho integration tests, measuring `|estimated_gas - actual_gas| / actual_gas` per protocol
- emit the gas_error_ratio metric
- buckets: dense up to 10% threshold (2% steps), then a few for outliers [0.0, 0.02, 0.04, 0.06, 0.08, 0.1, 0.25, 0.5, 1.0]

### DIFF
--- a/crates/tycho-integration-test/src/main.rs
+++ b/crates/tycho-integration-test/src/main.rs
@@ -1200,6 +1200,7 @@ async fn process_state(
             }
         };
         let duration_seconds = start_time.elapsed().as_secs_f64();
+        let estimated_gas = amount_out_result.gas;
         let expected_amount_out = amount_out_result.amount;
         debug!(
             event_type = "get_amount_out_duration",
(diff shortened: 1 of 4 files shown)


crates/tycho-integration-test/src/main.rs
```diff
@@ -1363,6 +1365,14 @@ fn process_execution_result(
 
             metrics::record_execution_slippage(&execution_info.protocol_system, slippage);
 
+            if let Some(estimated) = execution_info.estimated_gas.to_f64() {
```
dianacarvalho1, line 1368: it needs to be f64 for prometheus?

dianacarvalho1 (approved):
On a first look this seems good to me, thank you! 🙏🏼 However, the dashboard is quite finicky sometimes 😬 I would recommend you to check how some of the current histograms are being done now (and cringe 😐)!

crates/tycho-integration-test/src/main.rs
Troshchk: yes

tamaralipows (approved):
Thank you!  🙏🏼 🤞

PR #969, opened 2026-04-30
feat(router v3 model): Add missing executors

dianacarvalho1:
Adding AerodromeV1 and LiquidityParty. They are very similar to Univ2 but are added here anyway for completeness

Nothing sus here after running:
```
## Suspicious count by executor combination

- [UniswapV3]: 0
- [FluidV1]: 0
- [LiquidityParty]: 0
- [UniswapV2]: 0
- [ERC4626]: 0
- [AerodromeV1]: 0
- [Slipstreams]: 0
- [Curve]: 0
- [Weth]: 0
- [MaverickV2]: 0
```

### DIFF
--- a/crates/tycho-execution/model/src/model/executors.rs
+++ b/crates/tycho-execution/model/src/model/executors.rs
@@ -29,7 +29,9 @@ use crate::{
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, Serialize, PartialOrd, Ord)]
 pub enum Executor {
-    // TODO this list is incomplete.
+    // Only executors that give the caller control over the called pool contract were modeled.
+    // When a new executor that fulfills these criteria is added, it needs to be modelled here
+    // too.
     Curve,
     ERC4626,
     FluidV1,
@@ -38,6 +40,8 @@ pub enum Executor {
     UniswapV2,
     UniswapV3,
     Weth,
+    AerodromeV1,
+    LiquidityParty,
 }
 
 /// Return value of [Executor::get_transfer_data]
@@ -58,7 +62,7 @@ pub struct CallbackTransferData {
 
 impl Executor {
     /// Array containing all [Executor]s.
-    pub const VARIANTS: [Executor; 8] = [
+    pub const VARIANTS: [Executor; 10] = [
         Executor::Curve,
         Executor::ERC4626,
         Executor::FluidV1,
@@ -67,6 +71,8 @@ impl Executor {
         Executor::UniswapV2,
         Executor::UniswapV3,
         Executor::Weth,
+        Executor::AerodromeV1,
+        Executor::LiquidityParty,
     ];
 
     /// <https://github.com/propeller-heads/tycho-execution/blob/9b0512c9580617224c7a0d7de781674a2cdc6b62/foundry/interfaces/IExecutor.sol#L41>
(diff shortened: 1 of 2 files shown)


crates/tycho-execution/model/src/model/executors.rs
```diff
@@ -29,7 +29,9 @@ use crate::{
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, Serialize, PartialOrd, Ord)]
 pub enum Executor {
-    // TODO this list is incomplete.
+    // Only executors that give the caller control over the called pool contract were modeled.
+    // When a new executor that fulfills these criteria is added, it needs to be modelled here
+    // too.
```
tamaralipows, line 34: We should add this to our integration docs then right? Is this the integrator's responsibility?

crates/tycho-execution/model/src/model/executors.rs
```diff
@@ -263,6 +269,46 @@ impl Executor {
...
                 })
             }
+            // https://github.com/propeller-heads/tycho-indexer/blob/0d9b01ddbe72c5518fdc79a423ffd19dc7226709/crates/tycho-execution/contracts/src/executors/AerodromeV1Executor.sol#L80
+            Self::AerodromeV1 => Ok(TransferData {
+                transfer_type: TransferType::Transfer,
+                receiver: params.request(
+                    ParamKey::ProtocolData { swap_index, start: 0, end: 20 },
+                    // trying more variants might find some very obscure bugs
+                    // in the future but slows down simulation a lot
+                    // and currently is ignored anyway
```
tamaralipows, line 279: What does this mean?

crates/tycho-execution/model/src/model/executors.rs
```diff
@@ -484,6 +530,44 @@ impl Executor {
...
+                    Address::SENDER_CONTROLLED,
+                )?;
+                if pool.is_sender_controlled() {
+                    // if the sender controls the pool,
+                    // the actual swap logic doesn't matter
+                    Ok(())
+                } else {
+                    Err(Error::Ignore {
+                        reason: "aerodrome v1 pool not sender controlled. not low hanging fruit. would require simulating real pool".into(),
```
tamaralipows, line 548: not low hanging fruit?

tamaralipows (commented):
Thank you - I think this will make a lot more sense to me after I spend more time inspecting the executor code so I'm holding off my approval for a bit. I just have a couple questions

crates/tycho-execution/model/src/model/executors.rs
dianacarvalho1: oh good question.. not sure. actually I think we should keep control of this? 🤔 I think we need to rethink the whole "adding new executors security process"

crates/tycho-execution/model/src/model/executors.rs
dianacarvalho1: i copied this from the other executors 😅

crates/tycho-execution/model/src/model/executors.rs
dianacarvalho1: same here

crates/tycho-execution/model/src/model/executors.rs
dianacarvalho1: I think it means that Max didn't think it made sense to try more combinations here without falling into deep deep deep holes

crates/tycho-execution/model/src/model/executors.rs
dianacarvalho1: I'm also confused a bit about this design tbh. because at some other point we are defining the pool to be sender controlled.. so I'm not sure we even need this here

crates/tycho-execution/model/src/model/executors.rs
tamaralipows: very true

crates/tycho-execution/model/src/model/executors.rs
tamaralipows: Okay - will double check this later!

tamaralipows (approved):
Thank you! Went through all the executors and compared to our solidity code - I think what you did makes sense! (p.s. nice branch name)

PR #970, opened 2026-04-30
feat(router v3 model): Consider wasted router funds

tamaralipows:
We were only considering if the sender controlled accounts managed to steal funds. Now, we also consider if the router loses any vault balances that do not belong to the sender, even if the sender does not receive these funds (fund wastage).

Related fix:
We weren't actually making the router the balance owner of the eth that was sent. This was affecting our is_suspicious calculation for a nonzero msg.value.


dianacarvalho1 (approved):
Thank you @tamaralipows 🙏🏼 
I'm a bit confused that ETH and WETH are used interchangeably almost.. This means that if we add another token address to the Address enum we wouldn't catch things here? Or am I misunderstanding something? 🤔

PR #971, opened 2026-05-01
fix: update v3-style swap tick after partial price move

zach030:
## Summary

Fixes stale tick updates in V3-style swap simulations.

The swap loop was storing `sqrt_price_start` after updating `state.sqrt_price`, so partial price moves that did not cross an initialized tick failed to refresh `state.tick`. This could return an inconsistent post-swap state where `sqrt_price` had moved but `tick` was stale.

This applies the fix to Aerodrome Slipstreams, Velodrome Slipstreams, Uniswap V3, and Uniswap V4 simulations.

## Testing

- Added regression tests for partial price moves without crossing initialized ticks.
- Added boundary tests to ensure tick is not recomputed when price does not move.


zizou0x (approved):
LGTM, thanks!

PR #975, opened 2026-05-01
feat(integration test): Consider input token gas

tamaralipows:
[Task](https://propeller-heads.atlassian.net/browse/ENG-5934) (also left some questions there)

Please see this thread for reasoning: https://propellerswap.slack.com/archives/C05LQALE9L1/p1777952456968939

### DIFF
--- a/crates/tycho-integration-test/src/main.rs
+++ b/crates/tycho-integration-test/src/main.rs
@@ -1200,7 +1200,9 @@ async fn process_state(
             }
         };
         let duration_seconds = start_time.elapsed().as_secs_f64();
-        let estimated_gas = amount_out_result.gas;
+        let transfer_overhead =
+            estimate_transfer_overhead(&component.protocol_system, token_in, token_out);
+        let estimated_gas = amount_out_result.gas + transfer_overhead;
         let expected_amount_out = amount_out_result.amount;
         debug!(
             event_type = "get_amount_out_duration",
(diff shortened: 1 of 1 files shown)


crates/tycho-integration-test/src/main.rs
```diff
@@ -1488,6 +1490,54 @@ fn categorize_error(error_message: &str) -> &'static str {
     }
 }
 
+/// Gas cost for a native ETH value transfer into the router.
+const NATIVE_TRANSFER_GAS: u64 = 20_000;
+
+/// Gas cost for an ERC-20 `transferFrom`.
+const TOKEN_TRANSFER_GAS: u64 = 60_000;
```
dianacarvalho1: I think we can get these values from `Token.gas`. Sorry I missed adding this in the task!

dianacarvalho1: I think we still need a default for Tokens that don't have that though!

dianacarvalho1 (commented):
Thank you @tamaralipows ! 

> The transfer from the router into the first pool is already considered in tycho-simulation's gas_cost()

Can you tell me where this is done please? I only found `gas_costs` (plural) and this only exists for Ekubo. I thought we weren't considering the gas cost of the transfer in at all and I think this should be done application side (integration test/fynd) not directly in the simulation.

For now let's focus on single swap only (this is the case that is handled in the integration test)

tamaralipows:
@dianacarvalho1 

> Can you tell me where this is done please?

No this was an open question. I don't know if this is done really. I will double check everything today. Will tweak the task with that info!

dianacarvalho1 (approved):
Thank you @tamaralipows ! I just wonder if there would be a better way to define all those constants. We can merge this for now and I'll move it to execution later!

PR #977, opened 2026-05-05
fix: calibrate protocol specific gas costs

Troshchk:
Note:
Before merging, feel free to squash all commits, since only the final state matters.

Also, could you please remove this [debug print](https://github.com/propeller-heads/tycho-indexer/pull/977/changes#diff-26e2e8378ea4a8cf3761d062001b0af30e6be9c975653df2faf9cd8a7ba4ffaaR1369)
 before merging? It’s no longer needed.

**UniswapV2 / SushiSwapV2 / PancakeSwapV2**

Reduced the gas estimates to exclude token transfer costs.
Based on the tenderly simulations, the previous estimates included both the swap cost and token transfer cost. The updated values now only represent the swap itself.
This increases the gas error rate, but once token transfer gas is added explicitly as a separate cost, the overall error rate should decrease.

Tenderly references:
[UniswapV2](https://www.tdly.co/shared/simulation/cd042fcd-e956-4c34-b390-fd4d662adb52)
[PancakeSwapV2](https://www.tdly.co/shared/simulation/711f4f75-9e25-4c79-be06-4e02cffe7833)

**ERC4626**

Updated gas estimates based on Tenderly simulation data.

Tenderly references:
[ERC4626](https://www.tdly.co/shared/simulation/66081c01-5edb-4dd1-886b-28c48b6d7ff7)

**UniswapV3 / UniswapV4 / PancakeSwapV3 / Aerodrome Slipstreams / Velodrome Slipstreams**

Added missing cross-tick gas cost and increased the per iteration gas cost based on tenderly.

Tenderly references:
[UniswapV3](https://www.tdly.co/shared/simulation/09a3d46b-6b73-4528-9fa0-ee91db683ec2)
[UniswapV3](https://www.tdly.co/shared/simulation/6b2eb513-09e5-47ae-9833-fe45156bbf1b)
[UniswapV3](https://www.tdly.co/shared/simulation/09a3d46b-6b73-4528-9fa0-ee91db683ec2)
[UniswapV4](https://www.tdly.co/shared/simulation/8eb7c1d0-e6a7-4323-a43a-0b11d3609d83)
[UniswapV4](https://www.tdly.co/shared/simulation/8eb7c1d0-e6a7-4323-a43a-0b11d3609d83)
[UniswapV4](https://www.tdly.co/shared/simulation/4bf188d9-6e9f-4b39-94ce-9577356eaa9c)
[Aerodrome Slipstreams](https://www.tdly.co/shared/simulation/4d8b1cd1-9ff6-40e2-8e6a-8224d4cc6e27)
[Velodrome Slipstreams](https://www.tdly.co/shared/simulation/2b913851-99bb-45ec-9fb9-b103ac15c585)

**Rocketpool**

Added missing component costs based on the tenderly simulation.

Tenderly references:
[Rocketpool](https://www.tdly.co/shared/simulation/9fe7bb68-e296-4ba2-a686-d73466c53818)

**Ekubo v2** 

Added missing component costs based on the tenderly simulation.
 
Tenderly references:
[Ekubo v2](https://www.tdly.co/shared/simulation/336dbf17-71ad-4d0a-9693-9bc6989d6c5b)

**Ekubo v3** 

Updated gas estimates to include the full swap lifecycle cost, instead of only the pool swap cost.

For concentrated pools, the previous estimate only included the base swap gas. The updated estimate now also includes transfer/payment lifecycle costs: transfer from user, start payment, withdraw, and complete payment.

For stableswap pools, the same lifecycle cost pattern was added, with a different withdraw cost based on Tenderly simulation data.

This adds the missing fixed costs on top of the existing base swap gas and should make Ekubo v3 estimates closer to actual execution gas.

Tenderly references:
[Ekubo v3 stable pool](https://www.tdly.co/shared/simulation/072ae7bb-32d2-4597-a9be-9d93f597f48b)
[Ekubo v3 concentrated pool](https://www.tdly.co/shared/simulation/e581c38e-6973-4170-b597-24cb18b9b451)

### DIFF
--- a/crates/tycho-simulation/src/evm/protocol/ekubo_v3/pool/concentrated.rs
+++ b/crates/tycho-simulation/src/evm/protocol/ekubo_v3/pool/concentrated.rs
@@ -33,7 +33,13 @@ use crate::{
 // Factor to account for computation inaccuracies due to not using tick bitmaps
 const WEI_UNDERESTIMATION_FACTOR: i128 = 2;
 
-const BASE_GAS_COST: u64 = 19_665;
+// The names of the constants reflect the exact method from the tenderly log.
+const TRANSFER_FROM_USER_GAS_COST: u64 = 40_000;
+const START_PAYMENT_GAS_COST: u64 = 6_000;
+const BASE_SWAP_GAS_COST: u64 = 19_665;
+const WITHDRAW_GAS_COST: u64 = 37_500;
+const COMPLETE_PAYMENT_GAS_COST: u64 = 2_000;
+
 const GAS_COST_OF_ONE_INITIALIZED_TICK_CROSSED: u64 = 14_259;
 const GAS_COST_OF_ONE_EXTRA_TICK_BITMAP_SLOAD: u64 = 2_000;
 const GAS_COST_OF_ONE_EXTRA_MATH_ROUND: u64 = 4_076;
@@ -173,7 +179,11 @@ pub(super) fn gas_costs(
     let (extra_distinct_bitmap_lookups, initialized_ticks_crossed) =
         (u64::from(extra_distinct_bitmap_lookups), u64::from(initialized_ticks_crossed));
 
-    BASE_GAS_COST +
+    START_PAYMENT_GAS_COST +
+        BASE_SWAP_GAS_COST +
+        WITHDRAW_GAS_COST +
+        COMPLETE_PAYMENT_GAS_COST +
+        TRANSFER_FROM_USER_GAS_COST +
         extra_distinct_bitmap_lookups * GAS_COST_OF_ONE_EXTRA_TICK_BITMAP_SLOAD +
         initialized_ticks_crossed * GAS_COST_OF_ONE_INITIALIZED_TICK_CROSSED +
         (initialized_ticks_crossed + extra_distinct_bitmap_lookups) *
--- a/crates/tycho-simulation/src/evm/protocol/erc4626/state.rs
+++ b/crates/tycho-simulation/src/evm/protocol/erc4626/state.rs
@@ -96,7 +96,7 @@ impl ProtocolSim for ERC4626State {
                     amount_in * self.asset_price /
                         U256::from(10).pow(U256::from(self.asset_token.decimals)),
                 ),
-                gas: 86107.to_biguint().expect("infallible"),
+                gas: 297000.to_biguint().expect("infallible"),
                 new_state: self.clone_box(),
             })
         } else if token_in.address == self.share_token.address &&
@@ -109,7 +109,7 @@ impl ProtocolSim for ERC4626State {
                     amount_in * self.share_price /
                         U256::from(10).pow(U256::from(self.share_token.decimals)),
                 ),
-                gas: 74977.to_biguint().expect("infallible"),
+                gas: 287000.to_biguint().expect("infallible"),
                 new_state: self.clone_box(),
             })
         } else {
--- a/crates/tycho-simulation/src/evm/protocol/rocketpool/state.rs
+++ b/crates/tycho-simulation/src/evm/protocol/rocketpool/state.rs
@@ -26,6 +26,15 @@ const DEPOSIT_FEE_BASE: u128 = 1_000_000_000_000_000_000; // 1e18
 /// 32 ETH — the hardcoded amount every megapool queue entry requests.
 const FULL_DEPOSIT_VALUE: u128 = 32_000_000_000_000_000_000u128;
 
+// The names of the constants reflect the exact method from the tenderly log.
+const BASE_DEPOSIT_GAS: u64 = 209_000;
+const RETH_MINT_GAS: u64 = 50_000;
+const VAULT_DEPOSIT_OR_EXCESS_GAS: u64 = 20_000;
+const ASSIGN_DEPOSITS_BASE_GAS: u64 = 20_000;
+// peekItem + withdrawEther + assignFunds + dequeueItem
+const GAS_PER_MEGAPOOL_ASSIGNMENT: u64 = 120_000;
+const BASE_WITHDRAWAL_GAS: u64 = 134_000;
+
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct RocketpoolState {
     pub reth_supply: U256,
@@ -179,12 +188,12 @@ impl RocketpoolState {
     ///
     /// The minimum of these three gives the number of entries assigned.
     /// Total ETH assigned = entries × 32 ETH.
-    fn calculate_assign_deposits(&self, deposit_amount: U256) -> U256 {
+    fn calculate_assign_deposits(&self, deposit_amount: U256) -> (U256, u64) {
         if !self.deposit_assigning_enabled ||
             self.megapool_queue_requested_total
                 .is_zero()
         {
-            return U256::ZERO;
+            return (U256::ZERO, 0);
         }
 
         let full_deposit_value = U256::from(FULL_DEPOSIT_VALUE);
@@ -205,7 +214,7 @@ impl RocketpoolState {
             .min(vault_cap)
             .min(queue_entries);
 
-        entries * full_deposit_value
+        (entries * full_deposit_value, entries.try_into().unwrap_or(u64::MAX))
     }
 }
 
@@ -287,18 +296,35 @@ impl ProtocolSim for RocketpoolState {
         };
 
         let mut new_state = self.clone();
+        let mut gas_used = 0;
+
         // Note: total_eth and reth_supply are not updated as they are managed by an oracle.
         if is_depositing_eth {
+            gas_used += BASE_DEPOSIT_GAS;
+
             // route ETH between rETH collateral buffer and vault.
             let (to_reth, to_vault) = new_state.compute_deposit_routing(amount_in)?;
             new_state.reth_contract_liquidity =
                 safe_add_u256(new_state.reth_contract_liquidity, to_reth)?;
             new_state.deposit_contract_balance =
                 safe_add_u256(new_state.deposit_contract_balance, to_vault)?;
+            if to_reth > U256::ZERO {
+                gas_used += RETH_MINT_GAS;
+            }
+
+            if to_vault > U256::ZERO {
+                gas_used += VAULT_DEPOSIT_OR_EXCESS_GAS;
+            }
 
             // Assign deposits: dequeue megapool entries and send ETH from vault to validators.
             // Both the vault balance and the queue requested are reduced by the same amount.
-            let eth_assigned = new_state.calculate_assign_deposits(amount_in);
+            let (eth_assigned, entries) = new_state.calculate_assign_deposits(amount_in);
+
+            if entries > 0 {
+                gas_used += ASSIGN_DEPOSITS_BASE_GAS;
+                gas_used += entries * GAS_PER_MEGAPOOL_ASSIGNMENT;
+            }
+
             if eth_assigned > U256::ZERO {
                 new_state.deposit_contract_balance =
                     safe_sub_u256(new_state.deposit_contract_balance, eth_assigned)?;
@@ -306,6 +332,7 @@ impl ProtocolSim for RocketpoolState {
                     safe_sub_u256(new_state.megapool_queue_requested_total, eth_assigned)?;
             }
         } else {
+            gas_used = BASE_WITHDRAWAL_GAS;
             // Withdraw from rETH contract first, spill remainder into deposit pool.
             let needed_from_deposit_pool =
                 amount_out.saturating_sub(new_state.reth_contract_liquidity);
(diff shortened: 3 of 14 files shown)


crates/tycho-simulation/src/evm/protocol/ekubo_v3/pool/concentrated.rs
```diff
@@ -33,7 +33,12 @@ use crate::{
 // Factor to account for computation inaccuracies due to not using tick bitmaps
 const WEI_UNDERESTIMATION_FACTOR: i128 = 2;
 
-const BASE_GAS_COST: u64 = 19_665;
+const TRANSFER_FROM_USER_GAS_COST: u64 = 20_000;
```
dianacarvalho1: I think this is quite low for a general token transfer.. [here](https://github.com/propeller-heads/tycho-indexer/blob/main/crates/tycho-execution/src/encoding/evm/gas_estimator.rs#L7) we are assuming 60k

crates/tycho-simulation/src/evm/protocol/ekubo_v3/pool/concentrated.rs
```diff
@@ -33,7 +33,12 @@ use crate::{
...
+const START_PAYMENT_GAS_COST: u64 = 6_000;
+const BASE_SWAP_GAS_COST: u64 = 19_665;
+const WITHDRAW_GAS_COST: u64 = 37_500;
```
dianacarvalho1, line 40: is this the out token transfer?

crates/tycho-simulation/src/evm/protocol/ekubo_v3/pool/concentrated.rs
```diff
@@ -173,7 +178,11 @@ pub(super) fn gas_costs(
     let (extra_distinct_bitmap_lookups, initialized_ticks_crossed) =
         (u64::from(extra_distinct_bitmap_lookups), u64::from(initialized_ticks_crossed));
 
-    BASE_GAS_COST +
+    START_PAYMENT_GAS_COST +
+        BASE_SWAP_GAS_COST +
+        WITHDRAW_GAS_COST +
+        COMPLETE_PAYMENT_GAS_COST +
+        TRANSFER_FROM_USER_GAS_COST +
```
dianacarvalho1, line 186: I'm not sure I see a big benefit of being this granular.. can you explain why you decided to split this up so much?

crates/tycho-simulation/src/evm/protocol/erc4626/state.rs
```diff
@@ -96,7 +96,7 @@ impl ProtocolSim for ERC4626State {
                     amount_in * self.asset_price /
                         U256::from(10).pow(U256::from(self.asset_token.decimals)),
                 ),
-                gas: 86107.to_biguint().expect("infallible"),
+                gas: 297000.to_biguint().expect("infallible"),
```
dianacarvalho1, line 99: this is a biiigggggg increase!! in the doc there is no justification for this change

crates/tycho-simulation/src/evm/protocol/rocketpool/state.rs
```diff
@@ -287,25 +294,43 @@ impl ProtocolSim for RocketpoolState {
...
+            }
+
             if eth_assigned > U256::ZERO {
                 new_state.deposit_contract_balance =
                     safe_sub_u256(new_state.deposit_contract_balance, eth_assigned)?;
                 new_state.megapool_queue_requested_total =
                     safe_sub_u256(new_state.megapool_queue_requested_total, eth_assigned)?;
             }
         } else {
+            gas_used = 209_000u64;
```
dianacarvalho1: what does this value represent?

crates/tycho-simulation/src/evm/protocol/uniswap_v3/state.rs
```diff
@@ -42,11 +42,17 @@ use crate::evm::protocol::{
...
-// This gas is estimated from UniswapV3Pool cross() calls on Tenderly
-const GAS_PER_TICK: u64 = 17_540;
+// Bitmap word scan
+const GAS_PER_BITMAP_WORD: u64 = 2_100;
+// swap math step: getSqrtRatioAtTick + computeSwapStep + amount accounting
+const GAS_PER_SWAP_MATH_STEP: u64 = 4_000;
+// Initialized tick crossing
+const GAS_PER_INITIALIZED_TICK_CROSS: u64 = 17_540;
+// Output transfer + balanceBefore + callback + balanceAfter.
+const V3_CALLBACK_SETTLEMENT_GAS: u64 = 55_000;
```
dianacarvalho1: this seems too small again (basing myself on the token transfer)

dianacarvalho1 (commented):
Thank you @Troshchk ! 🙏🏼 I have some comments..

crates/tycho-simulation/src/evm/protocol/ekubo_v3/pool/concentrated.rs
Troshchk: This number is based on `.transferFrom(` [here](https://dashboard.tenderly.co/shared/simulation/e581c38e-6973-4170-b597-24cb18b9b451?trace=0.4.0.8.1.0.0.3.0.1.0) and [here](https://dashboard.tenderly.co/shared/simulation/072ae7bb-32d2-4597-a9be-9d93f597f48b?trace=0.4.0.8.1.0.0.3.0.1.0).

Are those token transfers different to the general token transfer?

crates/tycho-simulation/src/evm/protocol/ekubo_v3/pool/concentrated.rs
Troshchk: yes, WITHDRAW_GAS_COST is out token transfer, it corresponds to `withdraw()` hence the name

crates/tycho-simulation/src/evm/protocol/ekubo_v3/pool/concentrated.rs
Troshchk: It makes the estimate less of a magic number. With the split, each part maps to a specific step we can see in the tenderly trace, so it’s easier to understand where the total comes from and adjust only the relevant part later if needed

crates/tycho-simulation/src/evm/protocol/erc4626/state.rs
Troshchk: It is based on the [tenderly data](https://dashboard.tenderly.co/shared/simulation/66081c01-5edb-4dd1-886b-28c48b6d7ff7?trace=0.4.0.8.1), I have simulated 3 and all of them had the same value

crates/tycho-simulation/src/evm/protocol/rocketpool/state.rs
Troshchk: it is base withdrawal cost, also based on tenderly
I have wrapped it into a constant to make it more clear

crates/tycho-simulation/src/evm/protocol/uniswap_v3/state.rs
Troshchk: This is also based on the numbers I saw in the logs. I am wondering whether these transfers are somehow different, than the ones estimated for 60k?

crates/tycho-simulation/src/evm/protocol/ekubo_v3/pool/concentrated.rs
dianacarvalho1: @tamaralipows maybe you can provide some context here on how you came with the 60k please?

crates/tycho-simulation/src/evm/protocol/erc4626/state.rs
dianacarvalho1: you simulated three: where they all the same token or different tokens?

Troshchk: that is a good point -- I have  simulated with the same token and I see that all simulations in the tycho integration test are done with the same token.
Should we update it there?

crates/tycho-simulation/src/evm/protocol/ekubo_v3/pool/concentrated.rs
tamaralipows: @dianacarvalho1 @Troshchk I understand the confusion. 

I got the 40-60k number from my own memory mostly lol but there are various articles quoting the same cost for an ERC20 token transfer. For example:  https://www.kucoin.com/learn/web3/understanding-ethereum-gas-fees 

When looking up some transactions on etherscan I also see around 40-60k.
WBTC: https://etherscan.io/tx/0x85e61e9ef10d9687ec4e1622e3daf759a8447490a64c4df2158265d87fec1d11
USDT: https://etherscan.io/tx/0xa0373d90a4fb00baa81af6641391d3aa1b306f8264d04351036cdc9446776ab9

Also, if you go to the gas profiler tool (same tenderly link you posted) you will often see transfer costs upwards of 35k. 

**HOWEVER, something I overlooked:** Performing a transfer as a part of another transaction might be cheaper because it doesn't include the "base transaction gas".  This means that all Ethereum transactions cost 21k gas (used for address recovery), so when the transfer is a part of another transaction and not standalone like the ones I linked, this 21k gas is already paid for.

You can probably verify that easily and quickly using a simple forge test and the `--gas-report` tool if you want to sanity check it: one test with an extra transfer, one test without. In any case I think putting at least 40k would be safer here.

crates/tycho-simulation/src/evm/protocol/ekubo_v3/pool/concentrated.rs
tamaralipows: I'm also a little bit worried about the numbers being so granular because I'm worried this will be overwhelming to maintain. For example if I want to fix the UniswapV3 gas the way it was before, I would right away just verify the number of ticks and the gas cost per tick. Here, I wouldn't know where to start and might lose some time verifying all of them... maybe EkuboV3 is super complicated, but I would just double check if it needs to be this way for all of the protocols

Troshchk: The names of the constants are reflecting the exact method from the tenderly log, so it should be quite easy to check them one by one instead of having just one number. 

Will it help if all these constants are summed up at the top in a new constant?
like const TOTAL_SWAP_COST=sum of all these steps
and only use TOTAL_SWAP_COST later in the code?

crates/tycho-simulation/src/evm/protocol/ekubo_v3/pool/concentrated.rs
Troshchk: Thank you for the explanation.

What’s still not clear to me is the gas increase you’re suggesting.
In the tenderly log I do see that gas for transfer inside the swap is around 20k(`.transferFrom(`).
The transfer of 35k is part of the ekubo v3 `_pay()` method and corresponds to `.transfer(` and that transfer is indeed around 35k.

Maybe `.transferFrom(` and `.transfer(` have different costs?

I’m not against increasing the number, but I’d like to understand the reasoning first. Why does 40k feel safer in this case if the tenderly cost is closer to 20k?

crates/tycho-simulation/src/evm/protocol/erc4626/state.rs
dianacarvalho1: why are all simulations in the integration test with the same token? 🤔 don't we support more? To verify this I would run the integration test locally with only this protocol and see how many pools we get for the default tvl

crates/tycho-simulation/src/evm/protocol/ekubo_v3/pool/concentrated.rs
tamaralipows: The tenderly cost is closer to 20k sometimes for some tokens. Other tokens it will be closer to 40k. It depends on the complexity of the token contract and also if balances are changing from nonzero to zero, nonzero to nonzero, or zero to nonzero. For that reason I'd like to stay on the higher end because overestimation is better than underestimation.
Afaik `transferFrom` should be the same if not more expensive than `transfer` since it needs to update allowances but you can verify that

Troshchk: I have verified and saw, that in our case `transferFrom` is often also around 20k, but I did not see allowances being updated as part of it.
Based on our discussion, I will decrease the the values for `DEFAULT_TOKEN_TRANSFER_GAS` and `TOKEN_APPROVAL_GAS` by 20k to not include the base transaction gas

crates/tycho-simulation/src/evm/protocol/erc4626/state.rs
Troshchk: I ran the tests locally as well and also saw only one pool.
This is expected, since the integration was exploratory and we currently support only one pool.

crates/tycho-simulation/src/evm/protocol/ekubo_v3/pool/concentrated.rs
dianacarvalho1: I think that is the same.. I will add a comment saying that the constants names represent solidity functions in the pool contracts

dianacarvalho1 (approved):
Thank you @Troshchk 🙏🏼

PR #978, opened 2026-05-05
fix(integration-test): add staleness watchdog for protocol sync metrics

louise-poole:
Adds a background task that marks all known protocols as Stale in the tycho_integration_protocol_sync_state gauge when no protocol update arrives within --stale-threshold-secs (default 30s). Prevents the gauge from freezing at its last value during a silent stream disconnection, ensuring alerting remains accurate during WS reconnects.

### DIFF
--- a/crates/tycho-integration-test/src/main.rs
+++ b/crates/tycho-integration-test/src/main.rs
@@ -7,7 +7,7 @@ use std::{
     fmt::Debug,
     str::FromStr,
     sync::{Arc, RwLock},
-    time::Duration,
+    time::{Duration, Instant},
 };
 
 use alloy::{
@@ -155,6 +155,12 @@ struct Cli {
     /// Enable partial block updates (flashblocks) on the tycho stream for lower latency
     #[arg(long, default_value_t = false)]
     partial_blocks: bool,
+
+    /// Seconds without a protocol update before marking all known protocols as stale in metrics.
+    /// Should be at least 2× the chain's expected block time (e.g. 30 for Ethereum at ~12s
+    /// blocks).
+    #[arg(long, default_value_t = 30)]
+    stale_threshold_secs: u64,
 }
 
 impl Debug for Cli {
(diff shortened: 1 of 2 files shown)


crates/tycho-integration-test/src/main.rs
```diff
@@ -155,6 +155,12 @@ struct Cli {
     /// Enable partial block updates (flashblocks) on the tycho stream for lower latency
     #[arg(long, default_value_t = false)]
     partial_blocks: bool,
+
+    /// Seconds without a protocol update before marking all known protocols as stale in metrics.
+    /// Should be at least 2× the chain's expected block time (e.g. 30 for Ethereum at ~12s
+    /// blocks).
+    #[arg(long, default_value_t = 30)]
+    stale_threshold_secs: u64,
```
dianacarvalho1: are you gonna change the deployments of the other chains to smaller values?

crates/tycho-integration-test/src/main.rs
```diff
@@ -329,6 +335,36 @@ async fn run(cli: Cli) -> miette::Result<()> {
     let mut protocol_stream_open = true;
     let mut rfq_stream_open = !cli.disable_rfq;
 
+    // Staleness watchdog: if no protocol update arrives within stale_threshold_secs, mark all
+    // known protocols as Stale in metrics. This catches stream disconnections where the
+    // SynchronizerState gauge would otherwise remain frozen at its last known value.
+    let known_protocols: Arc<RwLock<HashSet<String>>> = Arc::new(RwLock::new(HashSet::new()));
+    let last_protocol_update: Arc<RwLock<Instant>> = Arc::new(RwLock::new(Instant::now()));
```
dianacarvalho1: I wonder if we really need a new background task.. couldn't this be simplified a lot and still achieve the same result? Didn't think too much about it though, so I won't insist!

dianacarvalho1 (approved):
Thank you @louise-poole ! 🙏🏼

crates/tycho-integration-test/src/main.rs
louise-poole: Yes - though I think over-exaggerating this threshold value a bit isn't so bad. It's worst to have it too short than too long. So all the faster chains are fine with this default imo 👀

crates/tycho-integration-test/src/main.rs
louise-poole: True, I think you're right. Let me play a little with a simpler approach

tamaralipows (approved):
Thank you!

PR #983, opened 2026-05-05
feat(encoding): Add gas usage attributes 

dianacarvalho1:
- Also needed to bring back the Token in Swap instead of Bytes

### DIFF
--- a/crates/tycho-execution/CLAUDE.md
+++ b/crates/tycho-execution/CLAUDE.md
@@ -1,6 +1,7 @@
 # Tycho Execution
 
-DeFi swap execution framework: Solidity smart contracts (TychoRouter) + Rust encoding library. Multi-protocol token swaps with fee-taking, vault-based accounting, and 20+ DEX integrations.
+DeFi swap execution framework: Solidity smart contracts (TychoRouter) + Rust encoding library. Multi-protocol token
+swaps with fee-taking, vault-based accounting, and 20+ DEX integrations.
 
 **Docs**: https://docs.propellerheads.xyz/tycho
 **License**: BUSL-1.1 (Solidity), MIT (Rust)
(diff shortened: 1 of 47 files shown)


crates/tycho-execution/src/encoding/models.rs
```diff
@@ -269,6 +269,8 @@ impl PartialEq for Swap {
 /// * `interacting_with`: Address of the contract to be called.
 /// * `function_signature`: The signature of the function to be called.
 /// * `n_tokens`: Number of tokens in the swap.
+/// * `gas_usage`: Estimated gas usage for the encoded solution, summed from each swap's
+///   `estimated_gas_usage`.
```
tamaralipows: Do you think the `usage` is important? It couldn't just be `gas` or `estimated_gas`?

docs/for-solvers/execution/migration-guide-v2-to-v3.md
```diff
@@ -127,19 +150,29 @@ let swaps = encoded_solution.swaps();
...
+The `permit: Option<PermitSingle>` field has been removed from `EncodedSolution`. The encoder no longer creates or
+returns Permit2 data. If you use `TransferFromPermit2`, you must handle permit creation and signing yourself.
 
 The `Permit2` utility struct has been made public, so you can use it directly.
 
+**New `gas_usage` field:**
+
+`EncodedSolution` now exposes a `gas_usage: BigUint` (via `.gas_usage()`), summed from each
+swap's `estimated_gas_usage`. Solvers can use this as a single gas estimate per encoded solution without re-summing the
+swaps. Zero if no per-swap estimates were provided.
```
tamaralipows: Summing the swaps wouldn't work anyway since they don't include router overhead and some transfers?

tamaralipows (approved):
Thank you!

crates/tycho-execution/src/encoding/evm/gas_estimator.rs
```diff
@@ -0,0 +1,18 @@
...
+    // loop through solution to see the swaps and if they have estimated_gas_usage assigned
+    //   - if not -> skip and do nothing
+    //   - if yes ->
+    //      - sum up all the swap costs
+    //      - add router overhead
+    //      - add token transfer costs depend on the swap type. also depend on if fees are being
+    //        taken or not
+    //         - for single swaps: is it worth to model properly?
+    //         - for sequential swaps: is it worth to model properly?
+    //         - for split swaps: all transfers need to go through the router
```
louise-poole, line 15: Does this also depend on the protocol? From what I understood some protocols will have a token transfer included in their simulation swap cost no?

docs/for-solvers/execution/migration-guide-v2-to-v3.md
dianacarvalho1: yes, stupid claudio. i kept telling it it wouldn't be just summing up the swaps gas and it keep saying the same thing over and over. I will 🧹 this

crates/tycho-execution/src/encoding/models.rs
dianacarvalho1: will rename to `estimated_gas` then!

dianacarvalho1: in both structs actually! Diana from yesterday was not so coherent hein? annoying

crates/tycho-execution/src/encoding/evm/group_swaps.rs
```diff
@@ -80,6 +83,7 @@ pub fn group_swaps(swaps: &[Swap]) -> Vec<SwapGroup> {
                 protocol_system: current_swap_protocol.clone(),
                 swaps: vec![swap.clone()],
                 split: swap.split(),
+                estimated_gas_usage: BigUint::ZERO, // TODO: figure out what to do here
```
louise-poole, line 86: The more complex one is the if-branch above no? Where we group them and optimise gas. And then this part should be more clear... like token transfer in, swap and transfer out? How do transfers happen between groups? I assume one group directly transfers to the next group? Then we'll need to define which transfer is in which group... probably transfer in should always be included in each group and we tack on a transfer out at the very end of everything?

(I have very limited encoding codebase knowledge - this blurb is an attempt to validate some understanding and help me redirect if I'm on the wrong track)

crates/tycho-simulation/examples/quickstart/main.rs
```diff
@@ -543,8 +544,8 @@ fn get_best_swap(
     amount_in: BigUint,
     sell_token: Token,
     buy_token: Token,
-    amounts_out: &mut HashMap<String, BigUint>,
-) -> Option<(String, BigUint)> {
+    amounts_out: &mut HashMap<String, (BigUint, BigUint)>,
+) -> Option<(String, BigUint, BigUint)> {
```
louise-poole: I'm not such a fan of ambiguous tuples in our interfaces... 👀  especially as they start to get larger like this.

louise-poole: Ah - okay, this is just the quickstart. Might actually be even more important here no? To make it as easy as possible for others to follow? Like minimum we should use type aliases.

louise-poole (approved):
Looks neater/more contained than I was imagining! Nice!!! 
Just a couple questions from my side mainly to help my understanding.

crates/tycho-execution/src/encoding/models.rs
louise-poole: It's more than just the swap estimated gas summed though no? This I feel misrepresents the math behind the value/special case handling. Would rather even just have:
```suggestion
/// * `gas_usage`: Estimated gas usage for the encoded solution, derived from each swap's
///   `estimated_gas_usage`.
```

crates/tycho-execution/CLAUDE.md
```diff
@@ -169,6 +169,10 @@ Three implementations (`evm/strategy_encoder/`), each targeting a TychoRouter me
 
 There is no `ClientFeeParams` struct on the Rust side. The Solidity struct is encoded as a raw ABI tuple `(uint16,address,uint256,uint256,bytes)` embedded in the function signature strings within each strategy encoder. The caller provides the pre-signed bytes; the Rust encoder just passes them through.
 
+### Gas estimation
+
+`Swap::new(component, token_in: Token, token_out: Token, estimated_gas_usage: BigUint)` carries a per-swap simulation gas estimate (zero if unknown). Encoders aggregate these into `EncodedSolution.gas_usage`, exposing a single estimate for the whole solution. Tokens are passed as `tycho_common::models::token::Token` (not raw `Bytes`); the test helper `default_token(addr)` builds a minimal Token from an address.
```
louise-poole: `(zero if unknown)`

Are we fully confident with this? Will we never have a case that we might want to represent something as 0 gas, but it be considered known? I tend to prefer to use option if we want to represent known vs unknown.

louise-poole: Like if we just sum them with some zeros the final result looks like a nice low-gas situation when the reality is that it's actually unknown how much gas it will use and this is more of a min.

louise-poole: > Tokens are passed as `tycho_common::models::token::Token` (not raw `Bytes`)

Not sure this distinction is necessary. Claude has a habit of giving a nod to the pre-refactor version in its docs, but this info is clear from the fn signatures and doesn't need to take up claude doc context.

crates/tycho-execution/src/encoding/evm/gas_estimator.rs
dianacarvalho1: yes this depends on the protocol too. This was just a simplified description 😅 will be better in upcoming PRs!

crates/tycho-execution/src/encoding/evm/group_swaps.rs
dianacarvalho1: yeah.. figuring out gas estimations in group swaps is it's own independent task ([here](https://propeller-heads.atlassian.net/jira/software/projects/ENG/boards/24?selectedIssue=ENG-5962)). Maybe we can have a short call with you to explain group swaps a bit (it's also explained in the docs though!

crates/tycho-execution/CLAUDE.md
dianacarvalho1: humm tbh I'm not so sure what would be best here.. Ideally this should never be zero. the user should always pass this value and that's it. If this is zero (or None) then we wouldn't be able to calculate the solution's estimated gas. Do you think it makes more sense to make them both an option (in `Swap.estimated_gas` and in `EncodedSolution.estimated_gas`)? if one is None then the other will be None. 
I think I wouldn't like to even open up this option of not passing the `estimated_gas` to `Swap` in the first place though. I think removing this comment "use 0 if you don't know" from the docs would be sufficient

crates/tycho-simulation/examples/quickstart/main.rs
dianacarvalho1: made it even better by using the `GetAmountOutResult` directly instead of unwrapping the amount and gas individually!

Troshchk (approved):
Thank you!

crates/tycho-execution/CLAUDE.md
louise-poole: Yes - agreed. We need to choose one or the other and be firm on it. I'm all for it not being optional at all (remove all 'set to 0' suggestions).

PR #984, opened 2026-05-05
feat(router v3): use native ETH marker instead of address(0)

tamaralipows:
Reason:
address(0) served double duty in Solidity as both "unset" and "native ETH", which creates ambiguity. For example, tstore(_SWAP_INPUT_TOKEN_SLOT, 0) couldn't distinguish between "clear this slot" and "this slot holds ETH".

Main changes:
- This commit replaces address(0) with 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE as the ETH marker across the TychoRouter, Vault, and TransferManager
- Reverts if top-level token inputs are address(0)
- Updated model, reran with no exploits found
- Protocols that swap native ETH will now have eeee... encoded, and only translated back to address(0) in the executor if necessary, so from the user side, all is consistent between the TychoRouter and executors.

**To confirm: Should we move this boundary deeper into the Tycho stack?**

Double checked: Protocols which handle native ETH:

  - UniswapV4: (calldata encoded)
  - Ekubo: (calldata encoded)
  - EkuboV3: (calldata encoded)
  - FluidV1: Fixed (the ETH case is already hardcoded in solidity. double-verified via the rust encoded calldata in case)
  - Curve: (rust encoder has always used ETH - remove specific curve address)
  - WETH: (hardcoded ETH_ADDRESS, no addresses in calldata)
  - Rocketpool: (hardcoded ETH_ADDRESS, no addresses in calldata)
  - Etherfi: (hardcoded ETH_ADDRESS, no addresses in calldata)

Protocols which do not handle native ETH:
  - UniswapV2, UniswapV3, BalancerV2, BalancerV3, Slipstreams, MaverickV2, AerodromeV1, LiquidityParty, Bebop, Hashflow, Liquorice, ERC4626


crates/tycho-execution/model/src/model/executors.rs
```diff
@@ -209,11 +201,11 @@ impl Executor {
                 receiver: Address::Zero,
                 token_in: params.request(
                     ParamKey::ProtocolData { swap_index, start: 0, end: 20 },
-                    Address::POSSIBLY_ERC20_AND_ZERO,
+                    Address::POSSIBLY_ERC20_AND_NATIVE,
```
tamaralipows, line 204: We are using the NATIVE here in the model, even though the zero address is still passed to the executors, because the executors translate the 0 back to eeeee when `getTransferData` is called.

louise-poole (approved):
Looks 100s to me! 👌

crates/tycho-execution/contracts/src/executors/EkuboExecutor.sol
```diff
@@ -263,6 +264,8 @@ contract EkuboExecutor is IExecutor, ILocker, IPayer, ICallback {
         uint256 lastHopOffset = 20 + (hopsLength - 1) * _HOP_BYTE_LEN;
         tokenIn = address(bytes20(data[0:20]));
         tokenOut = address(bytes20(data[lastHopOffset:lastHopOffset + 20]));
+        if (tokenIn == address(0)) tokenIn = ETH_ADDRESS;
+        if (tokenOut == address(0)) tokenOut = ETH_ADDRESS;
```
louise-poole: Please update the dex integration docs for the `getTransferData` section with this required behaviour?

louise-poole: And this section: https://docs.propellerheads.xyz/tycho/for-dexs/protocol-integration/execution#native-token-address-handling

crates/tycho-execution/contracts/lib/NativeETH.sol
```diff
@@ -0,0 +1,4 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.26;
+
+address constant ETH_ADDRESS = 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
```
dianacarvalho1, line 4: my biggest question: how does `0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE` sound like? 😂

dianacarvalho1 (commented):
Thank you @tamaralipows 🙏🏼 Great job at adapting the maximodel to handle this!

However, I don't think I agree with allowing address(0) to be ETH in the executors too 😕 I understand that the motivation for this change is only true in the router, however from an understanding perspective of the user it is a bit weird to use 0xeeEEee in the router but then in the executor it's address(0). I would prefer if the boundary was in the interaction with the protocols and everything that is tycho related is consistent. But I don't feel very strongly about this and can be persuaded that the boundary is in the executor/swap encoder.

Troshchk (approved):
Thank you @tamaralipows

crates/tycho-execution/src/encoding/evm/utils.rs
```diff
@@ -18,7 +18,18 @@ use once_cell::sync::Lazy;
 use tokio::runtime::{Handle, Runtime};
 use tycho_common::Bytes;
 
-use crate::encoding::{errors::EncodingError, models::Swap};
+use crate::encoding::{errors::EncodingError, evm::constants::ROUTER_ETH_ADDRESS, models::Swap};
+
+/// Converts `Address::ZERO` (protocol-native ETH marker) to the
+/// `ETH_ADDRESS` marker (0xEeee…) used by the TychoRouter. Non-zero
+/// addresses pass through unchanged.
+pub fn native_to_router_eth(addr: Address) -> Address {
```
dianacarvalho1: not a fan of this function.. when reading where it is used it seems like the conversion to router eth will always happen..

dianacarvalho1: maybe `convert_to_router_token(..)`?

dianacarvalho1 (approved):
Thank you @tamaralipows 🙏🏼 yes I like it much better like this! And no, I don't think we should change this higher in the tycho stack. I think changing only in encoding + execution makes the most sense! 
Only have 1 small comment

PR #987, opened 2026-05-06
ci(cargo-update): exclude workspaces with .cargo-update-skip marker

zizou0x:
Add a marker-file based mechanism to skip workspaces from the automated cargo update workflow. Drop a .cargo-update-skip file in any workspace root to exclude it. 

Applied to protocols/substreams/ which we want to stay as frozen as possible (any code change would change the Substreams cache and might introduce bugs).


Troshchk (approved):
Thank you!

PR #993, opened 2026-05-06
fix(encoding): Estimate gas properly for grouped swaps

dianacarvalho1:
Assumes that all protocols that can be grouped include both token in and out transfer gas usage in the swap.estimated_gas (this is acceptable because all protocols that can group do so with a callback mechanism). If there is a group we need to remove the extra token transfers gas costs that now will be saved:
- for the first hop: remove the token out transfer gas
- for all in between hops: remove both token out and in transfer gas
- for the last hop: remove the token in transfer gas

Misc fixes:
- add vm:balancer_v3 to PROTOCOLS_CALLBACK list
- Rename estimated_gas_usage -> estimated_gas in SwapGroup
- Extend PartialEq for SwapGroup and Swap

TODO later in fynd:
- Actually measure that this is reasonable and we are not doing nonsense stuff here. Theoretically I think this should work but would like to verify!

### DIFF
--- a/crates/tycho-execution/src/encoding/evm/group_swaps.rs
+++ b/crates/tycho-execution/src/encoding/evm/group_swaps.rs
@@ -101,7 +101,8 @@ pub fn group_swaps(swaps: &[Swap]) -> Vec<SwapGroup> {
 /// Aggregate per-swap gas estimates for a group, discounting transfers skipped by batching.
 ///
 /// Each swap's `estimated_gas` is assumed to include its input and output token transfer
-/// costs. When swaps are batched (e.g. USV4's PoolManager flash accounting), intermediate
+/// costs (this is acceptable because all protocols that can group do so with a callback mechanism).
+/// When swaps are batched (e.g. USV4's PoolManager flash accounting), intermediate
 /// transfers are saved: the first swap's output transfer, every middle swap's input and
 /// output transfers, and the last swap's input transfer. The group's external input and
 /// output transfers remain.
--- a/crates/tycho-execution/src/encoding/evm/group_swaps.rs
+++ b/crates/tycho-execution/src/encoding/evm/group_swaps.rs
@@ -12,7 +12,7 @@ use crate::encoding::{evm::constants::GROUPABLE_PROTOCOLS, models::Swap};
 /// * `protocol_system`: String, the protocol system of the swaps
 /// * `swaps`: Vec<Swap>, the sequence of swaps to be executed as a group
 /// * `split`: f64, the split percentage of the first swap in the group
-/// * `estimated_gas_usage`: BigUint, the estimated gas usage of the swap group
+/// * `estimated_gas`: BigUint, the estimated gas usage of the swap group
 #[derive(Clone, Debug)]
 pub struct SwapGroup {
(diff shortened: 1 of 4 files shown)


crates/tycho-execution/src/encoding/evm/group_swaps.rs
```diff
@@ -74,33 +75,64 @@ pub fn group_swaps(swaps: &[Swap]) -> Vec<SwapGroup> {
...
+        group.estimated_gas = compute_group_gas(&group.swaps);
+        grouped_swaps.push(group);
     }
     grouped_swaps
 }
 
+/// Aggregate per-swap gas estimates for a group, discounting transfers skipped by batching.
+///
+/// Each swap's `estimated_gas` is assumed to include its input and output token transfer
+/// costs. When swaps are batched (e.g. USV4's PoolManager flash accounting), intermediate
```
tamaralipows: Did I missunderstand something? I thought we only add the input/output token transfer at the end, so this assumption isn't safe to make yet?

crates/tycho-execution/src/encoding/evm/group_swaps.rs
```diff
@@ -74,33 +75,64 @@ pub fn group_swaps(swaps: &[Swap]) -> Vec<SwapGroup> {
...
+    let n = swaps.len();
+    if n <= 1 {
+        return total_gas;
+    }
+    let safe_sub = |t: BigUint, v: BigUint| if t >= v { t - v } else { BigUint::ZERO };
+    total_gas = safe_sub(total_gas, swaps[0].token_out().gas_usage());
+    for swap in &swaps[1..n - 1] {
+        total_gas = safe_sub(total_gas, swap.token_in().gas_usage());
+        total_gas = safe_sub(total_gas, swap.token_out().gas_usage());
+    }
```
tamaralipows, line 123: My guess is that this will be very conservative (and on the high gas usage side), because I think the logic saves more than the token transfers, but also a lot of overhead. For example, it doesn't need to perform a call twice to the Uniswap contract (in the Uniswap case). But I also don't think there is any point in speculating. We can just test it concretely later.

All that to say I think this might be inaccurate but probably fine for now anyway.

crates/tycho-execution/src/encoding/evm/group_swaps.rs
tamaralipows: Oh nevermind I missed your PR description.

> (this is acceptable because all protocols that can group do so with a callback mechanism)

I would just add this to the docstring here pls

tamaralipows (approved):
I think this works!  Thank you!

louise-poole (approved):
Seems logical. I agree that we might see more savings than just the transfer costs. Maybe we should add a TODO or task somewhere to remember to verify this and adjust if needed?

dianacarvalho1:
> Seems logical. I agree that we might see more savings than just the transfer costs. Maybe we should add a TODO or task somewhere to remember to verify this and adjust if needed?

@louise-poole It was already in the PR description 😅 
image

and also already added a task for it [here](https://propeller-heads.atlassian.net/jira/software/projects/ENG/boards/24?selectedIssue=ENG-5937)

PR #998, opened 2026-05-07
feat(examples): Add Polygon and Arbitrum support to simulation examples

tvinagre:
## Summary
- Add `Chain::Polygon` support to `price_printer` and `quickstart` examples with exchanges: `uniswap_v2`, `quickswap_v2`, `uniswap_v3`, `uniswap_v4`
- Add `Chain::Arbitrum` support with exchanges: `uniswap_v2`, `uniswap_v3`, `pancakeswap_v3`, `uniswap_v4`
- Auto-detect TLS based on `TYCHO_API_KEY` presence — when unset, uses plain HTTP/WS and prints a warning, enabling local indexer usage
- Add default buy/sell tokens for Polygon (USDC → WPOL) and Arbitrum (USDC → WETH) in quickstart

ENG-5644 ENG-5612

## Test plan
- [x] `cargo check` passes for both examples
- [x] Tested `price_printer` against local Polygon indexer at `localhost:4242`
- [ ] Verify quickstart works end-to-end against a Polygon/Arbitrum indexer

### DIFF
--- a/crates/tycho-simulation/examples/price_printer/main.rs
+++ b/crates/tycho-simulation/examples/price_printer/main.rs
@@ -80,6 +80,20 @@ fn register_exchanges(
                 .exchange::<UniswapV3State>("uniswap_v3", tvl_filter.clone(), None)
                 .exchange::<UniswapV4State>("uniswap_v4", tvl_filter.clone(), None)
         }
+        Chain::Polygon => {
+            builder = builder
+                .exchange::<UniswapV2State>("uniswap_v2", tvl_filter.clone(), None)
+                .exchange::<UniswapV2State>("quickswap_v2", tvl_filter.clone(), None)
+                .exchange::<UniswapV3State>("uniswap_v3", tvl_filter.clone(), None)
+                .exchange::<UniswapV4State>("uniswap_v4", tvl_filter.clone(), None)
+        }
+        Chain::Arbitrum => {
+            builder = builder
+                .exchange::<UniswapV2State>("uniswap_v2", tvl_filter.clone(), None)
+                .exchange::<UniswapV3State>("uniswap_v3", tvl_filter.clone(), None)
+                .exchange::<UniswapV3State>("pancakeswap_v3", tvl_filter.clone(), None)
+                .exchange::<UniswapV4State>("uniswap_v4", tvl_filter.clone(), None)
+        }
         _ => {}
     }
     builder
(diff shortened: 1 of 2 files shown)


crates/tycho-simulation/examples/price_printer/main.rs
```diff
@@ -98,11 +112,12 @@ async fn main() {
         get_default_url(&chain).unwrap_or_else(|| panic!("Unknown URL for chain {}", cli.chain))
     });
 
-    let tycho_api_key: String =
-        env::var("TYCHO_API_KEY").unwrap_or_else(|_| "sampletoken".to_string());
+    let tycho_api_key: Option<String> = env::var("TYCHO_API_KEY").ok();
+    let no_tls = tycho_api_key.is_none();
+    if no_tls {
+        eprintln!("Warning: TYCHO_API_KEY not set. Using plain HTTP/WS (no TLS).");
+    }
```
zizou0x, line 122: I think it's overkill and not safe. Only useful when using on a local tycho instance no? 

If yes then I'd rather make it a cli arg to manually disable it than risking external people using no TLS without noticing just because they forgot to set an env variable

tvinagre: > Only useful when using on a local tycho instance no?
Yes, for self-hosted or port-forwarded cases

crates/tycho-simulation/examples/price_printer/main.rs
```diff
@@ -98,11 +115,12 @@ async fn main() {
         get_default_url(&chain).unwrap_or_else(|| panic!("Unknown URL for chain {}", cli.chain))
     });
 
-    let tycho_api_key: String =
-        env::var("TYCHO_API_KEY").unwrap_or_else(|_| "sampletoken".to_string());
+    let tycho_api_key: Option<String> = env::var("TYCHO_API_KEY").ok();
+    let no_tls = cli.no_tls;
+    if no_tls {
+        eprintln!("Warning: TLS is disabled. Only use this for local/self-hosted Tycho instances.");
+    }
```
louise-poole, line 122: The intermediate var is pretty shallow tbh... I would skip it and just use cli.no_tls directly.
```suggestion
    if cli.no_tls {
        eprintln!("Warning: TLS is disabled. Only use this for local/self-hosted Tycho instances.");
    }
```

crates/tycho-simulation/examples/price_printer/main.rs
```diff
@@ -98,11 +115,12 @@ async fn main() {
...
 
-    // Perform an early check to ensure `RPC_URL` is set.
-    // This prevents errors from occurring later during UI interactions for curve.
```
louise-poole, line 105: Why remove this? I like that this indicated to new users that this is needed for curve specifically.

crates/tycho-simulation/examples/quickstart/main.rs
```diff
@@ -128,8 +135,11 @@ async fn main() {
             .unwrap_or_else(|| panic!("Unknown URL for chain {chain}", chain = cli.chain))
     });
 
-    let tycho_api_key: String =
-        env::var("TYCHO_API_KEY").unwrap_or_else(|_| "sampletoken".to_string());
+    let tycho_api_key: Option<String> = env::var("TYCHO_API_KEY").ok();
+    let no_tls = cli.no_tls;
+    if no_tls {
+        eprintln!("Warning: TLS is disabled. Only use this for local/self-hosted Tycho instances.");
+    }
```
louise-poole, line 142: Same comment as above

PR #999, opened 2026-05-07
feat(tycho-client): replace dto types with model types in public API

kayibal:
## Summary

- Moves the dto deserialization boundary to the wire edges of `tycho-client` (`rpc.rs`, `deltas.rs`) — dto types no longer appear in public trait signatures, public error types, or internal data structures
- Introduces builder-style parameter structs for all 7 `RPCClient` required methods, making future parameter additions backwards-compatible
- Migrates `DeltasClient::subscribe` to accept `models::ExtractorIdentity`
- Migrates `Chain` in `protocols/testing` from dto to models throughout
- Adds `TracedEntryPoints` type alias and `From<TracedEntryPoints> for DCIUpdate` in `tycho-common::models::blockchain`, replacing the removed `From<TracedEntryPointRequestResponse> for dto::DCIUpdate`

## Key changes

**`tycho-common`**
- `models/blockchain.rs`: new `TracedEntryPoints` type alias + `From<TracedEntryPoints> for DCIUpdate`
- `dto.rs`: removed `From<TracedEntryPointRequestResponse> for DCIUpdate` (superseded by the model-layer impl)

**`tycho-client`**
- `rpc.rs`: 7 builder-style params structs (`ContractStateParams`, `ProtocolComponentsParams`, `ProtocolStatesParams`, `TokensParams`, `ProtocolSystemsParams`, `ComponentTvlParams`, `TracedEntryPointsParams`) — private fields, minimal `new()`, `with_*()` setters, internal-only `with_pagination()`; `RPCClient` trait methods updated to take these structs; dto construction moved fully inside `HttpRPCClient` method bodies
- `deltas.rs`: `DeltasClient::subscribe` takes `models::ExtractorIdentity`; `DeltasError::ServerError` no longer exposes `dto::WebsocketError` in its variant
- `feed/synchronizer.rs`, `feed/component_tracker.rs`, `stream.rs`: updated to new signatures
- `cli.rs`: `Chain` switched from dto to models

**`protocols/testing`**
- `TychoRunner`, `CommonArgs`, `main.rs`: `Chain` switched from dto to models
- `tycho_rpc.rs`: updated to new `RPCClient` method signatures

### DIFF
--- a/crates/tycho-client/src/feed/dto.rs
+++ b/crates/tycho-client/src/feed/dto.rs
@@ -1,6 +1,6 @@
 //! Serializable wrappers for the client feed pipeline types.
 //!
-//! [`FeedMessageDto`], [`StateSyncMessageDto`], etc. mirror the structures of their
+//! [`FeedMessage`], [`StateSyncMessage`], etc. mirror the structures of their
 //! model-based counterparts but use [`tycho_common::dto`] types that carry full
 //! `Serialize` / `Deserialize` support. The JSON format produced here matches the
 //! wire format (dto field names such as `contract_ids` and `state_updates`), which
@@ -8,6 +8,8 @@
 //! existing test fixture files.
 //!
 //! Conversion to and from the model-based pipeline types is provided via `From` impls.
+//! When both names are in scope use the module path to disambiguate, e.g.
+//! `feed::dto::ComponentWithState` vs `feed::synchronizer::ComponentWithState`.
 
 use std::collections::HashMap;
 
@@ -22,108 +24,49 @@ use tycho_common::{
 };
 
 use crate::feed::{
-    synchronizer::{ComponentWithState, Snapshot, StateSyncMessage},
-    BlockHeader, FeedMessage, HeaderLike, SynchronizerState,
+    synchronizer::{
+        ComponentWithState as ModelComponentWithState, Snapshot as ModelSnapshot,
+        StateSyncMessage as ModelStateSyncMessage,
+    },
+    BlockHeader, FeedMessage as ModelFeedMessage, HeaderLike, SynchronizerState,
 };
 
-/// Serializable counterpart of [`ComponentWithState`].
+/// Serializable counterpart of [`crate::feed::synchronizer::ComponentWithState`].
 #[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct ComponentWithStateDto {
+pub struct ComponentWithState {
     pub state: ResponseProtocolState,
     pub component: DtoProtocolComponent,
     pub component_tvl: Option<f64>,
     pub entrypoints: Vec<(DtoEntryPointWithTracingParams, DtoTracingResult)>,
 }
 
-/// Serializable counterpart of [`Snapshot`].
+/// Serializable counterpart of [`crate::feed::synchronizer::Snapshot`].
 #[derive(Debug, Clone, Serialize, Deserialize, Default)]
-pub struct SnapshotDto {
-    pub states: HashMap<String, ComponentWithStateDto>,
+pub struct Snapshot {
+    pub states: HashMap<String, ComponentWithState>,
     #[serde(with = "tycho_common::serde_primitives::hex_hashmap_key")]
     pub vm_storage: HashMap<Bytes, ResponseAccount>,
 }
 
-/// Serializable counterpart of [`StateSyncMessage`].
+/// Serializable counterpart of [`crate::feed::synchronizer::StateSyncMessage`].
 #[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct StateSyncMessageDto<H = BlockHeader> {
+pub struct StateSyncMessage<H = BlockHeader> {
     pub header: H,
-    pub snapshots: SnapshotDto,
+    pub snapshots: Snapshot,
     pub deltas: Option<BlockChanges>,
     pub removed_components: HashMap<String, DtoProtocolComponent>,
 }
 
-/// Serializable counterpart of [`FeedMessage`].
+/// Serializable counterpart of [`crate::feed::FeedMessage`].
 #[derive(Debug, Clone, Serialize, Deserialize, Default)]
-pub struct FeedMessageDto<H = BlockHeader> {
-    pub state_msgs: HashMap<String, StateSyncMessageDto<H>>,
+pub struct FeedMessage<H = BlockHeader> {
+    pub state_msgs: HashMap<String, StateSyncMessage<H>>,
     pub sync_states: HashMap<String, SynchronizerState>,
 }
 
-// ── FeedMessageDto → model types ──────────────────────────────────────────────
+// ── dto types → model types ───────────────────────────────────────────────────
 
-impl From<ComponentWithStateDto> for ComponentWithState {
-    fn from(value: ComponentWithStateDto) -> Self {
-        Self {
-            state: value.state.into(),
-            component: value.component.into(),
-            component_tvl: value.component_tvl,
-            entrypoints: value
-                .entrypoints
-                .into_iter()
-                .map(|(ep, tr)| (ep.into(), tr.into()))
-                .collect(),
-        }
-    }
-}
-
-impl From<SnapshotDto> for Snapshot {
-    fn from(value: SnapshotDto) -> Self {
-        Self {
-            states: value
-                .states
-                .into_iter()
-                .map(|(k, v)| (k, v.into()))
-                .collect(),
-            vm_storage: value
-                .vm_storage
-                .into_iter()
-                .map(|(k, v)| (k, v.into()))
-                .collect(),
-        }
-    }
-}
-
-impl<H: HeaderLike> From<StateSyncMessageDto<H>> for StateSyncMessage<H> {
-    fn from(value: StateSyncMessageDto<H>) -> Self {
-        Self {
-            header: value.header,
-            snapshots: value.snapshots.into(),
-            deltas: value.deltas.map(Into::into),
-            removed_components: value
-                .removed_components
-                .into_iter()
-                .map(|(k, v)| (k, v.into()))
-                .collect(),
-        }
-    }
-}
-
-impl<H: HeaderLike> From<FeedMessageDto<H>> for FeedMessage<H> {
-    fn from(value: FeedMessageDto<H>) -> Self {
-        Self {
-            state_msgs: value
-                .state_msgs
-                .into_iter()
-                .map(|(k, v)| (k, v.into()))
-                .collect(),
-            sync_states: value.sync_states,
-        }
-    }
-}
-
-// ── model types → FeedMessageDto ──────────────────────────────────────────────
-
-impl From<ComponentWithState> for ComponentWithStateDto {
+impl From<ComponentWithState> for ModelComponentWithState {
     fn from(value: ComponentWithState) -> Self {
         Self {
             state: value.state.into(),
@@ -138,7 +81,7 @@ impl From<ComponentWithState> for ComponentWithStateDto {
     }
 }
 
-impl From<Snapshot> for SnapshotDto {
+impl From<Snapshot> for ModelSnapshot {
     fn from(value: Snapshot) -> Self {
         Self {
             states: value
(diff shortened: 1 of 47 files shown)


crates/tycho-simulation/src/evm/protocol/aerodrome_v1/decoder.rs
```diff
@@ -113,12 +113,18 @@ mod tests {
...
+            component: tycho_common::models::protocol::ProtocolComponent::from(
+                tycho_common::dto::ProtocolComponent {
+                    tokens: vec![token0, token1],
+                    static_attributes: HashMap::from([(
+                        "is_stable".to_string(),
+                        Bytes::from(vec![0]),
+                    )]),
+                    ..Default::default()
+                },
+            ),
```
louise-poole: Can't we create the model ProtocolComponent struct directly and skip the dto version? Or is there a purpose to creating it this way that I'm missing?

crates/tycho-simulation/src/evm/protocol/uniswap_v3/state.rs
```diff
@@ -854,57 +846,12 @@ mod tests {
...
-            .unwrap();
-
-        assert_eq!(&res.0, &BigUint::from_u128(155144999154).unwrap()); // Crazy amount because of this tick: "ticks/-887272/net-liquidity": "0x10d73d"
-
-        let out = usv3_state
-            .get_amount_out(res.0, &t0, &t1)
-            .expect("swap for limit in didn't work");
-
-        assert_eq!(&res.1, &out.amount);
+        unimplemented!("needs rewrite: ComponentWithState is no longer Deserialize");
```
louise-poole: I mean the test is already marked as ignored... no need to remove it entirely? This will be added back once the test asset is adapted no?

crates/tycho-simulation/src/utils.rs
```diff
@@ -103,17 +103,8 @@ pub async fn load_all_tokens(
...
-                        "Unable to convert token `{symbol}` at {address} on chain {chain} into ERC20 token",
-                        symbol = token_clone.symbol,
-                        address = token_clone.address,
-                        chain = token_clone.chain,
-                    ))
-                })
+            let converted = token;
+            Ok((converted.address.clone(), converted))
         })
         .collect()
```
louise-poole, line 109: What's the purpose of `let converted = token`? We're no longer converting anything so the nomenclature here is outdated. 
```suggestion
    tokens
        .into_iter()
        .map(|token| Ok((token.address.clone(), token)))
        .collect()
```

crates/tycho-simulation/src/evm/protocol/uniswap_v3/state.rs
louise-poole: I understand this was an intermediary step, but doing it like this also makes it not so great to review what actually changed commit-by-commit. The diff I see is to unimplemented and then from unimplemented. Not the end of the world (I can also review to PR as a whole) but not so great when it's a huge PR and the by-commit review is saving my sanity.

crates/tycho-client/src/feed/dto.rs
```diff
@@ -22,47 +24,50 @@ use tycho_common::{
...
+        StateSyncMessage as ModelStateSyncMessage,
+    },
+    BlockHeader, FeedMessage as ModelFeedMessage, HeaderLike, SynchronizerState,
 };
 
-/// Serializable counterpart of [`ComponentWithState`].
+/// Serializable counterpart of [`crate::feed::synchronizer::ComponentWithState`].
 #[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct ComponentWithStateDto {
+pub struct ComponentWithState {
```
louise-poole: I'm a little confused... a previous commit literally just renamed these to somethingDto and now it's changing back? What was the purpose?

crates/tycho-simulation/src/evm/protocol/uniswap_v3/state.rs
```diff
@@ -856,14 +856,14 @@ mod tests {
 
     #[tokio::test]
     async fn test_get_limits() {
-        use tycho_client::feed::dto::ComponentWithStateDto;
+        use tycho_client::feed::dto::ComponentWithState as DtoComponentWithState;
```
louise-poole: Personally I preferred the SomethingDto version to the DtoSomething version. It felt naturally smoother to read... the second version is interrupting my read of the logic flow with Dto in the middle of things telling me it's a dto type when honestly that doesn't matter in a brief code read and it's more important I know WHAT the object represented.
Don't feel crazy strongly about this though so if you prefer it this other way we can leave it.

crates/tycho-client/src/rpc.rs
```diff
@@ -91,8 +182,23 @@ impl ContractStateParams {
...
+            None,
+            Some({
+                #[allow(deprecated)]
+                BlockParam {
+                    hash: None,
+                    chain: Some(self.chain.into()),
+                    number: Some(block_number as i64),
+                }
+            }),
+        );
```
louise-poole: This same code block is duplicated in a few places... How about implement the conversion in a VersionParam::from_block(chain, n) and just use that?

crates/tycho-client/src/feed/synchronizer.rs
```diff
@@ -339,11 +339,12 @@ where
...
                 .get_traced_entry_points_paginated(
-                    self.extractor_id.chain,
-                    &self.extractor_id.name,
-                    &component_ids,
-                    None,
-                    RPC_CLIENT_CONCURRENCY,
+                    TracedEntryPointsPaginatedParams::new(
+                        self.extractor_id.chain,
+                        self.extractor_id.name.as_str(),
+                        component_ids.clone(),
```
louise-poole: Do we need to clone here? I don't see this used after this point?

crates/tycho-client/src/rpc.rs
```diff
@@ -324,6 +445,260 @@ impl TracedEntryPointsParams {
...
 
+/// Parameters for [`RPCClient::get_protocol_components_paginated`].
+#[derive(Clone, PartialEq, Debug)]
+pub struct ProtocolComponentsPaginatedParams {
+    pub(crate) chain: Chain,
+    pub(crate) protocol_system: String,
+    pub(crate) component_ids: Option<Vec<ComponentId>>,
+    pub(crate) tvl_gt: Option<f64>,
+    pub(crate) chunk_size: Option<usize>,
+    pub(crate) concurrency: usize,
```
louise-poole: Why are these ones pub(crate) and all the others (except TracedEntryPointsPaginatedParams) are private? I would prefer it be consistent between similar structs.

crates/tycho-client/src/rpc.rs
```diff
@@ -881,20 +1200,15 @@ pub trait RPCClient: Send + Sync {
...
-                    let batch = try_join_all(tasks).await?;
+                    let responses = try_join_all(tasks).await?;
 
-                    let mut done = false;
-                    for resp in batch {
-                        if resp.len() < chunk_size {
-                            done = true;
+                    for resp in responses {
+                        for (key, value) in resp {
+                            *merged_tvl.entry(key).or_insert(0.0) += value;
```
louise-poole: Don't we want to overwrite with the latest value, not accumulate them?
```suggestion
                            *merged_tvl.entry(key).or_insert(0.0) = value;
```

crates/tycho-client/src/rpc.rs
```diff
@@ -42,17 +42,108 @@ use tycho_common::{
...
+    pub fn page(&self) -> i64 {
+        self.page
+    }
+
+    pub fn page_size(&self) -> i64 {
+        self.page_size
+    }
+}
+
+#[allow(private_bounds)]
```
louise-poole: I'm not sure I understand the intention here... we have a pub struct for Page, but it's bound by a more private pub(crate) trait. So while we can access the struct publicly we cannot call `len()` or `is_empty()`? Is there a point to this CollectionLen trait? I feel like this is a bit over-engineered... can we not just implement this directly on Page?

louise-poole (commented):
Overall I found this veeeery satisfying. So much cleaner 👌

crates/tycho-simulation/src/evm/protocol/uniswap_v3/state.rs
kayibal: So far in the repository things are namespaced with the module so I prefer the structs to have the same name and use crate namespacing instead so ideally it should be `dto::ComponentWithState`

crates/tycho-client/src/rpc.rs
kayibal: This super confused me. It was actually previously already a `+=`, it should never accumulate though, it basically just adds the value here to 0 which is equivalent to the assignment you suggested. I think this is confusing though... I wonder why we have such weird code here....

crates/tycho-client/src/feed/synchronizer.rs
```diff
@@ -1198,14 +1185,35 @@ mod test {
...
+                HashMap::new(),
+                Bytes::default(),
+                HashMap::new(),
+                Bytes::default(),
+                Bytes::default(),
+                Bytes::default(),
+                Bytes::default(),
+                None,
+            ),
+        ]
```
louise-poole, line 1215: This feels a little ugly now... I'm sad about losing the `..Default::default()`. Might be nice to make a helper fn to generate a default Account with a set address. Though I suppose this is only used twice here so not really worth it.

crates/tycho-client/src/feed/synchronizer.rs
```diff
@@ -1257,11 +1265,11 @@ mod test {
...
                         ComponentWithState {
-                            state: ResponseProtocolState {
+                            state: ProtocolComponentState {
                                 component_id: "Component1".to_string(),
-                                ..Default::default()
-                            }
-                            .into(),
+                                attributes: HashMap::new(),
+                                balances: HashMap::new(),
+                            },
```
louise-poole, line 1267: Why can't we still have the default unpacking here? Does it not implement default?

crates/tycho-client/src/feed/dto.rs
```diff
@@ -66,7 +56,7 @@ pub struct FeedMessage<H = BlockHeader> {
 
 // ── dto types → model types ───────────────────────────────────────────────────
 
-impl From<ComponentWithState> for ModelComponentWithState {
+impl From<ComponentWithState> for synchronizer::ComponentWithState {
```
louise-poole: I think it's enough to qualify just the dto structs. The rest it's obvious they're the non-dto version.

crates/tycho-client/src/feed/dto.rs
```diff
@@ -15,29 +15,19 @@ use std::collections::HashMap;
...
 };
 
-use crate::feed::{
-    synchronizer::{
-        ComponentWithState as ModelComponentWithState, Snapshot as ModelSnapshot,
-        StateSyncMessage as ModelStateSyncMessage,
-    },
-    BlockHeader, FeedMessage as ModelFeedMessage, HeaderLike, SynchronizerState,
-};
+use crate::feed::{self as feed_model, synchronizer, BlockHeader, HeaderLike, SynchronizerState};
```
louise-poole, line 22: not sure the feed -> feed_model alias was really necessary? It's only used to qualify `FeedMessage`... I'd rather we just import `FeedMessage` and skip this.

louise-poole (approved):
Just come very minor comments. Otherwise looks good to me!

crates/tycho-client/src/feed/dto.rs
kayibal: I think since we are in the dto module, it's fine to not qualify the local structs as DTO and instead ensure the model/crate scoped structs are qualified clearly.

crates/tycho-client/src/feed/dto.rs
kayibal: I think it's needed because in here we are defining the local `dto:FeedMessage`, see my comment above but we shouldn't be scoping it.

crates/tycho-client/src/feed/synchronizer.rs
kayibal: Yeah it doesn't implement default, probably because we didn't want to add a default since it carries an id and there is no method to set it afterwards. This is going away soon anyway because these models attributes will be made private.

PR #1000, opened 2026-05-08
feat: add metric rfq integration

zach030:


crates/tycho-execution/contracts/src/executors/MetricExecutor.sol
```diff
@@ -0,0 +1,234 @@
...
+    uint256 private constant _ORACLE_UPDATE_HEADER_LENGTH = 24;
+    uint256 private constant _INT128_MAX = uint256(uint128(type(int128).max));
+
+    bytes4 private constant _METRIC_CALLBACK_SELECTOR = 0xc3251075;
+    // The docs currently mention both names; accept the older selector while
+    // Metric finalizes the public callback naming.
+    bytes4 private constant _COOL_CALLBACK_SELECTOR = 0xa4b618b2;
+
+    bytes32 private constant _CURRENT_METRIC_POOL_SLOT =
+        0x680ea5c1857f782d9caf44a6f1020e0d3bb55355ec15df645080b605b3154b99;
```
zizou0x: How did you chose this?

crates/tycho-simulation/src/rfq/protocols/metric/client_builder.rs
```diff
@@ -0,0 +1,86 @@
...
+}
+
+impl MetricClientBuilder {
+    pub fn new(chain: Chain) -> Self {
+        Self {
+            chain,
+            tokens: HashSet::new(),
+            tvl: 0.0,
+            quote_tokens: None,
+            base_url: "http://54.199.103.16:8080".to_string(),
```
zizou0x: What is this URL?

crates/tycho-execution/contracts/src/executors/MetricExecutor.sol
```diff
@@ -0,0 +1,234 @@
...
+            address tokenIn,
+            address tokenOut,
+            bool outputToRouter
+        )
+    {
+        _validateDataLength(data);
+        tokenIn = address(bytes20(data[0:20]));
+        tokenOut = address(bytes20(data[20:40]));
+        return (
+            TransferManager.TransferType.None,
```
zizou0x, line 149: Missing context here, will this be used or it's just default params here? What does `None` means?

Are we paying back with callback or before? Would be nice to document this on the contract IMO

crates/tycho-simulation/src/rfq/protocols/metric/state.rs
```diff
@@ -0,0 +1,479 @@
...
+        let direction = self.direction(&token_in.address, &token_out.address)?;
+        let amount_in_human = amount_in.to_f64().ok_or_else(|| {
+            SimulationError::RecoverableError("Can't convert amount in to f64".into())
+        })? / 10_f64.powi(token_in.decimals as i32);
+
+        // Stream prices are only indicative; execution asks Metric for a fresh quote later.
+        let (amount_out_human, max_output) = match direction {
+            MetricDirection::ZeroForOne => {
+                let price = self.bid_ask.bid_price()?;
+                (amount_in_human * price, self.bid_ask.total_token1_available()?)
```
zizou0x, line 178: There is no price impact depending on the `amount_in`?

crates/tycho-simulation/src/rfq/protocols/metric/client_builder.rs
zach030: This is metric API [endpoint](https://oracle-based-omm.gitbook.io/metric/RSm94m71kqtGICv4iKRj/developers/api#metadata-endpoint)

crates/tycho-execution/contracts/src/executors/MetricExecutor.sol
zach030: Metric is callback-based, so we must skip the pre-swap transfer in `getTransferData()` and pay the pool only during the callback via `getCallbackTransferData()`. For more context, the tycho execution docs describe this callback-based executor flow [here](https://github.com/propeller-heads/tycho-indexer/blob/main/crates/tycho-execution/CLAUDE.md#executor-flow-callbacks--output-verification).

crates/tycho-execution/contracts/src/executors/MetricExecutor.sol
zach030: Updated, it’s `keccak256("MetricExecutor#CURRENT_METRIC_POOL")`.

crates/tycho-simulation/src/rfq/protocols/metric/state.rs
zach030: I went back to the docs and realized I had missed the depth field from the `/bid_ask `endpoint.

Metric exposes orderbook-like depth bins on both the bid and ask sides. The volume is cumulative across bins, and within each bin the price follows a linear segment instead of staying flat. 

I updated the implementation to handle those bins properly, and added comments with the formulas used for calculations.

You can also refer to the Metric docs for the depth description [here].(https://oracle-based-omm.gitbook.io/metric/RSm94m71kqtGICv4iKRj/developers/api#depth)

crates/tycho-execution/src/encoding/evm/swap_encoder/metric.rs
```diff
@@ -0,0 +1,387 @@
...
+        let token_out = swap.token_out().address.clone();
+        bytes_to_address(&token_in)?;
+        bytes_to_address(&token_out)?;
+
+        let pool = contract_address(swap, 0, "pool")?;
+        let metric_router = contract_address(swap, 2, "Metric router")?;
+        let zero_for_one = zero_for_one(swap)?;
+        let price_limit_x64 =
+            if zero_for_one { Bytes::zero(32) } else { u256_bytes(&BigUint::from(u128::MAX)) };
+        let oracle_update = if self.request_oracle_update {
```
tvinagre: Does this mean that, on Mainnet, the user will always pay for its transaction + the Oracle Update TX? If that's the case, can't we make this configurable?

tvinagre: Or, it would be nice to try to swap without the oracle update and, if the TX reverts because the oracle didn't update, send the oracle update. Not sure how much that would add in gas or if that's feasible

crates/tycho-simulation/src/rfq/protocols/metric/client.rs
```diff
@@ -0,0 +1,727 @@
...
+                .to_vec()
+                .into(),
+        );
+        attributes.insert(
+            "depth".to_string(),
+            serde_json::to_vec(&bid_ask.depth)
+                .unwrap_or_default()
+                .into(),
+        );
+
```
tvinagre, line 184:
```
let entries: [(&str, Vec<u8>); 9] = [
    ("bid_adj", bid_ask.bid_adj.as_bytes().to_vec()),
    ("ask_adj", bid_ask.ask_adj.as_bytes().to_vec()),
    ("total_token0_available", bid_ask.total_token0_available.as_bytes().to_vec()),
    ("total_token1_available", bid_ask.total_token1_available.as_bytes().to_vec()),
    ("latest_block", bid_ask.latest_block.to_string().into_bytes()),
    ("block_ts", bid_ask.block_ts.to_string().into_bytes()),
    ("server_ts", bid_ask.server_ts.to_string().into_bytes()),
    ("quote_expiration", bid_ask.quote_expiration.to_string().into_bytes()),
    ("depth", serde_json::to_vec(&bid_ask.depth).unwrap_or_default()),
];

for (key, bytes) in entries {
    attributes.insert(key.to_string(), bytes.into());
}
```

is easier to read IMO - what do u think?

crates/tycho-simulation/src/rfq/protocols/metric/client.rs
```diff
@@ -0,0 +1,727 @@
...
+        response.json().await.map_err(|e| {
+            RFQError::ParsingError(format!("Failed to parse Metric metadata response: {e}"))
+        })
+    }
+
+    async fn fetch_bid_ask(&self, pool: &Bytes) -> Result<MetricBidAskResponse, RFQError> {
+        let endpoint =
+            format!("{}/{}/bid_ask", self.chain_endpoint, bytes_to_address_string(pool)?);
+        let http_client = Client::new();
```
tvinagre: By recreating the client here every time, you need to always redo TLS handshake and that adds latency. If you create the client once and reuse it you will get a performance increase on the requests

crates/tycho-simulation/src/rfq/protocols/metric/client.rs
```diff
@@ -0,0 +1,727 @@
...
+            amount_in: params.amount_in.clone(),
+            amount_out,
+            quote_attributes,
+        })
+    }
+}
+
+#[async_trait]
+impl RFQClient for MetricClient {
+    fn stream(
```
tvinagre, line 383: Shouldn't this be async fn?

crates/tycho-simulation/src/rfq/protocols/metric/client.rs
```diff
@@ -0,0 +1,727 @@
...
+    Ok(calldata.into())
+}
+
+fn biguint_decimal_to_u256(value: &str) -> Result<U256, RFQError> {
+    let value = BigUint::from_str(value)
+        .map_err(|_| RFQError::ParsingError(format!("Failed to parse uint value: {value}")))?;
+    Ok(biguint_to_u256(&value))
+}
+
+fn stable_decimals(address: &Bytes) -> Option<u8> {
```
tvinagre: Not a big fan of this - if the RFQ adds a new stable we'd need to release a new Tycho-Simulation version? 
Can we use Tycho data here?

crates/tycho-simulation/src/rfq/protocols/metric/client_builder.rs
tvinagre: Can we make this configurable somehow? If they change this URL - which is not unlikely, we'd need to release another version and users won't be able to fix until we act

crates/tycho-simulation/src/rfq/protocols/metric/state.rs
```diff
@@ -0,0 +1,864 @@
...
+                .ok_or_else(|| {
+                    SimulationError::RecoverableError("Can't convert amount out to BigUint".into())
+                })?;
+        let capped_amount = amount_out
+            .clone()
+            .min(effective_max_output.clone());
+        let res = GetAmountOutResult {
+            amount: capped_amount.clone(),
+            gas: BigUint::from(170_000u64),
+            new_state: self.clone_box(),
```
tvinagre, line 205: Shouldn't we calculate the depth impact of this swap and update the `new_state` with it?

crates/tycho-simulation/src/rfq/protocols/metric/client_builder.rs
zach030: This is configurable now: users can override it with `.base_url(...)` [function](https://github.com/propeller-heads/tycho-indexer/blob/8da897e118420b99d6b8f88389acd65a69a91df7/crates/tycho-simulation/src/rfq/protocols/metric/client_builder.rs#L67-L70), and I also added [env config](https://github.com/propeller-heads/tycho-indexer/blob/8da897e118420b99d6b8f88389acd65a69a91df7/crates/tycho-simulation/src/rfq/constants.rs#L72-L84) for the default URL.

crates/tycho-simulation/src/rfq/protocols/metric/client.rs
zach030: fixed, now the client is cached with `OnceLock`, so we create it once and reuse it across requests.

crates/tycho-simulation/src/rfq/protocols/metric/client.rs
zach030: `stream` returns a `BoxStream` directly and does the async work inside the stream

crates/tycho-simulation/src/rfq/protocols/metric/client.rs
zach030: yes, agreed. I [changed](https://github.com/propeller-heads/tycho-indexer/blob/8da897e118420b99d6b8f88389acd65a69a91df7/crates/tycho-simulation/src/rfq/protocols/metric/client_builder.rs#L48-L55) this so we can pass token metadata in from tycho data via `.token_metadata(...)`.

crates/tycho-simulation/src/rfq/protocols/metric/state.rs
zach030: Hum, I’m not sure how a Metric swap should update the [state](https://github.com/propeller-heads/tycho-indexer/blob/f107dfea91cf3aae4aa9c3ee50c3618c53ea65fa/crates/tycho-simulation/src/rfq/protocols/metric/state.rs#L27-L28), and the other RFQ sims also leave `new_state` unchanged.

crates/tycho-execution/src/encoding/evm/swap_encoder/metric.rs
```diff
@@ -263,25 +339,31 @@ mod tests {
...
-            request_oracle_update.to_string(),
-        )]))
+    fn metric_config(oracle_address: &Bytes) -> HashMap<String, String> {
+        HashMap::from([(ORACLE_ADDRESS_CONFIG.to_string(), oracle_address.to_string())])
+    }
+
+    fn encoder(oracle_address: &Bytes) -> MetricSwapEncoder {
+        MetricSwapEncoder::new(
+            Bytes::zero(20),
+            Chain::Ethereum,
```
tvinagre: Should chain be hardcoded here?

crates/tycho-simulation/src/rfq/protocols/metric/client.rs
```diff
@@ -533,29 +613,86 @@ fn biguint_decimal_to_u256(value: &str) -> Result<U256, RFQError> {
...
+) -> Option<f64> {
+    let mut best: Option<(f64, f64)> = None;
+
+    for (pool, bid_ask) in pool_quotes {
+        let Some(mid_price) = metric_mid_price(bid_ask) else {
+            continue;
+        };
+        let candidate = if &pool.token0 == token && quote_tokens.contains(&pool.token1) {
+            let Some(quote_token) = token_metadata.get(&pool.token1) else {
+                continue;
```
tvinagre, line 555: Aren't these `else` cases an error? Shouldn't we log something here?

crates/tycho-execution/src/encoding/evm/swap_encoder/metric.rs
zach030: this is hardcoded for testing, and the test tokens are also on mainnet, so I think it should be fine.

crates/tycho-simulation/src/rfq/protocols/metric/client.rs
zach030: fixed, I added some [logs](https://github.com/propeller-heads/tycho-indexer/commit/5e034a230bc185b7bc826b44382640a6b29292f3) when there's no candidate.

crates/tycho-execution/config/protocol_specific_addresses.json
```diff
@@ -14,6 +14,10 @@
   "base": {},
   "arbitrum": {},
   "polygon": {},
-  "unichain": {},
+  "unichain": {
+    "vm:curve": {
+      "native_token_address": "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE"
+    }
+  },
```
zizou0x: What is this?

crates/tycho-execution/contracts/src/executors/MetricExecutor.sol
```diff
@@ -0,0 +1,291 @@
...
+        if (amountIn > _INT128_MAX) {
+            revert MetricExecutor__AmountInTooLarge();
+        }
+
+        if (oracleUpdateMode == OracleUpdateMode.Always) {
+            // Always mode updates the configured Metric oracle before the first swap.
+            // Keep the target on the immutable instead of calldata since this executor runs
+            // through TychoRouter delegatecall.
+            // slither-disable-next-line unused-return
+            oracle.functionCall(oracleCalldata);
```
zizou0x: Is this safe? @tamaralipows

crates/tycho-integration-test/src/main.rs
```diff
@@ -88,6 +88,10 @@ struct Cli {
     #[arg(long, default_value_t = false)]
     disable_rfq: bool,
 
+    /// Run PAMM RFQ protocols.
+    #[arg(long, default_value_t = false)]
```
zizou0x: Why false by default?

crates/tycho-simulation/examples/rfq_quickstart/main.rs
```diff
@@ -70,6 +71,9 @@ struct Cli {
     tvl_threshold: f64,
     #[arg(long, default_value = "ethereum")]
     chain: Chain,
+    /// Run PAMM RFQ protocols.
+    #[arg(long, default_value_t = false)]
+    run_pamm_protocols: bool,
```
zizou0x: Same here, shouldn't we make this true by default?

crates/tycho-simulation/src/rfq/protocols/metric/client.rs
```diff
@@ -0,0 +1,949 @@
...
+    chain_endpoint: String,
+    tokens: HashSet<Bytes>,
+    #[serde(default)]
+    token_metadata: HashMap<Bytes, Token>,
+    tvl: f64,
+    quote_tokens: HashSet<Bytes>,
+    #[serde(skip_serializing, default)]
+    secret_key: Option<String>,
+    #[serde(skip, default = "OnceLock::new")]
+    http_client: OnceLock<Client>,
```
zizou0x: Why is it a OnceLock?

zach030: just to reuse the same `Client` across requests, address this [comment](https://github.com/propeller-heads/tycho/pull/1000#discussion_r3242522690)

zizou0x: Couldn't we reuse it if it was just `http_client: Client`?

crates/tycho-simulation/src/rfq/protocols/metric/state.rs
```diff
@@ -0,0 +1,835 @@
...
+            false
+        }
+    }
+}
+
+fn depth_max_output(bins: &[MetricDepthBin]) -> Result<Option<BigUint>, SimulationError> {
+    let mut max_output = None;
+    for bin in bins {
+        max_output = Some(bin.cumulative_volume()?);
+    }
```
die-herdplatte: Is that done to catch parsing errors or why not just use `cumulative_volume` from the last bin?

crates/tycho-simulation/src/rfq/protocols/metric/client.rs
die-herdplatte: I think so too.
Also, currently each component creates its own `Client` in `MetricState::try_from_with_header`.
To make them share a `Client` (and reuse the connections to the same host), you could use
```rust
static METRIC_HTTP_CLIENT: LazyLock<Client> = LazyLock::new(Client::new);
```

crates/tycho-execution/contracts/src/executors/MetricExecutor.sol
```diff
@@ -0,0 +1,306 @@
...
+    function _updateOracle(bytes calldata oracleArgs) internal {
+        (
+            address feedCreator,
+            uint256 deadline,
+            uint256 newSlotValue,
+            bytes memory signature
+        ) = abi.decode(oracleArgs, (address, uint256, uint256, bytes));
+
+        oracle.updateBySignature(feedCreator, deadline, newSlotValue, signature);
+    }
```
die-herdplatte, line 275: I think the Oracle args could either be passed in a packed form (to the executor) or be passed completely (including the selector) such that the executor can just forward it.
Right now the decoding is mostly for readability reasons (but seeing Metric Oracle interfaces isn't really required to understand the rest of the executor imo) and verification (ABI-coder should catch this on the Oracle side and the length is already verified in the executor).

zach030: Thanks! It’s mainly for safety. Since the oracle is unverified, we shouldn’t allow arbitrary args or arbitrary function calls here. The ABI encoder checks the args returned by the oracle call, but we can’t guarantee those args won’t be changed before the executor is called.

crates/tycho-simulation/src/rfq/protocols/metric/state.rs
zach030: Thanks! I’ve updated it to use `cumulative_volume` from the last bin.

crates/tycho-simulation/src/rfq/protocols/metric/client.rs
zach030: Thanks! I’ve updated it to share the `Client` now.

zizou0x (approved):

PR #1001, opened 2026-05-08
feat(tycho-executor): extend interface for convenience

louise-poole:
- Add a `new_with_defaults` constructor for `SwapEncoderRegistry`. This replaces the lessor intuitive `SwapEncoderRegistry::new(Chain::Ethereum).add_default_encoders(None)`. NOTE: ideally we should change `add_default_encoders`'s param to not be optional anymore, but this is a breaking change and isn't of any major consequence so I decided to leave it.
-  Add a `ClientFeeParams` helper struct

### DIFF
--- a/crates/tycho-execution/src/encoding/models.rs
+++ b/crates/tycho-execution/src/encoding/models.rs
@@ -1,5 +1,7 @@
 use std::sync::Arc;
 
+#[cfg(feature = "evm")]
+use alloy::primitives::{Address, U256};
 use clap::ValueEnum;
 use num_bigint::BigUint;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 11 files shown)


crates/tycho-execution/src/encoding/models.rs
```diff
@@ -34,6 +36,50 @@ pub enum UserTransferType {
...
+    pub client_fee_receiver: Bytes,
+    /// Maximum amount the client will contribute from their vault if slippage reduces the output
+    /// below `min_amount_out`.
+    #[serde(with = "biguint_string")]
+    pub max_client_contribution: BigUint,
+    /// Deadline for the fee signature as a unix timestamp.
+    #[serde(with = "biguint_string")]
+    pub deadline: BigUint,
+    /// EIP-712 signature over the fee parameters and swap intent.
+    pub client_signature: Bytes,
```
dianacarvalho1: not a huge fan that the attributes are all public like this. Can we have them private, have a new method and then setter methods please? The super mandatory attributes that should be in new are `client_fee_receiver`, `client_signature` and `deadline`. The other two (`client_fee_bps`  and `max_client_contribution`) depend on the use case of the user.

dianacarvalho1 (commented):
Thank you @louise-poole ! I have one request!

crates/tycho-execution/src/encoding/models.rs
louise-poole: Can you clarify for me why only those are mandatory? What's the point of setting `client_fee_receiver` when you don't set `client_fee_bps` and are taking 0 fees?

dianacarvalho1: great question madam. and the answer is: we are using the `client_fee_receiver` to see if the client has special fees in the router in the first place. so it might happen that the client doesn't want to take a fee themselves but that they still want to identify themselves to get discounts (by setting the `client_fee_receiver`)

dianacarvalho1 (approved):
✅ ✅ ✅

PR #1005, opened 2026-05-11
feat: LiquidityParty updated adapter

timolson:
- The `swap()` interface has changed, so a new Executor will need to be deployed.
- The Substreams adapter handles `Killed` events by setting all pool token balances to zero. If we have to disable a pool, this workaround prevents solvers from attempting a swap that would revert, without us needing to support PC deletion.
- `swapToLimit()` is no longer supported and was removed from the SwapAdapter.
- `swapAmounts()` moved from `PartyPool` to `PartyInfo`
- The new test pool has only 3 tokens instead of 10 for faster test runs.
- Tightened bounds on `LiquidityPartyExecutorTest` with the exact amount outs expected.
- Other trivial updates to contract addresses, swap test amounts, etc. `PartyPoolSwapImpl` is called `PartyPoolExtraImpl` now.
- No major flags from slither just a loop warning
- I absolutely could not get integration testing to run because Substreams keeps rejecting my JWT token for some reason. I created fresh ones. I used the CLI. I even tried hitting the API directly. Nothing. This happened to me before, and it's hella frustrating. Yes, I'm using JWT from The Graph and not API tokens from substreams.dev. All of the tests for each component pass, so I'm assuming integration testing will work, too.


protocols/substreams/ethereum-liquidityparty/src/modules.rs
```diff
@@ -227,27 +227,46 @@ fn map_protocol_changes(
...
+                        Ok(token_addrs) => {
+                            for token in token_addrs {
+                                builder.add_balance_change(&BalanceChange {
+                                    token,
+                                    balance: vec![0],
+                                    component_id: component.id.as_bytes().to_vec(),
+                                });
+                            }
+                        }
+                    },
```
zach030: Please remove the `substreams::log`. 

I don’t think setting the TVL to zero is a good way to make this pool untradable in Tycho. If a pool is killed, you can update a mutable entity associated with it and filter it out in tycho-simulation. 
You can use the Curve [filter](https://github.com/propeller-heads/tycho-indexer/blob/057a88d3b553716b72703243d29ebfea8bcc02cb/crates/tycho-simulation/src/evm/protocol/filters.rs#L90-L104) as a reference. FYI, static attributes are immutable, so they’re not suitable for this kind of kill case.

protocols/substreams/ethereum-liquidityparty/src/modules.rs
zach030: Iterating over all ERC20 `Transfer` events would add a lot of indexing latency. Does the Liquidity Party pool emit any `Swap` / `Mint` / `Burn` events we could use instead to make the filtering and balance detection more efficient?

timolson: Yes, all of those. I originally had a full suite of our events and switched it out for Transfers last time... You'll notice that this part of the code was not changed in this update. I'll rewrite this to use our custom events.

protocols/substreams/ethereum-liquidityparty/src/modules.rs
timolson: Ok, will do.

Any plans in the future to have a PC delete? Seems like the system was designed with CRUD in mind... I thought I saw a DELETE event somewhere although in the previous review I got the impression it wasn't supported. I'll match whatever Curve does.

zach030: Yes, it was designed around CRUD, and supporting deletion would take quite a bit of work. I wouldn’t expect it anytime soon.

protocols/substreams/ethereum-liquidityparty/src/modules.rs
zach030: Yes, I didn’t notice that in my last review. Switching to the full suite of your events would be a big performance improvement.

timolson: Yah makes total sense. Done.

timolson:
- Removed all Substreams logging
- Removed excessive sanity checks in the interest of lower latency
- Balance tracking updated to use only Liquidity Party events. TVL excludes protocol fees sitting in the pool awaiting collection
- Killed pools set a dynamic attribute and are filtered out in `protocol_stream_processor.rs`. The Curve filter doesn't seem to be wired, so I followed the `balancer_v2_pool_filter` example.
- Killed pools no longer track balances or send any other Substreams messages. A killed pool's Substreams will report balances at killed time, even if they are subsequently withdrawn by LP's. (Kill still allows LP's to `burn()` but for Tycho purposes no more Substreams messages will be sent for the pool after the kill.)

I'm going to test the new balance tracking tomorrow. Our test pool already has an "exercise" series of transactions that invokes one of every operation, so I just need to run Substreams on that sequence and verify the final balances.

timolson:
I found a few more nits in our DEX I'd like to clean up as long as we're re-deploying. The current Tycho adapter should be good to review, but I will probably make one more update to the singleton contract addresses before we're really ready to merge. But no changes to API's or anything like that, just updated deployment addresses.

timolson:
I updated the contract addresses and start blocks. There was also an ABI touchup, but it doesn't affect anything in Tycho. That's all the updates I wanted to get in.

timolson:
Our auditor found an issue and we're going to need to redeploy again. Minor issues, nothing close to value extraction. I'd understand if you want to pause review until the audit is complete; should be next week. But the ABI's are stable for swaps and there have been no issues found with swap. The adapter code should be stable as-is except for future contract addr changes.

zach030:
> Our auditor found an issue and we're going to need to redeploy again. Minor issues, nothing close to value extraction. I'd understand if you want to pause review until the audit is complete; should be next week. But the ABI's are stable for swaps and there have been no issues found with swap. The adapter code should be stable as-is except for future contract addr changes.

Thanks for the heads up. I haven’t seen any other major issues so far, and I’ll take another look once the audit is done.

timolson:
Turns out we do need to make some pretty significant changes. It's gonna be a few weeks, and I'll probably write a native simulator instead of the EVM simulator. We're moving the bulk of the LMSR solving off-chain and just using an on-chain validation approach instead. Basically we found regimes where our on-chain math wasn't mathing, and the fix is too much gas. So it will be a swap() ABI change to use an offchain solver + on-chain validator model. Don't expect a proper PR for a while.

zach030:
> Turns out we do need to make some pretty significant changes. It's gonna be a few weeks, and I'll probably write a native simulator instead of the EVM simulator. We're moving the bulk of the LMSR solving off-chain and just using an on-chain validation approach instead. Basically we found regimes where our on-chain math wasn't mathing, and the fix is too much gas. So it will be a swap() ABI change to use an offchain solver + on-chain validator model. Don't expect a proper PR for a while.

Thanks for the heads up! Sounds like you’re planning a native integration?
Before you start integration, please read this integration [guide](https://docs.propellerheads.xyz/tycho/for-dexs/protocol-integration/indexing#integration-modes) first to make sure Tycho supports it.
Feel free to ping me if you have any questions.

timolson:
Our audit will likely wrap up this week, so I should have an updated adapter ready for your review next week.

There will not be any significant changes to the adapter. Should just be contract addresses, start blocks, etc.

timolson:
Not going to be this week; we have a little more work to do.

timolson:
We've completed an audit and redeployed, and our integration is ready for review. There are only minor differences:

- Support for exact-out (BUY) added
- Flash loans removed
- Our PartyInfo marginal price reports BUY convention so we take the reciprocal for Tycho SELL convention.
- PartyInfo helper contract is now required by the SwapAdapter
- Delegate-call libraries renamed to ...ExtraImpl1 and ...ExtraImpl2
- The usual start block & contract address changes.

I *was* able to run the integration test this time, using the same Substreams API keys that failed previously :shrug:

zach030:
> We've completed an audit and redeployed, and our integration is ready for review. There are only minor differences:
> 
> * Support for exact-out (BUY) added
> * Flash loans removed
> * Our PartyInfo marginal price reports BUY convention so we take the reciprocal for Tycho SELL convention.
> * PartyInfo helper contract is now required by the SwapAdapter
> * Delegate-call libraries renamed to ...ExtraImpl1 and ...ExtraImpl2
> * The usual start block & contract address changes.
> 
> I _was_ able to run the integration test this time, using the same Substreams API keys that failed previously 🤷

Thanks for the update. I’ll start reviewing it.

protocols/substreams/ethereum-liquidityparty/bin/generate_abi
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-LMSR_HOME=../../../lmsr-amm
+LMSR_HOME=../../../../lmsr-amm
 CHAIN_ID=1
```
zach030: Is this code needed?

protocols/substreams/ethereum-liquidityparty/src/modules.rs
```diff
@@ -227,27 +252,32 @@ fn map_protocol_changes(
...
                 .for_each(|component| {
-                    builder.add_protocol_component(component);
+                    builder.add_entity_change(&EntityChanges {
+                        component_id: component.id.clone(),
+                        attributes: vec![Attribute {
+                            name: "killed".to_string(),
+                            value: vec![1u8],
+                            change: ChangeType::Update.into(),
+                        }],
+                    });
```
zach030, line 261: Can a killed pool be restored later?
We already have a generic `paused` attribute that supports update and delete. If this behaves similarly to `paused`, we could probably reuse it here. There’s already a [helper](https://github.com/propeller-heads/tycho/blob/632ee0153340f6ee4ecd2770f524c5b1521f450e/protocols/substreams/crates/tycho-substreams/src/models.rs#L104-L136) for it, and `tycho-simulation` also has a generic filter for paused pools.

protocols/substreams/ethereum-liquidityparty/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "ethereum-liquidityparty"
-version = "0.1.1"
+version = "0.1.3"
```
zach030: why `0.1.3`?

zach030 (commented):
Thanks! Overall it looks good to me. I just left a few questions, please take a look.

protocols/substreams/ethereum-liquidityparty/bin/generate_abi
timolson: I've kept it around so we can re-create the ABI files from original source, but will remove it.

protocols/substreams/ethereum-liquidityparty/src/modules.rs
timolson: Killing is permanent and "should" be a delete operation. Last cycle you told me to write a filter ala Curve.

LMSR pools do not last forever and will sometimes need to be killed for economic reasons. Killed also serves as a security switch if any vulnerability is found.

Do you want me to switch to pause now? Pools cannot be "unpaused."

protocols/substreams/ethereum-liquidityparty/Cargo.toml
timolson: AI slop I didn't catch, sorry. Will use 0.1.2.

timolson:
I made the two minor touchups but didn't switch to `paused` unless you tell me to. Killing is irreversible, and I don't think re-using the existing filter would improve latency at all, since filters are already scoped per-protocol.

protocols/substreams/ethereum-liquidityparty/src/modules.rs
zach030: Ok, looks good to me. Let’s keep using `killed`.

timolson:
Anything else? Ok to merge? When is your next release cycle?

zach030:
> Anything else? Ok to merge? When is your next release cycle?

Sorry! I've got quite a few integrations going on right now. I'll sync and test LiquidityParty locally today.

timolson:
Ok, thanks for the update. I was just wondering what had happened since we seemed to be clear last week. I've updated the branch again.

Note that there was a fuzz test failure previously in Etherfi:

```
Encountered 1 failing test in test/EtherfiAdapter.t.sol:EtherfiAdapterTest
[FAIL: TRANSFER_AMOUNT_EXCEEDS_BALANCE; counterexample: calldata=0x23faaaa900000000000000000000000000000000000000000000313f6b5f1bd314f696d50000000000000000000000000000000000000000000000000000000000000001 args=[232565839470534566778581 [2.325e23], true]] testSwapFuzzEtherfiEethWeEth(uint256,bool) (runs: 241, μ: 262441, ~: 278908)
```

I'm assuming this is an Etherfi problem that should be addressed that doesn't occur during every fuzz, so it may be worth recording this fuzz instance and notifying the Etherfi maintainer.

zach030:
> Ok, thanks for the update. I was just wondering what had happened since we seemed to be clear last week. I've updated the branch again.
> 
> Note that there was a fuzz test failure previously in Etherfi:
> 
> ```
> Encountered 1 failing test in test/EtherfiAdapter.t.sol:EtherfiAdapterTest
> [FAIL: TRANSFER_AMOUNT_EXCEEDS_BALANCE; counterexample: calldata=0x23faaaa900000000000000000000000000000000000000000000313f6b5f1bd314f696d50000000000000000000000000000000000000000000000000000000000000001 args=[232565839470534566778581 [2.325e23], true]] testSwapFuzzEtherfiEethWeEth(uint256,bool) (runs: 241, μ: 262441, ~: 278908)
> ```
> 
> I'm assuming this is an Etherfi problem that should be addressed that doesn't occur during every fuzz, so it may be worth recording this fuzz instance and notifying the Etherfi maintainer.

Yes, this test isn't related to your changes — we're working on a fix. I'm still running a full sync locally, which takes some time. Will keep you posted.

zach030:
> > Ok, thanks for the update. I was just wondering what had happened since we seemed to be clear last week. I've updated the branch again.
> > Note that there was a fuzz test failure previously in Etherfi:
> > ```
> > Encountered 1 failing test in test/EtherfiAdapter.t.sol:EtherfiAdapterTest
> > [FAIL: TRANSFER_AMOUNT_EXCEEDS_BALANCE; counterexample: calldata=0x23faaaa900000000000000000000000000000000000000000000313f6b5f1bd314f696d50000000000000000000000000000000000000000000000000000000000000001 args=[232565839470534566778581 [2.325e23], true]] testSwapFuzzEtherfiEethWeEth(uint256,bool) (runs: 241, μ: 262441, ~: 278908)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'm assuming this is an Etherfi problem that should be addressed that doesn't occur during every fuzz, so it may be worth recording this fuzz instance and notifying the Etherfi maintainer.
> 
> Yes, this test isn't related to your changes — we're working on a fix. I'm still running a full sync locally, which takes some time. Will keep you posted.

Hey @timolson! Just a heads up — our RPC seems to be failing to fetch historical blocks in a certain range (unfortunately, the LiquidityParty start block falls right in that range). It's currently blocking my testing and syncing. I'm in touch with the Chainstack team to sort it out.

timolson:
Thanks for the heads up. If this is regarding debug_storageRangeAt, this method is really problematic. It’s only supported by Erigon and no other client. I’ve written several messages in the DEX Telegram group about it.

zach030:
> Thanks for the heads up. If this is regarding debug_storageRangeAt, this method is really problematic. It’s only supported by Erigon and no other client. I’ve written several messages in the DEX Telegram group about it.

We've fixed the RPC and I'm syncing it locally right now.

zach030:
Hey @timolson ! I synced the liquidity party locally and ran the tests — found a few issues:

1. Please update `crates/tycho-simulation/src/evm/protocol/vm/assets/LiquidityPartySwapAdapter.evm.runtime` to match the latest version of the LiquidityPartySwapAdapter contract.
2. I got this error during simulation: `MissingAccount(0x5fe46969172b707cb58957c7bdc7e255535f5faf)`. This is the bfStore contract, it gets used during swaps, so it needs to be attached to component.contracts.
3. The start block should be 25301738, this is the earliest block where PartyPoolExtraImpl2 was [deployed](https://etherscan.io/tx/0xa2e1f9398db0e6a7fa89a912fc7c644627c592645cdfad6b36da9b6cbd4b84ee).

Can you fix these and run the integration test to verify it works?

timolson:
- Updated the runtime bytecode
- Added the immutables SSTORE2 contract to the protocol component contracts
- Pushed the start block back to 25301734, which includes ExtraImpl1, just in case.

I could not get the integration test to run. Chainnodes is reporting "history not available," which is I guess what you ran into and they "fixed?" Alchemy doesn't support debug_storageRangeAt, nor does our own Reth node.

timolson:
I've reached out to ChainNodes support...

If you have 10 minutes, it shouldn't be hard for you to give the latest push a try.

zach030:
> I've reached out to ChainNodes support...
> 
> If you have 10 minutes, it shouldn't be hard for you to give the latest push a try.

Thanks! I'm syncing locally right now — it's a full sync so it'll take some time. I'll keep you posted once I have the results.

zach030:
Hey @timolson ! I ran the integration test again — the previous issues are fixed, but I hit a new one.

During the `get_amount_out` simulation, we get `unhandled: insufficient funds`. It looks like the limit isn't set correctly:

```solidity
// input limits based on a maximum target slippage, we merely return a
// fixed fraction of the input token's current inventory as a practical
// limit.
limits[0] = IERC20(sellToken).balanceOf(pool);

// output token limit: the pool's current balance
limits[1] = IERC20(buyToken).balanceOf(pool);
```

The sellToken limit (limits[0]) isn't an amount that can actually be traded on-chain — swapping it reverts, and only much smaller amounts go through. So the input limit seems too large for what the pool can actually fill (my guess is the resulting output exceeds what the pool can pay out, but I haven't dug into the algorithm).

I'm not familiar with the internals here, so just a suggestion for direction: could we instead determine the max output amount first, then use something like `swapAmountsForExactOutput` to derive the corresponding max input for limits[0]? That way the input/output pair stays self-consistent.

I've attached the full simulation log (including the vm traces), feel free to take a look.
[liquidityparty.log](https://github.com/user-attachments/files/29624261/liquidityparty.log)

timolson:
Thanks for debugging that, Zach! I believe your analysis is exactly correct: we do not emit "insufficient balance" ourselves, and it must be coming from the output token code.

I've adjusted _both_ our limits to target 1/2 of the output balance, which should be plenty for any reasonable slippage, and use the exact-out quoter to set the input limit.

zach030:
Hey @timolson ! Gave it another run, Two things to share.
1. A small heads-up: after updating the adapter contract, please also update the `LiquidityPartySwapAdapter.evm.runtime` file.
2. The fix helped but isn't complete. One sell direction now prices correctly, but most still revert with unhandled: insufficient funds (pool = USDC/WETH/AAVE, 6 directions)

Direction | get_amount_out | Notes
-- | -- | --
WETH → USDC | ✅ passes | 
WETH → AAVE | ✅ passes | 
USDC → WETH | ❌ insufficient funds | input 2000463 (~62% of USDC reserve)
USDC → AAVE | ❌ insufficient funds | input 2034399
AAVE → USDC | ❌ insufficient funds | input 30812809527293704
AAVE → WETH | ❌ insufficient funds | input 31949333840898745

For the failing ones, even with limits[0] from `swapAmountsForExactOutput(buyBalance / 2)`, the input still reverts on the actual swap — looks like the exact-out quote and the exact-in swap aren't fully consistent.

Btw, did you manage to fix the RPC issue? You should be able to reproduce this by running the protocol testing.

Full log with vm traces attached. Thanks!
[liquidityparty_integration_test.log](https://github.com/user-attachments/files/29747486/liquidityparty_integration_test.log)

timolson:
This is actually an inconsistency between the test harness and the Substreams adapter.

Our pool retains protocol fees in a separate accounting bucket (not part of TVL) until a collection transaction. This saves swap gas by not transferring small protocol fee amounts on every swap, but holding onto them and aggregating larger amounts before they are swept out of the pool in a separate collection transaction.

I wrote our Substreams adapter to report the pool's tradable TVL, which is slightly smaller than the pool's balances due to uncollected protocol fees still sitting in the pool.

The protocol test harness assumes that the Substreams balance is also the balance of the pool, and overrides balanceOf() to report the Substreams number, which in our case is NOT the actual pool balance.

This causes our validation of the input amount to fail by the size of the uncollected protocol fees, throwing an `insufficient funds` error for any token that has uncollected protocol fees sitting in the pool.

IMO the correct fix is to have the test harness set the value for the pool's actual token balance from an RPC query of the forked block, instead of assuming that the TVL reported by Substreams is the same as the pool's balance, which is not true for our pools.

P.S. Chainnodes is still broken and I pinged them again today. Thanks so much for running the integration test for me. If CN isn't fixed by next cycle, I'll write a separate Foundry test that mimics the protocol integration swaps and make sure that works.

zach030:
> This is actually an inconsistency between the test harness and the Substreams adapter.
> 
> Our pool retains protocol fees in a separate accounting bucket (not part of TVL) until a collection transaction. This saves swap gas by not transferring small protocol fee amounts on every swap, but holding onto them and aggregating larger amounts before they are swept out of the pool in a separate collection transaction.
> 
> I wrote our Substreams adapter to report the pool's tradable TVL, which is slightly smaller than the pool's balances due to uncollected protocol fees still sitting in the pool.
> 
> The protocol test harness assumes that the Substreams balance is also the balance of the pool, and overrides balanceOf() to report the Substreams number, which in our case is NOT the actual pool balance.
> 
> This causes our validation of the input amount to fail by the size of the uncollected protocol fees, throwing an `insufficient funds` error for any token that has uncollected protocol fees sitting in the pool.
> 
> IMO the correct fix is to have the test harness set the value for the pool's actual token balance from an RPC query of the forked block, instead of assuming that the TVL reported by Substreams is the same as the pool's balance, which is not true for our pools.
> 
> P.S. Chainnodes is still broken and I pinged them again today. Thanks so much for running the integration test for me. If CN isn't fixed by next cycle, I'll write a separate Foundry test that mimics the protocol integration swaps and make sure that works.

Thanks for the explanation!

IMO the Substreams adapter should just strictly mirror the ERC20 Transfer events, protocol fees can keep accumulating in the component_balance, and we only deduct them on a `CollectProtocolFee` event, just like we do for Uniswap V3 and similar protocols.

Was there another reason for reporting tradable TVL instead of the real balance? For a VM integration, the key thing is keeping VM state consistent with on-chain state. If a LiquidityParty contract calls `token.balanceOf()` and we return the tradable amount rather than the real one, that'll break things. Do you think switching to the real balance would fix this?

timolson:
I thought Tycho used balances strictly as TVL filters, but now I understand that you fake the balanceOf() call in your simulator. It cannot be both a TVL and a true balance.

The **only** way to get a true balance is to go back to tracking ERC20 Transfer events for the pool. Even if we add tracking of our fee collection event, _that wouldn't be good enough_. If someone makes a donation to the pool by sending an unsolicited amount of token (e.g. dust attack), there is no pool event in that donation tx. The next `balanceOf()` call will be wrong, and all simulations from that point forward will generate swaps that revert during the actual on-chain execution attempt.

The reason your UniV3 implementation still works is because Uniswap always calls balanceOf() both before and after the funding event, so it gets a "consistently wrong value" and the delta ends up being correct. Our pool uses a cached value for the "before" in order to save gas, so the discrepancy between the real on-chain pre-transfer balanceOf() and the faked post-transfer balanceOf() is causing the problem. Also note that Tycho's UniV3 simulations will not be wei-exact for the swap following a donation.

I'm going to assume your hard requirement is no changes to the simulator, and that Tycho assumes the Substreams balances are true balances, not TVL's. Our only option in that case is to filter all ERC20 Transaction events.

If more than one protocol is doing a Transfer event scan, it would be more efficient for your indexer to stop using Substreams as the source of balances and instead just handle all Transfer events itself, keeping rolling balances for any registered protocol component. One global scan of the Transfer events in a tx will always be more efficient than having each protocol do it independently.

timolson:
* Switched Substreams balance tracking to use ERC20 Transfer events
* Updated swap adapter runtime

Chainnodes is still broken :(  I couldn't write an independent test for this, because the issue depends on Tycho's simulator behavior.

timolson:
I can do it the other way with pool events if you care more about latency than correctness, but like I said, _it will break_ if there is ever any overpayment or token donation.

Also, Chainnodes is really not helping. What did you guys do to get the integration tests to work?

zach030:
Hey @timolson! If the LiquidityParty pool calls `erc20.balanceOf` internally, then I'd suggest we include the token Transfer events. But if it uses a cached value instead, I think using the pool events is more appropriate. The donate or dust you mentioned won't affect the contract's swap logic, and shouldn't be counted toward TVL either.

The earlier Chainstack RPC issue was that queries failed within certain block ranges, the Chainstack team has fixed it now. IIRC, the RPC on Chainstack's basic plan also supports the features we need for testing. 
I just verified — only archive nodes support this, elastic nodes don't :(

timolson:
I signed up for Chainstack but am getting nulls for debug_storageRangeAt. This is so frustrating. They must have done something special for your account to spin up an Erigon client. I now have support tickets with both Chainstack and Chainnodes.

We will need to use Transfer events, which is how the current branch is set up. It "should" work in its current state.

timolson:
Chainstack does _not_ in fact support debug_storageRangeAt as far as I can tell, definitely not in the basic plan. I've gone through 2 email cycles with their support so far. They told me to upgrade my plan to get a dedicated node, but they only seem to offer Reth not Erigon. I'm not sure how your setup is working?

I'm still absolutely stuck, unable to run an integration test. I've tried three different node providers and multiple support requests with each.

You've been very generous and helpful running the test for me so far... please one more time? I switched the Substreams adapter to use ERC20 Transfer events which "should" fix the final swap tests.

timolson:
:heavy_check_mark: Integration test passes, including all six simulated swap directions.

I was able to get the test running with Tatum.io which hosts Erigon exclusively. I've submitted a Tycho PR to update the protocol testing docs: https://github.com/propeller-heads/tycho/pull/1193

I also updated the runtime again. Didn't seem to be actually updated in the last commit.

Everything should work now. Please review!

zach030:
> ✔️ Integration test passes, including all six simulated swap directions.
> 
> I was able to get the test running with Tatum.io which hosts Erigon exclusively. I've submitted a Tycho PR to update the protocol testing docs: #1193
> 
> I also updated the runtime again. Didn't seem to be actually updated in the last commit.
> 
> Everything should work now. Please review!

Thanks so much! I'll do another local sync test.

timolson:
Are you running a final check? I'll sync the branch nightly.

zach030:
Hey @timolson! I've restarted syncing it locally. After adding the ERC20 transfer event parsing, the indexing speed seems to be affected, so it'll take a few days. I'll let you know how it goes.

zach030:
Hey @timolson — re-ran the integration test and did a wei-level on-chain cross-check of both the indexed state and the swap simulation. Sharing the pricing/state results below.

**Run:** 100 blocks, `25643749–25643848`, pool `0x75f7c9ea0f9f976ea2652ad2eead9f4a3fcafc8c` (5-stable LMSR pool: DAI / USDC / USDT / USDe / `0x6c3ea9…`).

**Result**
- `get_limits`: 2000 success / 0 failure
- `get_amount_out`: 2000 success / 0 failure
- No `insufficient funds`, no pricing reverts

## On-chain verification (redone independently at this run's blocks)

**State — indexed component balances vs on-chain `balances()` @ block 25643848: 5/5 exact (to the wei)**

| token | indexed | on-chain |
|---|--:|--:|
| USDC | 5001025210 | 5001025210 |
| USDT | 5003557529 | 5003557529 |
| USDe | 5002306063095086756221 | 5002306063095086756221 |
| 0x6c3ea9 | 5001765623 | 5001765623 |
| DAI | 5000000000000000000000 | 5000000000000000000000 |

Token order matches on-chain `allTokens()`, and `allProtocolFeesOwed = [0,0,0,0,0]` at this block. `getLimits` maxOutput equals exactly half the on-chain balance.

**Swap — simulated `get_amount_out` vs on-chain `swapAmounts` quoter: 7/7 exact (to the wei)**

| block | direction | sim | on-chain |
|---|---|--:|--:|
| 25643749 | USDC→USDT | 2501778764 | 2501778764 |
| 25643749 | USDC→0x6c3ea9 | 2500882811 | 2500882811 |
| 25643749 | USDC→DAI | 2500000000140274130119 | 2500000000140274130119 |
| 25643749 | USDT→USDC | 2500512605 | 2500512605 |
| 25643848 | 0x6c3ea9→USDe | 2501153032203170753392 | 2501153032203170753392 |
| 25643848 | DAI→USDC | 2500512604 | 2500512604 |
| 25643848 | DAI→0x6c3ea9 | 2500882810 | 2500882810 |

Covers both blocks, both 6- and 18-decimal tokens, and both directions.

## Conclusion

- **State is correct** — indexed balances match on-chain to the wei; the LMSR engine storage that drives pricing (`qInternal` / `kappa` / `sigma` / fees) must be correct too, otherwise the simulation couldn't match the on-chain quoter exactly.
- **Swap math is correct** — the adapter's revm output equals the pool's on-chain quote down to the smallest unit, in both directions and across decimals.

crates/tycho-execution/contracts/src/executors/LiquidityPartyExecutor.sol
zach030: `allProtocolFeesOwed` is declared but never used, we can drop it.

zach030, line 81: It'd be nice to add an explicit length check.

zach030 (commented):
Looks good overall, just a few nits to make it clearer.

timolson:
I pushed the cleanups.

I didn't realize you were going to test against the blockchain head with live pools... The `0x75f7c9ea0f9f976ea2652ad2eead9f4a3fcafc8c` pool is not our test pool and had zero swap activity and zero retained protocol fees. Therefore, it would not exercise the problem we found before with balances.

For completeness, test again with the active pool `0xf1C8A3e2435dc55212cc432859768D1ef25DDb7C` which has been getting flow from Kyber Swap. Success on `0xf1C8...` would be definitive. It has the same token composition as the pool you just tried, so you just need to change the pool address in your test and nothing else.

zach030:
> I pushed the cleanups.
> 
> I didn't realize you were going to test against the blockchain head with live pools... The `0x75f7c9ea0f9f976ea2652ad2eead9f4a3fcafc8c` pool is not our test pool and had zero swap activity and zero retained protocol fees. Therefore, it would not exercise the problem we found before with balances.
> 
> For completeness, test again with the active pool `0xf1C8A3e2435dc55212cc432859768D1ef25DDb7C` which has been getting flow from Kyber Swap. Success on `0xf1C8...` would be definitive. It has the same token composition as the pool you just tried, so you just need to change the pool address in your test and nothing else.

Thanks! This pool wasn't created during my last round of testing. I'll do another round of integration testing against this pool.

zach030:
Re-ran against the active pool `0xf1C8A3e2435dc55212cc432859768D1ef25DDb7C` as you suggested — this time with **execution enabled**, so the whole pipeline (state → pricing → on-chain execution) is covered. Since the executor isn't deployed yet, I patched the compiled `LiquidityPartyExecutor` runtime into the fork locally (bytecode override + executor activation on the router) — local test scaffolding only, not part of the PR.

This pool has non-zero retained protocol fees, so it exercises the balance case we were worried about.

## Result

```
vm:liquidityparty (pool 0xf1C8…, blocks 25648862–25648961)
  Unique Pools: 1
  get_limits:      2000 success / 0 failure
  get_amount_out:  2000 success / 0 failure
  Simulations:     2000 → Success 2000 (100%), Reverted 0, Setup failures 0
  Slippage:        avg 0.0000% (every one of the 2000 samples = 0)
```

## On-chain cross-check (block 25648862)

**Fees are actually charged** — the on-chain `swapAmounts` quoter returns a non-zero `outFee` in every direction, and the simulated `get_amount_out` matches the quoter's net output to the wei:

| direction | sim `get_amount_out` | on-chain `swapAmounts` | `outFee` |
|---|--:|--:|--:|
| USDC→USDT | 2796964769 | 2796964769 | 97898 |
| USDC→USDe | 1898496393168989641799 | 1898496393168989641799 | 125309032345288106 |
| USDC→DAI | 2711732498436200412895 | 2711732498436200412895 | 225092480046048457 |
| USDT→USDC | 2026259785 | 2026259785 | 70922 |

**State matches too** — the indexed component balances equal the on-chain ERC20 `balanceOf(pool)` to the wei for all 5 tokens, including the fee-bearing ones (indexed `balanceOf` = pool `balances()` + `allProtocolFeesOwed`).

## Conclusion

On a live pool **with retained protocol fees**: fees are charged (`outFee > 0`), execution succeeds 2000/2000 with zero slippage, and state + pricing + execution all line up with on-chain truth to the wei. The tradable-TVL-vs-`balanceOf` concern doesn't surface here — the indexed balance is the full physical balance, so the swap that pays out from the real balance matches the simulation exactly.

zach030:
@timolson Can you fix the CI issue?

timolson:
Sorry, I got in the habit of completely ignoring the CI since it always seems to break, at least for forks.

I updated the runtime. The other two issues don't look related to Liquidity Party.

zach030:
> Sorry, I got in the habit of completely ignoring the CI since it always seems to break, at least for forks.
> 
> I updated the runtime. The other two issues don't look related to Liquidity Party.

CI's V4 failure is just a stale calldata.txt, merging main updates the encoder but not the committed fixture, so it's still the old format. Please regenerate it with `cargo test -p tycho-execution` and commit the updated calldata.txt.

timolson:
:heavy_check_mark:

zach030 (approved):
Thanks, LGTM! @timolson Could you rebase on main so we can merge it?

timolson:
:tada:

timolson:
Thanks for all the effort, Zach!

PR #1006, opened 2026-05-11
fix(quickstart): add -p flag to avoid error from missing PostgreSQL

brunoguerios:
**Summary**
- Running `cargo run --release --example quickstart` from the workspace root fails with `ld: library 'pq' not found` on machines without PostgreSQL installed
- Cargo resolves the full workspace graph, pulling `pq-sys` → `diesel` → `tycho-storage`/`tycho-indexer` into the link step even though the quickstart has no database dependency
- Fix: add `-p tycho-simulation` to all documented commands so the build is scoped to that crate's dependency tree, which does not include `pq-sys`

**Alternative considered**
- Documenting `brew install postgresql` / `apt install postgresql` as a prerequisite. Rejected because the quickstart targets simulation consumers (solver/routing developers) who will never run the indexer and have no reason to install a PostgreSQL client. The `-p` flag is also the semantically correct Cargo idiom for building a single workspace member.

**Also includes**
- Added `TYCHO_API_KEY` to all code blocks — it was previously undocumented, causing a confusing `Invalid authentication` key error on first run


zizou0x (approved):
Nice catch! LGTM, thanks!

brunoguerios:
Hey @zizou0x - thanks for reviewing it! 
I brought the branch up to date with latest changes on main, so it's ready to be merged ✅ 
Even though you approved the changes, seems that a few workflows are still pending.
Let me know if there's anything else you require from me to get this through or if you'll take it from here.
Cheers 🍻

PR #1007, opened 2026-05-12
test(tycho-execution): Add Arbitrum execution tests

louise-poole:


crates/tycho-execution/config/protocol_specific_addresses.json
```diff
@@ -12,5 +12,10 @@
     }
   },
   "base": {},
-  "unichain": {}
+  "arbitrum": {},
+  "unichain": {
+    "vm:curve": {
+      "native_token_address": "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE"
```
tamaralipows: No longer necessary since we are always passing eeee to the executors now :) Not sure if I missed some doc update about this? I tried to keep them up to date at least

tamaralipows (approved):
Thanks a lot - I like the refactor you did here. I just have a smol thing to look into but the rest looks good!

crates/tycho-execution/config/protocol_specific_addresses.json
louise-poole: Ah yes! Brain fart - thanks! 🙏

PR #1009, opened 2026-05-12
feat(UniV4 execution): Allow swap when already unlocked

tamaralipows:


crates/tycho-execution/contracts/test/protocols/UniswapV4.t.sol
```diff
@@ -860,18 +863,294 @@ contract ExternalSettlerNoUnlockTest is TychoRouterTestSetup {
...
             0,
+            "Settler USDE used for loan repayment"
+        );
+    }
+
+    function testSettlementWithLoanSequentialV2ThenV4() public {
+        //        WETH ──(UniV2)──> WBTC ──(UniV4 skipUnlock)──> USDT
+        // External settler with loan + skipUnlock on a sequential route:
+        //   1. Take WETH loan from PM
+        //   2. Sync WBTC — the V4 hop's input, deposited into PM by the V2 hop
```
tamaralipows, line 884: Oh wow I didn't even expect this use case honestly, I thought they were only going to take a loan for the V4 input token. Well it's cool to know this is supported too then :o

crates/tycho-execution/contracts/test/protocols/UniswapV4.t.sol
```diff
@@ -860,18 +863,294 @@ contract ExternalSettlerNoUnlockTest is TychoRouterTestSetup {
...
+                address(settler),
+                noClientFee(),
+                pleEncode(swaps)
+            )
+        );
+
+        Loan[] memory loans = new Loan[](1);
+        loans[0] = Loan({token: USDE_ADDR, amount: loanAmount});
+
+        vm.expectRevert();
```
tamaralipows, line 1152: Curious what it reverts with. Internal v4 error? Could we add that?

crates/tycho-execution/src/encoding/evm/swap_encoder/uniswap_v4.rs
```diff
@@ -248,6 +237,37 @@ impl SwapEncoder for UniswapV4SwapEncoder {
...
+    /// external-settler flows that take a PM loan around the swap).
+    /// Defaults to `false`; producers can omit the field entirely
+    /// and the field is skipped from the serialized JSON when false.
+    #[serde(default, skip_serializing_if = "std::ops::Not::not")]
+    pub skip_unlock: bool,
+    /// Bytes forwarded as `hookData` to the V4 pool's hook on this
+    /// swap. Ignored for Angstrom hooks, where the attestation bytes
+    /// are fetched separately.
+    #[serde(default)]
+    pub hook_data: Bytes,
```
tamaralipows, line 257: Shouldn't all pub structs have private fields with setters?

tamaralipows (commented):
Thank you! Just one question about the pub fields. I'd approve but its my PR :)

crates/tycho-execution/src/encoding/evm/swap_encoder/uniswap_v4.rs
dianacarvalho1: meh, I'll make it better in the next PR

crates/tycho-execution/contracts/test/protocols/UniswapV4.t.sol
dianacarvalho1: yes, it's a pool manager error. don't remember now which one but I will double check

dianacarvalho1 (approved):

crates/tycho-execution/src/encoding/evm/swap_encoder/uniswap_v4.rs
dianacarvalho1: this is pub only in this module actually so this is fine!

PR #1013, opened 2026-05-13
test(integration): Record absolute gas error

dianacarvalho1:
This is to show a table with the median error per protocol in the dashboard


tamaralipows (approved):
"signed absolute" sounds funny lol but it makes sense to me. Thank you!

PR #1015, opened 2026-05-14
fix(router v3): Account for final transfer in slippage check

tamaralipows:
Problem:
- In some cases, we were checking the amount in the router to perform the final slippage check. This did not account for the final transfer to the user, so users could end up receiving less than the min amount.

Important Notes:
- Our client contribution may not work if the token is already in the user's wallet (for example for an optimized final swap with no fees). We send an additional amount to the user to make the contribution, but after fees are taken on this contribution transfer, the contribution might still not be enough.
- We might totally miss making a contribution altogether: we might think all is fine and no contribution is necessary, and the swap will revert at the final output transfer due to the contribution. This is hard to fix without a bigger refactor (dangerous at this point). 
- In general: fee taking with client contribution is a very difficult thing to achieve nicely (unless we want to either find ourselves in a loop of check balance -> contribute -> check balance -> contribute, or perform some token fee estimation math). I think it would be sufficient to just mention that these specific cases can cause reverts in our docs.

TL;DR

What this PR does: ensures the min amount check works (even with a client contribution)
What this PR doesn’t do (and shouldn’t do): ensure the user receives enough client contribution to prevent a revert

### DIFF
--- a/crates/tycho-execution/contracts/src/TychoRouter.sol
+++ b/crates/tycho-execution/contracts/src/TychoRouter.sol
@@ -731,6 +731,11 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
 
         amountOut =
             _settleOutput(amountOut, amountIn, tokenIn, tokenOut, receiver);
+
+        // Check final amount to account for fee tokens or rebasing tokens
+        if (amountOut < minAmountOut) {
+            revert TychoRouter__NegativeSlippage(amountOut, minAmountOut);
+        }
     }
 
     /**
@@ -803,6 +808,11 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
 
         amountOut =
             _settleOutput(amountOut, amountIn, tokenIn, tokenOut, receiver);
+
+        // Check final amount to account for fee tokens or rebasing tokens
+        if (amountOut < minAmountOut) {
+            revert TychoRouter__NegativeSlippage(amountOut, minAmountOut);
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


crates/tycho-execution/contracts/src/TychoRouter.sol
```diff
@@ -731,6 +731,11 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
 
         amountOut =
             _settleOutput(amountOut, amountIn, tokenIn, tokenOut, receiver);
+
+        // Check final amount to account for fee tokens or rebasing tokens
+        if (amountOut < minAmountOut) {
+            revert TychoRouter__NegativeSlippage(amountOut, minAmountOut);
+        }
```
louise-poole: FMI - this check is always done directly after you call `_settleOutput`... why not move the check into `_settleOutput` and reduce code duplication? 👀

crates/tycho-execution/contracts/test/TychoRouterSingleSwap.t.sol
```diff
@@ -621,3 +623,74 @@ contract TychoRouterSingleSwapTest is TychoRouterTestSetup {
...
     }
 }
+
+contract TychoRouterSingleSwapFeeTokenTest is TychoRouterTestSetup {
+    // TWIF/USDC V4 pool exists at this block
+    function getForkBlock() public view override returns (uint256) {
+        return 22689128;
+    }
+
+    function testFinalFeeBelowMinAmount() public {
```
louise-poole, line 633: Can we add a test for the more complex client contribution path too?

dianacarvalho1 (commented):
Thank you @tamaralipows ! 🙏🏼 Tbh I wouldn't worry much about FoT or rebasing tokens here.. For rebasing tokens the diff is usually a couple of WEI and I guess that people that want to trade FoT are used to their weird behaviours and at most we should add a warning when setting the minAmountOut (to be looser to account for the fees on the tokens) and that's it. I don't think it's worth it to make our router more complex (or add more balance checks) to account for these weird behaviours

crates/tycho-execution/contracts/src/TychoRouter.sol
tamaralipows: I thought about this and then saw that I'd need to change the fn signature to include min amount. Thinking about it again that's not a big deal actually xD

dianacarvalho1 (approved):
I agree I agree I agree

PR #1017, opened 2026-05-14
fix(integration-test): improve integration test stability and protocol state metrics

louise-poole:
- set to infinite reconnect attempts for tycho stream
- emit 'skipped' state for protocols whose data arrived too late for us to simulate on and we skipped it

### DIFF
--- a/crates/tycho-integration-test/src/main.rs
+++ b/crates/tycho-integration-test/src/main.rs
@@ -690,6 +690,9 @@ async fn process_update(
                     metrics::record_block_processing_duration(latency_seconds, block_type);
                 }
                 metrics::record_protocol_update_skipped();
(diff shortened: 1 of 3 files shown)


tamaralipows (approved):
Thank you for this. I'm curious what the retry amount was set to before? Were we just not retrying? :o

dianacarvalho1 (approved):

louise-poole:
> Thank you for this. I'm curious what the retry amount was set to before? Were we just not retrying? :o

It defaults to 127 retries. I saw in the logs that when we have a slow dev indexer redeployment we were exceeding those retries waiting for the indexer to be ready again. This mainly affected Base (it is very slow to start up due to the amount of data we cache etc).

PR #1019, opened 2026-05-15
feat(gas-estimations): Add router overhead gas

dianacarvalho1:
From user transfer and fee taking

### User transfer gas usage

From the current state of the dashboard it was quite difficult to infer these values 
image

I decided to do a first estimation using solidity tests. I used the already existing tests in `TychoRouterSingleSwapTest`: `testSingleSwapPermit2` and `testSingleSwapTransferFrom` (they do the same single univ2 swap WETH -> DAI) and created a corresponding `testSingleSwapVault` as well (thought it wasn't worth it to commit it). 
I assumed that a v2 swap takes around 70k gas (like we have in the PoolState) and that a WETH transfer is 30k (so 100k already accounted for in the current logic).
These are the gas usages of each test:
- singleSwap: 140168 (40k extra)
- singleSwapPermit2: 179883 (80k extra) 
- singleSwapUsingVault: 143317 (43k extra)

Conclusions: assigned 40k extra to TransferFrom and UseVaultsFunds and 80k to permit2 (maybe this is a bit too much though? not sure yet).

### Fee taking gas usage
Then I ran the same tests but with fees enabled. These are the results:
- singleSwap: 210063
- singleSwapPermit2: 249778 
- singleSwapUsingVault: 213190 

Overall there is a 70k increase when taking fees. This was a bit more than what I expected. I expected to have an extra transfer cost + some negligible gas for validating the signature and etc. The issue here is that the Vault's storage slots for the client and router that I used were cold, that's why there is such a high gas usage. I repeated the singleSwapUsingVault with warming the slots before doing that swap and then the value was much more reasonable 178,990 - only a ~40k increase. 
Conclusions: For the case when we take fees we add only an extra token transfer and no other gas overhead. 

### Split swaps
Also added a condition for split swaps: always account for an extra token out transfer to the router


crates/tycho-execution/src/encoding/evm/gas_estimator.rs
```diff
@@ -42,29 +44,42 @@ pub const PROTOCOLS_OUTPUT_TO_ROUTER: &[&str] = &["vm:curve", "rocketpool", "flu
...
+
+    // Add user transfer overhead
+    total_gas += BigUint::from(match *solution.user_transfer_type() {
+        UserTransferType::TransferFromPermit2 => 80_000u64,
+        _ => 40_000u64, // TransferFrom and UseVaultsFunds have similar overheads
+    });
+
+    // Add fees overhead: when fees are active and the last swap's protocol does not
+    // already route output through the router, the fee path adds an extra transfer.
+    if ROUTER_FEES_ACTIVE {
```
dianacarvalho1, line 70: This is not 100% true. We might not have router fees and have client fees only and this gas overhead is still needed. However:
1. We don't know about the client fees at this point (it's not in the Solution anymore)
2. We expect our Router to always have fees enabled anyway from here on out, so I thought it was fine to keep this simple.

tamaralipows: I agree this is fine!

crates/tycho-execution/src/encoding/evm/gas_estimator.rs
```diff
@@ -42,29 +44,42 @@ pub const PROTOCOLS_OUTPUT_TO_ROUTER: &[&str] = &["vm:curve", "rocketpool", "flu
...
             swap.token_in(),
             swap.token_out(),
+            &strategy,
         );
         total_gas += swap_transfer_overhead + swap.estimated_gas();
     }
+
+    // Add user transfer overhead
+    total_gas += BigUint::from(match *solution.user_transfer_type() {
+        UserTransferType::TransferFromPermit2 => 80_000u64,
```
tamaralipows, line 64: Actually this is exactly what I expected. I remember permit2 being more expensive. I vaguely remember it was around 40k - this is a good starting point

crates/tycho-execution/src/encoding/evm/gas_estimator.rs
```diff
@@ -96,22 +111,24 @@ fn estimate_transfer_overhead(
...
         overhead += BigUint::from(TOKEN_APPROVAL_GAS);
     }
 
-    // Output transfer: router -> receiver (only when outputToRouter).
-    if PROTOCOLS_OUTPUT_TO_ROUTER.contains(&protocol_system) {
-        overhead += transfer_gas(token_out);
+    // Output transfer: router -> receiver/next pool (only when outputToRouter or strategy is
+    // Split).
+    if PROTOCOLS_OUTPUT_TO_ROUTER.contains(&protocol_system) || *strategy == Strategy::Split {
+        overhead += transfer_token_gas(token_out);
```
tamaralipows, line 131: good catch on the split swap here

tamaralipows (approved):
Thank you! All makes sense

PR #1020, opened 2026-05-15
feat: Router V3 Deployment

tamaralipows:


crates/tycho-execution/contracts/scripts/deploy-router.js
```diff
@@ -26,6 +26,15 @@ async function main() {
...
+        permit2 = "0x000000000022D473030F116dDEE9F6B43aC78BA3";
+    } else if (network === "bsc") {
+        // permit2 address is the same as on ethereum
+        permit2 = "0x000000000022D473030F116dDEE9F6B43aC78BA3";
+    } else if (network === "polygon") {
+        // permit2 address is the same as on ethereum
+        permit2 = "0x000000000022D473030F116dDEE9F6B43aC78BA3";
     } else {
         throw new Error(`Unsupported network: ${network}`);
     }
```
louise-poole: Could we rather use a hashmap here? Would feel much cleaner and easier to update.

louise-poole (approved):

crates/tycho-execution/contracts/scripts/deploy-router.js
louise-poole: Actually from my understanding the permit2 address will be the same on every chain. Do we really need a per-chain check?

docs/for-dexs/protocol-integration/execution/README.md
```diff
@@ -194,6 +194,21 @@ If your protocol uses `address(0)` (e.g., UniswapV4, Ekubo) or another sentinel
...
 
+## Security Requirements
+
+TychoRouter calls executors via `delegatecall`, so executor code runs within TychoRouter's context — it can freely transfer the router's assets and write to the router's storage, including users' vault balances. Follow this checklist when building or reviewing an executor.
+
+{% hint style="danger" %}
+**Executors run with TychoRouter's full privileges.** A bug or malicious executor can steal all router assets.
+{% endhint %}
+
+* **Never call `ERC20.transfer`, `ERC20.transferFrom`, or `Permit2.transferFrom` directly.** Communicate transfer intent through `getTransferData` and `getCallbackTransferData` instead — TychoRouter performs the actual transfers with its own safeguards.
```
tamaralipows: Should we mention the native case here?
Also not to do any approvals

docs/for-dexs/protocol-integration/execution/README.md
```diff
@@ -194,6 +194,21 @@ If your protocol uses `address(0)` (e.g., UniswapV4, Ekubo) or another sentinel
...
+* **Never assign to a state variable** or perform any operation that writes to TychoRouter's storage.
+* **Do not execute `delegatecall`.** If a protocol makes it unavoidable, ensure the caller cannot control the `delegatecall` target — attacker-controlled targets enable arbitrary code execution within TychoRouter's context.
+* **Verify callback origin.** Call `verifyCallback` within `handleCallback` to confirm `msg.sender` is a valid pool from the expected protocol.
```
tamaralipows: We actually don't do this. I think it's rather: avoid using data sent via the callback. For example, the input token and the input amount are determined entirely in the dispatcher. If necessary for some reason to use the callback data, validate the pool?

docs/for-solvers/execution/encoding/README.md
```diff
@@ -214,6 +214,14 @@ EIP712Domain(string name, string version, uint256 chainId, address verifyingCont
 
 with `name = "TychoRouter"`, `version = "1"`, and `verifyingContract` set to the TychoRouter contract address.
 
+{% hint style="warning" %}
+**Replay attack risk.** The signature contains no nonce. Once a signed `ClientFeeParams` appears on-chain, anyone who sees it can reuse it for the same swap parameters until `deadline` expires. If `maxClientContribution > 0`, the swap's `receiver` can repeatedly replay the transaction — each replay debits the client's vault balance by up to `maxClientContribution` — until the balance is exhausted or the deadline passes.
```
tamaralipows: It can only replay if all the swaps and input params are the same - should we specify?

tamaralipows (commented):
Thank you!

docs/for-solvers/execution/encoding/README.md
dianacarvalho1: it kinda already says it "can reuse it for the same swap parameters". will add "and input" only

dianacarvalho1 (approved):
✨ _amazing_ ✨ Thank you @tamaralipows and @louise-poole 🙏🏼

louise-poole (approved):

PR #1021, opened 2026-05-18
feat(simulation): add partial blocks flag to Update

louise-poole:
- add partial block flag to `Update` struct to differentiate between a partial and full update. 
- [tycho-integration-test] add flag to run with execution disabled (useful for testing on new chains)
- [tycho-integration-test] use chain-specific TVL defaults to better mimic what our users will recieve

### DIFF
--- a/crates/tycho-integration-test/src/main.rs
+++ b/crates/tycho-integration-test/src/main.rs
@@ -35,7 +35,7 @@ use tycho_simulation::{
         hashflow::{client::HashflowClient, state::HashflowState},
         liquorice::{client::LiquoriceClient, state::LiquoriceState},
     },
-    tycho_common::models::Chain,
+    tycho_common::models::{Chain, TvlThresholdTier},
     utils::load_all_tokens,
 };
 use tycho_test::{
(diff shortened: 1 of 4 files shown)


dianacarvalho1 (approved):
Thank you @louise-poole ! Did you already test the new tvl threshold values? does this add a lot of new pools/make the test much slower/should we adjust the number of simulations per block?

louise-poole:
> Did you already test the new tvl threshold values? does this add a lot of new pools/make the test much slower/should we adjust the number of simulations per block?

Yes I did - we definitely need to adjust the max simulations per block on the faster chains. I'm making a PR for that shortly.

crates/tycho-integration-test/src/main.rs
```diff
@@ -161,6 +162,11 @@ struct Cli {
     /// Seconds without a protocol update before marking all known protocols as stale in metrics.
     #[arg(long, default_value_t = 30)]
     stale_threshold_secs: u64,
+
+    /// Disable on-chain swap execution via Tenderly (simulation only, no execution validation).
+    /// Useful for diagnosing stream latency without Tenderly congestion.
```
tamaralipows: We aren't using tenderly for simulation though? Or am I brainfarting? Like, we provide a tenderly link but I think we are just simulating against the RPC

tamaralipows (approved):
Thank you! just a point about the docstring

crates/tycho-integration-test/src/main.rs
louise-poole: Ah you right - I was mislead by the Tenderly debugging output that we generate. Will fix this

PR #1023, opened 2026-05-18
fix(integration-test): Handle router fee when computing slippage

dianacarvalho1:
The current router takes a fee on all outputs so the executed amount will be lower than expected and this influences the slippage calculations. Solution: remove the fee from the simulated/expected amount out before calculating slippage


louise-poole (approved):
Well spotted - I totally missed that we'd need this. Thanks!

tamaralipows (approved):
Verryyyy good - thank you!

PR #1025, opened 2026-05-18
fix: preserve UnknownExtractor through paginated fetch + ws retry cooldown

kayibal:
## Summary

Three fixes for Ethereum integration test failures.

**`rpc.rs`**: `get_protocol_components_paginated` wrapped the first-page fetch in `.map_err(|err| RPCError::Fatal(...))`, converting any `RPCError` — including the typed `UnknownExtractor` — into `Fatal` before it reached `BlockSynchronizer`. `BlockSynchronizer` only skips gracefully on `UnknownExtractor`; everything else propagates as a fatal `InitializationError`, crashing the stream build when an extractor is not yet deployed on the server. Since `get_protocol_components` already returns `RPCError`, the `map_err` was unnecessary — replaced with `?`.

**`feed/mod.rs`**: `BlockSynchronizerError::InitializationError` dropped the extractor identity, producing an opaque `"Failed to initialize synchronizer: ..."` message with no indication of what failed. Changed to a struct variant carrying the `ExtractorIdentity`, so the error now reads `"Failed to initialize extractor 'ethereum:vm:liquidityparty': ..."`.

**`protocol_stream_processor.rs`**: the websocket infinite-retry config used a 10 s cooldown, which exceeds the state-sync cooldown (3 s on Ethereum, 2 s on every other chain). This produced `"Websocket cooldown should be < than state synchronizer cooldown"` on every startup. Changed to 1 s, which is always below the minimum.

### DIFF
--- a/crates/tycho-client/src/feed/mod.rs
+++ b/crates/tycho-client/src/feed/mod.rs
@@ -102,8 +102,12 @@ impl HeaderLike for BlockHeader {
 
 #[derive(Error, Debug)]
 pub enum BlockSynchronizerError {
-    #[error("Failed to initialize synchronizer: {0}")]
-    InitializationError(#[from] SynchronizerError),
+    #[error("Failed to initialize extractor '{extractor}': {source}")]
+    InitializationError {
+        extractor: ExtractorIdentity,
+        #[source]
+        source: SynchronizerError,
+    },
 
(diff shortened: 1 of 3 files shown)


louise-poole (approved):
Thanks!!

PR #1026, opened 2026-05-19
feat: add extractor supervisor with fault tolerance and exponential backoff

zizou0x:
## What

Adds `ExtractorSupervisor` and `ExtractorFactory` to handle extractor restarts
with exponential backoff, replacing the one-shot `ExtractorBuilder` pattern.

## Why

Any extractor failure would cause the whole indexer process to panic. This
removes that coupling — each extractor is now isolated and restarts
independently on failure without affecting others or the RPC server.

## Changes

**New `ExtractorSupervisor`** (`extractor/supervisor.rs`)
- Runs a loop: build runner → run → on failure, clear WS subscriptions, signal
  `PendingDeltas` to reset its buffer, wait with exponential backoff, rebuild.
- Exposes `handle()` so callers get an `ExtractorHandle` without managing the
  control channel directly.

**New `ExtractorFactory`** (same file)
- Async constructor: one RPC call for `ChainState`, one `populate()` for
  `ProtocolMemoryCache`. Both are reused across restarts (cache via cheap Arc
  clone; chain state is `Copy`).
- `build_runner()` creates a fresh `ProtocolExtractor`, DCI plugin, and
  Substreams stream per restart, with a fresh `CachedGateway` instance for
  write isolation.

**`ExtractorRunner` simplified**
- No longer handles control messages. Receives a `oneshot` stop signal and
  shared `ws_subscriptions`/`pending_deltas_tx` directly.
- WS subscription management and stop handling moved to the supervisor.

**Config types moved**
- `ExtractorConfig`, `ProtocolTypeConfig`, `DCIType` moved from `runner.rs`
  to `supervisor.rs`. Internal fields are now private; only
  `initialized_accounts`, `initialized_accounts_block`, and `dci_plugin`
  remain `pub`.
- `max_restarts: Option<u32>` added to `ExtractorConfig` (YAML-configurable,
  default `None` = restart forever).

**`PendingDeltas` updated**
- `run()` now accepts pre-built receivers and a reset channel instead of
  subscribing through extractor handles dynamically. On restart the supervisor
  sends the extractor name on the reset channel to clear its buffer.

**Supporting changes**
- `CachedGateway::new_instance()` — fresh gateway with independent LRU cache
  and `open_tx`, sharing the same write channel and connection pool.
- `Chain::block_time()` — per-chain block time in seconds.
- `ExtractionError::variant_name()` — static label for Prometheus counters.


kayibal (commented):
🤖 The three-struct split (`ExtractorFactory` / `ExtractorSupervisor` / `ExtractorRunner`) is cleaner than the old builder, but it's worth asking whether `ExtractorRunner` pulls its weight now. Before this PR, Runner owned both the stream loop and control message handling. After the refactor, control messages moved to the supervisor, and Runner's only remaining job is driving the `select!` loop over the stream. All of its fields (`ws_subscriptions`, `pending_deltas_tx`, `stop_rx`) are owned by the supervisor and injected per restart — Runner holds no state of its own.

Not necessarily a blocker, but worth a conversation: is the stream loop complex enough to justify a dedicated struct, or would a private `run_stream()` method on the supervisor give the same isolation with one fewer abstraction layer?

crates/tycho-indexer/src/extractor/supervisor.rs
```diff
@@ -0,0 +1,803 @@
...
+///   exponential backoff, then rebuilds from scratch.
+/// - Forwards `ControlMessage::Subscribe` from the `ExtractorHandle` to the WS subscription map.
+/// - Forwards `ControlMessage::Stop` by signalling the runner's stop channel.
+pub struct ExtractorSupervisor {
+    factory: ExtractorFactory,
+    ctrl_tx: Sender<ControlMessage>,
+    control_rx: Receiver<ControlMessage>,
+    ws_subscriptions: Arc<Mutex<SubscriptionsMap>>,
+    pending_deltas_tx: Sender<ExtractorMsg>,
+    reset_tx: Sender<String>,
```
kayibal: 🤖 The `reset_tx: Sender<String>` side channel is fragile: it's a global mux keyed by extractor name strings, so a mismatch between the key used at `PendingDeltas::new()` construction time and `ExtractorIdentity::name` at runtime silently fails with just a `warn!` — the buffer isn't cleared and the restarted extractor replays into stale state.

A cleaner approach: replace `pending_deltas_tx: Sender<ExtractorMsg>` with a typed `Sender<DeltaCommand>`:

```rust
pub enum DeltaCommand {
    Block(ExtractorMsg),
    ExtractorRestarted,
}
```

The supervisor sends `DeltaCommand::ExtractorRestarted` over the per-extractor channel it already owns; `PendingDeltas` handles it by clearing its buffer. This eliminates `reset_tx`/`reset_rx` entirely, gives an ordering guarantee (the event arrives in sequence with the last block messages on the same channel), and keeps the "what to do on restart" decision inside `PendingDeltas` rather than leaking it into the supervisor.

As a side effect, `ExtractorRestarted` is also the right hook for notifying WS clients without a separate subscription-clearing call in the supervisor.

crates/tycho-indexer/src/extractor/supervisor.rs
```diff
@@ -0,0 +1,803 @@
...
+///
+/// Reused across restarts:
+/// - `protocol_cache`: populated once at construction; Arc-based so cloning is cheap and all runs
+///   share the same live cache. The TTL mechanism refreshes stale entries automatically.
+/// - `chain_state`: estimated once at construction (block number via RPC); `Copy` so each run gets
+///   its own copy at no cost.
+/// - `cached_gw`: each `build_runner` call creates a fresh instance via `new_instance()` (fresh
+///   `open_tx` and LRU cache; shared write channel to `DBCacheWriteExecutor`).
+/// - `token_pre_processor`, `rpc_client`: stateless RPC wrappers.
+pub struct ExtractorFactory {
```
kayibal: 🤖 `ExtractorFactory` and `ExtractorSupervisor` solve different problems — construction vs. lifecycle — but they live in the same 800-line file. The supervisor ends up importing `aws_sdk_s3`, `prost`, `protocol_extractor`, all the DCI types etc. purely because the factory is co-located.

Splitting `ExtractorFactory` (along with `ExtractorConfig`, `DCIType`, `ProtocolTypeConfig`, `ensure_spkg`, `download_file_from_s3`) into `extractor/factory.rs` would be a pure file move with no logic changes, but it makes `supervisor.rs` about one thing — restart lifecycle — and keeps the heavy construction imports where they belong.

crates/tycho-indexer/src/extractor/supervisor.rs
```diff
@@ -0,0 +1,803 @@
...
+    database_insert_batch_size: usize,
+    partial_blocks: bool,
+    runtime_handle: Option<Handle>,
+    protocol_cache: ProtocolMemoryCache,
+    chain_state: ChainState,
+}
+
+impl ExtractorFactory {
+    #[allow(clippy::too_many_arguments)]
+    pub async fn new(
```
kayibal: 🤖 `new()` conventionally signals a cheap, synchronous, infallible constructor. This one makes an RPC call (`get_block_number()`), populates a DB-backed cache (`protocol_cache.populate()`), and returns `Result`. Consider renaming to `init()` or `create()` to set accurate expectations at call sites.

crates/tycho-indexer/src/extractor/supervisor.rs
```diff
@@ -0,0 +1,803 @@
...
+        .await?;
+
+    let data = resp.body.collect().await.unwrap();
+
+    if let Some(parent) = download_path.parent() {
+        std::fs::create_dir_all(parent)
+            .with_context(|| format!("Failed to create directories for {parent:?}"))?;
+    }
+
+    std::fs::write(download_path, data.into_bytes()).unwrap();
```
kayibal: 🤖 Two `unwrap()` calls in a non-test function that already returns `anyhow::Result`:
```rust
let data = resp.body.collect().await.unwrap();
std::fs::write(download_path, data.into_bytes()).unwrap();
```
Both failure modes (body read error, file write error) are realistic in production and should propagate via `?` rather than panic.

crates/tycho-indexer/src/extractor/supervisor.rs
```diff
@@ -0,0 +1,240 @@
...
+                    extractor = %self.id,
+                    error = %err,
+                    "Failed to send ExtractorRestarted to PendingDeltas"
+                );
+            }
+
+            // Exponential backoff: 120s, 240s, 480s, 960s, 1920s, 3840s, 7680s, 14400s
+            // (capped at 4 hours).
+            let exp = restart_count.min(7); // 120 * 2^7 = 14400s = 4 hours, cap here to avoid overflow.
+            let backoff = std::time::Duration::from_secs(120 * 2u64.pow(exp));
```
kayibal: 🤖 **Backoff starts at 120s — first retry is 2 minutes**

The first restart delay is `120 * 2^0 = 120s` regardless of failure cause. A momentary network blip or DB connection drop will keep the extractor dark for 2 minutes before the first retry.

Typical supervisory backoffs start at 1–10s (with optional jitter) and cap at minutes/hours — not start there. Consider `5s, 10s, 20s, 40s, ..., capped at 3600s`. The 4-hour cap is reasonable; it's the floor that's too high.

crates/tycho-indexer/src/extractor/factory.rs
```diff
@@ -0,0 +1,584 @@
...
+sync_batch_size: 1000
+start_block: 21688329
+protocol_types:
+  - name: uniswap_v4_pool
+    financial_type: Swap
+spkg: substreams/ethereum-uniswap-v4/ethereum-uniswap-v4-v0.2.1.spkg
+module_name: map_protocol_changes
+dci_plugin:
+  type: uniswap_v4_hooks
+  router_address: "0x2e234DAe75C793f67A35089C9d99245E1C58470b"
```
kayibal, line 579: 🤖 **`router_address` in YAML is silently ignored**

`DCIType::UniswapV4Hooks` only deserialises `pool_manager_address`:
```rust
UniswapV4Hooks { pool_manager_address: String },
```
The `router_address` field here is silently dropped by serde (no `deny_unknown_fields`). The factory hardcodes that address:
```rust
// random address to deploy our mini router to
let router_address = Address::from("0x2e234DAe75C793f67A35089C9d99245E1C58470b");
```
An operator who copies this YAML and changes `router_address` will get no error and no effect. Either add the field to the variant and use it, or remove the key from this YAML to avoid misleading config documentation.

crates/tycho-indexer/src/extractor/factory.rs
```diff
@@ -0,0 +1,584 @@
...
+        rpc_client: EthereumRpcClient,
+        database_insert_batch_size: usize,
+        partial_blocks: bool,
+        runtime_handle: Option<Handle>,
+    ) -> Result<Self, ExtractionError> {
+        let block_number = rpc_client
+            .get_block_number()
+            .await
+            .map_err(|e| ExtractionError::Setup(format!("Failed to get block number: {e}")))?;
+        let chain_state = ChainState::new(
```
kayibal, line 201: 🤖 **Stale `chain_state` timestamp after restarts** 
`current_block_ts` is captured once at factory construction (`chrono::Local::now()`) and reused across all subsequent restarts. After a long-running extractor fails and restarts hours later, the timestamp in `chain_state` will be stale by the full uptime of the extractor.

If `chain_state` is used anywhere to compute lag metrics, estimate current time, or derive block timestamps, this will silently produce wrong values post-restart. Worth a note confirming this is safe for all consumers.

crates/tycho-indexer/src/extractor/runner.rs
```diff
@@ -419,8 +392,22 @@ impl ExtractorRunner {
...
+        if let Some(tx) = pending_deltas_tx {
+            if let Err(err) = tx
+                .send(DeltaCommand::Block(message.clone()))
+                .await
+            {
+                error!(error = %err, "Failed to send to PendingDeltas channel");
+            }
+        }
+
         // TODO: rename variable here instead
```
kayibal: 🤖 **Stale TODO + wrong variable name — clean up now**

`arced_message` is not an `Arc` — it's just the message moved into a new binding. This rename has been deferred for a while and the surrounding function was significantly refactored in this PR, making it a natural moment to finish the job.

crates/tycho-indexer/src/main.rs
```diff
@@ -511,24 +544,40 @@ async fn build_all_extractors(
...
+            token_pre_processor.clone(),
+            rpc_client.clone(),
+            database_insert_batch_size,
+            partial_blocks,
+            Some(runtime_handle),
+        )
+        .await?;
+
+        // Create dedicated PendingDeltas channel for this extractor
+        let (pd_tx, pd_rx) = tokio::sync::mpsc::channel(256);
```
kayibal: 🤖 **Magic number for PendingDeltas channel buffer**

`256` is a tunable that affects backpressure and memory usage under load. Embedding it as a literal makes it easy to overlook and inconsistent with the `128` used for the supervisor control channel. A named constant makes the intent visible and the value easy to adjust.

crates/tycho-indexer/src/extractor/factory.rs
kayibal: This is only used to compute sync log messages.

crates/tycho-indexer/src/extractor/factory.rs
kayibal: Low prio, but indeed if you specify:

```
dci_plugin:
  type: rpc
  router_address: "0x2e234DAe75C793f67A35089C9d99245E1C58470b"
```

It won't complain... kind of out of scope.

crates/tycho-indexer/src/extractor/runner.rs
kayibal: It's actually an Arc here but indeed we should probably just do the renaming now.

crates/tycho-indexer/src/extractor/supervisor.rs
kayibal: I agree we should immediately restart on the first error. So I'd change this to be 1s or so initially. 

We already depend on tokio-retry in this crate, we could reuse it instead of implementing our own exponential backoff:

https://docs.rs/tokio-retry/latest/tokio_retry/

crates/tycho-indexer/src/extractor/factory.rs
```diff
@@ -0,0 +1,584 @@
...
+
+    /// Builds a fresh, ready-to-run [`ExtractorRunner`].
+    ///
+    /// Creates a fresh gateway instance, constructs the DCI plugin if configured, and establishes
+    /// the Substreams stream from the last committed block (or the config start block on first
+    /// run). The protocol cache and chain state are reused from factory construction.
+    pub async fn build_runner(
+        &self,
+        ws_subscriptions: Arc<Mutex<SubscriptionsMap>>,
+        pending_deltas_tx: Option<Sender<DeltaCommand>>,
```
kayibal: Wait so PendingDeltas, used to operate with the standard subscription mechnism and now we use something separate?

crates/tycho-indexer/src/extractor/runner.rs
```diff
@@ -164,23 +144,26 @@ impl MessageSender for ExtractorHandle {
...
-    control_rx: Receiver<ControlMessage>,
+    /// WS subscribers — managed by the supervisor, shared across restarts.
+    ws_subscriptions: Arc<Mutex<SubscriptionsMap>>,
+    /// Dedicated channel for PendingDeltasBuffer — survives restarts.
+    pending_deltas_tx: Option<Sender<DeltaCommand>>,
+    /// Oneshot stop signal from the supervisor.
+    stop_rx: oneshot::Receiver<()>,
     /// Handle of the tokio runtime on which the extraction tasks will be run.
-    /// If 'None' the default runtime will be used.
+    /// If `None` the default runtime will be used.
```
kayibal: Would be good to mention here in the comment that this refers to the tokio runtime.

crates/tycho-indexer/src/extractor/runner.rs
```diff
@@ -164,23 +144,26 @@ impl MessageSender for ExtractorHandle {
...
     extractor: Arc<dyn Extractor>,
     substreams: SubstreamsStream,
-    subscriptions: Arc<Mutex<SubscriptionsMap>>,
-    next_subscriber_id: u64,
-    control_rx: Receiver<ControlMessage>,
+    /// WS subscribers — managed by the supervisor, shared across restarts.
+    ws_subscriptions: Arc<Mutex<SubscriptionsMap>>,
+    /// Dedicated channel for PendingDeltasBuffer — survives restarts.
+    pending_deltas_tx: Option<Sender<DeltaCommand>>,
+    /// Oneshot stop signal from the supervisor.
```
kayibal: I don't like that these are split now. I also dislike that to correctly model the old behaviour the `pending_deltas_tx` has to be optional. I think ideally DeltaCommand should be moved into SubscriptionMap. Then any other component that requires it independently if internal or external can subscribe to it via subscriptions.

crates/tycho-indexer/src/extractor/supervisor.rs
```diff
@@ -0,0 +1,240 @@
...
+                )
+                .increment(1);
+                return runner_result
+                    .map_err(|e| ExtractionError::Unknown(format!("Runner panicked: {e}")))?;
+            }
+
+            // Clear WS subscriptions — clients must reconnect after a restart.
+            // TODO: can we keep the ws connections alive and handle this on the client side?
+            {
+                let mut subs = self.ws_subscriptions.lock().await;
```
kayibal: I really dislike that the Supervisor and the ExtractorRunner now "know" about websocket subscriptions and treat them differently than PendingDelta subscriptions, previously the only concept was to have subscribers. In my opinion this can be kept now that we introduced DeltaCommand type.

Atm you are forcefully clearing the websocket subscriptions, making the normal subscription service potentially unusable for any other internal future use case. I think the connection restart should be decided at the websocket layer not here. So services::ws should instead receive a DeltaCommand::ExtractoRestarted and act upon it as it thinks is best.

kayibal (commented):
I think removing websocket knowledge out of the supervisor should still happen before we merge this.

crates/tycho-indexer/src/extractor/supervisor.rs
zizou0x: Done in 74dbc8ceb: the supervisor no longer touches WS subscriptions. It broadcasts `DeltaCommand::ExtractorRestarted` to every subscriber on the same channel as block messages, and `services::ws` decides how to react — it ends the affected client subscriptions with `SubscriptionEnded` while keeping the connection (and other subscriptions) alive. Added `test_subscription_ended_on_extractor_restart` covering this.

crates/tycho-indexer/src/extractor/runner.rs
zizou0x: Done in 74dbc8ceb: `SubscriptionsMap` now carries `Sender<DeltaCommand>`, so PendingDeltas registers as a regular subscriber (`ExtractorSupervisor::add_subscriber`) and the `Option<Sender<DeltaCommand>>` on the runner is gone. Runtime subscribers keep going through `ExtractorHandle::subscribe`, which now also delivers `DeltaCommand`.

crates/tycho-indexer/src/extractor/factory.rs
zizou0x: No longer — as of 74dbc8ceb PendingDeltas is back on the standard subscription mechanism: it's just subscriber #0 in the shared subscription map, and the dedicated channel parameter has been removed from `build_runner`/`ExtractorRunner`.

crates/tycho-indexer/src/extractor/supervisor.rs
zizou0x: Fixed in 048715313: switched to tokio-retry's `ExponentialBackoff` starting at 1s, doubling per restart, capped at 4 hours (`MAX_RESTART_BACKOFF`).

crates/tycho-indexer/src/extractor/runner.rs
zizou0x: Renamed in 74dbc8ceb — the extra binding and the TODO are gone; `propagate_msg` now builds the `DeltaCommand` once and clones it per subscriber.

crates/tycho-indexer/src/main.rs
zizou0x: Fixed in 5ac23ad45: `PENDING_DELTAS_CHANNEL_SIZE` (256) in main.rs and `CONTROL_CHANNEL_SIZE` (128) in supervisor.rs.

crates/tycho-indexer/src/extractor/runner.rs
zizou0x: Updated in 74dbc8ceb: the doc now reads "If `None` the default tokio runtime will be used."

crates/tycho-indexer/src/extractor/supervisor.rs
```diff
@@ -71,6 +75,10 @@ impl ExtractorSupervisor {
     /// `ControlMessage::Stop` or has exhausted all restart attempts.
     pub async fn run(mut self) -> Result<(), ExtractionError> {
         let mut restart_count: u32 = 0;
+        // Exponential backoff: 1s, 2s, 4s, ... capped at 4 hours.
```
zizou0x: 1s is too small for a first backoff for an extractor. We should start with at least 60s.

crates/tycho-indexer/src/extractor/mod.rs
```diff
@@ -109,12 +109,13 @@ pub enum RPCError {
...
-/// per-extractor channel.
+/// Messages delivered to extractor subscribers over their subscription channel.
 ///
-/// Using a single typed channel (rather than a separate reset side-channel) gives an ordering
-/// guarantee: `ExtractorRestarted` always arrives after every `Block` message the runner sent
-/// before it stopped.
+/// Every subscriber (`PendingDeltas`, WebSocket clients, ...) receives the same commands and
+/// decides for itself how to react to a restart. Sending `ExtractorRestarted` on the same
+/// channel as `Block` messages gives an ordering guarantee: it always arrives after every
+/// `Block` message the runner sent before it stopped.
```
zizou0x, line 128: Isn't this a breaking change for clients running with an old tycho-client version?

crates/tycho-indexer/src/extractor/supervisor.rs
zizou0x: Changed in 69e1ce249: 60s, 120s, 240s, ... capped at 4 hours.

crates/tycho-indexer/src/extractor/mod.rs
zizou0x: Checked — no, it isn't. `DeltaCommand` is server-internal and never crosses the wire; this PR makes no changes to `tycho-common/src/dto.rs` (the only `tycho-common` change is the additive `Chain::block_time()`).

The only client-visible effect is an unsolicited `Response::SubscriptionEnded` when an extractor restarts. That variant has been in the dto and handled by the client since Jan 2024 (`2ba1dabf5`): `remove_subscription` logs "Subscription ended unexpectedly!", drops the subscription sender, and the synchronizer sees "Deltas channel closed" (`b95902181`, Feb 2024). That is the exact same internal path old clients already take today when the server restarts: on reconnect `WsDeltasClient` creates a fresh `Inner` with empty subscription maps, so the old senders are dropped and the synchronizer hits the same channel-closed error. Pre-PR an extractor failure panicked the whole indexer, so old clients went through that path on every failure anyway — now it is scoped to the one affected extractor and clients since `5ae277b11` (Apr 2025) auto-retry and resubscribe.

One behavioural note: a client resubscribing while the extractor is in backoff gets its `NewSubscription` immediately (the subscribe request is buffered on the supervisor's control channel) and first deltas only after the extractor recovers, so the synchronizer's first-message timeout can fire during long backoffs — same retry-budget semantics as a server crash-loop today.

crates/tycho-indexer/src/extractor/supervisor.rs
```diff
@@ -0,0 +1,250 @@
...
+                Ok(r) => r,
+                Err(err) => {
+                    error!(extractor = %self.id, error = %err, "Failed to build extractor");
+                    metrics::counter!(
+                        "extractor_restart_failed",
+                        "extractor" => self.id.name.clone()
+                    )
+                    .increment(1);
+                    return Err(err);
+                }
```
brunoguerios, line 124: A `build_runner` error bypasses the whole retry mechanism: the supervisor returns `Err`, and upstream `join_all` (main.rs:300) only logs it. The extractor is dead for the process lifetime while the process keeps running.

Rebuild hits the DB, RPC, and possibly S3 — the same transient failures that kill a run. So the likeliest sequence (extractor fails during a DB blip → backoff → rebuild hits the same blip) permanently kills the extractor. This is the exact scenario the supervisor exists to survive.

Suggestion: route build errors through the same backoff/`max_restarts` path as run errors. Consider treating config errors (`ExtractionError::Setup`) as fail-fast instead — on main, a bad config aborts startup; here it only logs.

crates/tycho-indexer/src/extractor/supervisor.rs
```diff
@@ -0,0 +1,250 @@
...
+                    "Extractor permanently stopped — restart limit reached"
+                );
+                metrics::counter!(
+                    "extractor_permanently_stopped",
+                    "extractor" => self.id.name.clone()
+                )
+                .increment(1);
+                return runner_result
+                    .map_err(|e| ExtractionError::Unknown(format!("Runner panicked: {e}")))?;
+            }
```
brunoguerios, line 230: Both permanent-stop paths (max restarts here, build failure above) return before the `ExtractorRestarted` broadcast. WS clients are fine — channel close ends the subscription — but `PendingDeltas` only sees its receiver stream end and never clears the buffer (deltas_buffer.rs:346). Every later RPC request for this protocol merges the frozen deltas on top of DB state, indefinitely.

Suggestion: send the broadcast (or a dedicated `Stopped` command) on both exit paths so the buffer is dropped.

crates/tycho-indexer/src/extractor/supervisor.rs
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    /// Runs the supervision loop. Returns when the extractor has been stopped via
+    /// `ControlMessage::Stop` or has exhausted all restart attempts.
+    pub async fn run(mut self) -> Result<(), ExtractionError> {
+        let mut restart_count: u32 = 0;
+        // Exponential backoff: 60s, 120s, 240s, ... capped at 4 hours.
+        let mut backoff_strategy = ExponentialBackoff::from_millis(2)
+            .factor(30_000)
+            .max_delay(MAX_RESTART_BACKOFF);
```
brunoguerios: `restart_count` and `backoff_strategy` live outside the loop and never reset, so they measure failures per process lifetime, not consecutive failures. An extractor that fails once a day reaches the 4h cap after ~9 days and stays there. With `max_restarts: Some(n)`, the n-th sporadic failure — weeks apart — permanently stops a healthy extractor.

Suggestion: reset both after a healthy run (e.g. runner ran longer than the current backoff, or processed N blocks).

crates/tycho-indexer/src/extractor/supervisor.rs
```diff
@@ -0,0 +1,250 @@
...
+            let backoff = backoff_strategy
+                .next()
+                .expect("backoff strategy is infinite");
+            warn!(
+                extractor = %self.id,
+                ?backoff,
+                restart_count,
+                "Waiting for backoff before restarting extractor"
+            );
+            tokio::time::sleep(backoff).await;
```
brunoguerios: `control_rx` is only polled while the runner runs. A `Stop` that arrives during this sleep (up to 4h) or during `build_runner` waits until the full rebuild completes — DCI init, stream setup — before it takes effect. Process shutdown currently works only because the services-side task wins the control-channel race in main.rs:333 and the runtime drop cancels this task; the supervisor should not depend on that.

Suggestion: `select!` over `control_rx` and the sleep (and ideally the build) so `Stop` is honored immediately.

brunoguerios (commented):
Changes seem to make sense 👍 
I added a few comments that Claude suggested and I thought were worth sharing.
They feel worth looking into, before approving/merging.
Let me know your thoughts 🙏

brunoguerios (approved):
LGTM ✅

PR #1029, opened 2026-05-19
feat: add PendingBlockProcessor and TxDeltaIndexer infrastructure

kayibal:
## Summary

- Adds `TxDeltaIndexer` trait and `TxInput`/`LogInput` structs to `tycho-common` as a native-code substitute for Substreams packages
- Adds `PendingBlockProcessor` to `tycho-simulation` — accepts a slice of `TxInput`, runs registered indexers, applies deltas ephemerally (without touching confirmed state), and returns a labelled `PendingUpdate`
- Adds `apply_deltas_ephemeral` to `TychoStreamDecoder`: read-only counterpart of `decode()` that clones pool states, applies pending deltas, injects block context, and returns an `Update`
- Adds `build_with_pending` to `ProtocolStreamBuilder` to wire a `PendingBlockProcessor` alongside the live feed via an unbounded channel

## Design notes

- `PendingBlockProcessor::generate_pending_update` drains confirmed blocks from the feed channel with `try_recv` before running, so its view of pool state is always up to date
- Returns `PendingError::ParentNotYetConfirmed` immediately if the target block's parent hasn't been confirmed yet
- `PendingUpdate` wraps `Update` with a caller-defined `label` string — avoids modifying the shared `Update` type

## Test plan

- [ ] CI passes (unit tests, clippy, format)
- [ ] Integration tests with UniswapV3 processor (covered in follow-up PR)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,12 +45,7 @@ serde = { version = "1.0.204", features = ["rc", "derive"] }
 serde_json = "1.0.120"
 toml = "0.8"
 typetag = "0.2.21"
-uuid = { version = "1.4.1", features = [
-    "serde",
-    "v4",
-    "fast-rng",
-    "macro-diagnostics",
-] }
+uuid = { version = "1.4.1", features = ["serde", "macro-diagnostics"] }
 
 # --- Async & concurrency ---
 tokio = { version = "1.47.1", features = ["full"] }
@@ -77,7 +72,7 @@ alloy = { version = "1.1.0", default-features = false }
 ethabi = "18.0.0"
 
 # --- Utility ---
-chrono = { version = "0.4.26", features = ["serde"] }
+chrono = { version = "0.4.26", default-features = false, features = ["serde", "clock", "std"] }
 hex = "0.4.3"
 hex-literal = "0.4.1"
 rand = "0.8.5"
--- a/crates/tycho-common/src/traits.rs
+++ b/crates/tycho-common/src/traits.rs
@@ -50,7 +50,7 @@ pub trait TxDeltaIndexer: Send {
     /// * `block` — a finalised [`BlockAggregatedChanges`] as received from the Tycho client. On the
     ///   first call this carries the full component and state snapshot; on later calls it carries
     ///   only the changed attributes and balances.
-    fn apply_block(&mut self, block: &BlockAggregatedChanges);
+    fn apply_block(&mut self, block: &BlockAggregatedChanges) -> anyhow::Result<()>;
 
     /// Applies a batch of in-flight transactions against the current state and
     /// returns the protocol state deltas they would produce.
--- a/crates/tycho-common/src/traits.rs
+++ b/crates/tycho-common/src/traits.rs
@@ -5,13 +5,74 @@ use async_trait::async_trait;
 
 use crate::{
     models::{
-        blockchain::{Block, BlockTag, EntryPointWithTracingParams, TracedEntryPoint},
+        blockchain::{
+            Block, BlockAggregatedChanges, BlockTag, EntryPointWithTracingParams, TracedEntryPoint,
+            TxInput,
+        },
         contract::AccountDelta,
         token::{Token, TokenQuality, TransferCost, TransferTax},
         Address, Balance, BlockHash, StoreKey,
     },
     Bytes,
 };
+
+/// Indexes protocol state deltas from raw EVM transactions.
+///
+/// A `TxDeltaIndexer` is a native-code substitute for a Substreams package.
+/// It consumes a stream of finalised [`BlockAggregatedChanges`] to keep its internal
+/// protocol state current, then — on demand — applies a batch of in-flight
+/// [`TxInput`]s and returns the resulting [`BlockAggregatedChanges`]. The primary
+/// consumer is an Ethereum block builder that needs to know how a candidate
+/// transaction bundle alters DEX state before deciding whether to include it.
+///
+/// # Lifecycle
+///
+/// 1. **Hydrate** — call [`apply_block`][TxDeltaIndexer::apply_block] for each finalised block
+///    received from the Tycho client. The first call serves as initialisation: `state_deltas` and
+///    `component_balances` will contain the full snapshot at that height rather than a sparse
+///    delta. Subsequent calls apply incremental deltas.
+///
+/// 2. **Query** — call [`generate_deltas`][TxDeltaIndexer::generate_deltas] with a batch of
+///    in-flight transactions at any point. The indexer applies them against its current internal
+///    state and returns a [`BlockAggregatedChanges`] describing what would change. Internal state
+///    is **not** mutated by this call; it always operates on the state left by the most recent
+///    `apply_block`.
+pub trait TxDeltaIndexer: Send {
+    /// Advances internal protocol state by applying a finalised block.
+    ///
+    /// Must be called in block-height order. The first call initialises the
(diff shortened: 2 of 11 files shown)


crates/tycho-simulation/src/evm/stream.rs
```diff
@@ -443,6 +449,94 @@ impl ProtocolStreamBuilder {
...
+    pub async fn build_with_pending(
+        self,
+    ) -> Result<
+        (impl Stream<Item = Result<Update, StreamDecodeError>>, PendingBlockProcessor),
+        StreamError,
+    > {
+        initialize_hook_handlers().map_err(|e| {
+            StreamError::SetUpError(format!("Error initializing hook handlers: {e:?}"))
+        })?;
+        let (_, rx) = self.stream_builder.build().await?;
```
zizou0x, line 495: Should we forward the handle to the caller instead of dropping it or how does the `Stream` handle this?

Cargo.toml
```diff
@@ -45,12 +45,7 @@ serde = { version = "1.0.204", features = ["rc", "derive"] }
 serde_json = "1.0.120"
 toml = "0.8"
 typetag = "0.2.21"
-uuid = { version = "1.4.1", features = [
-    "serde",
-    "v4",
-    "fast-rng",
-    "macro-diagnostics",
-] }
+uuid = { version = "1.4.1", features = ["serde", "macro-diagnostics"] }
```
zizou0x, line 48:
```suggestion
uuid = { version = "1.4.1", features = ["serde", "macro-diagnostics"] }
```

crates/tycho-simulation/src/evm/pending.rs
```diff
@@ -0,0 +1,279 @@
...
+            .map(|s| s.header.number)
+            .max()
+            .unwrap_or(0);
+
+        for (extractor, state_msg) in &msg.state_msgs {
+            let Some(indexer) = self.indexers.get_mut(extractor) else {
+                continue;
+            };
+
+            if !state_msg.snapshots.states.is_empty() {
```
zizou0x, line 193: To properly warm the in-memory stores of the substreams we need a full snapshot here. Is it currently the case? I have the feeling we're using the normal stream so applying TVL filters

crates/tycho-simulation/src/evm/decoder.rs
```diff
@@ -923,6 +927,72 @@ where
...
 
+    /// Applies pending deltas from one or more `TxDeltaIndexer`s against the current confirmed
+    /// state and returns an ephemeral `Update`.
+    ///
+    /// This is the read-only counterpart of `decode()`. It clones pool states, applies the
+    /// supplied `pending_deltas`, and returns the result — **without writing back** to
+    /// `DecoderState`. Calling this method twice with the same input produces identical results.
+    ///
+    /// Only native protocols are supported: VM protocols that require `update_engine()` are
+    /// silently skipped (their pools will not appear in the returned `Update`).
```
zizou0x: Where is this happening? I didn't see anything related to this? 

From what I could see there is a chance we update the protocol state for a vm component without updating VM which would return a wrong state. I think we need the explicitly ignore all VM protocols.

crates/tycho-common/src/traits.rs
```diff
@@ -5,14 +5,75 @@ use async_trait::async_trait;
...
+/// 2. **Query** — call [`generate_deltas`][TxDeltaIndexer::generate_deltas] with a batch of
+///    in-flight transactions at any point. The indexer applies them against its current internal
+///    state and returns a [`BlockAggregatedChanges`] describing what would change. Internal state
+///    is **not** mutated by this call; it always operates on the state left by the most recent
+///    `apply_block`.
+pub trait TxDeltaIndexer: Send {
+    /// Advances internal protocol state by applying a finalised block.
+    ///
+    /// Must be called in block-height order. The first call initialises the
+    /// indexer from a full snapshot; subsequent calls apply incremental state
```
zizou0x, line 44: One thing to note here: if the substreams module uses a store on something that end up not being pushed to Tycho then we need a resync. 

From the top of my head I didn't find any case but for example if we have a store of `component -> type` but it's only used internally in Substreams and not emitted as attribute then we would be stuck (because we wouldn't be able to initialize the `TxDeltaIndexer` from Tycho db). In that case we would need to resync.

crates/tycho-common/src/traits.rs
```diff
@@ -5,14 +5,75 @@ use async_trait::async_trait;
...
+    /// deltas. After this call returns, [`generate_deltas`][TxDeltaIndexer::generate_deltas]
+    /// will produce deltas relative to the new state.
+    ///
+    /// # Parameters
+    ///
+    /// * `block` — a finalised [`BlockAggregatedChanges`] as received from the Tycho client. On the
+    ///   first call this carries the full component and state snapshot; on later calls it carries
+    ///   only the changed attributes and balances.
+    fn apply_block(&mut self, block: &BlockAggregatedChanges);
```
zizou0x: IMO we will need a result returned here, else we wouldn't know if something was off and if the states are reliable still.

crates/tycho-common/src/traits.rs
kayibal: Yeah in that case the information needs to come from somewhere else or indeed we'd need to change the substreams.

crates/tycho-simulation/src/evm/decoder.rs
kayibal: Yes correct, we should probably have a if somewhere that dissallows adding vm protocols.

crates/tycho-simulation/src/evm/pending.rs
kayibal: Why is this a problem? I am assuming here that we only need state of pools we are interested in, so the tvl filter shouldn't be an issue.

crates/tycho-simulation/src/evm/stream.rs
kayibal: Seems fine, It's done the same way in the `build` method

Cargo.toml
kayibal: Probably too tired but it's unclear what you are suggesting to change here.

kayibal: 🤖 The `v4` and `fast-rng` features were intentionally dropped from the workspace-level `uuid` entry. `tycho-common` needs to compile to WASM (for use in browser/edge environments), and both `v4` and `fast-rng` pull in `getrandom` with OS-specific entropy sources that are unavailable under WASM targets. The features are still present where UUID generation actually happens: `tycho-indexer` and `tycho-simulation` (prod deps) and `tycho-common` (dev-deps only, for tests). Crates that only use `Uuid` as a type (`tycho-client`) inherit the workspace entry as-is and compile fine.

zizou0x: Oh, there was a weird indent at this line but it disapeared, all good

crates/tycho-simulation/src/evm/pending.rs
zizou0x: I think there could be some complex cases where the store needs to be complete for the whole protocol, for example in balancer we have a store that keeps track of all tokens used by all components, on fluid_v2 we also have some cross-components stores (IIRC it's `token -> set(components)`).

I'm fine with starting with this for the PoC. We just need to keep in mind that these stores aren't 1 <-> 1 with their equivalence in Substreams and we can't 100% assume that all our Substreams module will be convertible.

zizou0x (approved):

PR #1031, opened 2026-05-20
feat(tycho-client): move runtime snapshot fetches off the delta hot path

louise-poole:
New components discovered during delta processing are now fetched via background tokio tasks instead of blocking the delta loop. The synchronizer accumulates deltas while tasks are in-flight and applies them at drain time so each snapshot arrives fully up-to-date.

Introduces a SnapshotStatus state machine (Deferred → InFlight → RetryNext / Blacklisted) replacing four separate HashSets. Transient fetch failures are retried on the next delta; permanent failures blacklist the component until restart. In partial-blocks mode, new components remain Deferred until the block boundary, then fire at the previous block's height.

In local testing on ethereum:
Measured delta processing time: 1–4 ms vs 400+ ms previously when a new native component snapshot had to be fetched synchronously.

### DIFF
--- a/crates/tycho-client/src/feed/synchronizer.rs
+++ b/crates/tycho-client/src/feed/synchronizer.rs
@@ -702,7 +702,7 @@ where
                             }
 
                             // Drain any background snapshot tasks that have completed.
-                            let background_snapshots = self.drain_pending_snapshots();
+                            let background_snapshots = self.drain_completed_snapshots();
 
                             // Trim buffered_deltas: discard blocks no longer needed by any pending task.
                             if self.pending_snapshots.is_empty() {
@@ -981,7 +981,7 @@ where
 
     /// Drains any background snapshot tasks that have completed. Returns a `Snapshot` containing
     /// all ready results, with buffered deltas applied to bring each snapshot up to date.
-    fn drain_pending_snapshots(&mut self) -> Snapshot {
+    fn drain_completed_snapshots(&mut self) -> Snapshot {
         let mut result = Snapshot::default();
         let pending = std::mem::take(&mut self.pending_snapshots);
 
@@ -1004,24 +1004,10 @@ where
                     self.component_tracker
                         .components
                         .extend(fetch_result.components);
-                    // Mirror start_tracking → update_contracts: walk pre-existing
-                    // self.entrypoints for any entrypoints already linked to these components
-                    // and add their contracts. Without this, a component that shares an
-                    // entrypoint with an already-tracked component would miss those contracts
-                    // until the next DCI update.
-                    self.component_tracker.contracts.extend(
-                        self.component_tracker
-                            .get_contracts_by_component(&new_component_ids),
-                    );
-                    // Also add DCI contracts from this fetch that aren't in self.entrypoints yet.
-                    self.component_tracker.contracts.extend(
-                        fetch_result
-                            .contract_ids
-                            .iter()
-                            .cloned(),
-                    );
                     self.component_tracker
                         .process_entrypoints(&fetch_result.dci_update);
+                    self.component_tracker
+                        .update_contracts(new_component_ids);
                     let mut snapshot = fetch_result.snapshot;
                     self.apply_deltas_to_snapshot(
                         &mut snapshot,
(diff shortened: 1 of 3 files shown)


crates/tycho-client/src/feed/synchronizer.rs
```diff
@@ -592,63 +696,130 @@ where
...
+                            if !self.pending_snapshots.is_empty() {
+                                self.buffered_deltas.push(deltas.clone());
+                            }
 
-                            let (snapshots, removed_components) = {
-                                // 1. Remove components based on latest changes
-                                // 2. Add components based on latest changes, query those for snapshots
-                                let (to_add, to_remove) = self.component_tracker.filter_updated_components(&deltas);
+                            // Drain any background snapshot tasks that have completed.
+                            let background_snapshots = self.drain_pending_snapshots();
```
tamaralipows: So `pending_snapshots` in this case is actually completed? 😅

crates/tycho-client/src/feed/synchronizer.rs
```diff
@@ -713,6 +883,198 @@ where
...
+    fn apply_deltas_to_snapshot(
+        &self,
+        snapshot: &mut Snapshot,
+        snapshot_block: u64,
+        contract_ids: &HashSet<Bytes>,
+    ) {
+        for delta in &self.buffered_deltas {
+            if delta.block.number <= snapshot_block {
+                continue;
+            }
```
tamaralipows, line 897: Why are we skipping the delta if it's older than this block? Don't we still need it? Or I guess, the latest delta will contain all of the necessary state and account deltas? Haha this is a bit confusing

crates/tycho-client/src/feed/synchronizer.rs
```diff
@@ -713,6 +883,198 @@ where
...
+            self.snapshot_queue
+                .insert(id.clone(), SnapshotStatus::InFlight);
+        }
+        self.pending_snapshots
+            .push(PendingSnapshot { component_ids, snapshot_block, receiver: rx });
+    }
+
+    /// Drains any background snapshot tasks that have completed. Returns a `Snapshot` containing
+    /// all ready results, with buffered deltas applied to bring each snapshot up to date.
+    fn drain_pending_snapshots(&mut self) -> Snapshot {
```
tamaralipows: Maybe this would be clearer if called `drain_completed_snapshot_tasks` or something. I was confused when I saw this called at a higher level

crates/tycho-client/src/feed/synchronizer.rs
```diff
@@ -713,6 +883,198 @@ where
...
+                    // self.entrypoints for any entrypoints already linked to these components
+                    // and add their contracts. Without this, a component that shares an
+                    // entrypoint with an already-tracked component would miss those contracts
+                    // until the next DCI update.
+                    self.component_tracker.contracts.extend(
+                        self.component_tracker
+                            .get_contracts_by_component(&new_component_ids),
+                    );
+                    // Also add DCI contracts from this fetch that aren't in self.entrypoints yet.
+                    self.component_tracker.contracts.extend(
```
tamaralipows: Hmm none of this can be deduplicated? I thought I saw a similar codeblock above

tamaralipows (commented):
Thank you! I'm definitely not the best person to approve this, but here are some questions for my understanding. Glad to know this works!

tamaralipows:
@claude review once

crates/tycho-client/src/feed/synchronizer.rs
louise-poole: Yes - they're pending being included in the stream messages... though I see how this is ambiguous. I'll improve the wording here!

crates/tycho-client/src/feed/synchronizer.rs
louise-poole: So the buffered deltas are ones that have already been emitted to users and we're only saving them on the chance they update a component we're waiting on a snapshot for. Snapshots are a versioned query - we send the block we want a snapshot at. This fn is just to apply any delta changes that happened after a snapshot's block. Deltas before the snapshot block are already reflected in the snapshot data and do not need to be manually applied.

tamaralipows: ah got it!

tamaralipows (approved):
Thank you!

PR #1033, opened 2026-05-20
fix(gas-estimation): Add logic for PROTOCOLS_OPTIMIZABLE_TRANSFER_IN

dianacarvalho1:
I stumbled on the need for this for single swaps because of ERC4606 case: we were counting a token transfer twice in that case (saw it in the grafana dashboard).

- Ended up adding logic for sequential and split swaps too
- Added test cases

### DIFF
--- a/crates/tycho-execution/src/encoding/evm/gas_estimator.rs
+++ b/crates/tycho-execution/src/encoding/evm/gas_estimator.rs
@@ -2,12 +2,15 @@ use num_bigint::BigUint;
 
 use crate::encoding::models::{Solution, Strategy, UserTransferType};
 
-/// Default gas cost for an ERC-20 `transferFrom` or `transfer`. Used as fallback when the token
-/// has no measured gas cost.
+/// Default gas usage for an ERC-20 `transferFrom` or `transfer`. Used as fallback when the token
+/// has no measured gas usage.
 pub const DEFAULT_TOKEN_TRANSFER_GAS: u64 = 40_000;
 
-/// Gas cost for an ERC-20 `approve`.
+/// Gas usage for an ERC-20 `approve`.
 pub const TOKEN_APPROVAL_GAS: u64 = 25_000;
+
+/// Gas usage for a permit2 transfer from the user
+pub const USER_PERMIT2_TRANSFER: u64 = 80_000;
 
 /// Callback-based protocols: the input `transferFrom` happens inside the callback, which is
 /// inside `swap()`. The gas is thus already included in `get_amount_out`.
@@ -22,6 +25,11 @@ pub const PROTOCOLS_CALLBACK: &[&str] = &[
     "velodrome_slipstreams",
     "vm:balancer_v3",
 ];
+
+/// Protocols where the router-to-pool input transfer is skipped (funds are sent directly without
+/// an intermediate router hop). The input transfer cost is therefore not double-counted — it is
+/// only charged once, on the first hop or when the optimized path applies.
+pub const PROTOCOLS_OPTIMIZABLE_TRANSFER_IN: &[&str] = &["erc4626", "maverick_v2", "uniswap_v2"];
 
 /// ProtocolWillDebit: the router must `approve(protocol)` before swapping.
 /// The protocol's `transferFrom` is inside `swap()` and already in the gas computation of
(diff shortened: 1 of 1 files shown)


crates/tycho-execution/src/encoding/evm/gas_estimator.rs
```diff
@@ -117,19 +161,153 @@ fn estimate_transfer_overhead(
...
         overhead += transfer_token_gas(token_in);
     }
 
     if PROTOCOLS_NEEDING_APPROVAL.contains(&protocol_system) {
         overhead += BigUint::from(TOKEN_APPROVAL_GAS);
     }
 
-    // Output transfer: router -> receiver/next pool (only when outputToRouter or strategy is
-    // Split).
-    if PROTOCOLS_OUTPUT_TO_ROUTER.contains(&protocol_system) || *strategy == Strategy::Split {
```
tamaralipows, line 130: I guess this shouldn't have checks for split swap since it's still possible that the output transfer is optimized? or? 

edit: nevermind I see you moved this to a higher level. Are we sure it's correct that for split swaps it always passes through the router even for the input and output too? If so I think a comment would help, not sure where to put it though. All these intricacies get a bit lost with time

crates/tycho-execution/src/encoding/evm/gas_estimator.rs
```diff
@@ -117,19 +161,153 @@ fn estimate_transfer_overhead(
 
     // Input transfer: only needed when it happens outside swap().
     // Callback protocols handle it inside the callback (part of swap gas).
-    if !PROTOCOLS_CALLBACK.contains(&protocol_system) {
+    // Protocols that can have an optimizable transfer in should not be included here either because
+    // the extra transfer is skipped
+    if !PROTOCOLS_CALLBACK.contains(&protocol_system) &&
+        !PROTOCOLS_OPTIMIZABLE_TRANSFER_IN.contains(&protocol_system) &&
+        *strategy != Strategy::Split
```
tamaralipows: And for the input of split I also don't remember - we always transfer to the router beforehand? Maybe adding a small note in the docstring here could clear that up

crates/tycho-execution/src/encoding/evm/gas_estimator.rs
```diff
@@ -60,19 +72,44 @@ pub fn estimate_gas_usage(solution: &Solution, strategy: Strategy) -> BigUint {
...
-    if ROUTER_FEES_ACTIVE {
-        if let Some(last_swap) = solution.swaps().last() {
-            let protocol: &str = &last_swap.component().protocol_system;
-            if !PROTOCOLS_OUTPUT_TO_ROUTER.contains(&protocol) && strategy != Strategy::Split {
-                total_gas += transfer_token_gas(last_swap.token_out());
-            }
+    if let Some(last_swap) = solution.swaps().last() {
+        let protocol: &str = &last_swap.component().protocol_system;
+        if (ROUTER_FEES_ACTIVE || strategy == Strategy::Split) &&
+            !PROTOCOLS_OUTPUT_TO_ROUTER.contains(&protocol)
```
tamaralipows: I would love if we assigned these if clauses to a variable to make this more readable. Then we wouldn't need the comment below. We could call this `final_swap_output_to_router` or something?

tamaralipows (approved):
Thank you! This looks good to me - I only have comments about readability, etc. And please double check if the input/output transfers of the split swaps are truly never optimized (I don't have this loaded in my head atm). I can also check later!

crates/tycho-execution/src/encoding/evm/gas_estimator.rs
dianacarvalho1: yeah I know.. this was wrong because it was assuming an extra token transfer for ALL branches of a split. which might not be true.. it depends on what comes next 😕

crates/tycho-execution/src/encoding/evm/gas_estimator.rs
dianacarvalho1: yes I think we always transfer to the router

dianacarvalho1: ai I think this condition is wrong actually 🤔

dianacarvalho1: fixed it!

PR #1034, opened 2026-05-21
feat: fermiswap integration

zach030:
## Summary

This PR adds the FermiSwap integration based on the VM integration plan:

https://propeller-heads.atlassian.net/wiki/spaces/~71202094d3d73818b5447b85231f4bd027b78c/pages/3679518721/FermiSwap+Integration+Plan+VM

There is one extra part compared to the original plan: balance tracking for the `trader_vault`.

FermiSwap keeps all tokens in the `trader_vault`, but we still need balances on each component to calculate TVL. To handle that, I added two stores:

- `store_token_pairs`: {key: token, value: list of pairs that use this token}
- `store_token_balances`: {key: token, value: absolute token balance in the `trader_vault`}

With these two stores, when a token balance changes in the `trader_vault`, we can update all related components that depend on that token.

This PR also emits token balance [updates](https://github.com/propeller-heads/tycho-indexer/blob/7f2f0720de92849d219b0e2370bdad0bd9442b5e/protocols/substreams/ethereum-fermiswap/src/modules.rs#L355-L371) for the `trader_vault` itself, since the VM integration needs to read the vault token balances during simulation.


crates/tycho-execution/contracts/src/executors/FermiSwapExecutor.sol
```diff
@@ -0,0 +1,96 @@
...
+
+import {IExecutor} from "@interfaces/IExecutor.sol";
+import {TransferManager} from "../TransferManager.sol";
+
+interface IFermiSwapper {
+    function quoteAmounts(
+        address tokenIn,
+        address tokenOut,
+        int256 amountSpecified
+    ) external view returns (uint256 amountIn, uint256 amountOut);
```
zizou0x: Is this needed?

protocols/substreams/ethereum-fermiswap/src/pb/.last_generated_hash
```diff
@@ -0,0 +1 @@
+d599594e43fdc3f2ac04d85f34aaff61dd9435c9e6c5a26cf54db6984cabe89c
```
zizou0x: What is this?

protocols/substreams/ethereum-fermiswap/substreams.yaml
```diff
@@ -0,0 +1,121 @@
...
+      map_token_balance_deltas: "engine_address=0ba5665583a4593a86cdc935ea2c8ff378c11317&trader_vault=585d44727129B9C69791B10238Ca605932938B4F&swapper_address=b1076fE3AB5e28005C7c323Bac5AC06a680d452e"
+      map_protocol_changes: "engine_address=0ba5665583a4593a86cdc935ea2c8ff378c11317&trader_vault=585d44727129B9C69791B10238Ca605932938B4F&swapper_address=b1076fE3AB5e28005C7c323Bac5AC06a680d452e"
+
+params:
+  map_protocol_components: "engine_address=0ba5665583a4593a86cdc935ea2c8ff378c11317&trader_vault=585d44727129B9C69791B10238Ca605932938B4F&swapper_address=b1076fE3AB5e28005C7c323Bac5AC06a680d452e"
+  map_token_balance_deltas: "engine_address=0ba5665583a4593a86cdc935ea2c8ff378c11317&trader_vault=585d44727129B9C69791B10238Ca605932938B4F&swapper_address=b1076fE3AB5e28005C7c323Bac5AC06a680d452e"
+  map_protocol_changes: "engine_address=0ba5665583a4593a86cdc935ea2c8ff378c11317&trader_vault=585d44727129B9C69791B10238Ca605932938B4F&swapper_address=b1076fE3AB5e28005C7c323Bac5AC06a680d452e"
```
zizou0x: Why is this duplicated?

protocols/substreams/ethereum-fermiswap/src/utils.rs
```diff
@@ -0,0 +1,29 @@
...
+    input.extend_from_slice(base_asset);
+    input.extend_from_slice(quote_asset);
+
+    let mut out = [0u8; 32];
+    let mut hasher = Keccak::v256();
+    hasher.update(&input);
+    hasher.finalize(&mut out);
+
+    format!("0x{}", hex::encode(out))
+}
```
zizou0x, line 33: We need to confirm that they can only create one pool per token pair. Has it been confirmed?

protocols/substreams/ethereum-fermiswap/src/modules.rs
```diff
@@ -0,0 +1,395 @@
...
+) {
+    let mut on_pair_registered = |event: PairRegistered, _tx: &TransactionTrace, _log: &Log| {
+        let tycho_tx: Transaction = _tx.into();
+        let component_id = component_id(&event.base_asset, &event.quote_asset);
+        let new_component = ProtocolComponent::new(component_id.as_str())
+            .with_attributes(&[
+                ("swapper_address", config.swapper_address.to_hex().into_bytes()),
+                ("oracle_address", config.engine_address.to_hex().into_bytes()),
+                ("registry_address", config.registry_address.to_hex().into_bytes()),
+                ("trader_vault_address", config.trader_vault.to_hex().into_bytes()),
```
zizou0x: I wouldn't add these, especially because they not static (the swapper is just and interface, they could change it without recreating the pool).

protocols/substreams/ethereum-fermiswap/src/modules.rs
```diff
@@ -0,0 +1,395 @@
...
+        .map(Transaction::from);
+
+    for token_key in &new_token_keys {
+        let token = hex::decode(token_key)?;
+        let Some(tx) = &last_tx else {
+            continue;
+        };
+        let balance = erc20::functions::BalanceOf { owner: config.trader_vault.clone() }
+            .call(token.clone())
+            .unwrap_or_default();
```
zizou0x, line 161: Do you know if this includes changes at the current block or not? Else it could be double accounted (once here and once below)

protocols/substreams/ethereum-fermiswap/src/modules.rs
```diff
@@ -0,0 +1,395 @@
...
+            let tx: Transaction = tx.into();
+            balance_deltas.push(BalanceDelta {
+                ord: log.ordinal,
+                tx: Some(tx),
+                token: log.address.clone(),
+                delta: delta.to_signed_bytes_be(),
+                component_id: vec![],
+            });
+        }
+    }
```
zizou0x: I'd refactor this to be like this:

- Extract all erc20 transfer events for the block, store them in a Vec.
- Filter out those not involving `config.trader_vault` -> if empty after this, early return
- From the filtered list, read the ones that are in the store and emit the corresponding BalanceDelta

This will be way more efficient

protocols/substreams/ethereum-fermiswap/src/modules.rs
```diff
@@ -0,0 +1,395 @@
...
+                    _ => continue,
+                };
+
+            let tx: Transaction = tx.into();
+            balance_deltas.push(BalanceDelta {
+                ord: log.ordinal,
+                tx: Some(tx),
+                token: log.address.clone(),
+                delta: delta.to_signed_bytes_be(),
+                component_id: vec![],
```
zizou0x: This is a bug no?

crates/tycho-execution/contracts/src/executors/FermiSwapExecutor.sol
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    function fundsExpectedAddress(
+        bytes calldata /* data */
+    )
+        external
+        view
+        returns (address receiver)
+    {
+        return msg.sender;
```
zizou0x, line 38: How does it work? Do they call transferFrom the caller? I wonder if this is correct

crates/tycho-simulation/src/evm/protocol/fermiswap/decoder.rs
```diff
@@ -0,0 +1,149 @@
...
+                ))
+            })?;
+        let _swapper_address = snapshot
+            .component
+            .static_attributes
+            .get("swapper_address")
+            .ok_or_else(|| {
+                InvalidSnapshotError::ValueError("Missing swapper address in component".to_string())
+            })?;
+        let _oracle_address = snapshot
```
zizou0x: Not used?

crates/tycho-simulation/src/evm/protocol/fermiswap/vm.rs
```diff
@@ -0,0 +1,121 @@
...
+    )?;
+    let quote_to_base_update_timestamp = quote_to_base_state.updateTimestamp;
+    let quote_to_base_price = quote_to_base_state
+        .slots
+        .first()
+        .copied()
+        .ok_or_else(|| SimulationError::FatalError("getStateCall returned no slots".into()))?;
+
+    Ok((
+        base_to_quote_update_timestamp.min(quote_to_base_update_timestamp),
```
zizou0x: Is that how it work on the contract or then use the swap direction? (meaning if only one direction is updated you can only swap in this direction)

crates/tycho-simulation/src/evm/protocol/fermiswap/state.rs
```diff
@@ -0,0 +1,416 @@
...
+            ));
+        }
+
+        let price = self.price_for_direction(token_in, token_out)?;
+        if price.is_zero() {
+            return Err(SimulationError::RecoverableError("FermiSwap oracle price is zero".into()));
+        }
+
+        let priced = amount_in * u256_to_biguint(price) / BigUint::from(PRICE_SCALE);
+        let amount_out = priced * token_scale(token_out.decimals) / token_scale(token_in.decimals);
```
zizou0x: This doesn't include price impact?

crates/tycho-simulation/src/evm/protocol/fermiswap/state.rs
```diff
@@ -0,0 +1,416 @@
...
+        pool_id: Bytes,
+        token0: Token,
+        token1: Token,
+        active: bool,
+        registry_address: Bytes,
+        trader_vault_address: Bytes,
+        trader_vault_balances: HashMap<Address, U256>,
+        oracle_update_timestamp: u32,
+        base_to_quote_price: U256,
+        quote_to_base_price: U256,
```
zizou0x: This approach looks quite complex. I'd start by using VM on each call. Doesn't need to be hybrid right now.

crates/tycho-execution/contracts/src/executors/FermiSwapExecutor.sol
zach030: removed

protocols/substreams/ethereum-fermiswap/src/pb/.last_generated_hash
zach030: It’s unused, already removed.

protocols/substreams/ethereum-fermiswap/substreams.yaml
zach030: removed

protocols/substreams/ethereum-fermiswap/src/utils.rs
zach030: The contract uses `keccak(token0, token1)` as the pair key, so it’s consistent here.

protocols/substreams/ethereum-fermiswap/src/modules.rs
zach030: yes, they are used for hybrid, all removed now.

protocols/substreams/ethereum-fermiswap/src/modules.rs
zach030: Based on the Substreams [docs](https://docs.substreams.dev/how-to-guides/develop-your-own-substreams/on-evm/eth-calls#block-state-execution), this uses the state of the block being processed, so the `balanceOf` result here will just overwrite the value in the store.

protocols/substreams/ethereum-fermiswap/src/modules.rs
zach030: I’ve updated it based on your suggestion.

protocols/substreams/ethereum-fermiswap/src/modules.rs
zach030: No, `map_token_balance_deltas` is used to track token balances in the `trader_vault`. They only get assigned to components later in `map_balance_deltas`. I added more comments here.

protocols/substreams/ethereum-fermiswap/abi/.gitkeep
```diff
@@ -0,0 +1 @@
+Please put any abi JSONs into this folder the buildscript will compile these to rust code.
```
zizou0x: We can remove it, this is just for the template

protocols/substreams/ethereum-fermiswap/src/modules.rs
```diff
@@ -0,0 +1,422 @@
...
+            .with_attributes(&[("balance_owner", config.trader_vault.as_slice())])
+            .as_swap_type("fermiswap_pool", ImplementationType::Vm);
+
+        new_pair_changes.push(TransactionEntityChanges {
+            tx: Some(tycho_tx.clone()),
+            entity_changes: vec![EntityChanges {
+                component_id: component_id.clone(),
+                attributes: vec![Attribute {
+                    name: ACTIVE_ATTRIBUTE.to_string(),
+                    value: vec![{ 0u8 }], // default false at creation
```
zizou0x: So when the pool is created it's not active by default?

zizou0x: Can we use the "paused" attribute instead please? To be consistent with other protocols. Ideally we will treat this as a special attribute and filter out paused pools on the client side..

https://github.com/propeller-heads/tycho/blob/024ded3966d1521cf2793ea6cb24acb6bfa339f5/protocols/substreams/ethereum-fluid/src/modules.rs#L336-L340


We need to set it to [1u8] (not like in the code shared above, this is not ideal) and delete it to unpause. Please see change_component_pause_state function in the SDK.

protocols/substreams/ethereum-fermiswap/src/modules.rs
```diff
@@ -0,0 +1,422 @@
...
+    }
+
+    let mut transfers = Vec::new();
+    for raw_tx in block.transactions() {
+        let tycho_tx: Transaction = raw_tx.into();
+        for log in raw_tx
+            .calls
+            .iter()
+            .filter(|call| !call.state_reverted)
+            .flat_map(|call| &call.logs)
```
zizou0x: Do you know the `logs_with_calls()` function implemented by Substreams? I think it's a helper for exactly this

protocols/substreams/ethereum-fermiswap/src/modules.rs
```diff
@@ -0,0 +1,422 @@
...
+        } else {
+            transfer.value
+        };
+
+        balance_deltas.push(BalanceDelta {
+            ord,
+            tx: Some(tx),
+            token,
+            delta: delta.to_signed_bytes_be(),
+            component_id: vec![],
```
zizou0x, line 243: Can you add a comment explaining why we don't need component id here please, it's not obvious

protocols/substreams/ethereum-fermiswap/src/modules.rs
```diff
@@ -0,0 +1,422 @@
...
+        balance_deltas.sort_unstable_by_key(|delta| delta.ord);
+        return Ok(BlockBalanceDeltas { balance_deltas });
+    }
+
+    for (tx, ord, token, transfer) in vault_transfers {
+        let token_key = hex::encode(&token);
+        if new_token_keys.contains(&token_key) ||
+            token_pairs_store
+                .get_last(&token_key)
+                .is_none()
```
zizou0x: Aren't these two checks redundant?

protocols/substreams/ethereum-fermiswap/src/modules.rs
```diff
@@ -0,0 +1,422 @@
...
+            .and_modify(|ord| {
+                if *ord >= delta.ord {
+                    panic!(
+                        "Invalid ordinal sequence for token balance {token_key}: {} >= {}",
+                        *ord, delta.ord
+                    );
+                }
+                *ord = delta.ord;
+            })
+            .or_insert(delta.ord);
```
zizou0x: Please avoid panics. We have everything we need to sort things again here instead of panicking. Also it would make this function easier.

protocols/substreams/ethereum-fermiswap/src/modules.rs
```diff
@@ -0,0 +1,422 @@
...
+                balance_deltas.push(BalanceDelta {
+                    ord: tx.index,
+                    tx: Some(tx.clone()),
+                    token,
+                    delta: balance.to_signed_bytes_be(),
+                    component_id: component.id.as_bytes().to_vec(),
+                });
+            }
+        }
+    }
```
zizou0x: Wouldn't this already be accounted for below? 

Shouldn't the logic here just be to collect all changes from the token balance store and plug them into the affected components for the token?

protocols/substreams/ethereum-fermiswap/src/modules.rs
```diff
@@ -0,0 +1,422 @@
...
+
+    for trx in block.transactions() {
+        let tx = Transaction {
+            to: trx.to.clone(),
+            from: trx.from.clone(),
+            hash: trx.hash.clone(),
+            index: trx.index.into(),
+        };
```
zizou0x: We have a From implementation for this

protocols/substreams/ethereum-fermiswap/src/modules.rs
```diff
@@ -0,0 +1,422 @@
...
+                    .is_some()
+                {
+                    builder.add_entity_change(&EntityChanges {
+                        component_id,
+                        attributes: vec![Attribute {
+                            name: ACTIVE_ATTRIBUTE.to_string(),
+                            value: vec![if ev.active { 1u8 } else { 0u8 }],
+                            change: ChangeType::Creation.into(),
+                        }],
+                    });
```
zizou0x: So here this should delete the "paused" attribute if it's active and set it to [1u8] if it's not active

protocols/substreams/ethereum-fermiswap/src/modules.rs
```diff
@@ -0,0 +1,422 @@
...
+
+            builder.add_contract_changes(&contract_change);
+        });
+
+    extract_contract_changes_builder(
+        &block,
+        |addr| {
+            addr == config.engine_address ||
+                addr == config.swapper_address ||
+                addr == config.registry_address
```
zizou0x: Don't we need the vault too? It's registered as a contract above: 

```
            .with_contracts(&[
                config.swapper_address.as_slice(),
                config.engine_address.as_slice(),
                config.trader_vault.as_slice(),
                config.registry_address.as_slice(),
            ])
```

protocols/substreams/ethereum-fermiswap/src/modules.rs
zach030: I changed the attribute to "paused" now and it's defaults to `false` now. I checked the contract, and yes, it’s [initialized](https://etherscan.io/tx/0xf266482bf30c6f815b1b08b56aea459762963b48ebc029650eb75635ffa34033) to `true`, but I have no idea why they [manually](https://etherscan.io/tx/0xae418cef9d842981ec7c36c18fa40c80e1f4a47c2a405a669792c719ae8b4af4) set it to `true` again after creating the pair.

protocols/substreams/ethereum-fermiswap/src/modules.rs
zach030: added more comments here

protocols/substreams/ethereum-fermiswap/src/modules.rs
zach030: No, `new_token_keys.contains(&token_key)` is used to filter out new tokens, while `token_pairs_store.get_last(&token_key).is_none()` is used to filter out tokens that don’t belong to FermiSwap, since the `trader_vault` contract may also receive transfers for other tokens.

protocols/substreams/ethereum-fermiswap/src/modules.rs
zach030: Not quite. The loop below only handles tokens that moved in the current block. For a new component, we need to initialize all token balances, including the ones that didn’t move.

For example, if a new `WETH/USDT` component is created in a block where only `WETH` moved in the vault, the loop below would only update `WETH` and the initial `USDT` balance would be missing.
So the first loop seeds new components from the global vault balance store, and the second loop only applies this block’s token movements to existing components.

protocols/substreams/ethereum-fermiswap/src/modules.rs
zach030: I switched it to use `change_component_pause_state` now.

protocols/substreams/ethereum-fermiswap/src/modules.rs
zach030: We don’t need the trader_vault contract storage. I only added it to `with_contracts` to avoid `Account not found` error, since we need to `upsert_token_balance` for trader_vault.

protocols/substreams/ethereum-fermiswap/src/modules.rs
zizou0x: There was a misunderstanding. We need to set this attribute to [1u8] when it's paused and deleted it when it's not paused. Setting it to [0u8] doesn't means "false".

I've checked and indeed at creation it's not active until they manually activate it:

```
cast call 0xb1076fe3ab5e28005c7c323bac5ac06a680d452e \
  0xae131deb000000000000000000000000a0b86991c6218b36c1d19d4a2e9eb0ce3606eb48000000000000000000000000dac17f958d2ee523a2206206994597c13d831ec7 \
  --from 0x0000000000000000000000000000000000000000 \
  --block 25194560 \
  --rpc-url "$ETH_RPC_URL"
```

returns false. So we can just set it to [1u8] here.

crates/tycho-execution/contracts/src/executors/FermiSwapExecutor.sol
tamaralipows: So this is actually called by our TychoRouter. It's msg.sender since we staticcall instead of delegatecalling. This should return the TychoRouter address - all correct!

protocols/substreams/ethereum-fermiswap/src/utils.rs
```diff
@@ -27,3 +29,18 @@ pub fn component_id(base_asset: &[u8], quote_asset: &[u8]) -> String {
 
     format!("0x{}", hex::encode(out))
 }
+
+pub fn lane_index(base_asset: &[u8], quote_asset: &[u8]) -> String {
```
zizou0x, line 40: Can you add docs that explains what this does with reference to the code please?

protocols/substreams/ethereum-fermiswap/src/modules.rs
```diff
@@ -420,3 +497,66 @@ fn map_protocol_changes(
...
+    target: &[u8],
+    lane_index: &BigInt,
+    update_timestamp: &BigInt,
+) -> Option<String> {
+    if target != config.engine_address.as_slice() {
+        return None;
+    }
+    if update_timestamp.to_u64() != block_timestamp {
+        return None;
+    }
```
zizou0x: What's the logic behind these? Do we expect them or not and why is it safe to ignore them?

protocols/substreams/ethereum-fermiswap/src/modules.rs
```diff
@@ -93,6 +106,18 @@ fn store_pairs(pair_changes: BlockEntityChanges, store: StoreSetIfNotExistsProto
...
 }
 
+#[substreams::handlers::store]
+fn store_lane_components(pair_changes: BlockEntityChanges, store: StoreSetIfNotExistsString) {
+    for tx_changes in pair_changes.changes {
+        for component in tx_changes.component_changes {
+            let base_asset = &component.tokens[0];
+            let quote_asset = &component.tokens[1];
+            store.set_if_not_exists(0, lane_index(base_asset, quote_asset), &component.id);
+            store.set_if_not_exists(0, lane_index(quote_asset, base_asset), &component.id);
```
zizou0x: Safe because there can only be one component per token pairs, right?

protocols/substreams/ethereum-fermiswap/src/modules.rs
```diff
@@ -420,3 +497,66 @@ fn map_protocol_changes(
...
+    block_overridden_components: &mut HashSet<String>,
+    component_id: String,
+    block_number: u64,
+    block_timestamp: u64,
+) {
+    // A batch can include repeated updates for the same component. The override value is
+    // block-wide, so emit it once per component per transaction.
+    if !block_overridden_components.insert(component_id.clone()) {
+        return;
+    }
```
zizou0x: I'd prefer to emit these whenever there is an update, even if it somehow happens twice in the same block

protocols/substreams/ethereum-fermiswap/src/modules.rs
```diff
@@ -420,3 +497,66 @@ fn map_protocol_changes(
...
+    if target != config.engine_address.as_slice() {
+        return None;
+    }
+    if update_timestamp.to_u64() != block_timestamp {
+        return None;
+    }
+
+    // Registry calldata only contains `laneIndex`, not the token pair. `laneIndex` is
+    // keccak256(abi.encode(tokenA, tokenB)), so this is a one-way lookup through the lane store
+    // populated when the pair was created.
```
zizou0x: Interesting, so updates target specific pairs, right?

protocols/substreams/ethereum-fermiswap/src/modules.rs
zach030: I think I need to remove this check. In theory `target` should be `engine_address` and `update_timestamp` should match the current block, but the registry already allows some deviation here: (`MAX_UPDATE_AGE` and `MAX_UPDATE_LEAD_TIME` are both 0 now)
```solidity
uint256 ts = updateTimestamp;
// Comparing a user-supplied timestamp against `block.timestamp` within a configurable
// window is the intended freshness check, not a vulnerability to miner manipulation.
// slither-disable-next-line timestamp
if (ts + MAX_UPDATE_AGE < block.timestamp) revert InvalidUpdateTimestamp();
// slither-disable-next-line timestamp
if (ts > block.timestamp + MAX_UPDATE_LEAD_TIME) revert InvalidUpdateTimestamp();
```
it’s better to stay consistent with the onchain state.

protocols/substreams/ethereum-fermiswap/src/modules.rs
zach030: From the contract code, `(token0, token1)` and `(token1, token0)` are actually two different pairs, so they should be treated as different components. I already changed this and now only keep the `lane_index(base_asset, quote_asset)` key.

protocols/substreams/ethereum-fermiswap/src/modules.rs
zach030: yes

protocols/substreams/ethereum-fermiswap/integration_test.tycho.yaml
```diff
@@ -0,0 +1,62 @@
...
+          balance_owner: "0x585d44727129B9C69791B10238Ca605932938B4F"
+        creation_tx: "0x74bfa8b999532fe15fadd2c1ead24f881e2f375abe26c1b730f71de1d8e440e5"
+        skip_simulation: true # Fair price for this pair is stale at this test's stop block.
+        skip_execution: true
+      - id: "0x2b2f5776e38002e0c013d0d89828fdb06fee595ea2d5ed4b194e3883e823e350"
+        tokens:
+          - "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2" # WETH
+          - "0xdAC17F958D2ee523a2206206994597C13D831ec7" # USDT
+        static_attributes:
+          balance_owner: "0x585d44727129B9C69791B10238Ca605932938B4F"
```
zizou0x: Is this still a static attribute?

zach030: removed

zizou0x (approved):
LGTM

PR #1045, opened 2026-05-25
fix(tycho-ethereum): retry next slot candidate on transport errors in slot detection

tvinagre:
## Summary

On Arbitrum, `debug_traceCall` with `prestateTracer` includes storage slots from precompile addresses (e.g. `0xa4b05fff...`) alongside the actual token contract slots. When the user has zero balance, a precompile zero-value slot ties with the real balance slot in priority and gets tested first. The `eth_call` state override on a precompile fails with `"overriding address ... not allowed"`, which was treated as a terminal error — never trying the remaining candidates.

- **Retry on transport error**: when an `eth_call` slot test fails with a transport error (e.g. the RPC rejects a state override on a precompile), drop that candidate and retry with the next one. Previously this was a terminal error.
- **Prioritise token-address slots**: sort candidates so slots stored at the token's own address come first, with value-distance as a tiebreaker. Slots from foreign addresses (precompiles, external storage contracts) are pushed to the end and still tested if all token-address candidates fail. This avoids wasting RPC calls on precompile slots in the common case while remaining correct for tokens that store balances in a separate contract.
- **Remove unsafe `.expect()`**: replace `all_slots.first().expect(...)` with a `let…else` that records `TokenNotInTrace` and continues, avoiding a panic if the invariant is ever violated.

## Test plan
- [x] New Arbitrum integration test (`test_detect_arb_usdc_balance_slots`) confirms detection succeeds for both zero-balance and non-zero-balance holders on Arbitrum USDC (requires `ARB_RPC_URL`)
- [x] New unit test (`test_sort_slots_by_priority_foreign_address_last`) verifies foreign-address slots are sorted after token-address slots regardless of value distance
- [x] New unit tests (`test_transport_error_schedules_retry_with_remaining`, `test_transport_error_with_no_remaining_slots_is_terminal`) cover the retry-on-error behavior without RPC
- [x] All existing slot detector tests pass

### DIFF
--- a/crates/tycho-ethereum/src/services/entrypoint_tracer/balance_slot_detector.rs
+++ b/crates/tycho-ethereum/src/services/entrypoint_tracer/balance_slot_detector.rs
@@ -89,8 +89,8 @@ mod tests {
 
     use super::{BalanceStrategy, SlotDetectionStrategy, *};
     use crate::test_fixtures::{
-        TestFixture, ARB_USDC_HOLDER_ADDR, ARB_USDC_STR, ARB_WETH_STR, STETH_STR,
-        USDC_HOLDER_ADDR, USDC_STR, USDT_STR, WETH_STR,
+        TestFixture, ARB_USDC_HOLDER_ADDR, ARB_USDC_STR, ARB_WETH_STR, STETH_STR, USDC_HOLDER_ADDR,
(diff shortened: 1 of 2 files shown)


louise-poole (approved):
Looks good to me!

PR #1048, opened 2026-05-26
feat: Inject NativeWrapperState into ProtocolStream

tamaralipows:
### Problem:
Automatically include wraps and unwraps in Fynd.

### Main change:
Add a stateless 1:1 bridge component between a chain's native token and its wrapped counterpart (e.g. ETH/WETH, BNB/WBNB, POL/WPOL).

The NativeWrapperState is auto-injected into the protocol stream by ProtocolStreamBuilder, so Fynd can include wrap/unwrap hops in routes without any manual registration.

**Note that this will result in an extra hop in fynd.**
So there were two hops, but we insert an unwrap in between, this will now count as three hops.

I initially considered doing this entirely on the fynd side but ultimately it turned out to be so messy and horrible, having to manually insert a "dummy wrap/unwrap" node at each mention of eth/weth, resulting in messier algorithm code. It was much simpler to tackle this at the source.

Main file to check: `crates/tycho-simulation/src/evm/stream.rs`

### Related necessary fix:
The WethExecutor and "weth" protocol system were Ethereum-specific in naming but chain-agnostic in implementation. Rename to WrapExecutor / "wrap" so the wrapping bridge works naturally on BSC (BNB/WBNB),
Polygon (POL/WPOL), and all ETH L2s without confusion.

Also add WrapExecutor to the deploy script for Arbitrum, Polygon, and BSC which were previously missing it.

### Tested using: 

Ethereum:
`cargo run --example quickstart -- --sell-token 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2 --buy-token 0x0000000000000000000000000000000000000000 --sell-amount 0.01`
[Transaction here.](http://etherscan.io/tx/0x23b0316d1ea2163af490b765b548e18fcda0b27462909874a20ce1f2245e438f)

BSC:
`cargo run --example quickstart -- --chain bsc --sell-token 0xbb4CdB9CBd36B01bD1cBaEBF2De08d9173bc095c --buy-token 0x0000000000000000000000000000000000000000 --sell-amount 0.01`
(no transaction since not deployed yet on bsc)

Result:
```
==================== Received block 100609622 ====================

The best swap (out of 1 possible pools) is:
Protocol: wrap
Pool address: "native_wrapper"
Swap: 0.010000 WBNB -> 0.010000 BNB 

            Price: 1.000000 BNB per WBNB,
            1.000000 WBNB per BNB
```


crates/tycho-execution/contracts/src/executors/NativeWrapExecutor.sol
```diff
@@ -15,20 +15,20 @@ interface IWETH is IERC20 {
     function withdraw(uint256) external;
 }
 
-error WethExecutor__InvalidDataLength();
-error WethExecutor__ZeroAddres();
```
tamaralipows, line 19: Yikes

crates/tycho-execution/model/src/model/executors.rs
```diff
@@ -244,7 +244,7 @@ impl Executor {
                 output_to_router: false,
             }),
             // https://github.com/propeller-heads/tycho-execution/blob/0454514f4f6ccff55dcaa8e3abbb4ac494d89eba/foundry/src/executors/WethExecutor.sol#L76
```
dianacarvalho1: this link will need to be updated too

crates/tycho-execution/src/encoding/evm/swap_encoder/wrap.rs
```diff
@@ -87,12 +87,12 @@ mod tests {
...
 
-        write_calldata_to_file("test_encode_weth_wrapping", hex_swap.as_str());
+        write_calldata_to_file("test_encode_wrap_wrapping", hex_swap.as_str());
     }
 
     #[test]
-    fn test_encode_weth_unwrapping() {
-        // wETH -> (weth) -> ETH
+    fn test_encode_wrap_unwrapping() {
+        // WETH -> (wrap) -> ETH
```
dianacarvalho1: unwrap*

dianacarvalho1: ah I see that it is "wrap" as in the protocol name. but it is a bit strange in the little diagram (won't insist though)

crates/tycho-simulation/src/evm/protocol/native_wrapper/state.rs
```diff
@@ -0,0 +1,153 @@
...
+const NATIVE_WRAPPER_PROTOCOL_TYPE: &str = "NativeWrapper";
+const WRAP_GAS: u64 = 25_000;
+
+/// Stateless 1:1 bridge between a chain's native token and its wrapped
+/// counterpart (e.g. ETH ↔ WETH).
+///
+/// This component is auto-injected by `ProtocolStreamBuilder` so every
+/// consumer automatically sees the bridge without manual wiring.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct WrapperState {
```
dianacarvalho1: wrapper or native wrapper (name of file)?

crates/tycho-simulation/src/evm/protocol/native_wrapper/state.rs
```diff
@@ -0,0 +1,153 @@
...
+    },
+    Bytes,
+};
+
+use crate::protocol::models::ProtocolComponent;
+
+pub const NATIVE_WRAPPER_ID: &str = "native_wrapper";
+const NATIVE_WRAPPER_PROTOCOL_SYSTEM: &str = "wrap";
+const NATIVE_WRAPPER_PROTOCOL_TYPE: &str = "NativeWrapper";
+const WRAP_GAS: u64 = 25_000;
```
dianacarvalho1: if I remember correctly wrapping and unwrapping have slightly different gas!

crates/tycho-simulation/src/evm/protocol/native_wrapper/state.rs
```diff
@@ -0,0 +1,153 @@
...
+        amount_in: BigUint,
+        token_in: &Token,
+        token_out: &Token,
+    ) -> Result<GetAmountOutResult, SimulationError> {
+        self.validate_tokens(token_in, token_out)?;
+        Ok(GetAmountOutResult::new(
+            amount_in,
+            WRAP_GAS
+                .to_biguint()
+                .expect("infallible"),
```
dianacarvalho1: 😂

crates/tycho-simulation/src/evm/protocol/native_wrapper/state.rs
```diff
@@ -0,0 +1,153 @@
...
+            self.clone_box(),
+        ))
+    }
+
+    fn get_limits(
+        &self,
+        sell_token: Bytes,
+        buy_token: Bytes,
+    ) -> Result<(BigUint, BigUint), SimulationError> {
+        let valid_pair = (sell_token == self.native_token.address &&
```
dianacarvalho1: we can just use the `self.validate_tokens` no?

dianacarvalho1 (commented):
Thank you so much @tamaralipows 🙏🏼 Overall I like this!! ✨ 
I have one extra request though: in encoding we are adding a weth<->eth bridge automatically if needed but I forgot to add there the gas for each operation. Can you add the same value there that you used in the state please? (it was introduced in [here in this PR](https://github.com/propeller-heads/tycho-indexer/pull/983/changes#diff-b82ab80cba721035acc83041363f0f183ac562d44da4645739ff4d03606a4188R402))

crates/tycho-simulation/src/evm/protocol/native_wrapper/state.rs
tamaralipows: I realized `WrapperState` might not have been the best name, since the word `Wrapper` is a somewhat reserved general programming word, it could refer to a function that wraps around a state to do something else...
I've renamed this to `NativeWrapperState`, but I kept the `WrapExecutor` and the "wrap" protocol name as they are since I thought they could be inferred better from context. I might be wrong though since I've been staring at this for too long and now the word "Wrap" starts to sound weird.
I'm open to changing those too to `NativeWrapExecutor`, let me know if you would prefer that. No strong opinion.

crates/tycho-execution/model/src/model/executors.rs
tamaralipows: Oh actually these are all referencing the old repo. Will update.

crates/tycho-execution/config/executor_addresses.json
```diff
@@ -28,33 +28,36 @@
...
-    "weth": "0x50905E0f94350f9e533Cd275667AFC728265960D"
+    "wrap": "0x50905E0f94350f9e533Cd275667AFC728265960D"
   },
   "arbitrum": {
     "uniswap_v2": "0xd0a1f74d4d77834feca0efa01ca6e0161dbf1f57",
     "uniswap_v3": "0xCaAac0C6193E3e2e3E8E94bAf6367F75BaE591C9",
     "pancakeswap_v3": "0xCaAac0C6193E3e2e3E8E94bAf6367F75BaE591C9",
-    "uniswap_v4": "0xdb696336F7A5F9048252664A3475C194dAe0e62f"
+    "uniswap_v4": "0xdb696336F7A5F9048252664A3475C194dAe0e62f",
+    "wrap": "0x0000000000000000000000000000000000000000"
```
louise-poole: Whenever we put a placeholder address like this it would be nice if we added also a comment indicating it's a placeholder until actually deployed. Sometimes the zero address represents special cases/handling or might mean wrapping/unwrapping is not available at all on that chain (for some reason).

crates/tycho-execution/src/encoding/evm/tycho_encoders.rs
```diff
@@ -87,7 +87,7 @@ impl TychoRouterEncoder {
     /// Returns a new solution with added wrapping/unwrapping swaps if the original solution
     /// contains a swap that goes from ETH to WETH or vice versa but doesn't include the
     /// corresponding wrapping or unwrapping swap.
-    fn add_weth_swaps(&self, solution: &Solution, chain: &Chain) -> Solution {
+    fn add_wrap_swaps(&self, solution: &Solution, chain: &Chain) -> Solution {
```
louise-poole: The docstring here still references ETH/WETH explicitly 👀

crates/tycho-execution/src/encoding/evm/tycho_encoders.rs
```diff
@@ -128,21 +128,15 @@ impl TychoRouterEncoder {
     fn _wrapping_bridge(&self, token_a: &Bytes, token_b: &Bytes, chain: &Chain) -> Option<Swap> {
         let eth = chain.native_token();
         let weth = chain.wrapped_native_token();
```
louise-poole: Can you update these var names to native/wrapped_native?

crates/tycho-simulation/src/evm/protocol/etherfi/state.rs
```diff
@@ -301,7 +301,7 @@ impl ProtocolSim for EtherfiState {
                 amount_out,
                 BigUint::from(60_182u32), // weeth.unwrap function gas used
                 Box::new(new_state),
-            ))
+            ));
```
louise-poole, line 304: I wonder why there are so many formatting changes 🤔

crates/tycho-simulation/src/evm/protocol/native_wrapper/state.rs
```diff
@@ -0,0 +1,135 @@
...
+        errors::{SimulationError, TransitionError},
+        protocol_sim::{Balances, GetAmountOutResult, ProtocolSim},
+    },
+    Bytes,
+};
+
+use crate::protocol::models::ProtocolComponent;
+
+pub const NATIVE_WRAPPER_ID: &str = "native_wrapper";
+const NATIVE_WRAPPER_PROTOCOL_SYSTEM: &str = "wrap";
```
louise-poole: Personally would prefer this be called `native_wrap`. Feels a bit less ambiguous to me.

crates/tycho-simulation/src/evm/protocol/native_wrapper/state.rs
```diff
@@ -0,0 +1,135 @@
...
+                ),
+                None,
+            ));
+        }
+        Ok(())
+    }
+}
+
+#[typetag::serde]
+impl ProtocolSim for NativeWrapperState {
```
louise-poole, line 75: Should we maybe add some small unit tests for this new NativeWrapperState?

crates/tycho-execution/contracts/test/protocols/WrapExecutor.t.sol
```diff
@@ -5,18 +5,18 @@ import {CommonBase} from "../../lib/forge-std/src/Base.sol";
...
 import {StdAssertions} from "../../lib/forge-std/src/StdAssertions.sol";
 import {StdChains} from "../../lib/forge-std/src/StdChains.sol";
 import {StdCheats, StdCheatsSafe} from "../../lib/forge-std/src/StdCheats.sol";
 import {StdUtils} from "../../lib/forge-std/src/StdUtils.sol";
 import {TestUtils} from "../TestUtils.sol";
 
-contract WethExecutorExposed is WethExecutor {
-    constructor(address wethAddress) WethExecutor(wethAddress) {}
+contract WrapExecutorExposed is WrapExecutor {
+    constructor(address wethAddress) WrapExecutor(wethAddress) {}
```
louise-poole: The constructor param should now be wrappedAddress here no?

crates/tycho-simulation/src/evm/stream.rs
```diff
@@ -595,6 +606,37 @@ impl ProtocolStreamBuilder {
...
         Ok(stream)
     }
 }
+
+/// Wraps a decoded protocol stream to inject a `NativeWrapperState` component
+/// on the first successful update.
+///
+/// Skips injection for chains where the native and wrapped-native tokens share
+/// the same address (e.g. Starknet).
+fn inject_native_wrapper(
```
louise-poole, line 620: Can you add a test for this? That is injects in the first message only?

louise-poole (commented):
Nice!! Looks very clean, thank you! Just some very minor comments/test requests, otherwise I'm happy 👍

crates/tycho-execution/config/executor_addresses.json
dianacarvalho1: you can't add comments in json files 😅

crates/tycho-simulation/src/evm/protocol/native_wrapper/state.rs
dianacarvalho1: > since the word Wrapper is a somewhat reserved general programming word, it could refer to a function that wraps around a state to do something else

I actually agree with your reasoning and would add to that that there are some other wrapped tokens (like wBTC in ethereum) that this code doesn't support. I would prefer if we say it's native wrapping everywhere for clarity (including the executor)

crates/tycho-simulation/src/evm/protocol/etherfi/state.rs
tamaralipows: ci was failing without this :(

crates/tycho-simulation/src/evm/protocol/native_wrapper/state.rs
tamaralipows: Ah I had them initially but honestly didn't feel they were testing much :/ I guess I can put them back to check the amount out and gas, etc

dianacarvalho1 (approved):
✨ perfection ✨

crates/tycho-execution/config/executor_addresses.json
louise-poole: Boooo - be better json!

louise-poole (approved):
Thanks!!! 🙏

crates/tycho-simulation/src/evm/stream.rs
```diff
@@ -595,6 +606,96 @@ impl ProtocolStreamBuilder {
...
+    stream: impl Stream<Item = Result<Update, StreamDecodeError>>,
+    chain: Chain,
+) -> impl Stream<Item = Result<Update, StreamDecodeError>> {
+    let has_distinct_wrapper = chain.native_token().address != chain.wrapped_native_token().address;
+    let mut injected = false;
+
+    stream.map(move |result| {
```
kayibal: I don't love the map usage here, is there no other way to achieve this? It basically means that you will have these if else check here on every message.

tamaralipows: I had looked into other options before but they seemed less readable to me. Could you look at the latest commit and let me know if you find that better? I find it a bit worse...

PR #1049, opened 2026-05-26
feat: add Balancer V3 reclamm pool types

johngrantuk:
Adds support for Balancer V3 reClamm pools (rebranded as AutoRange).
Initial test pool can be found here: https://balancer.fi/pools/ethereum/v3/0xda66e8ddf9959e4db759bfd06256730d8a8b2d13


zach030:
Hi! Thanks for the work.

This part wasn’t changed by you, but I suspect the logic [here](https://github.com/propeller-heads/tycho-indexer/blob/89a1dd68f8f4f01bd794807f4499de25bd521595/protocols/substreams/ethereum-balancer-v3/src/modules.rs#L478-L500) might be wrong.
I don’t think we should use `wrapped_token` as the `component_id` to look up the related tokens.

I have a rough idea you could use as a reference. You can look at the previous liquidityBuffer integration [PR](https://github.com/propeller-heads/tycho-protocol-sdk/pull/235) I worked on. The flow would be:

1. Detect the liquidityBuffer creation: [1_map_components.rs](https://github.com/propeller-heads/tycho-protocol-sdk/blob/151a2e4568bef6b6c06a7e428faf0b7a6eb00f51/substreams/ethereum-balancer-v3/src/modules/1_map_components.rs#L61-L70)
2. Write the mapping (`wrapped_token`->`underlying_token`) into a store: [4_store_token_mapping.rs](https://github.com/propeller-heads/tycho-protocol-sdk/blob/feat/boosted-pool/substreams/ethereum-balancer-v3/src/modules/4_store_token_mapping.rs)
3. When updating vault reserves, read the `wrapped_token` / `underlying_token` relationship from that store:
[7_map_protocol_changes.rs](https://github.com/propeller-heads/tycho-protocol-sdk/blob/151a2e4568bef6b6c06a7e428faf0b7a6eb00f51/substreams/ethereum-balancer-v3/src/modules/7_map_protocol_changes.rs#L414-L431)

zach030:
Hey @johngrantuk! I debugged it and pushed some updates.
There were mainly two issues:
- Reading balances from `Swap`, `LiquidityAdded`, and `LiquidityRemoved` can drift, because Balancer-v3 has some fees that don’t show up in those events.
- The `reservesOf` update logic also had issues. The previous `wrapped_token` / `underlying_token` handling couldn’t update balances correctly.

I changed the balance logic to decode the `_poolTokenBalances` mapping instead, and added a token mapping store to track the `wrapped_token` -> `underlying_token` relation. The tests pass locally now.

Could you review whether this matches the balancer-v3 design, especially around token balance handling, and try running more test cases?

crates/tycho-simulation/src/evm/protocol/vm/adapter_contract.rs
```diff
@@ -68,7 +68,7 @@ where
         let selector = "price(bytes32,address,address,uint256[])";
 
         let res = self
-            .call(selector, args, overwrites, None, U256::from(0u64), None, block_overrides)?
+            .call(selector, args, overwrites, Some(Address::ZERO), U256::from(0u64), None, block_overrides)?
```
die-herdplatte: afaict this exists due to the [assertion `tx.origin == address(0)`](https://github.com/balancer/balancer-v3-monorepo/blob/7861ea2785b96dd10681ff1b8dfe56b36cc202b6/pkg/solidity-utils/contracts/helpers/EVMCallModeHelpers.sol#L18).
Are all other integrations fine with this change?

I think it should at least be documented why the address here is set explicitly (and not on the other adapter contract calls).
Or somehow wire the protocol type into the struct and only override the default address for Balancer.

zach030:
### Integration-test validation (incl. reCLAMM pools)

Ran the `vm:balancer_v3` integration test end-to-end (84 blocks, 8 pools, incl. the two reCLAMM pools `0x3aa14d…` and `0x1bb792…`):

| Metric | Result |
|---|---|
| Simulations (`get_amount_out`) | 1344 |
| Reverts / failures / errors | 0 / 0 / 0 |
| Executed vs simulated (slippage) | 0.0000% on all 1344 |

Two independent comparisons, both exact to the wei:

1. **Simulation vs on-chain quote** — spot-checked `get_amount_out` against `querySwapExactIn` at the same block (`--from 0x0`) for 3 pools × both directions: 6/6 identical.
2. **Simulation vs execution** — 1344/1344 `executed_amount == simulated_amount` (fee-adjusted), 0 slippage, routed through the already-deployed `BalancerV3Executor` (`0xf9AE8127…00cA10`). `tycho-execution` is unchanged, so no redeploy is needed.

Both directions of the two reCLAMM pools pass — they previously reverted (`BalanceNotSettled` / arithmetic under/overflow) until the accompanying fixes landed.

Depends on the companion fixes: simulation #1234 (VM balance-map refresh + configurable price-query caller) and indexer #1232 (pending account balances in `PendingDeltas` snapshots).

protocols/substreams/ethereum-balancer-v3/src/pool_balances.rs
die-herdplatte, line 1: I'm a bit confused as to what this whole seeding mechanism should achieve.
From what I can tell, all this exists to determine `is_seeded`.
But storage slots are initialized to zero, so `value.new_raw.clone() - value.old_raw.clone()` would compute the correct value in any case.

And, since the absolute balances are read directly from storage, can't we drop `map_pool_balance_seed_events`, `store_seeded_pool_balances`, `map_relative_balances`, and `store_balances` as a whole?

protocols/substreams/ethereum-balancer-v3/integration_test_base_balancer_v3.tycho.yaml
```diff
@@ -0,0 +1,35 @@
...
+    expected_components:
+      - id: "0xCC8DAD6E87deb75fA82E7c4dd08B58d368214812"
+        tokens:
+          - "0x74299A718b2c44483a27325d7725F0B2646DE3B1"
+          - "0xC768c589647798a6EE01A91FdE98EF2ed046DBD6"
+        static_attributes:
+          normalized_weights: "0x5b22307830623161326263326563353030303030222c22307830326336386166306262313430303030225d"
+          fee: "0x0aa87bee538000"
+          pool_type: "0x5765696768746564506f6f6c466163746f7279"
+        skip_simulation: true
```
die-herdplatte: Why is this skipped?

protocols/substreams/ethereum-balancer-v3/arbitrum-balancer-v3.yaml
```diff
@@ -0,0 +1,112 @@
+specVersion: v0.1.0
+package:
+  name: "arbitrum_balancer_v3"
+  version: v0.5.0
```
die-herdplatte, line 4: Why are the versions of the new deployments starting at 0.5.0 and not 0.4.3?

die-herdplatte (commented):
Thanks for the PR!

crates/tycho-simulation/src/evm/protocol/vm/adapter_contract.rs
die-herdplatte: (This comment was made a long time ago, it will be resolved with #1234)

zach030:
### Removed the relative-balance pipeline — balances are now reported as absolute values

Following @die-herdplatte's review on `pool_balances.rs`: the seeding mechanism was dead code, and the relative-delta pipeline as a whole is unnecessary once balances come from storage. Removed in `eefbf1f47`:

- `map_pool_balance_seed_events`
- `store_seeded_pool_balances`
- `map_relative_balances`
- `store_balances`

`map_protocol_changes` now calls `pool_balances::absolute_pool_balances` and feeds `add_balance_change` directly, since `BalanceChange.balance` is absolute by definition. The module graph goes from 9 modules to 5; the diff is 121 insertions, 442 deletions across 7 files.

#### Why the seeding branch could never diverge

Audited every write to `_poolTokenBalances` in the Vault (balancer-v3-monorepo @ `80fd29ce`): `swap`, `addLiquidity`, `removeLiquidity`, `initialize` and `removeLiquidityRecovery` each emit one of the three gated events, and the only event-less write path — the sync on `disableRecoveryMode` — round-trips the raw value unchanged. A tracked pool's first observed write therefore always starts from a zero slot, so `new_raw - old_raw == new_raw` and both branches of the old `pool_balance_delta` returned the same value for every possible input.

Reporting absolute values also self-corrects: if a write were ever missed, the next observed write restores the correct balance, whereas the additive store would drift permanently.

#### What is kept, and why

- **Event gating.** `_poolTokenBalances` keys are keccak-derived, so pools cannot be enumerated from storage. The three events identify candidate pools, whose slot keys we then compute and match. Event *amounts* are not used — fees, hooks, rates and rounding are only reflected in the final storage write.
- **`!call.state_reverted`** filter on Vault calls.
- **Last-write-per-transaction dedup** per (pool, token): a yield-fee sync can write the same slot before the swap within one transaction, so the highest ordinal wins.

#### Verification

Static checks:

- `cargo clippy --all-targets`: clean. 9 unit tests pass, covering slot-key derivation, packed-balance decoding (low 128 bits), last-write-wins, and non-matching slot keys.
- wasm release build with `--locked` from the package directory (toolchain pinned at 1.96.0); `substreams info` resolves all four manifests (ethereum / arbitrum / base / gnosis).

Fresh full sync from block 21332121 (Vault deployment), followed by the `vm:balancer_v3` integration test:

| | |
|---|---|
| Blocks | 25643926–25644025 (100) |
| Pools | 8, including both reCLAMM pools `0x3aa14db7…` and `0x1bb7928d…` |
| Simulations | 1600 |
| `get_limits` | 1600 success / 0 failure |
| `get_amount_out` | 1600 success / 0 failure |
| Reverts / setup failures | 0 / 0 |
| Execution slippage | avg 0.0000%, min 0.0000%, max 0.0000% (1600 samples) |

Independent on-chain reconciliation, taken at the extractor head block 25643967 as a point-in-time snapshot of the indexer tables and compared against `eth_call` at the same block:

| Data | Source of truth | Compared | Mismatches |
|---|---|---|---|
| Pool token balances | `Vault.getPoolTokenInfo(pool)` → `balancesRaw` | 401 balances across 174 pools | **0** |
| Vault account balances | `Vault.getReservesOf(token)` | 169 tokens | **0** |

Every value matches to the wei. The account-balance path (`get_vault_reserves`) is untouched by this refactor; it is included so the reserve tracking is covered as well.

Two notes for anyone reproducing this: `component_balance` and `account_balance` are temporal tables, and one block can hold several versions of the same balance — two transactions in the same block share the block timestamp, so the earlier version is stored with `valid_to == valid_from`. The query has to select the version valid at the target block (`valid_from <= ts(B) AND (valid_to IS NULL OR valid_to > ts(B))`), and the snapshot must be taken at the same block the RPC calls use, or busy tokens such as USDC and WETH will appear to differ. Ground truth for reserves is `getReservesOf`, not `token.balanceOf(vault)`, since the latter also counts unsettled transfers.

The companion fixes this depended on, #1234 (simulation) and #1232 (indexer), are both merged.

protocols/substreams/ethereum-balancer-v3/arbitrum-balancer-v3.yaml
zach030: Already updated the version to 0.5.0 for all chains.

protocols/substreams/ethereum-balancer-v3/integration_test_base_balancer_v3.tycho.yaml
zach030: This test just covers component creation and static attributes, simulation is covered by other tests.

protocols/substreams/ethereum-balancer-v3/src/modules.rs
die-herdplatte: This can be dropped now

die-herdplatte (approved):
Approved, with the smallest comment

protocols/substreams/ethereum-balancer-v3/src/modules.rs
zach030: nice catch! Already removed `mark_component_as_updated` and `update_maker`.

PR #1053, opened 2026-05-28
feat: add LunarBase protocol integration

ThomasAqu1nas:


protocols/substreams/base-lunarbase/src/modules/1_map_protocol_components.rs
```diff
@@ -0,0 +1,37 @@
...
+    let config = Config::parse(&params)?;
+    if config.bootstrap_block != Some(block.number) {
+        return Ok(tycho::BlockTransactionProtocolComponents::default());
+    }
+
+    let component = to_tycho_protocol_component(lunarbase::protocol_component(
+        config.pool,
+        config.token_x,
+        config.token_y,
+    ));
```
zach030: Does LunarBase only have one pool for now? Any plans to add more later? If new pools are added, can we discover them through events?

protocols/substreams/base-lunarbase/src/modules/2_store_protocol_components.rs
```diff
@@ -0,0 +1,18 @@
...
+pub fn store_protocol_components(
+    components: tycho::BlockTransactionProtocolComponents,
+    store: StoreSetIfNotExistsProto<tycho::ProtocolComponent>,
+) {
+    for tx_components in components.tx_components {
+        for component in tx_components.components {
+            store.set_if_not_exists(0, component_key(&component.id), &component);
+        }
+    }
+}
```
zach030, line 14: This store isn’t used right now. I saw that `map_protocol_changes` checks whether the pair exists, I think it should read from this store for that check.

protocols/substreams/base-lunarbase/src/modules/3_map_protocol_changes.rs
```diff
@@ -0,0 +1,117 @@
...
+        new_components
+            .tx_components
+            .iter()
+            .flat_map(|tx| tx.components.iter())
+            .any(|component| component.id == component_id);
+
+    let known_components = component_is_known
+        .then(|| component_id.clone())
+        .into_iter()
+        .collect::<Vec<_>>();
```
zach030: Do you mean there could be multiple components here?

protocols/substreams/base-lunarbase/src/modules/1_map_protocol_components.rs
ThomasAqu1nas: currently, yes

protocols/substreams/base-lunarbase/src/tycho_mapper.rs
```diff
@@ -0,0 +1,109 @@
...
+    tycho::ProtocolComponent::new(&component.id)
+        .with_tokens(&component.tokens)
+        .with_contracts(&component.contract_addresses)
+        .with_attributes(
+            &component
+                .static_attributes
+                .into_iter()
+                .sorted_unstable_by(|(left, _), (right, _)| left.cmp(right))
+                .collect::<Vec<_>>(),
+        )
```
zach030: Why do we need to add the `static_attribute` into `attributes`? In Tycho, `static_attributes` are immutable, while `attributes` are mutable, so it’s better to keep them separate.

protocols/substreams/base-lunarbase/src/lunarbase/events.rs
```diff
@@ -0,0 +1,101 @@
...
+    pub const CONCENTRATION_K_SET: &str =
+        "0xcf34ec77e4a73dc1b2fdbb6eaec360819374b6412a8bf8096f91c4fdb76db3a8";
+    pub const WHITELIST_SET: &str =
+        "0x0aa5ec5ffdc7f6f9c4d0dded489d7450297155cb2f71cb771e02427f7dff4f51";
+    pub const BLACKLIST_FEE_MULTIPLIER_SET: &str =
+        "0xa15057886e6ebcdf47294bcb091d686031124d1041cafe00740e93667bacd186";
+    pub const SYNC: &str = "0x99e93fd94a51b80d7dd7ec3f69c4f09a43e7523f5a45ca09b88a178d9daaed1e";
+    pub const PAUSED: &str = "0x62e78cea01bee320cd4e420270b5ea74000d11b0c9f74754ebdbfc544b05a258";
+    pub const UNPAUSED: &str = "0x5db9ee0a495bf2e6ff9c91a7834c1ba4fdd244a5e8aa4e537bd38aeae4b073aa";
+}
```
zach030: Can we use the ABI-generated Rust code to distinguish the events instead?

protocols/substreams/base-lunarbase/src/lunarbase/indexed.rs
```diff
@@ -0,0 +1,173 @@
...
+        entry.deleted_attributes.insert(deleted);
+    }
+
+    for (name, value) in delta.updated_attributes {
+        entry.deleted_attributes.remove(&name);
+        entry
+            .updated_attributes
+            .insert(name, value);
+    }
+}
```
zach030: I see `deleted_attributes` is always an empty hashmap, so do we still need this merge here?

crates/tycho-simulation/src/evm/protocol/lunarbase/decoder.rs
```diff
@@ -0,0 +1,136 @@
...
+    insert_u256(&mut attrs, attrs::BLACKLIST_FEE_MULTIPLIER, state.blacklist_fee_multiplier);
+    insert_bool(&mut attrs, attrs::EXECUTOR_WHITELISTED, state.executor_whitelisted);
+    attrs
+}
+
+pub fn decode_state(attrs: &AttributeMap) -> Result<LunarBaseState, StateDecodeError> {
+    Ok(LunarBaseState {
+        pool: require_address(attrs, attrs::POOL)?,
+        token_x: require_address(attrs, attrs::TOKEN_X)?,
+        token_y: require_address(attrs, attrs::TOKEN_Y)?,
```
zach030: These three attributes are static_attributes, so they won’t be readable here.

crates/tycho-simulation/src/evm/protocol/lunarbase/decoder.rs
```diff
@@ -0,0 +1,136 @@
...
+        fee_ask_x24: require_u32(attrs, attrs::FEE_ASK_X24)?,
+        fee_bid_x24: require_u32(attrs, attrs::FEE_BID_X24)?,
+        latest_update_block: require_u64(attrs, attrs::LATEST_UPDATE_BLOCK)?,
+        reserve_x: require_u128(attrs, attrs::RESERVE_X)?,
+        reserve_y: require_u128(attrs, attrs::RESERVE_Y)?,
+        concentration_k: require_u32(attrs, attrs::CONCENTRATION_K)?,
+        block_delay: require_u64(attrs, attrs::BLOCK_DELAY)?,
+        paused: require_bool(attrs, attrs::PAUSED)?,
+        blacklist_fee_multiplier: require_u256(attrs, attrs::BLACKLIST_FEE_MULTIPLIER)?,
+        executor_whitelisted: require_bool(attrs, attrs::EXECUTOR_WHITELISTED)?,
```
zach030: These attributes don’t have defaults in the substreams, so they may not exist if the events never fire. Better to set defaults for them in the substreams.

crates/tycho-simulation/src/evm/protocol/lunarbase/mod.rs
```diff
@@ -0,0 +1,444 @@
...
+                .map(|(key, value)| (key, value.to_vec()))
+                .collect(),
+            deleted_attributes: delta
+                .deleted_attributes
+                .into_iter()
+                .collect(),
+        };
+        apply_delta(&mut self.state, &state_delta)
+            .map_err(|err| TransitionError::DecodeError(format!("{err:?}")))
+    }
```
zach030: `head_block` isn’t being updated.

crates/tycho-simulation/src/evm/protocol/lunarbase/mod.rs
```diff
@@ -0,0 +1,444 @@
...
+        buy_token: Bytes,
+    ) -> Result<(BigUint, BigUint), SimulationError> {
+        let sell = address_from_bytes(sell_token.as_ref())?;
+        let buy = address_from_bytes(buy_token.as_ref())?;
+        if sell == self.state.token_x && buy == self.state.token_y {
+            return Ok((BigUint::ZERO, BigUint::from(self.state.reserve_x)));
+        }
+        if sell == self.state.token_y && buy == self.state.token_x {
+            return Ok((BigUint::ZERO, BigUint::from(self.state.reserve_y)));
+        }
```
zach030: `get_limits` should return the max input amount for the `sell_token` and the max output amount for the `buy_token`. Right now, the first value is `0`, that doesn’t match the expected definition.

crates/tycho-simulation/src/evm/protocol/lunarbase/mod.rs
```diff
@@ -0,0 +1,444 @@
...
+        match params.swap_constraint() {
+            SwapConstraint::TradeLimitPrice { .. } | SwapConstraint::PoolTargetPrice { .. } => {
+                Err(SimulationError::InvalidInput(
+                    "LunarBase native simulator only supports exact-input get_amount_out"
+                        .to_owned(),
+                    None,
+                ))
+            }
+        }
+    }
```
zach030: This function doesn’t seem to be implemented yet.

protocols/substreams/base-lunarbase/src/modules/1_map_protocol_components.rs
ThomasAqu1nas: added support for multiple explicitly configured pools via the `pools=...` parameter, so each configured pool becomes a separate Tycho component. If an on-chain factory is added later, discovery can be moved from config to factory events.

protocols/substreams/base-lunarbase/src/modules/2_store_protocol_components.rs
ThomasAqu1nas: `map_protocol_changes` now reads from `store_protocol_components` when checking known components.

protocols/substreams/base-lunarbase/src/tycho_mapper.rs
ThomasAqu1nas: removed `with_attributes(...)` call here and stopped duplicating pool/token data into component static attributes

protocols/substreams/base-lunarbase/src/lunarbase/events.rs
ThomasAqu1nas: now using ABI-generated bindings instead of manual topic constants

protocols/substreams/base-lunarbase/src/lunarbase/indexed.rs
ThomasAqu1nas: the merge now only applies updated attributes

crates/tycho-simulation/src/evm/protocol/lunarbase/mod.rs
ThomasAqu1nas: now delegates to shared `query_pool_swap` helper

crates/tycho-simulation/src/evm/protocol/lunarbase/mod.rs
ThomasAqu1nas: now fully quote-derived

crates/tycho-simulation/src/evm/protocol/lunarbase/mod.rs
ThomasAqu1nas: `delta_transition` now reads `block_number`, filters it out of protocol attrs, and updates `head_block`

protocols/substreams/base-lunarbase/src/modules/3_map_protocol_changes.rs
```diff
@@ -0,0 +1,134 @@
...
+        component_store
+            .get_last(component_key(&component_id))
+            .is_some() ||
+        new_components
+            .tx_components
+            .iter()
+            .flat_map(|tx| tx.components.iter())
+            .any(|component| component.id == component_id);
+    known.then_some(component_id)
+}
```
zach030, line 103: can we just check `component_store`? New components should already be written to the store earlier.

protocols/substreams/base-lunarbase/src/modules/3_map_protocol_changes.rs
```diff
@@ -0,0 +1,134 @@
...
+                pool.bootstrap_state,
+            );
+        }
+    }
+
+    for tx in block.transactions() {
+        for (log, _) in tx.logs_with_calls() {
+            let Some(pool) = pool_by_address(&config, &log.address) else {
+                continue;
+            };
```
zach030, line 54: Should we filter out pools whose components haven’t been created yet here?

protocols/substreams/base-lunarbase/src/lunarbase/events.rs
```diff
@@ -0,0 +1,221 @@
...
+            insert_u64(&mut updated_attributes, attrs::BLOCK_DELAY, *block_delay);
+        }
+        LunarBaseEvent::ConcentrationKSet { concentration_k } => {
+            insert_u32(&mut updated_attributes, attrs::CONCENTRATION_K, *concentration_k);
+        }
+        LunarBaseEvent::WhitelistSet { account, whitelisted } => {
+            if *account == context.tycho_router {
+                insert_bool(&mut updated_attributes, attrs::SWAP_CALLER_WHITELISTED, *whitelisted);
+            }
+        }
```
zach030: I noticed the `tycho_router` config is still empty. How are you planning to handle the whitelist? Will all pools be whitelisted for `tycho_router`, or only some?

If all pools are whitelisted, can we remove this attribute? Then In `tycho-simulation`, pools are whitelisted by default.

protocols/substreams/base-lunarbase/src/lunarbase/indexed.rs
```diff
@@ -0,0 +1,215 @@
...
+    pub hash: [u8; 32],
+    pub from: Address,
+    pub to: Address,
+    pub index: u64,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct BlockChanges {
+    pub transactions: Vec<TransactionChanges>,
+}
```
zach030: I’m not really a fan of adding these extra structs. They already exist in `tycho-substreams`, and if we define new ones, we’ll need extra conversion logic in `tycho-mapper` too.

protocols/substreams/base-lunarbase/src/lunarbase/events.rs
ThomasAqu1nas: our partners are whitelisted by default, so I will remove this attribute

ThomasAqu1nas:
@zach030, pls check latest commit, fixed

crates/tycho-simulation/src/evm/protocol/lunarbase/state.rs
```diff
@@ -0,0 +1,475 @@
...
+    loop {
+        let quote = quote_exact_in(
+            state,
+            QuoteRequest {
+                token_in,
+                token_out,
+                amount_in: biguint_to_u256(&amount_in)?,
+                block_number: state.latest_update_block,
+            },
+        );
```
zach030: Here we use `latest_update_block`, but `get_amount_out` uses `head_block`. Could that cause issues?

crates/tycho-simulation/src/evm/protocol/lunarbase/state.rs
```diff
@@ -0,0 +1,475 @@
...
+    pub anchor_price_x96: u128,
+    pub fee_ask_x24: u32,
+    pub fee_bid_x24: u32,
+    pub latest_update_block: u64,
+    pub reserve_x: u128,
+    pub reserve_y: u128,
+    pub concentration_k: u32,
+    pub block_delay: u64,
+    pub paused: bool,
+    pub blacklist_fee_multiplier: U256,
```
zach030: We don’t use `blacklist_fee_multiplier` anymore, right? If so, we can remove it and the related logic from both `tycho-simulation` and the substreams.

crates/tycho-simulation/src/evm/protocol/lunarbase/decoder.rs
```diff
@@ -0,0 +1,390 @@
...
+    pub const PAUSED: &str = "paused";
+    pub const BLACKLIST_FEE_MULTIPLIER: &str = "blacklist_fee_multiplier";
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum AttributeError {
+    Missing(&'static str),
+    InvalidLength { name: &'static str, expected: usize, actual: usize },
+    IntegerOverflow(&'static str),
+}
```
zach030: no need to define a separate `AttributeError`. We already have the `InvalidSnapshotError` enum.

crates/tycho-simulation/src/evm/protocol/lunarbase/state.rs
```diff
@@ -0,0 +1,475 @@
...
+        U256::from(1u64)
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub struct LunarBaseTychoState {
+    pub state: LunarBaseState,
+    pub head_block: u64,
+}
+
```
zach030, line 130: I don’t think we need two state structs here. Just keeping `LunarBaseTychoState` with all fields should be enough.

crates/tycho-simulation/src/evm/protocol/lunarbase/decoder.rs
```diff
@@ -0,0 +1,390 @@
...
+        .to_be_bytes_vec()
+        .iter()
+        .rev()
+        .take(32)
+        .enumerate()
+        .for_each(|(idx, byte)| {
+            out[31 - idx] = *byte;
+        });
+    attrs.insert(name.to_owned(), out);
+}
```
zach030: I don’t think these need separate functions. They’re only used once, so we can just initialize them inline when creating the hashmap.

crates/tycho-simulation/src/evm/protocol/lunarbase/decoder.rs
```diff
@@ -0,0 +1,390 @@
...
+    }
+    let mut out = [0u8; 16];
+    out[16 - value.len()..].copy_from_slice(value);
+    Ok(u128::from_be_bytes(out))
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct StateDelta {
+    pub updated_attributes: AttributeMap,
+}
```
zach030: I don’t think this struct is needed either.

ThomasAqu1nas:
I checked the failed RPC-dependent CI jobs. The logs show ETH_RPC_URL, BASE_RPC_URL, RPC_URL and the other RPC env vars are all empty. It looks like repository secrets are not available for this fork PR rather than a LunarBase-specific failure. Could you rerun these checks in a trusted context, or confirm if RPC-dependent checks are expected to fail on external fork PRs?

zach030:
> I checked the failed RPC-dependent CI jobs. The logs show ETH_RPC_URL, BASE_RPC_URL, RPC_URL and the other RPC env vars are all empty. It looks like repository secrets are not available for this fork PR rather than a LunarBase-specific failure. Could you rerun these checks in a trusted context, or confirm if RPC-dependent checks are expected to fail on external fork PRs?

Thanks, I’ll review the code. This looks like a CI setup issue on our side, we’ll check it.

crates/tycho-execution/contracts/src/executors/LunarBaseExecutor.sol
```diff
@@ -0,0 +1,126 @@
...
+    function swap(uint256 amountIn, bytes calldata data, address receiver)
+        external
+        payable
+    {
+        (address pool, address tokenIn, address tokenOut,) = _decodeData(data);
+
+        if (tokenIn == ETH_ADDRESS) {
+            if (msg.value != amountIn) {
+                revert LunarBaseExecutor__MsgValueMismatch();
+            }
```
zizou0x: This is too restrictive and could wrongly error (for example if a user does a split swap). 

We can remove it, the router is responsible for accounting.

protocols/substreams/base-lunarbase/src/modules/3_map_protocol_changes.rs
```diff
@@ -0,0 +1,117 @@
...
+    block: eth::v2::Block,
+    new_components: tycho::BlockTransactionProtocolComponents,
+    component_store: StoreGetProto<tycho::ProtocolComponent>,
+) -> Result<tycho::BlockChanges> {
+    let config = Config::parse(&params)?;
+    let mut known_components = config
+        .pools
+        .iter()
+        .filter_map(|pool| known_pool_component(pool, &component_store))
+        .collect::<HashSet<_>>();
```
zizou0x, line 28: I wonder if we need this, wouldn't it be simple to entirely rely on `component_store.get_last(id)` instead? We could probably entirely get rid of the params input in this map. Also it's a good step toward having automated detection (if you ever deploy a pool factory and we stop relying on static config for pool detection).

crates/tycho-simulation/src/evm/protocol/lunarbase/state.rs
```diff
@@ -0,0 +1,377 @@
...
+) -> Result<f64, SimulationError> {
+    if reserve_in == 0 || reserve_out == 0 {
+        return Err(SimulationError::RecoverableError("zero LunarBase reserve".to_owned()));
+    }
+    let decimals_adjustment = 10f64.powi(token_in.decimals as i32 - token_out.decimals as i32);
+    Ok((reserve_out as f64 / reserve_in as f64) * decimals_adjustment)
+}
+
+fn soft_limit(reserve_in: u128) -> BigUint {
+    BigUint::from(reserve_in) * 2162u32 / 1000u32
```
zizou0x, line 277: What are these numbers? Could you document what's happening here please?

protocols/substreams/base-lunarbase/src/modules/3_map_protocol_changes.rs
ThomasAqu1nas: nice catch, will explore

zizou0x (approved):
LGTM! Thanks for the integration!

protocols/testing/src/test_runner.rs
```diff
@@ -104,6 +105,12 @@ pub struct TestRunner {
     reuse_last_sync: bool,
 }
 
+struct SnapshotDecodeOptions {
+    adapter_contract_path: Option<PathBuf>,
+    vm_simulation_traces: bool,
+    component_filter: Option<ComponentFilter>,
+}
+
```
zach030: Not sure why we need this, and why we also need to change the filter condition in `register_protocol`.

protocols/testing/src/test_runner.rs
```diff
@@ -1323,6 +1342,12 @@ impl TestRunner {
 
         info!("Batch execution complete: {} successes, {} failures", success_count, failure_count);
 
+        if failure_count > 0 {
+            return Err(miette!(
+                "Execution validation failed for {protocol_system}: {success_count} successes, {failure_count} failures"
+            ));
+        }
+
```
zach030: We probably don’t need this.

protocols/testing/src/test_runner.rs
```diff
@@ -624,6 +637,12 @@ impl TestRunner {
             &test.expected_components,
             &config.protocol_system,
         )?;
+        if execution_data.is_empty() && should_run_execution(&test.expected_components) {
+            return Err(miette!(
+                "No execution data generated for {}; simulation decoded no executable state",
+                config.protocol_system
+            ));
+        }
```
zach030: We probably don’t need this. I remember we already have a filter for components that shouldn’t execute.

protocols/testing/src/execution.rs
```diff
@@ -116,5 +119,24 @@ pub fn create_router_overwrites_data(
...
+            })?;
+        let bytecode_hex = if let Some(stripped) = bytecode_str.strip_prefix("0x") {
+            stripped
+        } else {
+            bytecode_str
+        };
+        hex::decode(bytecode_hex)
+            .into_diagnostic()
+            .wrap_err("Failed to decode fee calculator bytecode from hex")?
+    };
```
zach030: What is this `fee_calculator` contract used for?

crates/tycho-test/src/execution/mod.rs
```diff
@@ -137,6 +137,15 @@ pub async fn simulate_swap_transaction(
...
             }
+            if info.protocol_system == "lunarbase" {
+                let pool_address = Address::from_str(&info.component_id)
+                    .map_err(|e| (miette!("Invalid LunarBase component id: {e}"), None, None))?;
+                let swap_caller = bytes_to_address(&to_address)
+                    .map_err(|e| (miette!("Invalid Tycho router address: {e}"), None, None))?;
+                let lunarbase_overwrites = setup_lunarbase_overwrites(pool_address, swap_caller)
+                    .map_err(|e| (e, None, None))?;
+                state_overwrites.extend(lunarbase_overwrites);
+            }
```
zach030: Since you’ve already whitelisted the tycho router onchain, it looks like we probably don’t need the override anymore.

zach030 (commented):
Hey! I missed part of this in the previous review, mainly the testing related code. I think some of those changes may not be needed, could you check again?

protocols/testing/src/execution.rs
ThomasAqu1nas: @zach030
It is needed by the overwritten TychoRouter runtime used in execution tests.

The router reads its _feeCalculator dependency during execution, so the test harness needs to place fee calculator bytecode at the deterministic address configured in the router overwrite. This is not LunarBase-specific; it lets the router accounting path run in tests without depending on production fee configuration. The bytecode here is a zero-fee calculator, so it should not affect protocol quote behavior.

pls ping me if I'm doing something wrong

ThomasAqu1nas:
@zach030, you can check last commit now

protocols/testing/src/state_registry.rs
```diff
@@ -23,57 +23,63 @@ pub fn register_protocol(
     decoder_context: DecoderContext,
 ) -> miette::Result<ProtocolStreamBuilder> {
     let tvl = chain.default_tvl_threshold(TvlThresholdTier::Medium);
-    let tvl_filter = ComponentFilter::with_tvl_range(tvl, tvl);
+    let component_filter = ComponentFilter::with_tvl_range(tvl, tvl);
```
zach030: Can you rename this back to `tvl_filter`?

protocols/testing/src/test_runner.rs
```diff
@@ -867,7 +864,7 @@ impl TestRunner {
         let protocol_stream_builder =
             ProtocolStreamBuilder::new("", self.chain).skip_state_decode_failures(true);
 
-        let mut decoder_context = DecoderContext::new().vm_traces(vm_simulation_traces);
+        let mut decoder_context = DecoderContext::new().vm_traces(self.vm_simulation_traces);
```
zach030: The changes in `test_runner.rs` should be reverted as well.

protocols/testing/src/execution.rs
zach030: Does the `fee_calculator` contract affect your executor right now? If not, can we revert this for now? It’s currently left empty, and we may update it later. I’ll also check with the team on this.

zach030 (commented):
Thanks for the update! Could you take a look at this code? If it’s not needed, we can revert the changes here.

protocols/testing/src/test_runner.rs
ThomasAqu1nas: Agree, I’ll revert the test_runner/state_registry changes as well. They were only part of the execution-test harness adjustments and are not required for the LunarBase integration itself.

crates/tycho-test/src/execution/encoding.rs
```diff
@@ -37,8 +37,6 @@ const GAS_LIMIT: u64 = 100_000_000;
 // 1_000 native tokens (10^21 wei): covers 100M gas at up to ~10_000 gwei
```
zach030, line 37: The three files under `tycho-test/src/execution` shouldn’t need to be changed

protocols/testing/fixtures/FeeCalculator.runtime.json
```diff
@@ -0,0 +1,3 @@
+{
```
zach030: This is unnecessary, could you remove it?

protocols/testing/fixtures/TychoRouter.runtime.json
```diff
@@ -1 +1,3 @@
...1981018352826115c7565b51915af4612354611b68565b905f7f098a7a3b47801589e8cdf9ec791b93ad44273246946c32ef1fc4dbe45390c80e5d156123955760208151918180820193849201010312610170575190565b8051156123b8576101989060405191829162461bcd60e51b8352600483016121b0565b506101986040517f457865637574696f6e206661696c6564000000000000000000000000000000006020820152601081526101d86030826115c7565b6001600160a01b037f000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc21690813b15610170575f916024839260405194859384927f2e1a7d4d00000000000000000000000000000000000000000000000000000000845260048401525af180156110855761224f5750565b9190820391821161211657565b6001600160a01b03806124936124a397989495969886611c94565b94169116146124dc575b5061246b565b8181036124ae575050565b7ff7bf5832000000000000000000000000000000000000000000000000000000005f5260045260245260445ffd5b6124e6919261246b565b905f61249d565b906124f7826115fd565b61250460405191826115c7565b8281528092612515601f19916115fd565b0190602036910137565b9190820180921161211657565b9091821560011461255357813560f01c906002830193600283809501019260011991030190565b5f92508291508190819056fea26469706673582212209bbb904097193ebac430ded80f0a89f7c6de1ea7fdd33366abebcb5092d9f65264736f6c634300081a0033"}
```
zach030, line 1: Could you revert this change?

protocols/testing/src/execution.rs
```diff
@@ -116,5 +118,5 @@ pub fn create_router_overwrites_data(
 
     let executor_bytecode = load_executor_bytecode(protocol_system)?;
 
-    Ok(RouterOverwritesData { router_bytecode, executor_bytecode, fee_calculator_bytecode: vec![] })
+    Ok(RouterOverwritesData { router_bytecode, executor_bytecode })
```
zach030: plz revert this change

protocols/testing/src/test_runner.rs
```diff
@@ -1116,12 +1115,12 @@ impl TestRunner {
...
                     }
 
+                    let chain_model = self.chain;
                     let executors_json = json!({
-                        "ethereum": {
+                        (chain_model.to_string()): {
                             (protocol_system): EXECUTOR_ADDRESS
                         }
                     });
-                    let chain_model = self.chain;
```
zach030, line 1124: plz revert this change

ThomasAqu1nas:
Gimme few hours, i'll continue bit later

zach030 (approved):
Thanks! LGTM!

PR #1058, opened 2026-06-01
fix(router): Reset deltas to allow batched calls

tamaralipows:
Found thanks to Consensys Diligence's LLM-based analysis: https://hackmd.io/@DDxm0FVhSNCODtO8HMANSQ/SkI1XB6yze#finding-6

Because EIP-1153 transient storage is scoped to the full transaction (not the call), this stale state persists after _finalizeBalances returns and the nonReentrant lock is released. Any subsequent call to a TychoRouter swap function within the same transaction — for example from a smart-contract aggregator or MEV bot that batches two vault swaps in one tx — will encounter the residual count and revert.

### DIFF
--- a/crates/tycho-execution/contracts/test/TychoRouterVault.t.sol
+++ b/crates/tycho-execution/contracts/test/TychoRouterVault.t.sol
@@ -26,6 +26,48 @@ import "./TychoRouterTestSetup.sol";
 import {NativeWrapExecutor} from "../src/executors/NativeWrapExecutor.sol";
 
 /**
+ * @title VaultSwapBatcher
+ * @notice Helper that batches two vault swaps in one transaction,
+ *         exercising same-tx transient storage reuse.
+ */
+contract VaultSwapBatcher {
+    TychoRouter public router;
+
+    constructor(address router_) {
+        router = TychoRouter(payable(router_));
+    }
+
+    function depositAndBatchSwaps(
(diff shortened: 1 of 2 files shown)


crates/tycho-execution/contracts/test/TychoRouterVault.t.sol
```diff
@@ -25,6 +25,48 @@ import {
...
+
+    function depositAndBatchSwaps(
+        address tokenIn,
+        uint256 depositAmount,
+        uint256 amountIn1,
+        uint256 amountIn2,
+        address tokenOut,
+        bytes calldata swap1,
+        bytes calldata swap2
+    ) external {
```
louise-poole, line 48: I'm not sure I understand what we want here... we're only batching 2 swaps right? This still feels limited and not much better than calling single swap each time.

louise-poole (commented):
One quick question to help me understand our goal here 👀

crates/tycho-execution/contracts/test/TychoRouterVault.t.sol
louise-poole: Ooooh! This is just a test. Lol, cool cool!

louise-poole (approved):

dianacarvalho1 (approved):
Thank you @tamaralipows 🙏🏼

PR #1064, opened 2026-06-03
fix: move schema to inside rule scope

brunoguerios:
Even Better TOML config file had `schema` config set to all toml files, when it should only apply to `Cargo.toml` files. Moving it to under `rule` fixes that.


zizou0x (approved):
LGTM! Thanks!

tvinagre (approved):

PR #1067, opened 2026-06-04
feat: add BlockStepController to ProtocolStreamBuilder for gated testing

kayibal:
## Summary

Adds `BlockStepController`, an opt-in handle that lets callers control exactly when each block is released for decoding in `ProtocolStreamBuilder`. The primary use case is deterministic testing where you need to inspect or coordinate state before a block is processed.

Enable it via the new `with_step_controller()` builder method, which returns the builder alongside the controller. Zero overhead in production — the gating task is never spawned if the method is not called.

**Controller API:**
- `trigger_next_block()` — releases the buffered block into the decode pipeline
- `peek_next_block()` — async, waits until a block is buffered and returns it without consuming it
- `try_peek_next_block()` — sync, returns the buffered block immediately or `None`

**Drop behaviour:** dropping the controller ungates the stream — the buffered block is forwarded and subsequent blocks pass through without gating, so the stream runs to its natural end.

Works with both `build()` and `build_with_pending()`. The `advance_tx` send to the pending processor remains in the shared `.then()` closure, so both paths are structurally identical.

Includes three unit tests and one `#[ignore]`d network test (requires `TYCHO_AUTH_TOKEN`).

### DIFF
--- a/crates/tycho-simulation/src/evm/stream.rs
+++ b/crates/tycho-simulation/src/evm/stream.rs
@@ -115,7 +115,7 @@ use tracing::{debug, error, warn};
 use tycho_client::{
     feed::{
         component_tracker::ComponentFilter, synchronizer::ComponentWithState, BlockHeader,
-        FeedMessage, SynchronizerState,
+        BlockSynchronizerError, FeedMessage, SynchronizerState,
     },
     stream::{RetryConfiguration, StreamError, TychoStreamBuilder},
 };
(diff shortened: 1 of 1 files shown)


crates/tycho-simulation/src/evm/stream.rs
```diff
@@ -708,4 +870,67 @@ mod tests {
...
+
+        assert!(!peeked.sync_states.is_empty(), "peeked block should carry sync states");
+
+        // Release the block.
+        controller
+            .trigger_next_block()
+            .expect("trigger_next_block failed");
+
+        // Stream should now yield the decoded update within one block time.
+        let update = tokio::time::timeout(Duration::from_secs(30), stream.next())
```
zizou0x, line 936: Shouldn't we make sure that stream is empty before calling `controller.trigger_next_block`

crates/tycho-simulation/src/evm/stream.rs
```diff
@@ -708,4 +870,67 @@ mod tests {
...
+        use std::{env, time::Duration};
+
+        use crate::evm::protocol::uniswap_v2::state::UniswapV2State;
+
+        let auth = env::var("TYCHO_AUTH_TOKEN").expect("TYCHO_AUTH_TOKEN must be set");
+
+        // Track a single well-known pool to minimise startup latency.
+        let usdc_weth_v2 = "0xb4e16d0168e52d35cacd2c6185b44281ec28c9dc".to_string();
+        let (builder, controller) =
+            ProtocolStreamBuilder::new("tycho-beta.propellerheads.xyz", Chain::Ethereum)
```
zizou0x, line 900: I'd prefer to use dev

zizou0x (approved):

crates/tycho-simulation/src/evm/stream.rs
kayibal: Why? It's more likely to be flaky, this test is ignore by default so it will only run if you explicitly run it and also just pulls a single pool.

PR #1068, opened 2026-06-04
docs: clarify swap ordering requirement for split swaps

tamaralipows:
Add validation rule and warning about swap ordering when paths of different lengths converge on the same intermediate token. The router processes swaps sequentially and applies split fractions to the balance at execution time, so all upstream swaps producing a token must appear before any swap consuming it.


dianacarvalho1 (approved):
🙏🏼

PR #1072, opened 2026-06-08
fix: revert ChangeType::Creation attributes without DB lookup

louise-poole:
Chain reorgs on Base crashed with "Could not find Component" when a partial block was reverted that contained an attribute first set with ChangeType::Creation (e.g. a new Uniswap V4 tick slot). These attributes have no prior state anywhere — not in the reorg buffer (reverted range) and not in the DB (non-finalized component) — so looking up their previous value is both impossible and unnecessary.

Track ChangeType::Creation attributes in a new `created_attributes` field on ProtocolComponentStateDelta, populated during protobuf deserialisation. In handle_revert, filter these out of the buffer/DB lookup entirely and emit ChangeType::Deletion directly instead.


louise-poole (commented):
Looks good! I would approve if I could ✅

zizou0x (approved):
LGTM!

PR #1073, opened 2026-06-08
fix(client): improve reconnect handling

louise-poole:
- doubled the WS buffer size (flashblocks and faster chains fill this buffer too quickly)
- loop for the ensure connection check to avoid a brief reconnect failing the check when it should keep retrying
- unblock ensure_connection callers if `close()` was called on that connection (used to just hang for those callers)


crates/tycho-client/src/deltas.rs
```diff
@@ -1122,17 +1127,13 @@ mod tests {
...
-            "compression": compression,
-            "partial_blocks": false
+        serde_json::to_string(&Command::Subscribe {
+            extractor_id: dto::ExtractorIdentity::new(dto::Chain::Ethereum, "vm:ambient"),
+            include_state: true,
+            compression,
+            partial_blocks: false,
         })
-        .to_string()
+        .unwrap()
```
dianacarvalho1: unwrap? 😱

dianacarvalho1 (approved):
Thank you @louise-poole ! I'm not super familiar with this code but it looks good to me! Only one small comment

crates/tycho-client/src/deltas.rs
louise-poole: I am mortified! 😱

louise-poole: Oh lol - yeah it was a test helper so unwrap is fine. But I removed it anyway 👍

PR #1074, opened 2026-06-08
feat: support block env overrides in VM simulation

zach030:


crates/tycho-simulation/src/evm/protocol/vm/decoder.rs
```diff
@@ -163,6 +164,36 @@ impl TryFromWithBlock<ComponentWithState, BlockHeader> for EVMPoolState<PreCache
...
         }
+        let block_overrides = if let (Some(block_number), Some(block_timestamp)) = (
+            snapshot
+                .state
+                .attributes
+                .get("override_block_number"),
+            snapshot
+                .state
+                .attributes
+                .get("override_block_timestamp"),
```
zizou0x: I don't think we should assume any attribute name for now. We're not sure yet how fermi is gonna use it yet, I think this needs to be done in the next PR. Let's just expose it on the interface for don't assume it comes from protocol state.

crates/tycho-simulation/src/evm/protocol/vm/state.rs
```diff
@@ -711,6 +743,34 @@ where
         tokens: &HashMap<Bytes, Token>,
         balances: &Balances,
     ) -> Result<(), TransitionError> {
+        if let (Some(block_number), Some(block_timestamp)) = (
+            delta
+                .updated_attributes
+                .get("override_block_number"),
+            delta
+                .updated_attributes
+                .get("override_block_timestamp"),
```
zizou0x: Same here, let's not assume this for now

crates/tycho-simulation/src/evm/decoder.rs
```diff
@@ -999,14 +999,13 @@ where
...
-    /// Add block information (number and timestamp) to a ProtocolStateDelta
+    /// Add current block information (number and timestamp) to a ProtocolStateDelta.
     fn add_block_info_to_delta(
         mut delta: ProtocolStateDelta,
         block_header_opt: Option<BlockHeader>,
     ) -> ProtocolStateDelta {
         if let Some(header) = block_header_opt {
-            // Add block_number and block_timestamp attributes to ensure pool states
-            // receive current block information during delta_transition
+            // These are current block attributes for native protocols, not VM block env overrides.
```
zizou0x: I find it less clear than it was previously

zizou0x (approved):
LGTM

PR #1075, opened 2026-06-08
feat: add Custom variant to Chain enum to support user-defined chains

brunoguerios:
## Summary

- Adds a `Custom(CustomChainConfig)` variant to `models::Chain` and `dto::Chain`, allowing callers to define chains beyond the hardcoded set without forking the library
- Introduces `CustomChainConfig` — a fully `Copy`-compatible struct holding chain identity (`name`, `chain_id`, `block_time_secs`), native and wrapped native token config (`ChainTokenConfig`), and TVL thresholds (`TvlThresholds`). All fields use stack-allocated types (`ArrayString`, fixed-size byte arrays) to satisfy the `Copy` bound that `Chain` requires
- Adds `ChainAddress` — a `[u8; 32]` + length field for variable-length addresses up to 32 bytes, covering both EVM (20-byte) and Starknet (32-byte felt252)
- Replaces `EnumString`/`Display` strum derives on `Chain` with manual `FromStr` and `Display` impls. Known lowercase names map to their variants as before; `Custom` is never constructed by `FromStr` alone — full config is required. `dto::Chain` delegates both impls to `models::Chain`
- Wires `Custom` into all `Chain` methods: `id()`, `default_tvl_threshold()`, `native_token()`, `wrapped_native_token()`
- Adds `arrayvec` (with `serde` feature) as a direct dependency of `tycho-common`

## Test plan

- [ ]  `cargo clippy --workspace --all-features` passes
- [ ]  Construct a `Chain::Custom` with a known config and verify `Display`, `id()`, `native_token()`, and `default_tvl_threshold()` return the expected values
- [ ]  Confirm `Chain::from_str("custom")` returns `Err` (custom chains require full config, not just a name)

### DIFF
--- a/crates/tycho-client-py/python/tycho_indexer_client/dto.py
+++ b/crates/tycho-client-py/python/tycho_indexer_client/dto.py
@@ -48,6 +48,38 @@ class Chain(str, Enum):
     polygon = "polygon"
 
 
+class ChainTokenConfig(BaseModel):
+    address: str
+    symbol: str
+    decimals: int
+
+
+class TvlThresholds(BaseModel):
+    low: int
+    medium: int
+
+
+class CustomChainConfig(BaseModel):
+    name: str
+    chain_id: int
+    block_time_secs: int
+    native: ChainTokenConfig
+    wrapped_native: ChainTokenConfig
+    default_tvl_thresholds: TvlThresholds
+
+    def __str__(self) -> str:
+        return self.name
+
+    @root_validator(pre=True)
+    def _unwrap_serde_tag(cls, values):
+        # Rust serialises Custom(cfg) as {"custom": {...}} — unwrap the tag.
+        if "custom" in values and len(values) == 1:
+            return values["custom"]
+        return values
+
+
+
+
 class ChangeType(str, Enum):
     update = "Update"
     deletion = "Deletion"
@@ -56,7 +88,7 @@ class ChangeType(str, Enum):
 
 
 class ExtractorIdentity(BaseModel):
-    chain: Chain
+    chain: Union[Chain, CustomChainConfig]
     name: str
 
 
@@ -93,13 +125,13 @@ class Block(BaseModel):
     number: int
     hash: HexBytes
     parent_hash: HexBytes
-    chain: Chain
+    chain: Union[Chain, CustomChainConfig]
     ts: datetime
 
 
 class BlockParam(BaseModel):
     hash: Optional[HexBytes] = None
-    chain: Optional[Chain] = None
+    chain: Optional[Union[Chain, CustomChainConfig]] = None
     number: Optional[int] = None
 
 
@@ -120,7 +152,7 @@ class TokenBalances(BaseModel):
 
 class AccountUpdate(BaseModel):
     address: HexBytes
-    chain: Chain
+    chain: Union[Chain, CustomChainConfig]
     slots: Dict[HexBytes, HexBytes]
     balance: Optional[HexBytes] = None
     code: Optional[HexBytes] = None
@@ -137,7 +169,7 @@ class ProtocolComponent(BaseModel):
     id: str
     protocol_system: str
     protocol_type_name: str
-    chain: Chain
+    chain: Union[Chain, CustomChainConfig]
     tokens: List[HexBytes]
     contract_ids: List[HexBytes]
     static_attributes: Dict[str, HexBytes]
@@ -147,7 +179,7 @@ class ProtocolComponent(BaseModel):
 
 
 class ResponseToken(BaseModel):
-    chain: Chain
+    chain: Union[Chain, CustomChainConfig]
     address: HexBytes = Field(..., example="0xc9f2e6ea1637E499406986ac50ddC92401ce1f58")
     symbol: str = Field(..., example="WETH")
     decimals: int
@@ -158,7 +190,7 @@ class ResponseToken(BaseModel):
 
 class BlockChanges(BaseModel):
     extractor: str
-    chain: Chain
+    chain: Union[Chain, CustomChainConfig]
     block: Block
     finalized_block_height: int
     revert: bool
@@ -186,7 +218,7 @@ class ComponentWithState(BaseModel):
 
 
 class ResponseAccount(BaseModel):
-    chain: Chain
+    chain: Union[Chain, CustomChainConfig]
     address: HexBytes
     title: str
     slots: Dict[HexBytes, HexBytes]
@@ -257,13 +289,13 @@ class PaginationParams(BaseModel):
 
 
 class ProtocolId(BaseModel):
-    chain: Chain
+    chain: Union[Chain, CustomChainConfig]
     id: str
 
 
 class ContractId(BaseModel):
     address: HexBytes
-    chain: Chain
+    chain: Union[Chain, CustomChainConfig]
 
 
 class VersionParams(BaseModel):
@@ -324,7 +356,7 @@ class TokensParams(BaseModel):
 
 
 class ProtocolSystemsParams(BaseModel):
-    chain: Optional[Chain] = None
+    chain: Optional[Union[Chain, CustomChainConfig]] = None
     pagination: Optional[PaginationParams] = None
 
     class Config:
@@ -332,7 +364,7 @@ class ProtocolSystemsParams(BaseModel):
 
 
 class ComponentTvlParams(BaseModel):
-    chain: Optional[Chain] = None
+    chain: Optional[Union[Chain, CustomChainConfig]] = None
     protocol_system: Optional[str] = Field(default=None, alias="protocolSystem")
     component_ids: Optional[List[str]] = Field(default=None)
     pagination: Optional[PaginationParams] = None
@@ -342,7 +374,7 @@ class ComponentTvlParams(BaseModel):
 
 
 class TracedEntryPointParams(BaseModel):
-    chain: Optional[Chain] = None
+    chain: Optional[Union[Chain, CustomChainConfig]] = None
     protocol_system: str
     component_ids: Optional[List[str]] = Field(default=None)
     pagination: Optional[PaginationParams] = None
--- a/crates/tycho-client-py/python/tycho_indexer_client/dto.py
+++ b/crates/tycho-client-py/python/tycho_indexer_client/dto.py
@@ -55,8 +55,8 @@ class ChainTokenConfig(BaseModel):
 
 
 class TvlThresholds(BaseModel):
-    low: int
-    medium: int
+    low: float
+    medium: float
 
 
 class CustomChainConfig(BaseModel):
--- a/crates/tycho-client/src/feed/mod.rs
+++ b/crates/tycho-client/src/feed/mod.rs
@@ -1,5 +1,3 @@
-#![allow(clippy::result_large_err)]
-
 use std::{
     collections::{HashMap, HashSet},
     fmt::{Display, Formatter},
@@ -245,6 +243,7 @@ impl SynchronizerStream {
             rx,
         }
     }
+    #[allow(clippy::result_large_err)]
     async fn try_advance(
         &mut self,
         block_history: &BlockHistory,
@@ -420,6 +419,7 @@ impl SynchronizerStream {
     }
 
     /// Helper method to check if synchronizer should transition to stale based on time elapsed
+    #[allow(clippy::result_large_err)]
     fn check_and_transition_to_stale_if_needed(
         &mut self,
         stale_threshold: std::time::Duration,
@@ -458,6 +458,7 @@ impl SynchronizerStream {
     /// - Next expected block -> Ready state
     /// - Latest/Delayed block -> Either Delayed or Stale (if >60s since last update)
     /// - Advanced block -> Advanced state (block ahead of expected position)
+    #[allow(clippy::result_large_err)]
     fn transition(
         &mut self,
         latest_retrieved: BlockHeader,
@@ -943,6 +944,7 @@ where
     /// ## Note
     /// This method assumes that at least one synchronizer is in Advanced, Ready or
     /// Delayed state, it will return an error in case this is not the case.
+    #[allow(clippy::result_large_err)]
     fn reinit_block_history(
         sync_streams: &mut [SynchronizerStream],
         block_history: &mut BlockHistory,
@@ -986,6 +988,7 @@ where
     ///
     /// Used once before entering the main loop, where all-Stale is a fatal configuration
     /// problem (nothing to sync from), not a temporary disconnect.
+    #[allow(clippy::result_large_err)]
     fn require_active_stream(sync_streams: &[SynchronizerStream]) -> BlockSyncResult<()> {
         if sync_streams
             .iter()
@@ -1009,6 +1012,7 @@ where
     ///
     /// Returns `Err` if at least one synchronizer has permanently ended while all
     /// remaining ones are stale — no recovery path exists.
+    #[allow(clippy::result_large_err)]
     fn check_streams(sync_streams: &[SynchronizerStream]) -> BlockSyncResult<()> {
         let mut has_any_ended = false;
         let mut latest_ended_stream: Option<&SynchronizerStream> = None;
--- a/crates/tycho-client/src/feed/mod.rs
+++ b/crates/tycho-client/src/feed/mod.rs
@@ -1,3 +1,5 @@
+#![allow(clippy::result_large_err)]
+
 use std::{
     collections::{HashMap, HashSet},
     fmt::{Display, Formatter},
@@ -100,6 +102,7 @@ impl HeaderLike for BlockHeader {
     }
 }
 
+#[allow(clippy::large_enum_variant)]
 #[derive(Error, Debug)]
 pub enum BlockSynchronizerError {
     #[error("Failed to initialize extractor '{extractor}': {source}")]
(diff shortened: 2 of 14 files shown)


crates/tycho-client-py/python/tycho_indexer_client/dto.py
```diff
@@ -56,7 +88,7 @@ class ChangeType(str, Enum):
 
 
 class ExtractorIdentity(BaseModel):
-    chain: Chain
+    chain: Union[Chain, CustomChainConfig]
```
brunoguerios, line 91: Decided to go with a Union here to handle custom chains more explicitly. Alternative would be a type alias, but I don't see much value in it.

crates/tycho-common/src/models/mod.rs
```diff
@@ -58,6 +59,69 @@ pub type ProtocolSystem = String;
...
+}
+
+impl DeepSizeOf for CustomChainConfig {
+    fn deep_size_of(&self) -> usize {
+        0
+    }
+
+    fn deep_size_of_children(&self, _context: &mut deepsize::Context) -> usize {
+        0
+    }
```
brunoguerios: Not completely sure I got the requirements right on this one. Let me know if you think I should return something other than zeros here.

crates/tycho-common/src/models/mod.rs
```diff
@@ -165,6 +251,19 @@ fn native_pol(chain: Chain) -> Token {
...
 }
 
+fn native_custom(cfg: &CustomChainConfig) -> Token {
+    let addr = Bytes::from(cfg.native.address.as_bytes().to_vec());
+    Token::new(
+        &addr,
+        cfg.native.symbol.as_str(),
+        cfg.native.decimals as u32,
+        0,
+        &[Some(2300)],
```
brunoguerios, line 342: Not sure how to set gas over here. Maybe it's something we should require from chain config as well?

crates/tycho-common/src/models/mod.rs
```diff
@@ -173,6 +272,24 @@ fn wrapped_native_pol(chain: Chain, address: &str) -> Token {
...
+            .address
+            .as_bytes()
+            .to_vec(),
+    );
+    Token::new(
+        &addr,
+        cfg.wrapped_native.symbol.as_str(),
+        cfg.wrapped_native.decimals as u32,
+        0,
+        &[Some(2300)],
```
brunoguerios, line 368: Same here.

crates/tycho-common/src/models/mod.rs
```diff
@@ -223,6 +341,11 @@ impl Chain {
             // BSC (BNB ≈ $630): 32 BNB ≈ $20K, 320 BNB ≈ $200K
             (Chain::Bsc, TvlThresholdTier::Low) => 32.0,
             (Chain::Bsc, TvlThresholdTier::Medium) => 320.0,
+
+            (Chain::Custom(cfg), TvlThresholdTier::Low) => cfg.default_tvl_thresholds.low as f64,
+            (Chain::Custom(cfg), TvlThresholdTier::Medium) => {
+                cfg.default_tvl_thresholds.medium as f64
```
brunoguerios: Any reason why thresholds are f64?
In order to have comparison available, I set it as u64 and cast it to f64 here.
Let me know if I should update default_tvl_threshold type and handle accordingly.

crates/tycho-common/src/models/mod.rs
```diff
@@ -432,3 +557,81 @@ pub enum MergeError {
     #[error("Cannot merge: {0}")]
     InvalidState(String),
 }
+
+#[cfg(test)]
+mod tests {
```
brunoguerios, line 641: I asked Claude to create some unit tests, but they seem rather trivial checking parsing and error handling. Usually I think they don't really add much value, but I decided to leave them here so we can discuss keeping them or not.

crates/tycho-indexer/src/main.rs
```diff
@@ -696,6 +760,78 @@ async fn run_analyze_tokens(
...
+        let token = TokenConfig {
+            address: "0x0000000000000000000000000000000000000000".to_string(),
+            symbol: "TOOLONGSYM".to_string(),
+            decimals: 18,
+        };
+        assert!(parse_chain_token(&token).is_err());
+    }
+
+    #[test]
+    fn test_resolve_custom_chain_from_yaml() {
```
brunoguerios, line 752: Same here about unit tests likely being unnecessary. The only one that I think is more relevant is this one that covers not only the yaml deserialization, but also the custom chain resolution, making sure the whole custom chain config worked as expected end-to-end.

crates/tycho-client/src/feed/mod.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(clippy::result_large_err)]
```
brunoguerios: Clippy highlighted a few of these large errors or large enum variants. Claude says since it's only within errors (i.e. not in the hot path), it should be ok to allow them. The alternative would be to Box them, but I'm not fully aware of the impact of that change, so I decided to stick with allowing for now and refactor later based on your feedback.

brunoguerios (commented):
I left a few comments that should help the review process.
Let me know if you have any feedback 😉

crates/tycho-client/src/feed/mod.rs
zizou0x: We tend to dislike these module scoped "allow" because they could end up hidding real warning, especially on such a big module.

I think the best would be to apply it specifically to the places the raise it. But I think we can find a way to make clippy happy, will post in a later comment.

crates/tycho-common/src/models/mod.rs
zizou0x: It should return the size of the objects, so it's not 0. Not sure how ArrayString allocates memory, maybe for deep_size_of_children it is but not deep_size_of because we for sure have some stack allocations here

Let me know if you need my input for size calculation, we can look into it together.

crates/tycho-common/src/models/mod.rs
```diff
@@ -58,6 +59,69 @@ pub type ProtocolSystem = String;
...
+    pub fn as_bytes(&self) -> &[u8] {
+        &self.bytes[..self.len as usize]
+    }
+}
+
+#[derive(Error, Debug, PartialEq)]
+pub enum ChainAddressError {
+    #[error("address is {0} bytes, max is 32")]
+    TooLong(usize),
+}
```
zizou0x, line 104: I would either make it a simple struct `pub struct ParseAddressError(String);` or make it more generic and name it "ModelError" or something like this because I find this one too specific to ever be extended (which kills the purpose of having an enum)

crates/tycho-common/src/models/mod.rs
```diff
@@ -58,6 +59,69 @@ pub type ProtocolSystem = String;
...
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize, ToSchema)]
+pub struct ChainTokenConfig {
+    #[schema(value_type = String)]
+    pub address: ChainAddress,
+    #[schema(value_type = String)]
+    pub symbol: ArrayString<8>,
+    pub decimals: u8,
```
zizou0x: Do we need all these fields to be public too?

crates/tycho-common/src/models/mod.rs
```diff
@@ -97,6 +147,41 @@ pub enum Chain {
     Bsc,
     Unichain,
     Polygon,
+    Custom(CustomChainConfig),
```
zizou0x, line 231: Can we `Box<>` it? I think it would solve all the clippy warnings that you got

crates/tycho-common/src/dto.rs
```diff
@@ -66,6 +55,20 @@ impl Chain {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        models::Chain::from(*self).fmt(f)
+    }
+}
+
+impl FromStr for Chain {
+    type Err = strum::ParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        models::Chain::from_str(s).map(Chain::from)
```
zizou0x: I'd use this to make it easier to read
```suggestion
        models::Chain::from_str(s).map(Self::from)
```

crates/tycho-indexer/src/main.rs
```diff
@@ -68,14 +70,35 @@ use tycho_storage::postgres::{builder::GatewayBuilder, cache::CachedGateway};
 
 mod ot;
 
+#[derive(Debug, Deserialize, Clone)]
+struct TokenConfig {
+    address: String,
+    symbol: String,
+    decimals: u8,
+}
```
zizou0x: Wouldn't it make sense to directly re-use what we declared in tycho-common?

crates/tycho-client-py/python/tycho_indexer_client/dto.py
```diff
@@ -137,7 +169,7 @@ class ProtocolComponent(BaseModel):
     id: str
     protocol_system: str
     protocol_type_name: str
-    chain: Chain
+    chain: Union[Chain, CustomChainConfig]
```
zizou0x, line 172: Are we returning the whole `CustomChainConfig` together with each `ProtocolComponent`? I think that's really overkill and we should just return the name of the chain on the server side. Then this can become a str.

crates/tycho-common/src/models/mod.rs
zizou0x: The reason is because some people might want to have 0.1 to filter out only empty pools but keep low tvls ones. As this could be decoded in any token (let's say ETH is worth 20k$ one day 🤞) then having to choose between 0$ and 20k$ is too big of a step.

crates/tycho-common/src/models/mod.rs
zizou0x: I think user wouldn't know what to set it to either.

2300 is the native token transfer cost on most evm chains, this is fine like this I'd say, we can always come back to it if we ever feel that this is a problem.

crates/tycho-client-py/python/tycho_indexer_client/dto.py
brunoguerios: Yeah, I agree it feels a bit of an overkill. I didn't change any of the logic in that sense as it's not relevant to this task though. I'd say the best approach is to create a github issue to track this and refactor in a separate PR. What do you think?
Ps: likely that same logic applies to most other classes that contain Chain and could simply contain `chain_name` or `chain_id`?

crates/tycho-client/src/feed/mod.rs
brunoguerios: Agreed. I'll reduce the scope of the `allow` here.

crates/tycho-common/src/models/mod.rs
brunoguerios: Agreed. I'll refactor this a bit and then see which approach makes more sense.

crates/tycho-common/src/models/mod.rs
brunoguerios: If we construct ChainTokenConfig directly without a constructor (as we do on tycho-indexer main.rs), they need to be pub. I'll explore the idea of adding a constructor and removing pub to see how it looks like. It might make the code easier to read.

crates/tycho-common/src/models/mod.rs
brunoguerios: Ok, I'll revisit this. In case I can't figure out using Claude, I'll ping you so we can look into it together 👍

crates/tycho-common/src/models/mod.rs
brunoguerios: If we Box it, Chain loses Copy, which seems to outweigh the benefits.
Clippy warnings happened on a few errors that contained Chain in them. Since they are not in the hot path, it should be fine.
Let me know if that makes sense or if you still think we should Box it.

crates/tycho-common/src/models/mod.rs
brunoguerios: Ok, I'll try to refactor them to f64 and see how I can handle comparison here.

crates/tycho-indexer/src/main.rs
brunoguerios: It does feel a bit weird to have different (yet so similar) types here and in tycho-common, but they serve different purposes. 
Here is just a raw yaml input (i.e. address is a string) and there address it's a validated domain type (i.e. address is a byte array), so I couldn't find a way to reuse them.
Let me know if you think I'm missing something.

brunoguerios (commented):
Thanks for the review!
I answered all comments and will start working on the suggestions.

crates/tycho-common/src/models/mod.rs
brunoguerios: done ✅

crates/tycho-common/src/models/mod.rs
brunoguerios: done ✅

crates/tycho-client/src/feed/mod.rs
brunoguerios: done ✅

crates/tycho-common/src/models/mod.rs
brunoguerios: done ✅

crates/tycho-common/src/models/mod.rs
brunoguerios: done ✅

crates/tycho-common/src/dto.rs
brunoguerios: done ✅

crates/tycho-common/src/models/mod.rs
zizou0x: What bothers me is that in Rust an enum always takes the memory size of the biggest value. So here we are increasing the size of all `Chain` entities from 1 to a few hundreds bytes, even if its `Chain::Ethereum` it's getting a few hundred bytes allocated. 

It feels like we could hold a ref instead here. Maybe `Custom(&'static CustomChainConfig)` would work?

crates/tycho-common/src/models/mod.rs
zizou0x: Which one did you chose? I don't see the difference here

crates/tycho-indexer/src/main.rs
zizou0x: What's the issue when you try to deserialize from the yml config directly to the tycho-common type? We might be able to solve this with a custom yaml deserializer implementation on the tycho-common model

crates/tycho-client-py/python/tycho_indexer_client/dto.py
zizou0x: Would be nice to have a test for it, just to make sure that deserialization here matches what's being serialized by Tycho

crates/tycho-client-py/python/tycho_indexer_client/dto.py
zizou0x: Happy to refactor it in a separate PR 👍

crates/tycho-common/src/models/mod.rs
zizou0x: I would honestly keep them, they are all very simple tests but all of them are testing a path in the code that would be untested if we didn't have them.

crates/tycho-indexer/src/main.rs
zizou0x: Same answer here, if I'm not wrong they are testing a logic that would be untested without them, so I'd say they are relevant

crates/tycho-common/src/models/mod.rs
brunoguerios: I went with the more generic approach.
I realized there were more errors being thrown related to chain config, so I decided to rename the enum to `ChainConfigError` and move them all in there.

crates/tycho-common/src/models/mod.rs
brunoguerios: Apparently this doesn't work either because we lose runtime construction, so load config from yaml file won't work anymore. I asked Claude to evaluate other options and it says allowing the large variant is still the best we got 😬 

Here's a comparison table on the options it came up with:

Approach | Small variant | Copy | Serde | Runtime construction
-- | -- | -- | -- | --
Current (Custom(CustomChainConfig)) | ✗ | ✓ | ✓ | ✓
Custom(Box<CustomChainConfig>) | ✓ | ✗ | ✓ | ✓
Custom(&'static CustomChainConfig) | ✓ | ✓ | ✗ | ✗
Custom(Arc<CustomChainConfig>) | ✓ | ✗ | ✓ | ✓

I even explored the idea of splitting `Chain` and `ChainConfig` into separate concepts, but that is not super clean either 😕

crates/tycho-indexer/src/main.rs
brunoguerios: Yes, we can add a custom serialize/deserialize - I'll do that over here for `TokenConfig` 👍 
Doing the same for `ChainConfig` can't really be done as is, because chain `name` is the key in the yaml file. Restructuring the yaml file to have a custom chains `list` instead of a `hashmap` solves this issue, but might not be ideal. I pushed this change in separate commit so we can evaluate and simply revert if we decide it's better the other way.
Let me know your thoughts.

crates/tycho-client-py/python/tycho_indexer_client/dto.py
brunoguerios: done ✅

crates/tycho-common/src/models/mod.rs
zizou0x: As discussed on Slack, this will be refactored on the release branch

crates/tycho-indexer/src/main.rs
zizou0x: I think a list is nice 👍

zizou0x (approved):
Thanks! LGTM 🔥

PR #1076, opened 2026-06-09
feat: add baseline mercury swap adapter

ind-igo:
Integration for Baseline Markets (https://www.baseline.markets/)

Notes:

- Indexing, simulation and execution integration for Baseline pools
- Uses storage slot tracking
- Production executor deployment is not included in this PR. The executor is validated through protocol-testing state overrides. Prod addresses can be deployed and added after review.


zach030:
Hi! Thanks for the PR. I took a quick look at the substreams part.
I see you’re using the `getQuoteState` RPC call in substreams to fetch the live state. I think this can work, but it may impact indexer performance, so it’s probably worth looking for a better way.
From what I can see, `tycho-simulation` depends on those states right now. Would it be possible to update them by decoding events instead, like `Swap`, `AddLiquidity`, or `RemoveLiquidity`?
Another option would be to derive the state updates from storage diffs. That would mean digging into the contract storage layout, figuring out the slot for each state, and decoding the values from there. You can use the [storage code](https://github.com/propeller-heads/tycho/tree/main/protocols/substreams/ethereum-uniswap-v3/src/storage) in `uniswap-v3` as a reference.

ind-igo:
> Hi! Thanks for the PR. I took a quick look at the substreams part. I see you’re using the `getQuoteState` RPC call in substreams to fetch the live state. I think this can work, but it may impact indexer performance, so it’s probably worth looking for a better way. From what I can see, `tycho-simulation` depends on those states right now. Would it be possible to update them by decoding events instead, like `Swap`, `AddLiquidity`, or `RemoveLiquidity`? Another option would be to derive the state updates from storage diffs. That would mean digging into the contract storage layout, figuring out the slot for each state, and decoding the values from there. You can use the [storage code](https://github.com/propeller-heads/tycho/tree/main/protocols/substreams/ethereum-uniswap-v3/src/storage?rgh-link-date=2026-06-10T07%3A37%3A45Z) in `uniswap-v3` as a reference.

Thanks for the review! I explored trying to do this directly from events but this was not able to reconstruct the full state of the amm. That getQuoteState was specifically added for this purpose. Is it a deadend to use the getQuoteState flow?

I will look into the storage diff method in the meantime.

zach030:
> > Hi! Thanks for the PR. I took a quick look at the substreams part. I see you’re using the `getQuoteState` RPC call in substreams to fetch the live state. I think this can work, but it may impact indexer performance, so it’s probably worth looking for a better way. From what I can see, `tycho-simulation` depends on those states right now. Would it be possible to update them by decoding events instead, like `Swap`, `AddLiquidity`, or `RemoveLiquidity`? Another option would be to derive the state updates from storage diffs. That would mean digging into the contract storage layout, figuring out the slot for each state, and decoding the values from there. You can use the [storage code](https://github.com/propeller-heads/tycho/tree/main/protocols/substreams/ethereum-uniswap-v3/src/storage?rgh-link-date=2026-06-10T07%3A37%3A45Z) in `uniswap-v3` as a reference.
> 
> Thanks for the review! I explored trying to do this directly from events but this was not able to reconstruct the full state of the amm. That getQuoteState was specifically added for this purpose. Is it a deadend to use the getQuoteState flow?
> 
> I will look into the storage diff method in the meantime.

This approach should work, but it may affect indexing performance on Base. Since we can probably solve it by decoding storage diffs too, I’d prefer going with that instead.

ind-igo:
@zach030 I have updated the integration to use storage slot tracking.

protocols/substreams/ethereum-baseline/BASELINE_STORAGE_DIFF_DESIGN.md
zach030, line 1: Thanks for the nice doc! Can you move it to the Baseline folder?

protocols/testing/run.Dockerfile
zach030, line 1: Are these changes necessary?

crates/tycho-indexer/src/cli.rs
zach030, line 1: Why do you need to add the `--implementation-type` flag?

protocols/testing/src/config.rs
```diff
@@ -144,6 +148,8 @@ pub struct IntegrationTestsConfig {
     pub initialized_accounts: Option<Vec<String>>,
     pub skip_balance_check: bool,
     pub protocol_type_names: Vec<String>,
+    #[serde(default = "default_implementation_type")]
```
zach030: Same here. why add the `--implementation-type` flag?

protocols/testing/src/execution.rs
```diff
@@ -9,6 +9,7 @@ use std::{collections::HashMap, sync::LazyLock};
 use miette::{miette, IntoDiagnostic, WrapErr};
 use tycho_test::execution::models::RouterOverwritesData;
 pub const ROUTER_BYTECODE_JSON: &str = include_str!("../fixtures/TychoRouter.runtime.json");
+const FEE_CALCULATOR_BYTECODE_JSON: &str = include_str!("../fixtures/FeeCalculator.runtime.json");
```
zach030, line 12: `FEE_CALCULATOR` should be unrelated to Baseline. We’ll add this contract bytecode later, so please revert the related changes.

protocols/testing/src/test_runner.rs
zach030, line 1: This file probably doesn’t need to be changed. Under the `testing` directory, we should only need to add the Baseline runtime bytecode and register it in `state_registry`. Everything else should stay unchanged.

protocols/substreams/ethereum-baseline/src/pb/mod.rs
zach030, line 1: I don’t see Baseline using any proto files. Are these needed?

protocols/substreams/ethereum-baseline/src/pool_factories.rs
```diff
@@ -0,0 +1,144 @@
...
+use substreams_ethereum::Event;
+use tycho_substreams::models::{
+    Attribute, ChangeType, FinancialType, ImplementationType, ProtocolComponent, ProtocolType,
+};
+
+#[derive(Deserialize)]
+pub struct DeploymentConfig {
+    #[serde(with = "hex::serde")]
+    pub relay_address: Vec<u8>,
+    pub protocol_type_name: String,
```
zach030: IMO this can be hardcoded, no need to pass it as a param.

protocols/substreams/ethereum-baseline/src/pb/mod.rs
ind-igo: deleted

protocols/testing/src/execution.rs
ind-igo: apologies, this was from a bad merge. fixed

protocols/testing/run.Dockerfile
ind-igo: fixed, only kept the base -> ethereum mappings.

crates/tycho-indexer/src/cli.rs
ind-igo: Removed, I had a misunderstanding of how the implementation type worked. All of the tycho-level changes should be reverted and only baseline changes should remain.

protocols/substreams/ethereum-baseline/src/modules/slot_stores.rs
```diff
@@ -0,0 +1,358 @@
...
+                    .is_none()
+            })
+            .for_each(|change| {
+                store.set(
+                    change.ordinal,
+                    slot_index_key(&change.key),
+                    &slot_value_hex(&change.new_value),
+                )
+            });
+    });
```
zach030: IIUC, we'd index all the slots of the relay contract here. I'm wondering if that's necessary, could we figure out the boundary of the slot range that stores the pools, so we can avoid indexing every slot?

protocols/substreams/ethereum-baseline/src/modules/mod.rs
```diff
@@ -0,0 +1,343 @@
...
+
+    block.transactions().for_each(|tx| {
+        tx.logs_with_calls()
+            .filter_map(|(log, _call)| {
+                if log.address.as_slice() != RELAY_ADDRESS {
+                    return None;
+                }
+                Swap::match_and_decode(log)
+                    .map(|event| format!("0x{}", hex::encode(event.b_token)))
+                    .or_else(|| manual_updates::maybe_component_id(log))
```
zach030, line 170: Would it be clearer to write it as
```rust
Swap::match_and_decode(log)
    .map(|event| event.b_token)
    .or_else(|| CreatorFeePctSet::match_and_decode(log).map(|event| event.b_token))
    .or_else(|| LiquidityFeePctSet::match_and_decode(log).map(|event| event.b_token))
    .map(|b_token| format!("0x{}", hex::encode(b_token)))

```

protocols/substreams/ethereum-baseline/src/modules/mod.rs
```diff
@@ -0,0 +1,343 @@
...
+                state_tx.change.change_type(),
+            )
+            .unwrap_or_default();
+            if !quote_state_attributes.is_empty() {
+                builder.add_entity_change(&EntityChanges {
+                    component_id,
+                    attributes: quote_state_attributes,
+                });
+            }
+        });
```
zach030: I think the design here is too complex. My idea: for all `new_components`, we initialize the entity changes with default values in one pass and mark them as `Creation`, then update them based on the `state_deltas`.

crates/tycho-simulation/src/evm/protocol/baseline/decoder.rs
```diff
@@ -0,0 +1,529 @@
...
+                "Expected Baseline component id to be bToken address: {e}"
+            ))
+        })?;
+
+        if snapshot.state.component_id != snapshot.component.id {
+            return Err(InvalidSnapshotError::ValueError(format!(
+                "Baseline component id mismatch: state={} component={}",
+                snapshot.state.component_id, snapshot.component.id
+            )));
+        }
```
zach030: IMO this is unnecessary

crates/tycho-simulation/src/evm/protocol/baseline/decoder.rs
```diff
@@ -0,0 +1,529 @@
...
+        total_supply: get_u256(attributes, TOTAL_SUPPLY)?,
+        total_b_tokens: get_u256(attributes, TOTAL_B_TOKENS)?,
+        total_reserves: get_u256(attributes, TOTAL_RESERVES)?,
+        reserve_decimals: get_u256(attributes, RESERVE_DECIMALS)?,
+        liquidity_fee_pct: get_u256(attributes, LIQUIDITY_FEE_PCT)?,
+        pending_surplus: get_u256(attributes, PENDING_SURPLUS)?,
+        should_settle_pending_surplus: get_bool(attributes, SHOULD_SETTLE_PENDING_SURPLUS)?,
+        max_sell_delta: get_u256(attributes, MAX_SELL_DELTA)?,
+        snapshot_active_price: get_u256(attributes, SNAPSHOT_ACTIVE_PRICE)?,
+    })
```
zach030, line 150: Are these attributes all required? If so, we need to set default values for them in substreams.

crates/tycho-simulation/src/evm/protocol/baseline/math.rs
```diff
@@ -0,0 +1,1104 @@
...
+}
+
+pub(super) fn spot_price(state: &BaselineQuoteState, is_buy: bool) -> Result<f64, SimulationError> {
+    let price = compute_active_price(&state.snapshot_curve)?;
+    let fee_adjustment = u_to_bi(state.snapshot_curve.swap_fee) * 2u8;
+    let price = if is_buy {
+        mul_wad(&price, &(wad() + fee_adjustment))
+    } else {
+        mul_wad(&price, &(wad() - fee_adjustment))
+    };
```
zach030, line 84: Isn't this backwards? The fee should always make the trader's price worse, but here it's added on sells and subtracted on buys.

crates/tycho-protobuf/Cargo.toml
```diff
@@ -18,4 +18,3 @@ chrono.workspace = true
 num-bigint.workspace = true
 num-traits.workspace = true
 tracing.workspace = true
-
```
zach030, line 21: No change needed here.

zach030 (commented):
Thanks for the update! I know switching to parsing from storage added a lot of complexity, could we add more comments in the substreams to make it easier to follow and to help with future iterations? 🙏 
I also left a few small comments, please take a look.

crates/tycho-simulation/src/evm/protocol/baseline/decoder.rs
ind-igo: yes unfortunately all necessary. will set defaults.

PR #1077, opened 2026-06-09
feat(ekubo-v3): support new TWAMM extension

die-herdplatte:
A new version of the TWAMM extension was deployed, with the only noticeable change being that the cancellation fee for TWAMM orders was removed.
As this doesn't affect any logic in this integration, the required change is minimal.
The first pool for this extension was created [in block 24995117](https://etherscan.io/tx/0x79c97f33b095982d051667b3335533cacb7915a491308907bd2205a5f4dc1216).

On another note, I tried to setup execution simulation for Ekubo but:
- The `TychoRouter` runtime bytecode is stale (simulations fail because `singleSwap` isn't defined in the checked-in bytecode)
- The [`fee_calculator_bytecode` is empty](https://github.com/propeller-heads/tycho/blob/4f6eab7d2fa35d03efa0d53b4d89f7154cda0631/protocols/testing/src/execution.rs#L119), causing the call to never reach the executor because [calldata decoding fails](https://github.com/propeller-heads/tycho/blob/4f6eab7d2fa35d03efa0d53b4d89f7154cda0631/crates/tycho-execution/contracts/src/Dispatcher.sol#L295)

I have a fix which I can put in this or another PR if you want.


protocols/substreams/ethereum-ekubo-v3/integration_test.tycho.yaml
```diff
@@ -100,3 +100,19 @@ tests:
...
         creation_tx: "0xf394db915b31b99b2f8683e2d888e9406ea98211b8f87e4fb22079dbc0110afd"
+
+  - name: twamm_new
+    start_block: 24995117
+    stop_block: 24995117
+    expected_components:
+      - id: "0xedeaae143f233a3a5d4fabd3166afa0e2108fe7741489237274b939ca17fcff8"
+        tokens:
+          - "0x0000000000000000000000000000000000000000"
+          - "0x4c46e830bb56ce22735d5d8fc9cb90309317d0f"
```
zach030: This address is missing a leading zero.

zach030 (commented):
The logic is clean and clear! Could you bump the version and add a changelog for this?

zach030 (approved):

PR #1078, opened 2026-06-09
fix(execution): Address small audit comments to new executors

dianacarvalho1:
Liquorice:
- Added selector allowlist: only settleSingle (0x9935c868) and settle (0xcba673a7) are permitted; reverts with InvalidSelector otherwise
- Added NotAContract error and code.length == 0 checks for both constructor arguments
- Replaced address(0) with ETH_ADDRESS (from NativeETH.sol) for native ETH detection, consistent with the rest of the router
- Added new tests for it and updated existing ones
- Mention the selector allowlist recommendation in the docs

Etherfi:
- Added NotAContract error and code.length == 0 checks for all constructor arguments
- Added EtherfiExecutor__ZeroAddress error instead of just reverting
- Don't get the amountIn from balance (this is already done in the dispatcher level)
- Added missing test and updated existing ones

Ekubo V3:
- Remove unnecessary constructor

TODO:
- [ ] Redeploy them (after approval) (I think we don't need to redeploy the EkuboV3 one though)


crates/tycho-execution/contracts/src/executors/EtherfiExecutor.sol
```diff
@@ -86,14 +86,8 @@ contract EtherfiExecutor is IExecutor {
         direction = _decodeData(data);
 
         if (direction == EtherfiDirection.EethToEth) {
-            // eETH is share-based and rounds down on amount conversions;
-            // cap redeem amount to current balance to avoid 1-wei dust reverts.
-            uint256 redeemAmount = IERC20(eethAddress).balanceOf(address(this));
-            if (redeemAmount > amountIn) {
-                redeemAmount = amountIn;
-            }
```
dianacarvalho1, line 94: this balance check is done in Dispatcher anyway already

tamaralipows (approved):
Perfect - thank you so much!

Troshchk (approved):
Thank you!

PR #1079, opened 2026-06-09
feat(fee-calculator): Add getAllClientFees view method and sub-BPS fee precision support

dianacarvalho1:
Review commit by commit for a premium experience ✨


crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -80,11 +86,15 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
         view
         returns (uint256 amountOut, FeeRecipient[] memory feeRecipients)
     {
-        (uint16 routerFeeOnOutputBps, uint16 routerFeeOnClientFeeBps) =
+        (uint32 routerFeeOnOutputBps, uint32 routerFeeOnClientFeeBps) =
             _getFeeInfo(client);
 
+        // Scale clientFeeBps from legacy scale (10_000 = 100%) to internal scale
+        // (100_000_000 = 100%) so both fee types can be compared and combined.
+        uint32 scaledClientFeeBps = uint32(clientFeeBps) * 10_000;
```
kayibal, line 96: Where does this come from? From the signature? If so I would not scale this here instead change this also on the fynd binary.

Mh ok I understand this protects someone with an old fynd version from charging to little fee, right? Yeah it is somewhat dangerous. We should think a bit more about this. Would prefer if we didn't have to scale here.

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -95,16 +105,18 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
...
+            uint256 clientFeeNumerator = amountOut * scaledClientFeeBps;
+            uint256 totalClientFee = clientFeeNumerator / 100_000_000;
 
             // Calculate router's cut of the client fee
             if (routerFeeOnClientFeeBps > 0) {
+                // Both fees use the 100_000_000 scale, so denominator is 100_000_000^2
                 routerFeeOnClientFee =
-                    (clientFeeNumerator * routerFeeOnClientFeeBps) / 100_000_000;
+                    (clientFeeNumerator * routerFeeOnClientFeeBps)
+                        / 10_000_000_000_000_000;
```
kayibal: This should be a constant than can be queried by the RPC

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -116,7 +128,7 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
         // Calculate router fee on output amount if > 0
         if (routerFeeOnOutputBps > 0) {
             uint256 routerFeeOnOutput =
-                (amountOut * routerFeeOnOutputBps) / 10000;
+                (amountOut * routerFeeOnOutputBps) / 100_000_000;
```
kayibal: Also should be a constant

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -214,18 +228,47 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
...
+    {
+        CustomFees memory customFees = _customRouterFees[client];
+        uint32 fee = customFees.hasCustomFeeOnOutput
+            ? customFees.feeBpsOnOutput
+            : _routerFeeOnOutputBps;
+        if (fee == 0) return 0;
+        // Convert from internal scale (100_000_000 = 100%) to legacy BPS scale (10_000 = 100%).
+        // Return at least 1 so callers can detect that a fee is active.
+        uint32 legacyBps = fee / 10_000;
+        return uint16(legacyBps > 0 ? legacyBps : 1);
```
kayibal, line 261: Mh yeah not nice either, I'd prefer if we changed that completely without dragging legacy decisions around.

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -290,25 +334,50 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
...
+    {
+        uint256 count = _customFeeClients.length();
+        clients = new address[](count);
+        fees = new CustomFees[](count);
+        for (uint256 i = 0; i < count; i++) {
+            address client = _customFeeClients.at(i);
+            clients[i] = client;
+            fees[i] = _customRouterFees[client];
+        }
+    }
```
kayibal: I imagine that this can eventually run out of gas or memory on some rpcs, it would be better to allow specifying a range as input so you can paginate over it.

crates/tycho-execution/contracts/src/FeeCalculator.sol
tamaralipows: I guess then this will have to wait for a router redeployment? :/

kayibal: Really? The risk is that some clients take too little fee until they upgrade Fynd. I mean ideally this legacy calldata would lead to reverts, but I don't think we have a way to discover the tycho-execution version the calldata was created with?

tamaralipows: Oh I thought you meant that you would also like this to change on the router side. Misunderstanding. I'm also fine with clients potentially taking too little fee until they upgrade tbh. I'm not aware of any clients taking their own fees right now. We can easily verify if anyone is doing that.

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -2,23 +2,15 @@
 pragma solidity ^0.8.26;
 
 import {AccessControl} from "@openzeppelin/contracts/access/AccessControl.sol";
+import {
+    EnumerableSet
+} from "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
```
tamaralipows, line 7: Nice, you decided to use this. The logic looks really simple with this.

tamaralipows (commented):
Thank you! In general I agree with Alan's comments. Let me know what you think

crates/tycho-execution/contracts/src/FeeCalculator.sol
dianacarvalho1: no no, it's not about Fynd. It's about not redeploying the Router entirely. The `clientFeeBps` is defined in the [Router](https://github.com/propeller-heads/tycho/blob/main/crates/tycho-execution/contracts/src/TychoRouter.sol#L85) and passed in the [swap methods](https://github.com/propeller-heads/tycho/blob/main/crates/tycho-execution/contracts/src/TychoRouter.sol#L204). To get rid of this scaling we would have to redeploy the Router. This was discussed on Slack [here](https://propellerswap.slack.com/archives/C05LQALE9L1/p1781016565271469)

crates/tycho-execution/contracts/src/FeeCalculator.sol
dianacarvalho1: you mean the `10_000_000_000_000_000`?

kayibal (approved):

PR #1081, opened 2026-06-09
feat(substreams): add BopAMM VM integration

tvinagre:
Adds an `ethereum-bopamm` Substreams package that indexes Bebop's BopAMM PMM as a Tycho VM protocol.

## What it does
- Discovers books (`asset/USDC`, one component per `assetId`) from the pricing module's asset-config storage writes — the contracts emit no config events.
- Tracks the full storage + code of the settlement, pricing module, and registry contracts (fixed set, no DCI).
- Tracks the market-maker wallet's token balances as per-book TVL (USDC duplicated across books, since inventory is shared).
- Records per-book committed quote `ts`/`block` decoded from `updateState`/`batchUpdateStateWithSignature` calldata, and `Paused`/`Unpaused` state.
- Deployment addresses and storage slots are passed via substreams `params`, so the package can target another deployment without code changes.

Pricing is a pure function of the indexed storage (no external calls, no DCI). One module per file under `src/modules/`; shared helpers in `src/common.rs`; config in `src/config.rs`.

## Verification
- 8 unit tests (real `updateState` calldata decode, `bookId == assetId` cross-lock, `asset_config_slot` keccak vs on-chain, batch decode, token-index regression, params parse). `clippy -D warnings` and nightly `fmt` clean; host + wasm32 build; `substreams pack` succeeds.
- Live `substreams run` against mainnet over the book-creation and an active-trading range: both books created with the expected ids/tokens/attributes/creation_tx, committed_ts/block on quote updates, maker balances, and `balance_owner` all emitted as expected.

See `HANDOVER.md` for the protocol design, interfaces, storage layout, and lifecycle.

## Not included (follow-ups)
- `tycho-simulation` `BopAMMSwapAdapter` (`get_amount_out`) + the quote-override stream that pins `block.timestamp`/`block.number` and injects the maker balance — needed because `MAX_UPDATE_AGE() == 0` (exact-timestamp gate).
- `extractors.yaml` `vm:bopamm` entry (8 fields) — gated on releasing the spkg to S3 via `release.sh ethereum-bopamm`.
- Asset delisting and maker-rotation re-seed are documented limitations.

### DIFF
--- a/protocols/substreams/Cargo.lock
+++ b/protocols/substreams/Cargo.lock
@@ -774,6 +774,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "ethereum-bopamm"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "ethabi 18.0.0",
+ "getrandom 0.2.17",
+ "hex",
+ "itertools 0.13.0",
+ "keccak-hash",
+ "serde",
+ "serde_qs 0.13.0",
+ "substreams 0.5.22",
+ "substreams-ethereum 0.9.13",
+ "tycho-substreams 0.8.0 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "ethereum-cowamm"
 version = "0.1.3"
 dependencies = [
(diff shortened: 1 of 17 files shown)


tvinagre:
Superseded by #1095, which aggregates the substreams, simulation adapter, and execution layers into one branch (reconciled against current main — FermiSwap #1034 had since merged and shares the execution wiring). Closing in favor of that PR.

PR #1085, opened 2026-06-10
feat: include group swap logic in gas estimations

Troshchk:


crates/tycho-execution/src/encoding/evm/gas_estimator.rs
```diff
@@ -63,20 +65,25 @@ pub const ROUTER_FEES_ACTIVE: bool = true;
...
     }
 
     // Add user transfer overhead
     if strategy != Strategy::Split {
-        if let Some(first_swap) = solution.swaps().first() {
-            let protocol: &str = &first_swap.component().protocol_system;
+        if let Some(first_group) = swap_groups.first() {
+            let protocol: &str = &first_group.swaps[0]
+                .component()
+                .protocol_system;
```
louise-poole: I'm pretty sure the SwapGroup has protocol as a field. So this could be neatened to:
```suggestion
            let protocol = first_group.protocol_system.as_str();
```

crates/tycho-execution/src/encoding/evm/gas_estimator.rs
```diff
@@ -63,20 +65,25 @@ pub const ROUTER_FEES_ACTIVE: bool = true;
...
-        let swap_transfer_overhead = estimate_transfer_overhead(
-            &swap.component().protocol_system,
-            swap.token_in(),
-            swap.token_out(),
+    let swap_groups = group_swaps(solution.swaps());
+    for group in &swap_groups {
+        let first_swap = &group.swaps[0];
+        let last_swap = &group.swaps[group.swaps.len() - 1];
+        let group_transfer_overhead = estimate_transfer_overhead(
+            &first_swap.component().protocol_system,
```
louise-poole: Same here:
```suggestion
            &group.protocol_system,
```

crates/tycho-execution/src/encoding/evm/gas_estimator.rs
```diff
@@ -63,20 +65,25 @@ pub const ROUTER_FEES_ACTIVE: bool = true;
...
     let mut total_gas = BigUint::ZERO;
-    for swap in solution.swaps() {
-        let swap_transfer_overhead = estimate_transfer_overhead(
-            &swap.component().protocol_system,
-            swap.token_in(),
-            swap.token_out(),
+    let swap_groups = group_swaps(solution.swaps());
+    for group in &swap_groups {
+        let first_swap = &group.swaps[0];
+        let last_swap = &group.swaps[group.swaps.len() - 1];
```
louise-poole, line 71: A lot of these first and last swap lookups could be avoided if we made the token fields on SwapGroup `Token` instead of `Bytes`. I wonder if that change is worth it? 🤔

louise-poole (approved):
Some very minor comments, otherwise I've approved! Looks good to me!

crates/tycho-execution/src/encoding/evm/gas_estimator.rs
Troshchk: Nice catch! fixed 👍

crates/tycho-execution/src/encoding/evm/gas_estimator.rs
Troshchk: hmm, I am not sure it is worth it. I would prefer to avoid extra copying taking into account we do not have many lookups

dianacarvalho1: afff I guess this was because initially the tokens were `Bytes` on the `Swap` too. We only changed it to `Token` recently. but tbh I don't think it is worth changing `SwapGroup` just for these couple of lookups

crates/tycho-execution/src/encoding/evm/gas_estimator.rs
```diff
@@ -294,6 +319,44 @@ mod tests {
...
 
+    #[test]
+    fn test_sequential_grouped_two_hops_usv4() {
+        // Two consecutive USV4 hops get grouped: intermediate transfer is saved.
+        // A→B→C on uniswap_v4 becomes one group with flash accounting.
+        let solution = make_solution(vec![
+            make_swap_with_tokens("uniswap_v4", token_a(), token_b(), 100_000),
+            make_swap_with_tokens("uniswap_v4", token_b(), token_c(), 100_000),
+        ]);
+        let gas = estimate_gas_usage(&solution, Strategy::Sequential);
```
dianacarvalho1: actually.. a grouped swap like this would be marked as a Single swap (see [here](https://github.com/propeller-heads/tycho/blob/main/crates/tycho-execution/src/encoding/evm/tycho_encoders.rs#L67)). so we should pass the `Strategy::Single` here (not sure if this changes anything actually 🤔)

dianacarvalho1 (commented):
Thank you @Troshchk ! 🙏🏼 I have one small comment!

crates/tycho-execution/src/encoding/evm/gas_estimator.rs
Troshchk: It doesn't, but I have updated it for correctness

dianacarvalho1 (approved):
merci!

PR #1092, opened 2026-06-16
feat: add lunarbase to integration test

zach030:


zizou0x (commented):
We need to wait for the executor to be deployed and set on the router. But LGTM

zizou0x (approved):

PR #1087, opened 2026-06-11
feat(tycho-execution): add BopAMM executor and swap encoder

tvinagre:
## What

Execution support for BopAMM (Bebop's on-chain PMM): `BopAMMExecutor.sol`, `BopAMMSwapEncoder` (`vm:bopamm`), config entries, and tests — same structure as the FermiSwap execution files in #1034.

- **Executor**: `ProtocolWillDebit` — the TransferManager pulls tokenIn to the router and approves the BopAmmV2 settlement (`0xdB13ad0fcD134E9c48f2fDaEa8f6751a0F5349ca`), which debits via allowance and sends output to the `recipient` param (`outputToRouter = false`). `swap()` calls `settlement.swap(tokenIn, tokenOut, amountIn, 0, block.timestamp, receiver)`; protocol-level `minAmountOut` is 0 because the router enforces the user's minimum and the Dispatcher measures output via balance diff. No callbacks.
- **Encoder**: packed `(tokenIn, tokenOut)` — 40 bytes; the settlement is an executor immutable and the traded book follows from the token pair. Ethereum-only.
- **Wiring**: registry entry, `Constants.sol`, `TychoRouterTestSetup.sol` (executors 21→22), `test_executor_addresses.json`, `deploy-executors.js`.

## Timestamp gate

BopAMM swaps only succeed in a block whose timestamp equals the book's committed registry update (`StaleUpdate()` otherwise — on-chain takers need builder bundles). Fork tests pin block 25266710 (a quote-commit block for the WETH/USDC book) and fund the maker wallet, since the venue's quote caps at the committed lane size while settlement transfers come from the maker's inventory.

## Testing

- Rust: encoder unit tests (both directions + non-Ethereum rejection) and `test_single_encoding_strategy_bopamm_weth_usdc`; `clippy -D warnings` and fmt clean.
- Solidity: 9 executor tests (decode, transfer data, direct swap, stale revert, cross-language decode of Rust-generated calldata) + a full TychoRouter `singleSwap` integration test, all passing on a mainnet fork. Pre-existing `TychoRouterSingleSwap` suites re-run green (19/19).

## Note for whoever merges second (#1034 or this)

Both PRs append the 22nd executor to the deterministic test deployment sequence, so both predict the test executor address `0xe8dc788818033232EF9772CB2e6622F1Ec8bc840`. The PR that lands second must re-derive its `test_executor_addresses.json` entry (the executor moves to position 23) and regenerate `calldata.txt`.

[...]

PR #1082, opened 2026-06-10
feat(adapter-integration): add BopAMM swap adapter

tvinagre:
## What

Sell-side `ISwapAdapter` for BopAMM (Bebop's on-chain PMM), wrapping the BopAmmV2 settlement contract (`0xdB13ad0fcD134E9c48f2fDaEa8f6751a0F5349ca`), plus manifest and Foundry fork tests.

- **Pool id** packs `settlement (20 bytes) | assetId (12 bytes)`, matching the `vm:bopamm` substreams component ids (#1081).
- **price/getLimits** use the settlement's `quote()` view. `quote()` reverts `InsufficientLiquidity()` above the committed lane size, so limits are found by probing + bisection. The maker's inventory is not checked by `quote()` — limits can overestimate what `swap()` settles (the interface prefers overestimation, and the operator sizes lanes to inventory).
- **swap()** is sell-only: pull from caller, approve settlement, swap with `recipient = msg.sender`, output measured by balance diff. Buy orders revert `NotImplemented` — the venue quotes exact-input only.
- **Capabilities**: `SellOrder`, `PriceFunction`, `ConstantPrice` (verified constant up to the lane cap), `HardLimits`.
- **getTokens/getPoolIds** enumerate `getAssetConfig(0..64)` on the pricing module; every book is `asset/USDC`.

## Staleness gate

The update registry reverts `StaleUpdate()` unless `block.timestamp` equals the book's committed update timestamp (`MAX_UPDATE_AGE == 0`). Fork tests re-stamp the committed lane timestamp via `vm.store` (`keccak256(abi.encode(module, bookId))`, timestamp in the top 32 bits). At simulation runtime the timestamp is pinned via the `override_block_timestamp` component attribute (emitted by the substreams in #1081, consumed by tycho-simulation in #1034).

## Testing

14/14 fork tests pass against a mainnet fork (`ETH_RPC_URL`): pool/token enumeration, capabilities, limits in both directions on both books, constant-price check up to the limit, sell swaps both directions with balance assertions, and revert cases (above limit, stale timestamp, paused venue, pool/token mismatch, foreign settlement pool id, buy order). `BopAmmV2::swap` measures ~120k gas (manifest `protocol_gas: 120000`).

Depends on #1081 (component id format) and #1034 (`override_block_timestamp` support) for end-to-end simulation; the files here are self-contained and the tests run on main.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1095, opened 2026-06-16
feat: BopAMM (Bebop PMM) integration — indexing, simulation, execution

tvinagre:
Full BopAMM (Bebop's on-chain PMM) integration, aggregating the three previously separate PRs (#1081 substreams, #1082 adapter, #1087 execution) into one branch now that they've been reconciled against current main. Each layer is a self-contained commit series; the parts are disjoint except for the execution wiring (see below).

## Layers

**Substreams (`vm:bopamm`)** — indexes BopAMM as a VM protocol, one component per `asset/USDC` book. Books discovered from module asset-config storage writes (no events). Component id is the 32-byte `settlement ‖ assetId` packing (hex-decodable to the adapter's `bytes32 poolId`). Emits `balance_owner`, pause state, and `override_block_timestamp` (the committed quote ts, consumed by tycho-simulation to pass the registry's exact-timestamp `StaleUpdate()` gate). `initialized_accounts` bootstraps the registry, which predates the test range.

**Simulation adapter** — `BopAMMAdapter` under `protocols/adapter-integration`, wrapping the BopAmmV2 settlement. Sell-side (exact-input only), `ConstantPrice` + `HardLimits`; `getLimits` bisects `quote()`. `pricing`/`usdc` are resolved lazily (not in the constructor) because the simulation engine deploys the adapter in a context where the settlement code is injected at call time. Fork tests pinned to a quote-commit block.

**Execution** — `BopAMMExecutor` (`ProtocolWillDebit`, settlement as approval target, output to recipient) + `BopAMMSwapEncoder` (`vm:bopamm`, packed token pair). The router enforces `minAmountOut`; the Dispatcher measures output by balance diff.

## Reconciliation with FermiSwap (#1034, now on main)

FermiSwap merged while these branches were open and shares the execution wiring. Reconciled here: bopamm is appended as test executor index 22 (FermiSwap holds 21), with its re-derived deterministic test address `0x3Cff5E7eBecb676c3Cb602D0ef2d46710b88854E` and regenerated `calldata.txt`.

## Verification

- Substreams: unit tests; the `protocol-testing` harness passes `test_book_discovery` 1/1 end-to-end (component discovery + tracked-contract indexing).
- Adapter: 14/14 fork tests at the pinned block.
- Execution: 9 executor + 1 TychoRouter integration test pass; FermiSwap router and the TychoRouter single-swap suite (19 tests) re-run green after the executor-array change.
- `cargo clippy -D warnings` and rustfmt clean on tycho-execution.

Simulation remains skipped in the integration yaml until the `override_block_timestamp` support is exercisable on main; the harness logs the expected `StaleUpdate` warning there.

Supersedes #1081, #1082, #1087.


protocols/substreams/ethereum-bopamm/src/common.rs
```diff
@@ -0,0 +1,297 @@
...
+        .as_bytes()
+        .to_vec()
+}
+
+pub fn is_zero(value: &[u8]) -> bool {
+    value.iter().all(|&b| b == 0)
+}
+
+/// All currently-known book component ids, by probing the components store.
+pub fn enumerate_books(store: &StoreGetProto<ProtocolComponent>) -> Vec<String> {
```
die-herdplatte, line 84: Short-circuit possible since asset IDs are sequential?
If so, I think the iteration should be unbounded.

protocols/substreams/ethereum-bopamm/src/modules/map_components.rs
```diff
@@ -0,0 +1,71 @@
...
+/// bytes). Emits one `Creation` component per newly configured book.
+#[substreams::handlers::map]
+pub fn map_components(
+    params: String,
+    block: eth::v2::Block,
+) -> Result<BlockTransactionProtocolComponents> {
+    let config: DeploymentConfig = serde_qs::from_str(&params)?;
+    let slot_to_asset: HashMap<Vec<u8>, u64> = (0..MAX_ASSET_ID)
+        .map(|i| (asset_config_slot(i, config.asset_config_base_slot), i))
+        .collect();
```
die-herdplatte: To avoid this upper bound, keep computations to a minimum, and prevent missing out on books, one could do the following:

- Collect all relevant storage changes first
- If they're non-empty, compute the storage slot for the next asset ID, based on the highest recorded asset ID (needs a store read of some kind)
- If one of the changed slots equals the computed slot, emit the new component and start from the previous step with the next asset ID

This assumes that asset IDs increase by 1.
If delisting should be supported, one would have to add the storage slots of the existing books to the comparison.

protocols/substreams/ethereum-bopamm/src/modules/map_protocol_changes.rs
```diff
@@ -0,0 +1,291 @@
...
+/// Covers the maker being configured *after* the books are created (the live case) and maker
+/// rotation. New books created while the maker is already known get it via
+/// `add_new_components`.
+fn extract_maker_changes(
+    block: &eth::v2::Block,
+    config: &DeploymentConfig,
+    components_store: &StoreGetProto<ProtocolComponent>,
+    transaction_changes: &mut HashMap<u64, TransactionChangesBuilder>,
+) {
+    let books = enumerate_books(components_store);
```
die-herdplatte: This can also be computed lazily

protocols/substreams/ethereum-bopamm/src/modules/map_protocol_changes.rs
```diff
@@ -0,0 +1,291 @@
...
+
+/// Reflects settlement `Paused`/`Unpaused` events onto every book (the venue is paused as a
+/// whole). Paused books should not be routed through until unpaused.
+fn extract_pause_state(
+    block: &eth::v2::Block,
+    config: &DeploymentConfig,
+    components_store: &StoreGetProto<ProtocolComponent>,
+    transaction_changes: &mut HashMap<u64, TransactionChangesBuilder>,
+) {
+    let books = enumerate_books(components_store);
```
die-herdplatte: Same here

protocols/substreams/ethereum-bopamm/src/modules/map_protocol_changes.rs
```diff
@@ -0,0 +1,291 @@
...
+///
+/// This is deliberately over-inclusive: any module storage write (not just config slots)
+/// re-marks all books, which is safe (never misses an update) at the cost of occasional
+/// redundant re-simulation.
+fn mark_books_updated_on_module_changes(
+    config: &DeploymentConfig,
+    components_store: &StoreGetProto<ProtocolComponent>,
+    transaction_changes: &mut HashMap<u64, TransactionChangesBuilder>,
+) {
+    let books = enumerate_books(components_store);
```
die-herdplatte: And here

protocols/adapter-integration/evm/src/bopamm/BopAMMAdapter.sol
```diff
@@ -0,0 +1,345 @@
...
+        if (asset == address(0)) {
+            revert InvalidOrder("Unknown pool");
+        }
+        tokens = new address[](2);
+        tokens[0] = asset;
+        tokens[1] = usdc();
+    }
+
+    /// @inheritdoc ISwapAdapter
+    function getPoolIds(uint256 offset, uint256 limit)
```
die-herdplatte, line 194: Similar argument as in the Substreams package, if the asset IDs are assumed to be sequential, this can be changed to support the case where there are more than `MAX_ASSET_ID` assets and be made more efficient by breaking early.

protocols/substreams/ethereum-bopamm/src/modules/map_relative_balances.rs
```diff
@@ -0,0 +1,53 @@
...
+    let Some(maker_hex) = maker_store.get_last("maker") else {
+        return Ok(BlockBalanceDeltas::default());
+    };
+    let maker = hex::decode(maker_hex)?;
+
+    let mut balance_deltas = Vec::new();
+    for log in block.logs() {
+        let Some(Transfer { from, to, value }) = Transfer::match_and_decode(log.log) else {
+            continue;
+        };
```
zach030: When I was integrating FermiSwap, WETH was a special case here. Users can wrap/unwrap ETH <> WETH without an erc20 Transfer log in the flow we track, so the indexed WETH balance may drift.
Not sure if that’s an issue here too.

protocols/substreams/ethereum-bopamm/src/modules/map_relative_balances.rs
```diff
@@ -0,0 +1,53 @@
...
+        }
+        for comp_id in books {
+            let delta = if to_maker { value.clone() } else { value.clone().neg() };
+            balance_deltas.push(BalanceDelta {
+                ord: log.ordinal(),
+                tx: Some(log.receipt.transaction.into()),
+                token: token.clone(),
+                delta: delta.to_signed_bytes_be(),
+                component_id: comp_id.into_bytes(),
+            });
```
zach030: If the maker already held tokens before `start_block`, the balance may be wrong. We may need to snapshot the balance first, then apply deltas.

protocols/substreams/ethereum-bopamm/src/modules/map_protocol_changes.rs
```diff
@@ -0,0 +1,291 @@
...
+                .entry(tx.index)
+                .or_insert_with(|| TransactionChangesBuilder::new(&tx));
+            balances
+                .values()
+                .for_each(|token_bc_map| {
+                    token_bc_map.values().for_each(|bc| {
+                        builder.add_balance_change(bc);
+                    })
+                });
+        });
```
zach030, line 55: Do we need to upsert the token balance for the `maker`? Also, do we need to call `balanceOf(maker)` during simulation?

protocols/adapter-integration/evm/src/bopamm/BopAMMAdapter.sol
```diff
@@ -0,0 +1,345 @@
...
+    /// construction, so a constructor that called into it would revert.
+    function pricing() public view returns (IBopAmmPricing) {
+        return IBopAmmPricing(settlement.pricing());
+    }
+
+    /// @notice The USDC (hub) token, read from the settlement contract.
+    /// @dev Resolved lazily for the same reason as {pricing}.
+    function usdc() public view returns (address) {
+        return settlement.usdc();
+    }
```
zach030, line 54: Do we need to expose these two functions in the adapter?

protocols/substreams/ethereum-bopamm/src/modules/map_relative_balances.rs
tvinagre: Did you solve it?

protocols/substreams/ethereum-bopamm/src/common.rs
tvinagre: Done in a008a4ac — `enumerate_books` now iterates from id 0 and breaks on the first missing `book:{i}` (unbounded + short-circuit). Safe because asset ids are sequential with no gaps and books are never delisted; that invariant is documented in code, with a note to revisit if delisting is ever added.

protocols/substreams/ethereum-bopamm/src/modules/map_components.rs
tvinagre: Addressed the per-block computation cost in bb14074c — the `MAX_ASSET_ID` slot map is now built lazily on the first module storage change, so the keccak loop is skipped entirely on the (vast majority of) blocks with no asset-config write.

I kept the bounded approach rather than the store-backed incremental discovery you sketched: the venue has 2 assets and the 64 cap is generous headroom, so supporting >64 (which needs a new store + a module-graph change) felt speculative for now. Happy to switch to the unbounded discovery if you'd prefer it.

protocols/substreams/ethereum-bopamm/src/modules/map_protocol_changes.rs
tvinagre: Done in a008a4ac — `extract_maker_changes` now computes the book list lazily via `get_or_insert_with`, only after a relevant maker-slot write is found. Same for the other two below.

protocols/substreams/ethereum-bopamm/src/modules/map_protocol_changes.rs
tvinagre: Done in a008a4ac — `extract_pause_state` now enumerates books lazily, only when a pause/unpause log on the settlement is present.

protocols/substreams/ethereum-bopamm/src/modules/map_protocol_changes.rs
tvinagre: Done in a008a4ac — `mark_books_updated_on_module_changes` enumerates lazily, only when a builder actually touched the module contract.

protocols/adapter-integration/evm/src/bopamm/BopAMMAdapter.sol
tvinagre: Done in 633d9ac3 — `getPoolIds` now `break`s on the first unconfigured asset id (sequential ids, no delisting). Kept `MAX_ASSET_ID` as a loop safety cap and pagination is unchanged. `getPoolIds` gas dropped ~858k → 63k.

protocols/substreams/ethereum-bopamm/src/modules/map_relative_balances.rs
tvinagre: Good catch — it applies here too. Fixed in 5d57584f: `map_relative_balances` now applies WETH `Deposit{dst==maker}` / `Withdrawal{src==maker}` alongside `Transfer`, mirroring your FermiSwap handling, so the maker wrapping/unwrapping ETH no longer drifts the indexed WETH balance.

protocols/substreams/ethereum-bopamm/src/modules/map_relative_balances.rs
tvinagre: Fixed in 5d57584f, same approach as your FermiSwap `map_token_balance_deltas`: on a write to the maker slot (first designation or rotation) we snapshot via `balanceOf` eth_call and re-seed every tracked token by `balanceOf(M_new) - balanceOf(M_old)`; on first designation (`M_old`=0) that seeds whatever the maker already held. New books seed their asset the same way. (On this venue the maker is designated at 25171618, after the books are created at 25171616, so the designation snapshot is what captures the initial inventory.)

One honest caveat documented in code: on a *true* maker rotation the re-seed uses `balanceOf(M_old)` as a proxy for the accumulated store value (the additive store can't be read back inside this module without a graph cycle). It's exact on first designation and effectively exact for these non-rebasing tokens; the venue has never rotated.

protocols/substreams/ethereum-bopamm/src/modules/map_protocol_changes.rs
tvinagre: Two parts:
1. **TVL/balances:** yes — we now seed and maintain the maker's per-book token balances (5d57584f, described above).
2. **Simulation:** no, we do *not* need to index `balanceOf(maker)`. The tycho-simulation VM engine pulls the maker's token balances from RPC on demand when it runs the adapter swap. Verified empirically: the protocol-testing harness simulates both books end-to-end (real amount-out, both directions) without us injecting the maker balance.

protocols/adapter-integration/evm/src/bopamm/BopAMMAdapter.sol
tvinagre: Kept them `public`: they're used internally for lazy resolution (the constructor can't call into the settlement in the simulation deploy context, so `pricing()`/`usdc()` resolve it on demand), and they're also asserted by the fork tests. They're cheap view wrappers, so exposing them is harmless.

protocols/substreams/ethereum-bopamm/src/modules/map_relative_balances.rs
tvinagre: Yes — solved in 5d57584f (balanceOf seeding on maker designation + WETH Deposit/Withdrawal handling). A follow-up code review also caught a same-block create+designate double-count, fixed in the same commit. Harness passes 2/2 with both books simulated.

die-herdplatte (approved):
lgtm!
I'm fine with keeping the `MAX_ASSET_ID` for now.

zizou0x (approved):
Approved in behalf of the integration team

zach030 (approved):
LGTM!

PR #1100, opened 2026-06-18
feat: update vm gas estimations

Troshchk:
## Summary

This PR improves gas estimation accuracy across Curve, Balancer V2, and Uniswap V4.

### Changes
- Curve gas measurement now wraps only the pool exchange() call, avoiding double counting transfer and approval overhead.
- Balancer V2 now pre warms cold storage to account for the difference between revm simulations and real transaction execution.
- Uniswap V4 gas constants were recalibrated for hook calls and tick crossings, improving estimates for hook pools.

#### Curve

Curve estimates now fall within ±25% of on-chain values.
Before this change, Curve estimates were consistently overestimated by 25-50%+.
Screenshot 2026-06-19 at 13 53 36

Screenshot 2026-06-19 at 13 52 43
Screenshot 2026-06-19 at 13 56 17

####  Balancer V2

Half of the Balancer V2 estimates now fall within ±6% of on-chain values.
Overestimates in range 10%+ are not influenced by this change.
Screenshot 2026-06-19 at 13 54 00
Screenshot 2026-06-23 at 09 50 28

Screenshot 2026-06-23 at 09 50 10

####  Uniswap V4 hooks

For uniswap_v4_hooks, severe underestimation improved:
25-50% underestimation: 33% → 17% of swaps
Around 50% of simulations now fall into the 0 to -6% estimation range.
Screenshot 2026-06-19 at 13 54 17

Screenshot 2026-06-19 at 13 52 23
Screenshot 2026-06-19 at 13 58 41

####  Uniswap V4 hookless

For uniswap_v4, there is a slight increase in overestimation, in the 10-25% range.
Screenshot 2026-06-19 at 13 54 37

Screenshot 2026-06-19 at 13 51 42


crates/tycho-simulation/src/evm/simulation.rs
```diff
@@ -325,49 +333,52 @@ fn interpret_evm_success(
     state: EvmState,
     transient_storage: HashMap<Address, HashMap<U256, U256>>,
 ) -> SimulationResult {
+    let mut accessed_slots_count: u64 = 0;
```
Troshchk: This change is needed only to get the `accessed_slots_count`

crates/tycho-simulation/src/evm/protocol/uniswap_v4/state.rs
```diff
@@ -58,10 +58,15 @@ const SWAP_BASE_GAS: u64 = 185_000;
...
-const GAS_PER_TICK: u64 = 17_540;
+// Warm ≈ 10–17k, cold ≈ 40–52k. We use a blended estimate that
+// weights toward cold costs.
+const GAS_PER_TICK: u64 = 29_000;
 // Settlement overhead within swapExactInputSingle: _settle() + _getFullCredit() + misc.
 const V4_CALLBACK_SETTLEMENT_GAS: u64 = 30_000;
+// PoolManager Hooks.sol wrapper overhead per hook call: ABI encode params,
+// external CALL dispatch, decode return, validate selector, and process
+// the returned BeforeSwapDelta / AfterSwapDelta.
+const PM_PER_HOOK_CALL_OVERHEAD: u64 = 25_000;
```
dianacarvalho1, line 69: aaah niceeeee!

crates/tycho-simulation/src/evm/protocol/vm/assets/BalancerV2SwapAdapter.evm.runtime
dianacarvalho1, line 1: why is this different?

crates/tycho-simulation/src/evm/protocol/vm/adapter_contract.rs
```diff
@@ -20,6 +20,18 @@ use crate::evm::{
     simulation::BlockEnvOverrides,
 };
 
+/// Cold SLOAD costs 2100 gas; warm SLOAD costs 100 gas.
+const COLD_WARM_SLOAD_DIFF: u64 = 2000;
```
dianacarvalho1: this is gas right? it should be explicit in the variable name

crates/tycho-simulation/src/evm/protocol/vm/adapter_contract.rs
```diff
@@ -20,6 +20,18 @@ use crate::evm::{
...
+/// Empirically tuned estimate of storage slots accessed outside the adapter's
+/// gas-metered section (pre-swap transfers/approvals and post-swap price
+/// queries). Subtracting these avoids correcting for cold penalties that
+/// were never charged to the adapter's reported gas.
+const NON_METERED_SLOT_ESTIMATE: u64 = 10;
+
+/// Protocols whose adapter gas measurement includes internal token transfers
+/// that the gas estimator also accounts for separately.
```
dianacarvalho1: I think this is not the real reason we need this correction. It's mostly because the protocol uses a contract that is mostly warm on chain for real txs and here in simulations it's always cold. For balancer v2 this is the vault contract

crates/tycho-simulation/src/evm/protocol/vm/adapter_contract.rs
```diff
@@ -20,6 +20,18 @@ use crate::evm::{
     simulation::BlockEnvOverrides,
 };
 
+/// Cold SLOAD costs 2100 gas; warm SLOAD costs 100 gas.
+const COLD_WARM_SLOAD_DIFF: u64 = 2000;
+/// Empirically tuned estimate of storage slots accessed outside the adapter's
+/// gas-metered section (pre-swap transfers/approvals and post-swap price
+/// queries). Subtracting these avoids correcting for cold penalties that
+/// were never charged to the adapter's reported gas.
+const NON_METERED_SLOT_ESTIMATE: u64 = 10;
```
dianacarvalho1: this is a bit not obvious (at least to me on a first glance). Can you give more context on why you decided to implement it this way? Would this even scale nicely to other protocols or is it fine tuned to balancer v2?

dianacarvalho1: I wonder if a better approach couldn't be to actually warm the storage slots before hand. there is [this](https://docs.rs/revm-context-interface/10.2.0/revm_context_interface/journaled_state/trait.JournalTr.html#tymethod.warm_account_and_storage) method in revm Journal. We could warm the vault/pool and tokens before the simulation (not sure how to wire everything from the outside though, I'm just throwing ideas now, not fully baked solutions)

dianacarvalho1 (commented):
Thank you @Troshchk ! I agree with the univ4 and curve changes! For balancer v2 I would like to discuss things a bit more (feel free to break the PR apart if you prefer though)

crates/tycho-simulation/src/evm/protocol/vm/adapter_contract.rs
tamaralipows: I also am not really sure what this means. We were charging cold slot read penalties and not including it in our gas estimations or?

tamaralipows (commented):
Thank you, I understand the UniV4 and _somewhat_ the Curve changes. Maybe we can have a call on Monday so I can understand Balancer better?

crates/tycho-simulation/src/evm/protocol/vm/assets/BalancerV2SwapAdapter.evm.runtime
Troshchk: good catch, it shouldn't be, will regenerate it

crates/tycho-simulation/src/evm/protocol/vm/adapter_contract.rs
dianacarvalho1: The issue is: these simulations happen inside of an revm context - so the vault's storage is always cold :cold_face: which is not true in a real tx. For an example, I saw a case that our revm simulation reported 114k gas and then the tenderly simulation reported only 67k.

Troshchk: Warming it is definitely better -- I thought the only way to do it is to do something like dry run. I did not go that way, because it would increase the simulation time.
Thank you for pointing to the method to warm the account and storage-- will try it

Troshchk: I made the change and it is in the final state. Please, do not review commit by commit, only the final state

crates/tycho-simulation/src/evm/protocol/vm/adapter_contract.rs
Troshchk: I have removed that comment, it was incorrect, thanks for spotting

crates/tycho-simulation/src/evm/simulation.rs
```diff
@@ -152,6 +153,12 @@ where
                         }
                     }
                 }
+                if let Some(overrides) = &params.overrides {
```
dianacarvalho1, line 156: @Troshchk can you give me an example of what are the `params.overrides` like in a balancer v2 swap please? I don't remember

dianacarvalho1: does this have any impact in other vm simulations (curve, univ4 hooks) gas estimations?

Troshchk: I did not observe impact on the other vm simulations gas estimations.

For balancer v2 we override the vault balance for every token in the pool. 
For the sell token, we additionally override the simulated caller’s balance and allowance to the adapter.

dianacarvalho1: > I did not observe impact on the other vm simulations gas estimations.

This sentence can have two meanings: you didn't observe because you didn't run it or you didn't observe because there are actually no differences 😅 Can you clarify which one do you mean please?

Troshchk: true 😅 
I ran it and the signed gas estimation error distribution looks very similar before and after this change. The only significant change I saw due to this change was balancer v2

dianacarvalho1 (approved):
🙏🏼

PR #1103, opened 2026-06-18
feat: Cap integration-test swap inputs using token prices

dianacarvalho1:
Swaps were sized at 0.1% of get_limits, which is near-infinite for tick-based protocols (Uniswap V3/V4), giving unrealistic gas estimates. Cap the input to ~10k USD (in ETH) using the weekly token prices instead.

Add a shared tycho_test::token_prices module to load the S3 price snapshot and cap an amount to a target ETH value. Prices load at startup (fatal on failure) and refresh every 6h so a new weekly dump is picked up without a restart.

### DIFF
--- a/crates/tycho-integration-test/src/main.rs
+++ b/crates/tycho-integration-test/src/main.rs
@@ -27,7 +27,7 @@ use tokio::{signal, sync::Semaphore};
 use tracing::{debug, error, info, warn};
 use tracing_subscriber::EnvFilter;
 use tycho_client::feed::SynchronizerState;
-use tycho_common::simulation::protocol_sim::ProtocolSim;
+use tycho_common::{simulation::protocol_sim::ProtocolSim, Bytes};
 use tycho_simulation::{
     evm::protocol::cowamm::constants::PROTOCOL_SYSTEM as COWAMM_PROTOCOL_SYSTEM,
     protocol::models::ProtocolComponent,
@@ -44,6 +44,7 @@ use tycho_test::{
         models::{TychoExecutionInput, TychoExecutionResult},
         simulate_swap_transaction, tenderly,
     },
+    token_prices::{cap_amount_to_eth_value, load_token_prices},
     validation::{batch_validate_components, get_validator, Validator},
 };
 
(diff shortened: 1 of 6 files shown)


crates/tycho-test/src/token_prices.rs
```diff
@@ -0,0 +1,144 @@
+//! Loads token prices and uses them to bound swap input amounts in tests.
```
dianacarvalho1, line 1: decided to add this in tycho-test in case someone else wants to use these functions somewhere else

crates/tycho-integration-test/src/main.rs
```diff
@@ -193,6 +194,19 @@ struct TychoState {
     component_ids_by_protocol: HashMap<String, HashSet<String>>,
 }
 
+/// Shared, periodically-refreshed token-price snapshot (raw token units per ETH).
+type SharedTokenPrices = Arc<RwLock<Arc<HashMap<Bytes, f64>>>>;
+
+/// How often to reload the token-price snapshot. The S3 dump is refreshed weekly; reloading every
+/// few hours picks up a new dump without restarting this long-running binary.
+const TOKEN_PRICE_REFRESH_INTERVAL: Duration = Duration::from_secs(6 * 60 * 60);
```
tamaralipows: If the dump is weekly, not sure I understand why it makes sense to load every 6 hours?

tamaralipows (commented):
Thanks Diana! I think I'm missing a tad bit of context. We already have weekly token prices saved to s3? Where is this happening?

crates/tycho-integration-test/src/main.rs
dianacarvalho1: I dunno when in the week the dump happens 😂 do you think we should try to sync it?

dianacarvalho1:
> Thanks Diana! I think I'm missing a tad bit of context. We already have weekly token prices saved to s3? Where is this happening?

yes we already have token prices saved weekly to s3. I'm not sure but my guess is the NTQ? @louise-poole can you confirm please?

crates/tycho-integration-test/src/main.rs
tamaralipows: Meh I just thought 6 hours was overkill but it probably doesn't make that big of a different runtime wise so its ok

tamaralipows (approved):
Thank you! Code wise this looks good, only would like more information about the saved token prices eventually

crates/tycho-integration-test/src/main.rs
louise-poole: The dump happens every Monday around midday.

louise-poole: It's probably good enough to only refresh every 24hrs? It's not super critical to have the absolute latest prices here...

louise-poole (approved):
Looks good to me!

PR #1104, opened 2026-06-19
fix(tycho-simulation): Apply pool blocklist by default

dianacarvalho1:
Done:
- ProtocolStreamBuilder now uses the default blocklist in new(); blocklist_components() adds to it. 
- Drop the now-redundant blocklist loading from the quickstart example
- Document the default in the solver simulation docs.

### DIFF
--- a/crates/tycho-simulation/examples/quickstart/main.rs
+++ b/crates/tycho-simulation/examples/quickstart/main.rs
@@ -59,7 +59,7 @@ use tycho_simulation::{
     protocol::models::{ProtocolComponent, Update},
     tycho_client::feed::component_tracker::ComponentFilter,
     tycho_common::{dto::TvlThresholdTier, models::Chain},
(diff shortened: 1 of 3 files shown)


crates/tycho-simulation/examples/quickstart/main.rs
```diff
@@ -248,10 +245,6 @@ async fn main() {
         _ => {}
     }
 
-    let blocklist =
-        load_blocklist(cli.blocklist_file.as_deref()).expect("Failed to load blocklist");
-    protocol_stream = protocol_stream.blocklist_components(blocklist);
-
     let protocol_stream = protocol_stream
```
Troshchk, line 248: It looks somewhat weird before the change 😬

Troshchk (approved):
Thank you @dianacarvalho1! 💫

PR #1106, opened 2026-06-19
feat: live pAMM state-override stream integration

zizou0x:
## Summary

Adds a generic, protocol-agnostic layer in `tycho-simulation` that injects live per-block VM storage overrides and block environment into pAMM pools, sourced from the Titan pAMM quote stream. Consumers keep using `ProtocolStreamBuilder`; the Titan provider auto-registers when a known pAMM venue is in the stream.

Jira: [ENG-6118](https://propeller-heads.atlassian.net/browse/ENG-6118)

## How it works

- A `StateOverrideProvider` maintains a live `watch` channel of `OverrideSnapshot`s (block number/timestamp + storage) per `protocol_system`.
- The matching receiver is wired into each `EVMPoolState` at construction (via `DecoderContext`), so there is no per-block re-attachment or downcasting.
- The pool reads the freshest snapshot at simulation time (`get_amount_out`, `get_limits`, spot prices), so sub-block oracle price changes are reflected without waiting for a Tycho block update.
- Live storage overrides and block environment take precedence over the pool's static values on conflict, and are dropped once the snapshot passes its provider-set TTL — reverting the pool to Tycho's indexed state, so a stalled stream never keeps serving stale prices.

## Registration

- Providers are registered per `protocol_system`. A default registry supplies built-in providers (Titan) for any pAMM venue that is a registered exchange and not already claimed by the consumer.
- `ProtocolStreamBuilder::with_override_provider(protocol_system, provider)` swaps a single protocol onto a custom provider; explicit registrations win over the default registry.
- A multi-venue provider is shared via `Arc`, so one Titan connection serves all its venues.

## Titan provider

- `TitanProvider` connects to the Titan pAMM quote stream over WebSocket, parses each message into per-venue `OverrideSnapshot`s, and publishes them on the matching channel, reconnecting with capped exponential backoff.
- Serves FermiSwap, Kipseli and bopAMM. The block timestamp is resolved from the freshest registry lane timestamp so the oracle-staleness guard treats streamed prices as current.
- `tokio-tungstenite` moved to the `evm` feature, since the override stream is an EVM concern, not an RFQ one.

## Changes

- `evm/override_stream/` (new): `OverrideSnapshot`, `StateOverrideProvider`, the default registry, and `TitanProvider`.
- `evm/protocol/vm/state.rs`: `EVMPoolState` live read-through (`live_overrides`, live-aware `get_overwrites` and block env).
- `evm/protocol/vm/decoder.rs` + `protocol/models.rs`: receiver wired in at construction through `DecoderContext`.
- `evm/decoder.rs`: per-`protocol_system` provider registry.
- `evm/stream.rs`: `with_override_provider`, default-registry install at build time.

## Test plan

- [x] Unit tests for message parsing, lane-timestamp resolution, venue mapping, registration, snapshot expiry, and live-override resolution in pool state (block-env precedence + storage merge).
- [x] Validated live mid-block price movement against tycho-dev with FermiSwap (prices update between blocks, faster than block cadence).

[ENG-6118]: https://propeller-heads.atlassian.net/browse/ENG-6118?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/tycho-simulation/Cargo.toml
+++ b/crates/tycho-simulation/Cargo.toml
@@ -113,6 +113,7 @@ evm = [
     "dep:revm-inspectors",
     "dep:alloy",
     "dep:reqwest",
+    "dep:tokio-tungstenite",
 ]
 rfq = [
     "dep:reqwest",
--- a/crates/tycho-simulation/src/evm/override_stream/mod.rs
+++ b/crates/tycho-simulation/src/evm/override_stream/mod.rs
@@ -43,6 +43,22 @@ pub struct OverrideSnapshot {
     /// can read the latest snapshot on every simulation with a refcount bump rather than a deep
     /// clone of the nested map.
     pub storage: Arc<HashMap<Address, HashMap<U256, U256>>>,
+    /// Unix-seconds instant after which these overrides are stale and must not be used. Set by the
+    /// provider from its own freshness rules (e.g. Titan uses the quote timestamp plus one block
+    /// time). `None` means the snapshot never expires — e.g. the empty initial snapshot.
+    pub expires_at: Option<u64>,
+}
+
+impl OverrideSnapshot {
+    /// Returns whether these overrides are stale at `now` (unix seconds).
+    ///
+    /// A snapshot without an [`expires_at`](Self::expires_at) never expires. Callers that read the
+    /// snapshot on every simulation should skip it entirely once this returns `true`, falling back
+    /// to whatever state they held before.
+    pub fn is_expired(&self, now: u64) -> bool {
+        self.expires_at
+            .is_some_and(|expires_at| now >= expires_at)
+    }
 }
 
 /// Supplies a *live* stream of resolved per-block VM overrides for one or more protocols.
@@ -72,4 +88,23 @@ pub(crate) fn default_override_providers(
     let mut providers: HashMap<String, Arc<dyn StateOverrideProvider>> = HashMap::new();
     providers.extend(titan::default_providers(protocols));
     providers
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn snapshot_without_expiry_never_expires() {
+        let snapshot = OverrideSnapshot::default();
+        assert!(!snapshot.is_expired(u64::MAX));
+    }
+
+    #[test]
+    fn snapshot_expires_at_or_after_deadline() {
+        let snapshot = OverrideSnapshot { expires_at: Some(100), ..Default::default() };
+        assert!(!snapshot.is_expired(99));
+        assert!(snapshot.is_expired(100));
+        assert!(snapshot.is_expired(101));
+    }
 }
--- a/crates/tycho-simulation/src/evm/override_stream/mod.rs
+++ b/crates/tycho-simulation/src/evm/override_stream/mod.rs
@@ -7,9 +7,11 @@
 //! block.
 //!
 //! This module defines the generic plumbing:
-//! - [`OverrideSnapshot`] — the resolved overrides for one protocol at a point in time.
-//! - [`StateOverrideProvider`] — a source that maintains a *live* [`watch`] channel of snapshots
-//!   per protocol (kept fresh in the background, e.g. from a WebSocket).
+//! - [`OverrideSnapshot`](crate::evm::override_stream::OverrideSnapshot) — the resolved overrides
+//!   for one protocol at a point in time.
+//! - [`StateOverrideProvider`](crate::evm::override_stream::StateOverrideProvider) — a source that
+//!   maintains a *live* [`watch`](tokio::sync::watch) channel of snapshots per protocol (kept fresh
+//!   in the background, e.g. from a WebSocket).
 //!
 //! Providers are registered per `protocol_system` (see
 //! [`ProtocolStreamBuilder::with_override_provider`](crate::evm::stream::ProtocolStreamBuilder::with_override_provider)).
@@ -17,7 +19,7 @@
 //! freshest snapshot at simulation time.
 //!
 //! It deliberately knows nothing about any specific venue or stream format; all such specifics live
-//! in the concrete provider implementations (see [`titan`]).
+//! in the concrete provider implementations (see [`titan`](crate::evm::override_stream::titan)).
 
 pub mod titan;
 
@@ -73,4 +75,3 @@ pub(crate) fn default_override_providers(
     providers.extend(titan::default_providers(registered_exchanges, covered));
     providers
 }
-
--- a/crates/tycho-simulation/src/evm/override_stream/mod.rs
+++ b/crates/tycho-simulation/src/evm/override_stream/mod.rs
@@ -23,10 +23,7 @@
 
 pub mod titan;
 
-use std::{
-    collections::{HashMap, HashSet},
-    sync::Arc,
-};
+use std::{collections::HashMap, sync::Arc};
 
 use alloy::primitives::{Address, U256};
 use tokio::sync::watch;
@@ -64,16 +61,15 @@ pub trait StateOverrideProvider: Send + Sync {
 
 /// The default registry of built-in override providers, keyed by `protocol_system`.
 ///
-/// Given the `registered_exchanges` and the set of protocols already `covered` by explicit
-/// consumer registrations, returns the providers that should serve the remaining protocols. This
-/// is the single place that wires concrete providers (e.g. the Titan pAMM stream) into the
-/// otherwise protocol-agnostic stream builder, keeping all venue-specific knowledge out of both the
-/// builder and the generic override core. Protocols in `covered` are left to the consumer.
+/// `protocols` yields the protocol systems the caller wants a built-in default for (i.e. registered
+/// exchanges minus any the consumer explicitly overrode — the builder computes that difference and
+/// forwards it as an owned iterator, avoiding clones). This is the single place that wires concrete
+/// providers (e.g. the Titan pAMM stream) into the otherwise protocol-agnostic stream builder,
+/// keeping all venue-specific knowledge out of both the builder and the generic override core.
 pub(crate) fn default_override_providers(
-    registered_exchanges: &[String],
-    covered: &HashSet<String>,
+    protocols: impl IntoIterator<Item = String>,
 ) -> HashMap<String, Arc<dyn StateOverrideProvider>> {
     let mut providers: HashMap<String, Arc<dyn StateOverrideProvider>> = HashMap::new();
-    providers.extend(titan::default_providers(registered_exchanges, covered));
+    providers.extend(titan::default_providers(protocols));
     providers
 }
--- a/crates/tycho-simulation/src/evm/override_stream/mod.rs
+++ b/crates/tycho-simulation/src/evm/override_stream/mod.rs
@@ -1,0 +1,76 @@
+//! Generic, protocol-agnostic core for injecting live per-block VM state overrides into pools.
+//!
+//! Some protocols (e.g. pAMMs such as FermiSwap) rely on off-chain oracle prices that are not part
+//! of the on-chain state Tycho indexes, and that change *within* a block. To simulate them
+//! accurately, a side channel publishes per-block VM storage overrides and an overridden block
+//! environment that must be reflected by the pool on every simulation — not just once per Tycho
+//! block.
+//!
+//! This module defines the generic plumbing:
+//! - [`OverrideSnapshot`] — the resolved overrides for one protocol at a point in time.
+//! - [`StateOverrideProvider`] — a source that maintains a *live* [`watch`] channel of snapshots
+//!   per protocol (kept fresh in the background, e.g. from a WebSocket).
+//!
+//! Providers are registered per `protocol_system` (see
+//! [`ProtocolStreamBuilder::with_override_provider`](crate::evm::stream::ProtocolStreamBuilder::with_override_provider)).
+//! A pool subscribes to the provider registered for its protocol at creation time and reads the
+//! freshest snapshot at simulation time.
+//!
+//! It deliberately knows nothing about any specific venue or stream format; all such specifics live
+//! in the concrete provider implementations (see [`titan`]).
+
+pub mod titan;
+
+use std::{
+    collections::{HashMap, HashSet},
+    sync::Arc,
+};
+
+use alloy::primitives::{Address, U256};
+use tokio::sync::watch;
+
+/// Resolved per-block VM overrides for a single protocol at a point in time.
+///
+/// `block_number` and `block_timestamp` are already resolved by the provider (e.g. Titan computes
+/// `block_timestamp` from lane timestamps) so this core stays protocol-agnostic. Either field may
+/// be `None`, in which case the pool's existing block environment is left intact.
+#[derive(Clone, Default, Debug)]
+pub struct OverrideSnapshot {
+    /// The L1 block number these overrides apply to, if known.
+    pub block_number: Option<u64>,
+    /// The block timestamp to inject into the pool's simulation environment, if known.
+    pub block_timestamp: Option<u64>,
+    /// Storage overrides keyed by contract address, then by storage slot.
+    pub storage: HashMap<Address, HashMap<U256, U256>>,
+}
+
+/// Supplies a *live* stream of resolved per-block VM overrides for one or more protocols.
+///
+/// Implementations maintain their snapshots in the background (e.g. from a WebSocket stream) and
+/// expose them through a [`watch`] channel per protocol. A single provider may serve several
+/// protocols sharing one underlying connection; the same provider can be registered for each of
+/// them (it is shared via `Arc`, not duplicated).
+pub trait StateOverrideProvider: Send + Sync {
+    /// Returns the live override channel for `protocol_system`, or `None` if unsupported.
+    ///
+    /// The returned [`watch::Receiver`] always reflects the latest snapshot; reading it
+    /// ([`watch::Receiver::borrow`]) never blocks, so a pool may read it on every simulation.
+    fn subscribe(&self, protocol_system: &str) -> Option<watch::Receiver<OverrideSnapshot>>;
+}
+
+/// The default registry of built-in override providers, keyed by `protocol_system`.
+///
+/// Given the `registered_exchanges` and the set of protocols already `covered` by explicit
+/// consumer registrations, returns the providers that should serve the remaining protocols. This
+/// is the single place that wires concrete providers (e.g. the Titan pAMM stream) into the
+/// otherwise protocol-agnostic stream builder, keeping all venue-specific knowledge out of both the
+/// builder and the generic override core. Protocols in `covered` are left to the consumer.
+pub(crate) fn default_override_providers(
+    registered_exchanges: &[String],
+    covered: &HashSet<String>,
+) -> HashMap<String, Arc<dyn StateOverrideProvider>> {
+    let mut providers: HashMap<String, Arc<dyn StateOverrideProvider>> = HashMap::new();
+    providers.extend(titan::default_providers(registered_exchanges, covered));
+    providers
+}
+
--- a/crates/tycho-simulation/src/evm/override_stream/mod.rs
+++ b/crates/tycho-simulation/src/evm/override_stream/mod.rs
@@ -42,8 +42,10 @@ pub struct OverrideSnapshot {
     pub block_number: Option<u64>,
     /// The block timestamp to inject into the pool's simulation environment, if known.
     pub block_timestamp: Option<u64>,
-    /// Storage overrides keyed by contract address, then by storage slot.
-    pub storage: HashMap<Address, HashMap<U256, U256>>,
+    /// Storage overrides keyed by contract address, then by storage slot. `Arc`-wrapped so a pool
+    /// can read the latest snapshot on every simulation with a refcount bump rather than a deep
+    /// clone of the nested map.
+    pub storage: Arc<HashMap<Address, HashMap<U256, U256>>>,
 }
 
 /// Supplies a *live* stream of resolved per-block VM overrides for one or more protocols.
--- a/crates/tycho-simulation/src/evm/override_stream/titan.rs
+++ b/crates/tycho-simulation/src/evm/override_stream/titan.rs
@@ -345,7 +345,7 @@ impl TitanProvider {
         // as current.
         let block_timestamp = Self::max_lane_timestamp(&storage);
 
-        Ok(Some(OverrideSnapshot { block_number, block_timestamp, storage }))
+        Ok(Some(OverrideSnapshot { block_number, block_timestamp, storage: Arc::new(storage) }))
     }
 
     /// Returns the newest lane update timestamp (in seconds) across all overridden slots.
--- a/crates/tycho-simulation/src/evm/override_stream/titan.rs
+++ b/crates/tycho-simulation/src/evm/override_stream/titan.rs
@@ -72,7 +72,8 @@ pub fn default_providers(
     if needed.is_empty() {
         return HashMap::new();
     }
-    let provider: Arc<dyn StateOverrideProvider> = Arc::new(TitanProvider::spawn(TITAN_URL.to_string()));
+    let provider: Arc<dyn StateOverrideProvider> =
+        Arc::new(TitanProvider::spawn(TITAN_URL.to_string()));
     needed
         .into_iter()
         .map(|protocol| (protocol.to_string(), provider.clone()))
@@ -320,7 +321,9 @@ mod tests {
 
     #[test]
     fn parse_message_extracts_venue_state_diff() {
-        let venue = FERMISWAP_VENUE.parse::<Address>().unwrap();
+        let venue = FERMISWAP_VENUE
+            .parse::<Address>()
+            .unwrap();
         let snapshot = TitanProvider::parse_message(SAMPLE_MESSAGE, &venue)
             .expect("parse must succeed")
             .expect("venue is present");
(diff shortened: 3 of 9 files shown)


crates/tycho-simulation/src/evm/stream.rs
```diff
@@ -636,6 +650,39 @@ impl ProtocolStreamBuilder {
...
+
+    /// Installs override providers onto the decoder before the stream is built.
+    ///
+    /// Explicit consumer registrations win; the built-in default registry (see
+    /// [`default_override_providers`](crate::evm::override_stream::default_override_providers))
+    /// then fills every remaining protocol it can serve.
+    fn install_override_providers(&mut self) {
+        let explicit = std::mem::take(&mut self.override_providers);
+        let covered: HashSet<String> = explicit.keys().cloned().collect();
+        let defaults = crate::evm::override_stream::default_override_providers(
```
brunoguerios: Seems like you're already importing override_stream, so the inline import is not really necessary here, right?
```suggestion
        let defaults = override_stream::default_override_providers(
```

crates/tycho-simulation/src/evm/override_stream/titan.rs
```diff
@@ -0,0 +1,450 @@
...
+    }
+
+    /// Returns the newest lane update timestamp (in seconds) across all overridden slots.
+    ///
+    /// FermiSwap registry lanes pack a uint32 update timestamp in the first 4 bytes of the slot
+    /// value. This extracts that prefix from every overridden slot and returns the maximum value
+    /// that looks like a plausible unix timestamp, or `None` if there are none. The result is used
+    /// as the resolved `block_timestamp` so the registry's staleness guard treats the streamed
+    /// oracle prices as current.
+    fn max_lane_timestamp(overrides: &VenueOverrides) -> Option<u64> {
```
brunoguerios, line 378: This one feels hacky 😅 
Just as an exercise, what happens if an incorrect timestamp gets returned here?

brunoguerios (commented):
Nice one 👏 

Even without much context it was somewhat straightforward to follow the implementation.

One thing that came to my mind is that I didn't see `chain` being mentioned anywhere. This applies to Base only, right? Should we mention it somewhere?

crates/tycho-simulation/src/evm/override_stream/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+///
+/// `block_number` and `block_timestamp` are already resolved by the provider (e.g. Titan computes
+/// `block_timestamp` from lane timestamps) so this core stays protocol-agnostic. Either field may
+/// be `None`, in which case the pool's existing block environment is left intact.
+#[derive(Clone, Default, Debug)]
+pub struct OverrideSnapshot {
+    /// The L1 block number these overrides apply to, if known.
+    pub block_number: Option<u64>,
+    /// The block timestamp to inject into the pool's simulation environment, if known.
+    pub block_timestamp: Option<u64>,
```
tvinagre, line 41: Should we add a `updated_at` timestamp? Might be useful to identify stale streams

crates/tycho-simulation/src/evm/override_stream/titan.rs
```diff
@@ -0,0 +1,450 @@
...
+                .iter()
+                .any(|e| e.as_str() == protocol) &&
+                !covered.contains(protocol)
+        })
+        .collect();
+    if needed.is_empty() {
+        return HashMap::new();
+    }
+    let provider: Arc<dyn StateOverrideProvider> =
+        Arc::new(TitanProvider::spawn(TITAN_URL.to_string()));
```
tvinagre: Potentially could be overridable

crates/tycho-simulation/src/evm/protocol/vm/state.rs
```diff
@@ -230,7 +278,7 @@ where
                         buy_token_address,
                         vec![sell_amount_limit / U256::from(100)],
                         overwrites,
-                        self.block_overrides.clone(),
+                        self.effective_block_overrides(),
```
tvinagre: I think we have a conceptual problem here.

For each simulation, we're reading the watch channel at least twice per simulation at different points. This exposes one risk that is quite clear, specially on set_spot_prices, which is the price changing intra-simulation.

If the Titan background task publishes a new snapshot between these borrows, a single simulation runs with storage from snapshot A but block number/timestamp from snapshot B. Worse, in this case, get_spot_price runs two sub-simulations (x1, x2) that each re-borrow, so even the two halves of one price calc can see different block envs. Since the resolved block_timestamp is specifically engineered to satisfy the pool's oracle-staleness guard against the lane timestamps in the storage override, a mismatch can spuriously trip that guard or produce an inconsistent quote. 

Maybe it would be better to borrow once at the top of each simulation and thread that single OverrideSnapshot into both the overwrites and the block-env construction.

crates/tycho-simulation/src/evm/override_stream/titan.rs
```diff
@@ -0,0 +1,450 @@
...
+                            // connection is healthy, so reset the reconnect backoff.
+                            Ok(Message::Text(text)) => {
+                                attempt = 0;
+                                for (venue, sender) in &senders {
+                                    match Self::parse_message(text.as_str(), venue) {
+                                        Ok(Some(snapshot)) => {
+                                            // A send error here only means every receiver (the
+                                            // provider handle and all pools) has been dropped, i.e.
+                                            // nothing consumes overrides anymore; safe to ignore.
+                                            let _ = sender.send(snapshot);
```
tvinagre: Safe to ignore, but we'd have a background task running that doesn't produce any value. We should either kill the bg task or, at least, raise a warning

crates/tycho-simulation/src/evm/override_stream/titan.rs
tvinagre: yeah - agree it feels very hacky. I don't know Titans internals to verify if there's a risk, but I'd advise we double check and test this well

crates/tycho-simulation/Cargo.toml
```diff
@@ -113,6 +113,7 @@ evm = [
     "dep:revm-inspectors",
     "dep:alloy",
     "dep:reqwest",
+    "dep:tokio-tungstenite",
```
tvinagre, line 116: Claudio flagged this:
crates/tycho-simulation/Cargo.toml declares tokio-tungstenite = "0.28.0" (with its own native-tls) instead of workspace = true, while the workspace root and tycho-client/tycho-indexer use 0.20.0. Cargo.lock now contains both 0.20.1 and 0.28.0 (and the
  duplicated tungstenite/TLS stacks). This bloats the build and breaks the workspace-dependency convention. Either bump the workspace pin to 0.28 and migrate the other crates, or reuse 0.20.

crates/tycho-simulation/src/evm/override_stream/titan.rs
```diff
@@ -0,0 +1,450 @@
...
+
+    /// Extracts a single venue's snapshot (`stateDiff` overrides + resolved block environment)
+    /// from a Titan stream message.
+    ///
+    /// Returns `Ok(None)` if the venue is absent from the message. The block timestamp is resolved
+    /// from the freshest lane update timestamp (see
+    /// [`max_lane_timestamp`](Self::max_lane_timestamp)) so the pool's oracle-staleness guard
+    /// sees the overridden prices as current.
+    fn parse_message(text: &str, venue: &Address) -> Result<Option<OverrideSnapshot>, String> {
+        let root: Value = serde_json::from_str(text).map_err(|e| format!("invalid JSON: {e}"))?;
```
tvinagre: This seems inneficcient. We're parsing the same JSON 3 times just to check if the override matches a venue? Shouldn't we deserialize it once and then pass the obj to this  function?

crates/tycho-simulation/src/evm/protocol/vm/state.rs
```diff
@@ -143,9 +152,48 @@ where
...
+        self.live_overrides = Some(receiver);
+    }
+
+    /// Storage overwrites from the live override channel, if one is attached and has data.
+    fn live_storage_overwrites(&self) -> Option<HashMap<Address, Overwrites>> {
+        let snapshot = self.live_overrides.as_ref()?.borrow();
+        if snapshot.storage.is_empty() {
+            None
+        } else {
+            Some(snapshot.storage.clone())
```
tvinagre: claudinho pointed:
live_storage_overwrites() does snapshot.storage.clone() (a nested HashMap<Address, HashMap<U256,U256>>) on every simulation, then merge() clones again. For a solver issuing many quotes this is repeated allocation per quote. Consider Arc-wrapping the
snapshot storage so reads are cheap.

crates/tycho-simulation/src/evm/override_stream/titan.rs
```diff
@@ -0,0 +1,450 @@
...
+    /// Connects, reads messages, and publishes each parsed [`OverrideSnapshot`] on the matching
+    /// venue channel. Any disconnect, read error, server-side close, or idle timeout drops back to
+    /// the reconnect loop, which retries forever with capped exponential backoff. The backoff is
+    /// reset only after a message is actually received, so a socket that connects and immediately
+    /// drops still backs off instead of busy-looping. This task never returns and never panics:
+    /// every failure mode is logged and retried.
+    async fn run(url: String, senders: Vec<(Address, watch::Sender<OverrideSnapshot>)>) {
+        let mut attempt: u32 = 0;
+        loop {
+            match connect_async(url.as_str()).await {
```
tvinagre: no timeout for the connection - can be blocked forever

crates/tycho-simulation/src/evm/protocol/vm/state.rs
```diff
@@ -143,9 +152,48 @@ where
...
         }
     }
 
+    /// The block environment to apply to adapter simulations, preferring the live snapshot's
+    /// resolved block number/timestamp over the statically configured [`Self::block_overrides`].
+    fn effective_block_overrides(&self) -> Option<BlockEnvOverrides> {
```
tvinagre: missing tests

crates/tycho-simulation/src/evm/protocol/vm/state.rs
```diff
@@ -143,9 +152,48 @@ where
...
+            return base;
+        }
+        let mut overrides = base.unwrap_or_default();
+        if snapshot.block_number.is_some() {
+            overrides.number = snapshot.block_number;
+        }
+        if snapshot.block_timestamp.is_some() {
+            overrides.timestamp = snapshot.block_timestamp;
+        }
+        Some(overrides)
```
zach030: do we need to add a check for this? if titan stops streaming for whatever reason but we're still getting a fresh `override_block_timestamp` from substreams, we should use the timestamp from substreams instead of the stale timestamp from titan.

zach030:
I'm wondering if we should add another example showing how to simulate locally with frequent Titan stream updates.
Right now, `price_printer` and `integration_test` only trigger simulation when they get a new Tycho update. Since Tycho updates are usually block-driven, there's often a gap between simulations, and during that gap the Titan watch channel just keeps the latest snapshot, so we basically skip over anything in between.

crates/tycho-simulation/Cargo.toml
zizou0x: True but not related to this PR so I'd prefer to treat this as a different problem. The correct fix would be to update to 0.28.0 in the workspace

crates/tycho-simulation/src/evm/override_stream/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+/// The default registry of built-in override providers, keyed by `protocol_system`.
+///
+/// Given the `registered_exchanges` and the set of protocols already `covered` by explicit
+/// consumer registrations, returns the providers that should serve the remaining protocols. This
+/// is the single place that wires concrete providers (e.g. the Titan pAMM stream) into the
+/// otherwise protocol-agnostic stream builder, keeping all venue-specific knowledge out of both the
+/// builder and the generic override core. Protocols in `covered` are left to the consumer.
+pub(crate) fn default_override_providers(
+    registered_exchanges: &[String],
+    covered: &HashSet<String>,
```
die-herdplatte: Why should this function (and any of its callees) receive `registered_exchanges` and `covered` when they're only meant to add providers for `registered_exchanges - covered`?
imo this subtraction should be done by the `ProtocolStreamBuilder` and the result passed to this function.

crates/tycho-simulation/src/evm/override_stream/titan.rs
```diff
@@ -0,0 +1,450 @@
...
+impl TitanProvider {
+    /// Opens ONE WebSocket connection to Titan serving all known pAMM venues and spawns the
+    /// background reconnect/parse task that keeps the per-protocol snapshots up to date.
+    ///
+    /// Returns immediately; the connection is established and maintained in the background, so a
+    /// transient WebSocket failure never blocks or crashes the caller.
+    pub fn spawn(url: String) -> Self {
+        let mut receivers = HashMap::new();
+        let mut senders = Vec::new();
+        for &protocol in known_pamm_protocols() {
```
die-herdplatte: Shouldn't this only iterator over the `needed` protocols?

crates/tycho-simulation/src/evm/override_stream/titan.rs
```diff
@@ -0,0 +1,450 @@
...
+                            // keep the (otherwise healthy) connection rather than reconnecting.
+                            Ok(Message::Binary(bytes)) => {
+                                warn!(len = bytes.len(), "Ignoring unexpected binary Titan frame");
+                            }
+                            // Keep-alive frames. tokio-tungstenite answers pings with pongs
+                            // automatically while the stream is polled, so there is nothing to do.
+                            Ok(Message::Ping(_)) | Ok(Message::Pong(_)) => {}
+                            // Raw frames only surface with non-default tungstenite settings;
+                            // ignore.
+                            Ok(Message::Frame(_)) => {}
```
die-herdplatte: It seems like this is only ever emitted on writes, not reads.
So the comment is slightly wrong.

crates/tycho-simulation/src/evm/override_stream/titan.rs
die-herdplatte: I also think it would be good to use a struct with a derived `Deserialize` for the remainder of this function to avoid the pain of operating with `serde_json::value::Value`

crates/tycho-simulation/src/evm/override_stream/titan.rs
die-herdplatte: Another potential solution:

Use the "slot" property on the root JSON object to derive the block timestamp.
Slots are always 12 seconds apart and we can compute the timestamp of the pending block with `genesisTimestamp + slot * 12` ([`genesisTimestamp` is 1606824023](https://beaconcha.in/slot/0)).

It doesn't necessarily equal the timestamp committed to the `PrioUpdateRegistry` but if the [timestamp check](https://github.com/flashbots/priority-update-registry/blob/da53117870c7bec96d71caebe1b3f94370aba3d6/src/PrioUpdateRegistry.sol#L99) would fail with the slot-derived (=exact) timestamp it would mean that the pAMM streams stale data to Titan.

This would also only work on mainnet, but that should be fine.

die-herdplatte: If this solution is kept, I'd harden it:
- filter state overrides to the `PrioUpdateRegistry` (afaik FermiSwap, BopAMM, and Kipseli all target it)
- the plausibility check [only reliably works if the pAMM doesn't write more than one slot](https://github.com/flashbots/priority-update-registry/blob/da53117870c7bec96d71caebe1b3f94370aba3d6/src/PrioUpdateRegistry.sol#L160)

The `PrioUpdateRegistry` writes to persistent storage in two occasions:
- for tracking updaters `mapping(address target => mapping(address updater => bool)) public isUpdater`
- for the state updates in `updateState`

The first type of storage write can be filtered out since the set of permissible values is just [0,1].
For the second type, one could filter out all slots where `slot - 1` has also been written to in this state diff (only the first slot of an update contains the timestamp).

crates/tycho-simulation/src/evm/protocol/vm/state.rs
```diff
@@ -466,9 +514,14 @@ where
         let token_overwrites = self.get_token_overwrites(tokens, max_amount)?;
 
         // Merge `block_lasting_overwrites` with `token_overwrites`
-        let merged_overwrites =
+        let mut merged_overwrites =
             self.merge(&self.block_lasting_overwrites.clone(), &token_overwrites);
 
+        // Live overrides (e.g. Titan pAMM oracle state) take precedence on conflict.
+        if let Some(live_overwrites) = self.live_storage_overwrites() {
+            merged_overwrites = self.merge(&merged_overwrites, &live_overwrites);
```
die-herdplatte: Theoretically could take ownership of both arguments and avoid multiple `HashMap` clones.

crates/tycho-simulation/src/evm/stream.rs
```diff
@@ -234,6 +235,13 @@ pub struct ProtocolStreamBuilder {
     /// Receiver half of the trigger channel. Held here until `build()` / `build_with_pending()`
     /// transfers ownership to the gating task. `Some` iff step-control mode is active.
     step_trigger_rx: Option<tokio::sync::mpsc::UnboundedReceiver<()>>,
+    /// State-override providers explicitly registered by the consumer, keyed by `protocol_system`.
+    /// These take precedence over the built-in default registry and are installed onto the decoder
+    /// at build time.
+    override_providers: HashMap<String, Arc<dyn StateOverrideProvider>>,
+    /// Names of all exchanges registered on the builder, used to decide which built-in override
+    /// providers to auto-register at build time.
+    registered_exchanges: Vec<String>,
```
die-herdplatte: `registered_exchanges` allows duplicates compared to the other properties which use `HashMap`s.
A repeated call to `exchange` or `exchange_with_decoder_context` inserts the exchange multiple times.
Not a big deal, it doesn't affect behavior, but is more consistent imo.

crates/tycho-simulation/src/evm/override_stream/titan.rs
zizou0x: @die-herdplatte could you elaborate on the pain of operating with `serde_json::value::Value`?

Parsing the JSON once is easily doable but it's hard to completely avoid using `serde_json::value::Value` because they use addresses as top level field and we want to be future proof (not break if they add a new field). 

I've implemented a middle ground that deserialize into DTO objects but still relies on `Value` internally. The only as future proof alternative that I found would be this but I didn't like it. 

```rust
struct TitanMessage {
    #[serde(rename = "blockNumber", default)] block_number: Option<u64>,
    #[serde(flatten)] venues: HashMap<String, VenueEntry>,
}
#[serde(untagged)] enum VenueEntry { Venue(VenueStateOverride), Other(serde::de::IgnoredAny) }
struct VenueStateOverride { #[serde(rename = "stateOverride", default)] state_override: HashMap<String, AccountStateOverride> }
struct AccountStateOverride { #[serde(rename = "stateDiff", default)] state_diff: HashMap<String, String> }
```

We could also consider using the `const *_VENUE` with a custom deserializer implementation but I also found it overkill.

die-herdplatte: With the pain I meant that around half of the lines in the function are about parsing / retrieving properties from the JSON object.
This could be circumvented by a derived `Deserialize` and make the function more concise.

I imagined something like your code example, except that `VenueEntry` doesn't exist and would be replaced with `Value` (one struct less & because we only need to parse a subset of venues).
The `String`s could be replaced by `U256` and `Address` if the `serde` feature of `alloy` gets enabled.
`TitanMessage` would be parsed from the `&str` once outside and the `VenueStateOverride` from the `Value` inside of `parse_message`.

If you don't like that, it's fine, the main issue is the repeated parsing.

zizou0x: Makes sense, looks quite similar to the middle ground I've implemented. I've added direct cast into alloy structs to make things even simpler 👍

crates/tycho-simulation/src/evm/override_stream/titan.rs
zizou0x: I agree it feels fragile, I'm still digging deeper into it, will try to confirm with titan if this is something we can expect to be stable. In the meantime I've tested with this and it works, given that this is private interfaces, I think we should proceed with it for now and leave the improvement for a future PR

crates/tycho-simulation/src/evm/override_stream/mod.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho/pull/1106/commits/e9f00b73322132d0f282157d923599182605318f. I've decided to go for a more flexible approach and leave the responsibility to set TTL to the provider, in case a provider wants a different behavior.

crates/tycho-simulation/src/evm/override_stream/titan.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho/pull/1106/commits/570782fb436d575192fd11aa0e2dfc04e75ae68b. 

We already had full flexibilty because anyone can create their own TitanProvider and set it on the stream but I've added a env variable so they can easily override the default URL without having to change code.

crates/tycho-simulation/src/evm/protocol/vm/state.rs
zizou0x: Nice catch, I've addressed this in https://github.com/propeller-heads/tycho/pull/1106/commits/888b5214cd56988753dc2d59f527be446bde538a.

We get the overrides atomically per function, so we ensure that it stays the same.

crates/tycho-simulation/src/evm/override_stream/titan.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho/pull/1106/commits/305466c619b7fba270d187e1b3fbc0b354f169ed

crates/tycho-simulation/src/evm/override_stream/titan.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho/pull/1106/commits/6521aac6261c440c3278e28ac9f4a5911a294ce4

crates/tycho-simulation/src/evm/protocol/vm/state.rs
zizou0x: Done in https://github.com/propeller-heads/tycho/pull/1106/commits/58d262af93a8443dbfe7a697bffe47bd047ed399

crates/tycho-simulation/src/evm/override_stream/titan.rs
zizou0x: Done in https://github.com/propeller-heads/tycho/pull/1106/commits/bcd3486eddecdfcf3b717a637294f826abf475ba

crates/tycho-simulation/src/evm/protocol/vm/state.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho/pull/1106/commits/1c2ce6fc9662bc8a6fa4d71ffba7ce986f4336fc

crates/tycho-simulation/src/evm/protocol/vm/state.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho/pull/1106/commits/e9f00b73322132d0f282157d923599182605318f

crates/tycho-simulation/src/evm/override_stream/mod.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho/pull/1106/commits/4ec9e05bf9c7584b13c12855519f3a598d3a161a

crates/tycho-simulation/src/evm/override_stream/titan.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho/pull/1106/commits/4ec9e05bf9c7584b13c12855519f3a598d3a161a

crates/tycho-simulation/src/evm/override_stream/titan.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho/pull/1106/commits/4ec9e05bf9c7584b13c12855519f3a598d3a161a

crates/tycho-simulation/src/evm/protocol/vm/state.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho/pull/1106/commits/b0b88c57f1f27bc82f8a017c86b99622b7bbadbc

crates/tycho-simulation/src/evm/stream.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho/pull/1106/commits/4ec9e05bf9c7584b13c12855519f3a598d3a161a

crates/tycho-simulation/src/evm/stream.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho/pull/1106/commits/a4cafd14ce74cde4c508df0f9ee431118bf763a0

crates/tycho-simulation/src/evm/override_stream/mod.rs
tvinagre: This becomes very opinionated on the quote. Maybe the user just wants to add a log to the quote to identify latency on any point of the system without adding an expiration logic. Not a blocker to merge this but I think this is less flexible than having a raw quote ts

crates/tycho-simulation/src/evm/override_stream/titan.rs
tvinagre: agreed!

crates/tycho-simulation/src/evm/protocol/vm/state.rs
```diff
@@ -29,13 +31,23 @@ use super::{
...
     protocol::{
         u256_num::{u256_to_biguint, u256_to_f64},
         utils::bytes_to_address,
     },
     simulation::BlockEnvOverrides,
 };
 
+/// Current unix time in seconds, or `0` if the system clock is before the epoch. Used to test live
+/// override snapshots against their expiry.
+fn unix_now_secs() -> u64 {
```
tvinagre: this is an aux function, shouldn't it be down on the code, right before tests?

tvinagre (approved):

zach030 (approved):
LGTM!

PR #1108, opened 2026-06-23
fix: fix cargo audit

Troshchk:
- Fixes:
[UNKNOWN] ID: RUSTSEC-2026-0185 | Crate: quinn-proto | Version: 0.11.14 | Error:  Remote memory exhaustion in quinn-proto from unbounded out-of-order stream reassembly
- Add comments to allows

### DIFF
--- a/crates/tycho-client/src/feed/synchronizer.rs
+++ b/crates/tycho-client/src/feed/synchronizer.rs
@@ -469,6 +469,8 @@ where
     ///
     /// The returned `end_rx` (if any) should be reused for retry attempts since the close
     /// signal may still arrive and we want to remain cancellable across retries.
+    // Newer clippy versions detect this large Err type. Keep the existing API and allow the lint
+    // explicitly.
     #[allow(clippy::result_large_err)]
     #[instrument(skip(self, block_tx, end_rx), fields(extractor_id = %self.extractor_id))]
     async fn state_sync(
@@ -1301,6 +1303,8 @@ mod test {
                 .await
         }
 
+        // clippy false positive: `'a` is required by the trait method signature and is
+        // used in `SnapshotParameters<'a>`, but `async_trait` makes Clippy miss it.
         #[allow(clippy::extra_unused_lifetimes)]
         async fn get_snapshots<'a>(
             &self,
--- a/crates/tycho-client/src/feed/synchronizer.rs
+++ b/crates/tycho-client/src/feed/synchronizer.rs
@@ -72,6 +72,7 @@ pub enum SynchronizerError {
 }
 
 pub type SyncResult<T> = Result<T, SynchronizerError>;
+pub type SyncError = (SynchronizerError, Option<oneshot::Receiver<()>>);
 
 impl SynchronizerError {
     /// Returns true if the error is transient and the failing operation can be retried.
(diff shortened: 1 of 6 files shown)


crates/tycho-client/src/feed/synchronizer.rs
```diff
@@ -469,6 +469,8 @@ where
     ///
     /// The returned `end_rx` (if any) should be reused for retry attempts since the close
     /// signal may still arrive and we want to remain cancellable across retries.
+    // Newer clippy versions detect this large Err type. Keep the existing API and allow the lint
```
zizou0x: I think we could address this one, the interface is private so there isn't much risk. Here is a solution that works: 

```
diff --git a/crates/tycho-client/src/feed/synchronizer.rs b/crates/tycho-client/src/feed/synchronizer.rs
index a4cc3492e..4a3fc490d 100644
--- a/crates/tycho-client/src/feed/synchronizer.rs
+++ b/crates/tycho-client/src/feed/synchronizer.rs
@@ -72,6 +72,7 @@ pub enum SynchronizerError {
 }
 
 pub type SyncResult<T> = Result<T, SynchronizerError>;
+pub type SyncError = (SynchronizerError, Option<oneshot::Receiver<()>>);
 
 impl SynchronizerError {
     /// Returns true if the error is transient and the failing operation can be retried.
@@ -469,13 +470,12 @@ where
     ///
     /// The returned `end_rx` (if any) should be reused for retry attempts since the close
     /// signal may still arrive and we want to remain cancellable across retries.
-    #[allow(clippy::result_large_err)]
     #[instrument(skip(self, block_tx, end_rx), fields(extractor_id = %self.extractor_id))]
     async fn state_sync(
         &mut self,
         block_tx: &mut Sender<SyncResult<StateSyncMessage<BlockHeader>>>,
         mut end_rx: oneshot::Receiver<()>,
-    ) -> Result<(), (SynchronizerError, Option<oneshot::Receiver<()>>)> {
+    ) -> Result<(), Box<SyncError>> {
         // initialisation
         let subscription_options = SubscriptionOptions::new()
             .with_state(self.include_snapshots)
@@ -487,7 +487,7 @@ where
             .await
         {
             Ok(result) => result,
-            Err(e) => return Err((e.into(), Some(end_rx))),
+            Err(e) => return Err(Box::new((e.into(), Some(end_rx)))),
         };
 
         let result = async {
@@ -879,7 +879,7 @@ where
                 // The error came from the inner async block. Since the async block
                 // can receive close signals (which would return Ok), any error means
                 // the close signal was NOT received, so we can return the end_rx for retry
-                Err((e, Some(end_rx)))
+                Err(Box::new((e, Some(end_rx))))
             }
         }
     }
@@ -1138,7 +1138,8 @@ where
                         );
                         return;
                     }
-                    Err((e, maybe_end_rx)) => {
+                        let (e, maybe_end_rx) = *e;
                         warn!(
                             extractor_id=%&self.extractor_id,
                             retry_count,
@@ -1301,7 +1302,6 @@ mod test {
                 .await
         }
 
-        #[allow(clippy::extra_unused_lifetimes)]
         async fn get_snapshots<'a>(
             &self,
             request: &SnapshotParameters<'a>,

```

zizou0x (approved):

PR #1115, opened 2026-06-24
feat: extend FeeCalculator to split positive slippage surplus

Troshchk:
[Task](https://propeller-heads.atlassian.net/browse/ENG-6053)

Extends FeeCalculator to split positive slippage surplus between the router and client. `calculateFee` now takes `actualAmountOut` and `expectedAmountOut` instead of a single `amountIn`, enabling surplus detection when `actualAmountOut > expectedAmountOut`. Fees are calculated on `expectedAmountOut` (the quoted amount), and any positive slippage is split according to a configurable per client share with a global default. 

A new `mustInterceptOutput` function tells the router whether it needs to receive funds before forwarding (true when slippage capture is enabled, or fees are nonzero). 

The legacy uint16 BPS scale is removed, all fee parameters now use the internal 8-decimal unit scale (uint32, 100_000_000 = 100%) end-to-end, eliminating lossy conversions. 

[This commit](https://github.com/propeller-heads/tycho/pull/1115/commits/1db063dc042a8bfc352993e21dd04bc7126dc2c9) is only for the CI to pass, won't be needed after `extend-router-fees/lp/ENG-6054-router-refactor` branch is merged

### DIFF
--- a/crates/tycho-execution/contracts/interfaces/IFeeCalculator.sol
+++ b/crates/tycho-execution/contracts/interfaces/IFeeCalculator.sol
@@ -5,7 +5,7 @@ import {FeeRecipient} from "../lib/FeeStructs.sol";
 
 /**
  * @notice Per-client custom fee configuration
- * @dev All fields pack into a single storage slot (10 bytes total).
+ * @dev All fields pack into a single storage slot (13 bytes total).
  *      Fee values use 8-decimal precision: 1 unit = 0.0001 BPS = 0.000001%.
  *      100% = 100_000_000 units.
  */
@@ -14,24 +14,47 @@ struct CustomFees {
     uint32 feeBpsOnOutput; // 4 bytes
     bool hasCustomFeeOnClientFee; // 1 byte
     uint32 feeBpsOnClientFee; // 4 bytes
+    bool hasCustomClientSlippageShare; // 1 byte
+    uint16 clientSlippageShareBps; // 2 bytes
 }
 
 interface IFeeCalculator {
     /**
-     * @notice Calculates fees from the swap output amount
+     * @notice Calculates all fees and slippage surplus from swap output
      * @dev Called from TychoRouter. Does not perform any accounting.
-     *      Router fee parameters are retrieved from contract storage based on the user address.
-     *      Client fee parameters are passed as function arguments.
-     * @param amountIn The amount before fee deduction
-     * @param client The client address to look up custom router fees for and to receive fees
-     * @param clientFeeBps Client fee in basis points
-     * @return amountOut The amount remaining after all fee deductions
-     * @return feeRecipients Array of (address, feeAmount) tuples for fee distribution
+     *      Handles both regular fees and positive slippage surplus
+     *      in a single call. Router fee parameters and slippage
+     *      share are retrieved from contract storage based on the
+     *      client address; client fee parameters are passed as
+     *      function arguments.
+     * @param grossAmountOut The actual amount received from the swap
+     * @param quotedAmountOut Caller-supplied quoted amount out.
+     *        Base for fee calculation and slippage surplus
+     * @param clientFeeBps Client fee in basis points (10_000 = 100%)
+     * @param client The client address to look up custom router fees
+     *        and slippage share for, and to receive the client fee
+     * @return feeRecipients Array of (address, feeAmount) tuples for
+     *         fee distribution. Returns [] when there is nothing to
+     *         capture (no fees, no surplus, or toggle off).
      */
-    function calculateFee(uint256 amountIn, address client, uint16 clientFeeBps)
-        external
-        view
-        returns (uint256 amountOut, FeeRecipient[] memory feeRecipients);
+    function calculateFee(
+        uint256 grossAmountOut,
+        uint256 quotedAmountOut,
+        uint32 clientFeeBps,
+        address client
+    ) external view returns (FeeRecipient[] memory feeRecipients);
+
+    /**
+     * @notice Whether the router must receive swap output before forwarding
+     * @dev Covers: slippage enabled, fees > 0, or any future condition.
+     * @param clientFeeBps Client fee in basis points
+     * @param client The client address to check
+     * @return True if the router must intercept output
+     */
+    function mustInterceptOutput(
+        uint32 clientFeeBps,
+        address client
+    ) external view returns (bool);
 
     /**
      * @dev Returns the effective router fee on output amount for a specific client
@@ -49,7 +72,7 @@ interface IFeeCalculator {
      * @param client The client address to check
      * @return The fee in fee units (custom if set, otherwise default)
      */
-    function getEffectiveRouterFeeOnOutputScaled(address client)
+   function getEffectiveRouterFeeOnOutputScaled(address client)
         external
         view
         returns (uint32);
@@ -61,7 +84,10 @@ interface IFeeCalculator {
      * @return clients Addresses of clients with at least one custom fee
      * @return fees Custom fee configuration for each client (parallel array)
      */
-    function getAllClientFees(uint256 start, uint256 count)
+    function getAllClientFees(
+        uint256 start,
+        uint256 count
+    )
         external
         view
         returns (address[] memory clients, CustomFees[] memory fees);
--- a/crates/tycho-execution/contracts/interfaces/IFeeCalculator.sol
+++ b/crates/tycho-execution/contracts/interfaces/IFeeCalculator.sol
@@ -5,7 +5,7 @@ import {FeeRecipient} from "../lib/FeeStructs.sol";
 
 /**
  * @notice Per-client custom fee configuration
- * @dev All fields pack into a single storage slot (13 bytes total).
+ * @dev All fields pack into a single storage slot (15 bytes total).
  *      Fee values use 8-decimal precision: 1 unit = 0.0001 BPS = 0.000001%.
  *      100% = 100_000_000 units.
  */
@@ -15,7 +15,7 @@ struct CustomFees {
     bool hasCustomFeeOnClientFee; // 1 byte
     uint32 feeBpsOnClientFee; // 4 bytes
     bool hasCustomClientSlippageShare; // 1 byte
-    uint16 clientSlippageShareBps; // 2 bytes
+    uint32 clientSlippageShareBps; // 4 bytes
 }
 
 interface IFeeCalculator {
@@ -30,7 +30,7 @@ interface IFeeCalculator {
      * @param actualAmountOut The actual amount received from the swap
      * @param expectedAmountOut Caller-supplied quoted amount out.
      *        Base for fee calculation and slippage surplus
-     * @param clientFeeBps Client fee in basis points (10_000 = 100%)
+     * @param clientFeeBps Client fee in fee units (100_000_000 = 100%)
      * @param client The client address to look up custom router fees
      *        and slippage share for, and to receive the client fee
      * @return feeRecipients Array of (address, feeAmount) tuples for
@@ -47,7 +47,7 @@ interface IFeeCalculator {
     /**
      * @notice Whether the router must receive swap output before forwarding
      * @dev Covers: slippage enabled, fees > 0, or any future condition.
-     * @param clientFeeBps Client fee in basis points
+     * @param clientFeeBps Client fee in fee units (100_000_000 = 100%)
      * @param client The client address to check
      * @return True if funds must pass through the router after the
      *         final swap instead of going directly to the receiver
--- a/crates/tycho-execution/contracts/interfaces/IFeeCalculator.sol
+++ b/crates/tycho-execution/contracts/interfaces/IFeeCalculator.sol
@@ -49,7 +49,8 @@ interface IFeeCalculator {
      * @dev Covers: slippage enabled, fees > 0, or any future condition.
      * @param clientFeeBps Client fee in basis points
      * @param client The client address to check
-     * @return True if the router must intercept output
+     * @return True if funds must pass through the router after the
+     *         final swap instead of going directly to the receiver
      */
     function mustInterceptOutput(
         uint32 clientFeeBps,
--- a/crates/tycho-execution/contracts/interfaces/IFeeCalculator.sol
+++ b/crates/tycho-execution/contracts/interfaces/IFeeCalculator.sol
@@ -27,8 +27,8 @@ interface IFeeCalculator {
      *      share are retrieved from contract storage based on the
      *      client address; client fee parameters are passed as
      *      function arguments.
-     * @param grossAmountOut The actual amount received from the swap
-     * @param quotedAmountOut Caller-supplied quoted amount out.
+     * @param actualAmountOut The actual amount received from the swap
+     * @param expectedAmountOut Caller-supplied quoted amount out.
      *        Base for fee calculation and slippage surplus
      * @param clientFeeBps Client fee in basis points (10_000 = 100%)
      * @param client The client address to look up custom router fees
@@ -38,8 +38,8 @@ interface IFeeCalculator {
      *         capture (no fees, no surplus, or toggle off).
      */
     function calculateFee(
-        uint256 grossAmountOut,
-        uint256 quotedAmountOut,
+        uint256 actualAmountOut,
+        uint256 expectedAmountOut,
         uint32 clientFeeBps,
         address client
     ) external view returns (FeeRecipient[] memory feeRecipients);
--- a/crates/tycho-execution/contracts/interfaces/IFeeCalculator.sol
+++ b/crates/tycho-execution/contracts/interfaces/IFeeCalculator.sol
@@ -58,22 +58,12 @@ interface IFeeCalculator {
     ) external view returns (bool);
 
     /**
-     * @dev Returns the effective router fee on output amount for a specific client
-     * @param client The client address to check
-     * @return The fee in basis points (custom if set, otherwise default)
-     */
-    function getEffectiveRouterFeeOnOutput(address client)
-        external
-        view
-        returns (uint16);
-
-    /**
-     * @dev Returns the effective router fee on output for a specific client in the internal
-     *      8-decimal fee unit scale (100_000_000 = 100%).
+     * @dev Returns the effective router fee on output for a specific client
+     *      in fee units (100_000_000 = 100%).
      * @param client The client address to check
      * @return The fee in fee units (custom if set, otherwise default)
      */
-   function getEffectiveRouterFeeOnOutputScaled(address client)
+    function getEffectiveRouterFeeOnOutput(address client)
         external
         view
         returns (uint32);
--- a/crates/tycho-execution/contracts/src/Dispatcher.sol
+++ b/crates/tycho-execution/contracts/src/Dispatcher.sol
@@ -297,16 +297,24 @@ contract Dispatcher is TransferManager {
 
     function _callCalculateFee(
         address feeCalculator,
-        uint256 amountIn,
-        uint16 clientFeeBps,
+        uint256 grossAmountOut,
+        uint256 quotedAmountOut,
+        uint32 clientFeeBps,
         address client
-    )
-        internal
-        view
-        returns (uint256 amountOut, FeeRecipient[] memory feeRecipients)
-    {
+    ) internal view returns (FeeRecipient[] memory feeRecipients) {
         // slither-disable-next-line calls-loop
-        (amountOut, feeRecipients) = IFeeCalculator(feeCalculator)
-            .calculateFee(amountIn, client, clientFeeBps);
+        feeRecipients = IFeeCalculator(feeCalculator)
+            .calculateFee(grossAmountOut, quotedAmountOut, clientFeeBps, client);
+    }
+
+    function _callMustInterceptOutput(
+        address feeCalculator,
+        uint32 clientFeeBps,
+        address client
+    ) internal view returns (bool) {
+        // slither-disable-next-line calls-loop
+        return
+            IFeeCalculator(feeCalculator)
+                .mustInterceptOutput(clientFeeBps, client);
     }
 }
--- a/crates/tycho-execution/contracts/src/Dispatcher.sol
+++ b/crates/tycho-execution/contracts/src/Dispatcher.sol
@@ -299,14 +299,14 @@ contract Dispatcher is TransferManager {
     // slither-disable-next-line dead-code
     function _callCalculateFee(
         address feeCalculator,
-        uint256 grossAmountOut,
-        uint256 quotedAmountOut,
+        uint256 actualAmountOut,
+        uint256 expectedAmountOut,
         uint32 clientFeeBps,
         address client
     ) internal view returns (FeeRecipient[] memory feeRecipients) {
         // slither-disable-next-line calls-loop
         feeRecipients = IFeeCalculator(feeCalculator)
-            .calculateFee(grossAmountOut, quotedAmountOut, clientFeeBps, client);
+            .calculateFee(actualAmountOut, expectedAmountOut, clientFeeBps, client);
     }
 
     // TODO(ENG-6054): remove dead-code suppression once _takeFees is wired up
--- a/crates/tycho-execution/contracts/src/Dispatcher.sol
+++ b/crates/tycho-execution/contracts/src/Dispatcher.sol
@@ -289,7 +289,7 @@ contract Dispatcher is TransferManager {
     function _callGetEffectiveRouterFeeOnOutput(
         address feeCalculator,
         address client
-    ) internal view returns (uint16 routerFeeOnOutputBps) {
+    ) internal view returns (uint32 routerFeeOnOutputBps) {
         // slither-disable-next-line calls-loop
         routerFeeOnOutputBps =
             IFeeCalculator(feeCalculator).getEffectiveRouterFeeOnOutput(client);
--- a/crates/tycho-execution/contracts/src/Dispatcher.sol
+++ b/crates/tycho-execution/contracts/src/Dispatcher.sol
@@ -306,7 +306,9 @@ contract Dispatcher is TransferManager {
     ) internal view returns (FeeRecipient[] memory feeRecipients) {
         // slither-disable-next-line calls-loop
         feeRecipients = IFeeCalculator(feeCalculator)
-            .calculateFee(actualAmountOut, expectedAmountOut, clientFeeBps, client);
+            .calculateFee(
+                actualAmountOut, expectedAmountOut, clientFeeBps, client
+            );
     }
 
     // TODO(ENG-6054): remove dead-code suppression once _takeFees is wired up
--- a/crates/tycho-execution/contracts/src/Dispatcher.sol
+++ b/crates/tycho-execution/contracts/src/Dispatcher.sol
@@ -295,6 +295,7 @@ contract Dispatcher is TransferManager {
             IFeeCalculator(feeCalculator).getEffectiveRouterFeeOnOutput(client);
     }
 
+    // slither-disable-next-line dead-code
     function _callCalculateFee(
         address feeCalculator,
         uint256 grossAmountOut,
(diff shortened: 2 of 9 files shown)


crates/tycho-execution/contracts/interfaces/IFeeCalculator.sol
```diff
@@ -14,24 +14,47 @@ struct CustomFees {
     uint32 feeBpsOnOutput; // 4 bytes
     bool hasCustomFeeOnClientFee; // 1 byte
     uint32 feeBpsOnClientFee; // 4 bytes
+    bool hasCustomClientSlippageShare; // 1 byte
+    uint16 clientSlippageShareBps; // 2 bytes
```
tamaralipows: Should we make this uint32 like the rest of the fee bps?

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -159,6 +217,68 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
         return client == address(0) ? tx.origin : client;
     }
 
+    /**
+     * @dev Calculates positive slippage surplus distribution
+     * @return Empty array if disabled or no surplus; otherwise 2-element array [router, client]
+     */
+    function _calculatePositiveSlippage(
+        uint256 grossAmountOut,
+        uint256 quotedAmountOut,
```
tamaralipows: I will say the same thing here like I said on Louise's PR about the naming: I'd much prefer `amountOutBeforeFees` and `expectedAmountOut`, but if you both insist on this way I wont fight it lol

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -61,40 +68,94 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
...
+        if (_positiveSlippageEnabled) return true;
+
+        address resolvedClient = _resolveClient(client);
         (uint32 routerFeeOnOutputBps, uint32 routerFeeOnClientFeeBps) =
-            _getFeeInfo(_resolveClient(client));
+            _getFeeInfo(resolvedClient);
+
+        if (clientFeeBps > 0) return true;
+        if (routerFeeOnOutputBps > 0) return true;
+        if (routerFeeOnClientFeeBps > 0) return true;
```
tamaralipows: I think this one only returns true if `clientFeeBps > 0`. Else we are taking the router portion of 0 fees, which will also be 0

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -418,4 +538,89 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
...
+    function removeCustomClientSlippageShare(address client)
+        external
+        onlyRole(ROUTER_FEE_SETTER_ROLE)
+    {
+        CustomFees memory customFees = _customRouterFees[client];
+        customFees.hasCustomClientSlippageShare = false;
+        customFees.clientSlippageShareBps = 0;
+        _customRouterFees[client] = customFees;
+
+        if (!customFees.hasCustomFeeOnOutput) {
```
tamaralipows: what about `hasCustomFeeOnClientFee`?

crates/tycho-execution/contracts/src/FeeCalculator.sol
tamaralipows: Here we will now also need to check about the slippage fees before removing the client from the list (please add tests for these cases)

tamaralipows: And here too, we will need to know about the custom slippage portion

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -61,40 +68,94 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
...
+        uint32 clientFeeBps,
+        address client
+    ) external view returns (FeeRecipient[] memory feeRecipients) {
+        address resolvedClient = _resolveClient(client);
+
+        // Calculate fees
+        FeeRecipient[] memory fees =
+            _calculateFee(quotedAmountOut, resolvedClient, clientFeeBps);
+
+        // Calculate slippage
```
tamaralipows: Unnecessary comments, same with the merge results. The fn names are descriptive enough

crates/tycho-execution/contracts/test/FeeCalculator.t.sol
```diff
@@ -776,3 +801,101 @@ contract FeeCalculatorConfigTest is Constants {
...
+
+    function setUp() public {
+        feeCalculator = new FeeCalculator(FEE_SETTER);
+        vm.startPrank(FEE_SETTER);
+        feeCalculator.setRouterFeeReceiver(ADMIN);
+        feeCalculator.setPositiveSlippageEnabled(true);
+        vm.stopPrank();
+    }
+
+    function testCalculatePositiveSlippage_routerKeepsAll() public view {
```
tamaralipows: Not a fan of the underscores in the test names - I know it was in the task (I commented the same in the task). Could we name them better, for example, `testRouterKeepsPositiveSlippage`

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -159,6 +217,68 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
...
+    function _calculatePositiveSlippage(
+        uint256 grossAmountOut,
+        uint256 quotedAmountOut,
+        address client
+    ) internal view returns (FeeRecipient[] memory) {
+        if (!_positiveSlippageEnabled || grossAmountOut <= quotedAmountOut) {
+            return new FeeRecipient[](0);
+        }
+
+        uint256 surplus = grossAmountOut - quotedAmountOut;
```
tamaralipows: We are dealing with two `uint256`s here, wouldn't this underflow if `grossAmountOut < quotedAmountOut`, which is entirely possible with negative slippage actually? Could we add a test for this case and also in the case where slippage is 0?

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -61,40 +68,94 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
...
+     * @param client The client address to check
+     * @return True if the router must intercept output
+     */
+    function mustInterceptOutput(uint32 clientFeeBps, address client)
         external
         view
-        returns (uint256 amountOut, FeeRecipient[] memory feeRecipients)
+        returns (bool)
     {
+        if (_positiveSlippageEnabled) return true;
```
tamaralipows, line 141: I guess we will have to call the fee calculator even for the cases we get negative slippage, right? Since we are using this to determine the final receiver, and at that point we don't know the true amount out yet. I think that's totally fine, just something to keep in mind (maybe write a comment here?)

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -61,40 +68,94 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
...
-     *               Pass address(0) to fall back to tx.origin for the custom fee lookup.
+     * @param grossAmountOut The actual amount received from the swap
+     *        (used for slippage surplus calculation)
+     * @param quotedAmountOut Caller-supplied quoted amount out.
      * @param clientFeeBps Client fee in basis points (10000 = 100%)
-     * @return amountOut The amount remaining after all fee deductions
+     * @param client The client address to look up custom router fees
+     *        and slippage share for and to receive fees.
+     *        Pass address(0) to fall back to tx.origin for the
+     *        custom fee lookup.
```
tamaralipows, line 98: I would like to make our slippage mechanism clearer in the docstring so it's transparent for the clients. i.e. first we take the slippage, and then **also** any existing fees on top of it. Else it's not clear whether fees are taken before or after slippage without having to read the code.

By the way related side note to @louise-poole: we will need to be careful about how we configure this on the Fynd side - because in this case we can't define the amount out to already be net fees, otherwise we will be taking fees on top of the amount out that already considered fees. xD I didn't double check the task right now so maybe you already mentioned that - just making a mental note. We will also have to mention this clearly in the docs to direct tycho users.

crates/tycho-execution/contracts/interfaces/IFeeCalculator.sol
```diff
@@ -14,24 +14,47 @@ struct CustomFees {
...
+    ) external view returns (FeeRecipient[] memory feeRecipients);
+
+    /**
+     * @notice Whether the router must receive swap output before forwarding
+     * @dev Covers: slippage enabled, fees > 0, or any future condition.
      * @param clientFeeBps Client fee in basis points
-     * @return amountOut The amount remaining after all fee deductions
-     * @return feeRecipients Array of (address, feeAmount) tuples for fee distribution
+     * @param client The client address to check
+     * @return True if the router must intercept output
```
tamaralipows: I'd say `return True` if the funds must pass through the router after the final swap instead of going directly to the client. "must intercept output" is a bit fancy language (which is fine for fn names)

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -61,40 +68,94 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
...
      * @return feeRecipients Array of (address, feeAmount) tuples for fee distribution
      */
-    function calculateFee(uint256 amountIn, address client, uint16 clientFeeBps)
+    function calculateFee(
+        uint256 grossAmountOut,
+        uint256 quotedAmountOut,
+        uint32 clientFeeBps,
+        address client
+    ) external view returns (FeeRecipient[] memory feeRecipients) {
```
tamaralipows, line 106: I'm actually very glad we stopped returning the `amountOut` here - good security decision. ✨

tamaralipows (commented):
Thanks Ksenia, good main logic, just some smol things! ✨

p.s. I already asked Louise to do this but didn't ask you yet - could you link the JIRA tasks in the PR descriptions? Makes it a lot easier for me to read the task without having to open JIRA and search myself :)

crates/tycho-execution/contracts/src/Dispatcher.sol
```diff
@@ -295,18 +295,28 @@ contract Dispatcher is TransferManager {
             IFeeCalculator(feeCalculator).getEffectiveRouterFeeOnOutput(client);
     }
 
+    // slither-disable-next-line dead-code
```
dianacarvalho1, line 299: can we add a TODO here too to remove this in the next task please?

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -61,40 +68,94 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
...
+        address resolvedClient = _resolveClient(client);
+
+        // Calculate fees
```
dianacarvalho1: these three comments are necessary - the function names are explicit enough

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -61,40 +68,94 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
...
+        address client,
+        uint32 clientFeeBps
+    ) internal view returns (FeeRecipient[] memory feeRecipients) {
+        (uint32 routerFeeOnOutputBps, uint32 routerFeeOnClientFeeBps) =
+            _getFeeInfo(client);
 
         // Scale clientFeeBps from legacy scale (10_000 = 100%) to internal scale
         // (100_000_000 = 100%) so both fee types can be compared and combined.
-        uint32 scaledClientFeeBps = uint32(clientFeeBps) * 10_000;
+        uint32 scaledClientFeeBps = clientFeeBps * 10_000;
```
dianacarvalho1: this scaling is not needed anymore if we are assuming that the clientFeeBps is already in uint32

crates/tycho-execution/contracts/test/FeeCalculator.t.sol
```diff
@@ -48,17 +56,19 @@ contract FeeCalculatorTest is Constants {
...
     function testCalculateOnlyRouterFeeOnClientFee() public {
-        // Test with only router fee on client fee set (requires client fee to be set too)
+        // Test with only router fee on client fee set
+        // (requires client fee to be set too)
         vm.prank(FEE_SETTER);
         feeCalculator.setRouterFeeOnClientFee(_10_PCT); // 10% of client fee
 
         uint256 amountIn = 1 ether;
-        uint16 clientFeeBps = 200; // 2% in legacy BPS scale
+        uint32 clientFeeBps = 200; // 2% in legacy BPS scale
```
dianacarvalho1: this is wrong now 😕 you are already assuming we are not using the legacy scale by making it a uint32 instead of a uint16

dianacarvalho1 (commented):
Thank you @Troshchk ! Overall it looks good - only have small comments. A more general one though: in all the temporary code/comments please leave a real TODO so it's easier to find them later in the next task and make sure everything will get cleaned.

crates/tycho-execution/contracts/src/FeeCalculator.sol
Troshchk: agree about the `routerFeeOnClientFeeBps `, but isn't `routerFeeOnOutputBps` a separate check?

crates/tycho-execution/contracts/src/FeeCalculator.sol
Troshchk: It should not underflow as there is a check earlier `grossAmountOut <= quotedAmountOut`. In that case empty array is returned `FeeRecipient[](0)`
Added test for these cases

crates/tycho-execution/contracts/src/FeeCalculator.sol
tamaralipows: Was only referring to last line!

crates/tycho-execution/contracts/src/FeeCalculator.sol
louise-poole: This is already noted in the task to update Fynd [here](https://propeller-heads.atlassian.net/browse/ENG-6115). But thanks for flagging it too! Prefer to double check it like this than have missed it entirely.

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -61,57 +65,111 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
...
+
+        FeeRecipient[] memory slippage = _calculatePositiveSlippage(
+            actualAmountOut, expectedAmountOut, resolvedClient
+        );
+
+        return _mergeFeeRecipients(fees, slippage);
+    }
+
+    /**
+     * @notice Whether the router must receive swap output before forwarding
```
louise-poole: I think this could be clearer... ('before forwarding' to who?)
I really like the wording you used for the return description - could you use similar here?

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -61,57 +65,111 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
...
         uint256 routerFeeOnClientFee = 0;
         uint256 clientPortion = 0;
 
         // Calculate client fee if > 0
-        if (scaledClientFeeBps > 0) {
+        if (clientFeeBps > 0) {
             // Save numerator for later routerFeeOnClientFee calculation to avoid
             // divide-before-multiply precision loss and warning
-            uint256 clientFeeNumerator = amountOut * scaledClientFeeBps;
+            uint256 clientFeeNumerator = expectedAmountOut * clientFeeBps;
```
louise-poole: Your doc string earlier says fees are calculated on the `actualAmountOut`, not `expectedAmountOut`?

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -61,57 +65,111 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
...
+        address client
+    ) external view returns (FeeRecipient[] memory feeRecipients) {
+        address resolvedClient = _resolveClient(client);
+
+        FeeRecipient[] memory fees =
+            _calculateFee(expectedAmountOut, resolvedClient, clientFeeBps);
+
+        FeeRecipient[] memory slippage = _calculatePositiveSlippage(
+            actualAmountOut, expectedAmountOut, resolvedClient
+        );
```
louise-poole: The doc string says slippage is taken first and then fees applied, yet here it is ordered in the opposite way? That won't affect the values, just found it a little unintuitive.

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -61,57 +65,111 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
...
+    function calculateFee(
+        uint256 actualAmountOut,
+        uint256 expectedAmountOut,
+        uint32 clientFeeBps,
+        address client
+    ) external view returns (FeeRecipient[] memory feeRecipients) {
+        address resolvedClient = _resolveClient(client);
+
+        FeeRecipient[] memory fees =
+            _calculateFee(expectedAmountOut, resolvedClient, clientFeeBps);
```
louise-poole: We should calculate this on the actualAmountOut.

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -159,6 +212,68 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
...
+            return customFees.clientSlippageShareBps;
+        }
+        return _defaultClientSlippageShareBps;
+    }
+
+    /**
+     * @dev Merges fee recipients with slippage recipients.
+     *      Combines amounts for the same recipient (router fee receiver).
+     */
+    function _mergeFeeRecipients(
```
louise-poole, line 271: ❤️

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -159,6 +212,68 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
...
+        FeeRecipient[] memory fees,
+        FeeRecipient[] memory slippage
+    ) internal pure returns (FeeRecipient[] memory) {
+        if (slippage.length == 0) return fees;
+
+        // fees[0] = router fees, slippage[0] = router slippage
+        // fees[1] = client fees, slippage[1] = client slippage
```
louise-poole, line 278: We assume they follow the same ordering - is this documented on the fns that create these FeeRecipients? Don't want a future engineer to not know the side-affect of changing this ordering.

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -159,6 +212,68 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
...
+        // Merge router amounts into fees[0]
+        fees[0].feeAmount += slippage[0].feeAmount;
+        // Merge client slippage into fees[1]
+        fees[1].feeAmount += slippage[1].feeAmount;
```
louise-poole, line 282: What happens if the client has no share? Is there still a `slippage[1]` value to unpack of 0?
Also what happens if the client has no custom router fee, but does have a slippage share? Is there still an appropriate `fees[1]` value to unpack here?

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -418,4 +516,92 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
...
+    {
+        if (bps > MAX_SLIPPAGE_SHARE_BPS) {
+            revert FeeCalculator__InvalidBps();
+        }
+        CustomFees memory customFees = _customRouterFees[client];
+        customFees.hasCustomClientSlippageShare = true;
+        customFees.clientSlippageShareBps = bps;
+        _customRouterFees[client] = customFees;
+        // slither-disable-next-line unused-return
+        _customFeeClients.add(client);
```
louise-poole, line 589: Would this error if they have a custom router fee already set (but not a slippage share set yet)?

crates/tycho-execution/contracts/test/FeeCalculator.t.sol
```diff
@@ -274,24 +297,15 @@ contract FeeCalculatorTest is Constants {
...
-    function testGetEffectiveRouterFeeOnOutputDetectsSubBpsFee() public {
-        // A sub-BPS fee must still be detectable as non-zero via the legacy uint16 interface
-        // used by Dispatcher for zero-checks.
-        vm.prank(FEE_SETTER);
-        feeCalculator.setRouterFeeOnOutput(5_000); // 0.5 BPS
-
-        assertEq(feeCalculator.getEffectiveRouterFeeOnOutput(BOB), 1);
-        assertEq(feeCalculator.getEffectiveRouterFeeOnOutputScaled(BOB), 5_000);
-    }
-
```
louise-poole, line 294: Why is the sub-bps test removed?

louise-poole (commented):
Most of these questions are for me to learn (and maybe there's a chance I'm pointing out something valid 👀 )
Thanks for the attention to detail in this PR! 🙏

crates/tycho-execution/contracts/src/FeeCalculator.sol
Troshchk: Fixed the doc string

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -31,6 +29,8 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
     uint32 public constant MAX_FEE_BPS = 100_000_000;
     // Combined denominator when both fees use the MAX_FEE_BPS scale (MAX_FEE_BPS^2)
     uint64 public constant MAX_FEE_BPS_SQUARED = 10_000_000_000_000_000;
+    // 100% in 8-decimal fee units (for slippage share validation)
+    uint32 public constant MAX_SLIPPAGE_SHARE_BPS = 100_000_000;
```
dianacarvalho1: this is the same value as `MAX_FEE_BPS`, can we use the have constant? maybe we can rename it

crates/tycho-execution/contracts/src/FeeCalculator.sol
Troshchk: I think we should calculate fees based on expectedAmountOut. Otherwise, the client may receive less than expected, because fees calculated from actualAmountOut would be higher when the actual output is higher.

crates/tycho-execution/contracts/src/FeeCalculator.sol
Troshchk: added documentation

crates/tycho-execution/contracts/src/FeeCalculator.sol
dianacarvalho1: not necessarily. but I feel like there are two big assumptions here: that the positive slippage will always be taken and that actualAmountOut>expectedAmountOut. 

Correct me if I'm wrong but from what I understand what is being done is:
- we take positive slippage using the actualAmountOut and expectedAmountOut
- take fees only over the expectedAmountOut -> which assumes that user will ever receive expectedAmountOut

I can find two issues with this:
1. What if slippage is not enabled? we are taking less fees than we could have
2. What if actualAmountOut<expectedAmountOut but still within the slippage tolerance? We would be taking a lot more fees than expected

The solution would be to actually return the amountAfterPositiveSlippage and use that instead of expectedAmountOut in _calculateFee

crates/tycho-execution/contracts/src/FeeCalculator.sol
Troshchk: Slippage is empty(length of 0) only `if (!_positiveSlippageEnabled || actualAmountOut <= expectedAmountOut)`. In other cases it is always created. So, if clientShareBps is 0, slippage[1] holds 0

_calculateFee always returns an array, so there is always an appropriate fees[1] value to unpack

crates/tycho-execution/contracts/test/FeeCalculator.t.sol
```diff
@@ -8,14 +8,21 @@ import {FeeRecipient} from "../lib/FeeStructs.sol";
 import "./Constants.sol";
 
 // Fee constants in the internal 8-decimal scale (100_000_000 = 100%).
-// clientFeeBps arguments to calculateFee use the legacy BPS scale (10_000 = 100%).
+// clientFeeBps arguments to calculateFee use the 8-decimal fee unit scale (100_000_000 = 100%).
```
dianacarvalho1: repeated comment

crates/tycho-execution/contracts/src/FeeCalculator.sol
Troshchk: No, it won't. add() returns false if the element already exists, but it does not error

crates/tycho-execution/contracts/test/FeeCalculator.t.sol
Troshchk: Before, `getEffectiveRouterFeeOnOutput` returned a uint16, where a sub-BPS fee like 0.5 BPS (5,000 units) would be downscaled to 1 just to pass non-zero checks. `getEffectiveRouterFeeOnOutputScaled` returned the full uint32 value. This test verified that behavior. 
Now the legacy uint16 interface is removed -- `getEffectiveRouterFeeOnOutpu`t returns uint32 directly (what getEffectiveRouterFeeOnOutputScaled used to do), so the tested behavior no longer exists and the test has no purpose

crates/tycho-execution/contracts/test/FeeCalculator.t.sol
```diff
@@ -775,4 +770,161 @@ contract FeeCalculatorConfigTest is Constants {
...
+        uint256 actualAmountOut = 0.9 ether;
+        uint256 expectedAmountOut = 1 ether;
+
+        FeeRecipient[] memory fees = feeCalculator.calculateFee(
+            actualAmountOut, expectedAmountOut, 0, BOB
+        );
+
+        // No surplus, but router fee on output still applies
+        // routerFee = 1 ether * 1% = 0.01 ether (based on expectedAmountOut)
+        assertEq(fees[0].feeAmount, 0.01 ether);
```
dianacarvalho1: this test shows the issue I pointed in the other comment: we are overcharging fees. The expectedAmountOut is a phantom high amount. The correct fee in this case should be 0.09 ether (using the actualAmountOut)

dianacarvalho1 (commented):
Thank you @Troshchk ! Almost there! 🤞🏼

crates/tycho-execution/contracts/src/FeeCalculator.sol
louise-poole: Yes yes - that's exactly what I was meaning 👍 We need it calculated on what the user actually receives. Which is essentially the expectedAmountOut if there is positive slippage and actualAmountOut if there is negative slippage. Or similarly (as Diana described) `actualAmountOut - positive slippage`.

Troshchk: I understand now, agree and fixed!

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -61,65 +63,129 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
...
-        // Scale clientFeeBps from legacy scale (10_000 = 100%) to internal scale
-        // (100_000_000 = 100%) so both fee types can be compared and combined.
-        uint32 scaledClientFeeBps = uint32(clientFeeBps) * 10_000;
+    /**
+     * @dev Calculates fees from the expected swap output amount
+     * @return feeRecipients 2-element array: [0] = router, [1] = client.
+     *         _mergeFeeRecipients relies on this ordering.
+     */
+    function _calculateFee(
+        uint256 expectedAmountOut,
```
louise-poole: Can you rename this to `feeBase` or `amountOut`?

louise-poole (commented):
Last little comment, otherwise I'm happy!

crates/tycho-execution/contracts/src/FeeCalculator.sol
Troshchk: Fixed!

louise-poole (approved):
Thanks!!!!

PR #1116, opened 2026-06-24
feat: add ChainConfigRegistry

brunoguerios:
Introduce the registry and config loading without touching the Chain enum.

# Changes

- Move chain value types (ChainAddress, ChainConfigError, ChainTokenConfig, TvlThresholds,
  TvlThresholdTier, CustomChainConfig) into tycho-common/src/models/chain_config.rs; re-export from
  models/mod.rs so existing import paths are unchanged.
- Add ChainConfigRegistry { custom: HashMap<String, CustomChainConfig> } with empty / from_configs /
  load_default / get / contains. load_default reads TYCHO_CHAIN_CONFIG else ./chain.yaml; missing file
  => empty registry.
- Add a lazily-initialised global accessor chain_registry() (OnceLock) plus init_chain_registry() for
  explicit init (used by the indexer).
- Add ChainConfigError::UnknownChain(String).

# Suggested commits

1. refactor(common): move chain value types into chain_config module
2. feat(common): add ChainConfigRegistry with file/env config loading

# Acceptance criteria

- No existing behavior changes; registry not yet wired into Chain.
- Tests: YAML round-trip; load_default with/without file; get/contains; explicit init wins.
- cargo test -p tycho-common passes.

# Files

tycho-common/src/models/chain_config.rs (new), tycho-common/src/models/mod.rs, tycho-common/Cargo.toml

### DIFF
--- a/crates/tycho-common/src/models/chain_config.rs
+++ b/crates/tycho-common/src/models/chain_config.rs
@@ -1,6 +1,8 @@
 //! Value types describing a blockchain's configuration: token metadata, TVL thresholds, and the
 //! full `CustomChainConfig` used to describe user-defined chains. Re-exported from `models`.
 
+use std::{collections::HashMap, sync::OnceLock};
+
 use arrayvec::ArrayString;
 use deepsize::DeepSizeOf;
 use serde::{Deserialize, Serialize};
@@ -61,6 +63,8 @@ pub enum ChainConfigError {
     SymbolTooLong(String),
     #[error("chain name '{0}' too long (max 32 chars)")]
     NameTooLong(String),
+    #[error("unknown chain '{0}': not a built-in chain and no custom config registered")]
+    UnknownChain(String),
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize, ToSchema)]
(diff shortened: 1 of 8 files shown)


crates/tycho-common/src/models/chain_config.rs
```diff
@@ -0,0 +1,342 @@
...
+            .map(|cfg| (cfg.name().to_owned(), cfg))
+            .collect();
+        Self { custom }
+    }
+
+    /// Reads custom chain configs from `TYCHO_CHAIN_CONFIG` (default `./chain.yaml`).
+    ///
+    /// Returns an empty registry when the file is absent, so runs on built-in chains need no config
+    /// file. Panics with context if the file exists but cannot be read or parsed.
+    pub fn load_default() -> Self {
```
zizou0x: Let's make it return a result instead. As this is in a lib (tycho-common) we really don't want to panic, it's better to leave it up to the user to decide how they want to react.

crates/tycho-common/src/models/mod.rs
```diff
@@ -1,17 +1,17 @@
...
+mod chain_config;
 pub mod contract;
 pub mod error;
 pub mod protocol;
 pub mod token;
 
 use std::{collections::HashMap, fmt::Display, str::FromStr};
 
-use arrayvec::ArrayString;
+pub use chain_config::*;
```
zizou0x: Any particular reason why we re-export it instead of just exporting normally?

crates/tycho-common/src/models/chain_config.rs
```diff
@@ -0,0 +1,342 @@
...
+            .map_err(|e| serde::de::Error::custom(format!("invalid hex address '{s}': {e}")))?;
+        ChainAddress::new(&raw).map_err(serde::de::Error::custom)
+    }
+}
+
+#[derive(Error, Debug, PartialEq)]
+pub enum ChainAddressError {
+    #[error("address is {0} bytes, max is 32")]
+    TooLong(usize),
+}
```
zizou0x: IMO are too many new errors for such a simple addition. Why don't we use `ChainConfigError::AddressTooLong` instead of this?

crates/tycho-common/src/models/chain_config.rs
```diff
@@ -0,0 +1,342 @@
...
+/// `chains:` section of the indexer's extractor config, so a consumer can point
+/// `TYCHO_CHAIN_CONFIG` at the same file the indexer uses.
+#[derive(Debug, Default, Deserialize)]
+struct ChainConfigFile {
+    #[serde(default)]
+    chains: Vec<CustomChainConfig>,
+}
+
+#[derive(Debug, Error)]
+pub enum ChainConfigFileError {
```
zizou0x: These don't need to be public, we could have a ChainConfigError::File error that inherit from this one. But I would only keep one Error enum public for simplicity (ChainConfigError).

crates/tycho-common/src/models/chain_config.rs
```diff
@@ -0,0 +1,342 @@
...
+    /// An empty registry — only built-in chains will resolve.
+    pub fn empty() -> Self {
+        Self { custom: HashMap::new() }
+    }
+
+    /// Builds a registry from custom chain configs, keyed by chain name.
+    pub fn from_configs(configs: impl IntoIterator<Item = CustomChainConfig>) -> Self {
+        let custom = configs
+            .into_iter()
+            .map(|cfg| (cfg.name().to_owned(), cfg))
```
zizou0x: We should log a warn if there is a duplicate here to avoid unclear behavior

crates/tycho-common/src/models/chain_config.rs
```diff
@@ -0,0 +1,342 @@
+//! Value types describing a blockchain's configuration: token metadata, TVL thresholds, and the
```
zizou0x, line 1: Overall comment on the whole file: in general it's better to keep everything as private as possible and interate. I have a feeling most of these should be private and exposed through `Chain` interface (except maybe `init_chain_registry` so people can initialize it customly)

zizou0x (commented):
Just a few comments but overall looks good!

crates/tycho-common/src/models/mod.rs
brunoguerios: The idea was to avoid touching imports elsewhere that could stick to using `models::CustomChainConfig` instead of having to include the namespace, such as `models::chain_config::CustomChainConfig`, but I agree there's benefit in keeping the standard - I'll export normally and change usages so they include namespace where needed.

crates/tycho-common/src/models/chain_config.rs
brunoguerios: I'll go ahead and fold these into ChainConfigError as well. I think it's even better to centralize all errors there. Let me know your thoughts.

crates/tycho-common/src/models/chain_config.rs
brunoguerios: Thanks for highlighting this one. It uncovered another issue.
Basically indexer and standalone tycho-common would have inconsistent behavior when loading these configs in case there was a duplicated entry.
Now they they are centralized in the same validation process.
And since a duplicate entry is clearly an error, I took one step further and made it an error instead of a warning.

crates/tycho-common/src/models/chain_config.rs
brunoguerios: I'd rather apply this refactor in the last PR on the stack to avoid too much rework - is that ok? 🙃

brunoguerios: Giving a bit more context, while evaluating making things more private, I felt like it could make sense to remove the global registry. If everything is accessible through Chain, then why have a global registry to access custom chain configs? Does that make sense?

crates/tycho-common/src/models/mod.rs
zizou0x: In general what you said above is correct

The reason why I think it's fine in that case is because `CustomChainConfig` isn't in main yet, it's still part of the release branch. So we're still free to change it as we want right now.

crates/tycho-common/src/models/chain_config.rs
```diff
@@ -254,10 +256,13 @@ impl ChainConfigRegistry {
...
-/// Returns the process-wide chain config registry, lazily loading it from the configured file on
-/// first access (see [`ChainConfigRegistry::load_default`]).
+/// Returns the process-wide chain config registry.
+///
+/// Defaults to an empty registry — only built-in chains resolve — unless one was installed with
+/// [`init_chain_registry`]. This never reads from disk; load custom chains explicitly via
+/// [`ChainConfigRegistry::load_default`] and install them before first access.
 pub fn chain_registry() -> &'static ChainConfigRegistry {
-    CHAIN_REGISTRY.get_or_init(ChainConfigRegistry::load_default)
+    CHAIN_REGISTRY.get_or_init(ChainConfigRegistry::empty)
```
zizou0x, line 265: I'd prefer to keep the old behaviour with default here. Given that it's a `OnceLock` I find it better to try its best to fill the registry with whatever it has. 

It's also less constraints on whoever uses it. We can easily make this return a `Result<>` too, no?

crates/tycho-common/src/models/chain_config.rs
zizou0x: Yes that's ok for me 👍 

I think I'd still tend toward keeping the registry as an internal component (public interface isn't the only motivation to have it as a component)

zizou0x (commented):
One last comment but looks good!

crates/tycho-common/src/models/chain_config.rs
brunoguerios: Problem is that the old behavior of lazy loading gets us back to where we were by allowing panic to happen down the line, instead of forcing it to happen during initialization.

Making it return a Result is also not viable due to the limitations we faced in the other PR, as we'd have to handle the error in multiple places.

Making it empty by default and then requiring loading as an explicit fallible step handled at startup feels more aligned with the whole proposed solution of failing early/fast and reducing the scenarios it can panic.

zizou0x (approved):
LGTM thanks! 🚀

PR #1117, opened 2026-06-24
feat: represent custom chains by name, resolve config via registry

brunoguerios:
# Summary

Depends on PR 1. Flips Chain::Custom to name-only and wires resolution through the registry. The enum  change forces all custom-chain construction/match sites to change together, so the core commit is atomic; the rest are additive. The ~60 chain.id()/native_token()/... call sites in tycho-simulation/tycho-execution/tests are untouched because the method signatures are preserved.

# Changes

- models::Chain::Custom(ArrayString<32>) and dto::Chain::Custom(ArrayString<32>) (stay Copy; default serde => known "ethereum", custom {"custom":"name"}; #[schema(value_type = String)] on the payload).
- Custom arms of id/default_tvl_threshold/native_token/wrapped_native_token read chain_registry() and panic with a clear message if absent; signatures unchanged. Add Chain::block_time_secs() for the client. Collapse native_custom/wrapped_native_custom into token_from(&ChainTokenConfig, Chain).
- FromStr: known => variant; else registry.contains => Custom; else Err(UnknownChain). From impls become passthrough Custom(name) => Custom(name).
- Indexer: init_chain_registry from the existing chains: config; replace resolve_chain with registry-aware parsing; validate configured chains at startup with actionable errors.
- Client: stream.rs default_timing uses block_time_secs(). Storage: chain-cache reconstruction works via registry-aware FromStr.
- Python client decodes {"custom":"name"}; drop the CustomChainConfig / ChainTokenConfig / TvlThresholds pydantic models and exports; update test_decode.py.
- Drop the 3 custom-chain schemas from OpenAPI (api_docs.rs).
- Remove the obsolete result_large_err / large_enum_variant clippy allows.

# Acceptance criteria

- `size_of::<Chain>() <= 40`; serde round-trips "ethereum" and {"custom":"x"}.
- FromStr handles known/registry/typo cases; panic on missing custom config has a clear message.
- Known chains work with no config file; indexer rejects unknown chains at startup.
- pytest crates/tycho-client-py passes; clippy on nightly with -D warnings is clean; CI green.

### DIFF
--- a/.github/workflows/ci-rust.yaml
+++ b/.github/workflows/ci-rust.yaml
@@ -66,7 +66,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@3c5f7ea28cd621ae0bf5283f0e981fb97b8a7af9  # nightly
         id: toolchain
         with:
-          toolchain: nightly
+          toolchain: nightly-2026-06-28
           components: rustfmt, clippy
 
       - uses: Swatinem/rust-cache@9d47c6ad4b02e050fd481d890b2ea34778fd09d6  # v2.7.8
--- a/crates/tycho-client-py/python/tycho_indexer_client/dto.py
+++ b/crates/tycho-client-py/python/tycho_indexer_client/dto.py
@@ -48,36 +48,21 @@ class Chain(str, Enum):
     polygon = "polygon"
 
 
-class ChainTokenConfig(BaseModel):
-    address: str
-    symbol: str
-    decimals: int
+class CustomChain(BaseModel):
+    """A user-defined chain. Only the chain name travels on the wire; the full config lives
+    server-side. Rust serialises ``Custom("name")`` as ``{"custom": "name"}``."""
 
-
-class TvlThresholds(BaseModel):
-    low: float
-    medium: float
-
-
-class CustomChainConfig(BaseModel):
     name: str
-    chain_id: int
-    block_time_secs: int
-    native: ChainTokenConfig
-    wrapped_native: ChainTokenConfig
-    default_tvl_thresholds: TvlThresholds
 
     def __str__(self) -> str:
         return self.name
 
     @root_validator(pre=True)
     def _unwrap_serde_tag(cls, values):
-        # Rust serialises Custom(cfg) as {"custom": {...}} — unwrap the tag.
+        # Rust serialises Custom("name") as {"custom": "name"} — unwrap the tag to the name.
         if "custom" in values and len(values) == 1:
-            return values["custom"]
+            return {"name": values["custom"]}
         return values
-
-
 
 
 class ChangeType(str, Enum):
@@ -88,7 +73,7 @@ class ChangeType(str, Enum):
 
 
 class ExtractorIdentity(BaseModel):
-    chain: Union[Chain, CustomChainConfig]
+    chain: Union[Chain, CustomChain]
     name: str
 
 
@@ -125,13 +110,13 @@ class Block(BaseModel):
     number: int
     hash: HexBytes
     parent_hash: HexBytes
-    chain: Union[Chain, CustomChainConfig]
+    chain: Union[Chain, CustomChain]
     ts: datetime
 
 
 class BlockParam(BaseModel):
     hash: Optional[HexBytes] = None
-    chain: Optional[Union[Chain, CustomChainConfig]] = None
+    chain: Optional[Union[Chain, CustomChain]] = None
     number: Optional[int] = None
 
 
@@ -152,7 +137,7 @@ class TokenBalances(BaseModel):
 
 class AccountUpdate(BaseModel):
     address: HexBytes
-    chain: Union[Chain, CustomChainConfig]
+    chain: Union[Chain, CustomChain]
     slots: Dict[HexBytes, HexBytes]
     balance: Optional[HexBytes] = None
     code: Optional[HexBytes] = None
@@ -169,7 +154,7 @@ class ProtocolComponent(BaseModel):
     id: str
     protocol_system: str
     protocol_type_name: str
-    chain: Union[Chain, CustomChainConfig]
+    chain: Union[Chain, CustomChain]
     tokens: List[HexBytes]
     contract_ids: List[HexBytes]
     static_attributes: Dict[str, HexBytes]
@@ -179,7 +164,7 @@ class ProtocolComponent(BaseModel):
 
 
 class ResponseToken(BaseModel):
-    chain: Union[Chain, CustomChainConfig]
+    chain: Union[Chain, CustomChain]
     address: HexBytes = Field(..., example="0xc9f2e6ea1637E499406986ac50ddC92401ce1f58")
     symbol: str = Field(..., example="WETH")
     decimals: int
@@ -190,7 +175,7 @@ class ResponseToken(BaseModel):
 
 class BlockChanges(BaseModel):
     extractor: str
-    chain: Union[Chain, CustomChainConfig]
+    chain: Union[Chain, CustomChain]
     block: Block
     finalized_block_height: int
     revert: bool
@@ -218,7 +203,7 @@ class ComponentWithState(BaseModel):
 
 
 class ResponseAccount(BaseModel):
-    chain: Union[Chain, CustomChainConfig]
+    chain: Union[Chain, CustomChain]
     address: HexBytes
     title: str
     slots: Dict[HexBytes, HexBytes]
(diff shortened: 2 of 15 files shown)


crates/tycho-client/src/feed/mod.rs
```diff
@@ -100,7 +100,6 @@ impl HeaderLike for BlockHeader {
     }
 }
 
-#[allow(clippy::large_enum_variant)]
```
zizou0x, line 103: Beautiful!

crates/tycho-client/src/stream.rs
```diff
@@ -117,7 +117,10 @@ impl TychoStreamBuilder {
             Chain::Bsc => (1, 12, 50),
             Chain::Unichain => (1, 10, 100),
             Chain::Polygon => (2, 12, 50), // ~2s block time
-            Chain::Custom(cfg) => (cfg.block_time_secs, cfg.block_time_secs * 3, 50),
+            Chain::Custom(_) => {
+                let block_time = chain.block_time_secs();
```
zizou0x, line 121: Nice, I think that will help addressing one comment on the other PR to make things more private

crates/tycho-common/src/models/mod.rs
```diff
@@ -209,13 +256,17 @@ fn wrapped_native_custom(cfg: &CustomChainConfig) -> Token {
...
         &[Some(2300)],
-        Chain::Custom(*cfg),
+        chain,
         100,
     )
 }
 
 impl Chain {
     pub fn id(&self) -> u64 {
+        self.id_in(chain_registry())
```
zizou0x: This is safe because we don't allow creating a Chain::Custom() if there isn't anything for it in the registry, correct?

crates/tycho-common/src/models/mod.rs
```diff
@@ -316,7 +391,26 @@ impl Chain {
...
+        match self {
+            Chain::Ethereum => 12,
+            Chain::Starknet => 2,
+            Chain::ZkSync => 3,
+            Chain::Arbitrum => 1,
+            Chain::Base => 2,
+            Chain::Bsc => 1,
+            Chain::Unichain => 1,
+            Chain::Polygon => 2,
+            Chain::Custom(name) => resolve_custom(name, registry).block_time_secs,
```
zizou0x: Do we need these pass through functions? I have a feeling there were made for testing purpose, but I think we could just do 

```
init_chain_registry(test_config)
chain.block_time_secs()
```

crates/tycho-client-py/python/tycho_indexer_client/dto.py
```diff
@@ -48,38 +48,23 @@ class Chain(str, Enum):
...
-    wrapped_native: ChainTokenConfig
-    default_tvl_thresholds: TvlThresholds
 
     def __str__(self) -> str:
         return self.name
 
     @root_validator(pre=True)
     def _unwrap_serde_tag(cls, values):
-        # Rust serialises Custom(cfg) as {"custom": {...}} — unwrap the tag.
+        # Rust serialises Custom("name") as {"custom": "name"} — unwrap the tag to the name.
```
zizou0x, line 62: Any particular reason to not change rust serialization to just "name"?

crates/tycho-indexer/src/main.rs
```diff
@@ -270,10 +266,16 @@ fn run_indexer(global_args: GlobalArgs, index_args: IndexArgs) -> Result<(), Ext
                 .parse()
                 .expect("Failed to parse retention horizon");
 
+            let chain_registry =
+                ChainConfigRegistry::from_configs(extractors_config.chains.clone());
```
zizou0x: I know in the plan it was mentioned to add this in the extractor config but with the current changes I wonder if it wouldn't be cleaner to have this in a separate chains.yaml. That way we could simplify things here and just keep using `Chain::from_str`

zizou0x (commented):
Looks so much better! Thanks for tackling this! Just have a few comments

crates/tycho-client-py/python/tycho_indexer_client/dto.py
brunoguerios: I actually already made that change on the following PR to simplify another area of the code ✅  🙃

crates/tycho-indexer/src/main.rs
brunoguerios: I challenged Claude about this while working on the PR and it said there's benefit in keeping configs in a single file. I think that separating in 2 config files makes more sense as well. I'll go ahead and do it.

crates/tycho-common/src/models/mod.rs
brunoguerios: Yeah, those were mostly created for tests, in order to prevent OnceLock issues in case they end up running multiple `init_chain_registry` on the same process.
Since we rely in nextest it should be fine because it starts a new process for each test.
I'll refactor them here.

crates/tycho-indexer/src/main.rs
brunoguerios: done ✅

crates/tycho-common/src/models/mod.rs
brunoguerios: done ✅

crates/tycho-common/src/models/mod.rs
brunoguerios: I spent a bit more time on this and it's actually not that safe 🫠 
I'd say it might be safe in practice because we expect this whole feature to be used by teams that will spin up their own indexer and likely reuse the same chains.yaml file as registry in all its processes.
But if they end up not loading it correctly downstream and access custom chain data through `From<dto::Chain>`, it will panic 😬 
This might be an argument to move away from the registry approach and rely on Chain only?
Let me know your thoughts on this.
Ps: if we decide to keep the registry approach and want to be safer, we can refactor `From<dto::Chain>` as `TryFrom`, but that has a significantly larger impact on the codebase 👀

.github/workflows/ci-rust.yaml
```diff
@@ -66,7 +66,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@3c5f7ea28cd621ae0bf5283f0e981fb97b8a7af9  # nightly
         id: toolchain
         with:
-          toolchain: nightly
+          toolchain: nightly-2026-06-28
```
zizou0x, line 69: I think this has been addressed on main, you might have conflicts now.

Usually the process if CI is broken because of external reasons is to fix it directly on main in another PR and rebase your PR at the end.

crates/tycho-common/src/models/mod.rs
zizou0x: Yeah this is quite bad, because it panics. Changing all these interfaces would also be a huge change. 

The only fix that I would see would be to "gate" the `Chain::Custom` creation to check that it's in the registry and error at creation time, lmk if you have another idea? How would moving out of the registry help?

zizou0x (commented):
Nice progress!

.github/workflows/ci-rust.yaml
brunoguerios: Yeah, I didn't want to rebase the whole feature branch to main just because of that small (likely temporary) change. I cherry picked it because ci was filing with the same root cause and the fix was trivially small. Will revert over here ✅

crates/tycho-common/src/models/mod.rs
brunoguerios: Yeah - definitely not trivial to get rid of this panic 😅 
When I said to move away from the registry approach I meant the need to have the registry on both the indexer and on downstream consumers. We can't enforce they will have the same registry available and that is error prone.
Going back to the original design that attached chain_config to Chain itself, the main benefit was that the chain config was set on the indexer and consumers would simply get all that from within Chain.
How about we get back to finding a way to have the indexer instantiate custom chains and then provide them to customers. If we can't have that within Chain, let's provide it in a separate component.
For that we could use a separate endpoint - resolving it at consumer initialization - which ensures it's fallible only at the start and panic becomes no longer reachable when accessing custom chain data.
What do you think?

brunoguerios: As discussed in slack, we'll defer the solution to provide custom chains via separate endpoint as a future improvement.
For now we've settled in an intermediary solution gating Chain::Custom at creation - which keeps panic within tycho-common, but fails early/fast and reduces surface for mistakes.
I'll create a doc with context on all these problems/solutions to ground and justify the decisions we made.

zizou0x (approved):
LGTM!

PR #1119, opened 2026-06-25
chore: Update contracts' bytecode for testing + automated script

dianacarvalho1:
[Task](https://propeller-heads.atlassian.net/jira/software/projects/ENG/boards/24?selectedIssue=ENG-6046)

Done in separate commits:
- Create executor_deployments.json. This new json has all the executors per chain and their respective deployment arguments (this is needed by the deploy-executors script and the new script coming in the next step)
- Create script to regenerate contracts bytecode for testing. Added it to CI to avoid having mismatches in the future
- Update contracts' bytecode to latest version for testing. Add missing fee calculator logic

I ran the test suite locally only for on vm protocol (balancer v2) and a native one (univ2) and everything was ✅


crates/tycho-execution/contracts/scripts/deploy-executors.js
```diff
@@ -228,7 +108,7 @@ async function main() {
         // CREATE2 address = keccak256(0xff ++ factory_address ++ salt ++ keccak256(bytecode))[12:]
         const bytecodeHash = ethers.utils.keccak256(bytecode);
         const computedAddress = ethers.utils.getCreate2Address(create2FactoryAddress, salt, bytecodeHash);
-        console.log(`${exchange} will be deployed to: ${computedAddress}`);
+        console.log(`${exchange} (${protocol}) will be deployed to: ${computedAddress}`);
```
louise-poole: What's the difference between exchange and protocol here?

protocols/testing/scripts/update_runtime_bytecode.sh
```diff
@@ -0,0 +1,197 @@
...
+    "BalancerV2|ethereum|vm:balancer_v2"
+    "BalancerV3|ethereum|vm:balancer_v3"
+    "Curve|ethereum|vm:curve"
+    "FermiSwap|ethereum|vm:fermiswap"
+    "MaverickV2|ethereum|vm:maverick_v2"
+    "EkuboV3|ethereum|ekubo_v3"
+    "FluidV1|ethereum|fluid_v1"
+    "LiquidityParty|ethereum|vm:liquidityparty"
+    "LunarBase|base|lunarbase"
+)
```
louise-poole, line 91: Is it intentional that we don't have fixtures for all protocols? Off the top of my head I see Rocketpool and Aerodrome etc missing...

protocols/testing/scripts/update_runtime_bytecode.sh
```diff
@@ -0,0 +1,197 @@
...
+# the bytecode, on every run. Pinning the block fixes the nonce and makes
+# generation deterministic, which --check relies on.
+FORK_BLOCK=21000000
+
+PERMIT2="0x000000000022D473030F116dDEE9F6B43aC78BA3"
+ADMIN="0x0000000000000000000000000000000000000001"
+
+# fixture | contract | literal constructor args
+NON_EXECUTOR_FIXTURES=(
+    "TychoRouter|TychoRouter|$PERMIT2 $PERMIT2 $ADMIN $ADMIN $ADMIN $ADMIN"
```
louise-poole, line 69: Why PERMIT2 twice? 👀

crates/tycho-execution/contracts/scripts/deploy-executors.js
```diff
@@ -2,203 +2,71 @@ require('dotenv').config();
...
-                "0x1f98400000000000000000000000000000000004",
-                // This is the Angstrom address for ethereum. There isn't one for unichain
-                "0x0000000aa232009084Bd71A5797d089AA4Edfad4"
-            ]
-        },
-        // Args: ETH address in curve pools, stETH address
-        {
-            exchange: "CurveExecutor", args: [
-                "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE",
-                "0x0000000000000000000000000000000000000000" // No stETH on unichain
```
louise-poole, line 133: The one down side of moving this all to a json is that we cannot have explanatory comments like this. Not critical in this case, but kinda sad anyway.

.github/workflows/ci-foundry.yaml
```diff
@@ -46,9 +47,20 @@ jobs:
...
+      # Inputs to the protocol-testing runtime bytecode fixtures.
+      - uses: tj-actions/changed-files@48d8f15b2aaa3d255ca5af3eba4870f807ce6b3c  # v45
+        id: filter-bytecode
+        with:
+          files: |
+            crates/tycho-execution/contracts/src/**
+            crates/tycho-execution/config/executor_deployments.json
+            protocols/testing/fixtures/**
+            protocols/testing/scripts/update_runtime_bytecode.sh
+            .github/workflows/ci-foundry.yaml
```
louise-poole: I don't think we need to trigger this update bytecode fixtures workflow if we edit this workflow file, no? We should just trigger when the executors, deployment params or bytecode generation script are edited (all the other files you already have here).

crates/tycho-execution/contracts/scripts/deploy-executors.js
```diff
@@ -213,8 +81,20 @@ async function main() {
     const create2FactoryAddress = "0x4e59b44847b379578588920cA78FbF26c0B4956C";
     console.log(`Using CREATE2 factory at: ${create2FactoryAddress}`);
 
-    for (const executor of executors_to_deploy[network]) {
-        const {exchange, args} = executor;
+    const protocols = deploy_protocols[network];
+    if (!protocols) {
+        throw new Error(`No deploy protocols configured for network: ${network}`);
+    }
+    const networkDeployments = executorDeployments[network] || {};
```
louise-poole: We should error here instead of fall back to en empty list no?

louise-poole (approved):
Was about time we did this - thanks!!!!! 🙏  Just a couple of small comments, otherwise have approved!

protocols/testing/scripts/update_runtime_bytecode.sh
dianacarvalho1: they were never added 😕 I only replaced the existing ones.. didn't add new tests

crates/tycho-execution/contracts/scripts/deploy-executors.js
dianacarvalho1: yeah agree 😕 but I still think this is nicer this way..

crates/tycho-execution/contracts/scripts/deploy-executors.js
dianacarvalho1: exchange is actually the contract name. let me rename it for clarity!

protocols/testing/scripts/update_runtime_bytecode.sh
dianacarvalho1: the second one it is acting as a placeholder for the fee calculator! I'm adding a comment to clarify!

tamaralipows (approved):
Thank you! Much needed!!

PR #1120, opened 2026-06-25
feat: update router and encoder to new FeeCalculator interface

louise-poole:
https://propeller-heads.atlassian.net/browse/ENG-6054

Updates the Router and encoding to reflect the latest FeeCalculator interface changes (done [here](https://github.com/propeller-heads/tycho/pull/1115)).

Replaces `minAmountOut` with `expectedAmountOut` (quoted) + `maxSlippageBps` across all 9 public swap functions in TychoRouter. The quoted `amountOut` is also passed to `FeeCalculator.calculateFee`, enabling the fee calculator to compute slippage surplus.


crates/tycho-execution/contracts/src/TychoRouter.sol
```diff
@@ -181,35 +182,38 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
...
      *
-     * @return amountOut The total amount of the output token received by the receiver.
+     * @return The total amount of the output token received by the receiver.
      */
     function splitSwap(
         uint256 amountIn,
         address tokenIn,
         address tokenOut,
-        uint256 minAmountOut,
+        uint256 amountOut,
```
tamaralipows: should we call this rather `expectedAmountOut` to avoid confusion with exact output swaps which would give exactly this amount? Or do you think the docstring is enough? No strong opinions. I like the simplicity of this name if you think it's clear enough

crates/tycho-execution/contracts/src/TychoRouter.sol
```diff
@@ -671,12 +708,13 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
...
+        uint16 maxSlippageBps,
         uint256 nTokens,
         address receiver,
         ClientFeeParams calldata clientFeeParams,
         bytes calldata swaps
-    ) internal returns (uint256 amountOut) {
+    ) internal returns (uint256 actualOut) {
```
tamaralipows: I'm more used to seeing the "expectedOut" and "amountOut" rather than "amountOut" and "actualOut" naming. amountOut and actualOut can also be easier confused at a glance since they have roughly the same length and first letters lol. Would you be fine changing it at least for these inner methods?

crates/tycho-execution/contracts/src/TychoRouter.sol
```diff
@@ -686,18 +724,20 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
...
+            revert TychoRouter__AmountOutZero();
         }
-
-        uint16 routerFeeOnOutputBps = _callGetEffectiveRouterFeeOnOutput(
-            _feeCalculator, clientFeeParams.clientFeeReceiver
-        );
+        if (maxSlippageBps > 10_000) {
+            revert TychoRouter__SlippageBpsTooHigh(maxSlippageBps);
+        }
+        uint256 effectiveMin = amountOut * (10_000 - maxSlippageBps) / 10_000;
```
tamaralipows: We will also need to update this in the docs, but I guess that's probably a later task. It might be easier to have claude do this while we go though?

crates/tycho-execution/contracts/src/TychoRouter.sol
```diff
@@ -708,29 +748,25 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
...
+            amountOutBeforeFees,
             amountOut,
-            minAmountOut,
+            clientFeeParams.clientFeeBps,
+            clientFeeParams.clientFeeReceiver
+        );
+
+        actualOut = _maybeAddClientContribution(
+            actualOut,
+            effectiveMin,
```
tamaralipows: `minAmountOut`?

crates/tycho-execution/contracts/src/TychoRouter.sol
```diff
@@ -762,48 +799,45 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
...
-                amountOutBeforeFees,
-                clientFeeParams.clientFeeBps,
-                client
-            );
-        }
-
-        amountOut = _maybeAddClientContribution(
+        actualOut = _takeFees(
+            tokenOut,
+            amountOutBeforeFees,
```
tamaralipows: I see we have removed the check to see if calling the fee calculator is necessary. Perhaps we could infer this from `mustInterceptOutput`? We actually save a massive amount of gas by not calculating fees if unnecessary (like if they are waived for a certain client).

But this depends on if we can do that while remaining readable

crates/tycho-execution/contracts/src/TychoRouter.sol
```diff
@@ -1172,20 +1204,27 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
...
      * @param clientFeeBps Client fee in basis points
      * @param client Address to receive client fees
      * @return amountOut The amount remaining after all fee deductions
      */
     function _takeFees(
         address token,
-        uint256 amountIn,
-        uint16 clientFeeBps,
+        uint256 grossAmountOut,
+        uint256 quotedAmountOut,
```
tamaralipows: `expectedAmountOut`?

crates/tycho-execution/contracts/src/TychoRouter.sol
```diff
@@ -83,7 +84,7 @@ error TychoRouter__ExpiredClientSignature(
 error TychoRouter__ZeroInput();
 
 struct ClientFeeParams {
-    uint16 clientFeeBps;
+    uint32 clientFeeBps;
```
tamaralipows, line 87: nice

tamaralipows (commented):
Thank you Louise! Everything looks sound! ✨ I only have minor naming nitpicks. Open to discussion, I've just been used to certain namings for so long

crates/tycho-execution/contracts/src/TychoRouter.sol
```diff
@@ -181,35 +182,38 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
...
         address tokenOut,
-        uint256 minAmountOut,
+        uint256 amountOut,
+        uint16 maxSlippageBps,
         uint256 nTokens,
         address receiver,
         ClientFeeParams calldata clientFeeParams,
         bytes calldata swaps
-    ) public payable whenNotPaused nonReentrant returns (uint256 amountOut) {
+    ) public payable whenNotPaused nonReentrant returns (uint256) {
```
Troshchk, line 210: why did you remove `amountOut` here?

crates/tycho-execution/contracts/src/TychoRouter.sol
```diff
@@ -686,18 +724,20 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
...
-        if (minAmountOut == 0) {
-            revert TychoRouter__UndefinedMinAmountOut();
+        if (amountOut == 0) {
+            revert TychoRouter__AmountOutZero();
         }
-
-        uint16 routerFeeOnOutputBps = _callGetEffectiveRouterFeeOnOutput(
-            _feeCalculator, clientFeeParams.clientFeeReceiver
-        );
+        if (maxSlippageBps > 10_000) {
```
Troshchk: I would move 10_000 to a named constant for clarity and since it is used several times. 
I have added it in FeeCalculator.sol, may be can be reused?
```
    // 100% in basis points (for slippage share validation)
    uint16 public constant MAX_SLIPPAGE_SHARE_BPS = 10_000;
```

crates/tycho-execution/contracts/src/TychoRouter.sol
```diff
@@ -762,48 +799,45 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
         ) {
             revert TychoRouter__AddressZero();
         }
-        if (minAmountOut == 0) {
-            revert TychoRouter__UndefinedMinAmountOut();
+        if (amountOut == 0) {
+            revert TychoRouter__AmountOutZero();
+        }
+        if (maxSlippageBps > 10_000) {
```
Troshchk: Same thing about the constant

crates/tycho-execution/contracts/src/TychoRouter.sol
```diff
@@ -835,47 +870,44 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
...
-                clientFeeParams.clientFeeBps,
-                client
-            );
-        }
-
-        amountOut = _maybeAddClientContribution(
+        actualOut = _takeFees(
+            tokenOut,
+            amountOutBeforeFees,
             amountOut,
```
Troshchk: I like `amountOutBeforeFees` name, I would rename `amountOut` for clarity too

crates/tycho-execution/contracts/src/TychoRouter.sol
Troshchk: I actually like the names here, I think it makes the difference between the two amounts clear

crates/tycho-execution/contracts/src/TychoRouter.sol
```diff
@@ -1172,20 +1204,27 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
...
-        (amountOut, fees) =
-            _callCalculateFee(_feeCalculator, amountIn, clientFeeBps, client);
+        FeeRecipient[] memory fees = _callCalculateFee(
+            _feeCalculator,
+            grossAmountOut,
+            quotedAmountOut,
+            clientFeeBps,
+            client
+        );
+        amountOut = grossAmountOut;
```
Troshchk: why is this needed?

crates/tycho-execution/tests/sequential_strategy_integration_tests.rs
```diff
@@ -148,16 +150,17 @@ fn test_sequential_swap_strategy_encoder_transfer_from_integration() {
...
+        "00000000000000000000000000000000000000000000000000000000000000c8", // max_slippage_bps=200
         "000000000000000000000000cd09f75e2bf2a4d11f3ab23f1389fcc1621c0cc2", // receiver
-        "00000000000000000000000000000000000000000000000000000000000000e0", /* clientFeeParams
-                     * offset = 224 */
-        "00000000000000000000000000000000000000000000000000000000000001a0", /* swapData offset =
-                                                                             * 416 */
+        "0000000000000000000000000000000000000000000000000000000000000100", /* clientFeeParams
+                     * offset = 256 */
+        "00000000000000000000000000000000000000000000000000000000000001c0", /* swapData offset =
+                                                                             * 448 */
```
Troshchk: I saw this formatting, but the added line looks a bit weird to me

crates/tycho-execution/contracts/src/TychoRouter.sol
```diff
@@ -181,35 +182,38 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
      *
      * @dev
      * - Swaps are executed sequentially using the `_swap` function.
-     * - Reverts with `TychoRouter__NegativeSlippage` if the output amount is less than `minAmountOut`
+     * - Reverts with `TychoRouter__NegativeSlippage` if the output amount is less than `effectiveMin`
```
Troshchk: `effectiveMin` is amount, right? I would add amount to the name to make it explicit.

crates/tycho-execution/contracts/src/TychoRouter.sol
```diff
@@ -181,35 +182,38 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
...
      *
      * @param amountIn The input token amount to be swapped.
      * @param tokenIn The address of the input token. Use `ETH_ADDRESS` for native ETH
      * @param tokenOut The address of the output token. Use `ETH_ADDRESS` for native ETH
-     * @param minAmountOut The minimum acceptable amount of the output token. Reverts if this condition is not met. This should always be set to avoid losing funds due to slippage.
+     * @param amountOut The quoted output amount used to compute slippage.
+     * @param maxSlippageBps Maximum slippage in basis points (0-10000).
```
Troshchk: may be 10_000 for readobility?

Troshchk (commented):
Thank you @louise-poole! Overall it looks good to me! 💫 

I agree with Tamara about the naming. We currently have more calculations involving amountOut and tbh I will not mind if the variable names are longer, but more explicit. At least for me, it makes reading the code easier.

Just to summarize, these are the amount related names, which I think could be named better:
`effectiveMin`
`amountOut` in `splitSomething` functions
`actualOut`

crates/tycho-execution/contracts/src/TychoRouter.sol
tamaralipows: It irks me that they're two different descriptors.

1. "gross/net" describes before or after the fee (while we are also using `amountOutBeforeFees` in other places, which I prefer)
2. "quoted/actual" describes whether it's what we actually received or what we expected (im much more used to the expected language than quoted but I won't insist)

and so we are mixing the two concepts here kindof, while also using language that I'm not used to seeing in the codebase. I get it after using my brain a bit but not at first glance. But this is a democracy so if you both like it, I will accept xD

crates/tycho-execution/contracts/src/TychoRouter.sol
louise-poole: I personally felt it was clear enough with the slippage param indicating slippage may apply and I might not get this exact amountOut. I didn't want to make long convoluted param names for these public interfaces, but am happy to adjust if you guys prefer?

crates/tycho-execution/contracts/src/TychoRouter.sol
louise-poole: I also don't want to add unnecessary overhead for our auditor... so while I agree with Ksenia about this feeling clearer to me as a newer contract writer, I think sticking to the language that is already used is probably better.

crates/tycho-execution/contracts/src/TychoRouter.sol
louise-poole: So I had a long think about this - it seems to be very rare that we do not take fees anymore, especially now that we're adding positive slippage taking to the fee logic. This intercept output check is done before we know if the swap actually has positive slippage or not, so we have to take a jump and decide to always intercept if globally the slippage taking is toggled on.
However having said all this I am happy to add back a similar check if you feel it's important? It's not adding too much complexity so I don't see much harm in keeping it.

crates/tycho-execution/contracts/src/TychoRouter.sol
louise-poole: Because you cannot use the same named var as an input and output param. We added `amountOut` to our input params so decided the returned output param could be unnamed to avoid issues.

crates/tycho-execution/contracts/src/TychoRouter.sol
louise-poole: Because the fn signature has: `) internal returns (uint256 amountOut) {`. This has a named returned var. In the line `amountOut = grossAmountOut;` I am populating the returned var with the value in `grossAmountOut`

crates/tycho-execution/tests/sequential_strategy_integration_tests.rs
louise-poole: Can you clarify what you mean here? I'm afraid I don't understand...

Troshchk: I mean `/* clientFeeParams
                     * offset = 256 */` and `/* swapData offset =
                                                                             * 448 */` being split in two lines

crates/tycho-execution/contracts/src/TychoRouter.sol
Troshchk: All clear, thanks!

crates/tycho-execution/contracts/src/TychoRouter.sol
```diff
@@ -181,35 +183,38 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
      *
      * @dev
      * - Swaps are executed sequentially using the `_swap` function.
-     * - Reverts with `TychoRouter__NegativeSlippage` if the output amount is less than `minAmountOut`
+     * - Reverts with `TychoRouter__NegativeSlippage` if slippage exceeds `maxSlippageBps`
```
Troshchk: why did you change this?
I think if slippage exceeds `maxSlippageBps`, may be the error name should be different? Like `TychoRouter__TooHighSlippage`

Troshchk (commented):
Thank you @louise-poole! 💫 
Left a couple of qs

crates/tycho-execution/contracts/src/TychoRouter.sol
```diff
@@ -181,35 +183,38 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
...
      *
      * @param amountIn The input token amount to be swapped.
      * @param tokenIn The address of the input token. Use `ETH_ADDRESS` for native ETH
      * @param tokenOut The address of the output token. Use `ETH_ADDRESS` for native ETH
-     * @param minAmountOut The minimum acceptable amount of the output token. Reverts if this condition is not met. This should always be set to avoid losing funds due to slippage.
+     * @param amountOut The quoted output amount used to compute slippage.
```
Troshchk: What do you think about not having pure `amountOut` name anywhere?
If this is a quoted one, I would rename it into `expectedAmountOut` so it is the same in the codebase. Let me know if I am missing something here

crates/tycho-execution/contracts/src/TychoRouter.sol
louise-poole: > why did you change this?

Because the user no longer sets a `minAmountOut`, we calculate it internally. And I do not think a doc string should reference an internal var name because then the user needs to read code to understand what it means.

> I think if slippage exceeds maxSlippageBps, may be the error name should be different? Like TychoRouter__TooHighSlippage

Really? To me, maxSlippageBps defines our slippage tolerance. If the actual slippage is within tolerance we consider the amount out to be within the expected range for the swap and do not report any slippage back to the user (as is the current behaviour when minAmountOut is passed). However if the negative slippage exceeds tolerance then we flag it as negative slippage. 

On that note, how about `TychoRouter__SlippageExceeded`?

Troshchk (approved):
Nice, thank you!

crates/tycho-execution/contracts/src/TychoRouter.sol
```diff
@@ -1285,23 +1356,17 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
...
         address receiver,
-        uint16 clientFeeBps,
-        uint32 routerFeeOnOutputBps
+        uint32 clientFeeBps,
+        address client
     ) internal view returns (address) {
-        // Fast path: if no fees at all, send directly to receiver
-        if (clientFeeBps == 0 && routerFeeOnOutputBps == 0) {
-            return receiver;
-        }
```
dianacarvalho1, line 1302: this is just a pass through method now?? In the `_splitSwapChecked` it's not being used anymore and `_callMustInterceptOutput` is being called directly. can we be coherent in all methods please?

crates/tycho-execution/src/encoding/models.rs
```diff
@@ -133,23 +133,28 @@ pub struct Solution {
...
     /// The token being bought
     token_out: Bytes,
-    /// Minimum amount that the receiver must receive at the end of the transaction.
+    /// Quoted output amount from simulation. Used as the baseline for slippage:
+    /// effective min = amount_out * (10000 - max_slippage_bps) / 10000.
     #[serde(with = "biguint_string")]
-    min_amount_out: BigUint,
+    amount_out: BigUint,
+    /// Maximum slippage in basis points (0–10000).
+    max_slippage_bps: u16,
```
dianacarvalho1: why are we using u16 here in the user interface? wouldn't it be nicer to use a float so that users don't need to worry about weird bps scales?

dianacarvalho1: also.. I just realised we don't really need these attributes inside of the Solution 🤔 they will be used only in the final step of encoding the full tycho router call

crates/tycho-simulation/examples/quickstart/main.rs
```diff
@@ -682,6 +682,7 @@ fn create_solution(
         buy_token.address,
         sell_amount,
         min_amount_out,
+        0u16,
```
dianacarvalho1: slippage 0 is a bit too ambitions and might lead to bad reverts

crates/tycho-simulation/examples/quickstart/main.rs
dianacarvalho1: this needs to be changed too

dianacarvalho1: same in the other RFQ quickstart

crates/tycho-simulation/examples/quickstart/main.rs
```diff
@@ -682,6 +682,7 @@ fn create_solution(
         buy_token.address,
         sell_amount,
         min_amount_out,
```
dianacarvalho1: this should be the real amount out from simulation, so the `expected_amount` at this point

crates/tycho-test/src/execution/encoding.rs
```diff
@@ -124,6 +124,7 @@ pub fn create_solution(
         // We want to keep track of how bad the slippage really is and not just error at execution
         // time. NEVER DO THIS IN PRODUCTION!
         BigUint::from(1u64),
+        0u16,
```
dianacarvalho1: huummm I'm not sure this will work anymore 🤔 I think we should pass the simulated amount out and then just add a big slippage tolerance anyway

crates/tycho-execution/examples/encoding-example/main.rs
```diff
@@ -57,6 +57,7 @@ fn main() {
         usdc.clone(),
         BigUint::from_str("1_000000000000000000").expect("Failed to create amount"),
         BigUint::from(1u64),
```
dianacarvalho1: can we make this a nicer value.. 1 is not ok anymore. Also probably add a comment that this needs to come from simulation

crates/tycho-execution/contracts/test/TychoRouterFees.t.sol
```diff
@@ -18,7 +18,7 @@ contract TychoRouterFeesTest is TychoRouterTestSetup {
 
     // TODO(ENG-6054): Rename skip_ → test once _takeFees is wired to
     // the new _callCalculateFee signature.
```
dianacarvalho1: remove TODO

crates/tycho-execution/contracts/test/TychoRouterSequentialSwap.t.sol
```diff
@@ -51,6 +51,7 @@ contract TychoRouterSequentialSwapTest is TychoRouterTestSetup {
             WETH_ADDR,
             USDC_ADDR,
             1000_000000, // min amount,
```
dianacarvalho1: all these comments are wrong now 😅

dianacarvalho1 (commented):
Thank you @louise-poole ! 🙏🏼 I have a few comments..

crates/tycho-execution/contracts/src/TychoRouter.sol
louise-poole: I had to call `_callMustInterceptOutput ` directly because calling `_determineFinalReceiver` and assigning the response to a var caused us to exceed the solidity variable stack limit. Should I make all methods just call `_callMustInterceptOutput` instead for consistency then? Not sure how to handle this...

dianacarvalho1: aaaff yes, I would prefer to be consistent everywhere! and honestly I think it's cleaner to call _callMustInterceptOutput directly now too

crates/tycho-execution/src/encoding/models.rs
louise-poole: This is quite a big refactor then no? Will touch on every encoder. I'll leave this as-is for now and we can do that refactor later as a clean-up (I want to unblock the audit step).

PR #1123, opened 2026-06-25
docs: add self-hosted EVM chain runbook

brunoguerios:
## What

Adds an operator-facing runbook for running `tycho-indexer` on an EVM chain that has **no hosted StreamingFast/Pinax Substreams endpoint**, using the self-hosted Firehose `substreams-endpoint` docker-compose profile.

- New page: `docs/for-solvers/self-hosted-evm-chain.md`
- Sidebar: `docs/SUMMARY.md` entry under **For Solvers**, after *Hosted Endpoints*

## Why

Operators indexing a chain with no hosted Substreams endpoint had no documented path. The self-hosted Firehose profile that makes this possible now exists in `docker-compose.yaml`; this page is the runbook for it.

## Contents

The runbook follows the ticket's structure: Overview (self-hosted vs. hosted), Prerequisites, Configuration (full env-var reference + writing `extractors.yaml` + custom-chain declaration + local SDK checkout), Running the stack (cold start, resume, hosted-endpoint alternative), Monitoring, Troubleshooting, and Performance tuning.

## Accuracy

Every env var, default, path, port, and command is grounded in `docker/docker-compose.yaml` (`substreams-endpoint` profile), the `.env` example, and `crates/tycho-indexer/extractors.example.yaml`. The page documents the now-shipped top-level `chains:` section for custom chains (resolved via `ChainConfigRegistry` in `main.rs`) — the earlier `CHAINS=<known-chain>` workaround has been removed, and `CHAINS` is described per its actual single-chain-context behavior rather than as an extractor filter.

## GitBook conventions

Active voice throughout, external links use `target="_blank" rel="noopener noreferrer"` anchors, tables set column widths, and `SUMMARY.md` is updated in the same change.

## Verification

- [ ] GitBook render check — new `SUMMARY.md` entry resolves and the page renders
- [ ] Cross-links to `hosted-endpoints.md` and `OBSERVABILITY.md` resolve
- [ ] Convention check — no passive voice, anchor-tag external links

### DIFF
--- a/docs/for-solvers/self-hosted-evm-chain.md
+++ b/docs/for-solvers/self-hosted-evm-chain.md
@@ -34,13 +34,13 @@ Configure the stack through `docker/.env`. The compose file reads it for both th
 <tr><td><code>START_BLOCK</code></td><td>poller</td><td>No</td><td><code>0</code></td><td>First block the poller fetches.</td></tr>
 <tr><td><code>CHAIN_NAME</code></td><td>poller</td><td>No</td><td><code>mainnet</code></td><td>Chain name the Firehose advertises (<code>--advertise-chain-name</code>).</td></tr>
 <tr><td><code>SUBSTREAMS_ENDPOINT</code></td><td>indexer</td><td>Yes</td><td><code>https://mainnet.eth.streamingfast.io:443</code></td><td>Substreams tier1 gRPC. Self-hosted: <code>http://substreams-endpoint:10016</code>.</td></tr>
-<tr><td><code>CHAINS</code></td><td>indexer</td><td>No</td><td><code>ethereum</code></td><td>Comma-separated chains to index; filters which extractors run. Custom-chain workaround until ENG-5738 (see below).</td></tr>
+<tr><td><code>CHAINS</code></td><td>indexer</td><td>No</td><td><code>ethereum</code></td><td>Active chain to index. The indexer uses only the first value (multichain is not yet supported). Name a built-in chain, or a custom chain you declare under <code>chains:</code> in the extractors config (see below).</td></tr>
 <tr><td><code>RETENTION_HORIZON</code></td><td>indexer</td><td>No</td><td><code>2000-01-01T00:00:00</code></td><td>Earliest block data the indexer retains.</td></tr>
 <tr><td><code>TYCHO_IMAGE</code></td><td>indexer</td><td>Yes</td><td>—</td><td>tycho-indexer image tag.</td></tr>
 <tr><td><code>EXTRACTORS_CONFIG</code></td><td>indexer</td><td>No</td><td><code>/opt/tycho-indexer/extractors.yaml</code></td><td>Path to the extractors config inside the container.</td></tr>
 <tr><td><code>TYCHO_PROTOCOL_SDK_PATH</code></td><td>indexer</td><td>No</td><td><code>../tycho-protocol-sdk</code></td><td>Host tycho-protocol-sdk checkout, mounted read-only.</td></tr>
 <tr><td><code>SUBSTREAMS_API_TOKEN</code></td><td>indexer</td><td>No</td><td><code>readme</code></td><td>Auth token for a hosted Substreams endpoint; unused self-hosted.</td></tr>
-<tr><td><code>TRACE_RPC_URL</code></td><td>indexer</td><td>For DCI</td><td>—</td><td>Trace-capable RPC for dynamic contract indexing.</td></tr>
+<tr><td><code>TRACE_RPC_URL</code></td><td>indexer</td><td>For DCI</td><td><code>readme</code> (placeholder)</td><td>Trace-capable RPC for dynamic contract indexing.</td></tr>
 <tr><td><code>OTLP_EXPORTER_ENDPOINT</code></td><td>indexer</td><td>No</td><td>empty (disabled)</td><td>OpenTelemetry collector. Set <code>http://lgtm:4317</code> with the <code>observability</code> profile.</td></tr>
 <tr><td><code>AUTH_API_KEY</code></td><td>indexer</td><td>No</td><td><code>local-dev-key</code></td><td>Tycho RPC API key.</td></tr>
 <tr><td><code>RUST_LOG</code></td><td>indexer</td><td>No</td><td><code>info</code></td><td>Log level.</td></tr>
@@ -59,8 +59,8 @@ START_BLOCK=24542000
 # Self-hosted Firehose, reachable by its docker service name
 SUBSTREAMS_ENDPOINT=http://substreams-endpoint:10016
 
-# Custom-chain workaround: reuse a known chain's extractors (see below)
-CHAINS=unichain
+# Custom chain — declared under `chains:` in extractors.yaml (see below)
+CHAINS=tempo
 
 RETENTION_HORIZON=2000-01-01T00:00:00
 SUBSTREAMS_API_TOKEN=local
@@ -71,7 +71,7 @@ OTLP_EXPORTER_ENDPOINT=
 
 ### Writing your extractors.yaml
 
-The indexer loads its extractors from `crates/tycho-indexer/extractors.yaml`, mounted into the container at `/opt/tycho-indexer/extractors.yaml`. Use that file as your template. Each entry under `extractors:` configures one protocol:
+Copy <a href="https://github.com/propeller-heads/tycho/blob/main/crates/tycho-indexer/extractors.example.yaml" target="_blank" rel="noopener noreferrer"><code>crates/tycho-indexer/extractors.example.yaml</code></a> to `crates/tycho-indexer/extractors.yaml` and edit it — the compose file mounts `extractors.yaml` into the container at `/opt/tycho-indexer/extractors.yaml`. The example file documents every field, including the optional `chains:` section. Each entry under `extractors:` configures one protocol:
 
 ```yaml
 extractors:
@@ -90,18 +90,45 @@ extractors:
 
 <table><thead><tr><th width="220">Field</th><th>Purpose</th></tr></thead><tbody>
 <tr><td><code>name</code></td><td>Unique extractor name; also the protocol system name exposed over the RPC.</td></tr>
-<tr><td><code>chain</code></td><td>Chain this extractor runs on. Must match an entry in <code>CHAINS</code> for the extractor to start.</td></tr>
+<tr><td><code>chain</code></td><td>Chain this extractor runs on — a built-in chain name, or a custom chain you declare under <code>chains:</code> below. The indexer rejects an unknown chain name at startup.</td></tr>
 <tr><td><code>implementation_type</code></td><td><code>Custom</code> for native substreams that emit Tycho protocol messages directly, or <code>Vm</code> for protocols simulated through the tycho-protocol-sdk.</td></tr>
 <tr><td><code>sync_batch_size</code></td><td>Number of blocks the indexer requests per Substreams batch.</td></tr>
 <tr><td><code>start_block</code></td><td>Block at which the protocol was deployed; the indexer starts streaming here.</td></tr>
 <tr><td><code>spkg</code></td><td>Path to the compiled <code>.spkg</code>, relative to <code>/opt/tycho-indexer/</code> (i.e. under <code>docker/substreams/</code>).</td></tr>
 <tr><td><code>module_name</code></td><td>Substreams output module to consume (e.g. <code>map_protocol_changes</code>).</td></tr>
-<tr><td><code>protocol_types</code></td><td>Protocol component types this extractor produces, each with a <code>name</code> and a <code>financial_type</code>.</td></tr>
+<tr><td><code>protocol_types</code></td><td>Protocol component types this extractor produces, each with a <code>name</code> and a <code>financial_type</code> (<code>Swap</code>, <code>Psm</code>, <code>Debt</code>, or <code>Leverage</code>).</td></tr>
 </tbody></table>
 
-{% hint style="info" %}
-A top-level `chains:` section that lets you declare a custom chain directly in `extractors.yaml` is pending ENG-5738. Until it ships, point `CHAINS` at a known chain whose extractors you want to reuse — for example `CHAINS=unichain` runs the Unichain extractors against the Tempo RPC.
-{% endhint %}
+### Declaring a custom chain
+
+Built-in chains (`ethereum`, `base`, `unichain`, …) need no extra config. To index a chain Tycho does not know, declare it in a top-level `chains:` section of the same file. Each extractor's `chain:` field resolves against these entries, and `CHAINS` names the active one. The indexer reads this section directly and fails fast at startup if an extractor references a chain that is neither built-in nor declared here.
+
+```yaml
+chains:
+  - name: tempo                       # the value used as `chain:` on extractors and in CHAINS
+    chain_id: 12345                   # EVM chain id
+    block_time_secs: 1
+    native:                           # native gas token
+      address: "0x0000000000000000000000000000000000000000"
+      symbol: "ETH"
+      decimals: 18
+    wrapped_native:                   # wrapped native token (e.g. WETH)
+      address: "0x0000000000000000000000000000000000000000"
+      symbol: "WETH"
+      decimals: 18
+    default_tvl_thresholds:           # TVL gates (in native token units) for component tracking
+      low: 1000
+      medium: 10000
+```
+
+<table><thead><tr><th width="220">Field</th><th>Purpose</th></tr></thead><tbody>
+<tr><td><code>name</code></td><td>Chain identifier; reference it from an extractor's <code>chain:</code> field and from <code>CHAINS</code>.</td></tr>
+<tr><td><code>chain_id</code></td><td>EVM chain id.</td></tr>
+<tr><td><code>block_time_secs</code></td><td>Average block time in seconds.</td></tr>
+<tr><td><code>native</code></td><td>Native gas token: <code>address</code>, <code>symbol</code>, <code>decimals</code>.</td></tr>
+<tr><td><code>wrapped_native</code></td><td>Wrapped native token (e.g. WETH): <code>address</code>, <code>symbol</code>, <code>decimals</code>.</td></tr>
+<tr><td><code>default_tvl_thresholds</code></td><td>TVL gates in native-token units (<code>low</code>, <code>medium</code>) for component tracking.</td></tr>
+</tbody></table>
 
 ### Pointing to a local tycho-protocol-sdk checkout
 
--- a/docs/for-solvers/self-hosted-evm-chain.md
+++ b/docs/for-solvers/self-hosted-evm-chain.md
@@ -19,8 +19,14 @@ The `substreams-endpoint` profile is a single-machine deployment: every Firehose
 ## Prerequisites
 
 - <a href="https://docs.docker.com/get-docker/" target="_blank" rel="noopener noreferrer">Docker</a> and Docker Compose.
+- The `tycho-indexer` image, built locally. From the repo root, build it and tag it to match `TYCHO_IMAGE`:
+
+  ```bash
+  docker build -f docker/tycho-indexer.Dockerfile -t tycho-indexer:local .
+  ```
+
 - An EVM JSON-RPC endpoint for your chain. The poller fetches every block from it, so prefer a low-latency node — RPC latency directly caps the fetch rate.
-- A compiled Substreams package (`.spkg`) for each protocol you want to index. Place it under `docker/substreams/`; the compose file mounts that directory into the indexer at `/opt/tycho-indexer/substreams/`.
+- A compiled Substreams package (`.spkg`) for **each protocol you want to index, built for your chain**. A chain Tycho has never indexed has no ready-made package, so you build one — see [Building a Substreams package for your chain](#building-a-substreams-package). Place each `.spkg` under `docker/substreams/`; the compose file mounts that directory into the indexer at `/opt/tycho-indexer/substreams/`.
 - Optionally, a local <a href="https://github.com/propeller-heads/tycho-protocol-sdk" target="_blank" rel="noopener noreferrer">tycho-protocol-sdk</a> checkout if you index VM-based protocols (see [Pointing to a local tycho-protocol-sdk checkout](#pointing-to-a-local-tycho-protocol-sdk-checkout)).
 
 ## Configuration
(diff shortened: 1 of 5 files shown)


docs/for-solvers/self-hosted-evm-chain.md
```diff
@@ -0,0 +1,272 @@
...
+- <a href="https://docs.docker.com/get-docker/" target="_blank" rel="noopener noreferrer">Docker</a> and Docker Compose.
+- The `tycho-indexer` image, built locally. From the repo root, build it and tag it to match `TYCHO_IMAGE`:
+
+  ```bash
+  docker build -f docker/tycho-indexer.Dockerfile -t tycho-indexer:local .
+  ```
+
+- An EVM JSON-RPC endpoint for your chain. The poller fetches every block from it, so prefer a low-latency node — RPC latency directly caps the fetch rate.
+- A compiled Substreams package (`.spkg`) for **each protocol you want to index, built for your chain**. A chain Tycho has never indexed has no ready-made package, so you build one — see [Building a Substreams package for your chain](#building-a-substreams-package). Place each `.spkg` under `docker/substreams/`; the compose file mounts that directory into the indexer at `/opt/tycho-indexer/substreams/`.
+- Optionally, a local <a href="https://github.com/propeller-heads/tycho-protocol-sdk" target="_blank" rel="noopener noreferrer">tycho-protocol-sdk</a> checkout if you index VM-based protocols (see [Pointing to a local tycho-protocol-sdk checkout](#pointing-to-a-local-tycho-protocol-sdk-checkout)).
```
zizou0x: Why? We now have everything in this monorepo

docs/for-solvers/self-hosted-evm-chain.md
```diff
@@ -0,0 +1,272 @@
...
+<tr><td><code>SUBSTREAMS_ENDPOINT</code></td><td>indexer</td><td>Yes</td><td><code>https://mainnet.eth.streamingfast.io:443</code></td><td>Substreams tier1 gRPC. Self-hosted: <code>http://substreams-endpoint:10016</code>.</td></tr>
+<tr><td><code>CHAINS</code></td><td>indexer</td><td>No</td><td><code>ethereum</code></td><td>Active chain to index. The indexer uses only the first value (multichain is not yet supported). Name a built-in chain, or a custom chain you declare in <code>chains.yaml</code> (see below).</td></tr>
+<tr><td><code>RETENTION_HORIZON</code></td><td>indexer</td><td>No</td><td><code>2000-01-01T00:00:00</code></td><td>Earliest version history the indexer keeps. <strong>Set this to a recent date when backfilling historical blocks</strong> — see <a href="#retention-horizon-when-backfilling-history">the warning below</a>. The default keeps all history and does not work for a from-scratch historical backfill.</td></tr>
+<tr><td><code>TYCHO_IMAGE</code></td><td>indexer</td><td>Yes</td><td>—</td><td>tycho-indexer image tag.</td></tr>
```
zizou0x: We should group all required at the top

docs/for-solvers/self-hosted-evm-chain.md
```diff
@@ -0,0 +1,272 @@
...
+# Self-hosted Firehose, reachable by its docker service name
+SUBSTREAMS_ENDPOINT=http://substreams-endpoint:10016
+
+# Custom chain — defined in chains.yaml (see below)
+CHAINS=tempo
+
+# Backfilling history: keep this recent, not the 2000-01-01 default (see the warning under "Running the stack")
+RETENTION_HORIZON=2026-07-01T00:00:00
+SUBSTREAMS_API_TOKEN=local
+AUTH_API_KEY=local-dev-key
```
zizou0x: This one isn't documented above

docs/for-solvers/self-hosted-evm-chain.md
```diff
@@ -0,0 +1,272 @@
...
+START_BLOCK=24542000
+
+# Self-hosted Firehose, reachable by its docker service name
+SUBSTREAMS_ENDPOINT=http://substreams-endpoint:10016
+
+# Custom chain — defined in chains.yaml (see below)
+CHAINS=tempo
+
+# Backfilling history: keep this recent, not the 2000-01-01 default (see the warning under "Running the stack")
+RETENTION_HORIZON=2026-07-01T00:00:00
```
zizou0x: I'd set it to a future date in the example, by default it's easier and better to not keep any historical state.

```suggestion
RETENTION_HORIZON=2100-01-01T00:00:00
```

docs/for-solvers/self-hosted-evm-chain.md
```diff
@@ -0,0 +1,272 @@
...
+The indexer builds **every** entry in this file at startup and fails if any referenced `.spkg` is missing. The shipped file lists Ethereum protocols whose packages are not in `docker/substreams/`. Replace them with only the extractors for the chain and protocols you are indexing, each pointing at a `.spkg` you have placed under `docker/substreams/`.
+{% endhint %}
+
+
+```yaml
+extractors:
+  uniswap_v3:
+    name: "uniswap_v3"
+    chain: "tempo"
+    implementation_type: "Custom"
```
zizou0x, line 89: Quick note, I wonder why this is needed, I have a feeling it's leftover from old implementation, maybe we can easily get rid of it now. 

My main concern is that people wouldn't know if it's custom or vm so it's a friction point.

No need to address it now but we should have a look into it

docs/for-solvers/self-hosted-evm-chain.md
```diff
@@ -0,0 +1,272 @@
...
+      - name: "uniswap_v3_pool"
+        financial_type: "Swap"
+    spkg: "substreams/ethereum-uniswap-v3/ethereum-uniswap-v3-logs-only-0.1.1.spkg"
+    module_name: "map_protocol_changes"
+```
+
+<table><thead><tr><th width="220">Field</th><th>Purpose</th></tr></thead><tbody>
+<tr><td><code>name</code></td><td>Unique extractor name; also the protocol system name exposed over the RPC.</td></tr>
+<tr><td><code>chain</code></td><td>Chain this extractor runs on — a built-in chain name, or a custom chain defined in <code>chains.yaml</code> (see below). The indexer rejects an unknown chain name at startup.</td></tr>
+<tr><td><code>implementation_type</code></td><td><code>Custom</code> for native substreams that emit Tycho protocol messages directly, or <code>Vm</code> for protocols simulated through the tycho-protocol-sdk.</td></tr>
```
zizou0x: I find the explanation not very accurate here

Native is for protocol for which we extract attribute and have custom pool logic implemented (natively implemented maths). Vm is for protocol for which we use a local vm inside tycho-simulation to reproduce the logic, we aren't indexing attribute but full contract instead.

docs/for-solvers/self-hosted-evm-chain.md
```diff
@@ -0,0 +1,272 @@
...
+<tr><td><code>sync_batch_size</code></td><td>Number of blocks the indexer requests per Substreams batch.</td></tr>
```
zizou0x: Not very accurate here too

It's not what we request Substreams but what batch before pushing to the DB. Substreams always send us messages as a continuous flow, this setting is how many inserts we wait for before we push the batch to the db, only during syncing (when it's live we have others mechanisms)

docs/for-solvers/self-hosted-evm-chain.md
```diff
@@ -0,0 +1,272 @@
...
+      medium: 10000
+```
+
+<table><thead><tr><th width="220">Field</th><th>Purpose</th></tr></thead><tbody>
+<tr><td><code>name</code></td><td>Chain identifier; reference it from an extractor's <code>chain:</code> field and from <code>CHAINS</code>.</td></tr>
+<tr><td><code>chain_id</code></td><td>EVM chain id.</td></tr>
+<tr><td><code>block_time_secs</code></td><td>Average block time in seconds.</td></tr>
+<tr><td><code>native</code></td><td>Native gas token: <code>address</code>, <code>symbol</code>, <code>decimals</code>.</td></tr>
+<tr><td><code>wrapped_native</code></td><td>Wrapped native token (e.g. WETH): <code>address</code>, <code>symbol</code>, <code>decimals</code>.</td></tr>
+<tr><td><code>default_tvl_thresholds</code></td><td>TVL gates in native-token units (<code>low</code>, <code>medium</code>) for component tracking.</td></tr>
```
zizou0x: Let's be slightly clearer here, where is this used, how to set it (we should advise with $ cost probably)

docs/for-solvers/self-hosted-evm-chain.md
```diff
@@ -0,0 +1,272 @@
...
+</tbody></table>
+
+### Building a Substreams package
+
+Tycho ships the Substreams module sources in <a href="https://github.com/propeller-heads/tycho/tree/main/protocols/substreams" target="_blank" rel="noopener noreferrer"><code>protocols/substreams/</code></a>, one directory per <code>{chain}-{protocol}</code>. The WASM logic is chain-agnostic — a package pins the chain-specific factory address and start block through its manifest. A chain Tycho has never indexed has no prebuilt `.spkg`, so you build one.
```
zizou0x: > one directory per <code>{chain}-{protocol}</code>.

Actually that's not completely true, we should address this but right now we have forks and other chains config together with the implementation. For example [ethereum-uniswap-v2](https://github.com/propeller-heads/tycho/tree/main/protocols/substreams/ethereum-uniswap-v2) contains pancakeswap and other chains config

docs/for-solvers/self-hosted-evm-chain.md
```diff
@@ -0,0 +1,272 @@
...
+
+To index a Uniswap-V2-style DEX on a new chain, reuse the `ethereum-uniswap-v2` module and add a manifest for your chain:
+
+1. Copy an existing manifest, e.g. `protocols/substreams/ethereum-uniswap-v2/ethereum-uniswap-v2.yaml`, to `<chain>-<dex>.yaml` in the same directory.
+2. Edit the copy: set the package `name`/`version`, set every module's `initialBlock` to the DEX factory's deployment block, and set the `params` line to `factory_address=<your factory>&protocol_type_name=<your pool type>`.
+3. Build the WASM and pack the package:
+
+   ```bash
+   cd protocols/substreams/ethereum-uniswap-v2
+   make build          # cargo build --target wasm32-unknown-unknown --release
```
zizou0x: I don't think all integrations have a makefile, we should rely on `cargo build --target wasm32-unknown-unknown --release` instead

docs/for-solvers/self-hosted-evm-chain.md
```diff
@@ -0,0 +1,272 @@
...
+   substreams pack <chain>-<dex>.yaml
+   ```
+
+4. Copy the resulting `.spkg` into `docker/substreams/`, then point your extractor's `spkg:` field at it.
+
+Pool detection is event-based — the module filters `PairCreated` by the factory address and tracks `Sync` events — so any standard Uniswap-V2 fork works with only the factory address and start block changed. Other protocols (Uniswap V3/V4, Balancer, Curve, …) have their own module directories under `protocols/substreams/`; follow the same pattern with the manifest for that protocol.
```
zizou0x: I think it's unnecessary context, wdyt?

docs/for-solvers/self-hosted-evm-chain.md
```diff
@@ -0,0 +1,272 @@
...
+
+{% hint style="info" %}
+`substreams pack` may warn that `network` is not set. This is harmless for the self-hosted stack — the Firehose advertises the chain through `CHAIN_NAME`, not the package.
+{% endhint %}
+
+### Pointing to a local tycho-protocol-sdk checkout
```
zizou0x: I didn't get this part, it seems like a really bad dependency, why do we need it?

docs/for-solvers/self-hosted-evm-chain.md
```diff
@@ -0,0 +1,272 @@
...
+<tr><td>Tycho RPC</td><td><code>4242</code></td><td><code>tycho-indexer</code></td></tr>
+</tbody></table>
+
+For dashboards, logs, and traces, enable the `observability` profile alongside `substreams-endpoint` and set `OTLP_EXPORTER_ENDPOINT=http://lgtm:4317`. See [Observability](../OBSERVABILITY.md).
+
+## Troubleshooting
+
+- **Blocks look years old.** If the poller's block `age` field reads years instead of seconds, `RPC_URL` points at the wrong chain or a stale node. Confirm the endpoint serves your chain and is fully synced.
+- **Firehose never becomes healthy.** A hub bootstrap deadlock leaves the stack waiting indefinitely. Stop the stack and bring it back up; on restart the poller resumes from stored one-block files instead of bootstrapping from scratch.
+- **Corrupt Substreams state cache.** If tier2 fails to produce segments after an unclean shutdown, clear the cached state and restart: remove `/data/substreams-states/` inside the `firehose-data` volume, then `up` again.
```
zizou0x: How does this happen? How do I detect it?

docs/for-solvers/self-hosted-evm-chain.md
```diff
@@ -0,0 +1,272 @@
...
+<tr><td>Firehose / Substreams tier1 gRPC</td><td><code>10016</code></td><td><code>substreams-endpoint</code></td></tr>
+<tr><td>Tycho RPC</td><td><code>4242</code></td><td><code>tycho-indexer</code></td></tr>
+</tbody></table>
+
+For dashboards, logs, and traces, enable the `observability` profile alongside `substreams-endpoint` and set `OTLP_EXPORTER_ENDPOINT=http://lgtm:4317`. See [Observability](../OBSERVABILITY.md).
+
+## Troubleshooting
+
+- **Blocks look years old.** If the poller's block `age` field reads years instead of seconds, `RPC_URL` points at the wrong chain or a stale node. Confirm the endpoint serves your chain and is fully synced.
+- **Firehose never becomes healthy.** A hub bootstrap deadlock leaves the stack waiting indefinitely. Stop the stack and bring it back up; on restart the poller resumes from stored one-block files instead of bootstrapping from scratch.
```
zizou0x: Would there be ways to address this in docker compose? Like detect when it's zombie and restart automatically?

docs/for-solvers/self-hosted-evm-chain.md
```diff
@@ -0,0 +1,272 @@
...
+<table><thead><tr><th width="320">Endpoint</th><th width="120">Port</th><th>Service</th></tr></thead><tbody>
+<tr><td>Firehose / Substreams tier1 gRPC</td><td><code>10016</code></td><td><code>substreams-endpoint</code></td></tr>
+<tr><td>Tycho RPC</td><td><code>4242</code></td><td><code>tycho-indexer</code></td></tr>
+</tbody></table>
+
+For dashboards, logs, and traces, enable the `observability` profile alongside `substreams-endpoint` and set `OTLP_EXPORTER_ENDPOINT=http://lgtm:4317`. See [Observability](../OBSERVABILITY.md).
+
+## Troubleshooting
+
+- **Blocks look years old.** If the poller's block `age` field reads years instead of seconds, `RPC_URL` points at the wrong chain or a stale node. Confirm the endpoint serves your chain and is fully synced.
```
zizou0x: > If the poller's block `age` field reads years instead of seconds

I didn't get this part?

docs/for-solvers/self-hosted-evm-chain.md
```diff
@@ -0,0 +1,272 @@
...
+- **Firehose never becomes healthy.** A hub bootstrap deadlock leaves the stack waiting indefinitely. Stop the stack and bring it back up; on restart the poller resumes from stored one-block files instead of bootstrapping from scratch.
+- **Corrupt Substreams state cache.** If tier2 fails to produce segments after an unclean shutdown, clear the cached state and restart: remove `/data/substreams-states/` inside the `firehose-data` volume, then `up` again.
+- **Cold start takes a long time.** A fresh chain must fetch and merge every block from `START_BLOCK` before your protocols' deployment blocks appear. This is expected — set `START_BLOCK` close to your earliest protocol `start_block` to avoid fetching irrelevant history.
```
zizou0x: START_BLOCK is a Substreams config that needs to be set to the earliest block for all protocols?

docs/for-solvers/self-hosted-evm-chain.md
brunoguerios: That was a mistake caused by a leftover from the original PR that added self-hosted chains. Thanks for flagging it. It's now removed. ✅

docs/for-solvers/self-hosted-evm-chain.md
brunoguerios: It's actually already in the table as a non-required env var. Do you think it's missing info on how to use it?

docs/for-solvers/self-hosted-evm-chain.md
brunoguerios: True - Claude confirmed it's stored as metadata, but the indexer never branches on it. It suggests a separate cleanup ticket to handle it. Maybe we should add it to the Maintenance epic? Wdyt?

docs/for-solvers/self-hosted-evm-chain.md
brunoguerios: Same as above. That was a mistake caused by a leftover from the original PR that added self-hosted chains. Thanks for flagging it. It's now removed. ✅

docs/for-solvers/self-hosted-evm-chain.md
brunoguerios: It basically logs errors and the indexer's committed block height stops advancing even though the poller keeps fetching. I'll rewrite to make it more clear.

docs/for-solvers/self-hosted-evm-chain.md
brunoguerios: Yeap, we can add a health check and restart logic, but I think it should be on a separate PR. Should we go ahead and do it or should we leave as a future improvement?

docs/for-solvers/self-hosted-evm-chain.md
brunoguerios: Yeah, it's the Firehose poller start block. It should be set to before the earliest extractor/protocol start block. I rewrote it a bit - let me know if there's something still not clear on the description.

docs/for-solvers/self-hosted-evm-chain.md
brunoguerios: True - I'll remove it.

docs/for-solvers/self-hosted-evm-chain.md
brunoguerios: done ✅

docs/for-solvers/self-hosted-evm-chain.md
brunoguerios: done ✅

docs/for-solvers/self-hosted-evm-chain.md
brunoguerios: fixed ✅

docs/for-solvers/self-hosted-evm-chain.md
brunoguerios: fixed ✅

docs/for-solvers/self-hosted-evm-chain.md
brunoguerios: done ✅

docs/for-solvers/self-hosted-evm-chain.md
brunoguerios: fixed ✅

docs/for-solvers/self-hosted-evm-chain.md
brunoguerios: done ✅

docs/for-solvers/self-hosted-evm-chain.md
brunoguerios: done ✅

docs/for-solvers/self-hosted-evm-chain.md
brunoguerios: The context here is that for a healthy stream `age` should read seconds, so if by any chance it reads years, then it's a sign that something is clearly off and should be investigated. I rewrote to make it more clear. Let me know what you think.

docs/for-solvers/self-hosted-evm-chain.md
zizou0x: Yes, sounds good

docs/for-solvers/self-hosted-evm-chain.md
zizou0x: I'm not sure, depends on how often it happens and how much effort it is to address it. But if it often go stale then it makes it unusable

docs/for-solvers/self-hosted-evm-chain.md
```diff
@@ -0,0 +1,260 @@
...
+  uniswap_v3:
+    name: "uniswap_v3"
+    chain: "tempo"
+    implementation_type: "Custom"
+    sync_batch_size: 1000
+    start_block: 12369621
+    protocol_types:
+      - name: "uniswap_v3_pool"
+        financial_type: "Swap"
+    spkg: "substreams/ethereum-uniswap-v3/ethereum-uniswap-v3-logs-only-0.1.1.spkg"
```
zizou0x: I find it confusing that the spkg is named ethereum for tempo, I think we should rename it. Is the start block correct for temp? I'd like this example to be as accurate as possible

docs/for-solvers/self-hosted-evm-chain.md
zizou0x: I wonder if it makes sense to have it here if it's set to the default value, wdyt?

docs/for-solvers/self-hosted-evm-chain.md
zizou0x: Have you seen this happening? Does it mean that you need to reprocess everything on the Substreams side?

docs/for-solvers/self-hosted-evm-chain.md
zizou0x: Have you seen this happening? I don't understand how wrong could `RPC_URL` changes how the `age` is given?

All EVM chains should have the same interface

docs/for-solvers/self-hosted-evm-chain.md
brunoguerios: Agreed - will rename and update the start block to the correct value.

docs/for-solvers/self-hosted-evm-chain.md
brunoguerios: Yeah - that was speculative indeed - I'll remove it over here.

docs/for-solvers/self-hosted-evm-chain.md
brunoguerios: I checked this and the complexity to add a health check + restart is low, but the problem is that it can't differentiate a deadlock from a slow-but-healthy bootstrap.
The test I made locally took more than 2 min to become healthy, but that may vary per chain, rpc provider, etc..
In other words, we'd have to tune a `start_period` wait time to infer it's a deadlock instead of a "normal" start up time.
Setting a `start_period` too low could flag a healthy startup as a deadlock and keep restarting it by mistake. That to me is a bit worse than what we currently have.
Based on that, I'd say we should leave this as a future improvement in case we get reports that it's happening frequently enough to be a problem.

docs/for-solvers/self-hosted-evm-chain.md
brunoguerios: No, I haven't. These were suggestions made by Claude on things that could go wrong. I agree we should stick with actual problems reported by users instead of attempting to anticipate things that could go wrong. I'll remove this one as well.

docs/for-solvers/self-hosted-evm-chain.md
brunoguerios: True. I'll remove the ones that are simply re-stating default values.

zizou0x (approved):
LGTM

PR #1124, opened 2026-06-25
feat: Add LunarBase to Maximodel

tamaralipows:
Ran with two hops sequential swaps, nothing suspicious.


dianacarvalho1 (approved):
Thank you @tamaralipows 🙏🏼

PR #1125, opened 2026-06-26
fix(integration-test): Fetch router fee from FeeCalculator contract

dianacarvalho1:
This is done only once at startup. There is no need for a complicated auto updating mechanism. The integration test is restarted often enough.

Before we had the router fee hardcoded to 10bps. Now that we reduced the fee we forgot to update this 🫠 so all the slippage results were skewed at 0.1%
image

after this fix they are all centered back at 0 ✨ 
image


Troshchk (approved):
Nice! ✨

PR #1126, opened 2026-06-26
feat: hybrid Curve implementation (vm:curve) with vendored MIT math

kayibal:
Moves `vm:curve` from the generic VM adapter to a hybrid implementation: quote math runs in pure Rust, while pool state is read from the locally indexed VM storage at decode and delta-transition time. Indexing (the `vm:curve` substreams) and the on-chain executor are unchanged.

## Commits

1. **Vendor MIT curve-math/curve-adapter as inline modules** — the Curve AMM math (`math`) and pool-construction adapter (`adapter`) are copied verbatim from the MIT-licensed `curve-math` / `curve-adapter` crates (tree of commit `fb3b25c`) into `crates/tycho-simulation/src/evm/protocol/curve/{math,adapter}`. This removes the external git dependency; the crate pulls only from crates.io. Adaptation is mechanical: module paths repointed, upstream feature gates made unconditional, provenance/lint header per module root. MIT attribution is preserved in `LICENSE-curve-math`.

2. **Re-derive post-MIT Curve features from Curve Vyper (clean-room)** — three features that exist only in post-MIT (BUSL-1.1) revisions of the upstream library are re-derived independently from Curve Finance's public deployed Vyper contracts and verified wei-exact against on-chain `get_dy`: the StableSwapSTETH variant, the StableSwapALend `get_D` `+1` divisor guard, and the TwoCryptoV1 `eth_variant` `newton_y` selection. No post-MIT upstream Rust was consulted. This commit touches only the `math`/`adapter` modules.

3. **Hybrid integration** — `CurveState: ProtocolSim` + decoder. Coin order/decimals/index come from the `coins` static attribute (on-chain order), ETH sentinel normalized to the zero address. Variant resolution from the deploying factory + `pool_type`, with a MATH-version check for TwoCrypto NG vs Stable and an on-chain probe fallback. Proxy/implementation and MATH contracts are loaded into the shared engine so getter delegatecalls resolve; rebasing pools are included. Registered for `"vm:curve"` in the examples and the integration test behind a `curve_filter`.

## Licensing

The vendored code is the last MIT-licensed revision of the upstream library; later revisions are BUSL-1.1 and are not included. The post-MIT features were re-derived clean-room from Curve's public Vyper. Provenance is documented in `LICENSE-curve-math` and the module headers.

## Verification

- 186 curve unit tests pass (165 inherited MIT inline tests + 21 wei-exact/clean-room tests). `cargo clippy --all-features -- -D warnings` clean.
- Integration run (282 pools, 10 blocks): `get_amount_out` 0 failures, 0 decode failures, average execution slippage 0.10%. The only pools above 1% slippage are a known pre-existing low-liquidity pool and a StableSwapNG rate-bearing-token pool (StableSwapNG math is unchanged from the prior implementation).


kayibal:
## Live soak test results

Ran the filtered build continuously for ~2 days against a live Tycho stream (`--protocols vm:curve --tvl-threshold 1 --max-simulations 100 --disable-rfq`), comparing each simulated `get_amount_out` against the on-chain swap executed via the TychoRouter at the same block.

**Window:** 14,122 blocks (25408758 → 25422888)

| Metric | Result |
|---|---|
| Executions simulated | 3,469,720 |
| `get_amount_out` failures | 0 |
| `get_limits` failures | 0 |
| Decode failures | 0 |
| Slippage events > 0.2% | 0 |
| Pools > 1% slippage | 0 |
| Execution reverts | 184,942 (5.3%) + 12,903 setup-failed — router/execution-side (executor unchanged), not quote errors |

No simulated quote exceeded 0.2% vs on-chain execution across all 3.47M executions, and the rate-bearing/rebasing filter held the entire run (no excluded pool was ever quoted).

For context, the pre-filter validation run found two rate-bearing pools (apyUSD/apxUSD StableSwapNG, ETH/ETHx) mispricing ±24–31% — their oracle/rebase rate is not in the indexed pool state. Those are the pools this PR's filter excludes; supporting them correctly is deferred to a follow-up (background RPC rate loader).

crates/tycho-simulation/src/evm/protocol/curve/vm.rs
```diff
@@ -0,0 +1,872 @@
...
+where
+    <D as DatabaseRef>::Error: Debug,
+    <D as EngineDatabaseInterface>::Error: Debug,
+{
+    ProbingResults {
+        has_gamma: call_opt::<_, ICurve::gammaCall, _>(engine, pool, ICurve::gammaCall {})
+            .is_some(),
+        n_coins,
+        has_math: read_math_address(engine, pool).is_some(),
+        math_version: None,
```
kayibal: **TwoCrypto NG-vs-Stable can't be resolved on the probe-fallback path** 🤖

`probe()` hardcodes `math_version: None`, so when variant resolution falls back to `detect_variant` (unknown factory / legacy pool), it can't distinguish `TwoCryptoNG` from `TwoCryptoStable` and defaults to NG — a v0.x-math 2-coin crypto pool reaching this path would be mispriced. The factory path covers this via `resolve_twocrypto`, but the probe path silently can't. Consider `math_version: read_math_address(engine, pool).and_then(|m| read_version(engine, &m))`.

crates/tycho-simulation/src/evm/protocol/curve/vm.rs
```diff
@@ -0,0 +1,872 @@
...
+    pool: &AlloyAddress,
+) where
+    <D as DatabaseRef>::Error: Debug,
+    <D as EngineDatabaseInterface>::Error: Debug,
+{
+    let Some(math) = read_math_address(engine, pool) else {
+        return;
+    };
+    if let Ok(code) = get_code_for_contract(&math.to_string(), None).await {
+        let _ = engine.state.init_account(
```
kayibal: **Silent failure degrades variant detection** 🤖

The `init_account` result is discarded (`let _ =`) and an RPC fetch failure is silent. If the pool's real `MATH()` code fails to load, TwoCrypto NG-vs-Stable detection silently falls back to a possibly-wrong default and misprices with no signal. `load_stateless_contracts` propagates its error; this should at least `warn!`/`debug!` on failure.

crates/tycho-simulation/src/evm/protocol/curve/state.rs
```diff
@@ -0,0 +1,229 @@
...
+                    self.pool_address
+                ))
+            })?;
+
+        let mut new_pool = self.pool.clone();
+        let (balance_in, balance_out) = {
+            let balances = new_pool.balances();
+            (balances[i], balances[j])
+        };
+        // Apply the swap to coin balances for multi-hop routing. Stored D / price_scale are kept
```
kayibal, line 159: **Multi-hop re-quote of the same CryptoSwap pool misprices** 🤖

`new_state` keeps `D`/`price_scale` fixed and only moves balances. Correct for StableSwap (D ~invariant across a swap), but a route hitting the same CryptoSwap pool twice misprices the second hop. Since `new_state` is the multi-hop contract for `ProtocolSim`, surface this as a limitation on `CurveState`'s public docstring (or recompute `D` in the clone) rather than only an inline comment.

kayibal (commented):
🤖 Automated code review (integration scope only; vendored `math`/`adapter` excluded).

crates/tycho-simulation/src/evm/protocol/curve/mod.rs
```diff
@@ -0,0 +1,17 @@
+//! Hybrid Curve implementation: pure-Rust quote math via the vendored `math` / `adapter` modules,
+//! with pool state read from the locally indexed VM storage at decode and update time.
+//!
+//! Indexing (the `vm:curve` substreams) and the on-chain executor are unchanged; only the consumer
+//! decoder differs — register [`CurveState`] for `"vm:curve"` instead of the generic VM adapter.
```
die-herdplatte: This seems to just be the description of this PR's changes

crates/tycho-simulation/src/evm/protocol/curve/state.rs
```diff
@@ -0,0 +1,229 @@
...
+        if balance_in.is_zero() || balance_out.is_zero() {
+            return Ok((BigUint::ZERO, BigUint::ZERO));
+        }
+        // Soft limit: cap the input at the pool's own balance of the sell token. Beyond this the
+        // solver math becomes unreliable and output approaches the available reserve.
+        let max_out_reserve = balance_out.saturating_sub(U256::from(1));
+        let max_out = self
+            .pool
+            .get_amount_out(i, j, balance_in)
+            .unwrap_or(max_out_reserve)
```
die-herdplatte: Return a `SimulationError` or 0 instead?

crates/tycho-simulation/src/evm/protocol/curve/vm.rs
die-herdplatte, line 1: The turbofishes on `call` and `call_opt` aren't required here, without them it's more readable imo

crates/tycho-simulation/src/evm/protocol/curve/variant.rs
die-herdplatte, line 1: I find the mappings here very hard to understand:

Image

crates/tycho-simulation/src/evm/protocol/curve/vm.rs
```diff
@@ -0,0 +1,872 @@
...
+{
+    let initial_a = call_opt::<_, ICurve::initial_ACall, _>(engine, pool, ICurve::initial_ACall {});
+    let future_a = call_opt::<_, ICurve::future_ACall, _>(engine, pool, ICurve::future_ACall {});
+    match (initial_a, future_a) {
+        (Some(ia), Some(fa)) if ia == fa => Ok(ia),
+        _ => {
+            Ok(call::<_, ICurve::ACall, _>(engine, pool, ICurve::ACall {})? *
+                U256::from(A_PRECISION))
+        }
+    }
```
die-herdplatte, line 256: [The docs recommend using `interpolate_a`](https://github.com/propeller-heads/tycho/blob/bd21ab3b4c1c2ef9b49c6b4edd619a6657b377e2/crates/tycho-simulation/src/evm/protocol/curve/adapter/build.rs#L141-L142) when `initial_a != future_a`.
You could then use `Pool#set_amp` before quoting to update the amplification (because a timestamp change doesn't trigger `delta_transition`).
It might also be overkill to do this and a recomputation in `delta_transition` could be sufficient.

If this doesn't get implemented, a comment here would be useful, detailing why the recommended approach wasn't taken.

die-herdplatte (commented):
Looks good!
Can the `CurveAdapter` be removed now?

die-herdplatte (approved):

zizou0x (approved):
Approving in behalf of the integration team

PR #1130, opened 2026-06-29
fix: Update required filters and refactor ekubo_v3 filter

louise-poole:
- The required filters list was outdated so still warned users to use the non-existant curve filter etc.
- All filters should live together in the central filters file - ekubo_v3's filter was moved to this file and the old fn was marked as deprecated.
- gitbook docs were updated with:
    - improved integration note on where to add filters
    - updated supported protocols table


dianacarvalho1 (approved):
beautiful ✨

docs/for-solvers/supported-protocols.md
```diff
@@ -33,9 +53,17 @@ fn register_exchanges(
                 .exchange::<UniswapV2State>("sushiswap_v2", tvl_filter.clone(), None)
                 .exchange::<PancakeswapV2State>("pancakeswap_v2", tvl_filter.clone(), None)
                 .exchange::<UniswapV3State>("pancakeswap_v3", tvl_filter.clone(), None)
-                .exchange::<EVMPoolState<PreCachedDB>>("vm:balancer_v2", tvl_filter.clone(), Some(balancer_v2_pool_filter))
+                .exchange::<EVMPoolState<PreCachedDB>>(
```
zizou0x: I don't think formatting here makes sense as it depends on our repo settings and this is gonna live in our client's repos. Also I feel it's cleaner and easier to read on the docs if we keep one like per protocol.

crates/tycho-simulation/src/evm/protocol/filters.rs
```diff
@@ -111,3 +114,15 @@ pub fn erc4626_filter(component: &ComponentWithState) -> bool {
...
+/// Filters out unsupported ekubo_v3 extensions.
+pub fn ekubo_v3_extension_filter(component: &ComponentWithState) -> bool {
+    component
+        .component
+        .static_attributes
+        .get("extension")
+        .is_some_and(|extension_bytes| {
+            Address::try_from(&extension_bytes[..])
+                .is_ok_and(|extension| extension_type(extension).is_some())
+        })
```
zizou0x: Shouldn't we reuse `ekubo_v3::filter_fn` here? Is there a big reason to want the function body to be in this file?

I'd say it's cleaner if protocol specific things stay in protocol's folder

zizou0x (commented):
Thanks! Two small comments!

crates/tycho-simulation/src/evm/protocol/filters.rs
louise-poole: Sure! Main reason was for consistency with the other protocol filters, but I am happy to change that!

zizou0x (approved):

PR #1131, opened 2026-06-29
feat: add Ramses V3 (Polygon) indexing, simulation, and execution

die-herdplatte:
## What

Adds native support for Ramses V3 (a Uniswap V3 fork) on Polygon across the three integration layers.

### Indexing (Substreams)
New `protocols/substreams/polygon-ramses-v3` package, forked from `ethereum-uniswap-v3-logs-only`. Indexes pool creation and the events that affect swap state: `Initialize`, `Mint`, `Collect`, `Burn`, `Swap`, `Flash`, and `CollectProtocol`. Ramses-specific differences handled:
- Pools are keyed by `tickSpacing` (stored as a static attribute).
- The fee is governance-mutable, tracked as a dynamic attribute (incl. `FeeAdjustment`).
- `Mint` carries an extra `index` field vs Uniswap V3.

The package references `tycho.evm.v1.Transaction` directly via prost `extern_path` rather than redefining it.

### Simulation
New `ramses_v3` `ProtocolSim` in `tycho-simulation`. The swap math is byte-identical to Uniswap V3, so it reuses the shared `utils::uniswap` concentrated-liquidity math. The decoder reads `fee`, `liquidity`, `sqrt_price_x96`, `tick` from state attributes and `tick_spacing` from static attributes; the mutable fee is updated on `delta_transition`. Gas estimates were calibrated against on-chain Polygon traces.

### Execution
`ramses_v3` reuses the existing Uniswap V3 executor via `SlipstreamsSwapEncoder` (registered in the swap encoder registry; executor addresses added for Polygon).

## Testing
- `integration_test.tycho.yaml` covers the USDC.e/USDC pool `0x7BCD99AE248E86EcaB7A5B23451D73825117F54D`, with the block range chosen so `Mint`, `Collect`, `Burn`, and `Swap` are each emitted at least once.
- Unit tests for the decoder, the mutable-fee transition, swap-math parity, and the slipstreams encoder all pass.

## Notes
Gauge/reward state is not indexed, as it does not affect swap state.


die-herdplatte:
I thought about calling the Substreams directory `evm-ramses-v3` instead of `polygon-ramses-v3` because it's a multi-chain deployment (also on Arbitrum and HyperEVM) and fixing one chain in the directory name seems inappropriate.
I didn't see any other protocol follow such a convention though.

The Uniswap V3 swap math could also be extracted in the future (to `tycho_simulation::evm::protocol::clmm`) and reused in the various forks.

The integration test depends on the fix in https://github.com/propeller-heads/tycho/pull/1127.

protocols/substreams/polygon-ramses-v3/src/modules/1_map_pool_created.rs
```diff
@@ -0,0 +1,125 @@
...
+    eh.handle_events();
+}
+
+// Track governance fee changes. The factory emits `FeeAdjustment(pool, newFee)` (carrying the pool
+// address) whenever `setFee` is called, so we update the pool's `fee` attribute accordingly.
+fn get_fee_adjustments(
+    block: &eth::Block,
+    new_pools: &mut Vec<TransactionEntityChanges>,
+    factory_address: Address,
+) {
```
zach030: Would it be more appropriate to move this part of the code into `maybe_pool_event`? It'd keep all the pool-related event handling in one place, and also ensure the event is only processed when the pool exists.

crates/tycho-execution/config/executor_addresses.json
```diff
@@ -61,6 +61,7 @@
     "uniswap_v2": "0xb9f140F4f8d674c453b3Ee00646189452a74E10c",
     "quickswap_v2": "0xb9f140F4f8d674c453b3Ee00646189452a74E10c",
     "uniswap_v3": "0x4C409a96DE96E06f061cE55B15Cd318D3Dd2239c",
+    "ramses_v3": "0x4C409a96DE96E06f061cE55B15Cd318D3Dd2239c",
```
zach030, line 67: Can we fully reuse the Uniswap V3 executor without any changes?

zach030 (commented):
Thanks for the PR! Just one small request, since most of the code is copied from `ethereum-uniswap-v3-logs-only`, so could you keep the copy commit separate? That way I can clearly see what you changed on top.

crates/tycho-execution/config/executor_addresses.json
die-herdplatte: Yes, see the comment [here](https://github.com/die-herdplatte/tycho/blob/0c051a356207c620f2a3ba5a346714f86dc9f865/crates/tycho-execution/src/encoding/evm/swap_encoder/swap_encoder_registry.rs#L170).
I expanded this comment now.

protocols/substreams/polygon-ramses-v3/src/modules/1_map_pool_created.rs
die-herdplatte: True, this was here because it's the only event (besides `PoolCreated`) that was emitted on the factory instead of the pools.
There is a similar event on the pool as well though, so that works just fine.

die-herdplatte:
@zach030 Thanks!
I inserted a copy commit such that you can more clearly see the changes.

protocols/substreams/polygon-ramses-v3/src/modules/3_map_events.rs
```diff
@@ -1,187 +1,90 @@
...
                 })
-                .collect::<Vec<_>>()
         })
-        .collect::<Vec<_>>();
+        .collect();
 
-    pool_events.sort_unstable_by_key(|e| e.log_ordinal);
-
-    Ok(Events { pool_events })
+    Events { pool_events }
```
zach030, line 33: I noticed that you remove the sort here: `pool_events.sort_unstable_by_key(|e| e.log_ordinal);`
Is there a specific reason for that?

die-herdplatte: Because `logs_with_calls` already returns the logs sorted by their ordinal

die-herdplatte:
## Live integration test results (Polygon)

Ran `tycho-integration-test` against a local indexer fully synced to the Polygon chain tip, with `ramses_v3` as the only protocol, over **1000 consecutive live blocks** (~35 min). For every sampled pool state and swap direction the test quotes via `RamsesV3State` (`get_limits` + `get_amount_out`), encodes the swap through the TychoRouter (Slipstreams encoder → Uniswap V3 executor), and executes it via `debug_traceCall`, comparing simulated vs executed output.

```
Blocks Processed: 1000

ramses_v3:
  Unique Pools: 11
  Simulations: 16149
  Operations:
    get_limits:     20040 success, 0 failure
    get_amount_out: 16934 success, 0 failure
  Simulation Results:
    Success: 16149 (100.0%)
    Failed: 0 (0.0%) — Reverted: 0, Setup Failures: 0
    Slippage: avg 0.0000%, min 0.0000%, max 0.0000% (16149 samples)
```

- **0 errors** over the whole run; the only warnings were 23 skipped updates (2.3%) where the RPC tip overtook the queued update — the harness's designed catch-up behavior, unrelated to the integration.
- **0.0000% slippage on all 16,149 executed simulations**: the simulated amount out matched on-chain execution exactly, every time.
- Two pools in the set are fully drained (no initialized ticks); they correctly report zero swap limits and are skipped rather than simulated.

die-herdplatte:
I also checked the prices given by the price printer and tested some pools (including the popular USDC/USDC.e pool) by varying the quote amounts.

zach030 (approved):

PR #1134, opened 2026-06-30
feat: update maximodel

Troshchk:
Original [task](https://propeller-heads.atlassian.net/browse/ENG-6119)

  - Replace `minAmountOut` with `expectedAmountOut` + `maxSlippageBps` in all swap entry points                                                         
  - Add positive slippage splitting: when enabled, surplus above expected output is split between router and client via configurable share
  - Replace `_callGetEffectiveRouterFeeOnOutput` with `_callMustInterceptOutput` to match Solidity refactor                                             
  - Bump fee BPS precision from 10_000 to 100_000_000 (uint16->uint32)

### DIFF
--- a/crates/tycho-execution/model/src/model/dispatcher.rs
+++ b/crates/tycho-execution/model/src/model/dispatcher.rs
@@ -170,19 +170,9 @@ pub fn _call_handle_callback_on_executor(
     Ok(())
 }
 
-/// <https://github.com/propeller-heads/tycho-execution/blob/9b0512c9580617224c7a0d7de781674a2cdc6b62/foundry/src/Dispatcher.sol#L337>
+/// Mirrors `Dispatcher._callMustInterceptOutput` in Solidity.
 ///
-/// <https://github.com/propeller-heads/tycho-execution/blob/9b0512c9580617224c7a0d7de781674a2cdc6b62/foundry/src/FeeCalculator.sol#L223>
-pub fn _call_get_effective_router_fee_on_output(
-    params: &Params,
-    _state: &mut State,
-) -> Result<i64, Error> {
-    Ok(if crate::config::ENABLE_NONZERO_FEE_BPS {
-        params.request(
-            ParamKey::String("router_fee_on_output_bps"),
-            [0, crate::model::fee_calculator::MAX_FEE_BPS],
-        )?
-    } else {
-        0
-    })
+/// Delegates to `FeeCalculator.must_intercept_output`.
+pub fn _call_must_intercept_output(params: &Params, client_fee_bps: i64) -> Result<bool, Error> {
+    crate::model::fee_calculator::must_intercept_output(params, client_fee_bps)
 }
--- a/crates/tycho-execution/model/src/model/dispatcher.rs
+++ b/crates/tycho-execution/model/src/model/dispatcher.rs
@@ -170,9 +170,3 @@ pub fn _call_handle_callback_on_executor(
     Ok(())
 }
 
-/// Mirrors `Dispatcher._callMustInterceptOutput` in Solidity.
-///
-/// Delegates to `FeeCalculator.must_intercept_output`.
-pub fn _call_must_intercept_output(params: &Params, client_fee_bps: i64) -> Result<bool, Error> {
-    crate::model::fee_calculator::must_intercept_output(params, client_fee_bps)
-}
--- a/crates/tycho-execution/model/src/model/dispatcher.rs
+++ b/crates/tycho-execution/model/src/model/dispatcher.rs
@@ -169,4 +169,3 @@ pub fn _call_handle_callback_on_executor(
 
     Ok(())
 }
-
--- a/crates/tycho-execution/model/src/model/fee_calculator.rs
+++ b/crates/tycho-execution/model/src/model/fee_calculator.rs
@@ -2,6 +2,7 @@
 use crate::{address::Address, error::Error, math::checked_subtract, params::Params};
 
 pub const MAX_BPS: i64 = 100_000_000;
+const MAX_BPS_SQUARED: i64 = 10_000_000_000_000_000;
 
 /// <https://github.com/propeller-heads/tycho-execution/blob/9b0512c9580617224c7a0d7de781674a2cdc6b62/foundry/lib/FeeStructs.sol#L9>
 pub struct FeeRecipient {
@@ -131,7 +132,7 @@ fn _calculate_fee(
 
         if fee_info.router_fee_on_client_fee_bps > 0 {
             router_fee_on_client_fee =
-                client_fee_numerator * fee_info.router_fee_on_client_fee_bps / (MAX_BPS * MAX_BPS);
+                client_fee_numerator * fee_info.router_fee_on_client_fee_bps / MAX_BPS_SQUARED;
         }
 
         client_portion = checked_subtract(total_client_fee, router_fee_on_client_fee)?;
--- a/crates/tycho-execution/model/src/model/fee_calculator.rs
+++ b/crates/tycho-execution/model/src/model/fee_calculator.rs
@@ -185,6 +185,8 @@ fn _merge_fee_recipients(
 
     // fees[0] = router fees, slippage[0] = router slippage
     // fees[1] = client fees, slippage[1] = client slippage
+    debug_assert!(fees.len() == 2, "expected exactly 2 fee recipients (router, client)");
+    debug_assert!(slippage.len() == 2, "expected exactly 2 slippage recipients (router, client)");
     fees[0].fee_amount += slippage[0].fee_amount;
     fees[1].fee_amount += slippage[1].fee_amount;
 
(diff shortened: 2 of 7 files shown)


crates/tycho-execution/model/src/model/fee_calculator.rs
```diff
@@ -1,7 +1,8 @@
 //! <https://github.com/propeller-heads/tycho-execution/blob/main/foundry/src/FeeCalculator.sol>
 use crate::{address::Address, error::Error, math::checked_subtract, params::Params};
 
-pub const MAX_FEE_BPS: i64 = 10000;
+pub const MAX_BPS: i64 = 100_000_000;
+const MAX_BPS_SQUARED: i64 = 10_000_000_000_000_000;
```
tamaralipows, line 5: This is only used in one place, couldn't we just square MAX_BPS where it's used

crates/tycho-execution/model/src/simulate.rs
```diff
@@ -36,9 +36,10 @@ pub fn simulate(params: &Params) -> Result<(State, Vault, impl Log + use<> + Ser
     let token_in = params.request("token_in", Address::VARIANTS_EXCEPT_ZERO)?;
     let token_out = params.request("token_out", Address::VARIANTS_EXCEPT_ZERO)?;
 
-    // assumption: all swap methods revert if `min_amount_out = 0`.
+    // assumption: all swap methods revert if `expected_amount_out = 0`.
     // no need to simulate it.
-    let min_amount_out = params.request("min_amount_out", [1, 10000])?;
+    let expected_amount_out = params.request("expected_amount_out", [1, 10000])?;
+    let max_slippage_bps = params.request("max_slippage_bps", [0, 500, 10000])?;
```
tamaralipows: I just realized now that the max slippage allows us to have 0 amount out. We should probably put a cap on this and revert if the max slippage is set to 10000, since that's what we did before with the 0 amount out

crates/tycho-execution/model/src/model/dispatcher.rs
```diff
@@ -170,19 +170,9 @@ pub fn _call_handle_callback_on_executor(
...
-    Ok(if crate::config::ENABLE_NONZERO_FEE_BPS {
-        params.request(
-            ParamKey::String("router_fee_on_output_bps"),
-            [0, crate::model::fee_calculator::MAX_FEE_BPS],
-        )?
-    } else {
-        0
-    })
+/// Delegates to `FeeCalculator.must_intercept_output`.
+pub fn _call_must_intercept_output(params: &Params, client_fee_bps: i64) -> Result<bool, Error> {
```
tamaralipows: This is now just a passthrough fn, I don't think we need it anymore

tamaralipows (commented):
Thanks @Troshchk ! Have you run this? What were the results?

Troshchk:
> Thanks @Troshchk ! Have you run this? What were the results?

Yes, I ran it and didn’t see any suspicious combinations. I didn’t run it for very long though, so I think a more extensive run should be part of this [task](https://propeller-heads.atlassian.net/browse/ENG-6113).

crates/tycho-execution/model/src/model/tycho_router.rs
```diff
@@ -300,25 +321,29 @@ fn _split_swap_checked(
...
         )?
     } else {
         0
     };
 
-    let final_receiver =
-        determine_final_receiver(receiver, client_fee_bps, router_fee_on_output_bps)?;
+    let must_intercept = must_intercept_output(params, client_fee_bps)?;
+
+    let final_receiver = determine_final_receiver(must_intercept, receiver);
```
louise-poole: In the latest router changes on my branch I removed this fn. Callsites now just call `must_intercept_output` directly. Can you update this to reflect the latest changes?

louise-poole (commented):
Just 1 thing I noticed quick:

crates/tycho-execution/model/src/model/fee_calculator.rs
```diff
@@ -12,58 +12,126 @@ pub struct FeeRecipient {
...
+/// positive slippage surplus and standard fees.
 pub fn calculate_fee(
     params: &Params,
-    amount_in: i64,
+    actual_amount_out: i64,
+    expected_amount_out: i64,
     client_fee_bps: i64,
-) -> Result<(i64, Vec<FeeRecipient>), Error> {
+) -> Result<Vec<FeeRecipient>, Error> {
+    let fee_info = _get_fee_info(params)?;
```
louise-poole, line 78: I see we call this both here and in `must_intercept_output`... I might be mistaken, but I think the model isn't deterministic during the same run so in that run the two calls to `_get_fee_info` could return different data and demonstrate behaviour we wouldn't actually see onchain? It's not clear to me if this is actually an issue here or not... maybe @tamaralipows can weigh in?

crates/tycho-execution/model/src/model/tycho_router.rs
```diff
@@ -522,6 +611,10 @@ fn _settle_output(
         }
     }
 
+    if amount_out < min_amount_out {
+        return Err(Error::revert("_settle_output: slippage exceeded"));
+    }
+
```
louise-poole, line 619: To better replicate the solidity code I think this check should be after finalize balances? (whether that is correct for the solidity code to do or not, I'm not sure).

crates/tycho-execution/model/src/model/fee_calculator.rs
```diff
@@ -72,16 +140,53 @@ pub fn calculate_fee(
...
+) -> Vec<FeeRecipient> {
+    if slippage.is_empty() {
+        return fees;
     }
 
-    amount_out = checked_subtract(amount_out, client_portion + total_router_fee)?;
+    // fees[0] = router fees, slippage[0] = router slippage
+    // fees[1] = client fees, slippage[1] = client slippage
+    fees[0].fee_amount += slippage[0].fee_amount;
+    fees[1].fee_amount += slippage[1].fee_amount;
```
louise-poole, line 192: Might be good to add a `debug_assert!(fees.len() == 2)` here to ensure this is the case no matter the scenario? Also would help us catch any future changes that might break this assumption we're operating under.

louise-poole (commented):
Just a couple things I spotted - not sure how important they are. If you're unsure too please defer to @tamaralipows for her opinion here?

crates/tycho-execution/model/src/model/fee_calculator.rs
Troshchk: I think the Params actually doesn't change during a run, so calling _get_fee_info twice always returns the same values

tamaralipows: Actually it looks like it will return the same value. This is some magic of params.request lol: 
```
  pub fn request(&self, key, variants) -> Result<V, ...> {
      let key = key.into();
      if let Some(param) = self.0.get(&key) {   // <-- lookup by the STRING key
          return Ok(param.try_into());          //     variants ignored entirely
      }
      // key not yet bound → error that forks the search into one Params per variant
      Err(RequestParamError::RequestParam(RequestParam::new(self, key, variants)))
  }

```
So if the key is not already there, only then will it request from the vec. But called twice in the same run it'll use the same key.

If this is problematic I think we will also see issues in the testing phase.

protocols/adapter-integration/evm/test/EtherfiAdapter.t.sol
```diff
@@ -77,14 +77,15 @@ contract EtherfiAdapterTest is Test, ISwapAdapterTypes {
...
+            /// work(balance is shares). Request extra eETH to cover
+            /// share-based rounding losses during transfer.
             deal(address(adapter), type(uint256).max);
             adapter.swap(
                 pair,
                 address(address(0)),
                 address(eEth_),
                 OrderSide.Buy,
-                requiredEeth
+                requiredEeth + 10
```
tamaralipows: Can you explain why this is necessary now but wasn't before?

crates/tycho-execution/model/src/model/fee_calculator.rs
tamaralipows: I see this was reflecting the solidity code actually :/ so either we should put it back here or update the solidity code to match. I will let you decide there sorry

protocols/adapter-integration/evm/test/EtherfiAdapter.t.sol
Troshchk: The previous code funded the adapter with limits[0] of eETH, so share based rounding losses during eETH transfers were always absorbed.
Due to change in this [commit](https://github.com/propeller-heads/tycho/commit/9e65aa4a1d58a0238e2a0d081c9230c4b5774416) adapter strated to be funded by `requiredEeth` and now the adapter sometimes ends up with a few wei less than needed after the transfer.
I have also refactored it, lmk it is it more clear that way

crates/tycho-execution/model/src/model/fee_calculator.rs
Troshchk: I have put it back

protocols/adapter-integration/evm/test/EtherfiAdapter.t.sol
tamaralipows: Oh this is just a flaky test unrelated to this PR then? i see

tamaralipows (approved):
Thank you!

PR #1136, opened 2026-06-30
fix: handle LunarBase zero quotes and marginal spot price

ThomasAqu1nas:
## Summary

- Return `0` for LunarBase `get_amount_out(0)` without transitioning state.
- Update LunarBase `spot_price()` to use marginal anchor price instead of reserve ratio.
- Apply directional base fees:
  - X -> Y: `anchor * (1 - fee_bid)`
  - Y -> X: `1 / anchor * (1 - fee_ask)`
- Add unit coverage for zero quotes and marginal spot price behavior.

## Testing

- `rustfmt +nightly --check crates/tycho-simulation/src/evm/protocol/lunarbase/state.rs`
- `cargo test --package tycho-simulation evm::protocol::lunarbase --lib`
- `cargo clippy --package tycho-simulation --lib --all-features -- -D warnings`
- `git diff --check`


zizou0x (approved):
LGTM

PR #1139, opened 2026-06-30
feat: add tokens to fee taking signature

Troshchk:
[Task](https://propeller-heads.atlassian.net/browse/ENG-6171)

This PR updates the fee calculation interfaces to accept `amountIn`, `tokenIn`, and `tokenOut`.

Implemented:
* Updated `IFeeCalculator.calculateFee` signature.
* Updated `FeeCalculator.sol` implementation.
* Updated `Dispatcher._callCalculateFee` to accept and forward the new parameters.
* Updated the Rust fee calculator model signature with unused stub parameters.
* Updated `FeeCalculator.t.sol` call sites.
* Updated `_takeFees` to accept and forward the new parameters

### DIFF
--- a/crates/tycho-execution/contracts/interfaces/IFeeCalculator.sol
+++ b/crates/tycho-execution/contracts/interfaces/IFeeCalculator.sol
@@ -30,6 +30,9 @@ interface IFeeCalculator {
      * @param actualAmountOut The actual amount received from the swap
      * @param expectedAmountOut Caller-supplied quoted amount out.
      *        Base for fee calculation and slippage surplus
+     * @param amountIn The input amount of the swap
+     * @param tokenIn The input token address
+     * @param tokenOut The output token address
      * @param clientFeeBps Client fee in fee units (100_000_000 = 100%)
      * @param client The client address to look up custom router fees
      *        and slippage share for, and to receive the client fee
@@ -40,6 +43,9 @@ interface IFeeCalculator {
     function calculateFee(
         uint256 actualAmountOut,
         uint256 expectedAmountOut,
+        uint256 amountIn,
+        address tokenIn,
+        address tokenOut,
         uint32 clientFeeBps,
         address client
     ) external view returns (FeeRecipient[] memory feeRecipients);
--- a/crates/tycho-execution/contracts/interfaces/IFeeCalculator.sol
+++ b/crates/tycho-execution/contracts/interfaces/IFeeCalculator.sol
@@ -28,11 +28,10 @@ interface IFeeCalculator {
      *      client address; client fee parameters are passed as
      *      function arguments.
      * @param actualAmountOut The actual amount received from the swap
-     * @param expectedAmountOut Caller-supplied quoted amount out.
-     *        Base for fee calculation and slippage surplus
+     * @param expectedAmountOut Caller-supplied quoted amount out
      * @param amountIn The input amount of the swap
-     * @param tokenIn The input token address
-     * @param tokenOut The output token address
+     * @param tokenIn The swap's input token address
+     * @param tokenOut The swap's output token address
      * @param clientFeeBps Client fee in fee units (100_000_000 = 100%)
      * @param client The client address to look up custom router fees
      *        and slippage share for, and to receive the client fee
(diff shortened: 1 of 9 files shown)


crates/tycho-execution/contracts/src/Dispatcher.sol
```diff
@@ -286,17 +286,21 @@ contract Dispatcher is TransferManager {
...
     }
 
-    function _callCalculateFee(
-        address feeCalculator,
-        uint256 actualAmountOut,
-        uint256 expectedAmountOut,
-        uint32 clientFeeBps,
-        address client
-    ) internal view returns (FeeRecipient[] memory feeRecipients) {
+    function _callCalculateFee(address feeCalculator, FeeInput memory f)
```
tamaralipows: Can we name this `feeInput` instead of `f`

crates/tycho-execution/contracts/src/TychoRouter.sol
```diff
@@ -1245,29 +1259,16 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
...
-    ) internal returns (uint256 amountOutAfterFees) {
-        FeeRecipient[] memory fees = _callCalculateFee(
-            _feeCalculator,
-            actualAmountOut,
-            expectedAmountOut,
-            clientFeeBps,
-            client
-        );
-        amountOutAfterFees = actualAmountOut;
+    function _takeFees(FeeInput memory f)
```
tamaralipows: same here, better naming?

crates/tycho-execution/contracts/test/FeeCalculator.t.sol
```diff
@@ -205,8 +214,9 @@ contract FeeCalculatorTest is Constants {
 
         uint256 amountIn = 1 ether;
 
-        FeeRecipient[] memory feeRecipients =
-            feeCalculator.calculateFee(amountIn, amountIn, 0, ALICE);
+        FeeRecipient[] memory feeRecipients = feeCalculator.calculateFee(
+            amountIn, amountIn, 0, address(0), address(0), 0, ALICE
```
tamaralipows: These tests are a bit confusing now with amountIn twice. Maybe we should call the `amountIn` `amountOutBeforeFees` and put another value for the "true" amountIn just for organization (I know it's not used right now in the test cases really). Thinking about future us

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -91,6 +91,9 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
      *        (used for slippage surplus calculation)
      * @param expectedAmountOut Caller-supplied quoted amount out.
      *        Fees are calculated on this amount.
```
tamaralipows: I know it's not a part of this PR but I realize this docstring is actually wrong... if there is no slippage then fees are calculated on the full amount. We can remove this `Fees are calculated on this amount.` I think. 

We can also remove `(used for slippage surplus calculation)` since that is kindof leaking implementation details into the docstring but not as strong opinions about that one

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -91,6 +91,9 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
...
+     * @param amountIn The input amount of the swap
+     * @param tokenIn The input token address
+     * @param tokenOut The output token address
```
tamaralipows: Smol thing but maybe specify that it's the _swap's_ input/output token address so it doesn't look like we are doing something funny with fees like taking it in a different token

crates/tycho-execution/contracts/interfaces/IFeeCalculator.sol
```diff
@@ -30,6 +30,9 @@ interface IFeeCalculator {
      * @param actualAmountOut The actual amount received from the swap
      * @param expectedAmountOut Caller-supplied quoted amount out.
      *        Base for fee calculation and slippage surplus
+     * @param amountIn The input amount of the swap
+     * @param tokenIn The input token address
+     * @param tokenOut The output token address
```
tamaralipows: Same here for the doc issues

tamaralipows (commented):
Thank you - mostly style things!

tamaralipows (approved):
Thank you!!

PR #1145, opened 2026-07-01
docs: Add new executor addresses across chains

tamaralipows:
Add missing executors that were deployed in the last weeks


Troshchk (approved):
Thank you @tamaralipows!

PR #1146, opened 2026-07-01
fix: retry on lagging RPC in integration test validation

tamaralipows:
On fast chains like Base, Tycho often delivers a block before the RPC node has caught up. The node then rejects calls for that block as "block not found". Our integration test was treating these lagging-node errors as real problems, which tanked the execution-success and validation-success rates across all protocols and set off the alerts (both dev and prod).

## What already existed

A `block-not-found` retry was added a while back for the swap-simulation path (originally for Arbitrum). It recognizes the lagging-node error, converts it into a retry, and backs off exponentially (1s doubling to a 5s cap, ~20 attempts). That machinery was working — for the error formats it knew about.

## Why it still failed on Base

The detection only matched the phrasings seen on other chains ("header not found", "block #N not found"). Base returns a different string ("block not found: 0x..."), which the check didn't recognize. So on Base the existing swap-sim retry silently never fired, and every lagging-node call fell through to a revert metric.

Separately, the state-validation path had no retry at all — it went straight to recording a validation failure.

## What this PR adds

- Broadens the lagging-node detection to also match Base's error format (this is what actually re-arms the existing swap-sim retry on Base).
- Adds a retry with the same backoff budget to the state-validation path, which previously had none.
- Skips (rather than counts as a failure) any validation call that still can't find the block after retries — that's node lag, not a state mismatch.

The metrics now reflect actual simulation and validation accuracy rather than transient node lag.

### DIFF
--- a/crates/tycho-integration-test/src/main.rs
+++ b/crates/tycho-integration-test/src/main.rs
@@ -46,6 +46,7 @@ use tycho_test::{
         models::{TychoExecutionInput, TychoExecutionResult},
         simulate_swap_transaction, tenderly,
     },
+    is_block_not_found,
     token_prices::{cap_amount_to_eth_value, load_token_prices},
     validation::{batch_validate_components, get_validator, Validator},
 };
@@ -900,6 +901,17 @@ async fn process_update(
                     }
                 }
                 Err(e) => {
+                    if is_block_not_found(&e.to_string()) {
+                        // The RPC node still lags behind Tycho after the batch-validation retries
+                        // exhausted: the block genuinely isn't available yet. This is infra
+                        // latency, not a state mismatch, so skip it rather than polluting the
+                        // validation-failure metric.
+                        warn!(
+                            component_id = %component_id,
+                            "Skipping validation: RPC block not yet available after retries"
+                        );
+                        continue;
+                    }
                     error!(
                         component_id = %component_id,
                         error = %e,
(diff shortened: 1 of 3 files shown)


crates/tycho-integration-test/src/main.rs
```diff
@@ -900,6 +901,17 @@ async fn process_update(
...
+                    if is_block_not_found(&e.to_string()) {
+                        // The RPC node still lags behind Tycho after the batch-validation retries
+                        // exhausted: the block genuinely isn't available yet. This is infra
+                        // latency, not a state mismatch, so skip it rather than polluting the
+                        // validation-failure metric.
+                        warn!(
+                            component_id = %component_id,
+                            "Skipping validation: RPC block not yet available after retries"
+                        );
+                        continue;
```
zizou0x, line 913: Are we skipping verification entirely or does this trigger a retry?

crates/tycho-test/src/validation.rs
```diff
@@ -189,6 +198,34 @@ pub async fn batch_validate_components(
...
+            Err(e) => crate::is_block_not_found(&e.to_string()),
+            Ok(_) => false,
+        });
+
+        if !block_lagging || attempt >= BLOCK_NOT_FOUND_MAX_RETRIES {
+            return results;
+        }
+
+        tokio::time::sleep(backoff).await;
+        backoff = (backoff * 2).min(Duration::from_secs(5));
```
zizou0x, line 222: Do we need the min here?

zizou0x (approved):

crates/tycho-integration-test/src/main.rs
tamaralipows: Haha was just about to comment about this. This is only hit when the retries are exhausted. In validation.rs we do retry with a backoff.

crates/tycho-test/src/validation.rs
tamaralipows: Yeah - trying to avoid exponential increase. I think it's fine to try 1s, 2s, 4s, 5s, 5s, and so on.

PR #1151, opened 2026-07-02
feat(tycho-client): send generic client metadata header

brunoguerios:
Adds a generic client-metadata capability to tycho-client, and forwards it through tycho-simulation's `ProtocolStreamBuilder`.

Consumers attach arbitrary key/value entries that are sent to the server in a dedicated `X-Tycho-Client-Metadata` header; tycho-client stays Fynd-agnostic. User-Agent is unchanged and an empty map sends no header, so the change is backward-compatible for all existing consumers.

- `TychoStreamBuilder::add_client_metadata` (and the `ProtocolStreamBuilder` passthrough) take any iterator of key/value pairs and insert each entry. `build()` serializes once before any network I/O; invalid or oversized metadata is dropped with a warning and no header is sent, so bad telemetry never blocks the stream.
- `HttpRPCClient` and the `WsDeltasClient` handshake send the header when set.
- A shared `client_metadata` module holds the header name, serializer and validation, so the RPC and WS paths cannot drift. Entries serialize in key order as `key=value;key=value`. Serialization enforces key/value character rules and size caps (<=16 entries, <=64B keys, <=128B values, <=1KiB header), each with a distinct error variant. The builder docstrings warn that values are self-reported and may reach server metrics/logs, so callers must not include secrets or PII.

Covers ENG-6190 (tycho-client) and ENG-6191 (tycho-simulation passthrough). Follow-up (separate PR): server-side capture + Prometheus (ENG-6180), which will also enforce the caps and ignore oversized input.

Testing: unit tests only — serializer validation/caps, RPC header presence and absence (mockito), and the WS handshake request builder. Ran nightly fmt + clippy and the stable unit suite for tycho-client and tycho-simulation.


crates/tycho-client/src/stream.rs
```diff
@@ -71,6 +72,7 @@ pub struct TychoStreamBuilder {
     compression: bool,
     partial_blocks: bool,
     max_messages: Option<usize>,
+    client_metadata: BTreeMap<String, String>,
```
zizou0x: Why BTreeMap instead of HashMap?

crates/tycho-client/src/client_metadata.rs
```diff
@@ -0,0 +1,205 @@
...
+}
+
+/// Serializes client metadata into the `X-Tycho-Client-Metadata` header value.
+///
+/// Entries are emitted in key order as `key=value; key=value`. Returns `Ok(None)` for an empty
+/// map, meaning no header should be sent (back-compatible default). Keys must be non-empty and
+/// match `[A-Za-z0-9_.-]`; values must be non-empty visible ASCII excluding `;` and `=`. These
+/// rules are stricter than `HeaderValue::from_str`, so any accepted output is always a valid
+/// header value and the RPC path can never fail on serialized input.
+pub fn serialize_client_metadata(
```
zizou0x: Do we need this to be pub?

crates/tycho-client/src/client_metadata.rs
```diff
@@ -0,0 +1,205 @@
...
+pub const MAX_ENTRIES: usize = 16;
+/// Maximum key length in bytes.
+pub const MAX_KEY_BYTES: usize = 64;
+/// Maximum value length in bytes.
+pub const MAX_VALUE_BYTES: usize = 128;
+/// Maximum serialized header length in bytes.
+pub const MAX_HEADER_BYTES: usize = 1024;
+
+#[derive(Debug, Error, PartialEq, Eq)]
+pub enum ClientMetadataError {
```
zizou0x: Same here, seems like this is more an internal error

crates/tycho-client/src/client_metadata.rs
```diff
@@ -0,0 +1,205 @@
...
+pub const CLIENT_METADATA_HEADER: &str = "x-tycho-client-metadata";
+
+/// Maximum number of entries a client may send.
+pub const MAX_ENTRIES: usize = 16;
+/// Maximum key length in bytes.
+pub const MAX_KEY_BYTES: usize = 64;
+/// Maximum value length in bytes.
+pub const MAX_VALUE_BYTES: usize = 128;
+/// Maximum serialized header length in bytes.
+pub const MAX_HEADER_BYTES: usize = 1024;
```
zizou0x, line 22: I'm fine to keep them here for checking on the client side but these should be enforced by the server too, it should ignore anything that is too big

crates/tycho-client/src/client_metadata.rs
```diff
@@ -0,0 +1,205 @@
...
+        }
+        if !is_valid_value(value) {
+            return Err(ClientMetadataError::InvalidValue(value.clone()));
+        }
+        if value.len() > MAX_VALUE_BYTES {
+            return Err(ClientMetadataError::ValueTooLong(value.clone()));
+        }
+        parts.push(format!("{key}={value}"));
+    }
+    let serialized = parts.join("; ");
```
zizou0x: Do we need the space here?

crates/tycho-client/src/stream.rs
```diff
@@ -189,6 +192,33 @@ impl TychoStreamBuilder {
...
 
+    /// Sets the client metadata sent to the server in the `X-Tycho-Client-Metadata` header.
+    ///
+    /// The map is opaque to tycho-client; consumers supply their own keys. Replaces any
+    /// previously set metadata. An empty map sends no header. Invalid keys/values and oversized
+    /// metadata are rejected at `build()` time.
+    ///
+    /// Values are self-reported and may surface in the server's metrics and logs. Do not include
+    /// secrets or personally identifiable information.
+    pub fn client_metadata(mut self, metadata: BTreeMap<String, String>) -> Self {
```
zizou0x: I think we can get both interface merged and more convenient to use if we do something like this:

```
pub fn add_client_metadata<I, K, V>(mut self, metadata: I) -> Self
where
    I: IntoIterator<Item = (K, V)>,
    K: Into<String>,
    V: Into<String>,
```

The it just inserts each entry into the `self.client_metadata`

crates/tycho-client/src/stream.rs
```diff
@@ -256,6 +286,10 @@ impl TychoStreamBuilder {
             ));
         }
 
+        // Serialize client metadata once, before any network I/O, so invalid input fails fast.
+        let metadata_header = serialize_client_metadata(&self.client_metadata)
+            .map_err(|e| StreamError::SetUpError(format!("Invalid client metadata: {e}")))?;
```
zizou0x: Erroring here seems overkill, I'd log a warn and continue without the header.

crates/tycho-simulation/src/evm/stream.rs
```diff
@@ -423,6 +423,33 @@ impl ProtocolStreamBuilder {
         self
     }
 
+    /// Sets the client metadata forwarded to the server in the `X-Tycho-Client-Metadata` header.
+    ///
+    /// See [`TychoStreamBuilder::client_metadata`]. Values are self-reported and may surface in the
+    /// server's metrics and logs — do not include secrets or personally identifiable information.
+    pub fn client_metadata(mut self, metadata: BTreeMap<String, String>) -> Self {
```
zizou0x: Same comment about the interface than on TychoStreamBuilder

crates/tycho-simulation/tests/client_metadata.rs
```diff
@@ -0,0 +1,27 @@
...
+
+use std::time::Duration;
+
+use tycho_simulation::{
+    evm::{protocol::uniswap_v2::state::UniswapV2State, stream::ProtocolStreamBuilder},
+    tycho_client::{feed::component_tracker::ComponentFilter, stream::StreamError},
+    tycho_common::models::Chain,
+};
+
+#[tokio::test]
```
zizou0x: Unit tests are at the bottom of the module in rust, I think this should be inside stream.rs

zizou0x (commented):
Mostly looks good, just a few comments!

crates/tycho-client/src/client_metadata.rs
brunoguerios: Yeah, it's already planned to be enforced by the server too.

crates/tycho-client/src/client_metadata.rs
brunoguerios: True. I'll make it `pub(crate)`.

crates/tycho-client/src/client_metadata.rs
brunoguerios: I don't think so. Will make it `pub(crate)`.

crates/tycho-client/src/client_metadata.rs
brunoguerios: No need. Will remove it.

crates/tycho-client/src/stream.rs
brunoguerios: It was Claude's suggestion to be honest. I challenged it and it says it doesn't have a strong reason to be one or the other. I'm happy to move to HashMap if you think it makes more sense.

crates/tycho-client/src/stream.rs
brunoguerios: I get the idea that telemetry failing might deserve a warn instead of an error, but if I'm the one setting it up, I'd rather see it failing fast, than having to figure out later because it's silently producing empty dashboards.
With that said, it's not a strong opinion. I'm fine refactoring if you still think it's best that way.
Let me know and I'm make the change over here 😉

crates/tycho-simulation/tests/client_metadata.rs
brunoguerios: Ok. I'll move them there.

crates/tycho-client/src/stream.rs
brunoguerios: Nice! I'll refactor them over here 👍

crates/tycho-client/src/stream.rs
brunoguerios: I refactored as HashMap as it seems to be the more common default in Rust.

crates/tycho-client/src/stream.rs
brunoguerios: Ok, gave this a bit more thought and I agree that observability failure should produce a warning instead of an error. Will refactor here.

zizou0x: Yes I really think it should just loudly warn but not error. It also gives us some space to change things without risking to break everyone's client

zizou0x (approved):
LGTM ✅

PR #1156, opened 2026-07-03
fix(substreams): pause and drain superseded BopAMM books

tvinagre:
BopAMM re-lists an asset under a new asset id instead of delisting the old book, so a token's superseded book kept its stale committed quote and its share of the maker's inventory. The old book stayed active — a dead market that reverts `StaleUpdate` — and held the token balance that belongs to the live book.

## Change
- When a new book is created for a token that already has one, pause the older book and move its asset balance onto the new book.
- The token index is now newest-wins (`set` instead of `set_if_not_exists`) so inventory routing and simulation resolve to the live book.
- Shared USDC stays duplicated across books (unchanged).

## Validation
Local re-sync (block 25,099,176 → chain tip) on a Docker indexer, crossing the re-list at block 25,286,302:
- superseded book `…0000`: `paused=true`, WETH drained to 0
- live book `…0002`: holds the maker's WETH = 6.098230, an exact match to on-chain `balanceOf(maker)` at the same block
- WBTC book returns no quote (maker holds 0 WBTC)
- live WETH price ≈ 1699 USDC/WETH vs Bebop's live 1700.65 / 1701.34 (~0.1%)

Unit tests, `wasm32` release build, clippy, and `cargo +nightly fmt` all pass.

### DIFF
--- a/protocols/substreams/ethereum-bopamm/src/common.rs
+++ b/protocols/substreams/ethereum-bopamm/src/common.rs
@@ -178,6 +178,37 @@ fn superseded_books_from(
         .collect()
 }
 
+/// Component id of the book that currently holds a token's inventory before a newer book replaces
+/// it: the highest asset id below `new_asset_id` that backs `asset_token`, or `None` if this is the
+/// token's first book.
+///
+/// Only this book carries a non-zero accumulated balance. Every book below it was already drained
+/// to zero when its own replacement was created, so draining the whole superseded set (rather than
+/// just this one) would push those already-emptied books negative.
+pub fn previous_live_book_for_token(
+    asset_token: &[u8],
+    new_asset_id: u64,
+    store: &StoreGetProto<ProtocolComponent>,
+) -> Option<String> {
+    previous_live_book_from(asset_token, new_asset_id, |i| store.get_last(format!("book:{i}")))
+}
+
+/// Testable core of [`previous_live_book_for_token`] over a probe of asset id -> component.
+fn previous_live_book_from(
+    asset_token: &[u8],
+    new_asset_id: u64,
+    probe: impl Fn(u64) -> Option<ProtocolComponent>,
+) -> Option<String> {
+    (0..new_asset_id).rev().find_map(|i| {
+        let component = probe(i)?;
+        component
+            .tokens
+            .iter()
+            .any(|t| t.as_slice() == asset_token)
+            .then_some(component.id)
+    })
+}
+
 /// Component ids of every book a higher-id book has superseded across the venue.
 ///
 /// A token is backed by exactly one live book — the highest asset id, since BopAMM re-lists an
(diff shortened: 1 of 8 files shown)


protocols/substreams/ethereum-bopamm/src/modules/map_relative_balances.rs
```diff
@@ -107,6 +113,32 @@ fn seed_new_books(
...
+                        ord: tx.index,
+                        tx: Some(tx.clone()),
+                        token: token.clone(),
+                        delta: balance
+                            .clone()
+                            .neg()
+                            .to_signed_bytes_be(),
+                        component_id: old_id.into_bytes(),
+                    });
+                }
```
zach030: Could this cause an older book's balance to go negative?
e.g. book0=WETH → book2=WETH (drains book0 to 0) → book5=WETH: now `superseded_books_for_token(WETH, 5)` returns `[book0, book2]`. book2 drains fine, but book0 is already 0 and would go negative.

protocols/substreams/ethereum-bopamm/src/modules/map_relative_balances.rs
die-herdplatte, line 103: Isn't there a seeding gap for USDC?
New and replaced components will start with a zero balance, not with the balance of the maker.

protocols/substreams/ethereum-bopamm/src/modules/map_relative_balances.rs
```diff
@@ -107,6 +113,32 @@ fn seed_new_books(
...
+                // is shared across every book by design).
+                //
+                // Invariant: the superseded book's accumulated asset balance equals
+                // `balanceOf(maker)` at this block — it was the token's live book and tracked the
+                // maker's holdings until now, so `-balanceOf(maker)` nets it to zero. The balance
+                // store is additive and cannot be read back here (module graph cycle), so an exact
+                // drain isn't possible. The invariant only breaks if the maker's balance moves in
+                // this very block (its own Transfer is skipped via `seeded_tokens`) or the maker is
+                // rotated in the same block; both leave a small residual on the now-paused,
+                // unrouted book. Neither has occurred for this venue.
```
die-herdplatte: Wouldn't it be much simpler to use `TransactionChangesBuilder::add_balance_change` and set the superseded books' balance to 0? Can be done in `pause_superseded_books`.
That would fix this and the [Zach's comment](https://github.com/propeller-heads/tycho/pull/1156/changes#r3542470435) (though the latter can also be hardened by only taking the book with the highest asset ID).

protocols/substreams/ethereum-bopamm/src/modules/map_relative_balances.rs
```diff
@@ -107,6 +113,32 @@ fn seed_new_books(
                         component_id: comp_id.into_bytes(),
                     });
                 }
```
die-herdplatte, line 115: If multiple books with the same asset are registered in one block, the last registered (live) book is being attributed a multiple of the maker's balance because each superseded book adds the maker's balance to the live book here.
Very unlikely to happen though.

protocols/substreams/ethereum-bopamm/src/common.rs
```diff
@@ -132,6 +134,94 @@ pub fn books_for_token(
...
+    probe: impl Fn(u64) -> Option<ProtocolComponent>,
+) -> HashSet<String> {
+    let mut books: Vec<(u64, String, Vec<u8>)> = Vec::new();
+    let mut i = 0u64;
+    while let Some(component) = probe(i) {
+        if let Some(token) = asset_token(&component, usdc) {
+            books.push((i, component.id, token));
+        }
+        i += 1;
+    }
```
die-herdplatte, line 237: This pattern is also used in `enumerate_books_from` and `tracked_tokens`.
I'd consider only spelling this out only once, especially because it's logic that may change in the future.
The new function would return an iterator over `ProtocolComponent`s.

protocols/substreams/ethereum-bopamm/src/common.rs
```diff
@@ -132,6 +134,94 @@ pub fn books_for_token(
...
+    books
+        .into_iter()
+        .filter(|(asset_id, _, token)| {
+            live_id
+                .get(token)
+                .is_some_and(|live| asset_id < live)
+        })
+        .map(|(_, id, _)| id)
+        .collect()
```
die-herdplatte, line 253: This can be simplified:
Iterate backwards over `books` (asset_id can be dropped because it's monotonically increasing w.r.t. the array index) and insert the asset token into a `HashSet`.
If it already exists, it is superseded.

protocols/substreams/ethereum-bopamm/src/modules/map_relative_balances.rs
tvinagre: Good catch — confirmed. The drain looped over the whole superseded set, so a second re-list (book0 → book2 → book5) re-drained book0, which was already zeroed when book2 replaced it, pushing it negative.

Fixed in 62f9ac662: drain only the *previous live book* (highest asset id below the new one that backs the token) via a new `previous_live_book_for_token` helper — that book is the sole holder of the token's inventory. Pausing still marks all older books (re-pausing an already-paused book is idempotent). Added a unit test covering the exact book0/book2/book5 scenario.

PR #1157, opened 2026-07-03
fix: replace transaction cleanup with lock-safe incremental procedure

tvinagre:
## What

Replaces the `clean_transaction_table()` pg_cron job and retires the rebuild-and-swap prune script.

**Migration `2026-07-03_incremental_transaction_cleanup`:**
- Drops `clean_transaction_table()` — it ran as a single 10–28h transaction whose `AccessShare` locks on `component_balance` starved pg_partman's midnight `AccessExclusive` partition maintenance, queueing the whole indexer behind it (repeated gridlocks on dev BSC; one partman deadlock abort).
- Adds `cleanup_orphaned_transactions()`, a procedure that:
  - deletes orphaned `transaction` rows in 50k-id batches, one `COMMIT` per batch — locks are held for seconds, never hours
  - sets a 2s `lock_timeout` per batch and yields (exits gracefully) on `lock_not_available` / `deadlock_detected` instead of queueing behind partman
  - resumes from a persistent cursor (`transaction_cleanup_progress`, single-row table) — interrupting a run loses at most one uncommitted batch
  - bounds each invocation with a time budget (default 45 min)
  - skips rows younger than `p_min_age` (default 35 days — partman's 1-month retention plus margin) via a binary-searched id horizon; the `NOT EXISTS` probes remain the correctness gate
- Re-points the cron entry (matched by command, so differing jobids across environments work) to 02:00 UTC, clear of partman's midnight window.
- `down.sql` restores the previous function and its 12:00 schedule byte-for-byte.

**`scripts/prune_transaction_table.md`:** runbook replacing the configmap-only rebuild-and-swap script. On PG 15, partitioned tables (`component_balance`, `protocol_state`, `contract_storage`) cannot take `NOT VALID` foreign keys, so any drop-and-re-add FK swap forces multi-hour validation inside the exclusive window. Physical disk reclaim now uses pg_repack (same-OID file swap; FKs untouched; `--no-kill-backend` so the indexer always wins lock races) or `VACUUM FULL` with the fleet scaled down. Includes a backfill section: freshly-synced orphans look young to the horizon for 35 days and are drained with manual low-`p_min_age` runs.

**Test:** `test_cleanup_orphaned_transactions_serial_db` seeds referenced (via `account.creation_tx` and `contract_code.modify_tx`) and orphaned transactions, `CALL`s the procedure, and asserts orphans are deleted while referenced rows survive.

## Deployment note

Applying this migration re-enables the (currently manually disabled) cleanup on dev, under the new procedure, at 02:00 UTC.

## Testing

- Full migration chain applied from scratch on the CI Postgres image (pg_cron + partman); `diesel migration redo` round-trips.
- tycho-storage suite CI-style against that DB: 103 parallel + 4 serial, all pass.
- An adversarial review verified empirically on a live DB: the 9 `NOT EXISTS` probes match `pg_constraint` exactly (a missing probe silently cascade-deletes protocol data — demonstrated); `lock_timeout` is re-applied after every in-loop `COMMIT` (an `ACCESS EXCLUSIVE` grabbed mid-run made batch 5 yield at exactly 2s); interrupt/resume drained the exact remainder; cursor wrap-around, self-heal, and clamp behave as written; pg_cron executes the procedure's internal commits; the test fails if a probe is removed (mutation-tested); the horizon boundary case (all rows younger than `p_min_age`) is guarded.
- `cargo +nightly fmt --check`, `clippy --all-targets --all-features`: clean.
- Not run: the full multi-crate workspace suite — the only Rust change is a `#[cfg(test)]` module in tycho-storage.

### DIFF
--- a/.github/workflows/cd-deploy-dev.yaml
+++ b/.github/workflows/cd-deploy-dev.yaml
@@ -31,6 +31,7 @@ on:
           - ethereum-tycho-indexer
           - ethereum-tycho-indexer-new-protocol
           - base-tycho-indexer
+          - bsc-tycho-indexer
           - unichain-tycho-indexer
       cargo_profile:
         description: >
--- a/.github/workflows/cd-deploy-dev.yaml
+++ b/.github/workflows/cd-deploy-dev.yaml
@@ -33,6 +33,7 @@ on:
           - base-tycho-indexer
           - bsc-tycho-indexer
           - unichain-tycho-indexer
+          - polygon-tycho-indexer
       cargo_profile:
         description: >
           Cargo build profile. "profiling" includes debug symbols for
--- a/crates/tycho-storage/migrations/2026-07-03_incremental_transaction_cleanup/up.sql
+++ b/crates/tycho-storage/migrations/2026-07-03_incremental_transaction_cleanup/up.sql
@@ -1,0 +1,179 @@
+-- Incremental transaction cleanup
+--
+-- Replaces clean_transaction_table() with cleanup_orphaned_transactions().
+--
+-- The old function ran as a single transaction that anti-joined the whole transaction table
+-- against every referencing table (O(all data), 10-28h on BSC volume). Its long-held
+-- AccessShare locks on component_balance starved pg_partman's midnight partition maintenance
+-- of the AccessExclusive lock it needs, which in turn queued the entire indexer behind
+-- partman and repeatedly gridlocked (and once deadlocked) the database.
+--
+-- The new procedure deletes orphaned transaction rows in small id-range batches and COMMITs
+-- after every batch, so no lock is held longer than a few seconds. A short lock_timeout makes
+-- it yield to partman instead of queueing behind it. A persistent cursor makes runs resumable
+-- after interruption, and a time budget bounds every invocation.
+--
+-- WARNING: the NOT EXISTS probes below are the correctness gate. On databases built purely
+-- from migrations the referencing FKs are ON DELETE CASCADE, so deleting a referenced
+-- transaction would silently cascade-delete protocol data; on environments where the legacy
+-- rebuild script re-created the FKs they are plain NO ACTION, so the delete would error the
+-- batch instead. If a new table ever adds a foreign key to "transaction", its column MUST be
+-- added to the probe list (and indexed).
+
+DROP FUNCTION IF EXISTS clean_transaction_table();
+
+-- Single-row cursor so each run resumes where the previous one stopped.
+CREATE TABLE IF NOT EXISTS transaction_cleanup_progress(
+    id boolean PRIMARY KEY DEFAULT TRUE CHECK (id),
+    last_processed_id bigint NOT NULL DEFAULT 0,
+    updated_at timestamptz NOT NULL DEFAULT CURRENT_TIMESTAMP
+);
+
+INSERT INTO transaction_cleanup_progress DEFAULT VALUES
+ON CONFLICT DO NOTHING;
+
+-- p_min_age: rows younger than this (by inserted_ts) are skipped. 35 days = partman's 1-month
+-- retention plus margin for month-length variance and late partition drops. NOTE: inserted_ts
+-- is DB insertion time, so during a backfill freshly-inserted-but-already-orphaned rows are
+-- skipped for 35 days; drain those with manual runs using a lower p_min_age (see
+-- scripts/prune_transaction_table.md). inserted_ts is used (rather than chain time) because
+-- the horizon binary search requires a timestamp monotone in id order, which chain time is not
+-- when a newly added extractor backfills alongside live ones.
+CREATE OR REPLACE PROCEDURE cleanup_orphaned_transactions(
+    p_batch_size bigint DEFAULT 50000,
+    p_time_budget interval DEFAULT '45 minutes',
+    p_min_age interval DEFAULT '35 days',
+    p_lock_timeout text DEFAULT '2s'
+)
+LANGUAGE plpgsql
+AS $$
+DECLARE
+    v_started timestamptz := clock_timestamp();
+    v_horizon_ts timestamptz := clock_timestamp() - p_min_age;
+    v_lo bigint;
+    v_hi bigint;
+    v_mid bigint;
+    v_mid_ts timestamptz;
+    v_horizon_id bigint;
+    v_cursor bigint;
+    v_batch_end bigint;
+    v_batch_deleted bigint := 0;
+    v_deleted bigint := 0;
+    v_lock_failed boolean := false;
+BEGIN
+    SELECT min(id), max(id) INTO v_lo, v_hi FROM "transaction";
+    IF v_lo IS NULL THEN
+        RAISE NOTICE 'cleanup_orphaned_transactions: transaction table is empty';
+        RETURN;
+    END IF;
+
+    -- Binary-search the largest id older than the horizon (~40 single-row PK lookups).
+    -- Rows younger than the partman retention window cannot be orphans yet, so probing them
+    -- would be wasted IO. The NOT EXISTS checks below are the correctness gate; this is only
+    -- an optimization.
+    WHILE v_lo < v_hi LOOP
+        v_mid := (v_lo + v_hi + 1) / 2;
+        SELECT inserted_ts INTO v_mid_ts
+        FROM "transaction"
+        WHERE id >= v_mid
+        ORDER BY id
+        LIMIT 1;
+        IF v_mid_ts IS NOT NULL AND v_mid_ts < v_horizon_ts THEN
+            v_lo := v_mid;
+        ELSE
+            v_hi := v_mid - 1;
+        END IF;
+    END LOOP;
+    v_horizon_id := v_lo;
+
+    -- The search only verifies ids it moved v_lo onto; the initial v_lo = min(id) is never
+    -- checked. If every row is younger than the horizon the loop converges there unverified,
+    -- so re-probe the boundary and bail out rather than sweep a too-young row.
+    SELECT inserted_ts INTO v_mid_ts
+    FROM "transaction"
+    WHERE id >= v_horizon_id
+    ORDER BY id
+    LIMIT 1;
+    IF v_mid_ts IS NULL OR v_mid_ts >= v_horizon_ts THEN
+        RAISE NOTICE 'cleanup_orphaned_transactions: no rows older than %, nothing to do',
+            p_min_age;
+        RETURN;
+    END IF;
+
+    SELECT last_processed_id INTO v_cursor FROM transaction_cleanup_progress;
+    IF v_cursor IS NULL THEN
+        -- Self-heal if the progress row is missing (e.g. wiped by test teardown).
+        INSERT INTO transaction_cleanup_progress DEFAULT VALUES
+        ON CONFLICT DO NOTHING;
+        v_cursor := 0;
+    END IF;
+    IF v_cursor >= v_horizon_id THEN
+        -- Previous sweep reached the horizon: wrap around. Orphans appear anywhere below the
+        -- horizon, not just in the newest band: versioned rows move to a dated partition when
+        -- their valid_to is set, and that partition is dropped a month later, orphaning
+        -- transactions of any age.
+        v_cursor := 0;
+    END IF;
+    -- Skip id space below the smallest existing row (earlier sweeps already emptied it).
+    SELECT min(id) - 1 INTO v_lo FROM "transaction";
+    v_cursor := greatest(v_cursor, v_lo);
+
+    WHILE v_cursor < v_horizon_id AND clock_timestamp() - v_started < p_time_budget LOOP
+        v_batch_end := least(v_cursor + p_batch_size, v_horizon_id);
+        BEGIN
+            -- Applies to the current batch transaction only. If partman or anything else
+            -- holds a conflicting lock, the batch aborts and the procedure yields instead of
+            -- queueing behind it (a queued AccessExclusive would freeze the indexer).
+            PERFORM set_config('lock_timeout', p_lock_timeout, true);
+
+            DELETE FROM "transaction" t
+            WHERE t.id > v_cursor
+              AND t.id <= v_batch_end
+              AND NOT EXISTS (SELECT 1 FROM contract_code cc WHERE cc.modify_tx = t.id)
+              AND NOT EXISTS (SELECT 1 FROM protocol_component pc WHERE pc.creation_tx = t.id)
+              AND NOT EXISTS (SELECT 1 FROM protocol_component pc WHERE pc.deletion_tx = t.id)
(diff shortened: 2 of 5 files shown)


crates/tycho-storage/scripts/prune_transaction_table.md
```diff
@@ -0,0 +1,151 @@
...
+
+```sql
+CALL cleanup_orphaned_transactions(
+    p_min_age     => interval '1 day',
+    p_time_budget => interval '6 hours'
+);
+```
+
+This is safe to run while the sync continues — every batch commits in seconds and yields via
+`lock_timeout` — but it competes for IO with the sync. Once drained, continue below to reclaim
```
zizou0x: Yeah I think this shouldn't run during syncing, else it will take forever to sync

crates/tycho-storage/scripts/prune_transaction_table.md
```diff
@@ -0,0 +1,151 @@
+# Reclaiming `transaction` table disk space
+
+This runbook replaces the old `prune_transaction_table.sql` rebuild-and-swap script (previously
```
zizou0x, line 3: So you wanna use the same process `cleanup_orphaned_transactions` for both cronjob and manual pruning?

I really hope we can use the same script, but I think realistically after a huge resync we will still need the other approach (move used transaction and drop the whole table).

crates/tycho-storage/migrations/2026-07-03_incremental_transaction_cleanup/up.sql
```diff
@@ -0,0 +1,179 @@
...
+    SELECT min(id), max(id) INTO v_lo, v_hi FROM "transaction";
+    IF v_lo IS NULL THEN
+        RAISE NOTICE 'cleanup_orphaned_transactions: transaction table is empty';
+        RETURN;
+    END IF;
+
+    -- Binary-search the largest id older than the horizon (~40 single-row PK lookups).
+    -- Rows younger than the partman retention window cannot be orphans yet, so probing them
+    -- would be wasted IO. The NOT EXISTS checks below are the correctness gate; this is only
+    -- an optimization.
```
zizou0x: I think there are wrong assumptions here

inserted_ts > retention window doesn't mean that the row isn't orphan. If we're syncing pancakeswap_v2 on bsc today we get millions of rows with inserted_ts > retention window and most of them would be orphans. 

Also the assumption that tx.id is sorted the same as timestamps is wrong, it depends on the syncing order.

crates/tycho-storage/scripts/prune_transaction_table.md
tvinagre: Yes, I'll update the md to make this clearer. 
The biggest advantage is that `cleanup_orphaned_transactions`, although might slow insertions, is non-blocking - so we could potentially run it during sync, and the penalty could be lower than running the script post-sync, as it would prevent the DB from accumulating 100M+ dangling transactions.

One of the reasons I've tried the current approach is bc the script took 25h to run, with dev scaled down, so no concurrency existed.

crates/tycho-storage/migrations/2026-07-03_incremental_transaction_cleanup/up.sql
tvinagre: > "inserted_ts > retention window doesn't mean the row isn't orphan — syncing pancakeswap_v2 today gives millions of young orphans."

You're partially right - the comment needs to be updated to make it clear that "Rows younger than the partman retention window cannot be orphans yet" - is only true for live indexing, not for backfills. However, the system already accounts for this: the horizon, in this case, is an optimization gate, not a correctness gate. Young orphans are deferred, never lost (they become sweepable once they cross p_min_age, and the NOT EXISTS probes are the only thing that decides deletion). The backfill case is explicitly documented ~30 lines up in the same file (the `p_min_age` docstring) and has a dedicated runbook section with the manual low-p_min_age drain.

tvinagre: > Also the assumption that tx.id is sorted the same as timestamps is wrong, it depends on the syncing order.

Not really, I think you're comparing different timestamps. 

The binary search orders by `inserted_ts` (DB wall-clock at insertion, DEFAULT CURRENT_TIMESTAMP), not by chain time. id is a bigserial assigned at insertion, so id order ≈ inserted_ts order by construction, regardless of syncing order. Syncing order breaks the correlation between id and block.ts (chain time) — which is precisely why the procedure deliberately does not use chain time

crates/tycho-storage/migrations/2026-07-03_incremental_transaction_cleanup/up.sql
```diff
@@ -0,0 +1,179 @@
...
+-- p_min_age: rows younger than this (by inserted_ts) are skipped. 35 days = partman's 1-month
+-- retention plus margin for month-length variance and late partition drops. NOTE: inserted_ts
+-- is DB insertion time, so during a backfill freshly-inserted-but-already-orphaned rows are
+-- skipped for 35 days; drain those with manual runs using a lower p_min_age (see
+-- scripts/prune_transaction_table.md). inserted_ts is used (rather than chain time) because
+-- the horizon binary search requires a timestamp monotone in id order, which chain time is not
+-- when a newly added extractor backfills alongside live ones.
+CREATE OR REPLACE PROCEDURE cleanup_orphaned_transactions(
+    p_batch_size bigint DEFAULT 50000,
+    p_time_budget interval DEFAULT '45 minutes',
```
zizou0x, line 47: I'm afraid 45 minutes won't be enough to delete transactions from the last day, are we planning on triggering this more often than once per day?

crates/tycho-storage/migrations/2026-07-03_incremental_transaction_cleanup/up.sql
```diff
@@ -0,0 +1,179 @@
...
+        EXCEPTION
+            WHEN lock_not_available OR deadlock_detected THEN
+                v_lock_failed := true;
+        END;
+
+        IF v_lock_failed THEN
+            RAISE NOTICE 'cleanup_orphaned_transactions: lock not available, yielding at id %',
+                v_cursor;
+            EXIT;
+        END IF;
```
zizou0x, line 210: Should we have a retry? I find this a bit extreme

crates/tycho-storage/migrations/2026-07-03_incremental_transaction_cleanup/up.sql
```diff
@@ -0,0 +1,179 @@
...
+              AND t.id <= v_batch_end
+              AND NOT EXISTS (SELECT 1 FROM contract_code cc WHERE cc.modify_tx = t.id)
+              AND NOT EXISTS (SELECT 1 FROM protocol_component pc WHERE pc.creation_tx = t.id)
+              AND NOT EXISTS (SELECT 1 FROM protocol_component pc WHERE pc.deletion_tx = t.id)
+              AND NOT EXISTS (SELECT 1 FROM account a WHERE a.creation_tx = t.id)
+              AND NOT EXISTS (SELECT 1 FROM account a WHERE a.deletion_tx = t.id)
+              AND NOT EXISTS (SELECT 1 FROM account_balance ab WHERE ab.modify_tx = t.id)
+              AND NOT EXISTS (SELECT 1 FROM component_balance cb WHERE cb.modify_tx = t.id)
+              AND NOT EXISTS (SELECT 1 FROM protocol_state ps WHERE ps.modify_tx = t.id)
+              AND NOT EXISTS (SELECT 1 FROM contract_storage cs WHERE cs.modify_tx = t.id);
```
zizou0x, line 190: Is this atomic? Just to be sure there is no race condition with this and a syncing protocol

crates/tycho-storage/migrations/2026-07-03_incremental_transaction_cleanup/up.sql
```diff
@@ -0,0 +1,179 @@
...
+
+    WHILE v_cursor < v_horizon_id AND clock_timestamp() - v_started < p_time_budget LOOP
+        v_batch_end := least(v_cursor + p_batch_size, v_horizon_id);
+        BEGIN
+            -- Applies to the current batch transaction only. If partman or anything else
+            -- holds a conflicting lock, the batch aborts and the procedure yields instead of
+            -- queueing behind it (a queued AccessExclusive would freeze the indexer).
+            PERFORM set_config('lock_timeout', p_lock_timeout, true);
+
+            DELETE FROM "transaction" t
```
zizou0x, line 179: Would be nice to have a safeguard, if anything is still referencing any transaction, just error. 

With this, we could be future proof if we ever add tables and new references.

crates/tycho-storage/migrations/2026-07-03_incremental_transaction_cleanup/up.sql
```diff
@@ -0,0 +1,179 @@
...
+-- retention plus margin for month-length variance and late partition drops. NOTE: inserted_ts
+-- is DB insertion time, so during a backfill freshly-inserted-but-already-orphaned rows are
+-- skipped for 35 days; drain those with manual runs using a lower p_min_age (see
+-- scripts/prune_transaction_table.md). inserted_ts is used (rather than chain time) because
+-- the horizon binary search requires a timestamp monotone in id order, which chain time is not
+-- when a newly added extractor backfills alongside live ones.
+CREATE OR REPLACE PROCEDURE cleanup_orphaned_transactions(
+    p_batch_size bigint DEFAULT 50000,
+    p_time_budget interval DEFAULT '45 minutes',
+    p_min_age interval DEFAULT '35 days',
```
zizou0x: Could we resolve this dynamically instead? Get it from the oldest of all partition + a security range.

crates/tycho-storage/migrations/2026-07-03_incremental_transaction_cleanup/up.sql
tvinagre: Fair point — the conservative estimate (0.5–2k rows/s in prod → 1.3–5.4M deletions per 45-min run) covers the ~2M/day steady-state orphan rate, but the margin was thin. Changed the schedule to twice daily (02:00/14:00 UTC) in 67f8cc782 — runs are lock-safe, so extra frequency is free capacity. Budget and schedule also stay tunable without a migration (`CALL` args / `cron.alter_job`), and we should sanity-check pacing on the first prod cycle via `cron.job_run_details`.

crates/tycho-storage/migrations/2026-07-03_incremental_transaction_cleanup/up.sql
tvinagre: Agreed — added 3 retries with 30s backoff before yielding (67f8cc782). Nothing is held during the backoff (the failed batch's subtransaction rolled back and the enclosing transaction is committed first), so a pending partman lock is never starved. Verified empirically: with an ACCESS EXCLUSIVE held on component_balance, the run made 3 attempts over ~67s, yielded with a NOTICE, and left the cursor on the contended batch for the next run.

crates/tycho-storage/migrations/2026-07-03_incremental_transaction_cleanup/up.sql
tvinagre: Yes — each batch is a single DELETE statement in its own transaction, so the NOT EXISTS evaluation and the row deletions share one snapshot, and FK row-locks (FOR KEY SHARE from RI checks) serialize any genuine collision. Structurally the race can't arise though: block flushes are atomic (a block's tx rows and their referencing rows commit together in DBCacheWriteExecutor), so no committed transaction is ever momentarily unreferenced — a syncing protocol only references transactions from the block it's currently flushing, never an old orphan candidate. The horizon adds a ≥31-day buffer on top for live indexing.

crates/tycho-storage/migrations/2026-07-03_incremental_transaction_cleanup/up.sql
tvinagre: Implemented in 67f8cc782 — the procedure now diffs the actual FK set referencing "transaction" (from pg_constraint, top-level constraints only so partition clones don't false-positive) against its probe list at startup, and raises an exception naming both sets on mismatch. An unprobed future reference now fails loudly instead of silently cascade-deleting. Verified with a canary table: unprobed FK → run refused; canary dropped → runs again.

crates/tycho-storage/migrations/2026-07-03_incremental_transaction_cleanup/up.sql
tvinagre: Done in 67f8cc782 — p_min_age now defaults to NULL and resolves to max(partman.part_config.retention) + a 4-day security margin (falls back to 31+4 days if partman has no retention configured), so retention changes propagate automatically. Explicit values still override it, which the backfill drain in the runbook relies on. I used the configured retention rather than the oldest physical partition: if partman falls behind on drops, the extra partitions only mean those transactions are still referenced — which the NOT EXISTS probes handle anyway.

tvinagre:
## First production runs on BSC dev (branch image `e2ca00ee9`, migration applied at pod startup in ~170ms)

**Run 1 — first scheduled run (21:36 UTC, defaults):**

| metric | result |
|---|---|
| duration | 45:00 exactly — time budget stopped it, cursor persisted |
| lock waiters | **0 on all 45 one-minute samples**, indexer syncing alongside |
| coverage | cursor 0 → 3.29B of 3.57B max id (~92% of the sparse id space in one run) |
| deletions | 19,814 orphans |
| wait profile | alternating `IO/DataFileRead` / `cpu`; zero `Lock` waits |

**Run 2 — next scheduled run (02:00 UTC):** succeeded in 46:09, ~1.35M orphans deleted (resumed from the saved cursor, wrapped, and started absorbing the week-old backlog as it aged past the horizon).

The low run-1 deletion count is the horizon working as documented: the Jul 1 prune had already removed all then-existing orphans, and the orphans created since reference June-inserted transactions — younger than the 35-day horizon, so deferred, not lost. Run 2 already picked up 1.35M as they aged in.

Re the 45-minute budget concern: at this hardware's observed pace one run covers ~90% of the entire 3.6B id space, and the twice-daily schedule gives far more deletion capacity than the ~2M/day steady-state orphan rate. A one-off manual drain (`p_min_age => '7 days'`, 2h budget) is currently running server-side to clear the accumulated June backlog faster; steady state needs no manual intervention.

Compare with what this replaces: the previous cleanup ran 10–28h holding locks throughout and repeatedly gridlocked partman and the indexer; these runs are bounded, resumable, and blocked nothing.

.github/workflows/cd-deploy-dev.yaml
```diff
@@ -31,6 +31,7 @@ on:
           - ethereum-tycho-indexer
           - ethereum-tycho-indexer-new-protocol
           - base-tycho-indexer
+          - bsc-tycho-indexer
```
zizou0x, line 34: Please add polygon too 👀

zizou0x (approved):

zizou0x:
🚀

PR #1159, opened 2026-07-06
fix(simulation): Titan override timestamps, venue aliases and indexed-state fallback

zizou0x:
## Summary

Hardens the Titan pAMM live-override pipeline (follow-up to #1106) based on live-stream and on-chain investigation:

- **Beacon-slot block timestamp** (`9b6fe290`): the override block timestamp is derived from the frame's beacon `slot` (`genesis + slot * 12`), matching Titan's reference taker ([gattaca-com/propamm](https://github.com/gattaca-com/propamm)) and the [LambdaClass propAMM SDK](https://github.com/lambdaclass/propamm-router-contracts), with plausibility validation against the frame's wall clock. The registry-lane extraction stays as a guarded fallback. Required because the [priority-update-registry](https://github.com/flashbots/priority-update-registry) enforces `updateTimestamp == block.timestamp` (verified on-chain: plain eth_call reverts `StaleUpdate()`, bumped timestamp succeeds).
- **Venue aliases** (`4e128a8b`): each pAMM subscribes to all known Titan venue alias addresses (oracle and router keys carry identical payloads, and deployments rotate). bopAMM's venue is the self-storage pAMM `0x28d9ccedf1`. `scripts/titan-venue-census.py` inspects the live venue set.
- **Indexed-state fallback** (`94f72385`): provider-set `FailurePolicy` on `OverrideSnapshot`; Titan snapshots opt into retrying failed simulations without overrides. Fixes pools being unquotable (and undecodable) whenever a pair's oracle lane is not re-stamped for the pending block — briefly true for every pair at block transitions (lanes re-stamp gradually over ~2s), and for minutes at a time during maker quoting droughts.
- **`titan_override_monitor` example** (`d46c3a26`): manual proof that quotes reprice mid-block, with per-block distinct-quote recaps and optional VM traces (`--vm-traces`).

### DIFF
--- a/crates/tycho-simulation/src/evm/override_stream/titan.rs
+++ b/crates/tycho-simulation/src/evm/override_stream/titan.rs
@@ -6,8 +6,45 @@
 //! `EVMPoolState` instances before simulation. See
 //! <https://docs.titanbuilder.xyz/propamms/takers>.
 //!
-//! All Titan/Fermi-specific knowledge (venue addresses, lane-timestamp resolution, endpoint URL)
-//! lives only in this module; the rest of the override machinery is protocol-agnostic.
+//! # Why the block timestamp must be overridden too
+//!
+//! Titan quotes target the *pending* L1 block. Registry-backed venues (FermiSwap, Kipseli)
+//! receive their oracle prices through Flashbots'
+//! [priority-update-registry](https://github.com/flashbots/priority-update-registry), whose lanes
+//! are stamped with the pending block's timestamp and whose reads revert with `StaleUpdate()` when
+//! the stored timestamp falls outside a freshness window derived from `block.timestamp`.
+//! Simulating against the latest indexed block therefore reverts even with the storage overrides
+//! applied; the simulated block timestamp must be bumped to the pending block's.
+//!
+//! The pending block timestamp is derived from the frame's beacon `slot` field as
+//! `BEACON_GENESIS_TIMESTAMP + slot * SECONDS_PER_SLOT`. This mirrors Titan's reference taker
+//! implementation ([gattaca-com/propamm](https://github.com/gattaca-com/propamm), `quoter.py`) and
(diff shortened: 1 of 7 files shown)


crates/tycho-simulation/src/evm/override_stream/titan.rs
```diff
@@ -72,12 +113,17 @@ const KIPSELI_PROTOCOL_SYSTEM: &str = "vm:kipseli";
...
-/// FermiSwap pAMM venue address on Titan's quote stream.
-const FERMISWAP_VENUE: &str = "0xb1076fe3ab5e28005c7c323bac5ac06a680d452e";
-/// Kipseli pAMM venue address on Titan's quote stream.
-const KIPSELI_VENUE: &str = "0x5cdbe59400cc2efdcc2b54acca4a99fe00dd588c";
-/// bopAMM pAMM venue address on Titan's quote stream.
-const BOPAMM_VENUE: &str = "0x160141a205f5ddcf096ba3f48b7ed21eb52c62ea";
+/// FermiSwap venue aliases on Titan's quote stream: the FermiSwap oracle and the FermiSwapper
+/// router. Both keys carry the same registry diffs; either serves the protocol.
+const FERMISWAP_VENUES: &[&str] =
+    &["0xb1076fe3ab5e28005c7c323bac5ac06a680d452e", "0x5979458912f80b96d30d4220af8e2e4925a33320"];
```
tvinagre, line 119: Should we make this overridable via a configmap, envvar or move it to Fynd Auto? I do expect these to change regularly

tvinagre (approved):

crates/tycho-simulation/src/evm/override_stream/titan.rs
zizou0x: Yes I think we should host a config that is pulled on startup. But better do it separately. So far it hasn't changed in the past months

PR #1167, opened 2026-07-07
feat: extend Ekubo V3 signed swap executor and encoder

Troshchk:
[Task](https://propeller-heads.atlassian.net/browse/ENG-6136)

Adds SignedExclusiveSwap (forward-only) support to the existing ekubo_v3 executor and encoder.

  - EkuboV3SwapEncoder: appends self-describing signed tail (meta | minBalanceUpdate | sigLen | signature) when user_data is present; byte-identical to normal hops otherwise
  - EkuboV3Executor: detects signed hops by poolConfig.extension(), routes through CORE.forward() instead of CORE.swap()
  - SIGNED_EXCLUSIVE_SWAP_ADDRESS is a placeholder pending ekubov3 extension pool deployment
  
  [ekubo signed exclusive swap extension](https://github.com/EkuboProtocol/evm-contracts/pull/319)


crates/tycho-execution/contracts/src/executors/EkuboV3Executor.sol
```diff
@@ -265,25 +256,64 @@ contract EkuboV3Executor is IExecutor, ICallback {
...
+                            ),
+                            // signature
+                            bytes(swapData[sigLenOff + _SIG_LEN_BYTES:sigEnd])
+                        )
+                    ),
+                    (PoolBalanceUpdate, PoolState)
+                );
+
+                offset = sigEnd;
+            } else if (poolConfig.extension() == MEV_CAPTURE_ADDRESS) {
```
tamaralipows, line 297: Is there any way to add a test for this? Like importing the Ekubo code locally and deploying it to the test fork? Would that be too difficult?

crates/tycho-execution/contracts/foundry.toml
```diff
@@ -2,7 +2,7 @@
 src = 'src'
 out = 'out'
 libs = ['lib']
-evm_version = 'cancun'
+evm_version = 'osaka'
```
Troshchk, line 5: The Ekubo V3 SDK uses the clz opcode (count leading zeros) in its tick bitmap code. clz is only available in the Osaka EVM — it doesn't exist in cancun.

Previously this didn't matter because our tests only imported Ekubo types and interfaces, which don't touch the bitmap code. The new test imports SignedExclusiveSwap.sol, which transitively pulls in bitmap.sol containing clz. Without evm_version = 'osaka', the compiler rejects it.

crates/tycho-execution/contracts/src/executors/EkuboV3Executor.sol
Troshchk: Added it, lmk what you think

crates/tycho-execution/contracts/src/executors/EkuboV3Executor.sol
```diff
@@ -43,6 +43,13 @@ address payable constant CORE_ADDRESS =
 ICore constant CORE = ICore(CORE_ADDRESS);
 address constant MEV_CAPTURE_ADDRESS =
     0x5555fF9Ff2757500BF4EE020DcfD0210CFfa41Be;
+// PLACEHOLDER non-zero address — must NOT be address(0). Signed Ekubo V3
+// (SignedExclusiveSwap, forward-only) pools set their pool config extension to
+// this address; the executor detects a signed hop by comparing each hop's
+// poolConfig.extension() against it and routes that hop through the signed path.
+// TODO: replace with the deployed SignedExclusiveSwap extension address.
```
louise-poole, line 50: Can we add a note to the deployment task to come back and do this? (if it is not noted there already)

crates/tycho-execution/contracts/src/executors/EkuboV3Executor.sol
```diff
@@ -147,7 +185,8 @@ contract EkuboV3Executor is IExecutor, ICallback {
     }
 
     function getCallbackTransferData(
-        bytes calldata, /* data */
+        bytes calldata,
+        /* data */
```
louise-poole: Was this format change intentional? I preferred what was here before (matches the `address` line too).

crates/tycho-execution/contracts/test/protocols/EkuboV3.t.sol
```diff
@@ -195,6 +216,206 @@ contract EkuboV3ExecutorTest is Constants, TestUtils {
...
+
+    constructor() {
+        vm.makePersistent(address(executor));
+        vm.makePersistent(address(liquidityHelper));
+    }
+
+    modifier setUpFork(uint256 blockNumber) {
+        vm.createSelectFork(vm.rpcUrl("mainnet"), blockNumber);
+        address(vm)
+            .call(abi.encodeWithSignature("setEvmVersion(string)", "osaka"));
```
louise-poole, line 326: Is the 'osaka' specification here needed if you set the version globally in the foundry config file?

crates/tycho-execution/contracts/test/protocols/EkuboV3.t.sol
```diff
@@ -195,6 +216,206 @@ contract EkuboV3ExecutorTest is Constants, TestUtils {
...
+        int128 liquidity
+    ) external {
+        // Start payment tracking before lock.
+        // slither-disable-next-line unused-return
+        LibCall.callContract(
+            CORE_ADDRESS,
+            abi.encodeWithSelector(
+                IFlashAccountant.startPayments.selector, poolKey.token0
+            )
+        );
```
louise-poole, line 238: Looks like these calls could silently fail (which would make debugging tests that use them pretty hard to do). Can we have a revert on failure logic here to help in that case?

louise-poole (commented):
Only some minor comments from me! Looks good!

crates/tycho-execution/contracts/test/protocols/EkuboV3.t.sol
Troshchk: It is not needed now, but I would like to keep it here if we change the version globally again

crates/tycho-execution/contracts/test/protocols/EkuboV3.t.sol
Troshchk: I think they will revert based on this [info](https://github.com/Vectorized/solady/blob/main/src/utils/LibCall.sol#L26) or do you mean something else?

crates/tycho-execution/contracts/test/protocols/EkuboV3.t.sol
louise-poole: I actually feel we should do the opposite: if we keep this here we're quietly pinning the version. If we then update the global version again to make sure the latest evm version deployed to ethereum would break anything, this test would quietly bypass that.

Troshchk: hmmm, ok, I see your point, will change it

crates/tycho-execution/contracts/test/protocols/EkuboV3.t.sol
louise-poole: Ahhhh nice! Okay, yes this is what I meant. Thanks!

louise-poole (approved):

crates/tycho-execution/contracts/test/protocols/EkuboV3.t.sol
```diff
@@ -85,6 +85,11 @@ contract EkuboV3ExecutorTest is Constants, TestUtils {
 
     modifier setUpFork(uint256 blockNumber) {
         vm.createSelectFork(vm.rpcUrl("mainnet"), blockNumber);
+        // Forks always use the default hardfork https://github.com/foundry-rs/foundry/issues/13040
+        // vm.setEvmVersion not exposed in forge-std 1.9.5 — use low-level cheatcode call
```
louise-poole: Can you change this to have a TODO to remove it once the newer Foundry stable is released?

Troshchk: Good point, changed 👍

PR #1169, opened 2026-07-08
fix: remove liquidityparty executor address

Troshchk:
Remove until the liquidityparty executor is ready


tamaralipows (approved):
Thanks! But just in the future pls include the "why" in the commit msg or PR description for our future reference

louise-poole (approved):

PR #1171, opened 2026-07-08
feat: support PropAMM pools as SignedExclusiveSwap

Troshchk:
[Task](https://propeller-heads.atlassian.net/browse/ENG-6135)

Changes summary:
- Register PropAMM (SignedExclusiveSwap) extension in the decoder
 - Extend EkuboV3 executor to route signed hops through CORE.forward(extension) 
 - Read dynamic fee from user_data and encode it


crates/tycho-execution/src/encoding/evm/swap_encoder/ekubo_v3.rs
```diff
@@ -13,6 +13,42 @@ use crate::encoding::{
...
+        return Ok(None);
+    };
+    if data.len() < SIGNED_USER_DATA_MIN_LEN {
+        return Err(EncodingError::FatalError(format!(
+            "signed user_data too short: {} bytes, need at least {SIGNED_USER_DATA_MIN_LEN}",
+            data.len()
+        )));
+    }
+    // Infallible: length validated above guarantees these slices are exact.
+    let fee = u64::from_be_bytes(data[..8].try_into().expect("8 bytes"));
```
tamaralipows: We could avoid this expect by just doing a `map_err` or something similar here no? Expects always make me nervous that someone will change the preceding code and we won't realize and cause panics

crates/tycho-execution/src/encoding/evm/swap_encoder/ekubo_v3.rs
```diff
@@ -13,6 +13,42 @@ use crate::encoding::{
...
+///
+/// Expected byte layout: `fee(8) | meta(32) | minBalanceUpdate(32) | signature(N)`.
+fn parse_signed_user_data(
+    user_data: &Option<Bytes>,
+) -> Result<Option<SignedSwapTail>, EncodingError> {
+    let Some(data) = user_data.as_ref() else {
+        return Ok(None);
+    };
+    if data.len() < SIGNED_USER_DATA_MIN_LEN {
+        return Err(EncodingError::FatalError(format!(
```
tamaralipows: Wouldn't this rather be `InvalidInput` than `FatalError`?

crates/tycho-execution/src/encoding/evm/swap_encoder/ekubo_v3.rs
```diff
@@ -13,6 +13,42 @@ use crate::encoding::{
...
+    }
+    // Infallible: length validated above guarantees these slices are exact.
+    let fee = u64::from_be_bytes(data[..8].try_into().expect("8 bytes"));
+    let meta: [u8; 32] = data[8..40]
+        .try_into()
+        .expect("32 bytes");
+    let min_balance_update: [u8; 32] = data[40..72]
+        .try_into()
+        .expect("32 bytes");
+    let signature = data[72..].to_vec();
```
tamaralipows, line 49: Isn't the signature also a fixed size?

crates/tycho-execution/src/encoding/evm/swap_encoder/ekubo_v3.rs
tamaralipows, line 295: We aren't writing this to file right? Would be nice to have an end-to-end rust-solidity test here but that depends how difficult it would be to deploy the extension contract on the test (see my comment in other PR)

tamaralipows (commented):
Thanks! Mostly questions for my understanding

crates/tycho-execution/src/encoding/evm/swap_encoder/ekubo_v3.rs
Troshchk: Ok, will change 👍

crates/tycho-execution/src/encoding/evm/swap_encoder/ekubo_v3.rs
Troshchk: The signature is variable length and depends on whether the controller is EOA or smart contract. 
Based on the [technical description](https://propeller-heads.atlassian.net/wiki/spaces/DEFI/pages/3742597126/Technical+Exploration+Fynd+PropAMM): `controller is set to the taker's key (they run Fynd locally and sign quotes)`, so I am not sure we know whether the controller is EOA or smart contract. Lmk if you think differently

crates/tycho-execution/src/encoding/evm/swap_encoder/ekubo_v3.rs
Troshchk: I will leave it open for now -- if the test in the prev PR looks good, I will also extend it here

crates/tycho-simulation/src/evm/protocol/ekubo_v3/decoder.rs
```diff
@@ -242,6 +246,18 @@ impl TryFromWithBlock<ComponentWithState, BlockHeader> for EkuboV3State {
 
                 Self::MevCapture(MevCapturePool::new(key, tick, concentrated_state, ticks)?)
             }
+            ExtensionType::SignedExclusiveSwap => {
```
louise-poole, line 249: From my understanding, this should be treated as a concentrated pool with 0 fee... Can you please make sure to overwrite the fee somewhere to reflect this?

louise-poole, line 249: Can you add a decoder test for this pool type?

crates/tycho-execution/src/encoding/evm/swap_encoder/ekubo_v3.rs
```diff
@@ -13,6 +13,45 @@ use crate::encoding::{
...
+    fee: u64,
+    meta: [u8; 32],
+    min_balance_update: [u8; 32],
+    signature: Vec<u8>,
+}
+
+/// Parses `user_data` into a signed swap tail when present.
+///
+/// Expected byte layout: `fee(8) | meta(32) | minBalanceUpdate(32) | signature(N)`.
+fn parse_signed_user_data(
```
louise-poole, line 29: Can you add tests for this? Particularly how it handles too short user_data or empty signature?

crates/tycho-execution/src/encoding/evm/swap_encoder/ekubo_v3.rs
```diff
@@ -13,6 +13,45 @@ use crate::encoding::{
...
+        )));
+    }
+    let fee = u64::from_be_bytes(
+        data[..8]
+            .try_into()
+            .map_err(|_| EncodingError::InvalidInput("fee should be 8 bytes".into()))?,
+    );
+    let meta: [u8; 32] = data[8..40]
+        .try_into()
+        .map_err(|_| EncodingError::InvalidInput("meta should be 32 bytes".into()))?;
```
louise-poole: Looks like it's impossible for this to error actually... Might be nicer to have something like:
```suggestion
    let mut meta = [0u8; 32];
    meta.copy_from_slice(&data[8..40]);
```

louise-poole: Can do the same for `fee` above this.

louise-poole (commented):
Looking good so far!

crates/tycho-simulation/src/evm/protocol/ekubo_v3/decoder.rs
Troshchk: The 0 fee is enforced at the pool initialization:
`if (poolKey.config.fee() != 0) revert PoolFeeMustBeZero();`
So I would say we do not need the overwrite here

crates/tycho-execution/src/encoding/evm/swap_encoder/ekubo_v3.rs
Troshchk: Good point -- I have added a test and change the check, so the signature is non empty. Since the length is variable, so I can only validate that it is not empty.
The actual signature validity is checked on-chain by the extension.

crates/tycho-simulation/src/evm/protocol/ekubo_v3/decoder.rs
```diff
@@ -414,7 +414,7 @@ mod tests {
             EkuboV3State::Oracle(_) => 2,
             EkuboV3State::Twamm(_) => 3,
             EkuboV3State::MevCapture(_) => 4,
-            // BoostedFees is new, no legacy format
+            // BoostedFees and SignedExclusiveSwap are new, no legacy format
```
louise-poole: Why is this comment added to the BoostedFees mapping? Isn't SignedExclusiveSwap a `EkuboV3State::Concentrated` pool?

louise-poole (approved):
One small comment, otherwise approved!

crates/tycho-execution/src/encoding/evm/swap_encoder/ekubo_v3.rs
Troshchk: Added the test here too

crates/tycho-execution/contracts/test/protocols/EkuboV3.t.sol
```diff
@@ -418,6 +418,70 @@ contract EkuboV3SignedSwapTest is Constants, TestUtils {
...
+            loadCallDataFromFile("test_ekubo_v3_signed_swap_integration"),
+            address(executor)
+        );
+
+        // --- 5. Assert ---
+        uint256 usdcSpent = usdcBefore - USDC.balanceOf(address(executor));
+        uint256 usdtReceived = USDT.balanceOf(address(executor)) - usdtBefore;
+
+        assertEq(usdcSpent, amountIn);
+        assertGt(usdtReceived, 0);
```
tamaralipows: Can we assert an exact amount here, since the block is fixed?

tamaralipows (approved):
Thanks! Just a small comment

PR #1175, opened 2026-07-09
fix(substreams): bump tycho-substreams to 0.8.0 in DCI packages

zizou0x:
## What

Bumps `tycho-substreams` to `0.8.0` in the four DCI-feeding packages that pinned pre-0.6.0 versions, and bumps each package version:

| Package | tycho-substreams before | Package version |
|---|---|---|
| ethereum-curve | git rev `655fae7` (2025-09-05) | v0.3.7 → v0.3.8 |
| ethereum-balancer-v2 | `0.5.1` | v0.4.1 → v0.4.2 |
| ethereum-balancer-v3 | git rev `51995f9` (2025-06-05) | v0.4.0 → v0.4.1 |
| ethereum-fluid | `0.5.1` | v0.3.0 → v0.3.1 (also aligns Cargo.toml, which was at 0.2.0) |

Adds a `CHANGELOG.md` to each package, and fixes clippy lints raised by the latest nightly in files rebuilt by this change (separate commit).

## Why

`get_block_storage_changes` emits native balance changes since `tycho-substreams` 0.6.0 (propeller-heads/tycho-protocol-sdk#292, 2025-10-06). Packages pinned to older versions produce block storage output without balances, so the DCI never receives balance updates for tracked contracts — their ETH balances stay frozen at the initial account snapshot.

Concrete impact: the ynETH/wstETH curve pool `0xb85d5b88…` quotes a rate built from the redemption vault's ETH balance frozen at 43.99 ETH (snapshot 2025-01-08) vs 25.90 on-chain — currently ~9.4% overpriced. Verified by running the deployed `ethereum-curve-v0.3.6.spkg` at block 25486933 (a 25.93 ETH vault outflow): 628 storage-change entries, zero `nativeBalance` fields.

## Notes

- Emitted `native_balance` is the absolute post-tx balance, so tracked accounts self-heal on their first balance change after the new spkgs are deployed. Rarely-touched accounts need a one-time balance backfill.
- New wasm = new module hashes; substreams servers will rebuild module caches from each package's initial block on first sync.
- The 2026-05-29 fix `b67b1bb57` (retain token-balance-only contract changes) exists only in the workspace crate and is not in any published release; publishing 0.9.0 and bumping again could be a follow-up.
- `ethereum-fluid` still pins `substreams-helper` to git rev `b8aeaa3` — untouched here, out of scope.

### DIFF
--- a/protocols/substreams/ethereum-balancer-v3/Cargo.toml
+++ b/protocols/substreams/ethereum-balancer-v3/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ethereum-balancer-v3"
-version = "0.4.0"
+version = "0.4.1"
 edition = "2021"
 
 [lib]
@@ -14,7 +14,7 @@ ethabi = "18.0.0"
 hex = "0.4.3"
 anyhow = "1.0.75"
 num-bigint = "0.4.4"
-tycho-substreams = { git = "https://github.com/propeller-heads/tycho-protocol-sdk.git", rev = "51995f9" }
+tycho-substreams = "0.8.0"
 itertools = "0.13.0"
 keccak-hash = "0.11.0"
 
--- a/protocols/substreams/ethereum-curve/src/pool_factories.rs
+++ b/protocols/substreams/ethereum-curve/src/pool_factories.rs
@@ -32,8 +32,10 @@ impl SerializableVecBigInt for Vec<BigInt> {
     }
     fn deserialize_bytes(bytes: &[u8]) -> Vec<BigInt> {
         bytes
-            .chunks_exact(32)
-            .map(BigInt::from_signed_bytes_be)
+            .as_chunks::<32>()
+            .0
+            .iter()
+            .map(|chunk| BigInt::from_signed_bytes_be(chunk))
             .collect::<Vec<BigInt>>()
     }
 }
--- a/protocols/substreams/Cargo.lock
+++ b/protocols/substreams/Cargo.lock
@@ -753,12 +753,12 @@ dependencies = [
  "substreams 0.5.22",
  "substreams-ethereum 0.9.13",
  "tiny-keccak",
- "tycho-substreams 0.2.1 (git+https://github.com/propeller-heads/tycho-protocol-sdk.git?rev=52d5021)",
+ "tycho-substreams 0.2.1",
 ]
 
 [[package]]
 name = "ethereum-balancer-v2"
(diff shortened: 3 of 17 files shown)


kayibal (approved):

protocols/substreams/ethereum-curve/src/pool_factories.rs
```diff
@@ -32,8 +32,10 @@ impl SerializableVecBigInt for Vec<BigInt> {
     }
     fn deserialize_bytes(bytes: &[u8]) -> Vec<BigInt> {
         bytes
-            .chunks_exact(32)
-            .map(BigInt::from_signed_bytes_be)
+            .as_chunks::<32>()
```
die-herdplatte, line 35: Requires Rust >= 1.88.0 but the toolchain is pinned to 1.83.0.
`SerializableVecBigInt` also seems to be dead (don't know why it's `#[allow(dead_code)]`).

protocols/substreams/ethereum-balancer-v3/Cargo.toml
```diff
@@ -14,7 +14,7 @@ ethabi = "18.0.0"
 hex = "0.4.3"
 anyhow = "1.0.75"
 num-bigint = "0.4.4"
-tycho-substreams = { git = "https://github.com/propeller-heads/tycho-protocol-sdk.git", rev = "51995f9" }
+tycho-substreams = "0.8.0"
```
die-herdplatte, line 17: From Claude:

Pinning crates.io 0.8.0 ships without the in-tree fix b67b1bb57 ("retain token balance contract changes"), and the workspace now contains two divergent crates both named tycho-substreams 0.8.0. The published 0.8.0 tarball predates that fix: its ContractChange::is_empty() doesn't check token_balances, while the in-repo protocols/substreams/crates/tycho-substreams (also versioned 0.8.0, used via path dep by base-lunarbase and ethereum-fermiswap) does. Balancer v3 is directly affected — map_protocol_changes builds vault InterimContractChanges carrying only upsert_token_balance entries; in a transaction where vault storage writes net out to no-ops while get_vault_reserves still records balances, the whole change is classified empty and silently dropped — the same class of silent balance-drift this PR exists to fix. Four independent finders confirmed this against the downloaded crates.io source. Fix options: use the in-tree path dependency (the pattern the newest packages already use, which also prevents this drift class recurring), or publish 0.9.0 with the fix first and pin that.

PR #1177, opened 2026-07-09
ci(substreams): make spkg builds reproducible

zizou0x:
## What

Makes substreams package builds reproducible and documents the publishing procedure:

- `release.sh` builds from inside the package directory, so rustup uses the package's own `rust-toolchain.toml` instead of the workspace root's floating `stable`, and passes `--locked` to enforce the committed `Cargo.lock`.
- The `publish-crates` job skips crate versions already on crates.io. It currently fails on `substreams-helper 0.0.2` (already published) before ever reaching `tycho-substreams`.
- Pins the workspace root toolchain to 1.96.0 and adds per-package pins to the five packages that had none (ethereum-balancer-v3, ethereum-ekubo-v2, ethereum-ekubo-v3, base-lunarbase, ethereum-liquidityparty), with minor version bumps and changelog entries.
- Bumps the three 1.75 pins (ethereum-sfrax, ethereum-sfraxeth, substreams-helper) to 1.96.0: cargo 1.75 cannot parse the workspace's version-4 `Cargo.lock`, so their release builds failed under the new per-package build. Minor version bumps and changelogs for the two spkg packages.
- Removes five stale nested `Cargo.lock` files ignored by cargo (workspace members resolve from the root lockfile). `ethereum-uniswap-v4/Cargo.lock` stays — it is its own nested workspace.
- Releases are now immutable at the registry level: the upload is an atomic S3 conditional write (`put-object --if-none-match`) that gets rejected if the spkg already exists (verified against the registry bucket: first PUT 200, second 412 PreconditionFailed). Pre-releases stay overwritable, they are keyed by commit sha.
- Fixes the `REPOSITORY` default in `release.sh`: it pointed at `s3://repo.propellerheads`, but the registry deployments consume from is `s3://repo.propellerheads-propellerheads`.
- Wires AWS credentials into the spkg release job via the same OIDC role `release.yaml` already uses (`ROLE_TO_ASSUME`/`AWS_REGION`). The job had no credentials and has never run in CI; that role already has `s3:PutObject` on the registry bucket, so no IAM change is needed.
- Rewrites the release section of `protocols/substreams/Readme.md`: the spkg job is dispatch-only, tagging alone triggers nothing. Documents release, pre-release, multi-manifest packages, and the reproducibility guarantees.

## Why

Deployed spkgs were built with whatever stable was current on release day: per-package toolchain pins were ignored in CI because the build ran from the workspace root, and nothing enforced the lockfile. Rebuilding the same commit could produce different wasm, and therefore a different substreams module hash.

## Notes

- Behavior change for all packages: the next CI rebuild of any package uses its pinned toolchain instead of current stable (e.g. ethereum-uniswap-v2 builds with its pinned 1.83.0).
- For the five newly pinned packages the toolchain that built their deployed spkgs is not recoverable, so they are pinned to current stable; their module hashes change on their next release regardless. The same applies to ethereum-sfrax and ethereum-sfraxeth: their 1.75 pin never took effect and cannot build against the current lockfile.
- Verified locally: `cargo metadata --locked` passes, wasm release builds of `base-lunarbase`, `ethereum-sfrax`, `ethereum-sfraxeth`, and `ethereum-uniswap-v2` from their package dirs succeed on their pinned toolchains, and all touched packages pass nightly `fmt` and `clippy -D warnings`.


protocols/substreams/release.sh
```diff
@@ -86,6 +89,13 @@ for yaml_file in "${yaml_files[@]}"; do
     REPOSITORY=${REPOSITORY:-"s3://repo.propellerheads/substreams"}
     repository_path="$REPOSITORY/$package/$version_prefix-$version.spkg"
```
zach030: Shouldn't the REPOSITORY be `s3://repo.propellerheads-propellerheads`? could you double-check?

protocols/substreams/release.sh
```diff
@@ -86,6 +89,13 @@ for yaml_file in "${yaml_files[@]}"; do
...
 
+    # Releases are immutable: refuse to overwrite an existing spkg. Pre-releases
+    # (pre.<sha>) may be rebuilt and overwritten.
+    if [[ "$version" != pre.* ]] && aws s3 ls "$repository_path" > /dev/null 2>&1; then
```
zach030: Does this CI job have AWS credentials?

die-herdplatte (changes requested):
`ci-substreams.yaml` is still using a stable toolchain.

`ethereum-sfrax` and `ethereum-sfraxeth` won't build and will fail with
```
error: failed to parse lock file at: .../protocols/substreams/Cargo.lock
Caused by: lock file version `4` was found, but this version of Cargo does not understand this lock file
```
Same issue in `protocols/substreams/crates/substreams-helper/rust-toolchain.toml` but the `release-substreams` workflow runs cargo from `protocols/substreams` so it won't fail.

`ethereum-uniswap-v4/Cargo.lock` is stale.

I also noticed that the Uniswap v4 packages `no-hooks`, `with-hooks`, and `shared` appear in both the root substreams manifest and in the one in `ethereum-uniswap-v4`.
This can cause some problems because the new `release.sh` builds from `ethereum-uniswap-v4`, picks the same directory as the root directory (not `../`) but the Substream manifests refer to `../target/wasm32-unknown-unknown/release/...`.

Also, (not really part of this PR but still) I'm wondering whether a shared Cargo.lock is appropriate if packages should remain reproducible and upgrading dependencies for one package touches the shared lockfile.

zizou0x:
> `ci-substreams.yaml` is still using a stable toolchain.

This is on purpose, only the build needs to be frozen. The actual CI staying up to date with latest rust version is nice. It should block because it only runs on package change. 

> `ethereum-sfrax` and `ethereum-sfraxeth` won't build and will fail with
> 
> ```
> error: failed to parse lock file at: .../protocols/substreams/Cargo.lock
> Caused by: lock file version `4` was found, but this version of Cargo does not understand this lock file
> ```
> 
> Same issue in `protocols/substreams/crates/substreams-helper/rust-toolchain.toml` but the `release-substreams` workflow runs cargo from `protocols/substreams` so it won't fail.

Nice catch, addressed in latest commit

> `ethereum-uniswap-v4/Cargo.lock` is stale.
> 
> I also noticed that the Uniswap v4 packages `no-hooks`, `with-hooks`, and `shared` appear in both the root substreams manifest and in the one in `ethereum-uniswap-v4`. This can cause some problems because the new `release.sh` builds from `ethereum-uniswap-v4`, picks the same directory as the root directory (not `../`) but the Substream manifests refer to `../target/wasm32-unknown-unknown/release/...`.

That's correct but it's pre-existing so I'd like to treat it separately. 

> Also, (not really part of this PR but still) I'm wondering whether a shared Cargo.lock is appropriate if packages should remain reproducible and upgrading dependencies for one package touches the shared lockfile.

Also correct, I've considered doing it in this scope but decided to stay targeted to this new CI workflow.

zach030 (approved):

die-herdplatte (approved):
Looks good to me!

> This is on purpose, only the build needs to be frozen. The actual CI staying up to date with latest rust version is nice. It should block because it only runs on package change.

I'd argue that it doesn't make sense to have CI use "stable" (`test` job) and "nightly" (`lint` job) when there is only one release configuration that actually matters.
By pinning to a channel, CI failures become unpredictable and potentially useless (because the release configuration may work fine).
(`rustfmt` is a different thing since it doesn't need to compile and likely breaks less often)

PR #1182, opened 2026-07-09
feat(simulation): add protocol registry and exchange_by_name builder API

tvinagre:
## What

Adds a centralized protocol registry to `tycho-simulation` so consumers can register protocol decoders by name instead of maintaining their own match blocks.

### New module: `evm/protocol/registry.rs`

A static `PROTOCOLS` table (26 rows) mapping every supported protocol name to:
- Its canonical decoder state type (via monomorphized fn pointer)
- Its default component filter (bundled, so callers can't forget required filters)

Both `entry()` and `supported_protocols()` derive from the same table — they cannot drift.

### New builder methods on `ProtocolStreamBuilder`

- `exchange_by_name(name, filter, filter_fn)` — registers a protocol by name, using the registry's default filter or a caller override. Returns `StreamError::SetUpError` for unknown names (fail-fast, listing valid protocols).
- `exchanges_by_name(names, filter)` — batch registration, errors on first unknown name.

`exchange::<T>()` is unchanged — purely additive API, backwards-compatible for all existing consumers.

### Protocols covered (26)

uniswap_v2, sushiswap_v2, quickswap_v2, pancakeswap_v2, uniswap_v3, pancakeswap_v3, uniswap_v4, uniswap_v4_hooks, ekubo_v2, ekubo_v3, aerodrome_v1, aerodrome_slipstreams, velodrome_slipstreams, cowamm, etherfi, rocketpool, erc4626, fluid_v1, lunarbase, vm:balancer_v2, vm:balancer_v3, vm:bopamm, vm:curve, vm:fermiswap, vm:maverick_v2, vm:liquidityparty

### Key design decisions
- `vm:curve` uses native `CurveState` with `curve_filter` bundled (not `EVMPoolState`)
- `velodrome_slipstreams` uses `VelodromeSlipstreamsState` (canonical, not `AerodromeSlipstreamsState`)
- Default filters bundled per protocol — no more silent missing-filter footgun
- RFQ protocols out of scope (different construction model)

## Why

Fynd's `register_exchanges` had a ~110-line match block duplicating this mapping. Every new protocol required a fynd PR; if it wasn't updated, the protocol was silently dropped. This eliminates the duplication — adding a protocol is now a one-line table entry here, and consumers pick it up automatically.

JIRA: [ENG-6246](https://propeller-heads.atlassian.net/browse/ENG-6246)

[ENG-6246]: https://propeller-heads.atlassian.net/browse/ENG-6246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/tycho-simulation/src/evm/stream.rs
+++ b/crates/tycho-simulation/src/evm/stream.rs
@@ -132,6 +132,7 @@ use crate::{
         protocol::{
             filters::uniswap_v4_non_angstrom_hook_pool_filter,
             native_wrapper::state::NativeWrapperState,
+            registry::{self, ComponentFilterFn},
             uniswap_v4::hooks::hook_handler_creator::initialize_hook_handlers,
         },
     },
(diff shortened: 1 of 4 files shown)


crates/tycho-simulation/src/evm/stream.rs
```diff
@@ -400,6 +401,51 @@ impl ProtocolStreamBuilder {
...
+    /// [protocol registry](crate::evm::protocol::registry).
+    ///
+    /// Passing `Some(filter_fn)` overrides the registry's default filter for this
+    /// protocol; `None` keeps the default (which may itself be `None`).
+    ///
+    /// # Errors
+    ///
+    /// Returns [`StreamError::SetUpError`] if `name` is not a known protocol. See
+    /// [`registry::supported_protocols`] for the valid names.
+    pub fn exchange_by_name(
```
zizou0x, line 414: Could we also have the equivalent of exchange_with_decoder_context but with name?

zizou0x, line 414: Can you use this in all location that currently have custom registry? (example, integration-tests, tycho-test, ect..)

PR #1183, opened 2026-07-09
feat(tycho-indexer): capture client metadata into Prometheus labels

brunoguerios:
Reads the generic X-Tycho-Client-Metadata header on RPC requests and WS upgrades and projects an allowlist of keys (fynd_version, fynd_preset) into Prometheus labels (ENG-6180, subtask ENG-6193). Independent of the client change: the header name is referenced by value, with no dependency on tycho-client, so this can merge before or after the client PR.

- New services/client_metadata module parses the header into a HashMap and applies defensive size caps (<=16 entries, <=64B keys, <=128B values, <=1KiB header; oversized input dropped) — the server never trusts the client to have validated.
- A key allowlist bounds label *names*, which protects the in-process metrics registry from arbitrary series. It is a projection policy (which keys become labels), not an interpretation of values: allowlisted values are emitted verbatim (missing => "none"), with no client-specific value semantics (no version-format or preset-name checks). Non-allowlisted keys are never emitted.
- Alongside the metadata labels, the same series also carry client_version (from the untouched User-Agent) and user_plan (from the upstream-injected X-User-Plan header, "none" when absent), so a connected client is identified by user_identity + client_version + user_plan + fynd_version + fynd_preset.
- Labels are added to rpc_requests, rpc_requests_failed, and websocket_connections_active. The RPC duration histogram stays endpoint-only. Per-protocol data stays on the existing websocket_subscriptions_active series, joinable by user_identity.

Bounding label *value* cardinality (a misbehaving client spamming distinct values on an allowlisted key) is left to the Prometheus scrape layer (metric_relabel_configs / sample+label limits), not tycho code. user_plan/client_version follow the same trust model as the existing user_identity label (upstream-injected headers).

Testing (unit-level, no DB/network):
- [x] `parse_client_metadata`: trims pairs, skips empty/malformed, drops oversized key/value, rejects over-cap header, caps entry count
- [x] `metric_labels`: missing keys => "none", allowlisted values pass through verbatim, non-allowlisted keys omitted
- [x] RPC middleware end-to-end (`test_rpc_metrics`): defaults (`client_version`=unknown, `user_plan`=none, metadata=none), real `client_version`/`user_plan`/metadata values, and a dropped non-allowlisted key
- [x] fmt + clippy via the CI-pinned nightly-2026-06-28
- [x] Tests run on stable

### DIFF
--- a/crates/tycho-indexer/src/services/middleware/metrics.rs
+++ b/crates/tycho-indexer/src/services/middleware/metrics.rs
@@ -203,7 +203,7 @@ mod tests {
             ("endpoint", "/v1/health"),
             ("user_identity", "alice"),
             ("fynd_version", "none"),
-            ("preset", "none"),
+            ("fynd_preset", "none"),
         ];
         assert_eq!(
             counter_value(&after, "rpc_requests", &success_labels) -
@@ -216,7 +216,7 @@ mod tests {
             ("status", "400"),
             ("user_identity", "alice"),
             ("fynd_version", "none"),
-            ("preset", "none"),
+            ("fynd_preset", "none"),
         ];
         assert_eq!(
             counter_value(&after, "rpc_requests_failed", &no_failure_labels),
@@ -240,7 +240,7 @@ mod tests {
             .insert_header(("user-identity", "alice"))
             .insert_header((
                 "x-tycho-client-metadata",
-                "fynd_version=1.2.3; preset=fast; secret=xyz",
+                "fynd_version=1.2.3; fynd_preset=fast; secret=xyz",
             ))
             .to_request();
         let response = test::call_service(&app, request).await;
@@ -252,7 +252,7 @@ mod tests {
             ("endpoint", "/v1/health"),
             ("user_identity", "alice"),
             ("fynd_version", "1.2.3"),
-            ("preset", "fast"),
+            ("fynd_preset", "fast"),
         ];
         assert_eq!(
             counter_value(&after, "rpc_requests", &metadata_labels) -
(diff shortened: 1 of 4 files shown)


crates/tycho-indexer/src/services/middleware/metrics.rs
```diff
@@ -32,7 +34,37 @@ where
...
-    counter!("rpc_requests", "endpoint" => endpoint.clone(), "user_identity" => user_identity.clone()).increment(1);
+    // Mirrors the WS `client_version` label (raw User-Agent), defaulting to a bounded sentinel.
+    let client_version = req
+        .headers()
+        .get("user-agent")
+        .and_then(|v| v.to_str().ok())
+        .unwrap_or("unknown")
+        .to_string();
+    // Client plan from the upstream-injected X-User-Plan header (`none` when absent).
+    let user_plan = req
```
brunoguerios, line 45: Original plan mentioned this as `tier`, but I think we should keep the vocabulary already present in the codebase, so I renamed it to `user_plan`. Let me know if you agree with the change.

crates/tycho-indexer/src/services/middleware/metrics.rs
```diff
@@ -32,7 +34,37 @@ where
 
     Span::current().record("user_identity", &user_identity);
 
-    counter!("rpc_requests", "endpoint" => endpoint.clone(), "user_identity" => user_identity.clone()).increment(1);
+    // Mirrors the WS `client_version` label (raw User-Agent), defaulting to a bounded sentinel.
+    let client_version = req
```
brunoguerios, line 38: The original plan required to parse `client_version` present in the `user-agent` into the rpc metrics as well, but the updated plan removed it by mistake. This addition follows the original plan intention. Let me know if you think it's unnecessary given it's already available the ws metrics and I'll remove it from here.

zizou0x (approved):

PR #1185, opened 2026-07-10
fix(tycho-client): skip catch-up wait when a synchronizer is advanced

louise-poole:
When a protocol (e.g. cowamm) reconnects and its snapshot comes back as Advanced, Delayed/Stale synchronizers were waiting their full catch-up timeout (block_time + latency_buffer ≈ 48 s on Ethereum) inside the join_all before the reinit could fire. This locked the consumer into a permanent lag behind the chain head.

Pass `skip_wait = any_advanced` to `try_advance` so that Delayed and Stale streams drain whatever is already buffered but return immediately when their channel is empty, letting the reinit happen without delay.


Troshchk (approved):
Thank you

tamaralipows (approved):
Thank you!

Issue #1189, opened 2026-07-11
[Base Aerodrome Slipstream] Stale quotes after swap-fee module replacement


0xdecr1pto:
# [Base Aerodrome Slipstream] Stale quotes after swap-fee module replacement

## Summary

Tycho’s Base Aerodrome Slipstream integration still uses fee configuration from a retired dynamic fee module after Aerodrome replaced the factory’s active module.

This causes Tycho Simulation to calculate fee `291` instead of the live fee `174`. Downstream consumers consequently receive quotes that differ from Aerodrome QuoterV2.

## Affected integration

- Chain: Base
- Protocol: `base-aerodrome-slipstreams`
- Factory: `0x5e7BB104d84c7CB9B682AaC2F3d509f5F406809A`
- Retired fee module: `0xDB45818A6db280ecfeB33cbeBd445423d0216b5D`
- Active fee module: `0x090b2A6bb475c00e2256e2095A60887cD710803b`

## Reproducible example

At Base block `48,464,451`:

- Pool: `0x4e962bb3889bf030368f56810a9c96b83cb3e778`
- Token in: cbBTC `0xcbb7c0000ab88b473b1f5afd9ef808440eed33bf`
- Token out: USDC `0x833589fcd6edb6e08f4c7c32d4f71b54bda02913`
- Amount in: `1,018,881`
- Tick spacing: `100`
- Aerodrome QuoterV2: `0x254cF9E1E6e233aa1AC962CB9B05b2cfeAaE15b0`

Observed results:

| Source | Effective fee | Amount out |
|---|---:|---:|
| Tycho Simulation | `291` | `651,057,234` |
| Aerodrome QuoterV2 | `174` | `651,133,296` |

The difference is `76,062` raw USDC, or approximately `1.1681 bps`.

A same-block local-fork A/B test confirmed causality:

1. With active module `0x090b…803b`, `pool.fee()` returns `174`, and QuoterV2 returns `651,133,296`.
2. Changing only the factory module pointer to retired module `0xDB45…6b5D` makes `pool.fee()` return `291`, and the same QuoterV2 call returns exactly `651,057,234`.

No pool state, timestamp, input, or Quoter code was changed between the calls.

## Fee calculation evidence

At the tested block:

```text
Current tick:    -64,606
600s TWAP tick:  -64,602
Absolute delta:         4
```

Retired module configuration:

```text
baseFee       = 279
feeCap        = 550
scalingFactor = 3,000,000

[...]

PR #1190, opened 2026-07-13
fix: support aerodrome slipstreams initial fees

zach030:
## Summary

Updates `base-aerodrome-slipstreams` for the replacement `DynamicSwapFeeModule` deployments and changes the rollout to a complete database rollback followed by Substreams back-processing and replay. The previously proposed SQL backfill utility has been removed.

Relates to #1189. Fee module selection remains explicit: this PR configures the two current module addresses in the SPKG rather than dynamically following `SwapFeeModuleChanged`.

Tycho Simulation consumption and fee calculation are split into #1196. Merge and deploy this PR first, let the extractor fully catch up, then deploy #1196.

### Substreams (`base-aerodrome-slipstreams` 0.1.2 → 0.1.3)

- Replace the retired fee modules with the current deployments:
  - `0x090b2A6bb475c00e2256e2095A60887cD710803b`
  - `0xF4Ecd78EBEB6d36CF7f80B5B6B41453515fe2785`
- Update the `DynamicSwapFeeModule` ABI and generated bindings for the five-field `dynamicFeeConfig`.
- Handle `InitialFeeSet` and `InitialFeeDisabled`, exposing `dfc_initialFeeEnabled` and `dfc_initialFee` alongside `dfc_baseFee`, `dfc_scalingFactor`, and `dfc_feeCap`.
- Accumulate events from the statically configured modules in a Substreams store keyed by pool and attribute.
- On the first configured-module event for a pool, emit all five configuration fields together with `dynamic_fee_module`. Fields absent from the replacement module are emitted as zero, replacing stale database attributes left by a retired module.
- On later events for that pool, emit only the fields changed by the event, preserving partial-update semantics.
- Use `dynamic_fee_module` as a version marker. The corresponding Simulation release only consumes dynamic fee attributes when this marker matches a configured replacement module.

### Migration: rollback and back-processing

Instead of patching the current database with generated SQL, restore the complete Tycho database snapshot from April 1, 2026. Its `aerodrome_slipstreams` extraction height must be lower than block `44_221_569`, before either replacement module emitted dynamic fee updates.

Because replay starts before the replacement-module history:

1. Substreams back-processes the package stores from initial block `13_843_704`, rebuilding the existing pool registry.
2. The extractor replays every block after the restored cursor with the v0.1.3 SPKG.
3. Each affected pool receives a complete five-field snapshot and module marker on its first replacement-module event.
4. Subsequent events apply partial updates normally.
5. Pools never configured by a replacement module have no matching marker, so #1196 uses default fee behavior.

The removed Python/SQL backfill remains available in the earlier Git history if it is needed later.

### Rollout

1. Stop the Slipstreams extractor and prevent Simulation from serving partially replayed state.
2. Restore the complete April 1 database snapshot, including protocol state, blocks, transactions, and extraction state. Do not roll back only the extractor cursor.
3. Build the v0.1.3 SPKG without an initial-block override. Run `substreams info <package>` and verify every module reports initial block `13_843_704`.
4. Restart the extractor at the restored cursor plus one and wait for back-processing and replay to reach the current finalized head.
5. Deploy the Tycho Simulation changes from #1196 only after the extractor is fully caught up.

A provider without matching module caches may need to back-process roughly 30 million blocks because the shared WASM change updates the existing module hashes. Pre-warm the package or budget for this initialization.

This produces the correct current state after the Factory rotations. It is not an exact reconstruction of the short historical interval between the April 1 snapshot and those rotations: retired-module updates in that interval are not replayed, while replacement-module configuration written before Factory activation is indexed immediately.

### Integration fixtures

The YAML documents two historical fixtures, currently disabled because their wide block ranges are too slow for routine integration runs:

- `0xfcd3f5ca230e7c1bd5b415eb85d5186346de0fec`: first complete snapshot at block `44_221_840`, followed by independent fee-cap and scaling-factor updates at block `44_227_070`.
- `0xb30540172f1b37d1ee1d109e49f883e935e69219`: `initialFeeEnabled = true` and `initialFee = 50`.

### Validation

- `cargo +nightly fmt --manifest-path protocols/substreams/Cargo.toml --package base-aerodrome-slipstreams -- --check`
- `cargo test --manifest-path protocols/substreams/Cargo.toml -p base-aerodrome-slipstreams --lib` (12 passed)
- `git diff --check`


protocols/substreams/base-aerodrome-slipstreams/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "base-aerodrome-slipstreams"
-version = "0.1.2"
+version = "0.1.5"
```
zizou0x: why from 0.1.2 to 0.1.5?

zizou0x: Can you add a changelog file please? I'd like to do this for all modules so we keep track of the changes

protocols/substreams/base-aerodrome-slipstreams/CHANGELOG.md
```diff
@@ -0,0 +1,111 @@
...
+  module. Later events emit only the fields changed by that event. Pools configured before the
+  SPKG cutover receive their complete initial snapshot from the backfill described below. Store
+  keys deliberately omit the module address because module rotation is handled as a static SPKG
+  configuration upgrade plus backfill, not as a runtime transition between modules for the same
+  pool.
+- Add the `dynamic_fee_module` pool attribute as a version marker. Tycho Simulation only applies
+  dynamic fee attributes when this marker matches one of the two configured modules above;
+  missing or unsupported markers fall back to the module defaults.
+- Update Tycho Simulation for the upgraded module's initial-fee behavior and use `30_000` as the
+  default fee cap.
```
die-herdplatte: This has been split into a new PR.

protocols/substreams/base-aerodrome-slipstreams/integration_test.tycho.yaml
```diff
@@ -5,9 +5,43 @@ protocol_type_names:
...
+          - "0x833589fcd6edb6e08f4c7c32d4f71b54bda02913" # USDC
+        static_attributes:
+          default_fee: "0x01f4"
+          tick_spacing: "0x0a"
+        creation_tx: "0x967196d2f84a5cb83dfdc0efa91b5e0b602b1f897f2ce81aff765b94a61724b9"
+        skip_simulation: false
+        skip_execution: true
+
+  # These historical cases remain disabled because their wide ranges are too slow for routine
+  # integration runs.
```
die-herdplatte, line 46: The above test has a way wider range than any of the disabled tests.

protocols/substreams/base-aerodrome-slipstreams/scripts/backfill_slipstreams_dynamic_fees.py
```diff
@@ -0,0 +1,780 @@
...
+    if args.input_file and (args.from_block is not None or args.modules):
+        raise ValueError("--input-file cannot be combined with --from-block or --module")
+
+    web3: Web3 | None = None
+    if args.input_file:
+        events = load_events_from_output(args.input_file)
+        if events[-1].block_number > args.to_block:
+            raise ValueError(
+                f"{args.input_file} contains an event after --to-block {args.to_block}"
+            )
```
die-herdplatte: Shouldn't there be a check for the other direction too?
Otherwise you could falsely assume to have generated snapshots for a newer block than you actually have.

die-herdplatte (commented):
lgtm!

protocols/substreams/base-aerodrome-slipstreams/Cargo.toml
zach030: fixed

zach030: Already added the [changelog](https://github.com/propeller-heads/tycho/blob/feat/update-dynamic-fee-module/protocols/substreams/base-aerodrome-slipstreams/CHANGELOG.md).

protocols/substreams/base-aerodrome-slipstreams/CHANGELOG.md
zach030: removed

protocols/substreams/base-aerodrome-slipstreams/integration_test.tycho.yaml
zach030: Already commented out.

protocols/substreams/base-aerodrome-slipstreams/scripts/backfill_slipstreams_dynamic_fees.py
zach030: Fixed in [614d06c45](https://github.com/propeller-heads/tycho/pull/1190/changes/6bf9dd9f610d1afb36edd1ed44b3c6dd7e04b169), the saved scan end must now exactly match the requested cutover block, so incomplete scans are rejected.

die-herdplatte (approved):

zizou0x (approved):

PR #1196, opened 2026-07-14
fix: support Slipstreams initial fees in simulation

zach030:
## Summary

Separates the Tycho Simulation portion of the Slipstreams dynamic-fee-module upgrade from #1190.

**Depends on #1190 (merged).** The Substreams PR introduces the versioned `dynamic_fee_module`
marker and the two initial-fee attributes consumed here.

### Changes

- Decode `dfc_initialFeeEnabled` and `dfc_initialFee` from Slipstreams component state.
- Apply the configured initial fee before the first observation is written in the current block.
- Preserve the module's sentinel behavior:
  - `initialFee = 0` falls back to the base fee.
  - `initialFee = 420` represents zero fee.
- Add `Observations::timestamp_at` for the observation-age check.
- Require a supported `dynamic_fee_module` marker. When the marker is absent or is not one of the
  two supported replacement modules, snapshot decoding fails and the pool is skipped, rather than
  trusting stale fee attributes left by a retired module. Delta updates that carry the marker apply
  the same check.
- Support partial attribute updates without replacing unchanged dynamic-fee fields.
- Raise `DEFAULT_FEE_CAP` from `10_000` to `30_000` to match the replacement modules.

### Behavioral note

Pools without a supported marker are now skipped, not served with default fees. This diverges from
the "default fee behavior" described in #1190 for unmatched pools. With `skip_state_decode_failures`
enabled the pool is dropped and logged; without it, the decode error is fatal for the stream, so the
extractor must have fully replayed the supported-module markers before this runs.

## Merge order

1. #1190 (Substreams, SPKG, attributes) — merged.
2. This PR (Tycho Simulation consumption and fee calculation), after the extractor has caught up.

## Validation

- `cargo clippy -p tycho-simulation --all-features --tests` — clean.
- `cargo test -p tycho-simulation --all-features slipstreams` — 31 passed.


crates/tycho-simulation/src/evm/protocol/aerodrome_slipstreams/state.rs
```diff
@@ -516,27 +516,11 @@ impl ProtocolSim for AerodromeSlipstreamsState {
...
-            .updated_attributes
-            .get("dfc_scalingFactor")
-        {
-            self.dfc
-                .update_scaling_factor(u64::from(dfc_scaling_factor.clone()));
-        }
+        self.dfc
+            .update_from_attributes(&delta.updated_attributes)
+            .map_err(|attribute| {
+                TransitionError::DecodeError(format!("Missing dynamic fee attribute: {attribute}"))
```
zizou0x: Small comment but `Missing dynamic fee attribute` isn't a nice error message IMO. It's leaking context that it shouldn't know about. Here we should just error with "Failed to update dynamic fee module config" and "Missing attribute..." should be given by the returned error from `update_from_attributes()`. That way each function scope is only logging things that it knows about and doesn't assume anything. 

The reason is, if we ever update `update_from_attributes` and now it can error with something like "failed to convert ... to u64" then we would get a very weird error "Missing dynamic fee attribute: failed to convert ... to u64"

crates/tycho-simulation/src/evm/protocol/utils/slipstreams/dynamic_fee_module.rs
```diff
@@ -1,34 +1,128 @@
...
-    pub fn update_scaling_factor(&mut self, scaling_factor: u64) {
-        self.scaling_factor = scaling_factor;
+    pub(crate) fn from_attributes(
+        attributes: &HashMap<String, Bytes>,
+    ) -> Result<Self, &'static str> {
+        if !attributes
+            .get(DYNAMIC_FEE_MODULE_ATTRIBUTE)
+            .is_some_and(|module| is_supported_dynamic_fee_module(module))
+        {
+            return Ok(Self::default());
```
zizou0x: So if we don't know the fee module we return Ok with all fees set to 0? Isn't this dangerous?

crates/tycho-simulation/src/evm/protocol/utils/slipstreams/dynamic_fee_module.rs
```diff
@@ -1,34 +1,128 @@
...
+        &mut self,
+        attributes: &HashMap<String, Bytes>,
+    ) -> Result<(), &'static str> {
+        if attributes.contains_key(DYNAMIC_FEE_MODULE_ATTRIBUTE) {
+            *self = Self::from_attributes(attributes)?;
+            return Ok(());
+        }
+        if !self.module_is_supported {
+            return Ok(());
+        }
```
zizou0x: What happens then? Are we ignoring this pool or continuing without fees?

crates/tycho-simulation/src/evm/protocol/aerodrome_slipstreams/state.rs
zach030: fixed, now only wraps with the context it actually knows.

crates/tycho-simulation/src/evm/protocol/utils/slipstreams/dynamic_fee_module.rs
zach030: It now returns an error instead of `Ok(default)` when the `dynamic_fee_module` is missing or unsupported, so the pool gets skipped at decode.

crates/tycho-simulation/src/evm/protocol/utils/slipstreams/dynamic_fee_module.rs
zach030: I removed the flag, so a pool with an unsupported module can no longer exist in memory, and won't be processed in `delta_transition`.

zizou0x (approved):

PR #1198, opened 2026-07-14
fix(substreams): pin tycho-substreams 0.8.1 across integrations

zizou0x:
## Why

`tycho-substreams 0.8.1` (published from #1197 + #1199) contains the token balance retention fix from #1056: with `0.8.0`, contract changes carrying only token balance updates are classified empty and dropped by `TransactionChangesBuilder`. Of the packages here, `ethereum-balancer-v3` and `ethereum-fermiswap` are concretely affected (they populate contract token balances via `upsert_token_balance`); the others are updated for consistency — the shared workspace lockfile resolves all `0.8.x` requirements to a single version anyway.

## What

- Pin `ethereum-curve`, `ethereum-balancer-v2`, `ethereum-balancer-v3` and `ethereum-fluid` to `tycho-substreams 0.8.1`.
- Switch `ethereum-fermiswap` and `base-lunarbase` from path dependencies on the in-tree crate to the published `0.8.1`.
- Remove the dead `SerializableVecBigInt` trait from `ethereum-curve`: it is `#[allow(dead_code)]`, referenced nowhere, and its `as_chunks` call requires Rust 1.88+, breaking builds on the package's pinned 1.83.0 toolchain.
- Bump all touched package versions, align the substreams manifest versions with them (balancer-v3 and lunarbase manifests were already lagging their Cargo.toml versions), and add changelog entries.
- Update `Cargo.lock` to the published `tycho-substreams 0.8.1`; all six packages pass `cargo check --locked`.


zach030 (approved):
lgtm

PR #1200, opened 2026-07-14
docs: Protocol testing re-added Chainnodes as supported provider

timolson:
Chainnodes just fixed `debug_storageRangeAt` (for the second time), so this PR re-adds it to the protocol integration testing docs as another supported RPC provider.


zizou0x (approved):
LGTM, thank you!

PR #1202, opened 2026-07-15
feat: add Plasma chain support

tamaralipows:
# Only merge after timelock expiry. Safe to merge Monday morning.

## Summary

Adds support for Plasma (chain id 9745), where the TychoRouter, FeeCalculator, and the uniswap_v3, fluid_v1, vm:curve, and native_wrapper executors are now deployed.

- Deploy tooling: plasma network config, executor deploy definitions, and role assignments. `ROUTER_FEE_SETTER` now comes from `roles.json` instead of an env var, so the fee calculator and router deploy scripts share one source of truth.
- `Chain::Plasma` as a first-class chain in tycho-common, mirrored in the dto, tycho-client, and Python client.
- Deployed contract addresses registered in the encoder config and documented on the contract addresses page. All addresses and constructor args were verified on-chain against https://rpc.plasma.to.


Troshchk (commented):
Thank you!
Why have you added only these executors, not the other ones?

crates/tycho-common/src/models/mod.rs
```diff
@@ -361,6 +382,10 @@ impl Chain {
             (Chain::Polygon, TvlThresholdTier::Low) => 200_000.0,
             (Chain::Polygon, TvlThresholdTier::Medium) => 2_000_000.0,
 
+            // Plasma (XPL ≈ $0.10): 200_000 XPL ≈ $20K, 2_000_000 XPL ≈ $200K
+            (Chain::Plasma, TvlThresholdTier::Low) => 200_000.0,
+            (Chain::Plasma, TvlThresholdTier::Medium) => 2_000_000.0,
+
```
Troshchk, line 388: I understand the setup is the same as for the other branches, but do we update it if the XPL price changes?

tamaralipows: Hmm this is just for the default TVL threshold so I don't think it's suuuuper important to update, the user can always override it. But yeah if it absolutely crashes or explodes and we see a noticeable difference in the pools being filtered then we can update it

Troshchk (approved):
Thank you!

PR #1211, opened 2026-07-16
feat: detect token slots at the latest block

die-herdplatte:
## What

Balance and allowance storage-slot detection no longer takes a block hash. Every detection request (the `debug_traceCall` prestate scan and the `eth_call` verification) now queries the latest block instead of a caller-supplied block.

- `BalanceSlotDetector::detect_balance_slots` and `AllowanceSlotDetector::detect_allowance_slots` drop their `block_hash` parameter.
- The `tycho-ethereum` slot-detector RPC methods and the entrypoint tracer's `create_trace_call_params` are updated accordingly; the entrypoint tracer's own tracing still pins the block it is given.
- Detection failures in the execution test harness are now logged with the token and error instead of being silently skipped.

## How it was discovered

Running `tycho-integration-test` against a local indexer synced to the Polygon chain tip, slot detection intermittently failed with "Couldn't find storage slots for token". The chain of causes:

1. Detection pinned the block hash of the update it was handed — often the freshly produced tip block.
2. Load-balanced RPC backends had not yet made that block canonical on the node that served the request, so the trace returned `result: null` alongside a "hash not currently canonical" error.
3. `alloy` rejects a JSON-RPC response carrying both `result` and `error`, and this failed the deserialization of the entire batched request — discarding the valid results for every other token in the batch.
4. The per-token errors were then swallowed with a bare `continue`, so the only visible symptom was missing slots downstream.

Querying the latest block removes the trigger: the node always has a canonical latest block, so the trace does not return the non-canonical error, and the batch is not poisoned.

## Why this is safe

A token's storage layout is fixed, so the slot location does not depend on chain state at any particular block. Detection results are already cached per `(token, params)` with no block component, so passing a block never affected a cached lookup. The only caller that pinned a block — the dynamic contract indexer — passed the hash of the block it was processing, which could only hurt: a `debug_traceCall` pinned to a historical block during sync needs archive state a full node may not retain, and pinning a fresh tip block exposes the same non-canonical-block race observed in integration testing. Querying `latest` avoids both. Cases that need a fixed block for reproducible on-chain values (e.g. the rebasing-token balance assertion) use the general RPC primitives directly and keep their block selection.

## Breaking change

`BalanceSlotDetector::detect_balance_slots` and `AllowanceSlotDetector::detect_allowance_slots` drop their `block_hash` parameter.

## Testing

- Format and clippy clean; workspace unit tests pass.
- Slot-detector tests exercising the latest-block detection path directly, against live nodes: batch trace/tests, allowance detection, balance detection, the stETH rebasing verify (Ethereum), and the Arbitrum USDC/WETH precompile-selection regression case — all pass.
- End-to-end check: a `tycho-integration-test` run on the Polygon tip drove the detection path for the pool token set and executed override-based swaps at 0% slippage, confirming the detected slots are correct. Detection is cached per token, so this covers first-time detection of each token rather than every block.

---

This supersedes #1192, which was opened from a fork. Fork-based PRs require extra permissions because CI deploys to the `integration-tests` environment, so this PR re-opens the same branch from within the main repository.


crates/tycho-common/src/traits.rs
```diff
@@ -233,10 +233,12 @@ pub trait BalanceSlotDetector: Send + Sync {
     /// Detect balance storage slots for multiple tokens from a single holder.
     /// Useful to allow overriding balances.
     ///
+    /// No block parameter is taken: a token's storage layout is fixed, so the slot location does
```
zizou0x: In the proxy case it's not technically fixed no? 

But changing the map slot it would mean losing all current value which makes it almost impossible, correct?

zizou0x (approved):
Makes sense, LGTM. 🚀

crates/tycho-common/src/traits.rs
die-herdplatte: Yes, not technically, but practically, migrations would be very expensive and idk this has ever happened.
Clarified it in [f60aaf9](https://github.com/propeller-heads/tycho/pull/1211/commits/f60aaf910d851331f316371578ae143d46e78a08)

zach030 (approved):
lgtm!

PR #1212, opened 2026-07-16
feat: integrate the Titan pAMM price level stream

die-herdplatte:
Adds a taker-side integration of Titan's pAMM price level stream
(https://docs.titanbuilder.xyz/propamms/takers): per-block quote-ladder snapshots are turned
directly into simulation `Update`s, and fills execute through a single generic executor for the
standard `IPropAMM` interface.

## tycho-simulation: `price_level_stream` module (default feature `price-level-stream`)

- `titan.rs`: client for the `pamm_price_levels` WebSocket — parses frames, reconnects forever
  with capped exponential backoff, configurable connect/read-idle timeouts. A captured live
  frame serves as a parse fixture.
- `PriceLevelStreamBuilder`: a new builder serves no venues; `with_known_pamms()` registers
  the known-executable ones (FermiSwap, Kipseli, Metric, Bebop, TaurusFi — router addresses;
  per-venue gas calibrated by replaying real fills) and denies the known-unexecutable ones
  (currently Tempest, whose `swap` enforces a taker allowlist: replays of real fills revert
  with `TakerNotAllowed()` for arbitrary callers, so routed swaps would revert). `add_pamm`
  registers or overrides a venue per address, `deny_pamm` excludes one (dropping any
  registration and blocking auto-detection), and opt-in `auto_detect` serves unknown streamed
  venues under their address, with `auto_detected_gas_cost` overriding the conservative default
  gas those venues are quoted with. Between `add_pamm` and `deny_pamm` the later call wins;
  `with_known_pamms` defaults never override either, regardless of call order. Components are
  identified as `pricelevelstream:{venue}`, with the venue address in the `pamm_address`
  static attribute and the component id `pamm ++ token0 ++ token1`. Snapshots never move
  backwards: a frame targeting an older block than an already processed one (e.g. replayed
  around a reconnect) is skipped wholesale.
- `PriceLevelStreamState` (`ProtocolSim`): linear interpolation between ladder quotes. Amounts
  below the smallest quote are rejected (extrapolating below the first genuine sample was
  measured to be wildly wrong); amounts above the largest get a partial result at the limit.
  Ladders that are not monotonically increasing in `amount_out` (stream glitches) are rejected
  as unreliable, and zero-amount quotes are skipped for the spot price. `new_state` returns
  consumed ladders so a second same-block fill cannot double-count the maker's liquidity.
- Liquidity may overlap with other integration paths of the same venue (e.g. `vm:fermiswap`);
  consumers deduplicate by venue where double-counting matters.

## tycho-execution: generic PropAMM execution

- `PropAMMExecutor`: one stateless executor for every pAMM implementing the standard `IPropAMM`
  interface (push payment; calldata is `pamm ++ tokenIn ++ tokenOut`). Both live routers
  implement the interface (verified on-chain). Because the caller supplies the venue address,
  the executor is modeled in the router security model (`model/src/model/executors.rs`).
- `PropAMMSwapEncoder` with a registry fallback: `get_encoder` resolves any
  `pricelevelstream:{venue}` — including auto-detected ones — through the single
  `pricelevelstream` family config entry; exact per-venue keys override.
- The gas estimator classifies `pricelevelstream:*` as optimizable-transfer-in (the venue is
  funded directly, like uniswap_v2).

## tycho-integration-test: stream processor

- Emits one sampled update per `--price-level-stream-block-interval` blocks (default 1). The
  chosen block's latest snapshot is held back until the stream moves to the next block, so the
  target block is already finalizing when block alignment starts polling.
- Execution is simulated at exactly the quoted block with no oracle overrides: it succeeds only
  when the pAMM's oracle state is fresh at that block (for the registry-priced venues, only
  when Titan built it and the maker's update landed on-chain). Freshness-guard reverts — the
  registry's `StaleUpdate()` for every venue, plus Metric's `FeedStalled()` — are recorded as
  the expected outcome (`tycho_integration_execution_stale_quotes_total`) instead of failures.
- `--max-blocks` terminates price-level-stream-only runs too: the stream's target blocks count
  toward the cap, deduplicated against the on-chain stream's blocks.
- Liveness watchdog (served venues marked stale on Titan silence), per-venue pair-count gauge,
  and a target-block-miss counter; target-block polling retries transient RPC errors instead of
  counting them as misses.
- The stream is also wired into the `price_printer` example.

Observed on live fill blocks: quote-vs-executed deviation within ±0.003%, venue swap gas
~180k (FermiSwap) / ~320k (Kipseli) / ~225k (Metric) / ~135k (Bebop) / ~105k (TaurusFi).

## Deployment status

The `PropAMMExecutor` is deployed at `0xc196f22d69d41A2B1DbF6ef8d28C09bc2CA4E4fa`, wired into
production `executor_addresses.json` as the single `pricelevelstream` entry, and active on the
router since 2026-08-16 (executor timelock elapsed). Validated end-to-end against the live
executor in a 30-minute integration-test run:
https://github.com/propeller-heads/tycho/pull/1212#issuecomment-5328848437

## Potential follow-ups

- A TVL-style filter on the stream: `tvl_filter` has no equivalent here, but `get_limits`
  exposes ladder depth that could gate small or malformed books.
- A `price_level_stream_quickstart` example, like the existing `quickstart` and
  `rfq_quickstart`.

---

This supersedes #1204, which was opened from a fork. Fork-based PRs require extra permissions because CI deploys to the `integration-tests` environment, so this PR re-opens the same branch from within the main repository.


crates/tycho-simulation/src/price_level_stream/state.rs
```diff
@@ -0,0 +1,471 @@
...
+                token1 = %self.token1,
+                %amount_in,
+                lower_amount_in = %lower.amount_in,
+                lower_amount_out = %lower.amount_out,
+                upper_amount_in = %upper.amount_in,
+                upper_amount_out = %upper.amount_out,
+                "Quote ladder is not monotonically increasing in amount_out; falling back to \
+                 the lower quote"
+            );
+            return lower.amount_out.clone();
```
zach030: Returning the lower quote when amount_out isn't monotonic feels a bit off. Can we fail fast and return an error here instead?

crates/tycho-integration-test/src/main.rs
```diff
@@ -618,6 +711,44 @@ async fn refresh_token_prices(chain: Chain, prices: SharedTokenPrices, interval:
...
+///
+/// Unlike [`poll_rpc_for_block`], an RPC that has already moved past the target is not treated
+/// as stale: the target block is fetched by number, since the caller wants to simulate at that
+/// exact (recent) block.
+async fn await_target_block(
+    rpc_tools: &tycho_test::RPCTools,
+    target_block: u64,
+    max_attempts: u32,
+    poll_interval: Duration,
+) -> Option<Block> {
```
zach030: I'd recommend returning `Result<Option<Block>>` and keeping the target block, attempt, and RPC operation context.

crates/tycho-simulation/src/price_level_stream/state.rs
```diff
@@ -0,0 +1,471 @@
...
+                new_state: self.consumed(),
+            };
+            return Err(SimulationError::InvalidInput(
+                format!(
+                    "Not enough liquidity to support complete swap. input amount: {amount_in}, maximum quoted amount: {}",
+                    last.amount_in
+                ),
+                Some(res),
+            ));
+        }
```
zach030: I tested it with the `titan_getPammQuoteVenue` API. The upper-bound check matches the live Titan quote helper: querying even one token unit above the max published amountIn returns `-32000: amountIn exceeds quoted depth`.

But the lower-bound doesn't match: even with an amountIn below the lower bound, it still returns a quote, and it looks like it scales quotes linearly between (0,0) and the first level.

crates/tycho-simulation/src/price_level_stream/stream.rs
```diff
@@ -0,0 +1,561 @@
...
+    }
+
+    /// Processes one frame into an [`Update`], or `None` if the frame contains nothing relevant
+    /// (no registered pAMM with at least one known pair or a pair removal).
+    fn process(&mut self, message: TitanPriceLevelMessage) -> Option<Update> {
+        let mut states: HashMap<String, Box<dyn ProtocolSim>> = HashMap::new();
+        let mut new_pairs = HashMap::new();
+        // The frame is a complete snapshot: every known component is presumed gone until the
+        // frame re-emits it below.
+        let mut previous = std::mem::take(&mut self.components);
```
zach030, line 275: I think we should check the block ordering here. If we replay block N-1, it would mess up the current components' state.

crates/tycho-simulation/src/price_level_stream/state.rs
die-herdplatte: Fixed in f728f5929

crates/tycho-integration-test/src/main.rs
die-herdplatte: Fixed in 733518c9bac728ec71189479449e7614560df9c4

crates/tycho-simulation/src/price_level_stream/state.rs
die-herdplatte: I tested the lower-bound side of this a bit more on FermiSwap, your observation of linear scaling below the lower bound is correct.
There are two issues with this though:
- FermiSwap (and possibly other pAMMs) enforces a minimum trade size (~1e13 wei WETH ≈ $0.02 reverts with `COR`)
- On the (still) broken cbBTC/USDC FermiSwap ladder, `titan_getPammQuoteVenue` returns 1 cbBTC -> 0.02 USDC

These two reasons show how unsafe it can be to extrapolate imo.
I added a comment in 35e038ca9.

crates/tycho-simulation/src/price_level_stream/stream.rs
die-herdplatte: Oh yeah, nice one, fixed in 68f14c0c0

crates/tycho-execution/contracts/src/executors/PropAMMExecutor.sol
```diff
@@ -0,0 +1,73 @@
...
+    // slither-disable-next-line locked-ether
+    function swap(uint256 amountIn, bytes calldata data, address receiver)
+        external
+        payable
+    {
+        (address pamm, address tokenIn, address tokenOut) = _decodeData(data);
+
+        // slither-disable-next-line unused-return
+        IPropAMM(pamm)
+            .swap(tokenIn, tokenOut, amountIn, 0, receiver, block.timestamp);
```
zizou0x, line 41: I wonder if this is acceptable security wise? It's similar to what we do for Uniswaps but given that this is given by websocket and not Tycho itself, I think we need a safeguard in case titan stream gets compromised. Not 100% sure if the safeguard should be here on if it's fine to have it in tycho-simulation directly. It kinda kills the auto-detect feature but I think it's the only way.

crates/tycho-simulation/src/price_level_stream/state.rs
```diff
@@ -0,0 +1,490 @@
...
+impl ProtocolSim for PriceLevelStreamState {
+    fn fee(&self) -> f64 {
+        0.0
+    }
+
+    fn spot_price(&self, base: &Token, quote: &Token) -> Result<f64, SimulationError> {
+        let quotes = self.quotes(&base.address, &quote.address)?;
+        let best = quotes
+            .iter()
+            .find(|q| q.amount_in > BigUint::ZERO)
```
zizou0x: Should we also filter out `q.amount_out == 0`? I think if people do `1/spot_price` they would get issues

crates/tycho-execution/contracts/src/executors/PropAMMExecutor.sol
die-herdplatte: Deemed safe from the execution team as per Slack

crates/tycho-simulation/src/price_level_stream/state.rs
die-herdplatte: Fixed in c5b0aec

die-herdplatte:
**Live validation of the Metric venue (30-minute fill-replay comparison, 2026-08-13)**

Setup: recorded the stream's Metric ladders per target block while watching all 6 Metric pools for fills (pool-level swap events, not just the router's `Swapped` — most Metric flow enters the pools directly through other fronts, e.g. aggregators using their UniswapV3-style interface). Each fill's amounts were reconstructed from its receipt's transfer events and run through the same linear interpolation `PriceLevelStreamState` uses, against the snapshot targeting exactly the fill block.

Coverage: 151 blocks, ~17k frames (Metric books present in 98% of frames), 42 fills across 5 of the 6 pools. Every fill block had a captured snapshot.

Results:

- **Quotes are execution-accurate.** 17 fills landed inside the quoted ladder: median deviation 0.09 bps, p95 1.4 bps, no systematic bias — holding up to fills consuming 58% of ladder depth. This matches the accuracy previously observed for FermiSwap (the observed-numbers line in the description) and validates the quote/interpolation path for Metric.
- **Half the fills (21/42) were on pairs the stream doesn't carry** — mostly USDe and sUSDe pairs. Metric's public fill volume overstates what is reachable through the price level stream (WETH/USDC/USDT books only). A few no-book fills were on nominally streamed directions, so Metric's books can occasionally drop out of a frame, like FermiSwap's.
- **Stream depth < venue capacity.** 4 fills exceeded the ladder's max input and executed at near-flat rates well above the ladder's clamped tail (up to ~4x on an 87k USDC→USDT fill). Same pattern as FermiSwap: `get_limits` under-advertises — the safe direction (lost flow, never bad fills).
- **0 of the 42 fills went through the router**, consistent with its ~28 fills/day base rate; execution through the `IPropAMM` router was verified separately by standalone replay (gas ~203k, permissionless).

Not covered by this run: the executor leg (deployment still timelocked) and below-first-level fills (none occurred).

die-herdplatte:
## Live-executor validation: 30-minute integration-test run (2026-08-18)

With the PropAMMExecutor's activation timelock expired (2026-08-16 09:25 UTC), I ran
tycho-integration-test against the real deployed executor — no state overrides, no bypass:

```
tycho-integration-test --chain ethereum --tycho-url tycho-beta.propellerheads.xyz \
  --disable-onchain --disable-rfq --price-level-stream-block-interval 1
```

~30 minutes / ~150 blocks. Every quote is simulated (`get_limits` + `get_amount_out`), encoded
through the router, and executed via `debug_traceCall` at exactly the quoted block.

### Per-venue execution outcomes

| pAMM | successes | stale-skips | real reverts | non-stale success rate |
|---|---|---|---|---|
| metric | 207 | 168 | 1 | 99.5% |
| fermiswap | 41 | 1,482 | 0 | 100% |
| kipseli | 4 | 386 | 0 | 100% |
| bebop | 2 | 288 | 0 | 100% |
| taurusfi | 0 | 140 | 0 | — (no fresh-oracle attempt) |
| `0xcf211b4d…` (auto-detected) | 0 | 122 | 0 | — |

- **Stale-skips are the designed outcome**, not failures: a pAMM quote is only executable in a
  block whose oracle update landed on-chain. The `StaleUpdate`/`FeedStalled` classification keeps
  them out of the revert metrics (`tycho_integration_execution_stale_quotes_total`).
- **Metric executes in ~55% of attempts** because its guard works differently from the registry
  venues'. `FeedStalled` checks that Metric's own oracle was updated within a staleness window,
  and those oracle updates land in ordinary blocks all day — Metric's pools serve aggregator flow
  directly (~1,700 fills/day), which only executes on a fresh feed, so the maker keeps the feed
  current regardless of who builds the block. The registry venues' `StaleUpdate` instead requires
  the priority-update lane to be stamped for the specific quoted block, which happens as part of
  Titan's fill flow — so their execution attempts only pass in Titan fill blocks (fermiswap's
  41/1,523 matches its known fill cadence).
- Metric's single real revert was `PartialFill(uint256,uint256)` — the venue could not fill the
  full amount; genuine and rare (0.5% of non-stale attempts).
- taurusfi produced no fresh-oracle attempt in the window, so its execution path is still
  unexercised end-to-end.

### Where execution ran, it matched simulation

Slippage between simulated and executed output was ~0 everywhere: metric mean ~0.000% over 207
executions, fermiswap −0.02% over 41, kipseli/bebop exactly 0. The PLS ladder interpolation
prices what the router actually executes.

### Simulation and stream health

- ~3,200 `get_limits`/`get_amount_out` calls across all venues: **zero failures**.
- Zero state-validation failures, zero target-block misses.
- A sixth, unknown pAMM appeared mid-run and was served by auto-detection under its address
  (`pricelevelstream:0xcf211b4dd0d2be5c173ea57bcf938fc61d1d3bd3`): an EIP-1967 proxy deployed
  2026-08-14 whose verified implementation is an unbranded reference-style `PropAMM` reading the
  shared PrioUpdateRegistry (`0xDa7AfeeD…`), absent from Titan's venue docs and pamm.wtf. Exactly
  the case the auto-detect fallback exists for; its `StaleUpdate` guard is already classified.

### Caveats

- One run window; the per-venue success counts for kipseli/bebop are small (their Titan fill
  cadence is low), so their rates are existence proofs rather than statistics.
- Coverage asymmetry stands: only quotes coinciding with on-chain oracle freshness are
  execution-tested (~2-3% of attempts outside metric).

crates/tycho-integration-test/src/stream_processor/price_level_stream_processor.rs
```diff
@@ -0,0 +1,467 @@
...
+                            // throttling — not with the consumer of the emitted updates.
+                            emitter.mark_stale();
+                            continue;
+                        }
+                    }
+                };
+                let Some(update) = next else { break };
+                if let Some(result) = emitter.handle(update) {
+                    if stream_tx.send(result).await.is_err() {
+                        warn!("Receiver dropped, stopping stream processor");
```
tvinagre, line 91: Can we add a reconnect here? Or are we doing this on the outside caller?

crates/tycho-integration-test/src/metrics.rs
```diff
@@ -44,6 +44,17 @@ pub fn initialize_metrics() {
         "tycho_integration_simulation_execution_reverts_total",
         "Total number of reverted execution simulations"
     );
+    describe_counter!(
+        "tycho_integration_execution_stale_quotes_total",
+        "Price level stream executions skipped because the quoted block was not built by Titan \
+         or the pAMM's oracle update did not land on-chain (freshness-guard revert; expected \
+         outcome)"
+    );
+    describe_counter!(
```
tvinagre, line 53: would probably add one for errors / disconnects

crates/tycho-simulation/src/price_level_stream/titan.rs
```diff
@@ -0,0 +1,346 @@
...
+                            Err(e) => {
+                                warn!(error = %e, "Titan price level stream read error; reconnecting");
+                                break;
+                            }
+                        }
+                    }
+                }
+                // Connection refused / TLS error — fall through to backoff and retry.
+                Ok(Err(e)) => {
+                    warn!(error = %e, "Failed to connect to Titan price level stream; retrying");
```
tvinagre, line 197: Nice!

tvinagre (approved):
small comments that can be addressed on a follow-up. LGTM!

PR #1213, opened 2026-07-16
feat: address audit findings

Troshchk:
[Task](https://propeller-heads.atlassian.net/browse/ENG-6113)

**slippageToleranceBps -> expectedAmountOut**
We originally used slippageToleranceBps to make it less obvious that users could change expectedAmountOut to avoid positive slippage fees.
But this does not really prevent bypassing and adds extra calculation and complexity.
So we decided to go back to passing expectedAmountOut and minAmountOut directly.


crates/tycho-execution/contracts/src/TychoRouter.sol
```diff
@@ -1267,9 +1299,20 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
...
 
+        uint256 totalFees = 0;
+        for (uint256 i = 0; i < fees.length; i++) {
+            totalFees += fees[i].feeAmount;
+        }
+        if (totalFees > feeInput.actualAmountOut) {
+            revert TychoRouter__FeesExceedOutput(
+                totalFees, feeInput.actualAmountOut
+            );
+        }
```
louise-poole, line 1271: Can you add a test case for this guardrail?

crates/tycho-execution/contracts/src/TychoRouter.sol
```diff
@@ -790,7 +805,8 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
     }
 
     /**
-     * @notice Internal implementation of the core swap logic shared between singleSwap() and singleSwapPermit2().
+     * @notice Internal implementation of the core swap logic shared
+     *         between singleSwap() and singleSwapPermit2().
```
louise-poole: Some of this reformatting seems unintentional?

crates/tycho-execution/contracts/src/TychoRouter.sol
louise-poole: Had an idea here to remove the duplication: Maybe we can group all of this into a `_finalizeSwap` fn and do what we did with _takeFees where we put all the input vars into a struct and pass only 1 pointer? This should help avoid stack too deep issues.

crates/tycho-execution/contracts/src/TychoRouter.sol
```diff
@@ -728,19 +735,24 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
...
-        if (slippageToleranceBps >= MAX_SLIPPAGE_BPS) {
-            revert TychoRouter__SlippageBpsTooHigh(slippageToleranceBps);
+        // Rejects minAmountOut == 0 implicitly: expectedAmountOut is non-zero,
+        // so the slippage floor is always positive.
+        if (
+            minAmountOut > expectedAmountOut
+                || minAmountOut * BPS_DENOMINATOR
+                    < expectedAmountOut
+                        * (BPS_DENOMINATOR - MAX_SLIPPAGE_TOLERANCE_BPS)
+        ) {
```
louise-poole: I find this math hard to follow laid out like this. Maybe it can be done first by calculating the floor:
```rust
  // Lowest acceptable output: at most MAX_SLIPPAGE_TOLERANCE_BPS below expected.
  uint256 minAmountOutFloor =
      (expectedAmountOut * (BPS_DENOMINATOR - MAX_SLIPPAGE_TOLERANCE_BPS)) / BPS_DENOMINATOR;
```
Then the if condition is easier to understand as `minAmountOutFloor <= minAmountOut <= expectedAmountOut`

louise-poole: These checks are duplicated 3 times. I think it might be cleaner if we deduplicate it and have 2 validation helpers: _validateAmounts and _validateAddresses?

louise-poole (commented):
Thanks! All audit findings look well addressed! I thought a bit deeper on some thing Max mentioned in the slack threads and left some comments that I think should help further than what he explicitly asked for. Let me know what you think?

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -226,66 +199,18 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
...
+     * @dev Calculates the positive slippage surplus, all of which goes to the router
+     * @return routerCut The full surplus (zero if disabled or no surplus)
      */
     function _calculatePositiveSlippage(
         uint256 actualAmountOut,
-        uint256 expectedAmountOut,
-        address client
-    ) internal view returns (FeeRecipient[] memory) {
+        uint256 expectedAmountOut
+    ) internal view returns (uint256 routerCut) {
```
tamaralipows: I think this is not the `routerCut` anymore since there is no client cut? it's just positiveSlippage

crates/tycho-execution/contracts/src/TychoRouter.sol
```diff
@@ -241,13 +248,13 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
...
-     * - Reverts with `TychoRouter__NegativeSlippage` if slippage exceeds `slippageToleranceBps`.
+     * - Reverts with `TychoRouter__NegativeSlippage` if the final output is below `minAmountOut`.
      *
      * @param amountIn The input token amount to be swapped.
      * @param tokenIn The address of the input token. Use `ETH_ADDRESS` for native ETH
      * @param tokenOut The address of the output token. Use `ETH_ADDRESS` for native ETH
-     * @param expectedAmountOut The quoted output amount used to compute slippage.
-     * @param slippageToleranceBps Maximum slippage in basis points (0-10_000).
+     * @param expectedAmountOut The quoted output amount; used to detect positive slippage.
+     * @param minAmountOut The minimum acceptable output amount (revert guardrail). Must not exceed `expectedAmountOut` and be at most 20% below it.
```
tamaralipows: Let's just refer to the const in the docstring. Afraid we will change that 20% const value one day and forget to change the docstring

crates/tycho-execution/contracts/src/TychoRouter.sol
```diff
@@ -728,19 +735,24 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
...
+        // so the slippage floor is always positive.
+        if (
+            minAmountOut > expectedAmountOut
+                || minAmountOut * BPS_DENOMINATOR
+                    < expectedAmountOut
+                        * (BPS_DENOMINATOR - MAX_SLIPPAGE_TOLERANCE_BPS)
+        ) {
+            revert TychoRouter__InvalidMinAmountOut(
+                minAmountOut, expectedAmountOut
+            );
```
tamaralipows: I wonder if it would be useful to know why exactly the min amount out is invalid instead of making this a general revert for all 3 cases? not a super strong preference though

crates/tycho-execution/contracts/src/TychoRouter.sol
```diff
@@ -754,6 +766,9 @@ contract TychoRouter is AccessControl, Dispatcher, EIP712 {
             );
         }
 
+        // DUPLICATE: exact copy in _singleSwap and
+        // _sequentialSwapChecked. Not extracted due to
+        // stack-too-deep. Edit all three together.
```
tamaralipows: huh?

crates/tycho-execution/contracts/interfaces/IFeeCalculator.sol
```diff
@@ -14,39 +14,24 @@ struct CustomFees {
...
-     *      function arguments.
-     * @param actualAmountOut The actual amount received from the swap
-     * @param expectedAmountOut Caller-supplied quoted amount out
-     * @param amountIn The input amount of the swap
-     * @param tokenIn The swap's input token address
-     * @param tokenOut The swap's output token address
-     * @param clientFeeBps Client fee in fee units (100_000_000 = 100%)
-     * @param client The client address to look up custom router fees
-     *        and slippage share for, and to receive the client fee
+     *      in a single call. The full surplus goes to the router.
```
tamaralipows: Not true if positive slippage is not enabled

crates/tycho-execution/tests/protocol_integration_tests.rs
```diff
@@ -802,7 +803,8 @@ fn test_single_encoding_strategy_curve_st_eth() {
         token_out,
         BigUint::from_str("1_000000000000000000").unwrap(),
         BigUint::from_str("1").unwrap(),
-        0.02,
+        // Dummy 1-wei quote: zero slippage keeps minAmountOut = 1 (accept any output)
+        0.0,
```
tamaralipows, line 807: Hmm this is not a 1-wei quote anymore, will this pass? Can this be 0? Shouldn't we have a guardrail against this

tamaralipows (commented):
Thank you! Looks a lot cleaner now! Just some comments

crates/tycho-execution/contracts/src/TychoRouter.sol
Troshchk: This is changed -- extracted in a separate function
The initial suggestion was to put a comment around the duplicated blocks

crates/tycho-execution/tests/protocol_integration_tests.rs
Troshchk: This 0.0 is slippage, the quote is still 1 wei(the line above `BigUint::from_str("1").unwrap()`)

But you are right -- we did not have guardrail against amountOut being 0 in the encoder -- added it

crates/tycho-execution/contracts/src/TychoRouter.sol
Troshchk: The error already contains `minAmountOut, expectedAmountOut`, so I think it is clear from that values why the revert happened, what do you think?

crates/tycho-execution/contracts/src/FeeCalculator.sol
```diff
@@ -226,66 +201,18 @@ contract FeeCalculator is AccessControl, IFeeCalculator {
     }
 
     /**
-     * @dev Calculates positive slippage surplus distribution
-     * @return Empty array if disabled or no surplus; otherwise 2-element array:
-     *         [0] = router, [1] = client. _mergeFeeRecipients relies on this ordering.
+     * @dev Calculates the positive slippage surplus, all of which goes to the router
+     * @return positiveSlippage The surplus (zero if disabled or no surplus)
      */
     function _calculatePositiveSlippage(
```
Troshchk, line 207: This method is shallow now, remove?

crates/tycho-execution/model/Cargo.lock
louise-poole, line 1: Was this intentional?

protocols/testing/Cargo.lock
louise-poole, line 1: Same here?

louise-poole (approved):

crates/tycho-execution/model/Cargo.lock
Troshchk: Yes, the `Cargo.lock` should only be in the root and `crates/tycho-execution/model` and `protocols/testing` are members of the workspace

protocols/testing/Cargo.lock
Troshchk: answered [here](https://github.com/propeller-heads/tycho/pull/1213#discussion_r3646609716)

PR #1215, opened 2026-07-17
fix(simulation): overwrite existing accounts on snapshot re-apply

zach030:
## What

`PreCachedDB::update` applied `ChangeType::Creation` with init-if-absent semantics (`AccountStorage::init_account`), silently discarding any re-fetched snapshot for a contract already present in the DB. This PR applies snapshots with `overwrite_account` instead — the semantics proxy-token accounts already use — and documents the `ChangeType` contract on `update()`.

## Why

Long-lived consumers re-fetch full snapshots for contracts the process-global `SHARED_TYCHO_DB` already knows: tycho-client fetches a healing snapshot after a non-contiguous reconnect, consumers rebuild streams in-process, and components re-enter the tracked set. With init-if-absent all of those snapshots were dropped, so every state change from the outage window was lost. Later deltas healed the account slot by slot at each slot's on-chain write frequency, leaving mixed-epoch storage (e.g. fresh Curve balances/D with a frozen price_scale or balance slot). CryptoSwap math converts that inconsistency into a constant additive quote error: in production this produced quotes like 0.14 WBTC (~$9,000) out for a $1.6 input on tricrypto2, flagged by hindsight on 2026-07-14.

Root-cause analysis: https://propeller-heads.atlassian.net/wiki/spaces/~71202094d3d73818b5447b85231f4bd027b78c/pages/3833790465/Root-Cause+Analysis+Curve+Pools+Over+Under-Quoting+in+Hindsight

## Testing

- `test_creation_reapply_overwrites_existing_account`: red on the old semantics (slot kept its pre-outage value after a snapshot re-apply), green with this fix.
- `test_delta_update_refreshes_existing_slot`: guards the delta path, green before and after.
- End-to-end reproduction (branch `codex/debug-curve-tricrypto2-overquote`, `--curve-rebuild-repro`): tears down and rebuilds a vm:curve stream in one process with one tampered tricrypto2 storage slot in between. Before this fix the re-fetched snapshot is discarded and a dust-input quote returns 13,988,869 raw WBTC (a ~$9,000 phantom constant); with it the snapshot overwrites the tampered slot and the same probe returns 2 raw WBTC.

Affects every VM/hybrid protocol and every long-lived consumer, not only Curve or hindsight.

### DIFF
--- a/crates/tycho-simulation/src/evm/account_storage.rs
+++ b/crates/tycho-simulation/src/evm/account_storage.rs
@@ -53,8 +53,12 @@ impl AccountStorage {
     /// # Notes
     ///
     /// This function checks if the `address` is already present in the `accounts`
-    /// collection. If so, it logs a warning and returns without modifying the instance.
+    /// collection. If so, it logs a trace message and returns without modifying the instance.
     /// Otherwise, it stores a new `Account` instance with the provided data at the given address.
+    ///
+    /// Init-if-absent is only safe for seeding placeholder accounts during engine setup;
+    /// authoritative data that can arrive again for a known account (e.g. a re-fetched contract
+    /// snapshot) must use [`overwrite_account`](Self::overwrite_account) instead.
     pub fn init_account(
         &mut self,
         address: Address,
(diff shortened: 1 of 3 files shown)


louise-poole (approved):
Nice!!! Looks good to me! 👍

PR #1217, opened 2026-07-17
feat(rfq): migrate Metric integration to v1 API

zach030:
## Summary

Replaces the legacy Metric RFQ integration with the v1 API at `https://api.metric.xyz`. Integration mode stays RFQ. The legacy integration is removed rather than kept in parallel.

Endpoints move to `/public/v1/evm/{chain_id}/...` with numeric chain ids (Ethereum, Base), Bearer auth on `bid_ask`, and paginated metadata. TVL is read from the metadata `tvlFiat` field, replacing the previous one-hop quote-token normalization.

## Oracle model

Under v1 the on-chain oracle is updated by a per-block heartbeat, so no signed oracle update is relayed with the swap. The signed oracle-update path is removed end to end (no `get_signed_data` fetch, no `oracle_update_0_args`, no `oracle_update_policy` attribute, no `MetricOracleUpdatePolicy`). The swap encoder always emits the `Never` mode byte, so the deployed `MetricExecutor` still validates and swaps unchanged — its `updateBySignature` path is now unused by Tycho.

Tradeoff: a swap can revert if the block's oracle update has not landed yet; there is no `RetryOnRevert` fallback.

## Changes

- **constants**: default URL to v1; `METRIC_SECRET_KEY` → `METRIC_API_KEY` (Bearer).
- **models**: add `PaginatedMetadataResponse`, `tvlFiat`, `serverTs`, nullable totals, `is_quotable()`; remove signed-oracle models and update policy.
- **client**: v1 endpoints, metadata pagination, Bearer auth, `chain_to_chain_id` (ETH=1, Base=8453); TVL from `tvlFiat`; skip pools with null availability.
- **state / decoder**: `server_ts` rename; `request_signed_quote` returns empty quote attributes.
- **execution encoder**: always `Never` mode; drop oracle-args encoding.
- Drop `.token_metadata(...)` at the quickstart and integration-test call sites.

`MetricExecutor.sol` and the shared `rfq/protocols/utils.rs` are unchanged.

## Test plan

- [x] `cargo check --workspace --all-features --all-targets`
- [x] `cargo test -p tycho-simulation -p tycho-execution` (metric) — passing; 2 live tests `#[ignore]`d
- [x] `cargo clippy` (both crates, all-targets/all-features)
- [x] `cargo +nightly fmt --check`
- [ ] Run the `#[ignore]` live tests with `METRIC_API_KEY` against `https://api.metric.xyz` to confirm real `bid_ask`/metadata shapes

## Open items before merge

- Confirm `bid_ask` availability semantics against the live API (v1 has no `quoteAvailable`; availability is inferred from non-null `totalToken{0,1}Available`).
- Confirm Base's permissionless oracle is updated frequently enough for heartbeat reliance to hold.

### DIFF
--- a/crates/tycho-execution/src/encoding/evm/swap_encoder/metric.rs
+++ b/crates/tycho-execution/src/encoding/evm/swap_encoder/metric.rs
@@ -1,37 +1,22 @@
 use std::collections::HashMap;
 
-use tokio::runtime::Handle;
-use tycho_common::{
-    models::{protocol::GetAmountOutParams, Chain},
-    Bytes,
-};
+use tycho_common::{models::Chain, Bytes};
 
 use crate::encoding::{
     errors::EncodingError,
-    evm::utils::{bytes_to_address, create_encoding_runtime, on_blocking_thread, SafeRuntime},
+    evm::utils::bytes_to_address,
     models::{EncodingContext, Swap},
     swap_encoder::SwapEncoder,
 };
 
-const ORACLE_UPDATE_POLICY_ATTR: &str = "oracle_update_policy";
-const ORACLE_UPDATE_ARGS_ATTR: &str = "oracle_update_0_args";
+// Oracle update mode byte consumed by MetricExecutor. Under the v1 heartbeat model Tycho never
+// relays an oracle update, so the encoder always emits `Never`. Kept as a named constant to stay
+// lined up with MetricExecutor's OracleUpdateMode enum.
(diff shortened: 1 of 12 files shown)


zach030:
Superseded by #1220 — same change rebased onto 0.339.0, with terminology corrected: the new api.metric.xyz API is v1 (its endpoints live under /public/v1/), so branch, title, and code comments now say v1 instead of the internal "v2" shorthand. Closed automatically when the head branch was renamed to feat/metric-v1-rfq-integration.

PR #1218, opened 2026-07-17
fix(tycho): stop producing code-less Creation account updates

pistomat:
## Summary

Two related fixes for `Bad account update: MissingCode - Account …` crashes in downstream simulation consumers (fynd solvers). During the Jul 13–15 tycho-beta websocket disconnect storms, every fynd deployment on the beta fleet (marketprice.xyz, staging-fynd ethereum/arbitrum/base/bsc/unichain, prod-fynd hindsight) crash-looped on this error after each reconnect; the crashing deltas were token accounts (WBTC `0x2260…`, stETH `0xae7a…`).

## Fix 1 — decoder: proxy-token updates decoded as code-less `Creation` (the production crash)

`tycho-simulation/src/evm/decoder.rs`: when a token account that already has a proxy contract receives a delta with `change=Creation` (routine after a websocket resubscribe), the decoder built the proxy storage write as `AccountUpdate { code: None, ..update }` — inheriting `Creation`. `PreCachedDB::update` rejects a code-less `Creation` (`MissingCode`), the decoder escalates to `StreamDecodeError::Fatal`, and the consumer's stream dies. The proxy account is guaranteed to exist in this branch, so the write is now sent as `change: Update`.

Note: downgrading to `Update` is one way to fix it — happy to discuss alternatives (e.g. attaching the proxy bytecode instead). The bug itself is confirmed: the new test reproduces the exact production error without the fix.

Evidence this is the prod path: crashes only fired on post-storm reconnects, the affected addresses are tokens, and fleet logs contain zero `FaultyCreationDelta` (ENG-4993 mitigation) and zero `AccountDelta without code marked as creation after merge!` warnings — ruling out the indexer-emission and model-merge producers.

## Fix 2 — dto: `AccountUpdate::merge` erased `code`/`balance` (latent, public API)

`tycho-common/src/dto.rs`: `merge` unconditionally overwrote `balance` and `code` with the newer update's values, so merging a `Creation { code: Some(…) }` with a later storage-only delta (`code: None`) produced a code-less `Creation`. `None` means "field untouched this block", not "clear it". Now absent fields keep their current values, matching `models::contract::AccountDelta::merge`. No in-repo production caller hits this today (the client pipeline merges at the model layer), but it's a public DTO API and a footgun for external consumers.

## Tests

- `test_decode_token_creation_delta_with_existing_proxy` (RED→GREEN): decodes a token `Creation` delta twice; unfixed code fails the second decode with the exact production error string.
- `test_account_update_merge_keeps_code_and_balance_when_other_carries_none` (RED→GREEN): merge keeps known `code`/`balance` when the newer update carries `None`.
- `cargo test -p tycho-common --lib dto::` 38 passed; `cargo test -p tycho-simulation --lib test_decode` 7 passed; fmt + clippy clean.


zizou0x (approved):
LGTM

PR #1219, opened 2026-07-17
feat(sim): lazy spot prices + limit caching for VM get_amount_out

tvinagre:
## Summary

Speeds up VM-protocol simulation (Balancer V2, Curve, …) by removing redundant work from the
`EVMPoolState` hot path, with byte-identical results.

`get_amount_out` previously recomputed spot prices for every token-pair permutation on the state it
returns, and recomputed the block-stable swap limit on every call. This change adds two thread-safe
read-through caches on `EVMPoolState`:

- `spot_price` computes a pair on demand and caches it, instead of eagerly recomputing all
  permutations inside `get_amount_out`.
- `get_amount_limits` caches the `(sell, buy)` limit and reuses it across calls.
- `get_amount_out` invalidates the returned `new_state`'s caches (normal pools) instead of eagerly
  recomputing; `update_pool_state` clears then re-warms per block; `delta_transition` clears the
  limit cache when the block environment changes.

Pools with live overrides (Titan pAMM), whose spot prices and limits are override-derived and change
sub-block, bypass the caches and keep the existing eager, uncached behaviour.

Also included: the criterion benchmark + concurrency harness used to baseline and verify the change
(committed fixtures, `get_amount_out`/`spot_price`/`concurrency_throughput` benches, an offline
fixture-replay smoke test, and a multi-threaded `spot_price`/`get_amount_out` equivalence test).

## Measured results (release, `cargo bench --baseline before`)

| metric | change |
|---|---|
| `get_amount_out` — balancer 2-token | −57% |
| `get_amount_out` — curve 3-token / 4-token | −79% |
| Fynd `find_best_route` (2 / 3 / 4-token) | −55% / −42% / −36% |
| contended throughput (1 / 2 / 4 / 8 threads) | +29% / +34% / +47% / +52% |
| `spot_price` | now a ~51 ns cache read |

Multi-token pools gain the most: they had the most permutations being eagerly recomputed. All
exact-value unit tests are unchanged (amounts, limits, spot prices byte-identical).

## Test plan

- [ ] `cargo nextest run -p tycho-simulation --lib` — 954 passed, 0 failed (16 network-gated skips)
- [ ] `cargo nextest run -p tycho-simulation -E 'test(/evm::protocol::vm::state/)'` — 25 passed
- [ ] `cargo nextest run -p tycho-simulation --test fixture_replay --test vm_concurrency` — 5 passed
- [ ] `cargo +nightly fmt -p tycho-simulation -- --check` and `cargo clippy -p tycho-simulation --all-targets` — clean

## Notes for reviewers

- Fixtures (`benches/fixtures/*.json`) total ~11 MB; happy to move them to LFS or split the harness
  into its own PR if preferred. They are `exclude`d from `cargo package` (along with the
  pre-existing `tests/assets`), so the published crate shrinks from 7.4MiB to 2.3MiB compressed.
- Limit freshness: cached limits are stamped with the engine block and block-env overrides they
  were computed under; a stamp mismatch is a miss, so limits self-correct each block even for
  pools receiving no deltas. The only residual staleness is same-block storage mutations that
  emit no pool-level event (e.g. token-proxy updates), matching the pre-existing spot-price
  staleness contract.
- The `vm::state` tests share a global `SHARED_TYCHO_DB` and can flake under single-process
  `cargo test`; CI's `cargo nextest` (process-per-test) is unaffected. A separate pre-existing flake
  (`test_failing_overrides_error_by_default`) is unrelated to this change.


crates/tycho-simulation/src/evm/protocol/vm/state.rs
```diff
@@ -507,6 +589,20 @@ where
...
+        if self.live_overrides.is_none() {
+            if let Some(limits) = self
+                .limit_cache
+                .read()
+                .expect("limit_cache poisoned")
+                .get(&key)
+                .copied()
+            {
+                return Ok(limits);
+            }
```
brunoguerios: The cache only clears on pool-level events, so a pool that receives no deltas serves a stale limit indefinitely.
This covers time-dependent adapter math: `update_engine` advances the block env every block, so limits that read `block.timestamp` (e.g. a Curve pool mid A-ramp) change with zero deltas.
A stale-high sell limit becomes an on-chain revert.
An idea would be to stamp each entry with the block number and treat a mismatch as a miss.
That bounds staleness to one block: the first read per pair per block pays the pre-PR sim cost, repeats still hit the cache.

crates/tycho-simulation/src/evm/protocol/vm/state.rs
```diff
@@ -936,12 +1064,40 @@ where
     fn spot_price(&self, base: &Token, quote: &Token) -> Result<f64, SimulationError> {
         let base_address = bytes_to_address(&base.address)?;
         let quote_address = bytes_to_address(&quote.address)?;
-        self.spot_prices
+
+        // Fast path: cache hit (all pools).
+        if let Some(price) = self
+            .spot_price_cache
+            .read()
+            .expect("spot_price_cache poisoned")
```
brunoguerios: These `expect("... poisoned")` calls turn one poisoned guard into a panic on every later read. These states are shared as `Arc<Box<dyn ProtocolSim>>` across threads, so a single poisoning cascades into every worker that touches the pool. The critical sections only do `HashMap` get/insert/clear — a poisoned lock cannot hold a torn value — so recovering is safe:
```suggestion
            .unwrap_or_else(std::sync::PoisonError::into_inner)
```
Same pattern for the other sites; a small `fn read_cache` / `fn write_cache` helper would avoid repeating it.

Even though poisoning here is near-unreachable, the change costs nothing and removes the cascade if that ever stops being true.

brunoguerios (approved):
I left a few small comments, but otherwise LGTM ✅

crates/tycho-simulation/src/evm/protocol/vm/state.rs
```diff
@@ -57,8 +57,16 @@ where
...
+    /// Spot prices by `(sell, buy)`, cleared whenever pool state changes. For regular pools this
+    /// is a read-through cache: lazily populated by `spot_price`, eagerly warmed by
+    /// `update_pool_state`. Pools with live overrides only ever populate it eagerly (their
+    /// override-derived prices change sub-block, so `spot_price` errors on a miss instead of
+    /// computing against a possibly newer snapshot).
+    spot_price_cache: RwLock<HashMap<(Address, Address), f64>>,
+    /// Read-through cache of `(sell_limit, buy_limit)` by `(sell, buy)`. Stable per
+    /// pool-state-version; cleared whenever pool state changes. Fully bypassed (no reads or
+    /// writes) for pools with live overrides.
+    limit_cache: RwLock<HashMap<(Address, Address), (U256, U256)>>,
```
carloszanella: why does this use "Address" while tokens is "Bytes"?

Also can you use fxhashmap if it is available in the codebase?

crates/tycho-simulation/src/evm/protocol/vm/state.rs
```diff
@@ -325,6 +387,132 @@ where
         )
     }
 
+    /// Computes the spot price for a single `(sell, buy)` pair against `live_snapshot`'s overrides
+    /// (or the plain indexed state when `None`), using the same logic as `set_spot_prices_with`:
```
carloszanella: if it uses the same logic, cant we share some of the logic with private methods?

carloszanella: ok looks like this is just AI slop comment, and the code was moved from that function into here

crates/tycho-simulation/src/evm/protocol/vm/state.rs
```diff
@@ -541,6 +644,14 @@ where
...
             })?;
         self.block_lasting_overwrites.clear();
+        self.spot_price_cache
+            .write()
+            .expect("spot_price_cache poisoned")
+            .clear();
+        self.limit_cache
+            .write()
+            .expect("limit_cache poisoned")
+            .clear();
```
carloszanella: this is very repetitive. I would write some small helpers like `write/read_price_cache()` and same for limits so this .write().expect() collapses to one line

carloszanella (approved):
only reviewed state.rs. looks good, just minor comments

crates/tycho-simulation/src/evm/protocol/vm/state.rs
carloszanella: yeah I think that would be cleaner in general

---

**[DESIGN `#1`]**

**Root cause**: `spot_price_cache` and `limit_cache` are separate `RwLock<HashMap>` fields whose "when is this stale" rule is written out again at every mutation point (`set_live_overrides`, `update_pool_state`, `get_amount_out_with`, `delta_transition`), and deliberately diverges in one of them (`delta_transition` clears only the limit cache). Nothing typed enforces "both caches reflect the same pool version", and any future field that invalidates derived state must be found and patched in all five places.

- **Architecture** *(was DESIGN `#1`)*: invalidation policy duplicated across five call sites with one intentional divergence a reader cannot distinguish from an omission. `crates/tycho-simulation/src/evm/protocol/vm/state.rs:647–654, 942–963, 1179–1183`
- **Complexity** *(was SIMPLIFY `#1`, PERSONAL `#7`)*: every access spells out `.read()`/`.write()` plus `.expect("<name> poisoned")` — 14 production sites — burying the logic in lock ceremony and giving 14 chances to get the message wrong. Maps to the `dry` directive. `crates/tycho-simulation/src/evm/protocol/vm/state.rs:219–228, 539–544, 593–618, 645–654, 952–963, 1067–1099, 1177–1182`
- **Complexity** *(was SIMPLIFY `#1`)*: because `RwLock` is not `Clone`, `#[derive(Clone)]` was replaced by a 38-line hand-written impl that re-lists all 17 fields. `crates/tycho-simulation/src/evm/protocol/vm/state.rs:109–146`

**Proposed fix**: Collapse both maps into one `DerivedCaches { spot_prices, limits }` behind a single `RwLock`, exposing `invalidate()` / `invalidate_limits()` methods that carry the staleness rationale once. The single named type gives the policy one home (first symptom), moves the poison `expect` from 14 sites to one (second), and its own manual `Clone` lets `EVMPoolState` go back to `#[derive(Clone)]` (third).

crates/tycho-simulation/src/evm/protocol/vm/state.rs
```diff
@@ -936,12 +1064,40 @@ where
...
+        ]);
+        let block_overrides = self.block_env(None);
+        let price =
+            self.compute_spot_price(&tokens, base_address, quote_address, None, block_overrides)?;
+        self.spot_price_cache
+            .write()
+            .expect("spot_price_cache poisoned")
+            .insert((base_address, quote_address), price);
+        Ok(price)
     }
```
carloszanella, line 1145: probably should add this since it was there before

---

**[DESIGN `#2`]**

**Description**: For non-override pools a cache miss now unconditionally calls `compute_spot_price`, with no check that `base`/`quote` are actually in `self.tokens`. Previously a bogus pair returned instantly with "Spot price not found"; now it runs one or two EVM swaps against the adapter before failing with a deeper, less actionable error — and any router probing pairs speculatively pays that cost per bad pair.

**Suggestion**: Guard the lazy path with a membership check against `self.tokens` and keep returning the existing "Spot price not found" `FatalError` for pairs the pool does not hold.

crates/tycho-simulation/src/evm/protocol/vm/state.rs
```diff
@@ -325,6 +387,132 @@ where
...
+    /// (or the plain indexed state when `None`), using the same logic as `set_spot_prices_with`:
+    /// the adapter `price` function when `PriceFunction` is supported, otherwise a two-swap
+    /// finite-difference. Does not touch the cache.
+    fn compute_spot_price(
+        &self,
+        tokens: &HashMap<Bytes, Token>,
+        sell_token_address: Address,
+        buy_token_address: Address,
+        live_snapshot: Option<&OverrideSnapshot>,
+        block_overrides: Option<BlockEnvOverrides>,
```
carloszanella: **[DESIGN `#3`]**

**Description**: `live_snapshot` and `block_overrides` always travel together and the second is always `self.block_env(live_snapshot)` at both call sites (`state.rs:533–539`, `state.rs:1093–1095`); the pair can be passed inconsistently with nothing to catch it. Six parameters for a function whose body is a single capability branch is more surface than the work needs.

**Suggestion**: Drop `block_overrides` and compute `self.block_env(live_snapshot)` at the top of `compute_spot_price`. That removes a positional parameter, two hoisted locals and the per-iteration `.clone()`, at the cost of calling the cheap `block_env` once per permutation in `set_spot_prices_with`'s loop instead of once per loop.

crates/tycho-simulation/Cargo.toml
```diff
@@ -131,3 +136,14 @@ rfq = [
...
+[[bench]]
+name = "get_amount_out"
+harness = false
+
+[[bench]]
+name = "concurrency_throughput"
+harness = false
+
+[[example]]
+name = "capture_vm_fixtures"
```
carloszanella: **[DESIGN `#10`]**

`examples/capture_vm_fixtures.rs` is auto-discovered and none of the five existing examples declare a block; drop the stanza. Maps to the `yagni` directive.

crates/tycho-simulation/src/evm/protocol/vm/state.rs
```diff
@@ -333,137 +521,27 @@ where
         live_snapshot: Option<&OverrideSnapshot>,
     ) -> Result<(), SimulationError> {
```
carloszanella, line 587: **[DESIGN `#11`]**

It now writes through `spot_price_cache`'s lock like `spot_price` does, so `&self` would be accurate and would make the two writers of the same cache consistent in ownership.

crates/tycho-simulation/src/evm/protocol/vm/state.rs
```diff
@@ -515,6 +611,13 @@ where
...
         )?;
 
+        if self.live_overrides.is_none() {
+            self.limit_cache
+                .write()
+                .expect("limit_cache poisoned")
+                .insert(key, limits);
+        }
+
         Ok(limits)
```
carloszanella, line 684: This is interesting although it depends on this being an actual possibility:
"For pools whose maximum tradable size depends on time rather than storage (amplification ramps, gradual weight shifts, maturity windows), the reported limit drifts from reality and never self-corrects."

---

**[IMPL `#1`]**

**Category**: bug

**Root cause**: `get_amount_limits` caches on `(tokens[0], tokens[1])` while the adapter's answer is fully determined by the `overwrites` and `block_overrides` it also accepts, plus the shared `PreCachedDB`'s current block. Invalidation is therefore event-driven (a delta, a swap, `update_pool_state`, `set_live_overrides`, a block-override attribute) rather than derived from what the value depends on.

**Plain English**: A pool whose on-chain storage did not change in a block, or one on a protocol that only refreshes on an explicit trigger, keeps reporting the trade-size limit it computed when it was last touched — possibly many blocks earlier. For pools whose maximum tradable size depends on time rather than storage (amplification ramps, gradual weight shifts, maturity windows), the reported limit drifts from reality and never self-corrects. An integrator sizing a trade against it either submits an oversized swap that reverts on chain or silently leaves liquidity unused. Separately, the routine that answers "what is the largest trade this pool can take" accepts a description of the simulated world but, after the first answer, returns that first answer no matter what world the caller asks about. The limit should be recomputed whenever the block or the simulated conditions it was computed under have moved on.

- **Behavior** *(was IMPL `#1`)*: the adapter's block env comes from `self.state.get_current_block()`, which advances every block for every pool, but a pool receiving no delta (and a `manual_updates` pool receiving one without an `update_marker`) never clears `limit_cache` — pre-PR every `get_limits`/`get_amount_out` re-simulated `getLimits` at the current block. `crates/tycho-simulation/src/evm/protocol/vm/state.rs:1171–1183`
- **Architecture** *(was DESIGN `#2`)*: the signature still takes `overwrites` and `block_overrides` and the doc still promises limits "for the given overwrites", but both are silently discarded on a hit. Correctness rests on an unwritten convention that every call site passes `get_overwrites(..., MAX_BALANCE/100, snapshot)` and `block_env(snapshot)` — it works by coincidence of today's three callers, not by construction. Maps to the `correctness` directive. `crates/tycho-simulation/src/evm/protocol/vm/state.rs:586–592`

**Proposed fix**: Make the cache key cover what the result depends on — fold the resolved block number/timestamp and a digest of the overwrites into the key (or store them alongside the entry and treat a mismatch as a miss). That fixes the block-staleness symptom because a new block no longer hits, and the signature symptom because the inputs and the key become the same thing. If the parameters are genuinely never meant to vary, the alternative single fix is to remove them from the signature and derive them inside `get_amount_limits`.

crates/tycho-simulation/src/evm/protocol/vm/state.rs
```diff
@@ -98,6 +106,45 @@ where
...
+            contract_balances: self.contract_balances.clone(),
+            manual_updates: self.manual_updates,
+            adapter_contract: self.adapter_contract.clone(),
+            disable_overwrite_tokens: self.disable_overwrite_tokens.clone(),
+            self_contained_tokens: self.self_contained_tokens.clone(),
+            block_overrides: self.block_overrides.clone(),
+            live_overrides: self.live_overrides.clone(),
+            spot_price_caller: self.spot_price_caller,
+        }
+    }
```
carloszanella: yeah I think we dont need to clone the caches too when creating new state, and we should just start with fresh ones

---

**[IMPL `#3`]**

**Category**: perf

**Root cause**: `Clone for EVMPoolState` copies both cache `HashMap`s under their locks, and `get_amount_out_with`'s non-override branch immediately clears both on the clone — so every `get_amount_out` allocates and drops n·(n−1) spot-price entries plus the limit entries, on exactly the path this PR exists to optimise. The manual `Clone` is also the only reason `#[derive(Clone)]` had to go.

- **Behavior** *(was IMPL `#4`)*: two lock acquisitions and two map deep-copies per `get_amount_out`, thrown away on the next statement. `crates/tycho-simulation/src/evm/protocol/vm/state.rs:951–964`
- **Architecture** *(was DESIGN `#6`, PERSONAL `#6`)*: a clone is semantically a fresh pool version, so copying derived caches into it is the wrong default; the eager re-warm is only needed on the override branch. Maps to the `simplify` directive. `crates/tycho-simulation/src/evm/protocol/vm/state.rs:115–145`

**Proposed fix**: Add a `clone_invalidated()` (or have `Clone` start the new state with empty caches and keep a full copy only for `clone_box`), and build `new_state` from it. That removes the copy on the hot path and removes the clear, because there is nothing left to clear.

carloszanella (commented):
Review comments below.

---

**[DOCS `#1`]**

**Description**: The doc still points at "the `spot_prices` map", a field the diff renamed to `spot_price_cache`, and still states "Spot prices need to be set before attempting to retrieve prices using `spot_price`" — now false for non-override pools, where `spot_price` computes on miss. A reader following this doc will believe eager warming is mandatory and will not learn that it is only required for live-override pools.

**Suggestion**: Update the field reference to `spot_price_cache` and rewrite the `# Usage` paragraph to say eager warming is required only for pools with live overrides, and is a latency optimisation elsewhere.

crates/tycho-simulation/src/evm/protocol/vm/state.rs
```diff
@@ -265,6 +323,10 @@ where
     ///
     /// * `Result<(), SimulationError>` - Returns `Ok(())` if the capability is supported, or a
     ///   `SimulationError` otherwise.
+    ///
+    /// Only used by tests now that `set_spot_prices_with` inlines the capability check via
+    /// `compute_spot_price`.
+    #[cfg(test)]
     fn ensure_capability(&self, capability: Capability) -> Result<(), SimulationError> {
         if !self.capabilities.contains(&capability) {
             return Err(SimulationError::FatalError(format!(
```
carloszanella: kill

---

**[SIMPLIFY `#1`]**

**Root cause**: `compute_spot_price` inlined the capability check, leaving `ensure_capability` with no production caller; instead of being deleted it was marked `#[cfg(test)]` and given a doc comment explaining that it exists only for tests. Its two remaining uses sit directly after `assert_eq!(capabilities_state, expected_capabilities)`, so the test now exercises a helper that exists only for that test.

- **Complexity** *(was SIMPLIFY `#3`, DESIGN `#13`)*: 23 lines of function plus doc comment retained in `src/` for no production purpose. `crates/tycho-simulation/src/evm/protocol/vm/state.rs:316–338`
- **Behavior** *(was TESTS `#6`, PERSONAL `#3`)*: the two assertion blocks re-check `HashSet::contains` against the same set the preceding `assert_eq!` already compares in full, so they test nothing the stronger assertion does not. Maps to the `yagni` directive. `crates/tycho-simulation/src/evm/protocol/vm/state.rs:1425–1436`
- **Documentation** *(was DOCS `#3`)*: the appended "Only used by tests now that `set_spot_prices_with` inlines the capability check via `compute_spot_price`" narrates the refactor, dangles below `# Returns`, and names the wrong symbol — the check lives in `compute_spot_price`, not `set_spot_prices_with`. `crates/tycho-simulation/src/evm/protocol/vm/state.rs:327–329`

**Proposed fix**: Delete `ensure_capability` and the two assertion blocks in `test_init` (~34 lines). The set-equality assertion above them is strictly stronger, so no coverage is lost; the misleading doc comment goes with the function.

crates/tycho-simulation/src/evm/protocol/vm/state.rs
```diff
@@ -1301,7 +1475,21 @@ mod tests {
...
         assert_eq!(result.amount, BigUint::from_str("137780051463393923").unwrap());
-        assert_ne!(new_state.spot_prices, pool_state.spot_prices);
+        // new_state has an invalidated (empty) spot-price cache; reading recomputes the
+        // post-swap price lazily, which differs from the pre-swap price because the swap moved
+        // the pool.
+        assert!(new_state
+            .spot_price_cache
+            .read()
+            .unwrap()
+            .is_empty());
```
carloszanella: **[SIMPLIFY `#4`]**

`assert!(new_state.spot_price_cache.is_empty())` sits immediately above the public pre/post-swap `spot_price` comparison that already covers the behavior; drop the map inspection.

crates/tycho-simulation/src/evm/protocol/vm/state.rs
```diff
@@ -1459,22 +1702,52 @@ mod tests {
...
+        assert_eq!(
+            pool_state
+                .spot_price_cache
+                .read()
+                .unwrap()
+                .get(&(dai_addr(), bal_addr()))
+                .copied(),
+            Some(price)
+        );
     }
```
carloszanella, line 1802: **[TESTS `#2`]**

**Description**: `test_spot_price_lazy_computes_on_miss` only asserts `price > 0.0` and that the value got cached. The PR's central claim is that the new lazy `compute_spot_price` path is byte-identical to the old eager `set_spot_prices` path, and nothing asserts that — a sign error or a wrong decimal correction would still pass. The exact value is already known: this is the same DAI/BAL pair `test_set_spot_prices` pins. Maps to the `ai-slop` and `test-quality` directives.

**Suggestion**: Assert the lazily computed DAI→BAL price equals the constant already pinned in `test_set_spot_prices` (`0.137_778_914_319_047_9`), and add the reverse pair against `7.071_503_245_428_246`.

crates/tycho-simulation/tests/fixture_replay.rs
```diff
@@ -0,0 +1,44 @@
...
+        });
+
+    assert!(
+        result.amount > BigUint::from(0u64),
+        "Expected a positive amount_out for {fixture}: {} -> {}, got {}",
+        t_in.symbol,
+        t_out.symbol,
+        result.amount
+    );
+}
```
carloszanella: **[TESTS `#5`]**

**Description**: `decode_and_simulate` asserts only `result.amount > 0`, so the committed fixtures currently prove little beyond "the JSON deserializes and the adapter does not revert". The fixtures pin an exact block state, so the outputs are deterministic and could pin exact amounts — which is precisely the byte-identical-results guard this PR needs and does not have. Maps to the `ai-slop` and `test-quality` directives.

**Suggestion**: Pin the expected `amount` (and `gas`) per fixture case as constants and assert equality, and add a `spot_price` assertion per fixture so the replay covers the cached path too.

crates/tycho-simulation/Cargo.toml
```diff
@@ -106,6 +110,7 @@ crossterm = { version = "0.28.1", features = ["event-stream"] }
 unicode-width = "0.1.13"
 tracing-appender = { workspace = true }
 dialoguer = "0.10.4"
+criterion = { version = "0.8.2", features = ["html_reports"] }
```
carloszanella: **[PERSONAL `#3`]**

Enabling `html_reports` adds plotters, plotters-svg and rayon to the dev-dep tree; drop the feature unless someone is actually reading the generated HTML. Maps to the `yagni` directive.

crates/tycho-simulation/src/evm/protocol/vm/state.rs
```diff
@@ -325,6 +387,132 @@ where
         )
     }
 
+    /// Computes the spot price for a single `(sell, buy)` pair against `live_snapshot`'s overrides
+    /// (or the plain indexed state when `None`), using the same logic as `set_spot_prices_with`:
+    /// the adapter `price` function when `PriceFunction` is supported, otherwise a two-swap
+    /// finite-difference. Does not touch the cache.
+    fn compute_spot_price(
```
carloszanella, line 459: yeah exactly what caused confusion to me

---

**[DOCS `#2`]**

**Description**: The doc says the function uses "the same logic as `set_spot_prices_with`", but the dependency now runs the other way: `set_spot_prices_with` loops over token permutations and delegates each pair to `compute_spot_price`. Describing a callee in terms of its caller inverts the relationship and rots the moment either side moves.

**Suggestion**: State the algorithm directly — adapter `price` when `PriceFunction` is supported, otherwise a two-swap finite difference — and drop the cross-reference.

crates/tycho-simulation/src/evm/protocol/vm/state.rs
```diff
@@ -828,12 +939,29 @@ where
...
 
-        // Update spot prices
-        let tokens = HashMap::from([
-            (token_in.address.clone(), token_in.clone()),
-            (token_out.address.clone(), token_out.clone()),
-        ]);
-        let _ = new_state.set_spot_prices(&tokens);
+        if new_state.live_overrides.is_some() {
+            // Override pools do not cache spot prices lazily (they are override-derived and
+            // change sub-block), so keep eagerly warming the returned state exactly as before.
```
carloszanella: **[DOCS `#3`]**

**Description**: Three added comments explain themselves relative to the pre-PR code: "keep eagerly warming the returned state exactly as before", "Preserve today's behavior", and "the old code recomputed limits fresh here, whereas the spot-price map was retained-stale in this path, so leaving `spot_price_cache` untouched preserves byte-identical behavior". Once this lands there is no "before" for a reader to compare against; in six months "the old code" means nothing. This belongs in the PR description. Maps to the `docs` directive.

**Suggestion**: Rewrite each as the invariant that holds now — e.g. at 1171: "Cached limits are computed against the block env, so a block-env change invalidates them; spot prices are deliberately allowed to go stale in this skip path, matching the override-pool contract."

crates/tycho-simulation/src/evm/protocol/vm/state.rs
```diff
@@ -57,8 +57,16 @@ where
     /// simulations. This has been deprecated in favor of `contract_balances`.
     #[deprecated(note = "Use contract_balances instead")]
     balance_owner: Option<Address>,
-    /// Spot prices of the pool by token pair
-    spot_prices: HashMap<(Address, Address), f64>,
+    /// Spot prices by `(sell, buy)`, cleared whenever pool state changes. For regular pools this
+    /// is a read-through cache: lazily populated by `spot_price`, eagerly warmed by
+    /// `update_pool_state`. Pools with live overrides only ever populate it eagerly (their
+    /// override-derived prices change sub-block, so `spot_price` errors on a miss instead of
+    /// computing against a possibly newer snapshot).
```
carloszanella: **[DOCS `#4`]**

**Description**: The field doc describes the key as `(sell, buy)`, but the new lazy path in `spot_price` inserts `(base_address, quote_address)` and the `ProtocolSim` trait documents that pair as base/quote. Two vocabularies for one key invites a reader to think the two writers disagree about direction.

**Suggestion**: Pick one frame in the field doc and note the equivalence explicitly, e.g. "keyed by `(sell, buy)` — equivalently `(base, quote)` as used by `ProtocolSim::spot_price`".

crates/tycho-simulation/src/evm/protocol/vm/state.rs
```diff
@@ -936,12 +1064,40 @@ where
     fn spot_price(&self, base: &Token, quote: &Token) -> Result<f64, SimulationError> {
         let base_address = bytes_to_address(&base.address)?;
         let quote_address = bytes_to_address(&quote.address)?;
-        self.spot_prices
+
+        // Fast path: cache hit (all pools).
```
carloszanella: **[DOCS `#7`]**

`// Fast path: cache hit (all pools).` sits directly above a cache read that says the same thing; delete it and keep the two comments below, which carry real information.

crates/tycho-simulation/src/evm/protocol/vm/state.rs
```diff
@@ -265,6 +323,10 @@ where
     ///
     /// * `Result<(), SimulationError>` - Returns `Ok(())` if the capability is supported, or a
     ///   `SimulationError` otherwise.
+    ///
+    /// Only used by tests now that `set_spot_prices_with` inlines the capability check via
+    /// `compute_spot_price`.
+    #[cfg(test)]
```
carloszanella: looks like this is unused

tvinagre:
Addressed the review — thanks both. Commits: 94c400f2a (cache rework + staleness fix), 21e918df4 (fixture pins), 1d0d01a96 (test hardening).

- **Stale limits for pools receiving no deltas (@brunoguerios, IMPL#1)** — cached limits are now stamped with the context they were computed under: the engine's current block (number, timestamp, partial index) plus the resolved block-env overrides. A stamp mismatch is a miss, so time-dependent limits self-correct each block with zero pool-level events. This also subsumes (and deletes) the special-cased limit-cache clear on `override_block_*` deltas. Guarded by `test_engine_block_advance_invalidates_cached_limits`, which advances the engine block with no pool delta and was verified to fail with the stamp removed. Residual staleness is narrowed to same-block storage mutations that emit no pool-level event.
- **Poison cascade** — all lock access goes through one `read()`/`write()` pair using `PoisonError::into_inner`.
- **DESIGN#1** — both maps collapsed into `DerivedCaches` behind one `RwLock`; the invalidation policy is documented once on the type; `EVMPoolState` is back to `#[derive(Clone)]`.
- **IMPL#3** — `get_amount_out` builds its returned state via `clone_invalidated()` (fresh caches, no map copies); `Clone` keeps the deep copy for `clone_box` users.
- **DESIGN#2** — membership check before any EVM sim; kept the existing "Spot price not found" error text as requested.
- **DESIGN#3** — `compute_spot_price` derives the block env itself; the parameter is gone.
- **DESIGN#11** — `set_spot_prices(&self)`.
- **SIMPLIFY#1/#4** — `ensure_capability` and the redundant assertions deleted.
- **TESTS#2/#5** — exact values pinned in the lazy spot-price test and per fixture (amount, gas, spot price) in `fixture_replay`; `test_get_limits_cached_matches_fresh` now genuinely guards the overwrite-amount independence assumption (fresh compute under different overwrites, byte-equal).
- **PERSONAL#3 / DESIGN#10** — `html_reports` feature and the `[[example]]` stanza dropped.
- **DOCS#2/3/4/7** — comments rewritten as invariants; `(sell, buy)` ≡ `(base, quote)` vocabulary unified.
- **fxhash** — adopted (`rustc-hash`, already in the workspace via `tycho-execution/model`).
- **Why `Address` keys while `tokens` is `Bytes`** — the simulations operate on alloy `Address` (converted once at the boundary via `bytes_to_address`), while `tokens` stays the wire-format `Bytes` the decoder produces; keying the cache by `Address` avoids a conversion on every lookup.

PR #1220, opened 2026-07-18
feat(rfq): migrate Metric integration to v1 API

zach030:
## Summary

Replaces the legacy Metric RFQ integration with the v1 API (`https://api.metric.xyz`). Endpoints move to `/public/v1/evm/{chain_id}/...` with numeric chain ids, Bearer auth on `bid_ask` (`METRIC_API_KEY`), and paginated metadata. TVL comes from the metadata `tvlFiat` field.

## Supported chains

Metric lists six chains. Tycho supports all except HyperEVM.

| Chain | chainId | Tycho chain |
|---|---|---|
| Ethereum | 1 | yes |
| Base | 8453 | yes |
| Robinhood | 4663 | yes |
| Arbitrum | 42161 | yes |
| BSC | 56 | yes |
| HyperEVM | 999 | no |

This PR maps Ethereum, Base and Robinhood. Arbitrum and BSC each need a `chain_to_chain_id` entry and an `executor_deployments.json` entry once an executor is deployed there.

## Execution

Under v1 a per-block heartbeat updates the on-chain oracle, so the swap relays no signed oracle update. The oracle-update path (`oracle_update_policy`, `oracle_update_0_args`, `MetricOracleUpdatePolicy`, `updateBySignature`, retry) is removed end to end. Calldata is 61 bytes:

```
tokenIn (20) | tokenOut (20) | pool (20) | direction (1)
```

`MetricExecutor` is constructorless, validates the payload, and calls the pool directly. A swap reverts if the block's heartbeat update has not landed yet.

The v1 executor is not deployed yet. The v0 `rfq:metric` addresses are removed from `executor_addresses.json` and `contract-addresses.md`; add the v1 addresses after deployment.

## Quoting

`get_amount_out` and `get_limits` walk Metric's `cumulativeVolume` / `cumulativeInputVolume` bins in raw integer units. Full bins use the API's cumulative input delta; partial bins are pro rata with conservative rounding. Aggregate inventory caps the output. Non-monotonic or inconsistent depth is rejected; pools without usable depth or with sentinel prices are not emitted. The top-of-book `f64` path remains only for sides without depth.

## Test plan

- [x] Unit tests: simulation state/decoder/models, encoder, registry, cross-language calldata fixture
- [x] `forge test --match-contract MetricExecutorTest` and `TychoRouterForMetricTest` against a Base fork
- [x] `#[ignore]` live tests with `METRIC_API_KEY` against `api.metric.xyz` and `beta-api.metric.xyz` (same `bid_ask` schema)
- [x] `cargo clippy --workspace --all-targets --all-features`, `cargo +nightly fmt --check`, `forge fmt --check`

## Before rollout

- Metric moves its beta deployment to production around Sept 8; deploy the v1 executor after that.
- Metric recommends executing the Metric leg first in split trades, since other legs move its reference price.


crates/tycho-execution/config/executor_deployments.json
die-herdplatte, line 1: I'm wondering why Metric is configured for Unichain, Polygon, and BSC here.
https://api.metric.xyz/public/v1/chains doesn't list them.

crates/tycho-simulation/src/rfq/protocols/metric/client_builder.rs
```diff
@@ -48,32 +35,18 @@ impl MetricClientBuilder {
         self
     }
```
die-herdplatte, line 40: It's unclear that leaving this empty (or explicitly passing an empty set) results in the [token filter being disabled](https://github.com/propeller-heads/tycho/blob/934bba20836f2e8d747c9fcf74b260e41990b959/crates/tycho-simulation/src/rfq/protocols/metric/client.rs#L284).
I'd add a comment here.

crates/tycho-simulation/src/rfq/protocols/metric/state.rs
```diff
@@ -168,70 +159,62 @@ impl ProtocolSim for MetricState {
...
+            MetricDirection::OneForZero => self.bid_ask.total_token0_available()?,
+        };
+
+        // Prefer size-aware depth when Metric exposes it. The depth walk runs entirely in raw
+        // integer units on Metric's own per-bin accounting, so its result is exact — including
+        // the cap returned when the depth is exhausted.
+        if let Some(quote) = self.quote_with_depth(direction, &amount_in, &max_output)? {
+            let res = GetAmountOutResult {
+                amount: quote.amount_out,
+                gas: BigUint::from(170_000u64),
```
die-herdplatte: This should be a constant

crates/tycho-simulation/src/rfq/protocols/metric/models.rs
die-herdplatte, line 85: The "raw" properties are always parsed twice, once trough the stream calling `is_quotable` and then by the `ProtocolSim` functions.
Parsing them directly into their expected format would avoid this and require less error handling / propagation.

die-herdplatte (commented):
Looks good to me!

A link to the Metric docs somewhere would be nice.

crates/tycho-execution/config/executor_deployments.json
zach030: Removed unconfigured chains

crates/tycho-simulation/src/rfq/protocols/metric/client_builder.rs
zach030: Added clarifying comments

crates/tycho-simulation/src/rfq/protocols/metric/models.rs
zach030: These fields are now parsed into `BigUint` once at deserialization

crates/tycho-simulation/src/rfq/protocols/metric/state.rs
zach030: Extracted into a METRIC_SWAP_GAS constant

crates/tycho-execution/contracts/src/executors/MetricExecutor.sol
```diff
@@ -5,14 +5,19 @@ import {IExecutor} from "@interfaces/IExecutor.sol";
 import {ICallback} from "@interfaces/ICallback.sol";
 import {TransferManager} from "../TransferManager.sol";
 
+// MetricOmmPool swap interface; see
+// https://oracle-based-omm.gitbook.io/metric -> Developers -> Smart Contracts Reference.
```
die-herdplatte: The link is a 401, it should be https://oracle-based-omm.gitbook.io/metric/RSm94m71kqtGICv4iKRj

crates/tycho-execution/contracts/test/protocols/Metric.t.sol
```diff
@@ -50,28 +44,34 @@ contract MockMetricPool {
...
+        address recipient,
         bool zeroForOne,
         int128 amountSpecified,
         uint128 priceLimitX64,
-        bytes calldata data
-    ) external {
+        bytes calldata callbackData,
+        bytes calldata extensionData
+    ) external returns (int128 amount0Delta, int128 amount1Delta) {
+        extensionData;
```
die-herdplatte: Can't we just ignore that with an underscore instead?

crates/tycho-simulation/src/rfq/protocols/metric/client.rs
```diff
@@ -1,3 +1,8 @@
+//! Metric RFQ client for the `api.metric.xyz` API.
+//!
+//! API and integration docs: <https://oracle-based-omm.gitbook.io/metric>
```
die-herdplatte: Same here, 401

die-herdplatte (approved):
Nice, only some nits remaining!

crates/tycho-simulation/src/rfq/protocols/metric/client.rs
die-herdplatte: This link is outdated

die-herdplatte (approved):

crates/tycho-simulation/src/rfq/protocols/metric/client.rs
zach030: Updated to https://docs.metric.xyz/RSm94m71kqtGICv4iKRj/developers/api

PR #1223, opened 2026-07-21
feat: support Ekubo V3 Ve33 pools

moodysalem:
## Summary

- index Ve33 `VoteWeightApplied` events and expose the current `swap_fee` as pool state
- quote Ve33 concentrated, full-range, and stableswap pools for exact-input and exact-output swaps
- execute Ve33 swaps through the required selectorless `Core.forward` path
- regenerate the Ekubo V3 runtime fixture and synchronize repository documentation

## Deployment

Robinhood mainnet and testnet use the deterministic Ve33 extension address `0xD18685a514E59b06d59824e16Db07e73345d9953`. The address is synchronized across Substreams, simulation, execution, and the runtime-bytecode fixture.

The executor tests are pinned to the merged Ve33 ABI and selectorless forwarding implementation from https://github.com/EkuboProtocol/evm-contracts/pull/360.

## Validation

- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `cargo test -p tycho-simulation evm::protocol::ekubo_v3` (58 passed)
- workspace Rustdoc and no-default-features checks
- Ekubo V3 Substreams Clippy, release WASM build, and package build
- `forge fmt --check` and `forge build`

The no-database nextest run reached 1,829 passes and 100 skips. The remaining 43 RPC-dependent tests failed or were interrupted because `RPC_URL` was unset; the Ve33-focused suites pass.


crates/tycho-execution/CLAUDE.md
```diff
@@ -161,6 +164,10 @@ the `TransferType`, receiver, tokenIn, tokenOut, and outputToRouter. Performs th
 `_currentSwappingExecutor` is stored in transient storage so `fallback()` knows which executor to route to. Cleared
 after the callback to prevent re-entrancy.
 
+Ekubo V3 normally calls `Core.swap`, but extension-owned swap call points require forwarding.
+Ve33 hops are detected by their extension address and sent through `Core.forward` with the Ve33
+selectorless swap payload; their packed hop encoding remains the normal fixed 52 bytes.
+
```
die-herdplatte: I don't think that's a necessary detail here

crates/tycho-simulation/src/evm/protocol/ekubo_v3/pool/ve33.rs
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    fn get_limit(&self, token_in: Address) -> Result<i128, SimulationError> {
+        match self {
+            Self::Concentrated(pool) => pool.get_limit(token_in),
+            Self::FullRange(pool) => pool.get_limit(token_in),
+            Self::Stableswap(pool) => pool.get_limit(token_in),
+        }
+    }
+}
```
die-herdplatte: You can just annotate the struct definition with `#[enum_delegate::implement(EkuboPool)]` and avoid this boilerplate.

crates/tycho-simulation/CLAUDE.md
```diff
@@ -35,6 +37,15 @@ fallback for protocols too complex to port, not a default.
...
 
+### Ekubo V3 Ve33 pools
+
+`ekubo_v3::pool::Ve33Pool` wraps full-range, stableswap, or concentrated native pool math. Ve33
+pools must have a zero Core fee; the simulator reads the extension-managed `swap_fee` state
+attribute and applies it to both exact-input and exact-output quotes. The Ekubo V3 Substreams
+package initializes that attribute and updates it from `VoteWeightApplied` events. On Robinhood,
+Ve33 is deployed at `0xD18685a514E59b06d59824e16Db07e73345d9953`; keep this address
+synchronized between the simulator, Substreams package, and executor.
+
```
die-herdplatte: Again, I don't think this fits here

protocols/substreams/ethereum-ekubo-v3/src/modules/3_map_filtered_events.rs
```diff
@@ -27,7 +27,8 @@ fn map_filtered_events(
                         {
                             Event::PoolInitialized(_) |
                             Event::VirtualExecution(_) |
-                            Event::RateUpdated(_) => true,
+                            Event::RateUpdated(_) |
+                            Event::SwapFeeUpdated(_) => true,
```
die-herdplatte: I think these three variants should all move to the conditional arm, they shouldn't pass the filter if the pool initializations haven't been registered.

A comment on this module explaining that this filter is a no-op as long as the package is indexed from its intended start block would be nice.
Right now it exists to make range integration tests easier, which don't sync from the start.

protocols/substreams/ethereum-ekubo-v3/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "ethereum-ekubo-v3"
-version = "0.1.1"
+version = "0.1.3"
```
die-herdplatte: This should probably be 0.2.2 like in the `Cargo.toml` (got out-of-sync before)

protocols/substreams/CLAUDE.md
```diff
@@ -22,8 +22,12 @@ contract) as a starting point. Implement the map modules, update the manifest, a
...
-merging to main — never merge changes without a version bump.
+Every PR that touches a package **must** bump that package's version in its `Cargo.toml` and add a
+`CHANGELOG.md` entry before merging to main — never merge package changes without both updates.
 
 - **Minor bump** (e.g. `0.3.2` → `0.3.3`): bug fixes, small adjustments
 - **Major bump** (e.g. `0.3.2` → `0.4.0`): significant changes, breaking output format
+
+The Ekubo V3 package also tracks extension-owned state. Ve33 pool creation initializes a
+`swap_fee` attribute, and `VoteWeightApplied` events update it so downstream native quotes use the
+current extension fee.
```
die-herdplatte: Same thing here

die-herdplatte (changes requested):
Hey, thanks for the PR!

I don't think that `VE33_ADDRESS` (and `SIGNED_EXCLUSIVE_SWAP_ADDRESS`, not part of this PR) should be hard-coded anywhere, they are chain-specific [according to the GitHub release notes](https://github.com/EkuboProtocol/evm-contracts/releases/tag/v3.2.0).

For the executor, I see two options:
- the addresses become `immutable`. Chains without support for any of the extensions pass a (non-zero) dead-address sentinel to the constructor.
- Convert the executor to an `abstract contract EkuboV3ExecutorBase` and expose a `_swapHop` function (see below). Add `EkuboV3EthereumExecutor` and `EkuboV3RobinhoodExecutor` that inherit from the base executor and override this function to handle the chain-specific extensions, delegating to the base contract for the common extensions.

```solidity
/// @dev Executes one hop against Core and returns its balance update and the
/// offset of the next hop. `offset` points just past the hop's fixed 52-byte
/// header (tokenOut | poolConfig), which the loop has already parsed and
/// bounds-checked. Implementations whose hops carry a self-describing tail
/// (e.g. signed hops) read it from `swapData` starting at `offset`, must
/// bounds-check it against `swapData.length`, and return the advanced offset;
/// fixed-size hops return `offset` unchanged.
function _swapHop(
    PoolKey memory poolKey,
    SwapParameters swapParameters,
    bytes calldata swapData,
    uint256 offset
) internal virtual returns (PoolBalanceUpdate balanceUpdate, uint256 nextOffset);
```

I like the second one better, it avoids sentinels and reduces the deployed contract to only contain reachable code.
There's only precedent in this repo for the first option though (`UniswapV4Executor`'s `_angstromHookAddress`).

On the Rust side, I checked and every current usage site somehow has access to a `tycho_common::models::Chain`, so you can branch based on that and rename the constant to `VE33_ROBINHOOD_ADDRESS`.

In substreams, you can pass string `params` to the modules that need the address (e.g. [like here](https://github.com/propeller-heads/tycho/blob/886655fc2ecd5336b7b42f944df5e18f4a2caea5/protocols/substreams/ethereum-uniswap-v3/ethereum-uniswap-v3.yaml)).
Keep it optional here such that Ethereum doesn't need to pass an address.

crates/tycho-execution/CLAUDE.md
moodysalem: Removed.

crates/tycho-simulation/CLAUDE.md
moodysalem: Removed.

protocols/substreams/CLAUDE.md
moodysalem: Removed.

crates/tycho-simulation/src/evm/protocol/ekubo_v3/pool/ve33.rs
moodysalem: Done. One wrinkle: the trait's `quote` signature referenced `super::pool::EkuboPoolQuote`, which doesn't resolve from the enum's module when `enum_delegate` re-expands it, so the trait now spells the return type with an absolute path.

protocols/substreams/ethereum-ekubo-v3/src/modules/3_map_filtered_events.rs
moodysalem: Moved all three variants to the conditional arm and added a module comment explaining that the filter is a no-op when synced from the intended start block and exists for range integration tests.

protocols/substreams/ethereum-ekubo-v3/Cargo.toml
moodysalem: Set to 0.2.2. Upstream main had meanwhile bumped the yaml to v0.2.1 and the Cargo.toml to 0.1.2 for the SignedExclusiveSwap change; after the merge both files are now aligned at 0.2.2, and the changelog notes the re-sync.

moodysalem:
Thanks for the review! All feedback is addressed, and the branch is synced with main.

**Executor** — went with the second option. `EkuboV3ExecutorBase` is now abstract and holds the chain-agnostic logic plus the extensions with universal deterministic addresses (MEVCapture). It exposes the suggested virtual `_swapHop`, plus a `_hopEnd` hook because the `getTransferData` hop walk also needs the chain-specific stride for hops with a self-describing tail. `EkuboV3EthereumExecutor` adds SignedExclusiveSwap (moved out of the base since it's also deployment-specific) and `EkuboV3RobinhoodExecutor` adds Ve33. `executor_deployments.json` maps ethereum → `EkuboV3EthereumExecutor` and robinhood → `EkuboV3RobinhoodExecutor`, and the `EkuboV3.runtime.json` fixture is regenerated from the Ethereum executor.

**Rust** — `extension_type` now takes a `tycho_common::models::Chain` (available at both usage sites via `ComponentWithState`) and only matches the renamed `VE33_ROBINHOOD_ADDRESS` on `Chain::Robinhood`.

**Substreams** — `map_events` and `map_components` take an optional string param carrying the Ve33 address (same style as the uniswap-v3 factory param). The default is empty, so Ethereum needs no change and Ve33 handling is disabled there; a Robinhood deployment passes the address via params.

protocols/substreams/ethereum-ekubo-v3/substreams.yaml
zach030, line 1: Could you add a `robinhood-ekubo-v3.yaml` with the Ve33 address in params? Plus a Robinhood integration test config if feasible.

crates/tycho-execution/contracts/test/protocols/EkuboV3.t.sol
```diff
@@ -298,11 +349,63 @@ contract LiquidityHelper {
...
+
+    // Runs on an Ethereum mainnet fork; Ve33 is planted at its Robinhood
+    // address via deployCodeTo, since Core is deployed at the same
+    // deterministic address on both chains.
+    function testVe33Swap() public setUpFork(24218590) {
+        deployCodeTo("Ve33.sol:Ve33", abi.encode(CORE, USDC_ADDR), VE33_ADDRESS);
+
+        PoolConfig poolConfig = createConcentratedPoolConfig({
+            _fee: 0, _tickSpacing: 100, _extension: VE33_ADDRESS
+        });
```
zach030, line 378: This tickspacing isn't a power of 4, and the underlying `checkValidPoolKey` checks for that, so this should error out.

die-herdplatte:
Superseded by #1337, which carries all commits from this PR unchanged (authorship preserved) plus the review fixes and the Robinhood rollout pieces.

Reason for the replacement: this PR's head branch lives in an org-owned fork, and GitHub does not allow maintainer pushes to those (`maintainerCanModify` is only honored for user-owned forks), so we could not update it. Closing in favor of #1337 — please carry any further changes over there.

PR #1224, opened 2026-07-21
fix: preserve block history across reinit for flashblock reverts

tvinagre:
## Problem

On Base with flashblocks enabled, `base-fynd` pods crash-loop with:

```
ERROR tycho_simulation::evm::stream: Block stream ended with terminal error:
Failed to process new block: Reverting block's insert position not found! History exceeded
```

The error is `BlockHistoryError::RevertPositionNotFound`, raised in `BlockHistory::push` when a revert drains the whole retained deque looking for its fork point.

## Root cause (confirmed from prod-fynd logs)

Reconstructed from the crashed container (`base-fynd-spot`, extractor `aerodrome_slipstreams`):

| Time | Event |
|------|-------|
| `14:59:58.338191` | `aerodrome_slipstreams` transition to **advanced** — jumped to `#48928326 (partial 1)` while the tip was `#48928325 (partial 11)` |
| `14:59:58.338235` | **Reinitialized block history** — rebuilt from stream headers only |
| `14:59:58.3–58.7` | history advances to `#48928327` |
| `14:59:59.711` | **crash** processing `aerodrome_slipstreams` revert to `#48928325 (partial 10)` → `RevertPositionNotFound` |

With flashblocks, one synchronizer regularly races a partial block ahead of the others. A detached higher block is classified `Advanced`, which triggers `reinit_block_history`. The old reinit rebuilt the history from `get_current_header()` of each stream only — one header per stream, all clustered at the leading blocks.

The precise mechanism: that seed rooted the new history at `#48928325 (p10)` (a slower stream's header, which is also the parent the advanced `#48928326 p1` connected to — hence it was `Advanced` rather than `NextExpected`, since its parent was `p10`, not the tip `p11`). Everything below that root — critically `#48928324` — was dropped. When the revert to `#48928325` arrived, `determine_block_position` classified it `NextExpected` (its hash is in history and `revert=true`, block_history.rs:236-251), so `push` drained the deque hunting for parent `#48928324`, never found it, and errored.

Note the ordering constraint: the revert target must itself still be **retained as the oldest block**. Had `#48928325` been dropped entirely, the revert would have been classified `Delayed` (`block.number < oldest.number`) and silently ignored — no crash. The failure needs the target retained as the root with its parent evicted.

## Why simply increasing the buffer size does not fix it

The initially suggested fix was to raise the two hardcoded `BlockHistory::new` sizes (`15` on startup, `10` on reinit). This would **not** have prevented the crash:

- The buffer did not overflow and evict the fork point. The reinit **rebuilt** the history from a seed that never contained `#48928324` at all, so `BlockHistory::new(seed, 10_000)` builds the identical `[325, 326]` chain and crashes the same way. Capacity is irrelevant when the block is absent from the seed.
- The startup `size = 15` is discarded by the reinit, so it has no bearing on this crash.
- Without the reinit, the revert from `#48928327` to `#48928325` is only two blocks deep and the existing `size = 10` buffer would have held it comfortably. The reinit is what threw away the history the revert needed.

## Fix

- `reinit_block_history` seeds the new history with the **retained blocks plus** current stream headers, so a revert to a block below the advanced tip still finds its fork point. `BlockHistory::new` keeps the connected chain ending at the highest-numbered header, so a genuinely detached advanced block (a real gap after a restart) still discards the old history rather than stitching a discontinuous chain.
- `BlockHistory::new` skips duplicate / same-height entries instead of truncating the chain, since the merged seed can contain the same block number twice. This half is load-bearing: without it the merged seed `[323,324,325,326,325]` breaks at the duplicate and yields `[325,326]` again.
- Startup and reinit share a single `BLOCK_HISTORY_SIZE` constant, so reinit no longer shrinks the retained depth from 15 to 10.

## Known residual (pre-existing, not introduced here)

If the advanced block's parent matches no retained or streamed header (possible with ephemeral flashblock hashes), the rebuilt history is still rooted at the advanced block. A later revert below it is then classified `Delayed` and silently dropped — the history never rewinds, and recovery waits for a subsequent reinit. That is a no-op rather than a crash, and is unchanged by this PR.

## Tests

- `test_reinit_preserves_history_for_revert_below_advanced_tip` reproduces the crash and fails with `RevertPositionNotFound` on the old code. It covers **both** halves of the fix — it also fails with the `mod.rs` change alone.
- `test_new_tolerates_duplicate_blocks` covers the duplicate-seed chain-building change.
- `test_reinit_discards_retained_history_on_detached_advanced_block` locks in that a real gap still discards old history.
- Full `tycho-client` suite passes (145 lib tests); workspace clippy and nightly fmt clean.

### DIFF
--- a/crates/tycho-client/src/feed/block_history.rs
+++ b/crates/tycho-client/src/feed/block_history.rs
@@ -267,7 +267,8 @@ impl BlockHistory {
                 // sibling is canonical, so we do not flip the tip here. Classify it as Delayed:
                 // if it is the losing fork it simply waits, and if it is canonical the next block
                 // (number > tip) arrives as Advanced and triggers a block-history reinit that
-                // rebuilds from the synchronizers' converged headers.
+                // rebuilds from the retained history merged with the synchronizers' converged
+                // headers.
                 BlockPosition::Delayed
             } else {
                 // anything else raises e.g. a completely detached, revert=false block
--- a/crates/tycho-client/src/feed/block_history.rs
+++ b/crates/tycho-client/src/feed/block_history.rs
@@ -66,6 +66,11 @@ impl BlockHistory {
 
             // Find connected blocks in sequence
             for block in history.iter().skip(1) {
+                // Skip duplicates or same-height forks. The input may contain overlapping blocks
+                // (e.g. when merging retained history with current stream headers on reinit).
+                if block.number >= current_number {
+                    continue;
+                }
                 // If we find a gap in block numbers, stop building the chain
                 if block.number != current_number - 1 {
                     break;
@@ -290,6 +295,11 @@ impl BlockHistory {
     pub fn oldest(&self) -> Option<&BlockHeader> {
         self.history.front()
     }
+
+    /// Returns the retained block headers, oldest first.
+    pub fn blocks(&self) -> impl Iterator<Item = &BlockHeader> {
+        self.history.iter()
+    }
 }
 
 #[cfg(test)]
@@ -400,6 +410,21 @@ mod test {
         assert_eq!(history.history, exp_history);
         assert!(history.reverts.contains(&int_hash(3)));
         assert!(history.reverts.contains(&int_hash(4)));
+    }
+
+    #[test]
+    fn test_new_tolerates_duplicate_blocks() {
+        // Reinit seeds `new` with retained history plus current stream headers, so the same block
+        // number can appear more than once. A duplicate must not truncate the connected chain.
+        let mut blocks = generate_blocks(4, 5, None); // blocks 5,6,7,8
+        blocks.push(blocks[2].clone()); // duplicate block 7
+
+        let history = BlockHistory::new(blocks, 10).expect("failed to create history");
+
+        // All four connected blocks are retained despite the duplicate.
+        assert_eq!(history.history.len(), 4);
+        assert_eq!(history.oldest().unwrap().number, 5);
+        assert_eq!(history.latest().unwrap().number, 8);
     }
 
     #[test]
--- a/crates/tycho-client/src/feed/mod.rs
+++ b/crates/tycho-client/src/feed/mod.rs
@@ -976,11 +976,12 @@ where
             .ok_or(BlockHistoryError::EmptyHistory)?
             .clone();
         // Preserve the previously retained history so a revert to a block below the advanced tip
-        // can still find its fork point. Seeding only from current stream headers would root the
-        // new history at the advanced block and drop the ancestors a later revert needs, which
-        // surfaces as `RevertPositionNotFound` ("History exceeded"). `BlockHistory::new` keeps the
-        // longest connected chain ending at the latest header, so detached older blocks (a genuine
-        // gap, e.g. after a restart) are still discarded.
+        // can still find its fork point. Seeding only from current stream headers roots the new
+        // history at the oldest header the streams happen to hold, dropping the ancestors a later
+        // revert needs: a revert targeting that root drains the deque looking for its parent and
+        // fails with `RevertPositionNotFound` ("History exceeded"). `BlockHistory::new` keeps the
+        // connected chain ending at the highest-numbered header, so detached older blocks (a
+        // genuine gap, e.g. after a restart) are still discarded.
         let mut blocks: Vec<BlockHeader> = block_history
             .blocks()
             .cloned()
@@ -1494,6 +1495,42 @@ mod tests {
         new_history
             .push(revert)
             .expect("revert below advanced tip must not exceed history");
+    }
+
+    /// Preserving history must not resurrect blocks across a real gap. When the advanced block is
+    /// genuinely detached (e.g. a synchronizer restarted and jumped ahead), the retained blocks do
+    /// not connect to it and must still be discarded, leaving the history rooted at the advanced
+    /// block rather than stitching together a discontinuous chain.
+    #[test]
+    fn test_reinit_discards_retained_history_on_detached_advanced_block() {
+        let old_blocks = vec![
+            full_header(323, 323, 322),
+            full_header(324, 324, 323),
+            full_header(325, 325, 324),
+        ];
+        let mut old_history = BlockHistory::new(old_blocks, BLOCK_HISTORY_SIZE).unwrap();
+
+        // Advanced block is far ahead and its parent is unknown to the retained history.
+        let mut streams = vec![stream_in_state(
+            "aerodrome",
+            SynchronizerState::Advanced(full_header(400, 400, 399)),
+        )];
+
+        let new_history = BlockSynchronizer::<MockStateSync>::reinit_block_history(
+            &mut streams,
+            &mut old_history,
+        )
+        .expect("reinit failed");
+
+        let retained: Vec<u64> = new_history
+            .blocks()
+            .map(|b| b.number)
+            .collect();
+        assert_eq!(
+            retained,
+            vec![400],
+            "detached advanced block must not be stitched to old history"
+        );
     }
 
     #[test(tokio::test)]
--- a/crates/tycho-client/src/feed/mod.rs
+++ b/crates/tycho-client/src/feed/mod.rs
@@ -33,6 +33,11 @@ pub mod component_tracker;
 pub mod dto;
 pub mod synchronizer;
 
+/// Number of block headers retained by the `BlockHistory` ring buffer. Bounds how deep a revert
+/// can unwind before the fork point is evicted. Used both on startup and on reinitialization so
+/// the retained depth stays consistent across an `Advanced`-triggered reinit.
+const BLOCK_HISTORY_SIZE: usize = 15;
+
 /// A trait representing a minimal interface for types that behave like a block header.
 ///
 /// This abstraction allows working with either full block headers (`BlockHeader`)
@@ -783,7 +788,7 @@ where
 
         // Fail fast if no synchronizer produced a ready first message.
         Self::require_active_stream(&sync_streams)?;
-        let mut block_history = BlockHistory::new(initial_headers, 15)?;
+        let mut block_history = BlockHistory::new(initial_headers, BLOCK_HISTORY_SIZE)?;
         // Determine the starting header for synchronization.
         // Safe: require_active_stream above guarantees at least one Ready stream,
         // so initial_headers is non-empty and block_history.latest() is Some.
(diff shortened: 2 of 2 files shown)


crates/tycho-client/src/feed/block_history.rs
```diff
@@ -290,6 +279,11 @@ impl BlockHistory {
     pub fn oldest(&self) -> Option<&BlockHeader> {
         self.history.front()
     }
+
+    /// Returns the retained block headers, oldest first.
+    pub fn blocks(&self) -> impl Iterator<Item = &BlockHeader> {
+        self.history.iter()
+    }
```
louise-poole, line 303: If this is only used in tests can you put it behind a test feature flag?

louise-poole (approved):
Looks good! Have you tested this locally already? It touches on some core decisions in the client so let's be ready to revert this if we see things go bad on the cluster tests.

crates/tycho-client/src/feed/block_history.rs
tvinagre: It's not test-only — it's the mechanism of the fix. `blocks()` is called in production at `feed/mod.rs:964`, inside `reinit_block_history`:

```rust
let mut blocks: Vec<BlockHeader> = block_history
    .blocks()
    .cloned()
    .collect();
```

That's the call that carries the previously retained history into the rebuilt `BlockHistory`, so a revert to a block below the advanced tip can still find its fork point. Putting it behind a test feature flag would break the non-test build.

On visibility: `block_history` is a private module (`mod block_history;`), so `pub` here is only crate-visible and doesn't widen the public API — same as the existing `latest()` and `oldest()`.

tvinagre:
> Have you tested this locally already?

Locally yes, on the cluster no — worth being explicit about where the confidence comes from and where it doesn't.

**What is verified locally**
- `test_reinit_preserves_history_for_revert_below_advanced_tip` reproduces the production crash and fails with `RevertPositionNotFound` against the pre-fix code. It covers both halves — it also fails with the `mod.rs` change alone, without the `BlockHistory::new` duplicate-skip.
- `test_new_tolerates_duplicate_blocks` and `test_reinit_discards_retained_history_on_detached_advanced_block` cover the chain-building change and the real-gap path.
- Full `tycho-client` suite: 145 passed, 0 failed. Workspace clippy `--all-targets --all-features`: 0 warnings. Nightly fmt clean.
- The root cause was reconstructed from the actual crashed pod's logs (`base-fynd-spot`, previous container), not inferred from the code alone.

**What is NOT verified**
No cluster/flashblocks run. The unit tests exercise the reinit-then-revert sequence with synthetic headers; they do not reproduce real flashblock ephemeral hashes. So I agree with staying ready to revert.

**One correction to my own description.** An independent review pass caught that my original root-cause narrative was internally inconsistent: I wrote that the history was rooted at `#48928326` with `#48928325` never inserted. That cannot produce this error — a revert to a block below `oldest` is classified `Delayed` and silently ignored (block_history.rs:243-246). The crash requires `#48928325` to have been **retained as the oldest block** with its parent `#48928324` evicted. The fix and the test were always correct for the real mechanism; the write-up was not. Description now updated.

**Since you flagged core-decision risk**, two things worth knowing for the revert calculus:
- This touches the same path as your recent `classify non-revert tip sibling as delayed`. That fix explicitly relies on reinit for recovery, and this PR strengthens exactly that path — the sibling swap now retains ancestry, so a later revert to the swapped tip resolves instead of crashing. They compose; neither is redundant.
- There is a pre-existing residual this PR does not close: if the advanced block's parent matches no retained or streamed header, a later revert below it is silently dropped rather than rewinding. That is a no-op, not a crash, and is unchanged here. Documented in the description.

If you would rather gate this behind a canary on one Base pod before it goes wide, that is easy to do.

tvinagre:
## Live verification — client run **locally** against `tycho-base-beta.propellerheads.xyz` (flashblocks enabled)

To be explicit about what this was: I built the client from this branch and **ran it locally on my own machine**, pointed at the hosted Base beta endpoint. **Nothing was deployed to any cluster** — no pods, no config, no infra were touched. The only thing exercised remotely was the hosted endpoint, as a normal read-only client.

Command shape:

```
TYCHO_URL=tycho-base-beta.propellerheads.xyz \
./target/release/tycho-client --chain base --partial-blocks \
  -e uniswap_v2 -e uniswap_v3 -e uniswap_v4 -e aerodrome_slipstreams -e pancakeswap_v3
```

Same extractor set as the crashed pod. I ran this branch and a pre-fix baseline (`origin/main`, i.e. this branch minus the fix) concurrently as two local processes against the same stream.

| | window | msgs | reverts | advanced | reinits | `History exceeded` |
|---|---|---|---|---|---|---|
| **This PR** (local) | 16:46:52–17:12:09 | 9855 | 525 | 3 | 2 | **0** |
| **Baseline / pre-fix** (local) | 16:51:45–17:12:08 | 8210 | 420 | 0 | 0 | 0 |

**Do not read the advanced/reinit columns as a behavioural difference — they are a measurement artifact.** I staggered the two starts, and all 5 events landed in the non-overlapping windows:

| Event cluster | Time | Baseline status |
|---|---|---|
| 2 advanced + 1 reinit | `16:51:13` | not yet started (first log `16:51:45`) |
| 1 advanced + 1 reinit | `17:12:09.598` | already exited (last log `17:12:08`) |

The baseline was never observing when either cluster occurred, so its `0` means "did not watch", not "did not happen". The PR cannot increase `Advanced` frequency in any case: `Advanced` is decided by `determine_block_position` purely from the **tip**, and this PR only retains ancestors *below* the tip — it never changes the tip. No production line in the diff touches `Advanced` classification or `transition()`.

Both exited cleanly, empty stderr, no terminal errors.

**The fixed client hit the exact production trigger and survived.** The reinit it went through has the same signature as the crash:

```
transition to advanced. extractor_id=base:pancakeswap_v3
  latest=Block #48974863 [0xc803807d..] (partial 10)  block=Block #48974864 (partial 2)
transition to advanced. extractor_id=base:aerodrome_slipstreams
  latest=Block #48974863 [0xc803807d..] (partial 10)  block=Block #48974864 (partial 2)
Advanced synchronizer detected. Reinitialized block history.
  previous=Block #48974863 [0xc803807d..] (partial 10)  latest=Block #48974864 [0xb7e605ad..] (partial 2)
```

Compare production: `aerodrome_slipstreams` advanced `#48928325 p11` → `#48928326 p1`, reinit `previous=325 p11 latest=326 p1`, then crashed on the next revert. Same shape — one synchronizer racing a partial ahead, reinit rebuilding at the leading partial — and here it went through 2 reinits with 525 reverts flowing and never raised `RevertPositionNotFound`.

### Limitation — read this before treating it as proof

This is **corroborating evidence, not a controlled A/B**. The baseline never hit an `advanced` transition in its window (0 of them), so I did not observe the old code crash. Absence of a baseline crash says nothing about the baseline — the trigger is stochastic, and the fixed run simply hit it three times while the baseline hit it zero times.

So: the fix demonstrably survives the real trigger under real flashblocks traffic, but I have not empirically shown the pre-fix binary failing on this endpoint. The evidence that the old code crashes remains the prod logs plus the unit test, which does fail with `RevertPositionNotFound` against pre-fix code.

Still worth staying revert-ready on rollout, as suggested.

tvinagre:
## Controlled A/B: this branch vs pre-fix baseline, 3h against `tycho-base-beta` (flashblocks)

Redo of the earlier comparison with proper controls. My first attempt staggered the start times, so the interesting events fell outside the shared window and the numbers were not comparable. This run fixes that.

**Setup.** Two locally-run clients, same machine, launched at the **same instant** and run for the **same duration**: `17:38:27Z → 20:38:27Z` (3h), both `exit=0`. Same endpoint, same extractors (`uniswap_v2/v3/v4`, `aerodrome_slipstreams`, `pancakeswap_v3`), `--partial-blocks`. Baseline = `origin/main` (this branch minus the fix), verified pre-fix. Nothing deployed to any cluster.

Both saw **the same stream**: 1212 reverts each, and identical message pacing throughout.

| | msgs | reverts | advanced | reinits | `History exceeded` | `Could not determine` | stderr |
|---|---|---|---|---|---|---|---|
| **This PR** | 58850 | 1212 | 15 | **6** | 0 | 0 | empty |
| **Baseline / pre-fix** | 58220 | 1212 | 32 | **13** | 0 | 0 | empty |

### The signal is in the reinit timestamps

```
FIXED     17:55:47  18:21:30                    19:03:07                      19:37:44           20:38:11 20:38:25
BASELINE  17:55:47  18:21:30 18:21:45 18:29:58  19:03:07 19:03:17 19:03:27    19:37:44 19:37:58  20:38:11 20:38:25
                             ^^^^^^^^ ^^^^^^^^           ^^^^^^^^ ^^^^^^^^             ^^^^^^^^
                                                19:27:35 19:30:34
```

**All 6 of the fixed client's reinits appear in the baseline at byte-identical timestamps.** Those are the genuine stream events — both clients react to them the same way. The baseline then has **7 additional reinits that the fixed client never has**, and they cluster within seconds of a shared one: `18:21:30 → +15s`, `19:03:07 → +10s, +20s`, `19:37:44 → +14s`.

That is the cascade the fix removes. When the pre-fix reinit rebuilds history rooted at the advanced block with no ancestors, the streams' next blocks fail to connect, so they get re-classified `Advanced` and force another rebuild seconds later. Several of these show **all five extractors going `Advanced` in the same millisecond** — the signature of a history tip that no stream agrees with:

```
18:21:45.535  transition to advanced  base:aerodrome_slipstreams
18:21:45.535  transition to advanced  base:pancakeswap_v3
18:21:45.535  transition to advanced  base:uniswap_v3
18:21:45.535  transition to advanced  base:uniswap_v4
18:21:45.535  transition to advanced  base:uniswap_v2
18:21:45.535  Advanced synchronizer detected. Reinitialized block history.
```

The fixed client recovers from the same stream event in one reinit, because the retained ancestry lets the streams reconnect. Net: **2.2x the reinits and 2.1x the advanced transitions on pre-fix**, plus slightly lower throughput (58220 vs 58850 msgs) from the extra rebuild work.

### What this does NOT show

**Neither side crashed.** `History exceeded` was 0 on both, so I still have not reproduced the actual `RevertPositionNotFound` failure on this endpoint. The crash needs `advanced → reinit → revert landing below the new root`; the first two are common (13 reinits in 3h on baseline) but the third did not occur in this window.

So this run demonstrates that the fix **substantially reduces reinit churn and keeps history continuous**, which is the same underlying defect. It does not itself prove the crash is gone. The evidence for that remains the prod logs plus the unit test, which fails with `RevertPositionNotFound` against pre-fix code.

Worth noting the churn reduction is a real independent benefit — the repeated shallow rebuilds are exactly the degraded state the crash emerged from.

PR #1228, opened 2026-07-22
fix(fermiswap): follow engine migration to 0x90f73fEA

die-herdplatte:
## Problem

`vm:fermiswap` stopped being simulatable on 2026-07-21:

```
StateDecodingFailure pool="0xa6175c23…" error=Unable to set up vm state on the engine:
Recoverable error: Storage error: MissingAccount(0x90f73fea1ee2dc514d4dbac0bff7ff04b933767f)
```

Fermi migrates engines by re-pointing the swapper's engine slot (storage slot 2). On 2026-07-21 they deployed a new engine `0x90f73fEA…` (block 25581530, upgraded bytecode — ABI is a strict superset of the old engine's), re-registered all 8 pairs on it (blocks 25581596–25581615), and re-pointed the swapper (tx `0xf4ed64dc…`, block 25581704). Our substreams params still tracked the old engine `0xe514A3c4…`, so:

- the new engine's storage was never indexed → every quote path through the swapper hit `MissingAccount`,
- `PairRegistered`/`PairActiveSet`/`PairUnregistered` on the new engine were invisible,
- registry `updateState` calls no longer matched the target filter → `override_block_timestamp` attributes froze → even decodable pools would price against stale oracle state.

## Changes

- **`ethereum-fermiswap` 0.2.0**: `engine_address` → `0x90f73fEA…`. `initialBlock` stays at the vault creation block 24936662 — it must cover the creation of every tracked contract, since account rows are only created for in-range deployments and storage deltas against absent accounts kill the extractor (now documented in the manifest).
- **Manifest dedup**: single anchored `initialBlock` + `params` entry (same pattern as base-aerodrome-slipstreams) instead of the 9-entry `networks` map.
- **Integration tests retargeted** to the new engine's history, execution enabled (executor is registered and deployed). Stop blocks are constrained by the engine's ~30s oracle staleness window (`StaleUpdate`, `0x666a2814`); the comments in the yaml document the constraints.
- **`tycho-test`**: `FERMISWAP_TARGET_ADDRESS` → new engine. `setup_fermiswap_overwrites` computes the registry lane slot as `keccak(abi.encode(target, laneIndex))`; with the old target it patched a dead slot, silently disabling the staleness workaround in both protocols/testing and tycho-integration-test.
- **Record `storage_changes`** in `map_protocol_changes` via `get_block_storage_changes` (same helper as balancer, uniswap-v4, fluid, curve, erc4626). DCI-only data — inert while fermiswap emits no entrypoints, but groundwork for DCI-based engine discovery.
- **CHANGELOG.md** added for the package.

## Operational rollout

1. Release the `ethereum-fermiswap` v0.2.0 spkg.
2. **Full resync of the `vm:fermiswap` extractor from block 24936662 (~650k blocks), with the extractor's existing data wiped first.** A rolling update is not possible: existing components carry the old engine in their contract set and there are no new `PairRegistered` deltas to re-create them from an existing cursor. Component ids are stable across the migration (`keccak(base ++ quote)` is engine-independent), so consumers keep their ids.
3. Wiping the stale components matters beyond hygiene: Fermi poisons the outgoing engine with garbage prices during wind-down (its last oracle update quoted WETH at 9.53 USDC vs ~1926 real). Old components left simulatable would mis-price by ~200×, not just revert.
4. `tycho-test` ships in the same deploy so the execution harnesses patch the live lane slot again.

**For the next migration** (this will recur — the engine pointer is a mutable slot): the signature is `MissingAccount(<unknown new address>)` on fermiswap pools, and `cast storage 0xb1076fE3…5C7c323Bac5AC06a680d452e 2` shows the current engine. `engine_address` (substreams) and `FERMISWAP_TARGET_ADDRESS` (tycho-test) must move together; the staleness window is engine-version-specific (old: ~12s, new: ~30s) and should be re-measured. Follow-up worth considering: DCI-based engine discovery via the swapper's slot 2 to remove the static coupling.

## Verification

All harness results below were produced at c2b1fa8ca; the follow-up 33120d102 only populates `storage_changes` and was verified separately with a live `substreams run` (module executes on the extended block model, output populated: 412/347 per-tx storage entries at blocks 24936663/24936664; ignored by the indexer absent DCI entrypoints).

- On-chain: registry writes the new-engine lane at the slot the code computes (timestamp prefix confirmed at block 25584584); a single-slot timestamp override flips a stale quote to success, proving the engine reads exactly that slot; both test pairs quote successfully at the test stop blocks.
- protocols/testing `range`: both yaml tests pass — indexing, simulation, and execution (12/12) at the pinned stop blocks.
- protocols/testing `full` (live, against a locally resynced DB): continuous execution checks at the chain tip. With the harness's default token-quality filter only WETH/USDC and WBTC/USDC are exercised (12/12 per update block); with all six tokens loaded (local quality curation), every decodable pool is exercised — 36/36 simulations + executions per update block, USDT pairs included.
- tycho-integration-test against the local resync (24936662 → tip): 456/456 simulations passed over 60 blocks (get_limits / get_amount_out / debug_traceCall execution), slippage 0.0000%; a second 30-block run passed 276/276 with slippage within ±0.11%. A live fill (block 25594963) was observed flowing through the pipeline in the same block.
- The 100% execution rate directly validates the tycho-test target fix: most tested blocks had oracle lanes older than the ~30s staleness window, so executions only pass if the lane overwrite patches the live registry slot.

Detailed test results and issues found during testing: see the test-results comment below.

### DIFF
--- a/protocols/substreams/ethereum-fermiswap/src/modules.rs
+++ b/protocols/substreams/ethereum-fermiswap/src/modules.rs
@@ -501,6 +501,8 @@ fn map_protocol_changes(
             .sorted_unstable_by_key(|(index, _)| *index)
             .filter_map(|(_, builder)| builder.build())
             .collect::<Vec<_>>(),
+        // Currently these storage changes don't have a consumer. We still emit them for the case
+        // that we manually add DCI entrypoints later.
         storage_changes: get_block_storage_changes(&block),
     })
 }
--- a/protocols/substreams/ethereum-fermiswap/src/modules.rs
+++ b/protocols/substreams/ethereum-fermiswap/src/modules.rs
@@ -33,6 +33,7 @@ use tycho_substreams::{
     abi::{erc20, weth},
     attributes::json_serialize_address_list,
     balances::aggregate_balances_changes,
+    block_storage::get_block_storage_changes,
     contract::extract_contract_changes_builder,
     prelude::*,
 };
@@ -500,6 +501,6 @@ fn map_protocol_changes(
             .sorted_unstable_by_key(|(index, _)| *index)
             .filter_map(|(_, builder)| builder.build())
             .collect::<Vec<_>>(),
-        storage_changes: vec![],
+        storage_changes: get_block_storage_changes(&block),
     })
 }
--- a/crates/tycho-test/src/execution/encoding.rs
+++ b/crates/tycho-test/src/execution/encoding.rs
@@ -40,7 +40,10 @@ pub const EXECUTOR_ADDRESS: &str = "0xaE04CA7E9Ed79cBD988f6c536CE11C621166f41B";
 // Fixed address used to plant FeeCalculator bytecode in state overrides.
 pub const FEE_CALCULATOR_ADDRESS: &str = "0xfEEcA1C0fEEcA1C0fEEcA1C0fEEcA1C0fEEcA1C0";
 const FERMISWAP_REGISTRY_ADDRESS: &str = "0xDA7AFeEd01fe625cF15D187A19F94B45F00b8C5f";
-const FERMISWAP_TARGET_ADDRESS: &str = "0xe514A3c48DA8B233f65b5d15BA1905d6d35BFE48";
+// The Fermi engine currently pointed at by the swapper's storage slot 2. Fermi migrates engines
+// by re-pointing the swapper (last: 2026-07-21, block 25581704); update this together with
+// `engine_address` in the fermiswap substreams params, or lane overwrites patch a dead slot.
+const FERMISWAP_TARGET_ADDRESS: &str = "0x90f73fEA1Ee2Dc514d4dbAc0bfF7ff04b933767f";
 // BopAMM prices its books from the same PrioUpdateRegistry as FermiSwap; only the registry
 // `target` differs — it is the pricing module, and the lane index is the book's `assetId`.
 const BOPAMM_REGISTRY_ADDRESS: &str = "0xDA7AFeEd01fe625cF15D187A19F94B45F00b8C5f";
--- a/protocols/substreams/Cargo.lock
+++ b/protocols/substreams/Cargo.lock
@@ -916,7 +916,7 @@ dependencies = [
 
 [[package]]
 name = "ethereum-fermiswap"
-version = "0.1.1"
+version = "0.2.0"
 dependencies = [
  "anyhow",
  "ethabi 18.0.0",
--- a/protocols/substreams/ethereum-fermiswap/CHANGELOG.md
+++ b/protocols/substreams/ethereum-fermiswap/CHANGELOG.md
@@ -1,0 +1,34 @@
+# Changelog
+
+## v0.2.0
+
+### Changed
+
+- Follow the Fermi engine migration of 2026-07-21: `engine_address` now points at the new engine
+  `0x90f73fEA1Ee2Dc514d4dbAc0bfF7ff04b933767f` (deployed block 25581530). Fermi migrates engines
+  by re-pointing the swapper's engine slot (re-point tx at block 25581704); with the old address
+  the package missed pair registrations and pause events on the new engine, no longer matched the
+  registry `updateState` calls that feed the `override_block_timestamp` attribute, and simulation
+  failed with `MissingAccount(<new engine>)` because the new engine's storage was never indexed.
+- Retarget the integration tests to the new engine's history. Stop blocks are constrained by the
+  engine's oracle staleness window (~30s, `StaleUpdate`): state decoding snapshots at the stop
+  block and quotes through the adapter, so the stop block must come after the swapper re-point and
+  the first `updateState` for every tested lane. Execution checks are enabled (the `vm:fermiswap`
+  executor is registered and deployed).
+- Deduplicate the manifest: single anchored `initialBlock` and params entry instead of the
+  per-module `networks` map. `initialBlock` stays at the trader-vault creation block 24936662 — it
+  must remain at or before the creation of every tracked contract, because account rows are only
+  created for in-range deployments and storage deltas against absent accounts kill the extractor.
+
+### Migration
+
+Requires a full resync of the `vm:fermiswap` extractor. Existing components carry the old engine
+in their contract set and cannot be patched forward; the new engine re-registered all pairs, so
+components are re-created (same component ids — `keccak(base ++ quote)` is engine-independent).
+Stale components from the old engine must be wiped, not left in place: Fermi poisons the outgoing
+engine with garbage prices during wind-down (last old-engine update quoted WETH at 9.53 USDC).
+
+The engine address is also hardcoded as `FERMISWAP_TARGET_ADDRESS` in
+`crates/tycho-test/src/execution/encoding.rs` (updated together with this release) — it must be
+kept in sync with `engine_address` on every future migration, or the execution harness's lane
+staleness overwrite patches a dead registry slot.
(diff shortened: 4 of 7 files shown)


die-herdplatte:
## Test results

Setup: full local resync of `vm:fermiswap` from block 24936662 to the chain tip with the v0.2.0 spkg, then all three harnesses against it. All results were produced at c2b1fa8ca; the follow-up commit 33120d102 (record `storage_changes`) is not covered by these runs — it was verified separately via a live `substreams run` (module executes on the extended block model, `storageChanges` populated, DCI-only and ignored by the indexer absent entrypoints).

**protocols/testing `range`** (Docker): both yaml tests pass.
- `test_pair_creation` (25581590–25581903): all 8 components indexed with correct tokens, static attributes, and creation txs.
- `test_pair_simulation` (25581590–25584584): simulation + execution enabled for WETH/USDC and WETH/USDT — 12/12 executions pass.

**protocols/testing `full`** (live, `--reuse-last-sync` on the resynced DB): continuous execution checks at the tip, 12/12 passing on every update block (WETH/USDC, WBTC/USDC; gas ~139–142k), quotes in line with market.

**tycho-integration-test** (local indexer, `vm:fermiswap` only):
- 60 blocks: 456/456 simulations passed (get_limits, get_amount_out, encode, debug_traceCall execution), slippage 0.0000% on all samples, 0 errors.
- 30 blocks (instrumented rerun): 276/276 passed, slippage within ±0.11%.
- A live fill at block 25594963 (WETH/USDC) arrived in the stream the same block: lane-stamp attribute plus vault balance fan-out to every pair sharing the moved tokens.
- Since fills are sparse, most tested blocks had lanes older than the ~30s staleness window — the 100% execution rate is only possible because `setup_fermiswap_overwrites` now patches the live registry slot (the `FERMISWAP_TARGET_ADDRESS` fix in this PR).

## Issues found during testing

> ⚠️ **The section below is superseded** — the failure is fully root-caused in the follow-up comment further down (adapter `getLimits` probe flaw; no tycho-simulation state divergence). Kept for history.

**Core — pre-existing, needs its own issue (not caused by this PR):** all four USDT pairs degrade in `tycho-simulation` via the spot-price probe in `EVMPoolState::set_spot_prices`:
- WBTC/USDT and cbBTC/USDT revert `IL` during snapshot decoding (`decoder.rs` calls `set_spot_prices` on construction) → the components are silently excluded from the stream for the whole session (`failed_components`).
- WETH/USDT and USDC/USDT decode but fail the per-block spot-price refresh with `COR` (occasionally `ArithmeticOver/Underflow`), on both the Titan-override path and the indexed-state retry → spot prices stay permanently stale and the pools drop out of `updated_states`.
- The live venue is healthy in every failing direction (verified via lane-timestamp-patched `quoteAmounts`, dust to 1M USDT: correct quotes, depth-clamps instead of reverting), and `get_amount_out`/execution simulate fine for the same pairs. The failure is specific to the probe environment (`getLimits/100` probe under `MAX_BALANCE/100` balance overwrites vs the engine's corridor/inventory math). Effect: live simulation coverage is currently capped at the USDC pairs, and USDT-pair spot prices never refresh for consumers (visible in price_printer). Repro: `range --reuse-last-sync --vm-simulation-traces` on this branch's yaml reproduces `IL` for WBTC/USDT deterministically at the stop block.

**Harness (protocols/testing):**
- `full` mode + `--reuse-last-sync` + `initialized_accounts` panics: account init inserts contracts against a synthetic transaction, but on a reused DB the start block already exists, the synthetic tx is skipped, and the insert fails its FK (`NoRelatedEntity("Transaction", …, "Account")`). Account init is redundant on reuse and should be skipped. The failed write rolls back cleanly. Workaround: comment out `initialized_accounts` for such runs.
- The synthetic `native_wrapper` pool simulates but has no swap encoder → one "Swap encoder not found" ERROR per full-mode session; ignorable.

die-herdplatte:
## Follow-up: USDT-pair spot-price failures root-caused (supersedes the "Core" section above)

The issue described above as "specific to the probe environment" is now fully root-caused — trace-proven in simulation (`--vm-simulation-traces`) and reproduced live with `cast` state overrides. It is a **FermiSwapAdapter flaw**; the simulation faithfully mirrors the venue (in-sim and live traces are frame-identical), there is no sim-state divergence.

**Mechanism.** `FermiSwapAdapter.getLimits` quotes an exact-output buy of the *entire* trader-vault balance of the buy token (`balanceOf(traderVault)`, read at call time). The probe is only serviceable while `vault_balance(buyToken) <= lane serviceable size`. When that invariant breaks, the engine refuses:

- clean state, thin lane → `Error("IL")` during snapshot decoding; `set_spot_prices` is fatal there (`vm/decoder.rs:251`), so the pool is silently benched for the whole session (WBTC/USDT, cbBTC/USDT);
- clean state serviceable, but the post-trade spot-price refresh — `get_amount_out` recomputes spot prices on the returned post-swap state (`state.rs:824`, result discarded) — re-probes against the trade-tightened corridor → `Error("COR")` or `ArithmeticOver/Underflow`, one WARN per trade size (WETH/USDT, USDC/USDT). Every captured `COR` follows a successful `getLimits` + simulated trade;
- vault holding ≤ lane depth (all USDC-quoted pairs) → the engine clamps; everything works.

The vault's USDT position (~1.0–1.4M, its largest) is the only holding statically above its lanes' serviceable size — hence the "only USDT/cbBTC pairs" pattern. Quoting, execution and per-block stored states are unaffected (full-mode live checks pass 36/36 per update block including USDT pairs once tokens are loaded); the damage is decode-time pool loss and stale spot prices plus WARN noise.

**Corrections to the section above:**
- The suggested repro (`range --reuse-last-sync --vm-simulation-traces`) is wrong on a retention-pruned local DB: contract-storage history is dropped while attributes time-travel, so the sim assembles era-mixed state and everything reverts `StaleUpdate` at the registry read — a different failure. Reproduce at the tip instead: full mode + `--reuse-last-sync` + `--vm-simulation-traces`.
- "USDC/USDT intermittently recovers" was lane-depth state changing between stamps, not client behavior.

**Fix directions (to be filed separately):**
1. Adapter (this repo): make the `getLimits` probe survivable — cap it below the serviceable size, binary-search, or catch the refusal and clamp. Single root fix.
2. Core: decode-time `set_spot_prices` failure should not permanently bench a pool; one-direction failures in the permutation loop should be partial; the discarded-result post-trade refresh should tolerate venue refusal quietly.
3. Observability: pools skipped for unloaded tokens and updates dropped for missing state (`MissingState`) log only at debug; and a range test with `skip_simulation: false` passes even when zero simulations ran.

Diagnosis note for future readers: trace `balanceOf` frames cannot distinguish clean from post-trade state — the balance overwrites pin vault balances to indexed values on every call; the engine's internal corridor storage is what differs.

protocols/substreams/ethereum-fermiswap/src/modules.rs
```diff
@@ -500,6 +501,6 @@ fn map_protocol_changes(
             .sorted_unstable_by_key(|(index, _)| *index)
             .filter_map(|(_, builder)| builder.build())
             .collect::<Vec<_>>(),
-        storage_changes: vec![],
+        storage_changes: get_block_storage_changes(&block),
```
zizou0x, line 506: Can you add docstring explaining that it's not used directly but might be use if we manually add entry points later

zach030 (approved):

zach030:
Just left a small note: we could add a `block_number` check before the main engine-address handling so we bail out early. That way we skip the unnecessary block log scans and speed up indexing a bit.

PR #1229, opened 2026-07-23
fix(ethereum-balancer-v2): track dynamic admin swap fee changes

dvansari65:
Fixes an issue where Balancer V2 pool swap fees were exported as static attributes during creation and never updated when the pool admin changed the fee dynamically.

What was changed

Added BasePool ABI:  Added base_pool.json containing the SwapFeePercentageChanged event to accurately decode fee updates directly from the pool contracts (since they are not emitted by the Vault).

Converted fee to dynamic attribute:  Modified map_protocol_changes to extract the initial "fee" from static_attributes during pool creation and emit it immediately as a dynamic attribute (ChangeType::Creation).

Added event handler:  Added a log listener in map_protocol_changes that tracks SwapFeePercentageChanged events emitted by recognized pools and emits ChangeType::Update for the "fee" attribute to ensure downstream indexers receive the live, on-chain swap fee.


zizou0x (commented):
Thanks for your contribution! Looks great! Could you bump the version (both of the crate and substreams) and add a small description of what was done in a changelog file (similar to what we did for Curve https://github.com/propeller-heads/tycho/blob/main/protocols/substreams/ethereum-curve/CHANGELOG.md)

dvansari65:
@zizou0x  take a look

protocols/substreams/ethereum-balancer-v2/src/modules.rs
```diff
@@ -235,15 +235,64 @@ pub fn map_protocol_changes(
...
-                    let entity_change = EntityChanges {
-                        component_id: component.id.clone(),
-                        attributes: default_attributes.clone(),
-                    };
+                    let mut component = component.clone();
+                    let fee_idx = component
+                        .static_att
+                        .iter()
+                        .position(|a| a.name == "fee");
+                    let fee_value = fee_idx.map(|idx| component.static_att.remove(idx).value);
```
zach030: Also need to remove this fee attribute from the test cases in `integration_test.tycho.yaml`.

zach030 (commented):
Thanks for the contribution! I think it works, but the code smell here feels a bit off, we add the fee to `static_attributes`, then remove it and emit it as a dynamic attribute. I get that it's due to the architecture limitation.

Could you look into whether `SwapFeePercentageChanged` is emitted in the pool creation tx? If it is, we could just rely on the event in `map_protocol_changes` and skip adding it to `static_attributes` altogether.

dvansari65:
@zizou0x  conflicts resolved , check it !!

dvansari65:
@zach030  can you review it!

zach030:
> @zach030 can you review it!

Thanks! Could you take a deeper look at this [comment](https://github.com/propeller-heads/tycho/pull/1229#pullrequestreview-4769355091)? If it's feasible, I'd like to improve the current implementation.

dvansari65:
> > @zach030 can you review it!
> 
> Thanks! Could you take a deeper look at this [comment](https://github.com/propeller-heads/tycho/pull/1229#pullrequestreview-4769355091)? If it's feasible, I'd like to improve the current implementation.

Sorry , I missed it

dvansari65:
> Thanks for the contribution! I think it works, but the code smell here feels a bit off, we add the fee to `static_attributes`, then remove it and emit it as a dynamic attribute. I get that it's due to the architecture limitation.
> 
> Could you look into whether `SwapFeePercentageChanged` is emitted in the pool creation tx? If it is, we could just rely on the event in `map_protocol_changes` and skip adding it to `static_attributes` altogether.

SwapFeePercentageChanged is emitted from pool creation tx successfully , so i removed static attr from pool factory and relying on event in map_protocol_changes

zach030:
> > Thanks for the contribution! I think it works, but the code smell here feels a bit off, we add the fee to `static_attributes`, then remove it and emit it as a dynamic attribute. I get that it's due to the architecture limitation.
> > Could you look into whether `SwapFeePercentageChanged` is emitted in the pool creation tx? If it is, we could just rely on the event in `map_protocol_changes` and skip adding it to `static_attributes` altogether.
> 
> SwapFeePercentageChanged is emitted from pool creation tx successfully , so i removed static attr from pool factory and relying on event in map_protocol_changes

Thanks, that's great! Could you add this to the comments? Also, the CI is failing, could you fix that?

protocols/substreams/ethereum-balancer-v2/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "ethereum-balancer-v2"
-version = "0.4.3"
+version = "0.4.4"
```
die-herdplatte: I think this should be a minor bump, the output format is changed

die-herdplatte (commented):
Thanks for the PR!
I just added one small comment.

zach030 (approved):
Thanks! lgtm!

PR #1231, opened 2026-07-23
fix: guard NG curve solvers against out-of-domain balances

zizou0x:
## What

- `curve/math/core/tricrypto_ng.rs`: ports the Vyper math contract's `get_y` safety asserts (`dev: unsafe values D`, `Unsafe values x[i]`) as a `check_solver_domain` guard at the top of `get_y_3_ng` and `newton_y_3`: `D` must be in `[1e17, 1e33]` and every non-solved balance must satisfy `x[k]*1e18/D ∈ [1e16, 1e20]`. Outside that domain the function returns `None`, matching the deployed contract's revert.
- `curve/math/swap/tricrypto_ng.rs` / `swap/twocrypto_ng.rs`: the `dx`/`desired_output` scaling (`bal[i] += dx`, `xp` products, `dy_native` products) is now checked arithmetic, so oversized amounts return `None` instead of wrapping into plausible-looking balances.

TwoCryptoNG's core solver already has the equivalent guard (`k0_i` vs `lim_mul` bounds); TriCryptoNG was missing it. All stableswap variants use fully checked math and are unaffected.

## Why

Production incident 2026-07-23 (fynd, ethereum): a quote request for ~7.9e28 USDC against a TricryptoNG pool killed every solver worker thread with

```
panicked at ruint-1.18.0/src/algorithms/div/mod.rs:66:17:
attempt to divide by zero
```

Mechanism: with a balance far out of proportion to `D`, the I256 cubic coefficients in `get_y_3_ng` overflow. `alloy_primitives::Signed` ops only `debug_assert!` on overflow, so release builds wrap silently; the garbage values force the `newton_y_3` fallback, where the iterate `y` truncates to zero and `yfprime / y` divides by zero. The deployed Vyper contract cannot reach this state because its `get_y` asserts the domain first — those asserts were not ported.

Reproduced exactly (same crate version, same panic location) by feeding the incident's amount into the vendored math in a release build; with this change the same sweep (dx from 1e18 to 1e40 and the incident amount, three pool shapes) returns `None` with no panics, and in-domain quotes are byte-identical to before.

## Tests

- `core::tricrypto_ng`: `get_y_3_ng`/`newton_y_3` reject out-of-domain balances and out-of-domain `D` (previously: overflow panic in debug, division-by-zero panic in release)
- `swap::tricrypto_ng`: TricryptoUSDC-shaped pool (6/8/18 decimals, real price scales) returns `None` for the incident amount and for `U256::MAX`, still quotes normal amounts
- `swap::twocrypto_ng`: same shape of tests as regression coverage

### DIFF
--- a/crates/tycho-simulation/src/evm/protocol/curve/math/swap/tricrypto_ng.rs
+++ b/crates/tycho-simulation/src/evm/protocol/curve/math/swap/tricrypto_ng.rs
@@ -27,12 +27,19 @@ pub fn get_amount_out(
     let wad = U256::from(WAD);
 
     let mut bal = *balances;
-    bal[i] += dx;
+    bal[i] = bal[i].checked_add(dx)?;
 
+    // Checked: an oversized dx must yield None, not wrap into a plausible-looking xp.
     let xp: [U256; 3] = [
-        bal[0] * precisions[0],
-        bal[1] * price_scale[0] * precisions[1] / wad,
-        bal[2] * price_scale[1] * precisions[2] / wad,
+        bal[0].checked_mul(precisions[0])?,
+        bal[1]
+            .checked_mul(price_scale[0])?
+            .checked_mul(precisions[1])? /
+            wad,
+        bal[2]
+            .checked_mul(price_scale[1])?
+            .checked_mul(precisions[2])? /
+            wad,
     ];
 
     // NG uses hybrid cubic+Newton solver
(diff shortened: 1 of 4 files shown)


crates/tycho-simulation/src/evm/protocol/curve/math/swap/tricrypto_ng.rs
```diff
@@ -27,12 +27,19 @@ pub fn get_amount_out(
     let wad = U256::from(WAD);
 
     let mut bal = *balances;
-    bal[i] += dx;
+    bal[i] = bal[i].checked_add(dx)?;
```
kayibal: If we already on removing panics, let's ensure `i < 3` as well here. Do the same for indices into `price_scale` and `precisions`

zizou0x: Good catch, will add a commit for it

zizou0x: Done in 669af6d06: i/j are now validated (`i >= N || j >= N || i == j` → `None`, mirroring the pool contract's `coin index out of range` assert) in `get_amount_out`/`get_amount_in` of both NG wrappers, plus index guards in the core solvers (`newton_y_3`, `get_y_2_ng`/`newton_y_2_ng` — the latter two underflowed on `x[1 - i]`). Covered by `rejects_out_of_range_coin_indices` tests in both swap modules and the core modules.

kayibal (approved):

PR #1232, opened 2026-07-23
fix(indexer): apply pending account balances in PendingDeltas snapshots

### LINKED ISSUE #1230: PendingDeltas snapshot serves stale token balances alongside fresh storage → Balancer V3 settle reverts with BalanceNotSettled
## Summary

`PendingDeltas` serves RPC snapshots (`/v1/contract_state`) in which a contract's **storage** reflects pending (not-yet-committed) blocks while its **token balances** do not. `PendingDeltas::update_vm_state` and `PendingDeltas::get_account` (`crates/tycho-indexer/src/services/deltas_buffer.rs`) apply pending `account_deltas` (contract storage) but **skip pending `account_balances`** (per-account token balances). Any consumer that reads both — storage for one value and the account's token balance for another — receives an internally inconsistent snapshot.

## Background

This surfaced during the **Balancer V3** integration. Balancer V3 uses a **single `Vault`** that holds the funds for every pool. The Vault records the same quantity in two places that simulation depends on simultaneously:

- **storage** — `_reservesOf[token]` (a storage slot), and
- **account token balance** — the ERC20 balance the indexer tracks separately for the Vault account.

At the end of a swap, the Vault's `unlock` runs `settle(token)`, which checks `token.balanceOf(vault) - _reservesOf[token]` against the amount transferred in. A VM simulation therefore needs the Vault's tracked token balance (used to build the `balanceOf` overwrite) and its `_reservesOf` storage slot to be **mutually consistent**. When the snapshot has storage ahead of the token balance, they disagree and `settle` reverts with `BalanceNotSettled()` (`0x20f1d86d`).

## How it was discovered

Running the `vm:balancer_v3` integration test, sell-into-Vault quotes for a low-activity token reverted with `BalanceNotSettled` while the identical swap succeeded on-chain.

Concrete case — pool `0x1bb7928d25ffb9d02625c7f793d1de002a47adca` (ReClamm AAVE/WETH), selling ~22.76 AAVE → WETH:

- On-chain `BatchRouter.querySwapExactIn` returns ~1.099 WETH (works).
- In simulation the swap computes an output, transfers, then the outer `Vault.unlock` reverts `0x20f1d86d`.
- From the revm trace, in `settle`:
  - `_reservesOf[AAVE]` (from the Vault **storage** slot) = **80.798 AAVE**
  - implied `balanceOf(vault)` base (from the tracked **token balance** overwrite) = **80.076 AAVE**
  - gap = **0.722 AAVE**, which is exactly the leftover unsettled delta that trips `BalanceNotSettled`.

The `80.076` value is the Vault's *previous* committed AAVE balance; `80.798` is the *latest* (also what's in storage and on-chain). i.e. the served snapshot's token balance lagged its storage by one update.

## Root cause

In `PendingDeltas::update_vm_state` the buffered-block loop applies storage deltas but not token-balance deltas:

```rust
for entry in guard.get_block_range(None, version)? {
    if let Some(delta) = entry.account_deltas.get(&db_state.address) {
        db_state.apply_delta(delta)?;      // applies pending STORAGE (+ native balance, code)
        change_found = true
    }
    // TODO: currently it is impossible to apply balance changes and state deltas since
    //  we don't know the component_id of the contract.
}
```

`PendingDeltas::get_account` has the same gap. The `BlockAggregatedChanges` entries do carry `account_balances: HashMap<Address, HashMap<Address, AccountBalance>>` (keyed by **contract address → token → balance**), so applying them here is straightforward. The existing TODO ("can't apply — we don't know the `component_id`") is only true for **component** balances (keyed by `component_id`); **account** balances are keyed by contract address (`db_state.address`), which is exactly what we have.

Meanwhile the RPC path (`get_contract_state_inner` → `PendingDeltas::update_vm_states`) returns the DB account with pending **storage** merged in, so the storage is fresh but the token balances are committed-only → inconsistent when a pending block changed that token's reserve.

## Impact

- Consumers that compare an ERC20 `balanceOf` overwrite (derived from the account's token balances) against a value read from storage get wrong results whenever they fetch a snapshot during the pending window after that token's reserve changed. For Balancer V3 this is a `BalanceNotSettled` revert on the **sell-into-Vault** direction (the side `settle` validates).
- **Directional**: only the token being sent into the Vault is settle-checked, so only that direction fails.
- **Persistent per subscription, not fleeting**: once a subscription loads an inconsistent snapshot, its in-memory state stays wrong until a WS delta resyncs that token's balance — which only happens when the token **next trades**. For a low-activity token (e.g. AAVE) that can be the entire run, so it errors on every block. High-activity tokens (e.g. WETH) resync on the next block and appear to "self-heal", which is why the bug is easy to miss.
- Not specific to Balancer V3 or to any pool math — it's a general snapshot-consistency gap that Balancer V3's single-Vault-with-both-balance-and-storage design happens to expose.

## Proposed fix

Apply pending `account_balances` (keyed by contract address) into the served account's `token_balances` in both `PendingDeltas::update_vm_state` and `PendingDeltas::get_account`, alongside the existing `account_deltas` application. Component balances still cannot be applied here (they need a `component_id`), so that part of the TODO remains.

## Reproduction

1. Sync `vm:balancer_v3` on a local stack.
2. Subscribe to a pool holding a low-activity token (e.g. `0x1bb7928d25ffb9d02625c7f793d1de002a47adca`, AAVE/WETH) during the pending window after that token's reserve changed.
3. Quote a sell of the low-activity token into the Vault (e.g. ~22 AAVE → WETH). It reverts `BalanceNotSettled` (`0x20f1d86d`) while the same `querySwapExactIn` succeeds on-chain.

zach030:
## What

`PendingDeltas::update_vm_state` and `PendingDeltas::get_account` applied pending `account_deltas` (contract **storage**) but not pending `account_balances` (per-account **token balances**). Snapshots served via `/v1/contract_state` during the pending window therefore returned a contract whose storage was fresher than its token balances.

This PR applies the pending `account_balances` (keyed by contract address) alongside `account_deltas` in both methods, so a served account's `token_balances` stay consistent with its storage.

## Why

Balancer V3 uses a single `Vault` that records reserves in two places simulation depends on together: the `_reservesOf[token]` **storage** slot, and the account's tracked **token balance**. The Vault's `settle` checks `token.balanceOf(vault)` against `_reservesOf[token]`. In a VM simulation `_reservesOf` comes from the tracked storage while `balanceOf(vault)` is an overwrite built from the tracked account balance.

When a snapshot carried the storage update but not the balance update, the two disagreed and `settle` reverted with `BalanceNotSettled` (`0x20f1d86d`) on the sell-into-Vault direction. It only bit tokens whose reserve changed in a pending block at subscribe time and that traded infrequently afterward (WS deltas resync active tokens; low-activity tokens like AAVE stayed wrong for the whole subscription). The identical swap succeeded on-chain.

Concrete case — pool `0x1bb7928d25ffb9d02625c7f793d1de002a47adca` (AAVE/WETH), selling ~22.76 AAVE → WETH: from the revm trace, `_reservesOf[AAVE]` (storage) = 80.798 while the `balanceOf(vault)` base (token balance) = 80.076; the 0.722 gap is exactly the leftover unsettled delta.

## Scope note

Only **account** balances are applied here — they are keyed by contract address (`db_state.address`), which these methods already have. **Component** balances remain unapplied because they are keyed by `component_id`, which cannot be resolved from a contract address in this context (the existing TODO's actual blocker); the native path (`update_native_state`) already applies component balances where the `component_id` is known.

## Testing

- Updated `test_update_vm_states` to assert the pending vault token balance is now reflected.
- `cargo test -p tycho-indexer --lib services::deltas_buffer` — 10/10 pass; `cargo +nightly fmt --check` and `cargo clippy -p tycho-indexer --lib` clean.

Fixes #1230


crates/tycho-indexer/src/services/deltas_buffer.rs
```diff
@@ -215,8 +215,24 @@ impl PendingDeltas {
...
+            {
+                for (token, balance) in token_balances {
+                    db_state
+                        .token_balances
+                        .insert(token.clone(), balance.clone());
+                }
+                change_found = true
+            }
+
+            // NOTE: component balances still can't be applied here — they are keyed by
```
die-herdplatte: Should this remain a TODO?

crates/tycho-indexer/src/services/deltas_buffer.rs
die-herdplatte, line 885: From what I can tell the new code in `get_account` isn't tested in here.
`usdc` remains in `missing_addresses` but doesn't have an entry in `account_balances` created in `vm_block_deltas`.

die-herdplatte (approved):

crates/tycho-indexer/src/services/deltas_buffer.rs
zach030: I don't think so. Component balances and state deltas are component_id-keyed (not resolvable from a contract address, and not part of the Account model anyway); they're already applied in `update_native_state`, where the component_id is known.

crates/tycho-indexer/src/services/deltas_buffer.rs
zach030: added a test for this.

PR #1236, opened 2026-07-24
feat(rfq): add Biconomy PropAMM RFQ protocol adapter

fichiokaku:
## What

RFQ protocol adapter for **PropAMM** (`rfq:propamm`) - streaming maker liquidity on Base by Biconomy. Makers sign price ladders (levels of cumulative size + price) off-chain; settlement enforces same-block freshness, per-block volume caps, and on-chain level-sweep pricing. This PR adds the simulation-side integration, patterned on the Bebop adapter.

Opening as a **draft** per your guidance to integrate via PR - happy to reshape anything to fit house conventions.

## Contents (crates/tycho-simulation, +1,957 lines)

- `rfq/protocols/propamm/`: `client` + `client_builder` (RFQClient impl), `models`, `state` (ProtocolSim + IndicativelyPriced), `decoder` (TryFromWithBlock), fixtures
- One deviation from Bebop, by necessity: our feed is a REST levels endpoint, so `stream()` is polling-based (default 1s, configurable), emitting the same absolute `StateSyncMessage<TimestampHeader>` snapshots per cycle. One `ProtocolComponent` per directed pair.
- `get_amount_out` is exact-integer (BigUint, no floats): marginal-merge across makers, best-price-first, then the settlement contract's exact floor chain per maker - so simulated output matches on-chain delivery to the wei (tests pin the boundary wei-loss vectors).
- `request_binding_quote()` hits our firm-quote endpoint: firm `amount_out` (delivered exactly or the tx reverts, no partial fills), with `quote_id`, ordered `calls` JSON, `valid_until`, `gas_estimate` packed into `quote_attributes`. Quotes are stateless: the encoder should fetch immediately before encoding and never replay - superseded call lists revert on-chain rather than filling stale.

## Checks

- `cargo check -p tycho-simulation` green; `cargo test -p tycho-simulation propamm` 35/35; `cargo +nightly-2026-06-28 fmt --check` clean (your CI toolchain)

## Live surface for validation

Base Sepolia staging is fully live: `GET /v1/levels` + `GET /v1/firm-quote` at `https://propamm-staging.biconomy.io` (e.g. levels: `?chainId=84532&tokenIn=0x8b414aD7005EeFd315aF2A16538885Eae229bab7&tokenOut=0xAbbdbbbd6d56593A9c5656c06cB30D61E4a544Df`). Base mainnet contracts are deployed and source-verified at the same addresses; production serving completes with our external audit. On-chain settlement target is a single venue call (`PropAMMAdapter.swap`, `0x1123da3Cf775EE932A83f3d3b9eDBe2e151F79B7` on both Base networks).

## Known gaps / follow-ups (flagged in code)

- `DEFAULT_PROPAMM_API_URL` is a placeholder (`PROPAMM_API_URL` env override works today) - we'll pin the production URL before undrafting
- tycho-execution SwapEncoder + executor contract: in progress as the follow-up commit (encoder consumes `quote_attributes.calls`; executor wraps our adapter with a fixed target)
- TVL normalization (`component_tvl: None` for now) and maker-expiry filtering at poll time

## Two questions for maintainers

1. Beyond the contributing guidelines + review, is there any process for an RFQ integration to become enabled for solvers?
2. For provider streams/APIs consumed by solver-run clients, how do you prefer auth handled (we saw the `ws_key` pattern in the Bebop client - our REST endpoints are currently unauthenticated)?


fichiokaku:
Hey @zach — thanks for the review! All three points addressed, and the big update since the draft: the protocol is now fully live in production with real liquidity, so this integration is testable end to end.

1. **Naming**: renamed throughout to `biconomy_propamm` (registry key `rfq:biconomy_propamm`, module paths, env vars `BICONOMY_PROPAMM_API_URL` / `BICONOMY_PROPAMM_API_KEY`).
2. **Call pattern**: unchanged per your confirmation — levels feed for local simulation, firm-quote once at encoding time.
3. **Testing**: the client now defaults to the production endpoint (`https://propamm.biconomy.io`, Base + BSC) and sends `x-api-key` auth. Added an ignored live integration test you can run against prod:
```
BICONOMY_PROPAMM_API_KEY=<key> cargo test -p tycho-simulation live_levels_and_firm_quote -- --ignored --nocapture
```
Just ran it against production: levels stream OK, firm quote OK (returns `calls`, `valid_until`, `gas_estimate`). Also wired the client into `examples/rfq_quickstart` alongside the other RFQ clients — happy to issue you an API key for testing, ping us.

Branch is rebased on latest main. Live pairs today: Base WETH/USDC, cbBTC/USDC, cbBTC/WETH; BSC WBNB/USDT, BTCB/USDT — with settlement contracts audited scope in prep and real maker liquidity behind every pair.

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/client.rs
```diff
@@ -0,0 +1,685 @@
...
+                        })?
+                        .as_secs(),
+                };
+
+                yield Ok(("propamm".to_string(), StateSyncMessage {
+                    header: TimestampHeader { timestamp },
+                    snapshots: Snapshot { states: new_components, vm_storage: HashMap::new() },
+                    deltas: None, // Deltas are always None - all the changes are absolute
+                    removed_components,
+                }));
```
zach030: The protocol name should be `biconomy_propamm` here.

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/client.rs
```diff
@@ -0,0 +1,685 @@
...
+/// Validates that PropAMM supports the chain and returns its numeric chain id.
+///
+/// The PropAMM API also serves Base Sepolia (chain id 84532), but tycho-common has no built-in
+/// testnet chain, so this integration is mainnet Base only.
+fn validate_chain(chain: Chain) -> Result<u64, RFQError> {
+    match chain {
+        Chain::Base | Chain::Bsc => Ok(chain.id()),
+        _ => Err(RFQError::FatalError(format!("Unsupported chain for PropAMM: {chain:?}"))),
+    }
+}
```
zach030: The docstring is outdated, it now should support both Base and Bsc.

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/models.rs
```diff
@@ -0,0 +1,264 @@
...
+    #[serde(rename = "inventoryContract")]
+    pub inventory_contract: Bytes,
+    /// Cumulative levels: each level's `size` is the total tokenIn wei tradable up to and
+    /// including that level's `price`.
+    pub levels: Vec<PropAmmLevel>,
+    /// Maker nonce as a decimal string.
+    pub nonce: String,
+    /// Unix timestamp (seconds) after which this ladder is no longer valid.
+    #[serde(rename = "expiresAt")]
+    pub expires_at: u64,
```
zach030: `expires_at` isn't used. Is it useful, or could we remove it?

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/client.rs
```diff
@@ -0,0 +1,685 @@
...
+    chain_id: u64,
+    levels_endpoint: String,
+    firm_quote_endpoint: String,
+    // Directed (token_in, token_out) pairs to poll levels for. PropAMM ladders are
+    // one-directional; poll the reverse pair separately if both directions are needed.
+    pairs: Vec<(Bytes, Bytes)>,
+    poll_interval: Duration,
+    quote_timeout: Duration,
+    // Hosted-API key, sent as `x-api-key` on every request. Issued by the Biconomy team.
+    api_key: Option<String>,
```
zach030, line 78: I'd recommend adding `#[serde(skip_serializing, default)]` for `api_key` to keep it safe when dumping the request.

crates/tycho-execution/contracts/src/executors/PropAMMExecutor.sol
```diff
@@ -0,0 +1,168 @@
...
+            address tokenOut,
+            bytes memory commitData,
+            IPropAMMAdapter.FillLeg[] memory legs
+        ) = _decodeData(data);
+
+        // Exact approval for the adapter's transferFrom. Under the router's
+        // delegatecall this approves router funds; the TransferManager's
+        // ProtocolWillDebit approval is equivalent, but the explicit approve
+        // keeps the executor usable standalone.
+        IERC20(tokenIn).forceApprove(propammAdapter, amountIn);
```
zach030: The approve here is redundant, it's already been approved in [TransferManager](https://github.com/propeller-heads/tycho/blob/2958893649cb65d3d4b0ff0c696356d4459b446a/crates/tycho-execution/contracts/src/TransferManager.sol#L151-L163) if the transfer type is marked as `ProtocolWillDebit`

crates/tycho-execution/src/encoding/evm/swap_encoder/biconomy_propamm.rs
```diff
@@ -0,0 +1,502 @@
...
+        _config: Option<HashMap<String, String>>,
+    ) -> Result<Self, EncodingError> {
+        // The PropAMM adapter also exists on Base Sepolia, but tycho-common
+        // has no built-in testnet chain, so this integration is Base only
+        // (mirroring the tycho-simulation PropAMM client).
+        if chain != Chain::Base {
+            return Err(EncodingError::FatalError(
+                "PropAMM swaps are only supported on Base".to_string(),
+            ));
+        }
```
zach030: Should be Base and BSC?

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/client.rs
```diff
@@ -0,0 +1,685 @@
...
+    }
+}
+
+/// Client for the PropAMM RFQ API.
+///
+/// Unlike Bebop's push-based WebSocket feed, PropAMM exposes a plain HTTP levels endpoint, so
+/// `stream()` polls `/v1/levels` for each configured pair at a fixed interval and converts every
+/// poll into the same absolute `StateSyncMessage` snapshot shape Bebop emits from its WS.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct PropAmmClient {
```
zach030: Could you rename this to `BiconomyClient`?

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/state.rs
```diff
@@ -0,0 +1,586 @@
...
+///
+/// `base_token` is the pair's tokenIn and `quote_token` its tokenOut. The levels feed is
+/// one-directional: simulating the reverse direction requires the reverse pair's own component.
+#[derive(Clone, Serialize, Deserialize)]
+pub struct PropAmmState {
+    pub base_token: Token,
+    pub quote_token: Token,
+    pub levels: PropAmmLevelsResponse,
+    pub client: PropAmmClient,
+}
```
zach030: rename this to `BiconomyState`.

crates/tycho-execution/src/encoding/evm/swap_encoder/biconomy_propamm.rs
```diff
@@ -0,0 +1,502 @@
...
+///
+/// # Fields
+/// * `executor_address` - The address of the executor contract that will perform the swap.
+#[derive(Clone)]
+pub struct PropAMMSwapEncoder {
+    executor_address: Bytes,
+    runtime_handle: Handle,
+    #[allow(dead_code)]
+    runtime: SafeRuntime,
+}
```
zach030, line 129: rename to `BiconomySwapEncoder`

crates/tycho-execution/contracts/src/executors/PropAMMExecutor.sol
```diff
@@ -0,0 +1,168 @@
...
+///      before encoding (quotes are valid for seconds; superseded ladders
+///      revert on-chain rather than filling stale) and re-shapes the quote's
+///      settlement calls into abi.encode(tokenIn, tokenOut, commitData,
+///      legs). The only external target is the immutable PropAMM adapter,
+///      called through a typed interface - no raw calldata forwarding, so
+///      there is no selector spoofing surface.
+/// @dev The adapter pulls exactly amountIn from msg.sender and reverts if
+///      the leg sum does not match, so partial fills are not supported:
+///      amountIn must equal the quoted amount.
+contract PropAMMExecutor is IExecutor {
```
zach030, line 19: Please rename all the relevant names in the contract to a Biconomy prefix, e.g. `BiconomyExecutor`, `BiconomyAdapter`, to avoid ambiguity.

crates/tycho-execution/contracts/src/executors/PropAMMExecutor.sol
zach030: We need to add biconomy to `PROTOCOLS_NEEDING_APPROVAL`  [list](https://github.com/propeller-heads/tycho/blob/2958893649cb65d3d4b0ff0c696356d4459b446a/crates/tycho-execution/src/encoding/evm/gas_estimator.rs#L36-L48).

crates/tycho-execution/src/encoding/evm/swap_encoder/biconomy_propamm.rs
```diff
@@ -0,0 +1,502 @@
...
+            .first()
+            .ok_or(EncodingError::FatalError(
+                "PropAMM firm quote contains no calls".to_string(),
+            ))?;
+        if is_fill_call(&commit.data) {
+            return Err(EncodingError::FatalError(
+                "The first PropAMM firm quote call must be the price commit, not a fill"
+                    .to_string(),
+            ));
+        }
```
zach030: Is it necessary to check whether call.to == biconomyAdapter? If the API gets updated, we can reject the rest of the logic.

crates/tycho-execution/src/encoding/evm/swap_encoder/biconomy_propamm.rs
```diff
@@ -0,0 +1,502 @@
...
+        // Re-derive the typed fill legs from the fillFromAnchor calls. The
+        // remaining calls (the ERC20 input transfer in particular) are covered
+        // by the router's TransferManager and the executor's approval, so they
+        // are skipped here.
+        let mut legs: Vec<FillLeg> = Vec::new();
+        let mut legs_total = U256::ZERO;
+        for call in &calls[1..] {
+            if !is_fill_call(&call.data) {
+                continue;
+            }
```
zach030: Shouldn't this be an error instead of continue?

zach030 (commented):
Thanks! Looks good overall. 

I've left some comments:
- A few names still need updating: we should rename them from propamm to biconomy consistently. 
- We'll also need a few more tests, like e2e integration tests for TychoRouter, you could use the Metric [PR](https://github.com/propeller-heads/tycho/pull/1000) as a reference.
- Also, please run clippy and fmt locally before committing.

fichiokaku:
Thanks for the thorough review @zach030! All comments addressed in 23315f933:

**Renames** — `BiconomyClient`, `BiconomyState`, `BiconomySwapEncoder`, `BiconomyExecutor` / `IBiconomyAdapter` (files renamed too), plus the module's model types for consistency. Stream key is now `biconomy_propamm`, and the outdated Base-only docstrings/gates now read Base + BSC (encoder chain gate included).

**`api_key`** — now `#[serde(skip_serializing, default)]` as suggested.

**`expires_at`** — removed. It was advisory only (snapshots refresh every poll, and firm quotes carry their own hard `valid_until` which we already enforce), so nothing used it.

**Redundant approve** — removed from the executor; you're right that the TransferManager's `ProtocolWillDebit` path covers it. `rfq:biconomy_propamm` added to `PROTOCOLS_NEEDING_APPROVAL` so gas estimation charges for that approval.

**Strict call validation (your two encoder comments)** — you were right to push on this, and it surfaced something bigger: the encoder was written against a decomposed call-list shape, but the production API returns exactly ONE settlement `swap()` call whose *step list* carries the price commit, input transfer and fill legs. The encoder now decodes that call (selector-checked) and walks the steps strictly: an unknown step selector, a missing/duplicated commit, a fill before the commit, a fill targeting a different contract than the commit, or a fill on an intermediate pair all refuse to encode — nothing is silently skipped anymore. While fixing this we also aligned the struct layouts with the deployed adapter (`PriceLadder` carries `provider`; `FillLeg` is `(ladder, amountIn)`). Verified against live production quotes across single-fill, split, and BSC routes — the step vocabulary matches exactly, and the live test passes.

Unit tests rebuilt around realistic settlement calldata with cast-cross-checked encodings and selector pins; Solidity suite updated and green (8/8).

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/client.rs
```diff
@@ -0,0 +1,701 @@
...
+
+        Box::pin(async_stream::stream! {
+            let mut current_components: HashMap<String, ComponentWithState> = HashMap::new();
+            let mut ticker = interval(client.poll_interval);
+
+            info!(
+                "Starting PropAMM polling every {} ms for {} pairs",
+                client.poll_interval.as_millis(),
+                client.pairs.len()
+            );
```
zach030, line 281: Most logs in this file still print PropAMM. Could you rename them all?

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/state.rs
```diff
@@ -0,0 +1,590 @@
...
+                "PropAMM levels are one-directional ({} -> {}). Got in={}, out={}; use the \
+                 reverse pair's component for the other direction",
+                self.base_token.address,
+                self.quote_token.address,
+                token_in.address,
+                token_out.address
+            )))
+        }
+    }
+}
```
zach030, line 75: So each pair only supports swapping in one direction? In Biconomy, are `(WETH, USDC)` and `(USDC, WETH)` treated as two separate pairs?

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/client.rs
```diff
@@ -0,0 +1,701 @@
...
+                        }
+                    };
+
+                    if levels.chain_id != client.chain_id {
+                        warn!(
+                            "PropAMM levels chain id mismatch: expected {}, got {}. Skipping.",
+                            client.chain_id, levels.chain_id
+                        );
+                        continue;
+                    }
```
zach030: Would this case actually happen? Did you need to add chain_id as a query param?

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/client.rs
```diff
@@ -0,0 +1,701 @@
...
+                for (token_in, token_out) in &client.pairs {
+                    let levels = match client.fetch_levels(token_in, token_out).await {
+                        Ok(levels) => levels,
+                        Err(e) => {
+                            warn!(
+                                "Failed to fetch PropAMM levels for {token_in} -> {token_out}: {e}"
+                            );
+                            continue;
+                        }
+                    };
```
zach030: Based on the current implementation I'd say WS is more suitable here, right now we poll the pairs sequentially, and a failed fetch is indistinguishable from an empty book, so we'd have to build timeout/backoff handling around that.

Can one socket subscribe to multiple pairs?

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/client.rs
```diff
@@ -0,0 +1,701 @@
...
+                .into(),
+        );
+
+        ComponentWithState {
+            state: ProtocolComponentState::new(&component_id, attributes, HashMap::new()),
+            component: protocol_component,
+            // PropAMM levels carry no USD normalization data, so no TVL is reported.
+            component_tvl: None,
+            entrypoints: vec![],
+        }
```
zach030, line 216: Is there a TVL field in the response? Ideally we'd have a TVL threshold filter.

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/state.rs
```diff
@@ -0,0 +1,590 @@
...
+        } else if base.address == self.quote_token.address
+            && quote.address == self.base_token.address
+        {
+            Ok(1.0 / price)
+        } else {
+            Err(SimulationError::RecoverableError(format!(
+                "Invalid token addresses: {}, {}",
+                base.address, quote.address
+            )))
+        }
```
zach030, line 223: If each pair only supports one direction, does it make sense to return the reverse price?

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/state.rs
```diff
@@ -0,0 +1,590 @@
...
+    size: BigUint,
+}
+
+/// Expands per-maker cumulative ladders into marginal segments, sorted best-price-first.
+///
+/// Levels with non-increasing cumulative sizes are skipped defensively. Ties in price keep the
+/// makers' original order (stable sort), so consumption is deterministic.
+fn expand_marginal_segments(
+    makers: &[BiconomyMakerLevels],
+) -> Result<Vec<MarginalSegment>, SimulationError> {
```
zach030, line 92: This function gets called three times in `get_limits`, could you simplify this?

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/state.rs
```diff
@@ -0,0 +1,590 @@
...
+        if self.levels.makers.is_empty() {
+            return Err(SimulationError::RecoverableError("No liquidity".into()));
+        }
+
+        let sweep = sweep_amount_out(&self.levels.makers, &amount_in)?;
+        let res = GetAmountOutResult {
+            amount: sweep.delivered,
+            gas: BigUint::from(265_000u64), // Gas estimate from PropAMM firm quotes
+            new_state: self.clone_box(),    // The state doesn't change after a swap
+        };
```
zach030, line 245: We can use the `gas_estimate` from firm quote response.

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/client.rs
```diff
@@ -0,0 +1,701 @@
...
+            api_key,
+        })
+    }
+
+    fn apply_auth(&self, request: reqwest::RequestBuilder) -> reqwest::RequestBuilder {
+        match &self.api_key {
+            Some(key) => request.header("x-api-key", key),
+            None => request,
+        }
+    }
```
zach030: Looks like `api_key` is required in all the API calls. Shouldn't we disallow it being missing?

zach030 (commented):
Thanks for the update! I've reviewed the simulation part. Please have a look.

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/client.rs
fichiokaku: fixed

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/client.rs
fichiokaku: fixed

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/models.rs
fichiokaku: removed

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/client.rs
fichiokaku: already there, skip_serializing + default on the field

crates/tycho-execution/contracts/src/executors/PropAMMExecutor.sol
fichiokaku: approve removed. biconomy is already in PROTOCOLS_NEEDING_APPROVAL on main

crates/tycho-execution/src/encoding/evm/swap_encoder/biconomy_propamm.rs
fichiokaku: fixed

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/client.rs
fichiokaku: renamed

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/state.rs
fichiokaku: renamed

crates/tycho-execution/src/encoding/evm/swap_encoder/biconomy_propamm.rs
fichiokaku: renamed

crates/tycho-execution/contracts/src/executors/PropAMMExecutor.sol
fichiokaku: renamed to BiconomyExecutor, adapter references too

crates/tycho-execution/src/encoding/evm/swap_encoder/biconomy_propamm.rs
fichiokaku: kept it. the decoded steps drive approvals and execution targets so we pin them to the expected adapter instead of trusting the api response. if the api ever changes this fails loudly instead of approving something unexpected

crates/tycho-execution/src/encoding/evm/swap_encoder/biconomy_propamm.rs
fichiokaku: changed to an error

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/client.rs
fichiokaku: all renamed

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/state.rs
fichiokaku: yes, each direction is its own signed ladder, so they're separate components. spot_price now rejects the reverse direction too, same as get_limits

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/client.rs
fichiokaku: one host serves multiple chains, chainId picks the book. the batch form uses it too now

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/client.rs
fichiokaku: switched to the batch endpoint instead, one request per poll covers every pair and a failed poll keeps the previous snapshot instead of looking like an empty book. the websocket today is our maker ingest side, a consumer push feed is on the roadmap

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/client.rs
fichiokaku: no tvl field, ladders are signed price commitments. closest analog is merged cumulativeSize, happy to add a min depth filter as a config knob if you want it

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/state.rs
fichiokaku: made it an error, same guidance as get_limits

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/state.rs
fichiokaku: restructured, one expansion serves both bounds

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/state.rs
fichiokaku: the binding path already carries the api's gasEstimate in quote_attributes. the 265k constant is only the routing stage estimate since the levels feed has no gas field, comment says that now

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/client.rs
fichiokaku: now enforced at the call sites. construction stays lenient so clients deserialized from snapshots still build, the key is never serialized

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/models.rs
```diff
@@ -0,0 +1,290 @@
...
+    /// Marginal segment size in tokenIn wei.
+    pub size: String,
+    /// Cumulative size in tokenIn wei up to and including this segment.
+    #[serde(rename = "cumulativeSize")]
+    pub cumulative_size: String,
+}
+
+/// A single maker's ladder for the pair.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct BiconomyMakerLevels {
```
zach030, line 81: `minFill` is in the live maker payload but not modelled, and simulation quotes below it.

Against staging, `USDC->WETH` currently has `minFill = 2000000000` (2000 USDC):

```
sim.get_amount_out(500 USDC)  -> 0.202 WETH        (returns fine)
GET /v1/firm-quote(500 USDC)  -> 400 {"error":"No routes found"}
```

So every swap under 2000 USDC in that direction gets quoted by simulation and then fails at encode time. The other five pairs are at `minFill = 1` right now, but it's a per-maker field. `get_amount_out` should reject below the minimum.

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/client.rs
```diff
@@ -0,0 +1,733 @@
...
+fn validate_chain(chain: Chain) -> Result<u64, RFQError> {
+    match chain {
+        Chain::Base | Chain::Bsc => Ok(chain.id()),
+        _ => Err(RFQError::FatalError(format!("Unsupported chain for Biconomy: {chain:?}"))),
+    }
+}
+
+/// Client for the Biconomy RFQ API.
+///
+/// The API key selects the served lane: firm quotes carry that key's minimum remaining
```
zach030, line 54: This says the levels book is filtered by the key's validity floor, which is the right design - but that floor is directly observable as `expiresAt - asOf`, and `expiresAt` was dropped from the model in 23315f93.

With the CI key it reads a stable ~45s. Without the field, the client can neither see which lane it's on nor notice if the server behaviour ever changes, so this docstring is an assertion nothing can check.

Putting `expiresAt` back gets three things at once: filtering expired ladders at poll time, an observable lane, and a claim that can be verified.

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/state.rs
```diff
@@ -0,0 +1,605 @@
...
+        if self.levels.makers.is_empty() {
+            return Err(SimulationError::RecoverableError("No liquidity".into()));
+        }
+
+        let sweep = sweep_amount_out(&self.levels.makers, &amount_in)?;
+        let res = GetAmountOutResult {
+            amount: sweep.delivered,
+            // Routing-stage estimate. Binding quotes carry the exact per-quote figure from
+            // the API's gasEstimate in quote_attributes["gas_estimate"].
+            gas: BigUint::from(265_000u64),
```
zach030, line 243: Live `gasEstimate` from staging is 1,099,015 - 1,344,949 depending on leg count, so this constant is 4-5x low, and it's the number routing actually compares on.

The comment points at `quote_attributes["gas_estimate"]` for the exact figure, but nothing in the workspace reads that key - it's written and dropped. Worth either feeding the quote value through or moving the constant closer to reality.

crates/tycho-execution/src/encoding/evm/swap_encoder/biconomy_propamm.rs
```diff
@@ -0,0 +1,813 @@
...
+/// `calls` quote attribute.
+#[derive(Debug, Deserialize)]
+struct BiconomyCall {
+    /// Call target: the Biconomy settlement contract. Raw calldata is never
+    /// forwarded on-chain - execution goes through the typed adapter call in
+    /// BiconomyExecutor - but step targets are validated against it.
+    to: Bytes,
+    /// Native value as a decimal string; Biconomy settlement is ERC20-only.
+    #[allow(dead_code)]
+    value: String,
```
zach030, line 99: `value` is never checked, on either the call or the steps.

Everywhere else the walk refuses to guess on an unexpected shape; this is the one silent gap. A quote carrying non-zero native value would encode as if it were pure ERC20. If settlement really is ERC20-only, asserting `value == 0` is a one-liner and keeps the same invariant.

crates/tycho-execution/contracts/src/executors/BiconomyExecutor.sol
```diff
@@ -0,0 +1,161 @@
...
+/// @title BiconomyExecutor
+/// @notice Executor for Biconomy Biconomy streaming-maker RFQ swaps
+///         (rfq:biconomy_propamm)
+/// @dev The Rust swap encoder requests a binding firm quote immediately
+///      before encoding (quotes are valid for seconds; superseded ladders
+///      revert on-chain rather than filling stale) and re-shapes the quote's
+///      settlement calls into abi.encode(tokenIn, tokenOut, commitData,
+///      legs). The only external target is the immutable Biconomy adapter,
+///      called through a typed interface - no raw calldata forwarding, so
+///      there is no selector spoofing surface.
```
zach030, line 53: `commitData` reaches the adapter unchecked on-chain - the executor only length-checks the payload.

The encoder's selector allowlist is off-chain, so anyone calling the router directly skips it. Whether that's actually safe depends on what the adapter does with `commitData` when it forwards to the anchor executor, and the adapter isn't in this repo.

Either validate the commit selector on-chain too, or point us at the adapter code so we can confirm the claim.

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/test_responses/firm_quote.json
```diff
@@ -0,0 +1,21 @@
+{
+    "quoteId": "0x3f8a1b2c4d5e6f708192a3b4c5d6e7f8091a2b3c4d5e6f708192a3b4c5d6e7f8",
+    "gasEstimate": "265000",
+    "chainId": 8453,
+    "amountIn": "15000000000000000000",
+    "amountOut": "28164999999",
+    "receiver": "0xfD0b31d2E955fA55e3fa641Fe90e08b677188d35",
+    "calls": [
```
zach030, line 8: This fixture no longer matches the API. A live staging quote is exactly one call with selector `0x1eada922`; this has two calls with `0xdeadbeef` and `0x0102`.

The encoder tests build their own calldata so nothing fails, but anyone changing the client against this fixture will be working from the wrong shape.

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/models.rs
```diff
@@ -0,0 +1,290 @@
...
+            )));
+        }
+        let amount_in = params.amount_in.to_string();
+        if self.amount_in != amount_in {
+            return Err(RFQError::FatalError(format!(
+                "Amount in mismatch: expected {}, got {}",
+                amount_in, self.amount_in
+            )));
+        }
+        if self.calls.is_empty() {
```
zach030, line 169: Since a firm quote is exactly one settlement call (the encoder already enforces it), `calls.len() != 1` would catch an API shape change here instead of several layers later.

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/models.rs
```diff
@@ -0,0 +1,290 @@
...
+
+use crate::rfq::errors::RFQError;
+
+/// PropAMM level prices are fixed-point integers scaled by 1e18: tokenOut wei per tokenIn wei,
+/// multiplied by 10^18. All level sizes are denominated in tokenIn wei.
+pub fn biconomy_propamm_price_scale() -> BigUint {
+    BigUint::from(10u32).pow(18)
+}
+
+/// Response of `GET {base_url}/v1/levels?chainId&tokenIn&tokenOut`.
```
zach030, line 15: Stale doc: the per-pair endpoint isn't called any more. This struct is now assembled from the chain-batch response, so it's an internal type rather than an API shape.

crates/tycho-simulation/src/rfq/protocols/biconomy_propamm/client.rs
```diff
@@ -0,0 +1,733 @@
...
+            })?;
+
+        if !response.status().is_success() {
+            return Err(RFQError::ConnectionError(format!(
+                "Biconomy levels HTTP error {}: {}",
+                response.status(),
+                response
+                    .text()
+                    .await
+                    .unwrap_or_default()
```
zach030, line 158: `unwrap_or_default()` turns a serialisation failure into an empty attribute, which the decoder then rejects as `Invalid makers JSON` - and with `skip_state_decode_failures` defaulting to false that's fatal for the stream. It can't realistically fail, but propagating beats writing a corrupt attribute.

zach030 (commented):
Re-reviewed at `c7aff11c6`. fmt, clippy and unit tests are all green (sim 35/35, encoder 13/13), and I ran the live test plus some ad-hoc comparisons against staging with the CI key.

**The pricing math checks out against live data.** Fetching levels immediately before each quote, `get_amount_out` matches the firm quote to the wei across the whole range:

| amount in | sim | firm | diff |
|---|---|---|---|
| 0.05 WETH | 123142425 | 123142425 | 0 |
| 0.5 WETH | 1231424256 | 1231424256 | 0 |
| 50% of depth | 53259101116 | 53259101125 | 9 wei |
| 99% of depth | 104683701617 | 104683701635 | 18 wei |
| 100% of depth | 105652989875 | 105652989876 | 1 wei |

The residual is the three-floor chain and it always lands on the safe side (sim <= firm). Batch polling, the strict step walk, the one-directional pricing and the removed approve all look right.

Comments inline. The two data-model ones (`minFill`, `expiresAt`) are the ones I'd want fixed before undraft — both reproduce against staging.

One thing that has no line to anchor to: the executor still isn't wired into `contracts/test/TychoRouterTestSetup.sol`, so there's no router-level test and no cross-language calldata check (`write_calldata_to_file` / `loadCallDataFromFile`). Right now the Rust encoder tests and the Solidity mock-adapter tests never meet in the middle. Bebop/Liquorice both have that layer.

Minor naming leftovers, not worth separate threads: three `PropAMM` mentions remain in `models.rs`, the helper is still `biconomy_propamm_price_scale`, and `BiconomyExecutor.sol` says "Biconomy Biconomy" on lines 7 and 45.

fichiokaku:
Updated: rebased on main (358 commits) and addressed the outstanding review items in one pass.

**Code changes** (c5f9b359d):
- `minFill` per maker and `minQuote` per pair are now modeled: the sweep drops any maker whose allocation lands below its signed minimum and re-runs (mirroring our route construction), and quotes below the pair floor decline the way the API does.
- Routing-stage gas estimate sized to the measured live band (1,099,015–1,344,949 depending on leg count); binding quotes carry the exact per-quote `gasEstimate` as before.
- Encoder rejects settlement calls with non-zero native value (settlement is ERC20-only).
- Executor pins the commit step's selector to `updatePrices` (0x86e97b02) — arbitrary calldata can no longer reach the adapter's commit hop even under a compromised API; covered by a revert test.
- Attribute serialization failures now fail that component instead of storing an empty ladder.
- `firm_quote.json` rebuilt as a real single settlement `swap()` call (selector 0x1eada922) with valid ABI-encoded commit + fill steps and wei-exact three-floor amounts, against our current contract generation (settlement `0x0000004c86FBE08cC6A39c6D4C87b59F71dE31bc`, executor `0x000000Ab52Bdb44411777412fe938776a766cc6a`, live on Base + BSC).
- Restored `propamm.rs` (your generic pAMM fallback executor) and `sky` which the rebase had displaced — no upstream files are modified by this PR anymore.

**Answers to open questions:**
- *One-directional pairs*: yes, `(WETH, USDC)` and `(USDC, WETH)` are separate ladders and separate components; a maker signs each direction independently, so no reverse price is derivable.
- *chain_id query param*: kept — one poll per chain via `/v1/levels?chainId=` batches every direction, which is also why sequential per-pair polling (the WS suggestion) matters less now; happy to move to WS when you'd take that dependency.
- *TVL field*: the levels response has no USD field by design; depth in tokenIn wei is in the ladders, so a threshold filter is possible on token amounts if you want one.
- *call.to == adapter check*: kept — with the selector pin above it's defense in depth, and both checks are cheap.

**Status**: we'd like to keep this PR open as a parked draft. Our maker network now serves RFQ-style demand from the same ladder stream that feeds our onchain venue, and the onchain venue is the integration path we're prioritizing with you (it's a plain IPropAMM venue, so it should ride your existing generic pAMM lane — separate thread coming). This branch stays current so the RFQ lane can be activated whenever it makes sense for both sides.

Tests: tycho-simulation biconomy 37/37, tycho-execution encoder 13/13, Biconomy.t.sol 9/9 (incl. the new selector-guard test), PropAMM.t.sol untouched and passing.

fichiokaku:
One follow-up (0c1af6536): the protocol identifier is now plain `rfq:biconomy`, matching the company-name convention of the other RFQ sources (bebop, hashflow, liquorice). Module dirs, `protocol_type_name` (`biconomy_pool`) and env vars (`BICONOMY_API_URL` / `BICONOMY_API_KEY`) follow. Types were already Biconomy-prefixed from the earlier rename round.

PR #1241, opened 2026-07-27
feat(execution): prefetch Angstrom attestations in the background

tamaralipows:
## Why

Every Angstrom hop made its own blocking API call while encoding, so a multi-hop route made several calls one after another. Measured against the live API: **902ms for the first read, 11µs once the cache is filled.**

Caching is correct because an attestation is valid for one block number and does not depend on the swap, the pool or the route, so <a href="https://docs.angstrom.xyz/l1/core-mechanisms/pool-unlock" target="_blank" rel="noopener noreferrer">one fetch covers every swap</a>. A background thread fetches a window every 6 seconds and stores it in one process-wide cache that encoding reads. Each fetch overwrites the previous window, so the cache always holds exactly one.

If the cache is empty or its window is more than one block old, encoding fetches a window itself rather than failing the route.

## What the transaction carries

`ANGSTROM_BLOCKS_IN_FUTURE=5` returns 6 entries, verified against the live API as `head+0 … head+5`:

```
6 × 93 bytes = 558 bytes of hook data per Angstrom hop
```

Same count as before. What changed: every hop now encodes the same block numbers. Each hop used to fetch on its own, so a block arriving between two fetches left them covering different ranges, and the transaction could only land in a block both of them covered.

## What Fynd needs

**No source changes.** Fynd's workspace compiles against this branch with only a `[patch]` added, and it already builds the registry once in `Solver::build()`.

1. `cargo update -p tycho-execution` once this releases — the version requirement is already `>=0.341.7`.
2. Set `ANGSTROM_API_KEY` before `Solver::build()` runs. It is read once when the registry is built, not on every fetch, so setting it after that does nothing. It is missing from Fynd's `.env` and compose files today.

Fetch failures now return `RecoverableError` instead of `FatalError`. Fynd converts both to `SolveError::FailedEncoding`, so this does not affect it.

### DIFF
--- a/crates/tycho-execution/CLAUDE.md
+++ b/crates/tycho-execution/CLAUDE.md
@@ -233,6 +233,27 @@ from `config/executor_addresses.json`. Protocol name prefixes: `vm:` (simulation
 e.g. `vm:balancer_v2`, `vm:curve`), `rfq:` (request-for-quote, e.g. `rfq:bebop`), bare (on-chain,
 e.g. `uniswap_v2`, `fluid_v1`).
 
+### Angstrom attestations (`evm/swap_encoder/angstrom.rs`)
+
+Angstrom's Uniswap V4 pools start every block locked. A swap against one carries a pool unlock attestation, signed by
+the current Angstrom leader, as its `hookData`. The attestation is scoped to a block number and says nothing about the
+swap, so one fetched window (covering `ANGSTROM_BLOCKS_IN_FUTURE` blocks, default 5) serves every swap, pool and route.
+
+`AttestationCache` therefore keeps the window in a process-wide cache instead of fetching it during encoding:
+
+- A dedicated OS thread (`angstrom-attestations`) refreshes the window every 2s. Not a `tokio` task — the encoder must
+  work without a runtime, and `reqwest`'s blocking client cannot be driven from inside one.
+- `AttestationCache::global()` starts the thread on first call. `UniswapV4SwapEncoder::new` calls it when
+  `angstrom_hook_address` is configured for the chain (`config/protocol_specific_addresses.json`), so registry
+  construction warms the cache. `ANGSTROM_API_KEY` / `ANGSTROM_API_URL` / `ANGSTROM_BLOCKS_IN_FUTURE` are read once, at
+  that point; without the API key the thread never starts and Angstrom swaps fail to encode with a `FatalError`.
+- `encode_swap` reads the cache. A window older than `ANGSTROM_ATTESTATION_MAX_AGE` (24s, ~2 blocks) triggers one
+  inline fetch on a scoped thread, so encoding degrades to the old behavior instead of failing. Fetch failures are
+  `RecoverableError`.
+- On chain, `UniswapV4Executor._selectAttestation` picks the 93-byte entry (8-byte block number + 85-byte attestation)
+  matching `block.number` and returns empty bytes when none match. Entries for blocks that already passed only cost
+  calldata; a window that covers no upcoming block falls back to Angstrom's protocol-driven empty-batch unlock.
+
 ### Gas estimation
 
 `Swap::new(component, token_in: Token, token_out: Token, estimated_gas: BigUint)` carries a per-swap simulation gas
--- a/crates/tycho-execution/CLAUDE.md
+++ b/crates/tycho-execution/CLAUDE.md
@@ -241,15 +241,15 @@ swap, so one fetched window (covering `ANGSTROM_BLOCKS_IN_FUTURE` blocks, defaul
 
 `AttestationCache` therefore keeps the window in a process-wide cache instead of fetching it during encoding:
 
-- A dedicated OS thread (`angstrom-attestations`) refreshes the window every 2s. Not a `tokio` task — the encoder must
-  work without a runtime, and `reqwest`'s blocking client cannot be driven from inside one.
+- A dedicated OS thread (`angstrom-attestations`) refreshes the window every
+  `ANGSTROM_ATTESTATION_REFRESH_INTERVAL`, twice per Ethereum block. Not a `tokio` task — the encoder must work without
+  a runtime, and `reqwest`'s blocking client cannot be driven from inside one.
 - `AttestationCache::global()` starts the thread on first call. `UniswapV4SwapEncoder::new` calls it when
   `angstrom_hook_address` is configured for the chain (`config/protocol_specific_addresses.json`), so registry
   construction warms the cache. `ANGSTROM_API_KEY` / `ANGSTROM_API_URL` / `ANGSTROM_BLOCKS_IN_FUTURE` are read once, at
   that point; without the API key the thread never starts and Angstrom swaps fail to encode with a `FatalError`.
-- `encode_swap` reads the cache. A window older than `ANGSTROM_ATTESTATION_MAX_AGE` (24s, ~2 blocks) triggers one
-  inline fetch on a scoped thread, so encoding degrades to the old behavior instead of failing. Fetch failures are
-  `RecoverableError`.
+- `encode_swap` reads the cache. A window older than `ANGSTROM_ATTESTATION_MAX_AGE` triggers one inline fetch on a
+  scoped thread, so encoding degrades to the old behavior instead of failing. Fetch failures are `RecoverableError`.
 - On chain, `UniswapV4Executor._selectAttestation` picks the 93-byte entry (8-byte block number + 85-byte attestation)
   matching `block.number` and returns empty bytes when none match. Entries for blocks that already passed only cost
   calldata; a window that covers no upcoming block falls back to Angstrom's protocol-driven empty-batch unlock.
--- a/crates/tycho-execution/src/encoding/evm/swap_encoder/angstrom.rs
+++ b/crates/tycho-execution/src/encoding/evm/swap_encoder/angstrom.rs
@@ -30,16 +30,16 @@ use crate::encoding::{
 
 static CACHE: OnceLock<Arc<AttestationCache>> = OnceLock::new();
 
-/// The attestation window every Angstrom swap encodes, kept warm by a background thread.
+/// Fetches attestations for the current block and the next `ANGSTROM_BLOCKS_IN_FUTURE` blocks.
+type WindowFetcher = Box<dyn Fn() -> Result<AttestationResponse, EncodingError> + Send + Sync>;
+
+/// How to fetch an attestation window, or why it cannot be fetched, plus the last one fetched.
 ///
-/// Every refresh replaces the whole window rather than appending to it, so the cache holds one
-/// window of `ANGSTROM_BLOCKS_IN_FUTURE + 1` attestations at a time and does not grow with
-/// uptime.
+/// A refresh replaces the whole window rather than appending to it, so this holds one window of
+/// `ANGSTROM_BLOCKS_IN_FUTURE + 1` attestations and does not grow with uptime.
 pub(crate) struct AttestationCache {
-    /// The configured API client, or the reason Angstrom swaps cannot be encoded.
-    api: Result<ApiConfig, String>,
-    /// The window from the most recent successful fetch, or `None` until the first one lands.
-    latest: RwLock<Option<CachedWindow>>,
+    fetcher: Result<WindowFetcher, String>,
+    window: RwLock<Option<CachedWindow>>,
 }
 
 impl AttestationCache {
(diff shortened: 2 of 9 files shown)


crates/tycho-execution/CLAUDE.md
```diff
@@ -233,6 +233,27 @@ from `config/executor_addresses.json`. Protocol name prefixes: `vm:` (simulation
...
 
+### Angstrom attestations (`evm/swap_encoder/angstrom.rs`)
+
+Angstrom's Uniswap V4 pools start every block locked. A swap against one carries a pool unlock attestation, signed by
+the current Angstrom leader, as its `hookData`. The attestation is scoped to a block number and says nothing about the
+swap, so one fetched window (covering `ANGSTROM_BLOCKS_IN_FUTURE` blocks, default 5) serves every swap, pool and route.
+
+`AttestationCache` therefore keeps the window in a process-wide cache instead of fetching it during encoding:
+
+- A dedicated OS thread (`angstrom-attestations`) refreshes the window every 2s. Not a `tokio` task — the encoder must
```
louise-poole: I think you iterated over some versions here and this doc got a bit outdated - from the code it looks like this is twice a block, not every 2s?

crates/tycho-execution/CLAUDE.md
```diff
@@ -233,6 +233,27 @@ from `config/executor_addresses.json`. Protocol name prefixes: `vm:` (simulation
...
+  work without a runtime, and `reqwest`'s blocking client cannot be driven from inside one.
+- `AttestationCache::global()` starts the thread on first call. `UniswapV4SwapEncoder::new` calls it when
+  `angstrom_hook_address` is configured for the chain (`config/protocol_specific_addresses.json`), so registry
+  construction warms the cache. `ANGSTROM_API_KEY` / `ANGSTROM_API_URL` / `ANGSTROM_BLOCKS_IN_FUTURE` are read once, at
+  that point; without the API key the thread never starts and Angstrom swaps fail to encode with a `FatalError`.
+- `encode_swap` reads the cache. A window older than `ANGSTROM_ATTESTATION_MAX_AGE` (24s, ~2 blocks) triggers one
```
louise-poole: I think remove the numbers given here for the max age so that this does not become outdates when we tweak values (I think it already is outdated?)

docs/for-solvers/supported-protocols.md
```diff
@@ -107,13 +107,20 @@ Interested in adding a protocol? Refer to the [Tycho Simulation for DEXs](../for
...
 
 **Required configuration**:
 
 * Set the `ANGSTROM_API_KEY` environment variable (request one from the Angstrom team directly)
 * Set `ANGSTROM_BLOCKS_IN_FUTURE` environment variable (if you want to override the <a href="https://github.com/propeller-heads/tycho-indexer/blob/main/crates/tycho-execution/src/encoding/evm/constants.rs" target="_blank" rel="noopener noreferrer">default value</a> of 5 blocks). **Important trade-off**: The more blocks you fetch, the more calldata will be sent to the Tycho Router, making execution more gas expensive.
 
 If `ANGSTROM_API_KEY` is not set, `ProtocolStreamBuilder` excludes Angstrom pools from `uniswap_v4_hooks` by default (unless you pass your own filter function), since routes over these pools would fail at encoding without attestations.
+
+**Attestations are prefetched**, so encoding an Angstrom swap makes no API call. A background thread refreshes the attestation window every two seconds and encoding reads the result from a process-wide cache. The thread starts when you build a `SwapEncoderRegistry` for Ethereum with `ANGSTROM_API_KEY` set, and it reads all three environment variables once, at that point.
```
louise-poole: Same here - it's not every 2 seconds

docs/for-solvers/supported-protocols.md
```diff
@@ -107,13 +107,20 @@ Interested in adding a protocol? Refer to the [Tycho Simulation for DEXs](../for
...
+
+Two consequences for your setup:
+
+* **Build the encoder once at startup and reuse it.** Encoding still works if you build a new encoder per quote, but the first Angstrom swap it encodes waits for the first fetch to finish.
+* **Watch your logs for `Angstrom attestation cache is cold or stale`.** When the cached window is older than 24 seconds, encoding fetches a fresh one inline and logs that warning. Encoding then pays the API round trip, so a steady stream of these means the background refresh is failing — the refresh logs its own error alongside them.
```
louise-poole: Here too - I think the check is that it's not older than 12s now?

crates/tycho-execution/src/encoding/evm/swap_encoder/angstrom.rs
```diff
@@ -0,0 +1,475 @@
...
+/// Every attestation takes exactly `8 + ANGSTROM_ATTESTATION_SIZE` bytes: a big endian block
+/// number followed by the attestation itself. Entries for blocks that have already passed are
+/// harmless, since the executor selects the entry matching `block.number`.
+///
+/// Returns an error if an attestation is not `ANGSTROM_ATTESTATION_SIZE` bytes long, which the
+/// executor would reject on chain.
+fn encode_attestations(response: &AttestationResponse) -> Result<Vec<u8>, EncodingError> {
+    let mut encoded =
+        Vec::with_capacity(response.attestations.len() * (8 + ANGSTROM_ATTESTATION_SIZE));
+    for data in &response.attestations {
```
louise-poole, line 254: As-is, if response.attestations is empty we'd skip this loop and return Ok. I think we should have a check for that here and raise an error in that case.

crates/tycho-execution/src/encoding/evm/swap_encoder/angstrom.rs
```diff
@@ -0,0 +1,475 @@
...
+    ///
+    /// Returns the reason Angstrom swaps cannot be encoded when `ANGSTROM_API_KEY` is unset,
+    /// which is how consumers that do not route over Angstrom opt out.
+    fn from_env() -> Result<Self, String> {
+        let key = env::var("ANGSTROM_API_KEY").map_err(|_| {
+            "ANGSTROM_API_KEY environment variable is required for Angstrom swaps".to_string()
+        })?;
+
+        let client = reqwest::blocking::Client::builder()
+            .timeout(ANGSTROM_API_TIMEOUT)
```
louise-poole, line 160: We never used to timeout on this, correct? What happens if this times out? Is 2s lenient enough?

crates/tycho-execution/src/encoding/evm/swap_encoder/angstrom.rs
```diff
@@ -0,0 +1,475 @@
...
+    }
+
+    /// Fetches the current attestation window into the cache and returns it.
+    fn refresh(&self) -> Result<Vec<u8>, EncodingError> {
+        let api = self
+            .api
+            .as_ref()
+            .map_err(|reason| EncodingError::FatalError(reason.clone()))?;
+
+        let encoded = encode_attestations(&api.fetch_off_runtime()?)?;
```
louise-poole: Ran this PR through my claude quick and it told me to tell you:
Make the fetch strategy a parameter of refresh so each call site states its own situation. Both functions are non-capturing, so a plain fn pointer is enough:

```rust
  /// Fetches the current attestation window into the cache and returns it.
  ///
  /// The blocking API client cannot send from a thread with a live `tokio` runtime, so callers
  /// pass the fetch that suits them: `ApiConfig::fetch` when they are already off a runtime,
  /// `ApiConfig::fetch_off_runtime` when they may not be.
  fn refresh(
      &self,
      fetch: fn(&ApiConfig) -> Result<AttestationResponse, EncodingError>,
  ) -> Result<Vec<u8>, EncodingError> {
      let api = self.api.as_ref()
          .map_err(|reason| EncodingError::FatalError(reason.clone()))?;
      let encoded = encode_attestations(&fetch(api)?)?;
      // ...unchanged
  }
```

  `hook_data` then calls `self.refresh(ApiConfig::fetch_off_runtime),` and the refresher loop calls `self.refresh(ApiConfig::fetch)`. Two call sites change, the docstring now says something a reader can verify, and the
  refresher stops paying for an escape hatch it doesn't use.

  If they'd rather not thread a parameter through, two named wrappers (refresh / refresh_off_runtime) over a shared `refresh_with` reads about as well. Either way this is a nit — worth fixing while the code is fresh,
  not worth blocking on.

  One thing that is worth stating in the same breath: the refresher discards refresh's Vec<u8> return value. That's harmless, but it's a hint that refresh is doing two jobs (update the cache, and hand the caller a
  window), and only `hook_data` needs the second.

crates/tycho-execution/src/encoding/evm/swap_encoder/angstrom.rs
```diff
@@ -0,0 +1,475 @@
...
+    }
+
+    #[test]
+    fn test_stale_window_triggers_a_fetch() {
+        let stale = ANGSTROM_ATTESTATION_MAX_AGE + Duration::from_secs(1);
+        let cache = unconfigured_cache(window_fetched_ago(stale));
+
+        let err = cache.hook_data().unwrap_err();
+
+        assert_eq!(err, EncodingError::FatalError("no API key".to_string()));
```
louise-poole: It feels a bit disjoined that a 'no API key' error being triggered is proof of something else's success. Is there maybe not a nicer/cleaner way we can test this?

crates/tycho-execution/src/encoding/evm/swap_encoder/angstrom.rs
louise-poole: Not sure how helpful this is. It doesn't seem too critical and is more about best practices here.

crates/tycho-execution/src/encoding/evm/swap_encoder/angstrom.rs
```diff
@@ -0,0 +1,475 @@
...
+            .read()
+            .unwrap_or_else(PoisonError::into_inner);
+        if let Some(window) = latest.as_ref() {
+            if window.fetched_at.elapsed() <= ANGSTROM_ATTESTATION_MAX_AGE {
+                return Ok(window.encoded.clone());
+            }
+        }
+        drop(latest);
+
+        warn!("Angstrom attestation cache is cold or stale, fetching while encoding");
```
louise-poole, line 84: It looks like we'd get this warning if no angstrom api key is set (as the prefetch would have failed). Should we not check for the key and explicitly fail if it's not available?

crates/tycho-execution/src/encoding/evm/swap_encoder/angstrom.rs
```diff
@@ -0,0 +1,475 @@
...
+            .timeout(ANGSTROM_API_TIMEOUT)
+            .build()
+            .map_err(|e| format!("Failed to build the Angstrom API client: {e}"))?;
+
+        let url =
+            env::var("ANGSTROM_API_URL").unwrap_or_else(|_| ANGSTROM_DEFAULT_API_URL.to_string());
+        let blocks_in_future = env::var("ANGSTROM_BLOCKS_IN_FUTURE")
+            .ok()
+            .and_then(|blocks| blocks.parse().ok())
+            .unwrap_or(ANGSTROM_DEFAULT_BLOCKS_IN_FUTURE);
```
louise-poole: If we fail to unwrap it means the env var exists but was not parsable. Maybe we should error in that case? It might cause unexpected behaviour to just fallback to a default like this.

crates/tycho-execution/src/encoding/evm/swap_encoder/angstrom.rs
tamaralipows: We did time out before, just at 30s — that's reqwest's blocking default. So this changes it from 30s to 3s.

Two callers can hit it. The background refresher logs the failure, leaves the previous window in the cache and retries after its 6s sleep, so encoding never notices. The other is `hook_data()`: when the cache is empty or its window is older than `ANGSTROM_ATTESTATION_MAX_AGE`, it calls the API right there inside encode_swap and waits. That's what the timeout bounds. On timeout it returns a `RecoverableError`, so the whole solution containing that Angstrom hop fails to encode and the caller gets an error instead of calldata for that route.

I bumped this to 3s. The cold read I measured was 902ms including DNS and TLS, so 3s leaves 3.3x headroom.

crates/tycho-execution/src/encoding/evm/swap_encoder/angstrom.rs
tamaralipows: Leaving it. [15a474c](https://github.com/propeller-heads/tycho/pull/1241/commits/15a474c23cf671d43a34121f2e9e9281a235e505)  changed the shape: the cache now holds one fetch (fetch: Result<Fetch, String>), set once in `into_fetch`, so there's no per-call-site choice left to make. It has to be `fetch_off_runtime`, because that same fetch serves both the refresher thread, which has no tokio runtime, and hook_data, which usually runs inside one.

The cost is one scoped thread spawn every 6s on a thread that doesn't need the hop. Cheaper than two fetch paths to keep straight.

On refresh doing two jobs: hook_data needs the returned window, the refresher doesn't. Splitting them makes hook_data re-take the read lock it dropped a line earlier - same work, extra round trip.

🤖

crates/tycho-execution/src/encoding/evm/swap_encoder/angstrom.rs
tamaralipows: Please let me know how you feel about [15a474c](https://github.com/propeller-heads/tycho/pull/1241/commits/15a474c23cf671d43a34121f2e9e9281a235e505)

louise-poole (approved):
Looks good!!! Thanks T!

PR #1242, opened 2026-07-28
fix(ekubo_v3): exclude SignedExclusiveSwap pools by default

louise-poole:
Ekubo's default `filter_fn` now drops `SignedExclusiveSwap` pools, which need a per-swap signature passed to the encoder as user_data. Consumers that can supply that signature can opt in with `filter_fn_with_signed_exclusive_swap` (re-exported as `ekubo_v3_extension_filter_with_signed_exclusive_swap`).


tamaralipows (approved):
Amazing, what a beautiful change

PR #1244, opened 2026-07-28
feat: integrate native RFQ protocol

dvansari65:
### Protocol Scaffold

- Added the `rfq:native` protocol module:
  - `client.rs`
  - `client_builder.rs`
  - `models.rs`
  - `decoder.rs`
  - `state.rs`
  - `mod.rs`
- Added authentication through the `NATIVE_API_KEY` environment variable.
- Registered Native in the RFQ protocol and swap-encoder registries.

### Orderbook Indexing

`NativeClient::stream()`:

- Polls Native Relay's orderbook endpoint.
- Groups bid and ask entries into one `NativePriceData` per token pair.
- Supports both two-sided and one-sided orderbooks.
- Emits complete `Snapshot` messages.
- Emits removed components when markets disappear or no longer pass configured filters.

### Component and State Modeling

- Protocol system: `rfq:native`
- Protocol type: `native_relay_pool`
- Stores the complete orderbook in the dynamic `book` attribute.
- Decodes indexed components into `NativeState`.
- Supports spot-price calculation, orderbook simulation, liquidity limits, and minimum trade amounts.

### TVL Filtering

- Calculates TVL from available orderbook levels.
- Averages bid-side and ask-side quote value when both sides are available.
- Uses the available side for one-sided orderbooks.
- Normalizes markets quoted in non-approved tokens through an available approved quote-token market.
- Skips markets when no normalization path exists.

### Firm Quotes

- Requests binding quotes from Native Relay's `/firm-quote` endpoint.
- Validates API success, returned orders, token addresses, quote expiry, and output amount.
- Stores the following `txRequest` fields in the signed quote:
  - `target`
  - `calldata`
  - `value`
- Preserves `txRequest.value` as decimal UTF-8 bytes for payable native-input swaps.

### Swap Encoding

- Added `NativeSwapEncoder` for `rfq:native`.
- Encodes token addresses, execution target, payable value, and Native calldata.
- Rejects missing, malformed, binary, or overflowing quote values instead of silently encoding zero.
- Supports ERC-20-input and native-input swaps.

### Native Executor

- Added `NativeExecutor.sol`.
- Restricts execution to the configured:
  - Native Router V4
  - Native Router V3, when available
  - CreditVault
- Validates the Native calldata selector.
- Supports chains such as Monad where Router V3 is unavailable.
- Forwards the required payable value for native-input swaps.
- Propagates failed Native calls as transaction reverts.

### Configuration

Added Native executor configuration for:

- Ethereum
- Base
- Arbitrum
- BNB Smart Chain
- Monad

Also added the corresponding Router V4, Router V3, and CreditVault addresses. Monad uses a zero Router V3 address because Native does not deploy V3 on that chain.

### Testing

Added coverage for:

- Orderbook deserialization and grouping.
- Two-sided and one-sided TVL calculations.
- Quote-token TVL normalization.
- Component decoding and API-key propagation.
- Orderbook simulation and liquidity limits.
- Firm-quote processing.
- Payable-value parsing and malformed-value rejection.
- Native executor constructor and target validation.
- ERC-20 and native-token transfer behavior.
- Rust SwapEncoder to Solidity Executor calldata integration.
- Full TychoRouter swaps with ERC-20 and native-token inputs.

## Verification

- Rust formatting passed.
- Clippy passed for `tycho-execution` and `tycho-simulation`.
- Native Rust execution and simulation tests passed.
- 20 Native executor Solidity unit tests passed.
- 2 RPC-backed TychoRouter integration tests passed.


crates/tycho-execution/config/executor_addresses.json
zach030, line 1: No need to change this file. I'll deploy the executor and set the address here once the integration's done.

crates/tycho-execution/config/executor_deployments.json
```diff
@@ -38,6 +38,10 @@
       "contract": "HashflowExecutor",
       "args": ["0x55084eE0fEf03f14a305cd24286359A35D735151"]
     },
+    "rfq:native": {
+      "contract": "NativeExecutor",
+      "args": ["0xWHITELISTED_ROUTER_ADDRESS_HERE"]
+    },
```
zach030, line 51: Is this `0xWHITELISTED_ROUTER_ADDRESS_HERE` chain-specific? If so, could you add the deployments for each chain and replace the placeholder with the actual address?

crates/tycho-execution/contracts/scripts/deploy-executors.js
```diff
@@ -30,6 +30,7 @@ const deploy_protocols = {
         "vm:fermiswap",
         "vm:bopamm",
         "rfq:metric",
+        "rfq:native"
```
zach030: Please expand to multi-chain.

crates/tycho-execution/config/test_executor_addresses.json
```diff
@@ -21,7 +21,8 @@
     "etherfi": "0xDB25A7b768311dE128BBDa7B8426c3f9C74f3240",
     "rfq:liquorice": "0x3381cD18e2Fb4dB236BF0525938AB6E43Db0440f",
     "vm:fermiswap": "0xe8dc788818033232EF9772CB2e6622F1Ec8bc840",
-    "vm:bopamm": "0x27cc01A4676C73fe8b6d0933Ac991BfF1D77C4da"
+    "vm:bopamm": "0x27cc01A4676C73fe8b6d0933Ac991BfF1D77C4da",
+    "rfq:native": "0x796f2974e3C1af763252512dd6d521E9E984726C"
```
zach030: Please expand to multi-chain.

crates/tycho-execution/contracts/src/executors/NativeExecutor.sol
```diff
@@ -0,0 +1,113 @@
...
+        // If tokenIn is not ETH, we must pass 0 to prevent spending Dispatcher's unrelated ETH.
+        uint256 ethValue = 0;
+        if (tokenIn == ETH_ADDRESS) {
+            ethValue = amountIn > value ? value : amountIn;
+        }
+
+        // Use OpenZeppelin's Address library for safe call
+        // This will revert if the call fails
+        // slither-disable-next-line unused-return
+        target.functionCallWithValue(payload, ethValue);
```
zach030: It's unsafe to allow arbitrary calls. We need to set an allowlist for the target and selector, see [CLAUDE.md](https://github.com/propeller-heads/tycho/blob/main/crates/tycho-execution/CLAUDE.md#building-executors-executor-checklist)

crates/tycho-execution/contracts/src/executors/NativeExecutor.sol
```diff
@@ -0,0 +1,113 @@
...
+        value = uint256(bytes32(data[60:92]));
+        
+        // The remaining bytes are the opaque Native Router calldata
+        payload = data[92:];
+    }
+
+    /**
+     * @dev Allow receiving ETH for settlement calls that require ETH
+     */
+    receive() external payable {}
```
zach030: This is dead code. The Tycho router delegatecalls to the executor.

crates/tycho-simulation/src/rfq/protocols/native/client.rs
```diff
@@ -0,0 +1,674 @@
...
+            entrypoints: vec![],
+        }
+    }
+
+    async fn fetch_orderbook(
+        &self,
+        http_client: &Client,
+    ) -> Result<Vec<NativeOrderbookEntry>, RFQError> {
+        let response = http_client
+            .get("{}/orderbook")
```
zach030: missing `format!` here

crates/tycho-simulation/src/rfq/protocols/native/models.rs
```diff
@@ -0,0 +1,383 @@
...
+    pub router_version: String,
+    #[serde(rename = "toWrap")]
+    pub to_wrap: bool,
+    #[serde(rename = "toUnwrap")]
+    pub to_unwrap: bool,
+    #[serde(rename = "amountInOffset")]
+    pub amount_in_offset: u32,
+    #[serde(rename = "amountOutMinimumOffset")]
+    pub amount_out_minimum_offset: u32,
+}
```
zach030, line 275: Could you check are these fields all required in the API response?

dvansari65: they are all required fields

dvansari65:
@zach030  tested with native's api , /orderbook and /firm-quote
quote endpoints giving successful responses. Also rechecked native executor end to end , passing  all the tests.

dvansari65:
Screenshot 2026-08-10 at 11 44 23 PM
Screenshot 2026-08-10 at 11 45 16 PM

crates/tycho-simulation/src/rfq/protocols/native/state.rs
```diff
@@ -0,0 +1,258 @@
...
+            ));
+        }
+
+        let amount_base = if is_sell_base { amount_in_f64 } else { amount_out_f64 };
+        if self.book.minimum_in_base > 0.0 && amount_base < self.book.minimum_in_base {
+            return Err(SimulationError::RecoverableError(format!(
+                "Amount below minimum. Base amount: {}, min amount: {}",
+                amount_base, self.book.minimum_in_base
+            )));
+        }
```
zach030: `amount_in_f64` is already divided by 10^decimals, so it's on a different scale from `self.minimum_in_base`.

crates/tycho-simulation/src/rfq/protocols/native/client.rs
```diff
@@ -0,0 +1,673 @@
...
+            let component_id =
+                format!("native_relay_{}_{}", entry.base_address, entry.quote_address);
+            match books.entry(component_id) {
+                Entry::Occupied(mut book) => {
+                    let book: &mut NativePriceData = book.get_mut();
+                    match entry.side {
+                        NativeOrderbookSide::Bid => book.bids.extend(entry.levels),
+                        NativeOrderbookSide::Ask => book.asks.extend(entry.levels),
+                    }
+                }
```
zach030: I checked the Native API. Native doesn't give "bid + ask for one pair"; instead, it gives two mirror-image one-sided books: one for `(WETH, USDT, bid)` and one for `(USDT, WETH, bid)`. 

So according to the current parsing rules, we get two different components, each containing only bids with empty asks. I think we should merge the two one-sided books here.

crates/tycho-simulation/src/rfq/protocols/native/client.rs
```diff
@@ -0,0 +1,673 @@
...
+                    Err(e) => {
+                        warn!(
+                            "Failed to read response: {}/{}  {}",
+                            attempt + 1,
+                            MAX_RETRIES - 1,
+                            e
+                        );
+                        last_error = Some(RFQError::ConnectionError(format!(
+                            "Failed to read response from Native failed requests:{e}"
+                        )));
```
zach030: The error handling isn't complete. I tested some API calls and Native returns an error code inside the response body, but the HTTP status stays 200. So we never hit this branch. Examples:

```
HTTP 200  {"code":171055,"message":"seller token amount too small, min: 100000000000"}
HTTP 200  {"code":171015,"message":"quoted token not available"}
```

crates/tycho-simulation/src/rfq/protocols/native/client.rs
```diff
@@ -0,0 +1,673 @@
...
+                        if attempt < MAX_RETRIES - 1 {
+                            tokio::time::sleep(Duration::from_millis(100)).await;
+                            continue;
+                        } else {
+                            return Err(last_error.unwrap());
+                        }
+                    }
+                };
+                last_error = Some(RFQError::ConnectionError(format!(
+                    "Failed to send Liquorice quote request: {err_msg}",
```
zach030: Should be `Native` here.

crates/tycho-simulation/src/rfq/protocols/native/models.rs
```diff
@@ -0,0 +1,420 @@
...
+pub enum NativeSupportedChain {
+    Ethereum,
+    Bsc,
+    Arbitrum,
+    Base,
+    Monad,
+    Xlayer,
+    Robinhood,
+    Morph,
+}
```
zach030, line 290: Tycho doesn't support Monad, Xlayer, Robinhood and Morph yet.

crates/tycho-execution/config/protocol_specific_addresses.json
```diff
@@ -9,13 +9,36 @@
     },
     "rfq:liquorice": {
       "balance_manager_address": "0xb87bAE43a665EB5943A5642F81B26666bC9E5C95"
+    },
+    "rfq:native": {
+      "router_v4": "0x8a2ddc0461Fcf96F81a05529Bed540d4f1eb2a00",
+      "router_v3": "0xa540ec8C73322200d68E1B86c471A5C850854f22",
+      "credit_vault": "0xe3D41d19564922C9952f692C5Dd0563030f5f2EF"
+    }
```
zach030, line 15: These fields aren't used in encoding.

zach030: I'd suggest keeping these fields and validating the address during encoding.

crates/tycho-simulation/src/rfq/protocols/native/state.rs
zach030, line 1: It'd be good to add more tests covering `get_limits` and `get_amount_out`.

crates/tycho-execution/contracts/test/TychoRouterTestSetup.sol
```diff
@@ -251,8 +253,23 @@ contract TychoRouterTestSetup is
...
+        if (nativeTestTarget.code.length == 0) {
+            vm.etch(nativeTestTarget, bytes("1"));
+        }
+        if (DUMMY2.code.length == 0) vm.etch(DUMMY2, bytes("1"));
+        if (DUMMY3.code.length == 0) vm.etch(DUMMY3, bytes("1"));
+        nativeExecutor = new NativeExecutor(
+            nativeTestTarget,
+            DUMMY2,
+            DUMMY3
+        );
```
zach030: Please move nativeTestTarget to `Constants.sol`. Is it a real address? This [address](https://etherscan.io/address/0xb2d1F342D2049684Fb2f8c4eF320633415598333) looks empty.

These dummy contracts have already been [deployed](https://github.com/propeller-heads/tycho/blob/2958893649cb65d3d4b0ff0c696356d4459b446a/crates/tycho-execution/contracts/test/TychoRouterTestSetup.sol#L156-L164), so I don't think we need to etch any bytecode.

crates/tycho-execution/config/executor_addresses.json
```diff
@@ -19,6 +19,7 @@
     "ekubo_v3": "0xD58B803DBE049903f3C1aDaacc2D82e9e514647C",
     "native_wrapper": "0x8FD892DD5F8E153fe15cc3A7567E18ed6f7b1E0A",
     "rfq:liquorice": "0x81c7febeed3fd67a5fdcf892d43489091caeb186",
+    "rfq:native": "0x0000000000000000000000000000000000000000",
```
zach030: Please remove this entry instead of setting it to the zero address.

crates/tycho-execution/config/executor_addresses.json
```diff
@@ -34,7 +34,8 @@
     "native_wrapper": "0x68AF9cDcc2e177e9D75c60f132b8DC52EAF1adBf",
     "lunarbase": "0xf435299a20c1405c957C3F407cc3E9b9df76ee7b",
     "aerodrome_v1": "0x76eeB7E737FDf441E0437F7051b4e0b84808c10a",
-    "rfq:metric": "0xb8644a264a4b765E60C4A65Fb8049EAE1A791D97"
+    "rfq:metric": "0xb8644a264a4b765E60C4A65Fb8049EAE1A791D97",
+    "rfq:native": "0x0000000000000000000000000000000000000000"
```
zach030: All these rfq:metric entries need to be removed.

crates/tycho-execution/config/executor_deployments.json
```diff
@@ -208,6 +240,16 @@
...
+  "monad": {
+    "rfq:native": {
+      "contract": "NativeExecutor",
+      "args": [
+        "0x332e56e0b91eC3F08553eC7964E19CC041A38672",
+        "0x0000000000000000000000000000000000000000",
+        "0xcD1D2D602C3e7394515DaAe96e4FFe16DE71e5B4"
+      ]
+    }
+  },
```
zach030: we don't support monad on tycho yet.

crates/tycho-execution/contracts/src/executors/NativeExecutor.sol
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        // Prevent Dispatcher ETH drain vulnerability:
+        // The `value` from the Native API could maliciously be set to drain the Dispatcher.
+        // If tokenIn is ETH, we strictly bound `ethValue` to the maximum available `amountIn`.
+        // If tokenIn is not ETH, we must pass 0 to prevent spending Dispatcher's unrelated ETH.
+        uint256 ethValue = 0;
+        if (tokenIn == ETH_ADDRESS) {
+            ethValue = amountIn > value ? value : amountIn;
+        }
```
zach030: Where does the `value` come from? Is it possible that `amountIn` and `value` don't match? Would it be better to revert here? Otherwise, we'd be silently modifying the swap's input amount.

crates/tycho-simulation/src/rfq/protocols/native/client.rs
```diff
@@ -0,0 +1,1144 @@
...
+            .eq_ignore_ascii_case(&seller_token.to_string()) ||
+            !order
+                .buyer_token
+                .eq_ignore_ascii_case(&buyer_token.to_string())
+        {
+            return Err(RFQError::ParsingError(format!(
+                "Native Relay quote token mismatch: expected {}/{}, got {}/{}",
+                seller_token, buyer_token, order.seller_token, order.buyer_token
+            )));
+        }
```
zach030: Did you check how the Native API handles native ETH? Same address as Tycho's, or could they mismatch?

crates/tycho-simulation/src/rfq/protocols/native/models.rs
```diff
@@ -0,0 +1,425 @@
...
+            1 => Ok(NativeSupportedChain::Ethereum),
+            56 => Ok(NativeSupportedChain::Bsc),
+            42161 => Ok(NativeSupportedChain::Arbitrum),
+            8453 => Ok(NativeSupportedChain::Base),
+            id => Err(format!("Chain ID {} not supported by Native API", id)),
+        }
+    }
+}
+
+// TODO: take intersection of the tycho supported chains and native supported chain
```
zach030: please remove this comment

crates/tycho-simulation/src/rfq/protocols/native/decoder.rs
```diff
@@ -0,0 +1,236 @@
...
+            .clone();
+
+        // Parse the Relay orderbook snapshot stored by the stream.
+        let empty_book_array: Bytes = "{}".as_bytes().to_vec().into();
+        let book_data = state_attrs
+            .get("book")
+            .unwrap_or(&empty_book_array);
+
+        let book: NativePriceData = serde_json::from_slice(book_data)
+            .map_err(|e| InvalidSnapshotError::ValueError(format!("Invalid book JSON: {e}")))?;
```
zach030, line 60: `unwrap_or(&empty_book_array)` is dead code. Since NativePriceData has a required field, we should return an error directly instead:
```
let book_data = state_attrs
    .get("book")
    .ok_or_else(|| InvalidSnapshotError::MissingAttribute("book".to_string()))?;
```

crates/tycho-simulation/src/rfq/protocols/native/client.rs
```diff
@@ -0,0 +1,1144 @@
...
+        http_client: &Client,
+    ) -> Result<Vec<NativeOrderbookEntry>, RFQError> {
+        let response = http_client
+            .get(format!("{}/orderbook", self.endpoint))
+            .query(&[("chain", self.chain.to_string())])
+            .header("accept", "application/json")
+            .header("apikey", &self.api_key)
+            .send()
+            .await
+            .map_err(|e| RFQError::FatalError(e.to_string()))?;
```
zach030: should be `ConnectionError` here?

crates/tycho-simulation/src/rfq/protocols/native/client_builder.rs
```diff
@@ -0,0 +1,75 @@
...
+};
+
+pub struct NativeClientBuilder {
+    chain: Chain,
+    api_key: String,
+    tokens: HashSet<Bytes>,
+    tvl: f64,
+    quote_tokens: Option<HashSet<Bytes>>,
+    poll_time: Duration,
+    quote_timeout: Duration,
```
zach030, line 18: `quote_timeout` is not configurable here

crates/tycho-execution/contracts/src/executors/NativeExecutor.sol
dvansari65: it comes from the native's /firm-quote response and they explicitly mention to add this in encoder to execute the swap.
ref: https://docs.native.org/native-dev/build-with-native/swap-aggregators/firmquote-swap-apis/miscellaneous/native-tokens-support#onchain-execution
so will look into it more , it'd be better to revert if value and amountin are different

crates/tycho-execution/config/executor_addresses.json
die-herdplatte: I think you meant `rfq:native`?

crates/tycho-execution/contracts/src/executors/NativeExecutor.sol
```diff
@@ -0,0 +1,155 @@
...
+            bytes memory payload
+        ) = _decodeData(data);
+
+        if (!_isValidTarget(target)) {
+            revert NativeExecutor__InvalidTarget();
+        }
+
+        // check payload against function selector
+        bytes4 selector = bytes4(payload);
+        if (selector != _SELECTOR) {
```
die-herdplatte: I think the `creditVault` doesn't expose this method, no?
It could never pass this check and, thus, is currently dead in this executor.

crates/tycho-execution/contracts/src/executors/NativeExecutor.sol
```diff
@@ -0,0 +1,155 @@
...
+contract NativeExecutor is IExecutor {
+    using Address for address;
+
+    address public immutable nativeRouterV4;
+    address public immutable nativeRouterV3;
+    address public immutable creditVault;
+
+    // this function selector is consistent across all versions
+    // of the native router (v3, v4)
+    bytes4 private constant _SELECTOR = 0x0947c2d9;
```
die-herdplatte: I'd name that `_TRADE_RFQT_SELECTOR` and add a comment that this is the entrypoint `tradeRFQT(RFQTQuote quote, uint256 actualSellerAmount, uint256 actualMinOutputAmount)`

crates/tycho-simulation/src/rfq/protocols/native/client.rs
die-herdplatte: Seconding this.
According to [their docs](https://docs.native.org/native-dev/build-with-native/swap-aggregators/firmquote-swap-apis/miscellaneous/native-tokens-support#orderbook), you have to opt in to receive native token books by adding a `showNative` query parameter.
The OpenAPI spec doesn't show that param though.

crates/tycho-execution/contracts/src/executors/NativeExecutor.sol
```diff
@@ -0,0 +1,155 @@
...
+        address _creditVault
+    ) {
+        // Some supported chains do not deploy V3. V4 and CreditVault are always required.
+        if (_nativeRouterV4 == address(0) || _creditVault == address(0)) {
+            revert NativeExecutor__ZeroAddress();
+        }
+        if (
+            _nativeRouterV4.code.length == 0 || _creditVault.code.length == 0
+                || (_nativeRouterV3 != address(0)
+                    && _nativeRouterV3.code.length == 0)
```
die-herdplatte: Now that Monad is gone, we can remove the zero-handling

crates/tycho-execution/contracts/src/executors/NativeExecutor.sol
```diff
@@ -0,0 +1,155 @@
...
+            address tokenIn,
+            address tokenOut,
+            address target,
+            uint256 value,
+            bytes memory payload
+        )
+    {
+        // Decode the 92 bytes fixed header injected by NativeSwapEncoder
+        // 20 (tokenIn) + 20 (tokenOut) + 20 (target) + 32 (value) = 92 bytes header
+        if (data.length < 92) {
```
die-herdplatte: Shouldn't this be +4 due to the selector bytes?
Or even more when we support setting `actualSellerAmount`.

crates/tycho-execution/contracts/test/protocols/Native.t.sol
```diff
@@ -0,0 +1,473 @@
...
+// Unit Tests
+
+contract NativeExecutorUnitTest is Test, Constants {
+    NativeExecutor executor;
+    MockNativeRouter mockV4;
+    MockNativeRouter mockV3;
+    MockNativeRouter mockVault;
+
+    address constant BAD_TARGET = address(0xdead);
+    bytes4 constant ALLOWED_SELECTOR = 0x0947c2d9;
```
die-herdplatte: This could be reused from the executor

crates/tycho-execution/contracts/src/executors/NativeExecutor.sol
```diff
@@ -0,0 +1,155 @@
...
+        bytes calldata /* data */
+    )
+        external
+        view
+        returns (address receiver)
+    {
+        return msg.sender;
+    }
+
+    function swap(uint256 amountIn, bytes calldata data, address receiver)
```
die-herdplatte: We should handler under-delivery here, the `amountIn` can differ by a bit during execution and Native reverts if there's only a 1 wei difference between the signed and the actual amount.
Its entrypoint `tradeRFQT(RFQTQuote quote, uint256 actualSellerAmount, uint256 actualMinOutputAmount)` exposes `actualSellerAmount` which you can use to override it:

```solidity
signed = payload[260:292]
if (amountIn < signed) {
    payload[amountInOffset:amountInOffset+32] = amountIn
}
```
roughly.

`amountInOffset` is returned by `/v1/firm-quote`

crates/tycho-execution/contracts/src/executors/NativeExecutor.sol
```diff
@@ -0,0 +1,155 @@
...
+            revert NativeExecutor__InvalidPayload();
+        }
+
+        // `value` comes from Native's txRequest. It must exactly match the ETH
+        // input, or be zero for ERC-20 swaps, so the executor never silently
+        // changes the quoted input or spends unrelated Dispatcher ETH.
+        uint256 expectedValue = tokenIn == ETH_ADDRESS ? amountIn : 0;
+        if (value != expectedValue) {
+            revert NativeExecutor__InvalidValue();
+        }
```
die-herdplatte: I don't understand what this should protect against.
`amountIn` is the authority on how much should be spent.
This check also makes under-delivery impossible, as detailed in another comment.

crates/tycho-execution/contracts/test/protocols/Native.t.sol
```diff
@@ -0,0 +1,473 @@
...
+}
+
+// Integration Tests
+
+contract TychoRouterNativeIntegrationTest is TychoRouterTestSetup {
+    function getForkBlock() public pure override returns (uint256) {
+        return 25747646;
+    }
+
+    function test_RecordedQuoteAgainstRealNativeRouter() public {
```
die-herdplatte: I'd move that to its own test contract, it doesn't use the Tycho Router

crates/tycho-execution/contracts/test/protocols/Native.t.sol
```diff
@@ -0,0 +1,473 @@
...
+
+        vm.prank(ALICE);
+        (bool success,) = tychoRouterAddr.call{value: amountIn}(callData);
+
+        assertTrue(success, "Call Failed");
+        assertEq(USDC.balanceOf(BOB) - balanceBefore, amountOut);
+        assertEq(target.balance, amountIn);
+        assertEq(USDC.balanceOf(tychoRouterAddr), 0);
+        assertEq(tychoRouterAddr.balance, 0);
+    }
```
die-herdplatte: These two tests are still mocked (`vm.etch`).
Would it be possible to record real quotes instead?

crates/tycho-simulation/src/rfq/protocols/native/models.rs
```diff
@@ -0,0 +1,424 @@
...
+    pub tx_request: TxRequest,
+    pub source: Vec<u32>,
+    #[serde(rename = "errorMessage")]
+    pub error_message: String,
+    #[serde(rename = "router_version")]
+    pub router_version: String,
+    #[serde(rename = "toWrap")]
+    pub to_wrap: bool,
+    #[serde(rename = "toUnwrap")]
+    pub to_unwrap: bool,
```
die-herdplatte: We should either assert this against `token_in`/`token_out` or drop it

crates/tycho-simulation/src/rfq/protocols/native/client.rs
```diff
@@ -0,0 +1,1192 @@
...
+        let chain = NativeSupportedChain::try_from(self.chain).map_err(RFQError::FatalError)?;
+
+        let request_data = FirmQuoteRequest {
+            src_chain: chain,
+            dst_chain: chain,
+            from_address: receiver.to_string(),
+            amount_wei: params.amount_in.to_string(),
+            token_in: token_in.to_string(),
+            token_out: token_out.to_string(),
+            version: 4,
```
die-herdplatte, line 774: This means we can drop V3 support completely?

crates/tycho-execution/model/src/model/executors.rs
die-herdplatte, line 36: Native doesn't belong in here, it doesn't allow callers to choose the target freely

crates/tycho-execution/src/encoding/evm/swap_encoder/mod.rs
```diff
@@ -17,6 +17,7 @@ mod liquorice;
 mod lunarbase;
 mod maverick_v2;
 mod metric;
+pub(crate) mod native;
```
die-herdplatte: Why is that `pub(crate)`?

crates/tycho-simulation/src/rfq/protocols/native/client.rs
```diff
@@ -0,0 +1,1192 @@
...
+    pub quote_timeout: Duration,
+}
+
+impl NativeClient {
+    pub const PROTOCOL_SYSTEM: &'static str = "rfq:native";
+    pub const DEFAULT_ENDPOINT: &'static str = "https://v2.api.native.org/swap-api-v2/v1";
+
+    fn classify_api_error(error: &NativeApiErrorResponse) -> (bool, RFQError) {
+        let message = format!("Native API error {}: {}", error.code, error.message);
+        match error.code {
```
die-herdplatte, line 97: A link to https://docs.native.org/native-dev/build-with-native/swap-aggregators/firmquote-swap-apis/miscellaneous/error-handling#error-codes would be nice here

crates/tycho-simulation/src/rfq/protocols/native/client.rs
```diff
@@ -0,0 +1,1192 @@
...
+    }
+
+    async fn request_binding_quote(
+        &self,
+        params: &GetAmountOutParams,
+    ) -> Result<SignedQuote, RFQError> {
+        let receiver = bytes_to_address(&params.receiver)?;
+        let token_in = bytes_to_address(&params.token_in)?;
+        let token_out = bytes_to_address(&params.token_out)?;
+        let http_client = Client::new();
```
die-herdplatte: This should mirror https://github.com/propeller-heads/tycho/blob/ca85c29ed8a77cf3f0130e4f5eaaf870c71adafa/crates/tycho-simulation/src/rfq/protocols/metric/client.rs#L44 Right now, we're paying TLS handshake costs on every firm quote request.

crates/tycho-simulation/src/rfq/protocols/native/client.rs
```diff
@@ -0,0 +1,1192 @@
...
+        let result = client
+            .request_binding_quote(&create_test_quote_params())
+            .await;
+
+        assert!(matches!(result, Err(RFQError::FatalError(_))));
+        assert_eq!(request_count.load(Ordering::SeqCst), 1);
+    }
+
+    #[test]
+    fn classifies_only_documented_native_api_codes() {
```
die-herdplatte: This can be dropped, it basically just copies the method in here, so it asserts an exact match instead of some behavior

crates/tycho-simulation/src/rfq/protocols/native/client.rs
```diff
@@ -0,0 +1,1192 @@
...
+                "Native Relay orderbook HTTP error {}: {}",
+                response.status(),
+                response
+                    .text()
+                    .await
+                    .unwrap_or_default()
+            )));
+        }
+
+        response.json().await.map_err(|e| {
```
die-herdplatte: Can't this endpoint also return `NativeApiErrorResponse` when the status is OK?
This would fail to parse here.

crates/tycho-simulation/src/rfq/protocols/native/client.rs
```diff
@@ -0,0 +1,1192 @@
...
+            from_address: receiver.to_string(),
+            amount_wei: params.amount_in.to_string(),
+            token_in: token_in.to_string(),
+            token_out: token_out.to_string(),
+            version: 4,
+            allow_multihop: false, // Multihop not implemented yet
+        };
+        const MAX_RETRIES: u32 = 3;
+        let mut last_error = None;
+        let start_time = std::time::Instant::now();
```
die-herdplatte: **Timeout handling: one deadline instead of five**

`quote_timeout` is a budget for the whole call, but it's enforced by hand at every await point — loop-head bail + `remaining_time` (L564-573), `timeout` on send (L580), recompute + bail again (L609-615), `timeout` on body read (L616), and `wait_before_retry`'s own `checked_sub`/`is_zero`/`min`/post-check (L78-92). Five derivations of the same quantity from the same `Instant`, already drifting: L583 logs `attempt/{MAX_RETRIES - 1}` while L627 logs `attempt/{MAX_RETRIES}`, and L601/L611/L641 build three identical "timed out after N seconds" strings.

One outer timeout owns the budget and nothing inside needs to know about it:

```rust
let mut last_error = None;

let attempts = async {
    for attempt in 1..=MAX_RETRIES {
        match try_quote().await {          // send -> read body -> classify -> parse
            Ok(quote)        => return Ok(quote),
            Err(Fatal(e))    => return Err(e),
            Err(Retry(e, d)) => { last_error = Some(e); sleep(d).await; }
        }
    }
    Err(last_error.take().unwrap_or(..))
};

// bind, don't inline: drops the future (and its borrow) before the match
let result = timeout(self.quote_timeout, attempts).await;
result.unwrap_or_else(|_| Err(last_error.unwrap_or(timed_out())))
```

`try_quote` does one attempt with no elapsed arithmetic at all, returning `Retry{error, delay}` or `Fatal(error)`. That deletes `remaining_time`, both interior bail-outs, the second `timeout`, and `wait_before_retry` — including its dead-sleep case, where a remaining budget ≤ the backoff is slept away only to return `false`.

Two more things in this function while you're here:

- `success == false` is handled twice, with two different verdicts: `FatalError` at L713, and `QuoteNotFound` inside `process_quote_response` (L296), which is unreachable because the caller already gates on `success`. Inconsistent — let's pick one and delete the other.
- A 5xx without an error envelope terminates on attempt 1 of 3 (L680: a bare `return Err`, no `continue`), while a connection error and even a malformed 200 body are retried three times. It should be retried too. Note it's currently conditional on cosmetics: a 503 that happens to carry `{"code":…,"message":…}` goes through `classify_api_error` and may be retried, whereas the same 503 with an HTML or empty body from an edge proxy does not.

crates/tycho-simulation/src/rfq/protocols/native/client.rs
```diff
@@ -0,0 +1,1192 @@
...
+                    return Err(last_error.unwrap());
+                }
+                Err(_) => {
+                    return Err(RFQError::ConnectionError(format!(
+                        "Native quote request timed out after {} seconds",
+                        self.quote_timeout.as_secs()
+                    )))
+                }
+            };
+            let response_text = String::from_utf8_lossy(&response_body);
```
die-herdplatte: This is only required in the error branch

crates/tycho-simulation/src/rfq/protocols/native/client.rs
```diff
@@ -0,0 +1,1192 @@
...
+                params.amount_in, quoted_amount_in
+            )));
+        }
+
+        if quote_response
+            .tx_request
+            .calldata
+            .is_empty()
+        {
+            return Err(RFQError::ConnectionError("Calldata not found!".to_string()));
```
die-herdplatte: This should probably be something else, mabye a `QuoteNotFound`?

crates/tycho-simulation/src/rfq/protocols/native/client.rs
```diff
@@ -0,0 +1,1192 @@
...
+        }
+        // Decode calldata (pre-built by Native Relay, ready to submit as-is)
+        let calldata = hex::decode(
+            quote_response
+                .tx_request
+                .calldata
+                .trim_start_matches("0x"),
+        )
+        .map_err(|e| RFQError::ParsingError(format!("Failed to decode calldata: {e}")))?;
```
die-herdplatte, line 503: We could assert that it matches the minimum length required by the executor here (or even the exact one, depending on how the executor calldata verification will be changed)

crates/tycho-simulation/src/rfq/protocols/native/client.rs
```diff
@@ -0,0 +1,1192 @@
...
+
+        // Signature may be empty for Relay-signed pool orders; store only if present.
+        if !order.signature.is_empty() {
+            quote_attributes.insert(
+                "signature".to_string(),
+                Bytes::from(hex::decode(order.signature.trim_start_matches("0x")).map_err(
+                    |e| RFQError::ParsingError(format!("Failed to decode signature: {e}")),
+                )?),
+            );
+        }
```
die-herdplatte: Why do we store those attributes here?
`deadline_timestamp` could be useful though if we validate it in the encoder.

crates/tycho-simulation/src/rfq/protocols/native/client.rs
```diff
@@ -0,0 +1,1192 @@
...
+                        entries.iter().min_by(|a, b| {
+                            a.base_address
+                                .as_ref()
+                                .cmp(b.base_address.as_ref())
+                        })
+                    })
+                    .unwrap_or(first)
+            } else {
+                first
+            };
```
die-herdplatte: We should just use the already sorted pairs as a stable reference for component IDs (the keys of the `HashMap`).
With the current logic, components could appear and disappear very frequently although they describe the same two tokens, e.g. if you have `(WETH, USDC)` in one poll and `(USDC, WETH)` in the next.

crates/tycho-simulation/src/rfq/protocols/native/models.rs
```diff
@@ -0,0 +1,424 @@
...
+    pub minimum_in_base: f64,
+    pub side: NativeOrderbookSide,
+    #[serde(deserialize_with = "deserialize_levels")]
+    pub levels: Vec<NativePriceLevel>,
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct NativePriceData {
+    pub base_symbol: String,
+    pub quote_symbol: String,
```
die-herdplatte: Those aren't used anywhere and can be dropped

crates/tycho-simulation/src/rfq/protocols/native/models.rs
```diff
@@ -0,0 +1,424 @@
...
+    pub router_version: String,
+    #[serde(rename = "toWrap")]
+    pub to_wrap: bool,
+    #[serde(rename = "toUnwrap")]
+    pub to_unwrap: bool,
+    #[serde(rename = "amountInOffset")]
+    pub amount_in_offset: u32,
+    #[serde(rename = "amountOutMinimumOffset")]
+    pub amount_out_minimum_offset: u32,
+}
```
die-herdplatte, line 275: These could use a `#[serde(rename_all = "camelCase")]`

crates/tycho-simulation/src/rfq/protocols/native/models.rs
```diff
@@ -0,0 +1,424 @@
...
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct NativePriceData {
+    pub base_symbol: String,
+    pub quote_symbol: String,
+    pub base_address: Bytes,
+    pub quote_address: Bytes,
+    pub minimum_in_base: f64,
+    #[serde(default)]
```
die-herdplatte: Why is this defaulted?

crates/tycho-simulation/src/rfq/protocols/native/models.rs
```diff
@@ -0,0 +1,424 @@
...
+impl TryFrom<tycho_common::models::Chain> for NativeSupportedChain {
+    type Error = String;
+
+    fn try_from(chain: tycho_common::models::Chain) -> Result<Self, Self::Error> {
+        match chain.id() {
+            1 => Ok(NativeSupportedChain::Ethereum),
+            56 => Ok(NativeSupportedChain::Bsc),
+            42161 => Ok(NativeSupportedChain::Arbitrum),
+            8453 => Ok(NativeSupportedChain::Base),
+            id => Err(format!("Chain ID {} not supported by Native API", id)),
```
die-herdplatte: This could match on the variants directly

crates/tycho-simulation/src/rfq/protocols/native/state.rs
```diff
@@ -0,0 +1,560 @@
...
+            ));
+        }
+        if self
+            .book
+            .bids
+            .iter()
+            .chain(self.book.asks.iter())
+            .any(|level| {
+                !level.quantity.is_finite() ||
+                    level.quantity <= 0.0 ||
```
die-herdplatte: Should we error if the quantity is 0?

crates/tycho-simulation/src/rfq/protocols/native/state.rs
```diff
@@ -0,0 +1,560 @@
...
+}
+
+#[typetag::serde]
+impl ProtocolSim for NativeState {
+    fn fee(&self) -> f64 {
+        0.0
+    }
+
+    fn spot_price(&self, base: &Token, quote: &Token) -> Result<f64, SimulationError> {
+        self.validate_book()?;
```
die-herdplatte: This should only happen once (also in `get_amount_out` and `get_limits`).

crates/tycho-simulation/src/rfq/protocols/utils.rs
```diff
@@ -26,3 +27,11 @@ pub fn default_quote_tokens_for_chain(chain: &Chain) -> Result<HashSet<Bytes>, R
         _ => Ok(HashSet::new()),
     }
 }
+
+pub fn bytes_to_address(address: &Bytes) -> Result<Address, RFQError> {
+    if address.len() == 20 {
+        Ok(Address::from_slice(address))
+    } else {
+        Err(RFQError::InvalidInput(format!("Invalid ERC20 token address: {address:?}")))
```
die-herdplatte: The error message could be more generic

die-herdplatte (changes requested):
Thanks for the PR!

Native still needs to be added to `crates/tycho-integration-test/src/stream_processor/rfq_stream_processor.rs`.

After you have applied the feedback, could you please merge main into this branch?
Thanks!

crates/tycho-simulation/src/rfq/protocols/native/client.rs
```diff
@@ -0,0 +1,1485 @@
...
+            entrypoints: vec![],
+        }
+    }
+
+    async fn fetch_orderbook(&self) -> Result<Vec<NativeOrderbookEntry>, RFQError> {
+        let chain = NativeSupportedChain::try_from(self.chain).map_err(RFQError::InvalidInput)?;
+        let response = self
+            .http_client()
+            .get(format!("{}/orderbook", self.endpoint))
+            .query(&[("chain", chain.as_str()), ("showNative", "true")])
```
zach030: `showNative=true` isn't a valid value. The docs only accept `showNative=0xE` or `showNative=0x0`, so the API is most likely ignoring this and we still won't get any native-token books. Let's use `0x0` since it matches Tycho's internal representation end to end.

Worth hitting the real API to confirm — the test we have only asserts on the URL we build ourselves, so it passes either way.

https://docs.native.org/native-dev/build-with-native/swap-aggregators/firmquote-swap-apis/miscellaneous/native-tokens-support#orderbook

crates/tycho-simulation/src/rfq/protocols/native/client.rs
```diff
@@ -0,0 +1,1485 @@
...
+        let mut books = HashMap::new();
+        for ((token0, token1), entries) in entries_by_pair {
+            let Some(first) = entries.first() else {
+                continue;
+            };
+            let has_reverse = entries.iter().any(|entry| {
+                entry.base_address == first.quote_address &&
+                    entry.quote_address == first.base_address
+            });
+            let canonical = if has_reverse {
```
zach030: The quote-token preference only kicks in when `has_reverse` is true. But Native drops a whole orderbook object when one direction runs dry, so one poll can carry `(WETH, USDC)` and the next only `(USDC, WETH)`. In that case we fall back to `entries.first()`, which is just API response order.

The component ID is stable now, but base/quote still flip under the same ID, and `stream()` filters on `quote_tokens.contains(&book.quote_address)` — so the market can drop in and out of the component set between polls depending on which direction Native happens to publish. That's the same flapping the stable ID was meant to fix.

Could we apply the preference unconditionally (falling back to the sorted pair order) and mirror whichever direction the API gives us? The mirroring code is already there.

crates/tycho-simulation/src/rfq/protocols/native/client.rs
```diff
@@ -0,0 +1,1485 @@
...
+                let is_direct =
+                    entry.base_address == base_address && entry.quote_address == quote_address;
+                if is_direct {
+                    minimum_in_base = minimum_in_base.max(entry.minimum_in_base);
+                    match entry.side {
+                        NativeOrderbookSide::Bid => direct_bids.extend(entry.levels),
+                        NativeOrderbookSide::Ask => direct_asks.extend(entry.levels),
+                    }
+                } else {
+                    minimum_in_quote = minimum_in_quote.max(entry.minimum_in_base);
```
zach030: Follows from the canonical-direction comment: when Native publishes only one direction, every entry counts as direct, so `minimum_in_quote` stays 0 and the min-amount check in `get_amount_out` is skipped entirely for quote->base. We'd hand out quotes that Native then rejects with 171055.

crates/tycho-simulation/src/rfq/protocols/native/client.rs
```diff
@@ -0,0 +1,1485 @@
...
+                    let levels = NativePriceData::invert_price_levels(&entry.levels);
+                    match entry.side {
+                        NativeOrderbookSide::Bid => mirrored_asks.extend(levels),
+                        NativeOrderbookSide::Ask => mirrored_bids.extend(levels),
+                    }
+                }
+            }
+
+            let bids = if direct_bids.is_empty() { mirrored_bids } else { direct_bids };
```
zach030: Mirrored levels get dropped when we have direct ones for the same side. I think that's intentional and fine, but a one-line comment would keep it from looking like a bug later.

crates/tycho-simulation/src/rfq/protocols/native/client.rs
```diff
@@ -0,0 +1,1485 @@
...
+        }
+
+        let effective_amount_in =
+            BigUint::from_str(&order.effective_seller_token_amount).map_err(|_| {
+                RFQError::ParsingError(format!(
+                    "Failed to parse effective seller token amount: {}",
+                    order.effective_seller_token_amount
+                ))
+            })?;
+        if effective_amount_in != params.amount_in {
```
zach030: Do we know what `effectiveSellerTokenAmount` actually means? The response also carries `tokenTransferFeeOnPercent`, which suggests fee-on-transfer tokens are supported — and that's exactly the case where this field should differ from `sellerTokenAmount`.

Right now we require all three of `amountIn`, `sellerTokenAmount` and `effectiveSellerTokenAmount` to equal the requested amount, so those markets would always fail here. Worth checking with Native before we keep the third check.

crates/tycho-execution/contracts/src/executors/NativeExecutor.sol
```diff
@@ -0,0 +1,185 @@
...
+
+        // Native treats a zero actualSellerAmount as "use the signed amount".
+        // Therefore zero cannot represent an actual zero input.
+        if (amountIn == 0 || signedAmountIn == 0) {
+            revert NativeExecutor__InvalidAmountIn();
+        }
+
+        _validateAmountInOffset(payload, amountInOffset);
+
+        if (amountIn != signedAmountIn) {
```
zach030, line 95: This now allows `amountIn` to be arbitrarily larger than the signed amount, but `actualMinOutputAmount` is left untouched. So for over-delivery we send more input against the min-output that was signed for the smaller amount — that extra input has no price protection at this hop, and the only backstop is the router-level min-amount-out for the whole solution.

Under-delivery is safe by construction (less in, same min out); over-delivery is the asymmetric direction. Two options I see: keep the `amountIn > signedAmountIn` revert and re-quote upstream, or scale `actualMinOutputAmount` proportionally (which means plumbing `amountOutMinimumOffset` too).

Also note the over-delivery test is mock-only — `MockNativeRouter` just records calldata, so it proves we write the word but not that the real Native router accepts `actualSellerAmount` above the signed amount, or at what price. Given we recorded a real quote for the 1-wei under-delivery case, could we do the same for over-delivery?

crates/tycho-execution/contracts/src/executors/NativeExecutor.sol
```diff
@@ -0,0 +1,185 @@
...
+        tokenOut = address(bytes20(data[20:40]));
+        target = address(bytes20(data[40:60]));
+        amountInOffset = uint32(bytes4(data[60:64]));
+        signedAmountIn = uint256(bytes32(data[64:96]));
+
+        // The remaining bytes are the opaque Native Router calldata
+        payload = data[_FIXED_HEADER_LENGTH:];
+    }
+
+    function _validateAmountInOffset(
```
zach030: Since the selector is pinned to `tradeRFQT(RFQTQuote, uint256, uint256)`, `actualSellerAmount` is always at 4 + 32 = 36 by ABI construction, and the docs sample confirms 36. We're plumbing a constant all the way through: API -> `quote_attributes` -> 4 header bytes -> executor, with validation on both ends plus three tests.

We also only check alignment and bounds here, so any aligned word is accepted — including `actualMinOutputAmount` at 68, or a word inside the signed quote. Most of those revert anyway so the risk is low, but pinning `amountInOffset == 36` (or dropping the field and hardcoding it) removes the surface and a fair bit of machinery. Worth deciding before merge since it changes the header layout.

zach030 (commented):
Went through the new commits against the last round. 31 of the 32 comments are properly addressed, and the fixes look right rather than just present — the under-delivery override, the native-ETH address normalization, and the timeout refactor all check out. Swapping the `vm.etch` mocks for a recorded 1-wei under-delivery against the real router is the right way to settle that one.

Comments below. The `showNative` value and the canonical-direction one look like they need fixing before merge, and I'd like to settle the over-delivery and `amountInOffset` questions now too since both affect the header layout / executor semantics.

crates/tycho-execution/contracts/src/executors/NativeExecutor.sol
```diff
@@ -0,0 +1,185 @@
...
+        bytes memory payload,
+        uint32 amountInOffset
+    ) private pure {
+        // Native's offset is relative to the complete Router calldata. The first
+        // ABI word begins after the 4-byte selector, and every argument word is
+        // 32-byte aligned from there.
+        if (
+            amountInOffset < 4 || (amountInOffset - 4) % 32 != 0
+                || payload.length < 32 || amountInOffset > payload.length - 32
```
zach030: `payload.length < 32` can't trigger — `_decodeData` already guarantees at least 100 bytes of payload.

crates/tycho-simulation/src/rfq/protocols/native/state.rs
```diff
@@ -0,0 +1,584 @@
...
+    },
+    Bytes,
+};
+
+use crate::rfq::{
+    client::RFQClient,
+    protocols::native::{client::NativeClient, models::NativePriceData},
+};
+
+#[derive(Clone, Serialize, Deserialize)]
```
zach030, line 26: `validate_book` only runs in `new()` now, but this derives `Deserialize`, so a deserialized state skips both the zero-quantity filter and the validation. Harmless today (zero-quantity levels just waste a loop iteration, and `delta_transition` is unimplemented so the book can't change), but the invariant now only holds on one construction path. Either validate via a `TryFrom`-based `Deserialize` or leave a note.

crates/tycho-simulation/src/rfq/protocols/native/models.rs
```diff
@@ -0,0 +1,393 @@
...
+    pub version: u32,
+    pub allow_multihop: bool,
+}
+
+// --- Response ---
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct WidgetFee {
+    pub signer: String,
+    #[serde(rename = "feeRecipient")]
```
zach030: Missed one — this could use `rename_all = "camelCase"` as well. (Nice catch keeping the explicit rename on `router_version`, by the way — the API really does send that one field in snake_case.)

crates/tycho-execution/contracts/src/executors/NativeExecutor.sol
```diff
@@ -96,30 +110,65 @@ contract NativeExecutor is IExecutor {
...
+            )
+            actualMinOutputAmount := mload(
+                add(add(payload, 0x20), _ACTUAL_MIN_OUTPUT_AMOUNT_OFFSET)
+            )
+        }
+
+        // Native's firm-quote calldata leaves both unsigned override arguments
+        // at zero. Tycho owns these fields during execution: accepting a pre-set
+        // seller amount could bypass amountIn, while a pre-set minimum would
+        // disable Native's automatic slippage adjustment.
```
die-herdplatte: I'd frame this as "those two fields are the only ones not covered by the signature"

crates/tycho-execution/contracts/test/Constants.sol
```diff
@@ -232,8 +240,12 @@ contract Constants is
     address LIQUORICE_BALANCE_MANAGER =
         0xb87bAE43a665EB5943A5642F81B26666bC9E5C95;
 
-    // Native Relay
+    // Native Relay V4 routers
     address NATIVE_ROUTER_V4 = 0x8a2ddc0461Fcf96F81a05529Bed540d4f1eb2a00;
```
die-herdplatte: This should have a chain suffix as well

crates/tycho-simulation/src/rfq/protocols/native/client.rs
```diff
@@ -52,43 +93,40 @@ impl NativeClient {
     pub const PROTOCOL_SYSTEM: &'static str = "rfq:native";
     pub const DEFAULT_ENDPOINT: &'static str = "https://v2.api.native.org/swap-api-v2/v1";
 
-    fn classify_api_error(error: &NativeApiErrorResponse) -> (bool, RFQError) {
+    fn http_client(&self) -> &Client {
```
die-herdplatte: I don't think this needs a method

crates/tycho-simulation/src/rfq/protocols/native/models.rs
```diff
@@ -40,23 +56,48 @@ where
     })
 }
 
+const NATIVE_TOKEN_ALIAS: Address = address!("EeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE");
+
+pub(super) fn normalize_native_address(address: Address) -> Address {
```
die-herdplatte: afaict this is essentially an identity function because no consumer can ever pass in `NATIVE_TOKEN_ALIAS` here.
Tycho itself and `showNative=0x0` result in the native token being represented as the zero address.

You can also drop `deserialize_address` then.

`accepts_native_eth_orderbook_using_e_address_alias` and `accepts_native_eth_response_using_e_address_alias` also need updating, change it to the zero address here.

crates/tycho-simulation/src/rfq/protocols/native/client.rs
```diff
@@ -35,6 +39,43 @@ use crate::{
...
+
+impl AggregatedLevels {
+    fn extend(&mut self, levels: Vec<NativePriceLevel>, minimum: f64) {
+        self.levels.extend(levels);
+        self.minimum = self.minimum.max(minimum);
+    }
+}
+
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct NativeClient {
```
die-herdplatte, line 77: Why are all the fields `pub`?
If you make them private, you can also drop the `validate_poll_time` call in `stream` and inline the function body in the constructor

crates/tycho-simulation/src/rfq/protocols/native/models.rs
```diff
@@ -65,19 +106,27 @@ pub struct NativeOrderbookEntry {
...
+    /// Atomic quote-token minimum input when selling quote into asks.
     pub minimum_in_quote: f64,
+    /// Atomic base-token minimum output when selling quote into asks. Defaults to zero for
+    /// snapshots written before output minimums were tracked.
+    #[serde(default)]
+    pub minimum_out_base: f64,
+    /// Atomic quote-token minimum output when selling base into bids. Defaults to zero for
+    /// snapshots written before output minimums were tracked.
+    #[serde(default)]
+    pub minimum_out_quote: f64,
```
die-herdplatte: We don't really need to be backward-compatible here, no?

die-herdplatte (approved):

crates/tycho-simulation/src/rfq/protocols/native/client.rs
```diff
@@ -0,0 +1,1958 @@
...
+        let chain = NativeSupportedChain::try_from(self.chain).map_err(RFQError::FatalError)?;
+
+        let request_data = FirmQuoteRequest {
+            src_chain: chain,
+            dst_chain: chain,
+            from_address: receiver.to_string(),
+            amount_wei: params.amount_in.to_string(),
+            token_in: token_in.to_string(),
+            token_out: token_out.to_string(),
+            version: 4,
```
zach030: Any idea why we're pinned to v4? Docs say v6 is the latest.

PR #1250, opened 2026-07-29
feat(execution): accept ERC-1271 client fee signatures

tamaralipows:
`clientSignature` only verified with `ECDSA.recover`, so a client whose treasury is a Safe, smart account, or DAO could not sign a fee at all. Verification now falls back to ERC-1271 when recovery does not match `clientFeeReceiver`.

ECDSA runs before that fallback rather than branching on `signer.code.length` the way `SignatureChecker.isValidSignatureNow` does. Code-length-first rejects an EOA carrying an EIP-7702 delegation designator, and it costs an extra cold `EXTCODESIZE` on the common path.

The second commit corrects the `ClientFee` typehash in the encoding docs. It omitted the swap intent fields the router has signed since v3, so the documented signing example produced signatures the router rejects.


louise-poole (commented):
Thank youuuuuu! Much smaller and well contained than I was expecting. Looks good to me! Please merge this into the `feature/extend-router-fees` branch? We're deploying from that branch with our newest router version.

PR #1251, opened 2026-07-29
refactor(execution): rename the TychoRouter contract to TychoRouterV3

tamaralipows:
**Two things kept their old name on purpose:** 
- the `TychoRouter__*` errors, so revert selectors don't change (in case clients are already matching by revert reason)
- the EIP-712 domain name `"TychoRouter"`, so client fee signature digests don't change. 

**To verify:** 
- Test files, the `tycho-router` CLI subcommand, and the rust encoder types are out of scope. Are we in agreement here?

**Verified:** 
- The runtime bytecode fixture needed no regeneration: `bytecode_hash = 'none'` drops the metadata trailer that would otherwise carry the contract name. 
- The monofile is safe to delete: nothing in `docs/` or `.gitbook.yaml` links to it, there's no generator for it in the repo, and it documents the V2-era `tycho_router` transfer type.
- The security model still runs: 37.8M simulations across all 11 executor combinations, 0 suspicious and 0 warnings.

**No client action needed:**

Domain separator, revert selectors, function selectors, deployed addresses, and the encoder API are all unchanged. Clients keep signing exactly as before.


docs/.gitbook/assets/Tycho (1).svg
louise-poole, line 1: Are these images even used anywhere?

louise-poole (approved):
Looks good! One minor docs comment, otherwise I've approved ✅

docs/.gitbook/assets/Tycho (1).svg
tamaralipows: orphans! I will remove them and do an asset sweep for more orphans

PR #1254, opened 2026-07-29
feat: tag exclusive components with is_exclusive static attribute

Troshchk:
Components whose swaps require off-chain authorization are now tagged with an  `is_exclusive` static attribute (value `0x01`) during stream decoding. Absence of  the attribute means the component is not exclusive.

  Currently this applies to ekubo_v3 SignedExclusiveSwap pools; the mechanism is generic and other protocols can register their own predicate.


crates/tycho-simulation/src/evm/protocol/mod.rs
```diff
@@ -24,6 +26,12 @@ pub mod utils;
 pub mod velodrome_slipstreams;
 pub mod vm;
 
+/// Extension contracts that gate swaps behind off-chain authorization. A component carrying one
+/// of these in its `extension` static attribute is exclusive, whatever its protocol.
+///
+/// Used by Fynd to filter exclusive components out of public-only routing graphs.
+pub const EXCLUSIVE_EXTENSIONS: &[Address] = &[ekubo_v3::addresses::SIGNED_EXCLUSIVE_SWAP_ADDRESS];
```
louise-poole: I like the plan to export the exclusive extension instead of maintaining it's address in Fynd, however I think the concept of 'extensions' is specific to Ekubo and Balancer, for example, might have a different identifier. So I would not pre-emptively export this info in a higher level const when others likely will never be added to this. Instead let's just export it from the ekubo_v3 module directly?

louise-poole (approved):
Looks good!

crates/tycho-simulation/src/evm/decoder.rs
```diff
@@ -571,10 +588,19 @@ where
...
                     );
+                    if self
+                        .exclusivity_fns
+                        .get(protocol.as_str())
+                        .is_some_and(|is_exclusive| is_exclusive(&snapshot))
+                    {
+                        component
+                            .static_attributes
+                            .insert("is_exclusive".to_string(), Bytes::from(vec![1u8]));
+                    }
```
louise-poole: Hmmm.... this doesn't feel so clean tbh. I don't like that this is in the general decoder (which is already too convoluted as it stands). I actually had a thought that maybe I should have thought of earlier: we should add this static attribute at indexing time (substreams side).
Originally I was trying to avoid this as I thought we might need to resync ekubo_v3, but actually we haven't started indexing our exclusive pool yet so have freedom there right now. Let's update the ekubo_v3 substreams package to add this and include it as a reserved attribute [here](https://docs.propellerheads.xyz/tycho/for-dexs/protocol-integration/indexing/reserved-attributes)? What do you think?

Troshchk: I agree with you -- that's the cleanest approach. I'll make the change 👍

protocols/substreams/ethereum-ekubo-v3/src/addresses.rs
```diff
@@ -4,3 +4,6 @@ pub const CORE_ADDRESS: Address = address!("0x00000000000014aA86C5d3c41765bb24e1
 pub const TWAMM_ADDRESS: Address = address!("0xd4F1060cB9c1A13e1d2d20379b8aa2cF7541eD9b");
 pub const BOOSTED_FEES_CONCENTRATED_ADDRESS: Address =
     address!("0xd4b54d0ca6979da05f25895e6e269e678ba00f9e");
+// TODO: replace with the deployed SignedExclusiveSwap extension address.
```
tamaralipows, line 7: We already have this deployed! https://etherscan.io/address/0x55b703eED01b35641963da2FB2E14885993605A3#code

tamaralipows (approved):
Looks much cleaner now, thank you! Please just add the proper address there for filtering since we already have it

docs/for-solvers/supported-protocols.md
```diff
@@ -16,7 +16,7 @@ Currently, Tycho supports the following protocols:
...
-<tr><td><code>ekubo_v3</code></td><td>Native (<code>EkuboV3State</code>)</td><td>9μs</td><td>Ethereum</td><td>Some extensions are unsupported. Use <code>ekubo_v3_extension_filter</code>. It also drops SignedExclusiveSwap pools, which need a per-swap signature passed to the encoder as <code>user_data</code>. If you can supply that signature, use <code>ekubo_v3_extension_filter_with_signed_exclusive_swap</code> instead to keep those pools.</td></tr>
+<tr><td><code>ekubo_v3</code></td><td>Native (<code>EkuboV3State</code>)</td><td>9μs</td><td>Ethereum</td><td>Some extensions are unsupported. Use <code>ekubo_v3_extension_filter</code>. It also drops SignedExclusiveSwap pools, which need a per-swap signature passed to the encoder as <code>user_data</code>. If you can supply that signature, use <code>ekubo_v3_extension_filter_with_signed_exclusive_swap</code> instead to keep those pools. Pools kept this way carry an <code>is_exclusive</code> static attribute (value <code>0x01</code>) — test for the attribute's presence to keep them out of public routing.</td></tr>
```
louise-poole: I don't think we need to have this level of detail in the supported protocols doc. Just a description of the filters that can be used is enough - let's remove this last sentence talking about attributes etc (most users never have to even know what 'attributes' are).

louise-poole (approved):
Looks very good! I like it! Let's just swap out the placeholder address for the deployed one and then we're good to go. Approved in the meantime.

PR #1257, opened 2026-07-30
fix(simulation): charge forward overhead on signed exclusive Ekubo pools

tamaralipows:
A `SignedExclusiveSwap` pool reverts in `beforeSwap`, so every swap goes through `Core.forward` with a signature to recover and a nonce to write. The curve math is identical either way, so the simulated gas missed all of it and an exclusive route was costed as a direct swap.

## The number

Measured in Ekubo's harness (`SignedExclusiveSwap.t.sol`, `--isolate`), adding a plain-swap baseline so the lock, settle, fee, tick spacing, position and amount match on both sides:

| case | gas |
|---|---|
| plain swap | 81,139 |
| signed, zero fee | 118,991 |
| signed, non-zero fee | 143,512 |

`forward` + signature = 37,852; fee accounting = 24,521. The constant is the non-zero-fee delta, **62,373** — Fynd normally signs a fee above zero, and erring high is the safe direction.

## Why here, not the estimator

`estimate_gas_usage` documents the split: pool gas comes from `get_amount_out`; the estimator adds only what `get_amount_out` cannot see (transfers, approvals, router). This overhead is inside the pool call, and the estimator adds `group.estimated_gas` to its own total, so it flows through either way. Execution's per-protocol lists classify token movement, not pool execution cost.

Happy to move it if you'd rather all gas accounting live in the estimator — that needs an address in `protocol_specific_addresses.json` and a `chain` param on a public function.

## Impact

Fynd uses output net of gas both to select exclusive routes and to set the taker's committed amount, so understating it biases against the taker twice. ~19 bps on a $1,000 trade at 10 gwei.

Dormant until `SIGNED_EXCLUSIVE_SWAP_ADDRESS` is the deployed address. Found via [fynd#365](https://github.com/propeller-heads/fynd/pull/365); tracks [ENG-6283](https://propeller-heads.atlassian.net/browse/ENG-6283).

### DIFF
--- a/crates/tycho-simulation/src/evm/protocol/ekubo_v3/state.rs
+++ b/crates/tycho-simulation/src/evm/protocol/ekubo_v3/state.rs
@@ -26,11 +26,38 @@ use super::pool::{
     twamm::TwammPool, EkuboPool,
 };
 use crate::evm::protocol::{
-    ekubo_v3::pool::{
-        boosted_fees::BoostedFeesPool, mev_capture::MevCapturePool, stableswap::StableswapPool,
+    ekubo_v3::{
+        addresses::SIGNED_EXCLUSIVE_SWAP_ADDRESS,
+        pool::{
+            boosted_fees::BoostedFeesPool, mev_capture::MevCapturePool, stableswap::StableswapPool,
+        },
     },
     u256_num::u256_to_f64,
 };
+
+/// Extra gas a swap on a `SignedExclusiveSwap` pool costs over the same swap on a plain pool.
+///
+/// The extension reverts in `beforeSwap`, so every swap on such a pool must go through
+/// `Core.forward`, carrying an EIP-712 signature that the extension recovers and a nonce it writes.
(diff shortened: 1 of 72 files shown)


louise-poole (approved):
Makes sense. Looks good to me 👍

crates/tycho-simulation/src/evm/protocol/ekubo_v3/state.rs
```diff
@@ -26,12 +26,20 @@ use super::pool::{
...
+    ekubo_v3::{
+        addresses::SIGNED_EXCLUSIVE_SWAP_ADDRESS,
+        pool::{
+            boosted_fees::BoostedFeesPool, mev_capture::MevCapturePool, stableswap::StableswapPool,
+        },
     },
     u256_num::u256_to_f64,
 };
 
+/// Gas cost of `Core.forward` plus the signature check, on top of the swap itself.
```
louise-poole: Plus fee accounting no?

Troshchk (approved):
Thank youuu ✨

PR #1261, opened 2026-07-31
chore: rename for consistency

Troshchk:


crates/tycho-execution/model/src/model/tycho_router.rs
```diff
@@ -483,11 +482,10 @@ fn _validate_amounts(
     if expected_amount_out == 0 {
         return Err(Error::revert("_validate_amounts: expected_amount_out == 0"));
     }
-    // Lowest acceptable output: at most MAX_SLIPPAGE_TOLERANCE_BPS below expected.
     // The division rounds down, which can make the floor 0 for tiny expected
     // amounts — hence the explicit zero check.
     let min_amount_out_floor =
-        expected_amount_out * (BPS_DENOMINATOR - MAX_SLIPPAGE_TOLERANCE_BPS) / BPS_DENOMINATOR;
+        expected_amount_out * (100 - MAX_SLIPPAGE_TOLERANCE_PERCENT) / 100;
```
tamaralipows, line 488: I didn't understand this from Max's message, I understood that we don't want to multiply the numerator and denominator by the same value (100) and instead say `MAX_SLIPPAGE_TOLERANCE_BPS` is `X BPS` of `expected_amount_out` rather than `1-Y` of it

tamaralipows:
not to be included in the next deployment

PR #1267, opened 2026-08-03
feat(testing): build the Substreams WASM before packing

die-herdplatte:
## What this changes

The protocol test runner now compiles a Substreams package's WASM binary before packing its `.spkg`, and reports a failed pack instead of continuing past it.

**Build the WASM before packing.** `substreams pack` only reads the binary the manifest points at, and neither failure mode pointed at the build. A missing binary meant no `.spkg` was produced, so `ExtractorRunnerBuilder::ensure_spkg` fell through to the S3 bucket that `cli.rs` configures by default and the run died on `Failed to download <spkg> from s3` — an infrastructure-looking error. A stale binary was quieter still: the old `.spkg` was already on disk, `ensure_spkg` returned early, and the test indexed against an outdated module without complaint.

`build_spkg` now compiles the package first, before it writes the YAML backup, so a failure there leaves no backup file behind.

The nested `cargo build` drops three inherited environment variables:

- `RUSTUP_TOOLCHAIN`, which rustup exports for this crate and which outranks a `rust-toolchain.toml`. Keeping it would build under `protocols/testing`'s `nightly` pin, where no `wasm32` target is installed, instead of the `1.96.0` that `protocols/substreams/rust-toolchain.toml` pins together with that target.
- `CARGO_TARGET_DIR` and `CARGO_BUILD_TARGET_DIR`, either of which redirects the build output. The manifest's `binaries.default.file` is relative to the package, so a shared target directory exported by a developer or a CI runner puts the binary where `substreams pack` does not look. Cargo exports neither variable itself, so this surfaces only where someone sets one — verified by building with each variable in turn, which succeeded while leaving `pack` to fail on the missing path.

**`--prebuilt-wasm`** skips that build for callers that already hold the binaries. The runtime stage of `run.Dockerfile` compiles every package in a builder stage and ships no Rust toolchain, so `entrypoint.sh` passes the flag. `run.Dockerfile` itself is unchanged.

**Fail on a failed pack.** A non-zero `substreams pack` exit only logged at `error!` level and the run continued with whatever `.spkg` happened to be on disk. It now returns an error carrying the pack stderr. The original YAML is restored from backup before the error is reported, so a failed pack does not leave the manifest holding a rewritten initial block.

**`TestRunner::new` takes a `RunnerConfig`.** The flag would otherwise have been a tenth positional argument beside two other `bool`s, which a caller can transpose without a type error. This drops the constructor's `too_many_arguments` exemption.

**`--execution-traces` is gone.** Nothing ever read it and no execution tracing hook exists, so passing it did nothing. Note this is user-visible: the flag is now rejected rather than accepted and ignored.

## Docs

The testing guide no longer lists the manual WASM build as a step before running tests, since the runner does it. The command stays on the page as a note, for driving a package directly with `substreams run` or `substreams gui`, pointing at the implementation guide that already pairs it with those tools.

`--prebuilt-wasm` is not documented there — the container entrypoint sets it, not people.

## Testing

`cargo fmt --check` (workspace), `cargo clippy -p protocol-testing --all-features --all-targets`, and `cargo nextest run -p protocol-testing --all-features` (7 tests) pass. Each commit compiles on its own.

Checked by hand against `ethereum-uniswap-v2`, which has no compiled WASM in the worktree:

| Invocation | Result |
|---|---|
| `full --package ethereum-uniswap-v2` | logs `Building WASM binary in …`, runs a real `cargo build` |
| same, plus `--prebuilt-wasm` | logs `Expecting a pre-built WASM binary in …`, then fails on `substreams pack` |

The second row is the new pack-failure path: with the flag set and no binary present, packing fails and the run stops.

No unit test covers `build_wasm`. Exercising the build branch means invoking `cargo` from a test, and the skip branch is a single `if`.

### Container path

Built the image locally with `PROTOCOLS=ethereum-uniswap-v2` and ran the range test through `entrypoint.sh`. It passed `1/1`, having skipped the build, packed the prebuilt binary, indexed blocks 19701395–19701397, and run 6 execution simulations with 0 failures.

The image reproduces the state the flag exists for: the package directory holds no `Cargo.toml`, the prebuilt `.wasm` is present, and there is no `cargo` on `PATH`. Running the binary there without the flag fails with `Failed to run cargo build for the Substreams package: No such file or directory`, which is the failure the explicit flag rules out.

Worth noting for reviewers: CI will not exercise that path on this PR. `ci-substreams-integration.yaml` triggers on `protocols/testing/**`, but `detect-changes` only selects target protocols when `protocols/substreams/**`, `run.Dockerfile`, or `protocols/testing/Cargo.toml` changed — none of which this PR touches. `target-protocols` comes out empty and `skip-tests` runs green, so nothing builds the image. Hence the local run above.

## Not covered

The full local CI suite has not been run on this branch — only the checks listed above.


docs/for-dexs/protocol-integration/3.-testing.md
```diff
@@ -193,31 +193,23 @@ export SUBSTREAMS_API_TOKEN=eyJhbGci...
 export RUST_LOG=protocol_testing=info,tycho_client=info,tycho_indexer=error,error
 ```
 
-**Step 2: Build the substreams wasm file**
-
-If you do not have one already, you must build the wasm file of the substreams package you wish to test. This can be done by navigating to the substreams package directory and running:
-
-```bash
-cargo build --target wasm32-unknown-unknown --release
-```
```
zach030, line 202: I'd suggest keeping this part. Aside from running the protocol testing, we sometimes need to manually run substreams run to test, I think this step should stay in the docs.

protocols/testing/src/utils.rs
```diff
@@ -9,12 +9,55 @@ use figment::{
...
+    }
+
+    info!("Building WASM binary in {package_dir}");
+    let status = Command::new("cargo")
+        .args(["build", "--release", "--target", "wasm32-unknown-unknown"])
+        .current_dir(package_dir)
+        // RUSTUP_TOOLCHAIN outranks a rust-toolchain.toml, so the value rustup exported for this
+        // crate would build the package with its nightly toolchain, which carries no wasm32
+        // target, instead of the version the Substreams workspace pins.
+        .env_remove("RUSTUP_TOOLCHAIN")
```
zach030, line 28: we could also add `env_remove("CARGO_TARGET_DIR")` ?

zach030 (commented):
Thanks! It really is a pain point. Left a few comments, please take a look.

docs/for-dexs/protocol-integration/3.-testing.md
die-herdplatte: Yes, left it in, but not as a separate step, but as a hint

protocols/testing/src/utils.rs
die-herdplatte: Added that and `CARGO_BUILD_TARGET_DIR`

zach030 (approved):
LGTM 🚀

PR #1268, opened 2026-08-03
fix(protocol-testing): read initialBlock from anchored substreams manifests

die-herdplatte:
## Summary

- read `initialBlock` by parsing the substreams manifest as YAML instead of matching it with a regex, so anchored declarations (`initialBlock: &initial_block 13843704`) and aliases (`initialBlock: *initial_block`) are picked up — the `full` test refused to start on packages like `base-aerodrome-slipstreams` unless `--initial-block` was passed
- derive the start block as the lowest across modules, reading both inline declarations and the per-module entries under `networks.<network>.initialBlock`, which previously had to be passed by hand for the four packages that only use the latter
- apply a requested start block to both spellings, and only to modules that declare one
- pack with `substreams pack -`, piping the manifest in, so the checked-in manifest is never rewritten
- fail the run when `substreams pack` fails, instead of returning the path of a package that was not built
- parse manifests and test configs with `serde_yaml` and drop the `figment` dependency

## Context

`build_spkg` rewrote the manifest on disk to pack it: backup, set every module's `initialBlock`, pack, restore. That had two problems beyond the regex. Any early return between the write and the restore left the manifest modified with a stray `.backup` beside it, and the rewrite ran even when no start block had been asked for, pulling modules that legitimately start later back to the earliest block. `substreams pack -` reads the manifest from stdin, so the file no longer has to be touched; relative paths inside a manifest resolve against the working directory, so pack runs from the directory holding it with the output file named explicitly. Packing from stdin was checked to produce a byte-identical spkg to packing the file, both with the local CLI (1.20.1) and the version pinned in `run.Dockerfile` (v1.16.4).

`ethereum-fluid`, `ethereum-ekubo-v2`, `ethereum-ekubo-v3` and `ethereum-template-singleton` declare their modules' start blocks only under `networks.<network>.initialBlock`, so both the derivation and the override have to look there. Packing fluid's manifest shows those entries taking precedence over an inline `initialBlock` — an inline 12345 is ignored while the network says 19239106, on both CLI versions — which the manifest reference does not state either way. Injecting the block into every module, as the old code did, was therefore ineffective for these packages: it pinned the stores while leaving the source map at 19239106. Rewriting the network entries pins all six of fluid's modules at the requested block.

Modules that leave `initialBlock` implicit are left alone, since the reference defines them as inferred: "If all the inputs have the same initialBlock, the field can be omitted and its value is inferred by its dependent inputs." Equalising the declared blocks can only make a module's inputs agree, so this cannot make an omitted `initialBlock` mandatory. Packing the aerodrome manifest with and without the extra keys gives identical `substreams info` output, module hashes included.

For the parser swap, all 31 `integration_test*.tycho.yaml` files were checked to deserialize into `IntegrationTestsConfig` identically under figment and serde_yaml.

## Testing

Unit tests cover the start block derivation for anchors, aliases, the lowest-of-several case, network blocks, network-over-inline precedence, unselected networks and a manifest declaring none; and the override for network blocks and for leaving implicit modules alone. The alias fixture anchors outside `modules` so the aliased value is the only way to reach the expected block — anchoring on a module's own `initialBlock`, as real manifests do, passes whether or not aliases resolve.

All 56 manifests under `protocols/substreams` resolve a start block with the derivation in this branch.

Ranged test, `cargo run -- range --package ethereum-uniswap-v2` (blocks 19701395–19701397): passed 1/1. Components and static attributes matched, both token balances matched onchain, spot price and limits resolved, 6/6 execution simulations passed. The packed spkg carried the test's start block on all modules.

Full test, `cargo run -- full --package base-aerodrome-slipstreams --chain base` with no `--initial-block`, aborted after ~45s: the start block was derived as 13843704 from the anchored manifest — the case that fails on `main` — the spkg packed, and the indexer streamed from there to block 13845998 with a cursor recorded, so the spkg is accepted by the Substreams endpoint. The manifest was byte-identical afterwards with no `.backup` beside it, and all eight modules in the spkg kept their declared 13843704.


protocols/testing/src/utils.rs
```diff
@@ -1,147 +1,278 @@
...
+    else {
+        return;
+    };
+
+    for module in modules {
+        if let Some(initial_block) = module.get_mut("initialBlock") {
+            *initial_block = start_block.into();
         }
     }
 }
```
zach030, line 216: This breaks manifests that declare `initialBlock` only under `networks`: (fluid, ekubo-v2/v3, template-singleton): no module declares it inline, so the override is a silent no-op. E.g. in fluid's ranged test, stores now derive 19239106 from networks: instead of being pinned to the test's start block 21609290, so store state gets built across ~2.37M extra blocks.

die-herdplatte: Nice one, fixed!

zach030 (approved):
LGTM!

PR #1270, opened 2026-08-04
feat(erc4626): add svZCHF (Frankencoin Savings Vault)

aummfi-bit:
Adds the Frankencoin Savings Vault (svZCHF) to the generic `ethereum-erc4626` substream.

| | |
|---|---|
| vault | `0xE5F130253fF137f9917C0107659A4c5262abf6b0` — svZCHF, 18 dec |
| asset | `0xB58E61C3098d85632Df34EecfB899A1Ed80921cB` — ZCHF, 18 dec |
| created | block `24118106`, tx `0x100220050204815b4445040fee1974455a4f43e5f78b7f666733a0bf4cda16e4` |

Creation block and tx were located by binary search on `eth_getCode` and confirmed against the receipt's `contractAddress`.

## Why it fits this stream

The Readme requires "any vault that follows the ERC4626 event and value semantics". The verified source (`contracts/vault/SavingsVaultZCHF_2.sol`) is `SavingsVaultZCHF_2 is ERC4626, Ownable2Step`, on OpenZeppelin's ERC4626:

- `_deposit` emits `Deposit(caller, receiver, assets, shares)` — `assets` is the ZCHF transferred in, `shares` the amount minted.
- `_withdraw` emits `Withdraw(caller, receiver, owner, assets, shares)`, same semantics.

Neither documented exception applies: it is not an Aave `StataToken` (no wrapped-aToken amounts in events), and unlike Ethena its `redeem` is enabled with no cooldown. Structurally it is the same shape as the already-supported **sDAI** and **sUSDS** — a savings vault over a stablecoin accruing interest via `totalAssets()`.

Context on size: the Frankencoin savings module this wraps holds ~12.2M ZCHF. The ERC-4626 wrapper itself is new (deployed 2025-12-29) and lightly used so far, because savers interact with the module directly through the Frankencoin app and only contracts need the tokenised form. Indexing it is what makes the wrapper reachable from solvers and routers.

## Changes

Config only, following the Readme's "Adding an ERC4626 pool" steps:

- `params.json` — svZCHF entry
- `substreams.yaml` — regenerated `map_protocol_components` via `python params.py`
- `integration_test.tycho.yaml` — `test_svZCHF`, 50k-block window matching the enabled `test_sUSDS`

## Test status — quota-limited, no assertion failure

Ran `range --package ethereum-erc4626` against an archive RPC. `test_sUSDS` completed; `test_svZCHF` then indexed `24118106 → 24158650` of `24168106` (~81%) before the Substreams free-tier egress quota stopped the stream:

```
ResourceExhausted: billable egress bytes quota exceeded
  (quota 5368709120, current 5914094370)
```

Nothing asserted false — the run was cut off by billing, not by a defect. Egress runs ~60 kB/block, so a 50k-block test costs ~3 GB and a 5 GB free tier cannot fit two tests in one run. Happy to post a full green run before merge if you'd like one.

Two notes from getting the harness running, both unrelated to this change and not included here:

1. `protocol-testing/postgres.Dockerfile` does not build on current base images. It installs `clang19 llvm19`, but `ghcr.io/dbsystel/postgresql-partman:15-5` now ships Alpine 3.24 (which has `clang16`), and PGXS then requires `clang-21` for JIT bitcode. Building pg_cron with `make with_llvm=no` avoids the version chase entirely.
2. The Substreams endpoint is hardcoded in `protocol-testing/src/tycho_runner.rs` with no env override, so testers can't point at an alternate provider when a quota runs out.


aummfi-bit:
Heads-up on CI: all checks are green, but `test-protocols` **skipped**, so this PR's integration test never ran. I think that's a bug affecting every PR rather than anything specific to this change.

From the `detect-changes` job on this PR ([run 30910408267](https://github.com/propeller-heads/tycho/actions/runs/30910408267)):

```
HEAD_SHA: 886655fc2ecd5336b7b42f944df5e18f4a2caea5
Changed substreams:
Target protocols:
```

`886655fc…` is the current tip of `main`, not this PR's head.

`.github/workflows/ci-substreams-integration.yaml` triggers on `pull_request_target`, where `github.sha` is the **base branch tip** rather than the PR head. The checkout step accounts for this:

```yaml
ref: ${{ github.event.pull_request.head.sha || github.sha }}
```

but the *Get changed substreams folders* step passes plain `${{ github.sha }}`:

```yaml
env:
  HEAD_SHA: ${{ github.sha }}
```

so its `git diff --name-only --diff-filter=d origin/main "$HEAD_SHA"` compares `main` against `main` and returns nothing. `changed-substreams` comes out empty, `target-protocols` comes out empty, and `test-protocols` is gated out by `if: needs.detect-changes.outputs.target-protocols != ''`.

Note `substreams-changed` is `true` — the path-filter step does see the change; only the folder-level diff misses it.

The fix looks like one line, matching what checkout already does:

```yaml
env:
  HEAD_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
```

Happy to open that as a separate PR if useful.

**In the meantime**, this PR can be tested via `workflow_dispatch` with `protocols: ethereum-erc4626` — `INPUT_PROTOCOLS` takes precedence over the detection, so it runs on your RPC and Substreams quota. That would give the green run my local attempt couldn't finish (it reached ~81% before hitting the free-tier egress ceiling, with no assertion failure).

aummfi-bit:
Rebased onto current `main` (`766d189`) — content unchanged, all three files byte-identical to the previous head.

**Integration test: passing.** Since `test-protocols` is skipped on this PR (see below), I ran it locally via the Docker path against live mainnet:

```
PROTOCOLS="ethereum-erc4626=test_svZCHF" MODE=range docker compose up --build --abort-on-container-exit
```

```
INFO Running Tycho indexer from block 24118106 to 24168106...
INFO extractor:substreams_waiting: SubstreamSessionInit session.resolved_start_block=24118106
INFO extractor:substreams_waiting: Stream completed, reached end block
INFO Fetching protocol data from Tycho with stop block 24168106...
INFO Component 0xe5f130253ff137f9917c0107659a4c5262abf6b0 matches the expected state
INFO All expected components were successfully found on Tycho and match the expected state
INFO ✅ test_svZCHF passed
INFO Tests finished!
INFO Passed 1/1
```

Full 50,000-block range indexed against a Chainnodes archive node, ~23 minutes. The component resolves with both expected tokens (svZCHF + ZCHF `0xb58e61c3098d85632df34eecfb899a1ed80921cb`) and the `svZCHF` name attribute. `skip_execution: true` matches the other entries in this package, so balance/execution checks are skipped as expected.

---

**Note on CI:** `test-protocols` was skipped on this run, as it is on every PR touching `protocols/substreams/**`. This is the `HEAD_SHA` issue from #1271 — under `pull_request_target`, `github.sha` is the base tip, so `detect-changes` diffs `origin/main` against itself and `target-protocols` comes out empty.

Checking the workflow's run history: it has no `push:` or `schedule:` trigger, and of the last 40 runs every one was `pull_request_target` with zero `workflow_dispatch`. So as far as I can tell this integration suite hasn't run automatically for some time — I looked at 9 recently merged PRs touching `protocols/substreams/**` and `test-protocols` was skipped on all of them.

The fix looks like one line, matching what the checkout step already does:

```yaml
env:
  HEAD_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
```

Happy to open that as a separate PR if it's wanted — though it would start consuming RPC/Substreams quota on every substreams PR, so that trade-off is yours to make rather than mine. If you'd prefer to verify this PR through CI directly, `workflow_dispatch` with `protocols: ethereum-erc4626` bypasses the detection entirely.

PR #1273, opened 2026-08-04
fix: scope contract snapshot queries to the returned page

zizou0x:
## What this changes

`get_contracts` applies LIMIT/OFFSET to the `account` query, but then passed the request's original `ids` to `get_account_balances` and `get_contract_slots`. When `ids` is `None` — which is what a full snapshot request sends — both of those read the entire chain's balances and slots on every page. They now receive the addresses of the accounts on the returned page, so the work per page is proportional to the page.

For a 1200-account page against 10.5M live storage slots, the slots query went from reading the whole `contract_storage` set to reading only the page's accounts. The `codes` query in between already used the paginated account ids.

Second change, in the RPC layer: `get_contract_state_inner` slices `contract_ids` to the requested page before calling the gateway, and also passed the real `pagination_params` to it. The database then applied the offset a second time to a list that was already at most one page long, so any request with explicit `contract_ids` and `page > 0` came back with no accounts from the database. It now passes `None` when it has already sliced the list; the response `total` is unaffected because for the ids path it is the length of the requested address list.

`tycho-client` was never affected by that second bug: `get_contract_state_paginated` chunks ids client-side and always sends `page = 0`. Callers using the HTTP API directly with `contract_ids` plus paging were.

## Semantics of `total`

- No `contract_ids`: the count of accounts that have code at the requested version, from the count query.
- With `contract_ids`: the number of requested addresses, set by the RPC layer.

The gateway docstring now states that callers must pass the full address list and let the gateway cut the page, since the two cannot be combined.

## Tests

- `test_get_contract_state_second_page` (tycho-indexer, mock gateway, no database needed) asserts the gateway receives the page's address and no pagination params. It fails on the previous code with `Some(PaginationParams { page: 1, page_size: 1 })` instead of `None`.
- `test_get_contracts_pages_cover_all_accounts` (tycho-storage) walks all pages of a full snapshot request and checks each page returns its own accounts and a stable total. Page > 0 had no coverage before.

Verified: `cargo clippy -p tycho-storage --lib` and `cargo clippy -p tycho-indexer --lib --tests` are clean; 339 tycho-indexer lib tests pass. The tycho-storage tests and 8 tycho-indexer tests could not run in my environment (no Postgres, and a substreams spkg fixture that needs git-lfs), so the new storage test has not been executed against a database.

### DIFF
--- a/crates/tycho-indexer/src/services/rpc.rs
+++ b/crates/tycho-indexer/src/services/rpc.rs
@@ -294,8 +294,7 @@ where
         // When addresses are given, the slice above already is the page, so the db must not
         // paginate again - its offset would be applied a second time and skip the whole page.
         // Without addresses the db owns pagination and also reports the total page count.
-        let db_pagination =
-            if paginated_addrs.is_some() { None } else { Some(&pagination_params) };
+        let db_pagination = if paginated_addrs.is_some() { None } else { Some(&pagination_params) };
 
         // Get the contract states from the database
         let account_data = self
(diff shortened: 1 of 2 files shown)


brunoguerios (approved):
LGTM ✅

PR #1276, opened 2026-08-04
feat(balancer_v3): native quoting and multi-generation factory indexing

zach030:
Two halves that have to land together, because they meet at one static attribute.

**Simulation** — quote `vm:balancer_v3` through Balancer's own Rust maths instead of executing the Vault in REVM for every quote.

**Indexing** — index every factory generation per pool family, rather than one factory each. Ethereum was pointing at the deprecated December 2024 weighted and stable factories, so no pool built by a later factory was indexed at all.

The generation travels in the `pool_type` attribute, which the simulation side parses.

---

## Simulation: native quoting

`vm:balancer_v3` was the slowest protocol on the dashboard at **11.3 ms per `get_amount_out`**, against 2.7 ms for balancer_v2 and single-digit µs for the native ones: every quote walked adapter → BatchRouter → Vault → VaultExtension → pool → each rate provider, re-reading `getPoolTokens` each time. Measured here: **~1.8 µs**.

The shape follows Curve's migration. Indexing stays VM-based, the decoder resolves the pool family and reads state through the pool's own getters, and `delta_transition` rebuilds that state rather than patching it. Rate providers keep working untouched — the DCI tracks their storage and the getters execute against it.

The trade is that update-time work grows: every V3 pool re-reads its getters whenever any V3 pool moves, because the Vault is a linked contract on every component. Worth it while quotes outnumber updates.

### Coverage

| Family | Native | Notes |
|---|---|---|
| Weighted | yes | |
| Stable | yes | including pools mid amplification update |
| reCLAMM | yes | the generations we index reuse the library's V2 maths, which Balancer confirmed |
| QuantAMM | yes | weights interpolate with time, matched on its own factory |
| Swap-hook pools | rejected | a pool's factory says nothing about its hooks — see below |
| Gyro / LBP | rejected | maths exists upstream, decoder does not build it yet |

Rejection happens at decode time, so an unsupported pool never reaches the router with a wrong number.

### `get_limits`

The first version reported whatever the maths would solve — the entire reserve of a reCLAMM pool, three times the reserve of a stable pool — which would have handed the router sizes that revert on chain. Each family is now bounded by the check its pool contract runs:

| Family | Bound |
|---|---|
| Weighted | `MAX_IN_RATIO` (30% of the sell reserve), and where the pool registers one, the `MinTokenBalanceLib` minimum inverted through the exact-out formula. Tighter wins. |
| Stable | binary search against `ensureBalancesWithinMaxImbalanceRange`, which rejects a swap leaving `maxBalance / minBalance >= 10_000`. Applied to every generation, though only the v3 factory's pools enforce it on chain: without it the bound degenerates to all but draining the output side. |
| reCLAMM | the input that buys `MAX_TOKEN_OUT_RATIO` of the output reserve at the current virtual balances. |
| QuantAMM | binary search against the Vault path, since `maxTradeSizeRatio` bounds the output side too. |

`balancer-maths-rust` models the December 2024 genesis contracts, so the stable imbalance ratio and the weighted minimum balance — both added by later factory generations — have no counterpart upstream and are mirrored here.

`spot_price` still differs from the adapter by ~0.1% and is left alone: the adapter averages execution price over 0.3% of the reserve, this probes 1e-6 of it. Anything calibrated against the old figure may need rechecking.

---

## Indexing: every factory generation

`weighted_factory`, `stable_factory` and `reclamm_factory` become maps from version label to address:

```
weighted_factories[v1]=201efd…&weighted_factories[v2]=332694…
```

`create` and `PoolCreated` are byte-identical across every generation of these three families — checked against every artifact in the Balancer deployments repo — so one decoder covers them all and a new generation is configuration only.

All four chains now name every deployed weighted and stable generation: two weighted, three stable each. The gaps ran in opposite directions — Ethereum had only the oldest generation of each family, the three L2s only the newest. QuantAMM is configured the same way on every chain but Gnosis; mainnet's factory sits at a different address than Base and Arbitrum, from a different deployment task. Its `create` parameters arrive as one seventeen-element tuple, past the twelve Rust derives `Debug` and `PartialEq` for, so the build strips those derives from the generated bindings.

**reCLAMM stays pinned to the newest generation.** `balancer-maths-rust` carries the first generation as a separate `PoolState` variant with its own `on_swap`, and Balancer confirmed only that the second and third share their maths. Adding the first generation's address would price those pools with the wrong curve, silently.

For anyone extending this: `WeightedPoolFactory` and `StablePoolFactory` are also **Balancer V2** factory names, with a different `create` signature. Only tasks whose id contains `-v3-` belong here.

`pool_type` names the family alone, e.g. `WeightedPoolFactory`. The version label stays in the deployment params, where it names the generation for whoever reads the manifest and keeps one factory from being configured twice; it reaches no consumer, because a family prices the same whichever generation built the pool. Where generations do differ, the difference is resolved without the label — see `get_limits` above. Parsing rejects a factory configured under two labels, or an address that is not 20 bytes; `map_components` fails when no factory is configured at all.

---

## Verification

**Offline parity** (`cargo test -p tycho-simulation balancer_v3`, no RPC): 15 mainnet pools, 145 swaps, both directions, sizes from 0.01% of pool depth up to 50% — every one reproduces `BatchRouter.querySwapExactIn` **to the wei**. Committed with its regeneration script, so it doubles as a regression baseline. One caveat on reach: only one of the five reCLAMM pools sits in the time-dependent branch, so 48 of the 60 reCLAMM swaps would answer the same at any timestamp. That one is also the fastest-moving of the five.

**Live sweep against a full local resync** (mainnet, 637 pools indexed from the V3 genesis block, no TVL floor): 184 blocks, 2257 quotes, 1875 executions, all of them in the `slippage <= 0` bucket — the chain never returned less than the quote. Three reverted, all on one pool holding $16; see Still open.

That sweep is also what the `balancer-maths-rust` 0.5.0 bump buys. `get_limits` reported 92 recoverable failures, every one `no usable price range` on three reCLAMM pools holding 1 wei per token, where the invariant rounds to zero. Under 0.4.x that state panicked and took the quoting thread with it; 0.5.0 returns `ZeroInvariant`, surfaced here as a recoverable error. The same release drops 46 silent `unwrap_or(U256::ZERO)` fallbacks and types the stable zero-balance and exact-out overflow cases — all reported from this work.

**Protocol integration tests** now run through the native state rather than the swap adapter: `vm:balancer_v3` is registered in the testing framework and `adapter_contract` is gone from both manifests. 10 cases on Ethereum, all passing — weighted v1 (creation + three quote/execution cases), reCLAMM v3, and six new ones covering weighted v2, stable v2, stable v3 and QuantAMM. Without those a wrong address in the new configuration would have indexed nothing and gone unnoticed. Executed amounts match the native quotes to the wei.

The runs are flaky against a rate-limited RPC: cases intermittently fail with an HTTP 503 fetching a block header or contract code, and pass on a rerun with identical numbers.

---

## Deploying this

**A full resync is required.** The parameter names, the `pool_type` format and the factory set all change, and pool immutables — weights, rate providers — now travel as static attributes instead of being read back through the Vault. A component indexed by an earlier package carries neither those nor a family-only `pool_type`, and is rejected at decode time.

Substreams package goes to **v0.6.0**; `docker/extractors.yaml` needs the new spkg path.

---

## Still open

- **The stable limit sits on the imbalance boundary.** Three of 1875 executions reverted with `MaxImbalanceRatioExceeded`, all on one pool holding $16, at the amount `get_limits` reported. The rule matches the deployed `StableMath` — accept while `maxBalance < 10_000 * minBalance`, equivalent to its `>= 10_000` revert under integer division — so what remains is last-digit rounding between this reproduction of the post-swap balances and the Vault's, enough to flip the check on a pool already near 10,000:1. Backing off by a wei would close it, at the cost of marginally smaller quotable size on every stable pool. Left as is.
- **A Vault-wide pause is not modelled.** Per-pool pause and recovery mode come from `getPoolPausedState`; `_ensureVaultNotPaused` has no counterpart here.
- **Hooked pools drop out of the stream** rather than falling back to the VM adapter. Deliberate for now — `balancer-maths-rust` ships `hooks::stable_surge`, so supporting them natively is the better path.
- **The `block_timestamp` a reCLAMM quote uses only advances when the component receives a delta.** Any V3 activity refreshes every V3 pool, so the lag is small, but it is a regression against the adapter, which read the timestamp from global engine state. Unmeasured.

## Two findings worth knowing

**A pool's factory says nothing about its hooks.** A `StablePoolFactory` pool carrying a dynamic-fee hook was quoted with the static fee, and the execution reverted with `DynamicSwapFeeHookFailed`. The decoder now reads `Vault.getHooksConfig` and rejects any hook that takes part in swapping.

**reCLAMM quotes depend on when you ask.** It is the first family here whose maths recomputes virtual balances from the pool's last update to the block being quoted, so `BalancerV3State` carries the block timestamp — taken from the header at decode time, refreshed from each delta. A test asserts a quote moves when the timestamp advances, because losing that plumbing would not fail loudly: quotes would freeze at whatever moment the pool was decoded.


zach030:
## What the factory expansion actually buys — measured against the Balancer API

To quantify the indexing half of this PR, I pulled every protocol-v3 pool on our four chains from the [Balancer API](https://api-v3.balancer.fi/graphql) (2,052 pools total) and attributed each to a `(chain, factory)` pair, matched against the factory sets before and after this PR. Attribution is purely by factory — no rate-provider or hook filtering — so these are upper bounds on what actually gets indexed on the L2s, where `skip_rate_provider_pools=true` still applies.

### Pools and TVL by family and generation (all four chains)

| Family | Gen | Status | Pools | TVL (USD) | 24h volume | Before | After |
|---|---|---|---:|---:|---:|:-:|:-:|
| weighted | v1 | deprecated | 266 | $24,643,485 | $7,070 | partial¹ | ✓ |
| weighted | v2 | active | 71 | $295,922 | $533 | partial¹ | ✓ |
| stable | v1 | deprecated | 42 | $1,630,698 | $549,825 | partial¹ | ✓ |
| stable | v2 | deprecated | 54 | $5,869,077 | $45,028 | — | ✓ |
| stable | v3 | active | 26 | $36,892 | $390 | partial¹ | ✓ |
| reclamm | v1 | deprecated | 1 | $6 | $0 | — | — |
| reclamm | v2 | deprecated | 17 | $295 | $0 | — | — |
| reclamm | v3 | active | 15 | $142,739 | $71,479 | ✓ | ✓ |

¹ *Before this PR each chain named exactly one factory per family, and the generations differed by chain: Ethereum had weighted v1 + stable v1, the L2s had weighted v2 + stable v3. So each of these rows was indexed on some chains and missed on others.*

### Before vs. after

| | Pools | TVL | 24h volume |
|---|---:|---:|---:|
| Before | 143 | $2,384,231 | $617,425 |
| After | 474 | $32,618,813 | $674,325 |
| **Added** | **+331 (3.3×)** | +$30,234,582 | +$56,901 (+9.2%) |

### The headline TVL number is misleading

$23.8M of the added TVL — 79% of it — is a single pool: `0x83470106…` ("jujukiki", Base, weighted v1). One holder, zero lifetime volume, and 99.99% of its "TVL" is a self-priced token; the WETH side holds **$8**. It is the same self-valued-token pattern we hit on Curve. It will be indexed (it is a legitimate weighted v1 pool), it just carries no routable depth.

Excluding it:

| | Pools | TVL | 24h volume |
|---|---:|---:|---:|
| **Added (real)** | **+330** | **+$6,396,334 (3.7×)** | +$56,901 (+9.2%) |

The volume gain is small because the previously-indexed stable v1 factory already carried 89% of these families' volume ($549k of $617k/day). The single biggest real gain is **stable v2: 54 pools, $5.87M TVL** — the only generation with meaningful liquidity that was indexed nowhere before.

### reCLAMM v2 stays excluded, and the data confirms that costs nothing

reCLAMM v1+v2 together hold 18 pools with **$301 of TVL and zero volume**. Pinning reCLAMM to v3 (whose maths `balancer-maths-rust` verifiably implements) forgoes nothing measurable.

### A caution for anyone reproducing this

Balancer reuses deployment addresses across chains for *different* contracts. Two collisions inside our own factory set:

- `0x201efd50…` — weighted v1 on Ethereum, **reCLAMM v2 on Base**
- `0xeb1eeabf…` — stable v1 on Arbitrum, **weighted v1 on Gnosis**

Attribution must be by `(chain, address)`, never by address alone. My first pass got this wrong and mis-attributed a $266M StableSurge pool (Stream Surge xUSD-vgUSDC, `0xae255db0…` — a family we don't support at all, and itself a dead pool: 7 holders, zero volume, 99.98% one collapsed token) to reCLAMM v2. The numbers above are chain-aware.

### Not counted here

1,531 of the 2,052 v3 pools belong to families outside this PR's scope (Gyro, QuantAMM, LBP, StableSurge, …). They total **$2.6M TVL** combined.

crates/tycho-simulation/src/evm/protocol/balancer_v3/state.rs
```diff
@@ -0,0 +1,424 @@
...
+        Ok(GetAmountOutResult::new(
+            u256_to_biguint(amount_out),
+            SWAP_GAS
+                .to_biguint()
+                .expect("u64 fits in BigUint"),
+            Box::new(new_state),
+        ))
+    }
+
+    fn get_limits(
```
brunoguerios, line 648: There is an analytical way to get swap limits, so you don't have to probe/iterate. I checked and the rust implementation doesn't have it available, but it's already implemented in the TS version, so it should be quite straightforward to port to Rust. If you don't have bandwidth to contribute directly, ask them for it. I'd expect them to deliver this to you quite quickly.

brunoguerios: [Here](https://github.com/balancer/balancer-maths/blob/9f91ab369c3748a49323447e3f29f04f254d21c9/typescript/src/vault/vault.ts#L112) is the method in the TS implementation.

crates/tycho-simulation/src/evm/protocol/balancer_v3/state.rs
```diff
@@ -0,0 +1,492 @@
...
+        token_out: &Bytes,
+    ) -> Result<U256, SimulationError> {
+        let input = SwapInput {
+            amount_raw: amount_in,
+            swap_kind: SwapKind::GivenIn,
+            token_in: Self::token_key(token_in),
+            token_out: Self::token_key(token_out),
+        };
+        Vault::new()
+            .swap(&input, &PoolStateOrBuffer::Pool(Box::new(self.state.clone())), None)
```
die-herdplatte: `Vault#swap` only needs a reference, yet we have to clone the state here to satisfy their interface.
There is also `vault::swap::swap` which we could use here to avoid the clone of the `PoolState`.

crates/tycho-simulation/src/evm/protocol/balancer_v3/vm.rs
```diff
@@ -0,0 +1,677 @@
...
+        LiquidityManagement liquidityManagement;
+        uint256 staticSwapFeePercentage;
+        uint256 aggregateSwapFeePercentage;
+        uint256 aggregateYieldFeePercentage;
+        uint40 tokenDecimalDiffs;
+        uint32 pauseWindowEndTime;
+        bool isPoolRegistered;
+        bool isPoolInitialized;
+        bool isPoolPaused;
+        bool isPoolInRecoveryMode;
```
die-herdplatte, line 56: Shouldn't we also respect these two fields?

die-herdplatte (approved):
lgtm!

zach030:
Split into two so the simulation half can land on its own:

- #1326 — native quoting. Reads the `pool_type` attribute the deployed Substreams package already writes, so it needs no resync and no package release. Verified against tycho-dev: 1758 quotes, 1777 executions, no reverts.
- #1327 — every factory generation per pool family, plus QuantAMM. Stacked on #1326; needs a full resync.

The two branches together are byte-identical to this one. @die-herdplatte your two review comments were addressed here and carried over — the getter-count question is answered by the immutables now coming from the indexed component, and the pause check calls `getPoolPausedState` rather than approximating the window. Happy to re-open the threads on #1326 if you'd rather follow them there.

PR #1278, opened 2026-08-04
feat(execution): fall back to Uniswap V3 when a pAMM reverts

tamaralipows:
Stacked on #1212. Fynd side: propeller-heads/fynd#402

## Problem

A pAMM gets its price from a maker quote stream, which reaches the chain only in the block the maker quotes for. Simulate a pAMM route against a mined block, and the venue reverts. Integrators drop the quote.

## Change

Send the pAMM swap through the live PropAMMRouter at `0x4DdF368080CD7946db5b459aD591c350158175e1`. It calls the venue in a try/catch block, and falls back to one Uniswap V3 pool if the venue reverts.

We write no fallback code and do not touch the TychoRouter, so this needs no audit and no router deployment. A TychoRouter-level fallback covering all protocols needs both, and a longer timeline.

## We know which Uniswap V3 pool the fallback uses

`resolvedFee(tokenIn, tokenOut)` is a public view function, and the fallback is a single-hop `exactInputSingle`:

```
PropAMMRouter.resolvedFee(WETH, USDC)  -> 500
UniV3Factory.getPool(USDC, WETH, 500)  -> 0x88e6A0c2dDD26FEEb64F039a2c41296FcB3f5640
```

So integrators need no wide slippage value: the pool is known before the swap, and Fynd prices it in propeller-heads/fynd#402. That PR is required — `minAmountOut` priced off the pAMM quote alone reverts whenever the fallback pool's price sits outside the caller's slippage tolerance.
 
**bopAMM does not work.** The whitelist does not contain its venue, and the router authors state that Bebop does not work with the PropAMMRouter.

**A venue outside the whitelist is worse than not doing this.** Every swap then executes on Uniswap V3 at a worse price than the pAMM gives.

**Large trades get weak slippage protection.** One pool in one fee tier moves 60 bps on 1000 WETH to USDC (about $1.87M), and 7 bps on $2M USDC to USDT. The router cannot correct this. Fynd needs a cap on trade size or on the bps gap. Open decision.

**Governance can change the pool.** `setPairFee`, `setFallbackFee` and `setFallbackSwapRouter` need no upgrade, so Fynd refreshes the fee tiers.

**Native ETH does not work** as `tokenIn` or `tokenOut`.

## Before merge

Wait for the executor's timelock. It is deployed at `0xA2F9f6f4a3F64EDa6014ce71A8ec8A874d690ADF` and set on the TychoRouter with an activation timestamp of 1787424959; swaps revert `Dispatcher__ExecutorIsTimelocked` until then.


crates/tycho-execution/contracts/test/protocols/PropAMMFallback.t.sol
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /// The venue is inactive at this block, so a direct call would revert.
+    function testVenueQuoteAtForkBlock() public {
+        vm.expectRevert();
+        IPropAMM(FERMI_PROPAMM_VENUE).quote(WETH_ADDR, USDC_ADDR, 1 ether);
+    }
+
+    /// The whole point: the leg still delivers tokenOut, at the Uniswap V3 price.
+    function testSwapWithStaleVenue() public {
```
tvinagre, line 106: The PR's central claim — a stale pAMM venue reverts and the route completes via the Uniswap V3 fallback — is never exercised through the TychoRouter/Dispatcher. `testSwapWithStaleVenue` deals WETH directly to the executor and calls `executor.swap()` directly, so the Dispatcher's `_transfer`/`_approveIfNeeded`/balance-diff machinery never runs. The integration that is actually new here (ProtocolWillDebit approve of a third-party router, balance-diff measurement at the receiver, `_revokeUnconsumedApproval`, mid-route output landing at the router via fundsExpectedAddress, fee path with finalReceiver = router, minAmountOut binding) is untested for this executor: no router-level orchestrated swap, no sequential hop, no split, no fees-active path. The on-chain flow is coherent on inspection (Dispatcher.sol:119-193, TychoRouter.sol:1046-1047), but none of it is verified. Suggest adding a router-level test (swap through TychoRouter -> Dispatcher -> executor with a stale venue) before merge.

crates/tycho-execution/model/src/model/executors.rs
```diff
@@ -664,6 +682,30 @@ impl Executor {
...
+                // The PropAMMRouter's own ETH_SENTINEL path is not used.
+                if token_in == Address::NativeETH || token_out == Address::NativeETH {
+                    return Err(Error::revert(
+                        "PropAMMFallbackExecutor: native token is not supported",
+                    ));
+                }
+                // Unlike the sender-controlled counterparties above, the PropAMMRouter is
+                // governed and whitelist-gated: the swap data cannot make it call arbitrary code.
+                // It has no effect on router state beyond consuming the approval, so this models
+                // the router's accounting around the call, not the venue or the retry.
```
tvinagre: The adversarial model blesses the entire `propammfallback` family with `Ok(())` and never reads the venue — only tokenIn/tokenOut slices are requested. It does not simulate the router's `transferFrom`, the arrival of tokenOut, or the `UnknownVenue` revert. Consequences: (a) routes through non-whitelisted venues (which revert on-chain) are accepted as safe; (b) any real discrepancy in the router interaction is invisible to the fuzzer; (c) a zero-output leg is only 'rescued' via `_maybe_add_client_contribution`, and `is_suspicious` only checks ETH/WETH — a zero-output WETH leg with the router balance going negative and the client vault debited is not flagged (-1 < -1 is false). Note the contrast with the direct PropAMM path above, which returns `Error::Ignore` unless the pAMM is sender-controlled; the fallback path drops that guard. The model crate has no unit tests, and none of the new model code is tested. At minimum, a test asserting the family is only accepted for whitelisted venues (or a documented model limitation) would be valuable.

crates/tycho-execution/src/encoding/evm/gas_estimator.rs
```diff
@@ -195,10 +213,16 @@ fn estimate_transfer_overhead(
...
 
-    if PROTOCOLS_NEEDING_APPROVAL.contains(&protocol_system) {
+    if needs_approval(protocol_system) {
         overhead += BigUint::from(TOKEN_APPROVAL_GAS);
     }
 
+    // The venue swap gas from `get_amount_out` prices a direct call, not one wrapped by the
+    // PropAMMRouter.
+    if protocol_system.starts_with(PROPAMM_FALLBACK_PREFIX) {
+        overhead += BigUint::from(PROPAMM_FALLBACK_OVERHEAD_GAS);
```
tvinagre, line 236: Possible double-count of the router's `transferFrom` gas. `estimate_transfer_overhead` adds a full input transfer (`transfer_token_gas`, ~60k) for the family AND `PROPAMM_FALLBACK_OVERHEAD_GAS` (70k), which is described as measured on mainnet forks (+48,972 FermiSwap / +69,952 Kipseli). If that measured delta is router-wrapped-minus-direct-venue, it already includes the router's `transferFrom` (which the direct call lacks, since it is push-payment), so the separate input-transfer line double-counts. The unit test asserts the combined 355,000 purely from the comment's breakdown, with no independent measurement. If intentional, the comment should state that the fork delta excludes `transferFrom`; otherwise validate against a fork-measured full router call.

crates/tycho-execution/contracts/test/Constants.sol
```diff
@@ -211,6 +211,14 @@ contract Constants is
     // FermiSwap
     address FERMI_SWAPPER = 0xb1076fE3AB5e28005C7c323Bac5AC06a680d452e;
 
+    // PropAMMRouter (github.com/lambdaclass/propamm-router-contracts) and its Uniswap V3 fallback
+    address PROPAMM_ROUTER = 0x4DdF368080CD7946db5b459aD591c350158175e1;
+    address PROPAMM_FALLBACK_SWAP_ROUTER =
```
tvinagre, line 216: Dead test constants: `PROPAMM_FALLBACK_SWAP_ROUTER` (0x68b3...) and `KIPSELI_PROPAMM_VENUE` (0x71e7...) are declared but never referenced anywhere in the contracts tree. The Kipseli venue — which produced the higher measured gas overhead — is never used in any test; only FermiSwap is exercised. Either wire them into a test (recommended for the venue that drives the overhead constant) or remove them.

crates/tycho-execution/model/src/model/executors.rs
```diff
@@ -664,6 +682,30 @@ impl Executor {
...
+                    Address::POSSIBLY_ERC20_AND_NATIVE,
+                )?;
+                let token_out = params.request(
+                    ParamKey::ProtocolData { swap_index, start: 40, end: 60 },
+                    Address::POSSIBLY_ERC20_AND_NATIVE,
+                )?;
+                // The executor passes token addresses through unchanged and declares
+                // ProtocolWillDebit, so a native leg would have the router approve the ETH marker.
+                // The PropAMMRouter's own ETH_SENTINEL path is not used.
+                if token_in == Address::NativeETH || token_out == Address::NativeETH {
```
tvinagre, line 714: The native-token rejection branch here is reachable and untested: the requests use `Address::POSSIBLY_ERC20_AND_NATIVE`, so NativeETH can legitimately appear and the branch fires (safe direction — `Error::revert` rejects the route). Beyond that, the on-chain executor never reads the router's venue whitelist and neither does the model; a whitelist regression on the router would silently change route outcomes with no model signal. Worth a test for the native branch and, ideally, a way for the model to reflect whitelist state.

crates/tycho-execution/contracts/test/protocols/PropAMMFallback.t.sol
tamaralipows: Added `PropAMMFallbackRouterTest` in 0a89739a7: the same stale venue at the same fork block, run through TychoRouter -> Dispatcher -> executor.

- `testSingleSwap` — 1 WETH in, 1872190012 USDC to ALICE, router left holding nothing and no allowance to the PropAMMRouter.
- `testSingleSwapWithRouterFee` — 1% router fee, output lands at the router, fee credited to the receiver's vault.
- `testSequentialSwap` — the retry pays the next hop's pool (`fundsExpectedAddress`), then USDC -> DAI on Uniswap V2.
- `testSplitSwap` — two legs on the same executor, so the second approval survives the first leg's `_revokeUnconsumedApproval`.
- `testSingleSwapMinAmountOutBinds` — `minAmountOut` above the retry price reverts `TychoRouter__NegativeSlippage`.

17 tests pass in the file.

crates/tycho-execution/model/src/model/executors.rs
tamaralipows: Fixed in ffb673983. The model now reads the venue and no longer blesses the family:

```
let venue = params.request(ProtocolData { 0..20 }, Address::VARIANTS)?;
if venue.is_sender_controlled() {
    return Err(Error::revert("PropAMMRouter: UnknownVenue"));
}
...
state.erc20_safe_transfer_from(token_in, PROPAMM_ROUTER, Address::Router, PROPAMM_ROUTER, amount)?;
```

A venue the sender controls is never on a governance whitelist, so the route reverts — the model now says the swap data cannot reach code the sender wrote. Verified on-chain by `testUnknownVenueReverts`.

The router also consumes the approval it was given, so the pull and the leftover-approval revoke are exercised instead of skipped.

Unit tests added (the crate had none): `sender_controlled_venue_reverts`, `native_token_reverts`, `whitelisted_venue_consumes_the_approval`, `transfer_data_approves_the_propamm_router`.

Full run: 360,320 successes on `PropAMMFallback`, 489,600 `UnknownVenue` reverts, 326,400 native reverts, 0 suspicious, 0 warnings.

What the venue or the retry pays back is still not modeled — the model cannot make a counterparty the sender does not control pay out, so every leg here measures zero output, same as the other non-sender-controlled protocols. Stated in the crate's Limitations section.

crates/tycho-execution/src/encoding/evm/gas_estimator.rs
tamaralipows: You are right, it was a double count. Fixed in 5fc9d4cc7.

The measured delta is dominated by an ERC20 transfer — 48,972 is about the cost of one cold `transferFrom` on its own — so the pull is inside it. The leg is `ProtocolWillDebit`, so the Dispatcher performs no transfer of its own; it only approves. The separate input-transfer line was charging the pull a second time, and the family is now excluded from it. A single fallback leg estimates 295,000 instead of 355,000.

Measured the retry path in this repo to bound the other direction (`testFallbackPathGas`, `testDirectUniswapV3Gas`, block 25682938): 249,895 gas for the whole router call against 106,784 for the same Uniswap V3 swap called directly. So a leg that actually falls back costs about 180k more than the estimate. The estimate prices the venue path on purpose, since that is what a fresh quote takes; the constant's doc comment now says all of this.

crates/tycho-execution/contracts/test/Constants.sol
tamaralipows: Both wired into tests in 0a89739a7 rather than removed:

- `PROPAMM_FALLBACK_SWAP_ROUTER` — `testFallbackFillsOnUniswapV3` asserts `executedVenue` equals it, which is what proves the retry fired rather than the venue; `testDirectUniswapV3Gas` calls it directly for the gas baseline.
- `KIPSELI_PROPAMM_VENUE` — `testKipseliVenueFallsBack`. It is whitelisted and behaves the same as FermiSwap at this block.

crates/tycho-execution/model/src/model/executors.rs
tamaralipows: Both covered in ffb673983.

`native_token_reverts` exercises the branch for `NativeETH` as `token_in` and as `token_out`. The full model run hits it 326,400 times.

On the whitelist: the model now reverts `UnknownVenue` for any venue the sender controls, so a route through a non-whitelisted venue is no longer accepted (details in the other thread). The model reflects the whitelist's guarantee, not its current membership — a governance change to the on-chain set is invisible to it. `PropAMMFallbackExecutorTest.testUnknownVenueReverts` checks the on-chain enforcement, and `testKipseliVenueFallsBack` checks the two venues we route to are on it.

tvinagre (approved):

PR #1279, opened 2026-08-05
fix(execution): only add mid-route wrap swaps for dangling balances

Troshchk:
[Task](https://propeller-heads.atlassian.net/browse/ENG-6255)

## Problem
`add_native_wrap_swaps` finds ETH/WETH gaps with list adjacency, so in a split solution that contains both a WETH branch and a native-ETH branch it injects a 0%-split wrap swap between two parallel branches, and split validation then rejects the solution with "The 0% split for token ... must be the last swap". 
Not mentioned in the ticket: If the bridged token is the solution output, validation passes, the injected swap re-routes the user's payout through more pools, and the user receives less than the quote.

## Fix

The encoder now adds a mid-route bridge swap only if the balance is dangling: the bridged token has no consumer in the solution and is not the solution output. Bridge swaps at the start and at the end of a solution do not change, and sequential solutions do not change.

## Details

Adjacency means "chained" only in a sequential solution. In a flattened split solution, adjacent swaps can belong to different branches.

  The fix repairs two failure types.

  ### Type 1 — quote lost (validation error)

 A split route has both WETH and native ETH as intermediates. Observed in production (ENG-6255).

  **1a — WETH branch larger → error names ETH (0x0000…)**

  ```
  The solver emits:                    Today:                               After the fix:

  #1 USDT → WETH (0.7)         ★ ETH→WETH (0.0) is injected         ETH has a consumer (#4)
  #2 USDT → ETH  (0.0)           between #2 and #3                  → no injection
  #3 WETH → USDC (0.0)         → "The 0% split for token            → the solution encodes ✓
  #4 ETH  → USDC (0.0)           0x0000… must be the last swap"
  ```

  **1b — ETH branch larger → error names WETH (0xc02aaa…)**

  ```
  The solver emits (valid):    Today:                               After the fix:

  #1 USDT → ETH  (0.7)         ★ WETH→ETH (0.0) is injected         WETH has a consumer (#4)
  #2 USDT → WETH (0.0)           between #2 and #3                  → no injection
  #3 ETH  → USDC (0.0)         → "The 0% split for token            → the solution encodes ✓
  #4 WETH → USDC (0.0)           0xc02aaa… must be the last swap"
  ```

  The position of the fork does not matter. The same fix applies when the fork sits deeper (USDT→DAI→{WETH|ETH}→USDC).

  ### Type 2 — silent misroute (wrong execution, no error)

 The solution output is ETH or WETH, and the counterpart is an intermediate. This type does not appear in error logs. It shows as unexplained slippage.

  **2 -- buy ETH, large branch through WETH**

  ```
  The solver emits (valid):            Today:                            After the fix:

  #1 USDT → WETH (0.6)                 ★ ETH→WETH (0.0) is injected      ETH is the solution output
  #2 USDT → ETH  (0.0)  ← payout         between #2 and #3               → no injection
  #3 WETH → USDC (0.0)                 → validation passes               → the payout stays ETH
  #4 USDC → ETH  (0.0)  ← payout       → the 40% ETH payout is wrapped   → delivers as quoted ✓
                                         and re-traded through #3, #4
                                       → the user receives less than
                                         the quote
  ```

### DIFF
--- a/crates/tycho-execution/src/encoding/evm/tycho_encoders.rs
+++ b/crates/tycho-execution/src/encoding/evm/tycho_encoders.rs
@@ -103,6 +103,9 @@ impl TychoRouterEncoder {
     /// `available_tokens` tracks the tokens that still hold a balance at each point of the
     /// route. A 0%-split swap takes its input token's whole remainder, so it also removes
     /// that token again.
+    ///
+    /// `validate_native_wrap_gaps` has already rejected the solutions where a needed conversion
+    /// amount is ambiguous, so an unservable gap here simply inserts nothing.
     fn add_native_wrap_swaps(&self, solution: &Solution, chain: &Chain) -> Solution {
         let swaps = solution.swaps();
         let mut new_swaps: Vec<Swap> = Vec::with_capacity(swaps.len());
@@ -154,8 +157,8 @@ impl TychoRouterEncoder {
     ///
     /// The wrap swap takes the counterpart's whole remaining balance unless a later swap still
     /// consumes the counterpart, in which case it takes only the share `wrap_share` derives.
-    /// Without a derivable share the counterpart is left alone, because wrapping all of it would
-    /// starve that later swap.
+    /// Without a derivable share nothing is inserted; `validate_native_wrap_gaps` rejects those
+    /// solutions before they reach here.
     fn missing_wrap_swap(
         &self,
         consumed_token: &Bytes,
@@ -192,6 +195,61 @@ impl TychoRouterEncoder {
         Some(wrap_swap.with_split(share))
     }
 
+    /// Raises an `EncodingError` if a swap consumes a native or wrapped token that no swap
+    /// provides and the amount to convert from its counterpart does not follow from the solution.
+    ///
+    /// The amount is ambiguous when the counterpart's balance is the solution's payout, or when a
+    /// later swap also consumes the counterpart and `wrap_share` cannot derive a share. Either way
+    /// the solver has to emit an explicit `native_wrapper` swap carrying the split it wants.
+    ///
+    /// A token that neither the solution nor its counterpart can supply is left to the strategy
+    /// validators, which report it as an unconnected token.
+    fn validate_native_wrap_gaps(&self, solution: &Solution) -> Result<(), EncodingError> {
+        let swaps = solution.swaps();
+        let native = self.chain.native_token().address;
+        let wrapped = self
+            .chain
+            .wrapped_native_token()
+            .address;
+
+        for (i, swap) in swaps.iter().enumerate() {
+            let consumed_token = &swap.token_in().address;
+            let counterpart = if *consumed_token == native {
+                &wrapped
+            } else if *consumed_token == wrapped {
+                &native
+            } else {
+                continue;
+            };
+            if is_supplied(consumed_token, solution) || !is_supplied(counterpart, solution) {
+                continue;
+            }
+            if *counterpart == *solution.token_out() {
+                return Err(EncodingError::InvalidInput(format!(
+                    "A swap consumes {consumed_token} which no swap provides, and its counterpart \
+                     {counterpart} is the solution's output token. Converting that balance would \
+                     spend the payout. Add an explicit native_wrapper swap with the split to \
+                     convert."
+                )));
+            }
+            let remaining_swaps = &swaps[i..];
+            let counterpart_consumed_later = remaining_swaps
+                .iter()
+                .any(|swap| swap.token_in().address == *counterpart);
+            if counterpart_consumed_later &&
+                wrap_share(consumed_token, counterpart, swaps, remaining_swaps).is_none()
+            {
+                return Err(EncodingError::InvalidInput(format!(
+                    "A swap consumes {consumed_token} which no swap provides, and a later swap \
+                     also consumes its counterpart {counterpart}. The amount to convert does not \
+                     follow from the solution. Add an explicit native_wrapper swap with the split \
+                     to convert, or set estimated_amount_in on the swaps that consume both tokens."
+                )));
+            }
+        }
+        Ok(())
+    }
+
     /// Returns the swap that wraps or unwraps `token_in` into `token_out`, or `None`
     /// if the two tokens are not the chain's native/wrapped-native pair.
     fn wrap_swap_between(
@@ -215,6 +273,16 @@ impl TychoRouterEncoder {
             None
         }
     }
+}
+
+/// Returns whether `token` holds a balance at some point of the route, either because the
+/// solution starts from it or because a swap produces it.
+fn is_supplied(token: &Bytes, solution: &Solution) -> bool {
+    token == solution.token_in() ||
+        solution
+            .swaps()
+            .iter()
+            .any(|swap| swap.token_out().address == *token)
 }
 
 /// Returns the fraction of `counterpart`'s balance that must be wrapped to supply
@@ -292,7 +360,11 @@ impl TychoEncoder for TychoRouterEncoder {
     ///   `expectedAmountOut`).
     /// * The token cannot appear more than once in the solution unless it is the first and last
     ///   token (i.e. a true cyclical swap).
+    /// * Where a swap consumes a native or wrapped token that no swap provides, the amount to
+    ///   convert from its counterpart follows from the solution. `validate_native_wrap_gaps` lists
+    ///   what makes that amount ambiguous.
     fn validate_solution(&self, solution: &Solution) -> Result<(), EncodingError> {
+        self.validate_native_wrap_gaps(solution)?;
         if solution.swaps().is_empty() {
             return Err(EncodingError::FatalError("No swaps found in solution".to_string()));
         }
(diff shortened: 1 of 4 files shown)


crates/tycho-execution/src/encoding/evm/tycho_encoders.rs
```diff
@@ -88,6 +88,15 @@ impl TychoRouterEncoder {
     /// original solution contains a swap between a chain's native token and
     /// its wrapped counterpart but doesn't include the corresponding
     /// wrapping or unwrapping swap.
+    ///
+    /// A bridge swap carries a 0% split, meaning "consume the entire remaining balance of
```
tamaralipows: I'd call this wrap swap rather than bridge swap. Bridge makes me think of cross-chain, I was only find with it when it was linked with "wrapping" to say wapping_bridge but even that might be vague tbh

crates/tycho-execution/src/encoding/evm/tycho_encoders.rs
```diff
@@ -107,7 +116,13 @@ impl TychoRouterEncoder {
                 let token_out = &swaps[i].token_out().address;
                 let token_in = &swaps[i + 1].token_in().address;
                 if let Some(s) = self._wrapping_bridge(token_out, token_in, chain) {
-                    new_swaps.push(s);
+                    let bridge_input = &s.token_in().address;
+                    let is_consumed = swaps
+                        .iter()
+                        .any(|swap| &swap.token_in().address == bridge_input);
+                    if !is_consumed && bridge_input != solution.token_out() {
+                        new_swaps.push(s);
```
tamaralipows: I worry this might be too broad. I think we should check if another swap takes the remainder of the token rather than the whole token. A consumer with a non-zero split takes a fraction, and the rest still needs unwrapping (and we should add a test for this too)

tamaralipows (commented):
Thanks - I think the edge case goes slightly deeper than this though. Will take over!

crates/tycho-execution/src/encoding/evm/tycho_encoders.rs
```diff
@@ -88,53 +88,110 @@ impl TychoRouterEncoder {
...
+        let remainder_taken = solution
+            .swaps()
+            .iter()
+            .any(|swap| swap.token_in().address == counterpart && swap.split() == 0.0);
+        if !available_tokens.contains(&counterpart) ||
+            remainder_taken ||
+            counterpart == *solution.token_out()
+        {
+            return None;
+        }
```
louise-poole: I cannot recall - do we support cyclical swaps/routes? I think this check would make it so that we don't. Would like to make sure this is intentional.

crates/tycho-execution/src/encoding/evm/tycho_encoders.rs
```diff
@@ -88,53 +88,110 @@ impl TychoRouterEncoder {
     /// original solution contains a swap between a chain's native token and
     /// its wrapped counterpart but doesn't include the corresponding
     /// wrapping or unwrapping swap.
+    ///
+    /// An inserted wrap swap carries a 0% split, meaning "consume the entire remaining
+    /// balance of the input token". Mid-route, a wrap swap is therefore inserted directly
+    /// before the first swap that consumes a token which no earlier swap produces. This
+    /// position is after every swap that takes a fraction of the wrap swap's input token,
+    /// which the split ordering rules require. `missing_wrap_swap` lists the conditions
```
louise-poole: We assume the swaps are correctly ordered no? Maybe we should make that assumption clear in this doc string to make it easier to see that breaking that assumption might break this logic too.

crates/tycho-execution/src/encoding/evm/tycho_encoders.rs
```diff
@@ -88,53 +88,110 @@ impl TychoRouterEncoder {
...
         // Check if we need to add a wrapping swap at the beginning of the solution
-        if let Some(s) =
-            self._wrapping_bridge(solution.token_in(), &swaps[0].token_in().address, chain)
+        if let Some(wrap_swap) =
+            self.wrap_swap_between(solution.token_in(), &swaps[0].token_in().address, chain)
         {
-            new_swaps.push(s);
+            available_tokens.insert(wrap_swap.token_out().address.clone());
+            new_swaps.push(wrap_swap);
         }
```
louise-poole: What about this case:
  token_in = ETH, token_out = USDC
  #1 WETH → USDC (60%)
  #2 ETH  → USDC (rem)

If I understand correctly, the logic here will wrap all the solution input token to WETH for the first path and leave nothing for the ETH needed for the second path?

crates/tycho-execution/src/encoding/evm/tycho_encoders.rs
```diff
@@ -550,6 +608,160 @@ mod tests {
...
+                },
+                default_token(token_in.clone()),
+                default_token(token_out.clone()),
+                BigUint::ZERO,
+            );
+            if split > 0.0 {
+                swap.with_split(split)
+            } else {
+                swap
+            }
```
louise-poole: I mean Swap::new defaults the split to 0.0 anyway, so setting it again to 0.0 is okay. This piece of code is the same as:

```suggestion
            swap.with_split(split)
```

crates/tycho-execution/src/encoding/evm/tycho_encoders.rs
```diff
@@ -88,53 +88,110 @@ impl TychoRouterEncoder {
...
+    /// An inserted wrap swap carries a 0% split, meaning "consume the entire remaining
+    /// balance of the input token". Mid-route, a wrap swap is therefore inserted directly
+    /// before the first swap that consumes a token which no earlier swap produces. This
+    /// position is after every swap that takes a fraction of the wrap swap's input token,
+    /// which the split ordering rules require. `missing_wrap_swap` lists the conditions
+    /// for an insertion.
     fn add_native_wrap_swaps(&self, solution: &Solution, chain: &Chain) -> Solution {
         let swaps = solution.swaps();
         let mut new_swaps: Vec<Swap> = Vec::with_capacity(swaps.len());
+        let mut available_tokens = HashSet::from([solution.token_in().clone()]);
```
louise-poole, line 112: I like this `available_tokens` approach 👌

louise-poole (commented):
Some quick thoughts!

crates/tycho-execution/src/encoding/evm/tycho_encoders.rs
tamaralipows: Confirmed — and worse than an unwrapped branch: it encoded fine as a `sequentialSwap` and silently misrouted. The whole ETH input got wrapped, so the ETH branch had nothing left.

This works now. The inserted wrap carries a split, so each branch gets its share:

```
#0 ETH  -> WETH (60%)  <- inserted
#1 WETH -> USDC (rem)
#2 ETH  -> USDC (rem)  <- keeps its 40%
```

The 60% comes from the amounts the solver already puts on each swap (`estimated_amount_in`). Wrapping is 1:1, so the ratio carries straight over.

We only do this when the answer is unambiguous. If several swaps each take a percentage of the ETH, adding a wrap in front of them changes the pot those percentages are measured against — and nothing in the solution tells us whether the solver already allowed for that. Rather than guess, validation rejects it up front and names the two tokens, instead of the old unconnected-token error. A solver that wants that shape can add the wrap swap itself.

Not an issue for fynd either way — it can't produce this shape.

crates/tycho-execution/src/encoding/evm/tycho_encoders.rs
tamaralipows: Cyclical routes still work. The only thing blocked is this case — a cycle with a branch that needs the leftover WETH unwrapped:

```
        ┌──[50%]── USDC ──────┐
WETH ───┤                     ├── WETH
        └─ (ETH) ── DAI ──────┘
```

The leftover WETH is either the user's payout or feedstock for the ETH branch, and the solution doesn't say which. Unwrapping the payout would deliver less than quoted, so it now fails naming the two tokens instead of the old `"Single swap must have 0% split for token 0xc02aaa..."`.

Covered by `test_validate_fails_wrap_would_spend_payout`, plus `test_add_native_wrap_swaps_skips_cyclical_output_token` for the no-insertion side. A solver that wants this route can add the unwrap swap itself.

crates/tycho-execution/src/encoding/evm/tycho_encoders.rs
tamaralipows: Added:

> This assumes the swaps are already in a valid execution order: a swap comes after the swaps that produce its input token, and for each token the 0%-split swap that takes the remainder comes last.

crates/tycho-execution/src/encoding/evm/tycho_encoders.rs
tamaralipows: Done, taken as suggested.

tamaralipows:
Decided to remove support for encoding-side wrap insertion.

Red flag: I tried to fix those edge cases and it was just getting more and more complex.

The encoding-side solution came from a time before we were injecting wraps/unwraps into the TychoStream. We are now auto-injecting them, there’s no reason to pass a solution without a valid weth/eth hop into our encoding and expect us to auto-insert it.

PR #1281, opened 2026-08-05
refactor(rfq)!: emit ready states from clients, drop stream decode round-trip

die-herdplatte:
Stacked on #1280.

RFQ clients serialized live price data into Tycho wire-format snapshots which `RFQStreamBuilder` immediately decoded back in the same process. The decode step reconstructed each state's client from environment variables, silently dropping all builder configuration (credentials, Bebop origin identification, quote timeouts). Both endpoints of the encode→decode round-trip live in one process; nothing crosses a wire.

## Changes

- `RFQClient::stream` yields ready-to-simulate `Update`s. Each client constructs its protocol states directly and embeds a clone of itself, so binding quotes carry the full builder configuration.
- Client builders take the traded tokens as `HashMap<Bytes, Token>` (with metadata) instead of `HashSet<Bytes>`.
- `RFQStreamBuilder` merges and forwards client streams; `add_client` loses the decoder type parameter and `set_tokens` is gone.
- Deletes the four per-protocol snapshot decoders, the bids/asks attribute JSON round-trip, the decode-time credential reads and the RFQ `TimestampHeader`. Net −1,378 lines.
- Removes the `BEBOP_ORIGIN_*` environment variables introduced in #1280: with the decoder gone, origins set on the client builder reach the quoting states directly.
- The integration test sets Bebop origin identification (simulated user EOA, target router) on the client builder instead of the environment.

Breaking for `RFQClient` implementors and `RFQStreamBuilder`/client-builder consumers; the emitted `Update` type is unchanged for downstream consumers.

Verified against the live API with a router-mode (origin-enforcing) Bebop key and no origin environment variables: zero 101 errors; quotes fetched, validated, encoded and executed (eth_call) through the full TychoRouter pipeline.


die-herdplatte:
Superseded by #1440, which carries this change and the rest of the redesign it grew into: the decode round-trip is gone here too, and on top of it the merge layer is deleted in favour of a watch channel per provider behind a generic `SnapshotFeed`, with the five per-protocol loops collapsed into a shared `book/` module.

PR #1287, opened 2026-08-06
fix(simulation): stop CLMM swaps panicking on an empty tick list

pistomat:
## Problem

`TickList`'s bounds helpers index `ticks[0]` and `ticks[len - 1]` directly, so a pool holding no
ticks panicked with an index-out-of-bounds instead of reporting an error. `uniswap_v3`,
`uniswap_v4`, `velodrome_slipstreams` and `aerodrome_slipstreams` all reach those helpers after
guarding only `liquidity == 0`, and liquidity is updated independently of the ticks: a delta drops
a tick entry once its net liquidity reaches zero without touching the pool's liquidity attribute,
and a snapshot whose ticks all carry zero net liquidity decoded to an empty list because emptiness
was checked before those ticks were filtered out. The panic was therefore already reachable in
production.

## Changes

- `next_initialized_tick` and `next_initialized_tick_within_one_word` reject an empty list with a
  new `TickListErrorKind::Empty`, which every fork maps to `RecoverableError("No liquidity")` —
  the same answer they already give for zero liquidity, and the one `ramses_v3` already returns
  from its `has_initialized_ticks` guards.
- The match arms are now exhaustive over `TickListErrorKind`: adding a variant becomes a compile
  error instead of being silently swallowed as `FatalError("Unknown error")`.
- `set_tick_liquidity` skips inserting a zero-net tick for an absent index, mirroring its removal
  branch, so deltas can no longer re-create the inconsistent shape the guards protect against.
- `query_pool_swap`'s zero-liquidity guard on `uniswap_v3` and `uniswap_v4` returns
  `RecoverableError` instead of `FatalError`: an uninitialised pool stays in the state map so
  later deltas can populate it, instead of being blacklisted permanently by consumers.
- Decoders keep an empty tick list rather than rejecting the snapshot, matching `ramses_v3`. An
  empty pool is unquotable until its ticks arrive; a lifecycle test pins that such a pool quotes
  again once a delta supplies ticks. A snapshot carrying nonzero liquidity but no initialized
  ticks — impossible on-chain — is logged at decode as an upstream data-quality signal.
- Tick parse failures surface as their own `ValueError` instead of being reported as a missing
  `tick_liquidities` attribute. For `uniswap_v4` hook pools a malformed tick attribute previously
  decoded to an empty list silently and now rejects the snapshot, pinned by a hooked-pool decoder
  test.
- The `velodrome_slipstreams` decoder had no tests; it gains a module covering its required
  attributes alongside the empty-snapshot assertion.

## Testing

Rebased onto 0.349.0 main. 150 targeted tests across the five protocols plus `tick_list`, clippy
(`-D warnings`) and fmt green; the 5 uniswap_v4 hook tests that need a node RPC were not run
locally and are untouched by this change. A 5-agent local review produced 14 findings, all
addressed; the new tests were verified to fail against the unfixed code.

### DIFF
--- a/crates/tycho-simulation/src/evm/protocol/aerodrome_slipstreams/decoder.rs
+++ b/crates/tycho-simulation/src/evm/protocol/aerodrome_slipstreams/decoder.rs
@@ -165,15 +165,21 @@ impl TryFromWithBlock<ComponentWithState, BlockHeader> for AerodromeSlipstreamsS
             })
             .collect();
 
-        let mut ticks = match ticks {
-            Ok(ticks) if !ticks.is_empty() => ticks
-                .into_iter()
-                .filter(|t| t.net_liquidity != 0)
-                .collect::<Vec<_>>(),
-            _ => return Err(InvalidSnapshotError::MissingAttribute("tick_liquidities".to_string())),
-        };
+        // An empty tick list decodes to a state that quotes no liquidity; rejecting it would
+        // leave no state for later deltas to repopulate.
+        let mut ticks: Vec<_> = ticks?
+            .into_iter()
+            .filter(|tick| tick.net_liquidity != 0)
+            .collect();
 
         ticks.sort_by_key(|tick| tick.index);
+
+        if liquidity != 0 && ticks.is_empty() {
+            tracing::warn!(
+                pool_id = %snapshot.component.id,
+                "snapshot carries nonzero liquidity but no initialized ticks"
+            );
+        }
 
         let observations: Vec<Observation> = snapshot
             .state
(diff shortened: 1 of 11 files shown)


pistomat:
Force-pushed to reshape the history for review: the single commit is now three, following the mechanism —

1. `fix(simulation): error on an empty tick list instead of panicking` — the `TickList` root fix
2. `fix(simulation): treat an empty tick list as no liquidity in CLMM states` — the five fork states consuming the new error
3. `fix(simulation): keep empty tick lists when decoding CLMM snapshots` — the four decoders that stop producing the inconsistent state

Also rebased onto current main (was 3 behind). Payload unchanged — verified via `git diff --quiet` against the pre-rework tip (`ee9b7286f`); 149 offline targeted tests green per commit range, fmt checked on every commit.

pistomat:
Force-pushed again: comments trimmed to the file's own style. The rationale that was inline now lives in the commit messages, so the diff carries 2 comment lines per 100 added (repo guideline is 3-6). No code changes — same three commits, same payload.

pistomat:
Pushed a fourth commit after an independent review round (two reviewers, one of them on the whole diff):

`fix(simulation): report no price for an uninitialized CLMM pool`

Keeping empty tick lists makes a created-but-never-initialized pool the ordinary decoded shape, and that pool carries a zero sqrt price. `sqrt_price_q96_to_f64` had no lower guard, so `spot_price` answered 0 in one token direction and infinity in the other while every amount-based query correctly reported no liquidity — a fabricated number travelling further than a missing measurement would. The guard sits in the conversion, which covers all five forks; uniswap_v4 asks its hook first and only falls back to it, so hook-priced pools are unaffected.

Two tests pin it, both verified to fail without the guard. 211 spot-price and decoder tests across the crate pass, clippy `-D warnings` and fmt clean.

One behaviour change worth noting explicitly for reviewers, which the commit messages describe only in terms of swaps: `get_limits` on an empty tick list changes from panicking to returning zeros.

PR #1289, opened 2026-08-06
fix(tycho-client): partial-aware BlockHistory linkage for flashblock reverts

brunoguerios:
Flashblock partials carry ephemeral hashes. Only the last partial of a block carries the sealed
block hash (`sf.substreams.rpc.v2.BlockScopedData`). `BlockHistory` linked blocks by hash only.
When history retained a mid-block partial, three things broke: the connected-chain build in
`BlockHistory::new` could not stitch across it, so a reinit after an `Advanced` partial rooted the
new history at that partial and dropped all ancestry; `determine_block_position` could not find a
revert's fork point; and the revert drain in `push` emptied the deque looking for it. The error
propagates out of `BlockSynchronizer::handle_next_message`, which tears down the feed for every
extractor at once. That is the Base outage, about 20 times a day.

A partial at height H is now the same canonical block as any header at height H. Hash matching
still wins wherever a hash match exists: the stitch searches all candidates at a height for a hash
match before it accepts a partial, and the drain takes the height fallback only when the fork point
is not findable by hash. Full blocks still need a hash match, so a genuine gap or a detached
restart still roots the history at the advanced block.

Supersedes f1d7cee3d, which preserved the retained history across reinit but left it unconnectable.
Its regression test used a full block, so the partial path stayed untested.

Not verified against prod traffic. Six tests cover it, including two that replay the logged prod
sequence end to end: partial advanced block, reinit, then a shallow revert.

Base is the only chain running `--partial-blocks` today. Any chain that enables flashblocks
inherits the bug, so this needs a release before the next one does.

### DIFF
--- a/crates/tycho-client/src/feed/block_history.rs
+++ b/crates/tycho-client/src/feed/block_history.rs
@@ -257,6 +257,11 @@ impl BlockHistory {
                     // if this is not a revert it means this block is delayed.
                     BlockPosition::Delayed
                 }
+            } else if block.revert && self.partial_at_height(block.number) {
+                // A revert to a block retained only as a mid-block partial: the revert carries
+                // the sealed hash, which an ephemeral partial hash can never match. Same height
+                // means same canonical block, so this is the expected forward update.
+                BlockPosition::NextExpected
             } else if block.is_partial() && !block.revert {
                 // A non-revert partial block at or below the tip whose hash is not in history is a
                 // superseded/delayed partial. Partials share a block number but carry ephemeral
@@ -291,6 +296,12 @@ impl BlockHistory {
         self.history
             .iter()
             .any(|b| &b.hash == h)
+    }
+
+    fn partial_at_height(&self, number: u64) -> bool {
+        self.history
+            .iter()
+            .any(|b| b.number == number && b.is_partial())
     }
 
     pub fn latest(&self) -> Option<&BlockHeader> {
@@ -709,6 +720,31 @@ mod test {
                 .determine_block_position(&incoming)
                 .unwrap(),
             expected
+        );
+    }
+
+    #[test]
+    fn test_revert_to_block_retained_as_partial_is_next_expected() {
+        // History retains block 9 only as a mid-block partial (ephemeral hash). A revert to
+        // sealed 9 cannot be found by hash but is the same canonical block by height.
+        let mut blocks = generate_blocks(2, 7, None); // full blocks 7, 8
+        blocks.push(partial_block(9, 2, int_hash(8)));
+        blocks.push(partial_block(10, 0, int_hash(9)));
+        let history = BlockHistory::new(blocks, 15).unwrap();
+
+        let revert = BlockHeader {
+            number: 9,
+            hash: int_hash(9),
+            parent_hash: int_hash(8),
+            revert: true,
+            ..Default::default()
+        };
+
+        assert_eq!(
+            history
+                .determine_block_position(&revert)
+                .expect("must classify, not error"),
+            BlockPosition::NextExpected
         );
     }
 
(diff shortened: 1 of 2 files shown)


crates/tycho-client/src/feed/block_history.rs
```diff
@@ -108,14 +133,27 @@ impl BlockHistory {
...
                             .history
                             .back()
                             .ok_or(BlockHistoryError::RevertPositionNotFound)?;
 
-                        if head.hash == block.parent_hash {
+                        if head.hash == block.parent_hash ||
+                            (!hash_findable &&
+                                head.is_partial() &&
+                                head.number + 1 == block.number)
+                        {
```
louise-poole: When we fall back to the height check instead of a hash-verified check we should probably log something to make this easier to trace/debug later if it causes issues. Can you add a debug log for this case?

louise-poole (approved):
Looks good! Have you run this locally yet? If you add debug logs and then run this locally for a while and see those logs without it crashing I think we can proceed with merging.

crates/tycho-client/src/feed/block_history.rs
brunoguerios: Added debug logs at all three points where linkage falls back from a hash-verified check to a height check:

- `HistoryStitchHeightFallback` — the reinit stitch in `BlockHistory::new` picks a partial by height because no candidate at that height hash-matches.
- `RevertForkPointHeightFallback` — the drain loop stops at a partial head by height (the line you commented on).
- `RevertClassifiedByPartialHeight` — `determine_block_position` classifies a revert against a block retained only as a mid-block partial.

I left the post-drain `connects` check unlogged, since it mirrors the drain loop and would only double-log.

**Local run:** 61 minutes against `tycho-base-beta` with `--partial-blocks`, uniswap_v3 + aerodrome_v1. 1,756 blocks, 202 reverts, 17,310 partial updates. No crash, no `RevertPositionNotFound`, no `DetachedBlock`, no panic.

The fallback did fire, and on the exact path this PR fixes. The endpoint dropped the websocket at 19:47 (`ResetWithoutClosingHandshake`, then 502s for ~90s). On catch-up the advanced blocks triggered 6 reinits, and the stitch fell back to height 10 times:

```
19:50:38 DEBUG HistoryStitchHeightFallback number=49801039 hash=Bytes(0x6d64903dd45232ea..) expected_parent=Bytes(0xba784271..)
19:50:38 DEBUG InitBlockHistory tip=Block #49801040 [0x7966f251..] (partial 11)
19:50:38 DEBUG HistoryStitchHeightFallback number=49801041 hash=Bytes(0xf99299873ddf7976..) expected_parent=Bytes(0x3c863baa..)
19:50:38 DEBUG HistoryStitchHeightFallback number=49801039 hash=Bytes(0x6d64903dd45232ea..) expected_parent=Bytes(0xba784271..)
19:50:38 DEBUG InitBlockHistory tip=Block #49801042 [0x20943944..] (partial 11)
```

You can read the cause straight from the log: 49801039 is retained under its ephemeral partial hash `0x6d64903d..`, while the child expects parent `0xba784271..`. Before this PR that is where ancestry was dropped. The client stayed up and kept streaming.

The two revert-side fallbacks did not fire in this window — all 202 reverts resolved through the normal hash path, so those logs stay quiet unless the fork point is only retained as a partial. I verified their output separately by running the three regression tests under a subscriber:

```
RevertClassifiedByPartialHeight number=10 hash=Bytes(0x000000000000000a)
RevertForkPointHeightFallback revert_number=10 revert_parent=Bytes(0x..09) fork_number=9 fork_hash=Bytes(0x000000000000000900000002)
```

louise-poole (approved):
Looks good to me! Thank youuuu!

PR #1290, opened 2026-08-06
feat: add HTTP/1 RPC mode to Tycho client

tvinagre:
## Summary
- Add an explicit HTTP/1-only constructor for Tycho RPC clients.
- Propagate the opt-in through `TychoStreamBuilder` and `ProtocolStreamBuilder`.
- Preserve the existing h2c/HTTP/2-prior-knowledge default and unchanged WebSocket handling.

## Why
Fynd configured with `--disable-tls` must use the in-cluster nginx auth proxy on port 80. That proxy accepts HTTP/1.1, while Tycho RPC clients currently send an h2c prior-knowledge preface. Fynd derives RPC and WebSocket URLs from the same host and port, so a separate h2c-only proxy listener would break the ordinary HTTP/1.1 WebSocket upgrade. This opt-in keeps both paths compatible with the existing proxy.

## Validation
- `cargo test -p tycho-client --lib` — 148 passed, 1 ignored
- `cargo test -p tycho-simulation --lib evm::stream::tests` — 3 passed, 1 ignored
- `cargo clippy -p tycho-client -p tycho-simulation --all-targets --all-features -- -D warnings` — passed
- `cargo +nightly fmt --all -- --check` — passed
- Interactive Claude Code review — APPROVE; focused tests repeated 70 times with 0 failures

## Compatibility
`HttpRPCClientOptions` is unchanged. New APIs are additive; the default remains h2c.

### DIFF
--- a/crates/tycho-client/src/rpc.rs
+++ b/crates/tycho-client/src/rpc.rs
@@ -1278,7 +1278,21 @@ pub struct HttpRPCClient {
 }
 
 impl HttpRPCClient {
+    /// Creates a client that uses HTTP/2 prior knowledge.
     pub fn new(base_uri: &str, options: HttpRPCClientOptions) -> Result<Self, RPCError> {
+        Self::new_with_http_version(base_uri, options, false)
+    }
+
+    /// Creates a client that uses HTTP/1 instead of HTTP/2 prior knowledge.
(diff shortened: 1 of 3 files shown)


brunoguerios (approved):
LGTM ✅ 

Ps: approved assuming checks are failing due to the same github outage I saw in another PR

PR #1294, opened 2026-08-07
fix(integration-test): call MAX_BPS on the V3 FeeCalculator

louise-poole:
The FeeCalculator interface declared MAX_FEE_BPS, but the latest router V3 renamed it to MAX_BPS, so the eth_call hit a non-existent selector and reverted, failing start-up.


zach030 (approved):

Troshchk (approved):
Thank you!

PR #1299, opened 2026-08-10
fix(curve): stop toxic rate providers from detecting our quotes

zach030:
## What

Curve reads `stored_rates()` as a funded sender paying a 1 gwei gas price instead of as the zero address paying nothing, so the rate providers behind it cannot tell our read apart from a swap. Every other Curve getter, and every other protocol, keeps the context it has today.

## Why

Rate providers wired to Curve StableSwap-NG pools have been found branching on:

```
isSimulation = (tx.origin == <marker>) || (tx.gasprice == 0)
```

They wrap a real Chainlink price and multiply it by a factor chosen from that branch — inflated for callers that look like an off-chain quote, real for a swap — arming and disarming it from a storage slot. Our getters ran with `caller = Address::ZERO` and no gas price, which matches **both** markers, so the hybrid decoded the inflated rate into its math pool and quoted every trade off it. Routers then picked the pool and the swap under-delivered.

`stored_rates()` is the only Curve read that reaches attacker-supplied code (it forwards to the pool's configured providers), so it is the only call this changes. A scan of the last 183 days of the Ethereum `stableswap-ng` factory found 113 pools whose providers carry this branch, and one of them was routed through in production.

## Changes

- `SimulationParameters` gains `gas_price: Option<u128>`, wired into revm's `TxEnv`. `None` means zero, so behaviour is unchanged for every existing caller; the four other construction sites just spell out `gas_price: None`.
- `curve/vm.rs` gains `EXECUTION_SENDER` (`keccak256("tycho")[12..]`, deliberately an unremarkable address — a vanity one is what a provider would add to its markers), `EXECUTION_GAS_PRICE`, and `init_execution_sender`, which seeds the sender with a balance because revm still charges it `gas_limit * gas_price`.

## Testing

Three layers:

- **Mechanism** — `stored_rates_are_read_as_the_swap_would_see_them` seeds a pool whose runtime bytecode implements the real branch (`(tx.origin == 0 || tx.gasprice == 0) ? 0xDEAD : 0xBEEF`), calls the production `read_stored_rates` through an engine built exactly as the decoder builds it, and asserts the swap branch is taken.
- **Live pool** — `stored_rates_on_a_toxic_pool_match_the_swap` (RPC-gated) reads `0x509c6c89…` at block 25723040, where its providers were armed. Before this change the read returned `[0.9941e30, 0.9984e30]` (the quote answer); it now returns `[0.9999e30, 0.9993e30]`, matching what a swap gets on-chain. It fails loudly if the pool is ever disarmed rather than quietly becoming a tautology.
- **Blast radius** — `other_reads_keep_the_default_context` pins the default context so the change stays confined to the one call.

`cargo test -p tycho-simulation --lib curve::`: 194 passed. clippy (`--all-targets --all-features`) and `cargo +nightly fmt` clean.

## Scope

This defeats the mechanism these providers use today; a determined attacker can move the branch (our sender is in open source, and the pools' own getters could carry it just as easily). It is not a substitute for `curve_filter`, which already excludes every one of these pools (they declare `asset_types = ["0x01", ...]`), or for blocklisting them. Consumers that register `vm:curve` without `Some(curve_filter)` remain the larger exposure.

### DIFF
--- a/crates/tycho-simulation/src/evm/protocol/balancer_v3/vm.rs
+++ b/crates/tycho-simulation/src/evm/protocol/balancer_v3/vm.rs
@@ -758,6 +758,7 @@ fn params(to: AlloyAddress, data: Vec<u8>) -> SimulationParameters {
         gas_limit: None,
         transient_storage: None,
         block_overrides: None,
+        gas_price: None,
     }
 }
 
--- a/crates/tycho-simulation/src/evm/protocol/curve/vm.rs
+++ b/crates/tycho-simulation/src/evm/protocol/curve/vm.rs
@@ -8,10 +8,10 @@ use std::fmt::Debug;
 
 use alloy::{
     core::sol,
-    primitives::{Address as AlloyAddress, U256},
+    primitives::{Address as AlloyAddress, Keccak256, U256},
     sol_types::SolCall,
 };
-use revm::{state::AccountInfo, DatabaseRef};
+use revm::{primitives::KECCAK_EMPTY, state::AccountInfo, DatabaseRef};
 use tycho_common::simulation::errors::SimulationError;
 
 use crate::evm::{
@@ -21,7 +21,7 @@ use crate::evm::{
             adapter::{build_pool, detect_eth_variant, CurveVariant, ProbingResults, RawPoolState},
             math::Pool,
         },
-        vm::utils::get_code_for_contract,
+        vm::{constants::MAX_BALANCE, utils::get_code_for_contract},
     },
     simulation::{SimulationEngine, SimulationParameters},
 };
@@ -308,8 +308,14 @@ where
     <D as DatabaseRef>::Error: Debug,
     <D as EngineDatabaseInterface>::Error: Debug,
 {
+    // Read the providers the way the swap will: they answer a quote and a swap differently, and
+    // both the sender and the gas price have to say "swap" for that to hold.
     let res = engine
-        .simulate(&params(*pool, STORED_RATES_SELECTOR.to_vec()))
+        .simulate(&SimulationParameters {
+            caller: execution_sender(engine, pool)?,
+            gas_price: Some(EXECUTION_GAS_PRICE),
+            ..params(*pool, STORED_RATES_SELECTOR.to_vec())
+        })
         .ok()?;
     let out = res.result.as_ref();
     if out.len() < n_coins * 32 {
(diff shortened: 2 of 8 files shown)


die-herdplatte (commented):
Interesting!
Have you tested this on the 113 pools you identified?

I think this PRs replacement values are fine for now.
If attackers adapt to branch on other environmental values (many values on `BlockEnv` are still defaulted and can't land in a mainnet block) or specifically target Tycho, we may need to do more work (randomize & restrict to plausible values).

zach030:
Ran it across all 113, two ways.

**1. Rate level — `eth_call` probe, all 113 pools @ block 25821663.** Each pool's `stored_rates()` called twice, differing only in the two fields this PR changes: quote context (zero sender, zero gas price) vs execution context (this PR's sender and 1 gwei, with a balance override standing in for the funding `execution_sender` does).

| | pools |
|---|---|
| armed (quote ≠ execution) | **103** |
| disarmed (identical) | 10 |
| errored | 0 |

Divergence per pool, largest coin delta: 57 above 0.5%, 40 between 0.2–0.5%, 6 between 0.05–0.2%. Direction goes both ways (47 quote high, 56 quote low), which fits the rate being applied per coin rather than per side — either sign skews the quote depending on the swap direction. The 10 that matched are dormant, not fixed: same providers, same branch, just not armed at that block.

**2. Quote level — through `tycho-simulation`.** Rate divergence isn't the number that matters, so I also built the real `Pool` both ways (production `read_stored_rates` vs the pre-fix read, same balances/amp/fee otherwise) and quoted `get_amount_out` on each:

| pool | liquidity | quote gap |
|---|---|---|
| `0xB5555Ab6…` | ~$8.4k | **+49.23 bps** |
| `0x706dDBd9…` | ~$36k | **+49.22 bps** |
| `0x509C6c89…` | ~$46k | **−48.80 bps** |

So roughly ±0.5% on the quote, and near-identical across the three — same template, batch-deployed.

Only three because the other 108 currently sit at `balances = [0, 0]` and cannot be quoted at all; that is a property of the pools right now, not of the fix. They are drained shells whose providers are still armed at the rate level, so they price again the moment someone refunds them.

**On randomisation** — agreed, and done in the latest push: the sender is now derived per pool and per block as `keccak256(pool ++ block_hash)[12..]` instead of a constant. A fixed address sits in the open and costs a provider nothing to add to its markers; this one can't be enumerated ahead of the block.

I left the rest of `BlockEnv` alone for now (`basefee = 0`, `coinbase = 0x0`, `gas_limit = u64::MAX`, `prevrandao = 0x0` are all impossible on mainnet and just as usable as tells). Worth noting the asymmetry there: values that genuinely occur on mainnet are safe for us to use precisely because branching on them would misfire on real swaps too, so the fix is plausibility and internal consistency rather than randomness — e.g. `gasprice ≥ basefee`, a real builder coinbase, `prevrandao` from the block hash we already carry. The clean version is carrying the real header fields through the stream instead of synthesising them, which is a bigger change; happy to open a follow-up.

This still doesn't replace `curve_filter`, which already excludes all 113 (they declare `asset_types = ["0x01", …]`). The bigger exposure is consumers registering `vm:curve` without it.

PR #1300, opened 2026-08-10
feat(sushiswap-v2): add Base and Arbitrum manifests

zach030:
Adds two Substreams manifests to the `ethereum-uniswap-v2` package for SushiSwap V2 on Base and Arbitrum. No Rust changes — both reuse `ethereum_uniswap_v2.wasm`, since the map modules take the factory address and protocol type name as params and SushiSwap V2 pairs on both chains are standard 0.3% UniswapV2 pairs (verified on-chain: the pairs have no `swapFee()`, matching the hardcoded 30 bps `fee` attribute).

| Manifest | Factory | initialBlock |
|---|---|---|
| `base-sushiswap-v2.yaml` | `0x71524B4f93c58fcbF659783284E38825f0622859` | 2631214 |
| `arbitrum-sushiswap-v2.yaml` | `0xc35DADB65012eC5796536bD9864eD8773aBc74C4` | 70 |

Both use `protocol_type_name=sushiswap_v2_pool`, matching `ethereum-sushiswap-v2.yaml`.

`initialBlock` is the factory deployment block on each chain. On Arbitrum that is block 70 (2021-05-30, pre-Nitro); the first `PairCreated` is at 228460 and over a thousand pairs — including WETH/USDC.e — were created before the Nitro migration, so the start block cannot be moved up to the fork point. Backfill from block 70 requires the Substreams provider to serve Arbitrum One from genesis, including logs for classic-era blocks.

Package version bumped to 0.3.3 with a changelog entry, per `protocols/substreams/CLAUDE.md`.

Publishing is manual — `release.sh` only auto-selects `ethereum*` manifests in this package:

```
cd protocols/substreams
./publish.sh ethereum-uniswap-v2 base-sushiswap-v2
./publish.sh ethereum-uniswap-v2 arbitrum-sushiswap-v2
```

Not included, needed before these run in production: `sushiswap_v2` entries under `base` and `arbitrum` in `crates/tycho-execution/config/executor_addresses.json` (same executor address as `uniswap_v2` on each chain), and the extractor config for both chains.

## Testing

- `cargo +nightly fmt`, `cargo +nightly clippy -- -D warnings`, `cargo build --target wasm32-unknown-unknown`, `cargo test` for `ethereum-uniswap-v2` — all pass
- `substreams pack` succeeds for both manifests; `substreams info` confirms params and initial blocks, and the pack warnings match the existing `ethereum-sushiswap-v2.yaml`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1301, opened 2026-08-10
feat(sushiswap-v2): add Base support

zach030:
Adds SushiSwap V2 indexing support on Base, reusing the `ethereum-uniswap-v2` Substreams package unchanged. Scoped to `protocols/` — the consumer-side registration follows in a separate PR once this one merges.

## What

- **Substreams**: `base-sushiswap-v2.yaml` manifest next to the existing per-chain manifests, factory `0x71524B4f93c58fcbF659783284E38825f0622859`, `initialBlock` 2631214 (the factory's creation block). Package version bumped to 0.3.3 with a changelog entry.
- **Testing harness**: `base-sushiswap-v2` → `ethereum-uniswap-v2` clone mapping, plus `integration_test_base_sushiswap_v2.tycho.yaml`.

No Rust changes in the Substreams module. The map modules take the factory address and protocol type name as params, and SushiSwap V2 pairs on Base are standard 0.3% UniswapV2 pairs — verified on-chain: the pairs expose no `swapFee()`, so the hardcoded 30 bps `fee` static attribute holds.

## Follow-up PR (after this merges)

The consumer-side wiring is held back so this PR stays inside `protocols/`:

- `base.sushiswap_v2` in `crates/tycho-execution/config/executor_addresses.json`, reusing the UniswapV2 executor already deployed on Base (`0x321C00604355DD20A2CB472D44fe8f37d821A82c`) — no new contract.
- `sushiswap_v2` registered on Base in the quickstart and price_printer examples.
- `sushiswap_v2` added to the Base list in `get_default_protocols_for_chain` (`crates/tycho-integration-test/src/stream_processor/protocol_stream_processor.rs`). That binary subscribes to a live Tycho instance, so it can only be enabled once the extractor is actually running in production.

## Arbitrum is not included

Arbitrum was in scope initially and was dropped: Substreams cannot serve the blocks the integration needs.

- The StreamingFast `arb-one` endpoint's first streamable block is **22,207,900**, the start of the Nitro era. Arbitrum Classic blocks are not served ([StreamingFast integration announcement](https://www.streamingfast.io/blog/launching-the-new-arbitrum-one-substreams-integration-in-beta)).
- Pinax's `arbone` endpoint does cover the Classic range, but only as base blocks — no receipts, therefore no logs.

The SushiSwap V2 factory on Arbitrum (`0xc35DADB65012eC5796536bD9864eD8773aBc74C4`) was deployed at block 70, on 2021-05-30, and 1,045 pairs were created before the 2022-08-31 Nitro migration at block 22207817. Those pre-Nitro pairs are 3.2% of the pool count but hold **63.5% of the TVL** and, once two post-Nitro wash-volume pools are excluded, **78% of real 24h volume** (measured against GeckoTerminal's top 196 pools by volume and tx count; the TVL share is a floor, since the pools GeckoTerminal does not surface are dormant ones, which skew pre-Nitro).

Since the module derives components from `PairCreated` and reserves from `Sync`, starting at the Nitro boundary would silently omit those pools, and starting earlier is rejected by the endpoint. Revisit if Classic-era blocks become available.

## Publishing

Manual — `release.sh` only auto-selects `ethereum*` manifests in this package:

```
cd protocols/substreams
./publish.sh ethereum-uniswap-v2 base-sushiswap-v2
```

## Testing

- Substreams package checks (what Substreams CI runs for the changed package): `cargo +nightly fmt --check`, `cargo +nightly clippy -- -D warnings`, `cargo build --target wasm32-unknown-unknown`, `cargo test` — all pass.
- `substreams pack base-sushiswap-v2.yaml` succeeds; `substreams info` confirms params, network and initial block.
- Workspace: `cargo +nightly fmt --check` clean, `cargo clippy -p protocol-testing -p tycho-simulation --all-targets --all-features` clean.
- Integration test pool verified on-chain: TYBG/WETH `0xf65bb528ced09008603509c3fda43e1ccfddf935` comes from the Base SushiSwap factory, token order matches the config, it is created and funded in block 8167705, and 6 `Sync` events fall inside the configured range.

Replaces #1300, which GitHub closed when its head branch was renamed.


die-herdplatte (approved):
lgtm!

> Publishing is manual — release.sh only auto-selects ethereum* manifests in this package

I think the workflow supports this though, no need to manually publish.

PR #1302, opened 2026-08-10
feat: serve get_tokens from an in-memory token cache

zizou0x:
## What

Adds an in-memory copy of the token tables to `tycho-storage` and serves `get_tokens` from it instead of SQL. Enabled via `GatewayBuilder::enable_token_cache()` for the `index` and `rpc` commands; the `analyze-tokens` job keeps the SQL path, which also remains as the fallback when the cache is disabled.

Per chain the cache stores all tokens in one `Vec` in `token.id` order, so a token's position acts as a compact integer id. The quality filter is answered from roaring bitmaps over these positions, the last-traded filter from a flat timestamp array, and pagination totals come from bitmap cardinality instead of a `COUNT` query. See the module docs in `token_cache.rs` for the full design.

The cache stays in sync with the database three ways: a paged full load at startup, write-through on token and balance writes in the same process, and a background poll every 60s for token rows with a newer `modified_ts` (this catches quality updates written by the token analysis job from its own pod). The poll re-reads a 10 minute overlap window so writes from transactions that commit late are not skipped. One migration adds an index on `token(modified_ts)` to keep the poll cheap; note it takes a SHARE lock on `token` for the seconds the index build needs.

## Measurements

Benchmarked against dev databases through a port-forward (an `#[ignore]`d test, run with `DATABASE_URL=... BENCH_CHAIN=... cargo test -p tycho-storage --release --lib token_cache_benchmark -- --ignored --nocapture`). All totals and page contents matched the SQL path exactly.

Base dev, 4,978,873 tokens (load 35s, +1.87 GiB RSS):

| query | SQL | cache |
|---|---|---|
| all tokens, page 0 | 1.75 s | 0.26 ms |
| all tokens, deep page | 5.13 s | 0.46 ms |
| min_quality=51, page 0 | 1.86 s | 0.79 ms |
| min_quality=51 + traded 30d, page 0 | 2.50 s | 29 ms |
| 100 addresses | 58 ms | 0.17 ms |
| full sweep, q>=51 + 30d, 20 pages | ~58 s | 0.55 s |

Ethereum dev (605k tokens): load 5s, +234 MiB RSS, queries 0.13-6.1 ms.

## Tests

- Unit tests for the store (quality moves, pagination boundaries, traded-filter edges).
- DB equivalence tests asserting identical rows, order and totals between cache and SQL across the filter/pagination matrix, after an out-of-process quality update picked up by `refresh`, and after write-through inserts/updates/balance writes.

## Notes

- No API or DTO changes; clients are unaffected.
- Memory cost is ~1.9 GiB on Base and ~240 MiB on ethereum, in the indexer pod.
- Earlier attempt: #610.


zizou0x:
Deployed to base dev (`cd-deploy-dev`) and measured against the live endpoint `tycho-base-dev.propellerheads.xyz` through the auth proxy. Startup: token cache loaded 4,978,904 tokens in 29.5s; pod memory 9.8-10.2 GiB of the 14 GiB limit; extractors and the delta refresh run without errors.

`/v1/tokens` latency with uncached request bodies, measured from outside the cluster (includes TLS + proxy + transfer):

| request | before | after |
|---|---|---|
| min_quality=100 + traded 3d, page 0 (fynd shape) | 0.33 s | 0.18 s |
| all tokens, page 0 | 3.1 s | 0.11 s |
| all tokens, page 700 | 8.4 s | 0.07 s |
| min_quality=51 + traded 30d, page 0 | 5.9 s | 0.16 s |
| min_quality=51 + traded 30d, page 5 | 3.5 s | 0.11 s |
| min_quality=51, page 0 | 1.9 s | 0.11 s |

Remaining time is network and serialization; server-side work is now sub-millisecond to ~30 ms. Fynd startup against dev was measured before and after and is unchanged within noise (7-23 s) — its default filter returns a single 8.4k-token page, so its startup is dominated by the snapshot, not tokens.

crates/tycho-storage/src/postgres/token_cache.rs
```diff
@@ -0,0 +1,1557 @@
...
+    /// sync marker only on success, so a failed poll is retried on the next tick.
+    /// Returns the number of rows read in the lookback window, which is an upper
+    /// bound on (not a count of) actual changes.
+    ///
+    /// The query re-reads a window of [`REFRESH_OVERLAP_SECS`] before the sync
+    /// marker. This closes a race: a write from a transaction that was still open
+    /// during the previous poll carries a `modified_ts` from *before* that poll,
+    /// so a strict `> last_sync` filter would skip it forever. Re-reading recent
+    /// history is safe because writing the same token twice is a no-op.
+    pub async fn refresh(&self, conn: &mut AsyncPgConnection) -> Result<usize, StorageError> {
```
zizou0x, line 530: Biggest gap I see: `last_traded` only converges in the process that writes the balances. `refresh` polls `token.modified_ts`, but balance writes never touch the token row, so in a process that doesn't index — the `rpc` command, which `run_rpc` now also enables the cache for — `last_traded` is frozen at the startup snapshot. Tokens that start trading after boot (including freshly discovered ones, which arrive here via refresh as `NEVER_TRADED`) will never pass a "traded within N days" filter until the pod restarts, and the error grows with uptime. Prod currently only runs `index`, where write-through covers this, but anything running `rpc` serves silently shrinking results.

Suggest either a second delta poll over `component_balance_default.valid_from` (the same overlap-window trick would work), or dropping `enable_token_cache()` from `run_rpc` until that exists.

crates/tycho-storage/src/postgres/token_cache.rs
```diff
@@ -0,0 +1,1557 @@
...
+        let last_sync = *self
+            .last_sync
+            .read()
+            .expect("token cache lock poisoned");
+        let since = last_sync - chrono::Duration::seconds(REFRESH_OVERLAP_SECS);
+
+        let rows: Vec<(orm::Token, Address, i64)> = schema::token::table
```
zizou0x, line 538: This poll is unbounded — no LIMIT or paging, unlike the initial load. A bulk quality backfill (or a manual `UPDATE token ...`) touching millions of rows makes a single tick load them all into one `Vec` (GiB-scale transient allocation on Base), and `upsert_tokens` then holds the chain's write lock while applying the whole batch, stalling every `get_tokens` for the duration. Worth paging with a `(modified_ts, id)` cursor like `load_chain` does, applying in chunks so readers can interleave.

crates/tycho-storage/src/postgres/protocol.rs
```diff
@@ -1156,6 +1171,10 @@ impl PostgresGateway {
         .instrument(debug_span!("insert_tokens", count = token_count))
         .await?;
 
+        if let Some(token_cache) = &self.token_cache {
+            token_cache.add_tokens(tokens);
```
zizou0x, line 1175: Write-through fires inside the write executor's transaction (`DBCacheWriteExecutor` runs `execute_write_op` under `build_transaction().repeatable_read()` with retries), i.e. before commit. If the transaction rolls back — serialization failure on the last retry, or an error in a later op of the same batch — the cache keeps the phantom tokens / advanced `last_traded`, and since the store is append-only, refresh never undoes a token that was never committed. Retries that eventually commit the same data make this self-healing in the common case, but a permanently failed batch leaves divergence until restart. Consider buffering cache mutations and applying them after commit, or documenting why the transient divergence is acceptable. (Same applies to the write-throughs in `update_tokens` and `insert_component_balances`.)

crates/tycho-storage/src/postgres/token_cache.rs
```diff
@@ -0,0 +1,1557 @@
...
+            .select((schema::chain::id, schema::chain::name))
+            .load(conn)
+            .await
+            .map_err(PostgresError::from)?;
+
+        let mut chain_ids = HashMap::new();
+        let mut chains = HashMap::new();
+        let mut last_sync = NaiveDateTime::default();
+        for (chain_id, chain_name) in chain_rows {
+            let chain = Chain::from_str(&chain_name).map_err(|_| {
```
zizou0x: Two coupled issues here: (1) the cache loads every chain present in the `chain` table, ignoring `GatewayBuilder::set_chains` — a DB hosting more than the target chain pays the full memory cost for all of them; (2) an unrecognized chain name is a hard error that aborts gateway construction, so a chain row inserted by a newer version sharing the DB would crash-loop older pods at startup. Restricting the load to the builder's configured chains (and skip-with-warn for unknown rows) fixes both.

zizou0x (commented):
Reviewed for big gaps — four inline comments. The headline one is `last_traded` staleness in any process that doesn't write balances itself (the `rpc` command); the others are robustness: the unbounded refresh poll, write-through firing before commit, and the chain-table load scope/failure mode. Design, docs and test coverage otherwise look solid, and the schema.rs regen checks out against existing migrations.

crates/tycho-storage/src/postgres/token_cache.rs
zizou0x: It's a fine limitation, if we run rpc only it means the extractor doesn't run so it shouldn't move. We can leave a note for this but not urgent to address.

crates/tycho-storage/src/postgres/token_cache.rs
zizou0x: Also fine with this limit. We don't update millions of tokens.

crates/tycho-storage/src/postgres/protocol.rs
zizou0x: Discussed offline — deferring the fix, documented the limit in the module docs (8ba90b2fb) instead. Summary of the analysis:

**Today this is benign.** A non-deadlock write failure is fatal: the error propagates from `DBCacheWriteExecutor` through `commit_transaction` to the extractor, the runner exits, and the process dies — the cache is rebuilt from the DB at startup, wiping any divergence. Deadlock retries re-apply identical data through idempotent cache ops, so they converge. Worst case is a few seconds of phantom data on a pod that is already crashing.

**It becomes a real bug if in-process extractor restarts land (#1026).** There, `CachedGateway::new_instance()` shares `state_gateway` and thus the same `Arc<TokenCache>`, the process survives write failures, and the RPC keeps serving. Transient failures still self-heal (redo from cursor re-commits identical finalized data), but a persistent failure leaves phantoms in the cache indefinitely — the refresh poll cannot remove a token that is not in the DB, and a rolled-back update leaves the row's `modified_ts` old so it is never re-read.

**Fix to apply before #1026 (or similar) merges** (~150 lines, no gateway signature changes): stage cache mutations per DB transaction instead of applying them — a `CacheOp` enum + staging buffer + `begin/commit/discard` on `TokenCache`; the three write-through sites push ops (they already compute the needed context: the exists-in-DB filter, the tx→timestamp map); `DBCacheWriteExecutor::write` brackets its transaction with begin/discard per retry attempt and commits the staged ops only on `Ok`. The `DirectGateway` path keeps applying immediately since its calls autocommit.

zizou0x: Update: concluded this is acceptable by design, superseding the plan above. Only finalized block data reaches the write path, so on rollback the cache runs *ahead of* the database while staying faithful to on-chain state — a phantom token is a real on-chain token the DB failed to record, and the DB catches up when the write is retried or the extractor re-processes from its cursor. That holds under in-process restarts (#1026) too, so no post-commit staging is planned. The module doc note was reframed accordingly in 7660233d7.

crates/tycho-storage/src/postgres/protocol.rs
```diff
@@ -1156,6 +1171,10 @@ impl PostgresGateway {
         .instrument(debug_span!("insert_tokens", count = token_count))
         .await?;
 
+        if let Some(token_cache) = &self.token_cache {
```
tvinagre, line 1174: 🔴 **Critical: Write-through fires before DB transaction commits.**

The cache is updated here, inside the DB transaction, before it commits. On rollback (deadlock, constraint violation, connection loss), the cache has phantom tokens that don't exist in the DB — visible to all concurrent readers until the next 60s refresh cycle.

The module docs dismiss this as harmless because "only finalized block data reaches this path," but rollbacks do happen on the extractor hot path. Consider moving this to a post-commit callback.

crates/tycho-indexer/src/main.rs
```diff
@@ -401,6 +401,7 @@ async fn run_rpc(global_args: GlobalArgs, rpc_args: RpcServerArgs) -> Result<(),
 
     let direct_gw = GatewayBuilder::new(&global_args.database_url)
         .set_chains(&[chain])
+        .enable_token_cache()
```
tvinagre, line 404: 🔴 **Critical: `last_traded` goes stale in the `rpc` process.**

The `rpc` command doesn't write component balances, so write-through never fires for `last_traded`. The delta refresh only covers the `token` table, not `component_balance_default`. This means tokens that have their first trade after the rpc pod starts will have `NEVER_TRADED` forever, and the "recently traded" filter silently excludes them.

The gap grows with uptime. Consider not enabling the cache for `rpc` until the refresh also covers balance changes, or adding `last_traded` to the delta refresh query.

crates/tycho-storage/migrations/2026-08-10_token_modified_ts_index/up.sql
```diff
@@ -0,0 +1,2 @@
+-- Supports the token cache delta refresh, which polls for recently modified tokens.
+CREATE INDEX IF NOT EXISTS idx_token_modified_ts ON token (modified_ts);
```
tvinagre, line 2: ⚠️ **Warning: `CREATE INDEX` takes a SHARE lock on the `token` table.**

On Base with 5M tokens, this can block writes for 30-60 seconds during deployment. The extractor's hot path stalls — it cannot insert new tokens while the index builds.

Use `CREATE INDEX CONCURRENTLY` to avoid blocking writes. Note: this cannot run inside a transaction, so the migration needs a non-transactional execution path.

tvinagre (changes requested):
## Deep Code Review — In-Memory Token Cache

**Reviewer:** Thales (via Hermes Agent)
**Verdict: Changes Requested** — 2 critical issues, 3 warnings, 4 design questions.

---

### 🔴 Critical 1: Write-Through Fires Before DB Transaction Commits

**Files:** `protocol.rs` — `insert_tokens` (line ~1174), `update_tokens` (line ~1235), `add_component_balances` (line ~1413)

The write-through hooks are called *inside* the DB transaction, after the SQL `execute()` but before the transaction commits. The module docs acknowledge this:

> *"Known limit: write-through runs while the enclosing DB transaction is still open, so on rollback the cache can run ahead of the database."*

The dismissal ("This is harmless: only finalized block data reaches this path") is **not fully justified**:

1. **Transaction rollback**: If the DB transaction rolls back (deadlock, constraint violation, connection loss), the cache now has tokens/balances that don't exist in the DB. These are visible to all concurrent readers via `query_tokens` until the next refresh cycle (60s) or process restart.

2. **`add_tokens` uses `ON CONFLICT DO NOTHING`** (`overwrite: false`) — on retry, the re-inserted token is silently skipped because it's already in the cache. If quality or tax was updated in the failed write but differs from a subsequent retry, the cache holds stale values from the first attempt.

3. **Phantom reads**: An RPC client querying tokens immediately after a rollback sees phantom tokens — tokens that were "indexed" but never committed to the DB. For Fynd consumers this could mean serving token metadata that has no DB backing.

**Recommendation:** Move the write-through hooks to *after* transaction commit. If the gateway doesn't control commit boundaries (the `conn` is passed in), consider a post-commit callback mechanism, or document this as a known correctness gap with a concrete convergence bound.

---

### 🔴 Critical 2: `last_traded` Staleness in the `rpc` Process

**File:** `main.rs:404`, `token_cache.rs` module docs

The PR enables the cache for both `index` and `rpc` commands. But the delta refresh only covers the `token` table — it does NOT refresh `last_traded` timestamps from `component_balance_default`. The `last_traded` vector converges through write-through alone.

The `rpc` command doesn't write component balances (it's a read-only RPC server). So:

- At startup, `last_traded` is loaded from the DB.
- After startup, **no mechanism updates `last_traded`** in the rpc process.
- Tokens that have their first trade after the rpc process started will have `NEVER_TRADED` forever.
- The "recently traded" filter (`last_traded_ts_threshold`) will **silently exclude newly-traded tokens** from RPC responses, and this gap grows with uptime.

The author documented this (commit 09), but the PR still enables the cache for `rpc` in `main.rs:404`:

```rust
let direct_gw = GatewayBuilder::new(&global_args.database_url)
    .set_chains(&[chain])
    .enable_token_cache()   // <-- enabled for rpc, but last_traded goes stale
    .build_direct_gw()
    .await?;
```

**This is a silent data correctness issue.** An RPC consumer (like Fynd) filtering on `traded_30d` will miss tokens that started trading after the RPC pod launched. If the pod runs for days, the gap becomes significant.

**Recommendation:** Either:
- Don't enable the cache for `rpc` until the refresh also covers balance changes, OR
- Add `last_traded` to the delta refresh query (join `component_balance_default` for the lookback window), OR
- At minimum, add a startup warning log when the cache is enabled in a process that doesn't write balances, and document the expected staleness bound.

---

### ⚠️ Warning 1: Unbounded `refresh` Query Result Set

**File:** `token_cache.rs` — `refresh()`

The delta refresh query has no `LIMIT`:

```rust
let rows: Vec<(orm::Token, Address, i64)> = schema::token::table
    .inner_join(schema::account::table)
    .filter(schema::token::modified_ts.gt(since))
    .filter(schema::account::chain_id.eq_any(chain_db_ids))
    .order(schema::token::id.asc())
    .load(conn).await?;
```

The lookback window is 10 minutes (`REFRESH_OVERLAP_SECS = 600`). If the `analyze-tokens` job does a bulk quality update across millions of tokens (plausible — it re-evaluates quality scores), this query loads **all of them into memory at once**. On Base with 5M tokens, a full quality re-evaluation would:

1. Load 5M `orm::Token` rows + addresses into a `Vec` — hundreds of MB of transient allocation.
2. Hold them while iterating and upserting into the cache (write lock per chain).
3. Spike RSS by the size of the result set, which may not be reclaimed promptly.

The `LOAD_BATCH_SIZE` constant (500K) exists for the initial load but is not applied to the refresh query.

**Recommendation:** Add `.limit(LOAD_BATCH_SIZE)` and loop, similar to `load_chain`. The refresh doesn't need to be atomic — it's already re-reading a window, so batching is safe.

---

### ⚠️ Warning 2: `RwLock` Poisoning is Fatal and Silent

**File:** `token_cache.rs`

All lock accesses use `.expect("token cache lock poisoned")`:

```rust
let guard = store.read().expect("token cache lock poisoned");
let mut guard = store.write().expect("token cache lock poisoned");
```

If *any* write to the cache panics (e.g., a malformed token, an index overflow, an allocation failure), the `RwLock` is poisoned and every subsequent `get_tokens` call panics — taking down the entire indexer or RPC server. There's no recovery path.

In the `index` process, the write-through hooks run on the extractor's hot path. A single bad token in a block could poison the lock and crash the indexer.

**Recommendation:** Consider using `parking_lot::RwLock` (no poisoning), or catch the poison error and rebuild from the DB, or at minimum add `catch_unwind` around the write-through operations.

---

### ⚠️ Warning 3: Migration Takes a SHARE Lock on `token` Table

**File:** `migrations/2026-08-10_token_modified_ts_index/up.sql`

```sql
CREATE INDEX IF NOT EXISTS idx_token_modified_ts ON token (modified_ts);
```

On Base with 5M tokens, `CREATE INDEX` can take 30-60 seconds. During that time:
- The `token` table is locked with a SHARE lock (reads allowed, writes blocked).
- The indexer cannot insert new tokens — the extractor's hot path stalls.
- If the migration runs during a deployment, it blocks the very writes that the cache depends on for freshness.

**Recommendation:** Use `CREATE INDEX CONCURRENTLY` to avoid blocking writes. Diesel migrations run inside a transaction by default — this needs an explicit workaround or a migration runner that supports non-transactional migrations.

---

### 💡 Design Questions

**1. Why `std::sync::RwLock` instead of `tokio::sync::RwLock`?**
The docs say "the lock is never held across an `await`." `std::sync::RwLock` is correct for this constraint, but a `read()` call blocks the tokio worker thread if a write is in progress. If the write takes >100µs (possible with thousands of bitmap updates per block), a read on the same worker thread stalls. Consider measuring the write lock hold time under load.

**2. `Arc<Token>` — clone-on-read means every page query allocates.**
`query()` clones tokens via `(*self.tokens[idx as usize]).clone()`. For a 3000-token page, that's 3000 allocations per request. Could the API return `Arc<Token>` references instead, or use `Cow`?

**3. No eviction strategy for deleted tokens.**
The docs say "Tokens are never deleted from the DB; no eviction." True today, but if token deletion is ever added (spam token cleanup), the cache will serve stale entries forever. Consider a periodic full reconciliation or a tombstone mechanism.

**4. `RpcCache` removal changes behavior for `rpc`.**
The old `RpcCache` had a 7-minute TTL. Removing it means `last_traded` in the `rpc` process is now *more* stale than before (the old cache expired in 7 min; the new cache never refreshes `last_traded`). This is a net positive for `index` but a potential regression for `rpc`.

---

### ✅ Looks Good

- **Performance design** is sound: dense positional indices + RoaringBitmaps is the right approach. 4-order-of-magnitude improvement is credible.
- **Test coverage** is excellent: DB equivalence tests across a full filter/pagination matrix, write-through tests, refresh convergence tests, and a topology test mirroring the real `block_on` setup.
- **Documentation** is thorough — module docs explain the why, the how, the known limits, and the cost.
- **Opt-in via builder** is the right pattern — `analyze-tokens` keeps the SQL path.
- **The `refresh` overlap window** (10-minute lookback) is a clever race-condition fix for late-committing transactions.
- **Chain scoping** (commit 10) correctly prevents a shared DB from inflating memory or blocking startup.
- **Schema.rs changes** are reordering + new tables from other migrations, verified against the existing migration set.

---

*Reviewed by Thales via Hermes Agent (model: glm-5.2)*

crates/tycho-storage/src/postgres/token_cache.rs
```diff
@@ -0,0 +1,1644 @@
...
+        for pos in positions {
+            if self.last_traded[pos] <= threshold {
+                continue;
+            }
+            if total >= offset && entity.len() < limit {
+                entity.push((*self.tokens[pos]).clone());
+            }
+            total += 1;
+        }
+        WithTotal { entity, total: Some(total as i64) }
```
tvinagre, line 318: Out of scope, but total should be `u64` I believe

crates/tycho-storage/src/postgres/token_cache.rs
```diff
@@ -0,0 +1,1644 @@
...
+    }
+}
+
+/// The public face of the cache: one [`ChainTokenStore`] per chain plus the
+/// bookkeeping needed to keep them in sync with the database (see the module
+/// docs for the overall design).
+pub struct TokenCache {
+    chains: HashMap<Chain, RwLock<ChainTokenStore>>,
```
tvinagre, line 326: Do we need to have this in the interface? Chains are now somewhat deprecated. We're moving the application towards being single-chain by design - and all the hosted services also point to this design. I believe we should already move towards removing this from the interfaces.

crates/tycho-storage/src/postgres/token_cache.rs
```diff
@@ -0,0 +1,1644 @@
...
+    /// Maps the `chain` table's numeric ids to [`Chain`] values, so the delta
+    /// refresh can tell which store a returned row belongs to.
+    chain_ids: HashMap<i64, Chain>,
+    /// Largest `token.modified_ts` this cache has seen; `refresh` polls for rows
+    /// newer than this (minus a safety overlap).
+    last_sync: RwLock<NaiveDateTime>,
+}
+
+impl TokenCache {
+    pub async fn from_pool(
```
tvinagre, line 336: We have some traits that already expose this. We should be using them:
```
trait FromConnection<T> {
    async fn from_pool(pool: Pool<AsyncPgConnection>) -> Result<T, StorageError>;
    async fn from_connection(conn: &mut AsyncPgConnection) -> Result<T, StorageError>;
}
```

from tycho-storage

crates/tycho-storage/src/postgres/token_cache.rs
```diff
@@ -0,0 +1,1644 @@
...
+        chain: Chain,
+        chain_id: i64,
+    ) -> Result<(ChainTokenStore, NaiveDateTime), StorageError> {
+        let mut store = ChainTokenStore::default();
+        let mut idx_by_db_id: HashMap<i64, u32> = HashMap::new();
+        let mut max_modified_ts = NaiveDateTime::default();
+
+        let mut last_db_id = i64::MIN;
+        loop {
+            let batch: Vec<(orm::Token, Address)> = schema::token::table
```
tvinagre, line 406: it'd probably be cleaner (IMO) if we split cache operations from DB operations on the design

crates/tycho-storage/src/postgres/token_cache.rs
```diff
@@ -0,0 +1,1644 @@
...
+        Self::from_connection(&mut conn, chains).await
+    }
+
+    /// Loads the tokens of the given chains. Chains present in the `chain` table
+    /// but not requested are not loaded and not queryable; chain rows whose name
+    /// this build does not recognize are skipped with a warning, so a shared
+    /// database cannot prevent startup.
+    pub async fn from_connection(
+        conn: &mut AsyncPgConnection,
+        chains: &[Chain],
```
tvinagre, line 353: I'd prob bound this to a single chain as well - but there might be a reason why you added this slice on the interface?

crates/tycho-storage/src/postgres/token_cache.rs
```diff
@@ -0,0 +1,1644 @@
...
+            let batch: Vec<(orm::Token, Address)> = schema::token::table
+                .inner_join(schema::account::table)
+                .filter(schema::account::chain_id.eq(chain_id))
+                .filter(schema::token::id.gt(last_db_id))
+                .order(schema::token::id.asc())
+                .limit(LOAD_BATCH_SIZE)
+                .select((orm::Token::as_select(), schema::account::address))
+                .load(conn)
+                .await
+                .map_err(PostgresError::from)?;
```
tvinagre, line 415: Could we add a retry?

crates/tycho-storage/src/postgres/token_cache.rs
```diff
@@ -0,0 +1,1644 @@
...
+        let mut max_modified_ts = last_sync;
+        let mut refreshed = Vec::with_capacity(n_refreshed);
+        for (orm_token, address, chain_id) in rows {
+            let Some(chain) = self.chain_ids.get(&chain_id) else {
+                continue;
+            };
+            max_modified_ts = max_modified_ts.max(orm_token.modified_ts);
+            refreshed.push(to_model_token(&orm_token, &address, *chain));
+        }
+        self.upsert_tokens(&refreshed);
```
tvinagre, line 560: So, on refresh, a new token will get marked as "never_updated" and will never pass the filter until Tycho restarts. I'd like to see if we can find a way to solve this elegantly, as Tycho's design principle is to be an application that never restarts - and to have always the correct, latest data.

tvinagre (approved):
Added some comments for a future PR to address.
Because this is very high-important for production, and the suggestions don't touch public interfaces, I will merge this and we can address those in a future PR.
I have tested this PR in dev over multiple chains and compared the cache result vs the DB and we're reaching 100% accuracy. This gives me enough confidence to merge the PR.

PR #1303, opened 2026-08-10
feat(indexer): DeltaWindow skeleton for the entity cache

zizou0x:
## What

Skeleton for story 1 of the entity-cache plan: a fixed-depth in-memory block window (`DeltaWindow`) that will replace the drop-on-commit retention of `PendingDeltas`, folding evicted blocks into the entity cache. Interfaces and contracts only — method bodies are `todo!()` with the intended logic written as step comments. Design doc: https://propellerswap.slack.com/archives/C0B787HG10T/p1786093064886909

## Layout

- `services/state/window.rs` — `DeltaWindow` (retention rule `min(finalized, db_committed, tip - W, lowest active pin)`), `FoldSink` trait (fold-on-eviction callback, consumed by the entity cache in story 2), `WindowResolution` (in-window / below-floor / above-tip), `FloorPin` + `PinRegistry` (fill-time floor pinning, lazy expiry).
- `services/state/mod.rs` — module for the entity-cache serving layer; cache, fills, and routing land here in later stories.

Not in this PR: implementations, `PendingDeltas` rewiring, tests, metrics emission.

## Review feedback folded in

- `insert(message, sink)` folds-and-evicts as one operation; evicted blocks are never returned to the caller, so the fold-before-evict invariant is not caller-dependent.
- The reorg guard is keyed on `min(finalized, db_committed)` (not the folded height) and specifies `std::process::abort()` — the deltas pump unwraps in a spawned task, so a panic could kill only the pump and leave a frozen buffer serving.
- `commit_status` is derived from the db-commit watermark, not buffer bounds; the doc comment records the existing off-by-one at the oldest buffered block.
- Fold duration is part of the insert contract (`delta_window_fold_duration` histogram + slow-fold warning) since the fold runs under the lock all facade readers contend on.
- Pin expiry is lazy (checked at eviction, no timer task); `FloorPin::is_valid()` is answerable from the pin registry alone so the publish-time re-check cannot deadlock against an insert.

## Memory expectation

W ≈ 128 blocks per extractor vs today's measured 75–105 (`pending_deltas_buffer_size`), i.e. roughly 1.2–1.7× that gauge. The implementation PR must state the expected steady-state value for prod Ethereum.

## Update — 2026-09-15

Branch taken over for ENG-6290 and rebased onto main after the `ReorgBuffer` rewrite.

Contract changes since the review notes above:

- `FloorPin` and `PinRegistry` are removed. The retention rule is
  `min(finalized, db_committed, tip - W)`.
- `insert` no longer folds. Fold and eviction moved to `fold_and_evict`, which is a no-op
  until `min_fold_batch` blocks are evictable.
- Errors from `insert` and `fold_and_evict` return `Err` instead of aborting the process.
  The caller resets that extractor's window and keeps serving the other extractors.
- The window wraps the RPC-side `ReorgBuffer`, which never calls `drain_into_committing`,
  so its committing section stays empty.
- `W` is a constructor parameter. The default of 128 lands with the facade; per-chain
  values are ENG-6384.

Stacked next: ENG-6297 core, ENG-6298 fold, ENG-6300 facade and flags, ENG-6299 patch
capture.

### DIFF
--- a/crates/tycho-indexer/src/services/state/window.rs
+++ b/crates/tycho-indexer/src/services/state/window.rs
@@ -2,12 +2,12 @@
 //!
 //! The window retains roughly the last `W` blocks of [`BlockAggregatedChanges`] instead of
 //! dropping blocks as soon as the database commits them. Blocks leave the window only through
-//! [`DeltaWindow::insert`], which folds each evicted block into a [`FoldSink`] first — a block is
-//! removed only after its fold succeeded, so no block's deltas can be lost between the window and
-//! the long-lived store behind the sink.
+//! [`DeltaWindow::fold_and_evict`], which folds each evicted block into a [`FoldSink`] before
+//! removing it, so no block's deltas can be lost between the window and the long-lived store
+//! behind the sink.
 //!
 //! Retention rule: a block is evictable only when its number is at or below
-//! `min(finalized, db_committed, tip - W, lowest_active_pin - 1)`, all subtractions saturating.
+//! `min(finalized, db_committed, tip - W)`, the subtraction saturating.
 //!
 //! - `finalized`: folded data must never be affected by a reorg; reverts purge unfolded window
 //!   blocks only.
@@ -15,17 +15,13 @@
 //!   buffered; the database fallback path assumes every below-floor block is in the database.
 //! - `tip - W`: the serving depth. `W = max(finality horizon, maximum version age served from
 //!   memory) + margin` (~128 on Ethereum).
-//! - pins: an in-flight database fill pins the floor so the pinned block and everything above it
-//!   cannot be evicted underneath the fill (see [`FloorPin`]).
+//!
+//! Errors from [`DeltaWindow::insert`] and [`DeltaWindow::fold_and_evict`] are fatal: each one
+//! means the window no longer matches the chain or the database, with no in-process recovery —
+//! the caller must terminate the indexer process rather than keep serving.
 
 // Not yet constructed by production code; wired into `PendingDeltas` in a follow-up.
 #![allow(dead_code)]
-
-use std::{
-    collections::HashMap,
-    sync::{Arc, Mutex, Weak},
-    time::{Duration, Instant},
-};
 
 use deepsize::DeepSizeOf;
 use tycho_common::{
@@ -35,20 +31,14 @@ use tycho_common::{
 
 use crate::extractor::reorg_buffer::{BlockNumberOrTimestamp, CommitStatus, ReorgBuffer};
 
-/// How long a floor pin protects the window before eviction may pass it.
-///
-/// Expiry is lazy: it is checked when eviction computes its bound, not by a timer. A fill whose
-/// pin expired must discard its result (see [`DeltaWindow::publish_with_pin`]).
-const PIN_TIMEOUT: Duration = Duration::from_secs(30);
-
 /// Receives blocks evicted from a [`DeltaWindow`].
 pub(crate) trait FoldSink: Send + Sync {
     /// Merges one finalized, database-committed block into the long-lived store.
     ///
     /// Blocks arrive in ascending order. Delta values are absolute, so applying the same block
-    /// twice must be a no-op for implementations that tag values with their block number —
-    /// callers rely on this to retry a block whose fold previously failed. Implementations must
-    /// be fast: folds run synchronously and delay state reads while they run.
+    /// twice must be a no-op for implementations that tag values with their block number.
+    /// Implementations must be fast: folds run synchronously and delay state reads while they
+    /// run.
     ///
     /// An error means the block was not (fully) applied; the caller keeps the block buffered and
     /// propagates the error.
@@ -73,12 +63,15 @@ pub(crate) enum WindowResolution {
 /// Fixed-depth window of block deltas for a single extractor.
 ///
 /// Wraps the extractor's [`ReorgBuffer`] and owns the retention decision: today the buffer drains
-/// as soon as the database commits, here blocks are kept until they are finalized, committed,
-/// deeper than the configured depth, and above every active pin.
+/// as soon as the database commits, here blocks are kept until they are finalized, committed, and
+/// deeper than the configured depth.
 ///
 /// The type is `Sync` by composition but not internally synchronized: methods rely on the caller
 /// for exclusive access — one instance lives behind the per-extractor `Arc<Mutex<..>>` owned by
-/// the pending-deltas facade.
+/// the pending-deltas facade. That lock is also the coordination point between eviction and the
+/// database fill path: a fill must not publish state assembled from blocks that
+/// [`DeltaWindow::fold_and_evict`] folded out from under it, so the facade serializes the two
+/// behind the same lock (the fill path lands in a later story).
 ///
 /// Because committed blocks are retained, window contents and database rows overlap by up to
 /// `depth` blocks. Readers that merge window deltas with database queries and assume the two are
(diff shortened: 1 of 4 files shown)


zizou0x:
## Review Summary

Three-lens review (code quality / contract soundness / verification) of the interface-only skeleton. Since the deliverable is the documented contracts, contract defects were treated as code defects. External reviewers (codex, gemini) unavailable in this environment — skipped.

### Findings

| # | Severity | Finding | Resolution |
|---|----------|---------|------------|
| 1 | P1 | Pin validity TOCTOU: `is_valid()` (registry-only, no window lock) can be observed `true`, then an insert's lazy-expiry sweep evicts+folds past the pin before the fill publishes — publishing a stale base over newer folded data | Fixed: added `publish_with_pin()`, which validates the pin and runs the publish closure under the same exclusive access `insert` requires; `is_valid()` re-documented as advisory fast-path |
| 2 | P1 | `FoldSink::apply_folded` returned `()` — a failed fold (e.g. `MergeError` from delta application) would silently lose the block's deltas, contradicting the module's headline invariant | Fixed: returns `Result<(), StorageError>`; insert keeps the failed block buffered, evicts only the folded prefix, propagates the error; idempotent refolds make retry safe |
| 3 | P1 | Retaining committed blocks breaks readers that assume window/DB disjointness: `get_new_components` + the `/protocol_components` total would double-count everything created in the last W blocks | Fixed (contract): documented that facade reads merging both sides must bound window reads by `db_committed + 1`, not `floor()` |
| 4 | P2 | `abort()` rationale factually wrong: a pump panic does not leave a frozen buffer serving — it propagates via `try_join_all` → control channel → main `expect` and the process exits | Fixed: rewrote the justification (immediacy; no dependency on `panic = "unwind"` or the join wiring). The revert-below-watermark abort itself is kept — it tightens an already-fatal path (a below-`db_committed` purge already errors and hits the pump unwrap today) |
| 5 | P2 | Eviction bound is inclusive but `drain_blocks_until` is exclusive, and it returns `Err(NotFound)` when the target is below the floor — the steady state right after startup; naive implementation would kill the pump | Fixed: step comments specify the below-floor skip guard and `drain(h + 1)`, safe by parent-hash contiguity and `bound ≤ tip − depth` |
| 6 | P2 | "Not `Sync`" doc claim is false — every field is `Send + Sync`, so the type is auto-`Sync` (compiler-verified) | Fixed: reworded to "Sync by composition but not internally synchronized" |
| 7 | P2 | The pinned block itself was evictable (`lowest_active_pin` sat inclusively in the min), evicting exactly the block that makes a fill's base contiguous | Fixed: bound term is now `lowest_active_pin − 1` |
| 8 | P3 | `tip − depth` underflows on chains shorter than W (wraps in release → depth term silently ignored); `depth: 0` unvalidated | Fixed: saturating subtraction documented; `assert!(depth >= 1)` in `new` |
| 9 | P3 | `insert` documented no error conditions or the full-block precondition | Fixed: `# Errors` section + precondition added |
| 10 | P3 | `FoldSink` trait doc described the caller's locking scheme (implementation detail, prohibited on trait docs by the project guidelines) | Fixed: moved to `insert`; trait doc is now implementation-agnostic |
| 11 | P3 | `invalidated` flag described inconsistently in three places, and expired entries were never removed (unbounded registry growth if a handle leaks) | Fixed: sweep now removes expired entries; `invalidated` field dropped; `is_valid` = still registered |
| 12 | P3 | `pin_floor` returned `(u64, FloorPin)` — redundant with `FloorPin::height()`; empty-window pin height was undefined (no field holds the next insert height) | Fixed: returns `FloorPin` alone; empty-window semantics defined (`db_committed + 1`, else 0) |
| 13 | P3 | `FloorPin::drop` was an empty body — indistinguishable from an intentional no-op, though it's the one item with no dependency on unimplemented code | Fixed: implemented (registry-only removal, lock-ordered) |
| 14 | P3 | Timestamp resolution rounding unspecified; today's buffer lookup rounds up while a deltas_buffer docstring claims rounds down | Fixed: documented round-up; also noted `BelowFloor` is a deliberate fix vs today's below-floor fall-through to the buffer front |
| 15 | P3 | Module-wide `#![allow(unused_variables)]` would mask real bugs once bodies land under clippy `-D warnings` | Fixed: per-function allows; module keeps only `dead_code` |
| 16 | P3 | The commit-status boundary this design changes (`oldest buffered block → Committed`, off by one) had no test pinning today's behavior, so the follow-up couldn't demonstrate the fix | Fixed: added `committed_oldest_no` case to `test_get_commit_status` |
| 17 | P3 | The `finalized` gate is dead today — `db_committed ≤ finalized` by construction, so the min reduces to `db_committed`; a reader would believe finality is separately enforced | Fixed: documented as belt-and-braces against a change to the commit trigger |
| 18 | P3 | `floor() → Option<u64>` vs `tip() → Option<Block>` asymmetry; `u64` heights throughout instead of a newtype | Dismissed: retention math wants heights, `resolve` wants blocks; a `BlockNumber` newtype is follow-up-wide churn beyond this skeleton |
| 19 | P3 | `Option<CommitStatus>` expresses absence twice (`Unseen` exists) | Dismissed: `None` on empty window is load-bearing — the facade maps it to the DB-serving default; returning `Unseen` instead would turn startup requests into not-found errors |
| 20 | P4 | `mod state;` is private, narrower than the `pub(crate)` items inside | Dismissed: matches every sibling module in `services/`; widen when a consumer outside `services` appears |
| 21 | P4 | `pin_timeout` field duplicates `PIN_TIMEOUT` with no constructor override | Kept, documented as the test seam for expiry tests |

Verified as correct (no change): the `get_commit_status` off-by-one claim (confirmed against `drain_blocks_until(height + 1)` and the existing rstest), revert messages carrying `db_committed_block_height: None` with watermarks left alone (the revert's `finalized` comes from the abandoned branch — adopting it would be wrong), the timestamp-beyond-tip clamp, `Weak<PinRegistry>` soundness, and the `DeepSizeOf` composition.

### Verification

- **Tests**: 340/340 pass (nextest, DB-free filter; DB tests need postgres, not run locally)
- **Lint**: clean — `cargo +nightly clippy --workspace --all-targets --all-features -- -D warnings`
- **Format**: clean — `cargo +nightly fmt --all --check`
- **Docs**: clean — `RUSTDOCFLAGS="-D rustdoc::broken_intra_doc_links" cargo doc -p tycho-indexer --no-deps --all-features`

### Commit

63f44c09d — `fix(indexer): resolve code review findings for PR #1303`

crates/tycho-indexer/src/services/state/window.rs
```diff
@@ -0,0 +1,357 @@
...
+        &mut self,
+        message: &BlockAggregatedChanges,
+        sink: &dyn FoldSink,
+    ) -> Result<(), StorageError> {
+        // Step 1 — revert message (`message.revert == true`):
+        //
+        // If the purge target would remove any block at or below
+        // `min(self.finalized, self.db_committed)`, the database holds rows from the abandoned
+        // branch and persisted state is never rolled back, so there is no in-process recovery.
+        // Log at error level and `std::process::abort()`. Returning an error would also end the
```
zizou0x: I think we can return an error. But it needs to be considered fatal by the indexer process.

crates/tycho-indexer/src/services/state/window.rs
```diff
@@ -0,0 +1,357 @@
...
+        //
+        // Let `bound = self.eviction_bound()`. Skip this step when `bound` is `None` or below
+        // the oldest buffered block number — the steady state right after startup, where
+        // `ReorgBuffer::drain_blocks_until` would error because the target is not buffered.
+        //
+        // Fold before evicting: for each buffered block up to `bound` in ascending order call
+        // `sink.apply_folded(&self.extractor, &block)`, timing each call into a
+        // `delta_window_fold_duration` histogram (label: extractor) and logging a warning above
+        // a slow-fold threshold. On a fold error stop folding and evict only the successfully
+        // folded prefix before returning the error.
```
zizou0x: Fold error are also quite unexpected. We should error the whole process too for now until we have more knowledge on the fold errors and how to handle them.

crates/tycho-indexer/src/services/state/window.rs
```diff
@@ -0,0 +1,357 @@
...
+    /// Pins the current floor and returns a release-on-drop handle.
+    ///
+    /// While the pin is active (not dropped, not expired) eviction never removes the pinned
+    /// block or anything above it, so window deltas from the pinned height upward stay available
+    /// for a fill's top-up. The pinned height is [`FloorPin::height`].
+    ///
+    /// On an empty window the pin registers at `db_committed + 1` (the next height that can
+    /// enter the window), or at 0 before any commit is observed — nothing is evictable in either
+    /// case, so every block inserted after the pin is protected.
+    pub(crate) fn pin_floor(&mut self) -> FloorPin {
```
zizou0x: Why do we need this "pin" feature at all? Where do we plan to use it?

crates/tycho-indexer/src/services/state/window.rs
```diff
@@ -0,0 +1,357 @@
...
+    /// - `StorageError::Unexpected` when the message does not extend the buffered chain
+    ///   (parent-hash mismatch).
+    /// - `StorageError::NotFound` when a revert targets a hash that is not buffered.
+    /// - Any error returned by [`FoldSink::apply_folded`]; the failed block and everything above it
+    ///   stay buffered, and absolute delta values make the retry refold a no-op.
+    #[allow(unused_variables)]
+    pub(crate) fn insert(
+        &mut self,
+        message: &BlockAggregatedChanges,
+        sink: &dyn FoldSink,
```
zizou0x: I don't think this should be part of inserts. I'd prefer if the flow was:

- insert are straightforward
- we then call a different "fold and evict" method with the folding logic (basically the in memory cache)

crates/tycho-indexer/src/services/state/window.rs
zizou0x: Done in f586604d3 — `insert` returns the error, and the contract (module doc + `# Errors`) states every `insert`/`fold_and_evict` error is fatal with no in-process recovery. Today's pump wiring already terminates the process on an insert error (unwrap → JoinError → try_join_all → control channel → main exits), so the fatality requirement is documented for whoever rewires the pump rather than re-implemented.

crates/tycho-indexer/src/services/state/window.rs
zizou0x: Done in f586604d3 — `fold_and_evict` propagates the first fold error (after evicting the successfully folded prefix, so nothing already applied is replayed as lost), documented as fatal until fold failure modes are better understood.

crates/tycho-indexer/src/services/state/window.rs
zizou0x: It existed for the story-2 DB fill path: a fill reads a base from the database, then tops up from window deltas, and the pin was meant to stop eviction from folding those blocks into the cache mid-fill (and the fill from then overwriting newer folded values with a stale base). With fold-and-evict now an explicit method instead of a side effect of insert, the facade decides when eviction runs and can serialize fills against it with the per-extractor lock it already owns — the pin registry (weak refs, timeouts, lazy expiry sweeps) was machinery for a race that simpler coordination avoids. Removed in f586604d3; the `DeltaWindow` doc records the fill/eviction coordination point so story 2 doesn't reintroduce the race.

crates/tycho-indexer/src/services/state/window.rs
zizou0x: Done in f586604d3 — `insert(message)` only buffers/purges and raises the watermarks; `fold_and_evict(sink)` owns bound computation, folding (with the duration histogram), and eviction. Fold-before-evict stays a single operation per block inside `fold_and_evict`, so the no-loss invariant is unchanged.

crates/tycho-indexer/src/services/state/window.rs
```diff
@@ -0,0 +1,282 @@
...
+    depth: u64,
+    /// Highest `db_committed_block_height` seen on any inserted message. `None` until the first
+    /// commit is observed; nothing is evictable before that.
+    db_committed: Option<u64>,
+    /// Highest `finalized_block_height` seen on any inserted message.
+    finalized: Option<u64>,
+    /// Minimum number of evictable blocks required before a fold runs. Amortizes folding: with
+    /// 1 every evictable block is folded as soon as possible; larger values trade `min_fold_batch`
+    /// extra buffered blocks for folds that run `min_fold_batch` times less often.
+    min_fold_batch: u64,
```
brunoguerios: **`min_fold_batch` duplicates the DB commit batch**

Window depth is `tip - min(finalized, db_committed, tip - W)`, so the buffer holds
`max(tip - finalized, tip - db_committed, W)` blocks. `db_committed` advances in jumps of
`commit_batch_size`, because the extractor drains only once that many finalized blocks
have accumulated.

Values in prod today:

| Chain | `--database-insert-batch-size` |
| --- | --- |
| Arbitrum | 1000 |
| BSC, Polygon | 512 |
| Base, Robinhood | 32 |
| Ethereum, Unichain | unset (0) |

When the `db_committed` term binds, one commit makes that many blocks evictable at once.
Folding is already batched, by 1000 on Arbitrum and by 32 on Base.

When `tip - W` binds, blocks become evictable one at a time. That is Ethereum and
Unichain, where the batch is unset. On Ethereum it means one fold per 12 s, against a fold
we expect to take well under a millisecond.

So `min_fold_batch` amortizes only where amortization is not needed.

It also costs margin. Batched folding is what lets the cache move past a version a request
already resolved, which is the read-guard case in `state/mod.rs`. Folding every evictable
block keeps the cache floor on the window floor, and the read guard then fires only on
`db_committed` jumps, which we cannot avoid.

Proposal: remove `min_fold_batch`. If `delta_window_fold_duration` later shows a problem,
add it back with evidence behind it.

zizou0x: I think they semantically serve different purposes.

- `database-insert-batch-size` is used to batch inserts on the extractor side
- `min_fold_batch` is used to keep blocks available in the window

Having them separated can be useful. In a later plan we want to expose blocks from the window to clients through a new HTTP endpoint, this would allow them to catch up on disconnects and replay some recent historical state. We will need to be able to set these two parameters independently, for example keep inserting to the db whenever we can but keep blocks available in the window for a little while.

brunoguerios:
Taking this branch over for ENG-6290 as agreed. I will rebase it onto main and force-push,
then stack the implementation PRs on it. #1306 is based on this branch and will need a
rebase afterwards; ping me when you want to coordinate that.

brunoguerios:
Superseded by #1456, which reviews the whole DeltaWindow stack against main as one branch. The skeleton and the review fixes from this PR are in it (3d6c39f9f, 10f03fb19, e795685d4). We took the work over from @zizou0x. Closing to keep review in one place; the branch goes away with it.

PR #1316, opened 2026-08-12
fix(tycho-simulation): tolerate empty FeedMessages in stream decoder

brunoguerios:
## Problem

fynd crash loops from short extractor stalls showing `Fatal("Missing block!")` before dying

Over 2026-08-11/12 we lost fynd pods to stalls that were transient and recoverable:
• Robinhood — extractor stalls >3s on token metadata in the block hot path. 23 crashes in 24h.
• Arbitrum — an ingress-nginx reload closed all WebSockets. Both pods died 3.0s after the disconnect, while tycho-client was still reconnecting.
• Unichain/Polygon — a shared Substreams gRPC outage paused all extractors >3s.

## Root cause

The cause is in the Tycho stream decoder, not in fynd. When no extractor delivers a block within one block time (+ 3s timeout), `BlockSynchronizer` emits a `FeedMessage` with an empty `state_msgs` map. That is its status heartbeat — it carries `sync_states` so consumers can watch Delayed/Stale and recovery. `decode()` treated it as fatal, fynd propagated the `Err` with `?`, and the solver shut down. tycho-client already absorbs these stalls; the decoder killed
the consumer first.

## Fix

`decode()` now returns a status-only `Update` — empty state maps, `sync_states` passed through, block number and partial-block flag repeated from the last decoded block. An empty message before the first block stays an error.
Stall entry logs at `warn!` with `sync_states`, repeats at `debug!`, recovery at `info!`.

## Worth knowing

after this lands a stall no longer appears as a restart, it appears as aging data. Restart-based alerts will not catch it. 

## Follow-up

- raise `SolveError::MarketDataStale` in the fynd solve path
- expose per-protocol Stale/Delayed as metrics
- add staleness alerts.

### DIFF
--- a/crates/tycho-simulation/src/evm/decoder.rs
+++ b/crates/tycho-simulation/src/evm/decoder.rs
@@ -271,7 +271,12 @@ where
     }
 
     /// Decodes a `FeedMessage` into a `BlockUpdate` containing the updated states of protocol
-    /// components
+    /// components.
+    ///
+    /// A message with an empty `state_msgs` map (no synchronizer advanced this tick) decodes
+    /// into a status-only update: empty state maps, `sync_states` passed through, and the block
+    /// number of the last decoded message. Such a message before the first decoded block is an
+    /// error.
     pub async fn decode(&self, msg: &FeedMessage<H>) -> Result<Update, StreamDecodeError> {
         // stores all states updated in this tick/msg
         let mut updated_states = HashMap::new();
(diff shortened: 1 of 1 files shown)


crates/tycho-simulation/src/evm/decoder.rs
```diff
@@ -280,6 +291,41 @@ where
...
+                // otherwise repeat every block time for as long as the stall lasts.
+                warn!(
+                    current_block_number,
+                    sync_states = ?msg.sync_states,
+                    "All synchronizers stalled; emitting a status-only update"
+                );
+            } else {
+                debug!(current_block_number, "Empty FeedMessage; emitting status-only update");
+            }
+            return Ok(Update::new(current_block_number, HashMap::new(), HashMap::new())
```
tvinagre, line 324: This is a breaking change in the sense that, previously, all the consumers of the Decoder could rely on the fact that we're emitting a single message per-block - which is not true anymore. 
So, before we proceed with this rollout, please verify if this behavior won't break Fynd, or any other downstream consumer of TychoSim stream, including the integration tests, or Hindsight.

brunoguerios: Nice catch! Thanks for flagging this. 
Fynd is safe/fine with empty updates like these, but Hindsight and Integration Tests indeed require a small adjustment to behave the same way (i.e. ignore empty updates).
This should be fixed alongside the `SolveError:MarketDataStale` which is suggested as a follow-up, but should ideally be coordinated to land alongside this PR, otherwise we risk to keep solving on stale market data indefinitely.

zizou0x: Would be nice to have a document explaining the whole update message flow semantics (tycho-indexer emitting per extractor update -> tycho-client receiving and syncing them -> tycho-simulation decoder handling the synced messages, and maybe even how Fynd treats the sync state). 

I'm unsure about some particularities, for example when `tycho-client` emits a message with a stall extractor, if the extractor then emits a new update, is it emitted together with the next update message or right when it arrives with a new message pointing to the same block we already emitted the message for?

brunoguerios:
> This should be fixed alongside the SolveError:MarketDataStale which is suggested as a follow-up, but should ideally be coordinated to land alongside this PR, otherwise we risk to keep solving on stale market data indefinitely.

Closing this PR as it's currently high effort/low benefit.
High effort because it would require a bigger refactor on downstream consumers due to the context mentioned above.
Low benefit because most of the failures it handles were already fixed or reduced significantly on the indexer side.

PR #1318, opened 2026-08-13
feat: integrate Tempest propAMM (vm:tempest)

JonathanChow2000:
Adds Tempest, Flowdesk's propAMM on Ethereum mainnet, as `vm:tempest`: a substreams package, a VM simulation adapter, and execution through the existing generic `PropAMMExecutor`.

Router (ERC1967 proxy) `0x00000003f1ec2379e79F58E12EC6C4F51Ee92149` · vault `0xc9d748e601d9984a43da0b80e5b91dc28d31d9fb` · shared Flashbots `PrioUpdateRegistry` `0xDa7AfeeD01fe625CF15d187a19f94B45f00b8C5F`, the same registry `ethereum-fermiswap` and `ethereum-bopamm` already depend on.

> **Commit 1 is a change to shared simulation behaviour.** It was briefly split out as #1458; that is now closed and it stays here. See [TokenProxy](#tokenproxy-shared-simulation-behaviour) below — please review it on its own terms, since it affects every protocol that simulates through the proxy.

## How it works

Tempest prices from a quote **lane** the maker commits to the shared registry. The router only reads a lane whose `updateTimestamp` is inside `laneWindow` of `block.timestamp` and reverts `StaleUpdate` otherwise, so the package emits an `override_block_timestamp` attribute to pin simulation to the committed quote.

**Indexing** — 8 substreams modules. One component per registered pair, identified by `keccak256(router, token0, token1)`. The router is mixed in deliberately: `keccak256(token0, token1)` is the venue's own `laneFor` value, which is not deployment-specific and already collides with `ethereum-fermiswap`'s WETH/USDT component id. Because the id can no longer be inverted from a registry lane index, `store_pair_registered` records a `lane:{laneFor}` → id mapping at registration.

The vault is event-sourced from `VaultUpdated` rather than configured, since the router can migrate it. A rotation re-reads every tracked token's balance and repoints `balance_owner` on existing components. Vault balances live in one `set_sum` store: rebases `set`, movements `sum`. The router implementation is reached through DCI rather than the component contract set, which is frozen at creation and would go stale on the next `upgradeToAndCall`.

**Simulation** — `TempestAdapter` settles a real `swap` and reports measured gas. Tokens carry `self_contained_tokens`, which is what surfaced the `allowance` behaviour described below.

## TokenProxy: shared simulation behaviour

Commit 1 makes `TokenProxy.allowance` agree with `transferFrom`. `transferFrom` skips the allowance check when the approval system is globally disabled, but `allowance` returned the custom-approval mapping, which nothing populates unless `approve` was called in-simulation. A holder flagged by `set_has_custom_approval` — what `self_contained_tokens` does — therefore reported a zero allowance for state a transfer would happily spend. Tempest is the first protocol to hit it, because its quote reads `allowance(vault, router)` and so saw zero and refused to price.

**Why a 21-line source change rewrites the whole artifact.** The committed artifact had drifted from source. `assets/TokenProxy.bin-runtime` on `main` was compiled with **solc 0.8.29** while `foundry.toml` pins **0.8.20** — visible in the metadata tail (`736f6c6343` `00081d` vs `000814`). It was also built from older source: the committed runtime contains no `decimals`, `name`, `symbol` or `totalSupply` selectors at all, so the metadata overrides already present in `TokenProxy.sol` were never in the deployed bytecode. Rebuilding at the pinned compiler therefore ships this fix *and* previously-unbuilt behaviour other commits had added to source. That is the whole of the bytecode diff — it is not this one-line change doing it.

**A narrower alternative, if you would prefer it.** This makes an `allowance` read stop being a meaningful safety check under simulation, globally and permanently. Two options that unblock Tempest without touching shared semantics:

1. Have `TokenProxyOverwriteFactory::set_has_custom_approval` write the approval amount as well, so the mapping the current `allowance` reads is actually populated.
2. Add an `approval_owner` / `approval_spender` static attribute mirroring the existing `balance_owner`, and seed the approval from it.

I implemented the global one because it needs no per-protocol attribute, but I would rather you chose than inherit it.

**Execution** — no dedicated executor. Tempest implements `IPropAMM`, so `vm:tempest` resolves to the generic `PropAMMExecutor` and the venue address travels in the swap data via a `pamm_address` static attribute. Verified end-to-end: Rust-encoded calldata → TychoRouter → `PropAMMExecutor` → live venue.

## Taker access — please read

Settlement is open to any caller **only because an operator flag is on**. `Tempest._checkTaker` is intact:

```solidity
if (blockedTaker[msg.sender]) revert TakerIsBlocked();
if (!openTakerAccess && !allowedTaker[msg.sender]) revert TakerNotAllowed();
```

On chain now: `openTakerAccess() == true`, `allowedTaker(TychoRouter) == false`, `blockedTaker(TychoRouter) == false`.

Neither `isActive` nor `quote` consults taker access. If Flowdesk turns the flag off, or blocks the router address, Tycho keeps quoting while every fill reverts — quotable-but-unfillable. Flowdesk is being asked to `addTaker(<TychoRouter>)` so the integration survives a flag flip. Nothing here blocks merging; it is a durability matter, and worth knowing before this serves flow.

## Freshness constraint for solvers

A fill must land within `laneWindow` maxAge — currently 12s on Ethereum — of the maker's lane commit, or the router reverts `StaleUpdate`. In practice the builder orders the maker's quote transaction directly ahead of the fill. **A solver that cannot order behind our quote transaction will not fill.** This is a property of the venue, not of the integration.

## Testing

| suite | result |
|---|---|
| `ethereum-tempest` unit tests | 15 passed |
| `substreams pack` | clean |
| `TempestAdapter` fork tests | 15 passed |
| Tempest router fork test (Rust calldata → router → venue) | passed |
| `cargo test -p tycho-execution --all-features` | 153 passed, 1 failed |
| `token-proxy-contracts` | 12 passed, 1 skipped |
| `cargo test -p tycho-simulation --all-features` | unchanged vs `main` — see below |

The one `tycho-execution` failure is `permit2::tests::test_permit`, which fails identically on a clean `origin/main` checkout.

`tycho-simulation`, run twice on each side on the same machine, to show the TokenProxy change moves nothing:

| | run 1 | run 2 |
|---|---|---|
| `main` | 1276 passed / 1 failed | 1275 passed / 2 failed |
| this branch | 1276 passed / 1 failed | 1275 passed / 2 failed |

Identical. The variance is two tests that are flaky on `main` independently of this change — `evm::protocol::vm::state::tests::test_engine_block_advance_invalidates_cached_limits` and `evm::decoder::tests::test_euler_hook_low_pool_manager_balance`. `clippy`, `cargo +nightly fmt` and `forge fmt` clean; runtime bytecode fixtures up to date.

## Open question: Tempest is already served, so this exposes one vault twice

This is the one item here that can produce a **wrong route** rather than a loud failure, so it needs your decision before merge.

`main` already serves Tempest on the generic Titan path. `default_served_pamms()` carries `("tempest", "0x00000003f1ec2379e79f58e12ec6c4f51ee92149", 160_000)` — the same router this PR indexes — and `default_denied_pamms()` is now `Vec::new()`. (An earlier draft of this description claimed the opposite, that the venue was still denied over the taker allowlist. That was wrong and has been corrected; the restriction was lifted on 2026-09-11 when Tempest was added to the served list.)

So with this PR merged the same router, backed by the **same shared vault**, is published twice:

| | source | gas |
|---|---|---|
| `pricelevelstream:tempest` | Titan stream, on `main` today | 160_000 |
| `vm:tempest` | this PR's indexing | 130_000 |

Both settle through `PropAMMExecutor`. A solver consuming both sees two components whose depth is the same vault inventory — double-counted depth. Worse, a split route across "both venues" is exactly what the venue's vault-cover check rejects at settlement: each component independently says the size is fine, and the fill reverts `InsufficientLiquidity`.

Three ways out, and it is your call which:

1. Keep `vm:tempest` and drop Tempest from `default_served_pamms()`.
2. Keep the stream path and drop this PR's indexing.
3. Keep both and state how they are de-duplicated downstream.

I have not touched `price_level_stream/config.rs` — this branch does not modify that file — because the choice is yours, not mine.

**While you are here:** the stream config measures 120k–155k and budgets 160_000; this PR's manifest declares `protocol_gas: 130000`. The adapter reports measured gas, so the manifest value is only a fallback constant and neither number is load-bearing — but two figures for one venue in one repo will get asked about, and whichever option above you pick should probably collapse them.

## Known limitations

- **The protocol-testing harness has never run against this package.** I have no Substreams API token. Everything above is unit-, fork- and code-path verified only. `test-protocols` is currently skipping in CI because the changed-substreams gate resolves empty while the branch conflicts with `main`; it would be the most valuable check to get green.
- Both harness tests carry `skip_execution: true`. Execution is only exercised by the Foundry fork test, which restamps the lane with `vm.store`. Enabling it in the harness means moving `stop_block` past the upgrade at block 25744018 to the next committed lane, growing the indexed range from ~36k blocks to ~330k.
- Only USDC/WETH had a committed lane within the harness block range. All three registered pairs carry a ladder by block 25963848.
- `executor_deployments.json` adds a `vm:tempest` recipe for `PropAMMExecutor`. That executor is stateless and already deployed at `0xc196f22d69d41A2B1DbF6ef8d28C09bc2CA4E4fa` for `pricelevelstream` — pointing `vm:tempest` at the existing deployment instead would avoid a second identical contract, but that means an `executor_addresses.json` entry, which a previous review asked me not to add. Your call.

- `docs/for-solvers/supported-protocols.md` and `config/executor_addresses.json` are deliberately untouched, per earlier review.


docs/for-solvers/supported-protocols.md
zach030, line 1: No need to update this doc in this PR, we'll update it once the integration is done.

crates/tycho-execution/config/executor_addresses.json
zach030, line 1: No need to update this config in this PR, we'll update deploy the executor and update the config.

protocols/substreams/ethereum-tempest/src/modules.rs
```diff
@@ -0,0 +1,769 @@
...
+                    .get_last(&id)
+                    .is_none()
+                {
+                    continue;
+                }
+                if !ev.registered {
+                    builder.change_component_pause_state(&id, true);
+                } else if !first_registrations.contains(&id) {
+                    builder.change_component_pause_state(&id, false);
+                }
```
zach030: There's a case here that could cause an error. If `addPair` → `removePair` → `addPair` happens in the same block (three txs), the deltas are ("", "1"), ("1", "0"), ("0", "1"), and `first_registrations` contains this id because of the first one. So the three events are processed in order:

- registered=true, in first_registrations → skipped
- registered=false → pause(true)
- registered=true, still in first_registrations → skipped again, so the pause is never lifted

Result: the component is created but stays paused, even though the pair is actually in a registered state onchain.

zach030 (commented):
Thanks! This looks great. Left a few small comments.

protocols/testing/fixtures/Tempest.runtime.json
zach030, line 1: Could you add Tempest into the [update_runtime_bytecode.sh](https://github.com/propeller-heads/tycho/blob/main/protocols/testing/scripts/update_runtime_bytecode.sh) script?

docs/for-solvers/supported-protocols.md
JonathanChow2000: Removed — the file is back to its state on `main`. Done in e5561e4.

crates/tycho-execution/config/executor_addresses.json
JonathanChow2000: Removed. Kept the `executor_deployments.json` entry, since `update_runtime_bytecode.sh` resolves the contract name and constructor args from it to build the fixture — happy to drop that too if you would rather it waited for the deployment as well. Done in e5561e4.

protocols/substreams/ethereum-tempest/src/modules.rs
JonathanChow2000: Good catch — that was a real bug, thanks. Fixed in e5561e4.

The decision is now `pause_transition`, and the creation entry is **consumed** the first time a registration is seen rather than just tested:

```rust
fn pause_transition(
    pending_creation: &mut HashSet<String>,
    id: &str,
    registered: bool,
) -> Option<bool> {
    if !registered {
        return Some(true);
    }
    // `remove` reports whether this is the still-unconsumed creation event.
    if pending_creation.remove(id) {
        None
    } else {
        Some(false)
    }
}
```

So for your three-tx block the first event consumes the entry and emits nothing, the second pauses, and the third no longer looks like a creation and lifts the pause.

Three tests cover it: add/remove/add, add/remove (ends paused, matching the chain), and re-registration in a later block with no pending creation. I checked the add/remove/add one fails against the previous `contains` logic, so it genuinely pins the case rather than just passing.

protocols/testing/fixtures/Tempest.runtime.json
JonathanChow2000: Added in e5561e4, alongside the other VM executors:

```
    "FermiSwap|ethereum|vm:fermiswap"
    "Tempest|ethereum|vm:tempest"
```

The contract and constructor args resolve from `executor_deployments.json`, so `--check` now catches drift on the Tempest fixture like every other executor.

protocols/substreams/ethereum-tempest/src/modules.rs
```diff
@@ -0,0 +1,827 @@
...
+/// reverts. Rather than depend on that storage arriving, write the slot explicitly from the
+/// `Upgraded` event the proxy emits at deployment and on every upgrade.
+///
+/// Re-asserting an already-correct value is a no-op, so this is safe even once the underlying
+/// extraction gap is closed.
+fn assert_implementation_slot(
+    builder: &mut TransactionChangesBuilder,
+    config: &Config,
+    implementation_store: &StoreGetString,
+) {
```
zach030: Is this handling here to work around the [test case](https://github.com/JonathanChow2000/tycho/blob/db32d583696be6aa2cbdff61021dfd0f0416f00c/protocols/substreams/ethereum-tempest/integration_test.tycho.yaml#L10-L17) where initializing an account only gives you the code but not the storage? My understanding is that if we sync from the start block, this slot would be populated normally and wouldn't need special handling. Is that right?

zach030: I think this is related to the `initialize_account` bug. Opened an issue to track it: #1358

protocols/adapter-integration/evm/src/tempest/TempestAdapter.sol
```diff
@@ -0,0 +1,325 @@
...
+        if (sellAmount == 0) {
+            revert TooSmall(0);
+        }
+
+        uint256 amountOut = tempest.quote(sellToken, buyToken, sellAmount);
+        if (amountOut == 0) {
+            revert TooSmall(0);
+        }
+
+        calculatedPrice = Fraction(amountOut, sellAmount);
```
die-herdplatte: This isn't the marginal price after selling `sellAmount`.
I think dropping the `PriceFunction` capability is the best move here, simulation will fall back to
```
x1 = limit / 100
x2 = 1.01 · x1
price = (quote(x2) − quote(x1)) / (x2 − x1)
```
which is more accurate.

protocols/substreams/ethereum-tempest/src/utils.rs
```diff
@@ -0,0 +1,160 @@
...
+    let (token0, token1) = sort_tokens(token_a, token_b);
+
+    let mut out = [0u8; 32];
+    let mut hasher = Keccak::v256();
+    hasher.update(token0);
+    hasher.update(token1);
+    hasher.finalize(&mut out);
+
+    format!("0x{}", hex::encode(out))
+}
```
die-herdplatte: This should contain some deployment-specific input.
Component IDs are chain-global and collisions are silently discarded.
I know FermiSwap currently has the same format but this is exactly where collisions could take place.

protocols/substreams/ethereum-tempest/src/modules.rs
```diff
@@ -0,0 +1,753 @@
...
+        }
+
+        let tycho_tx: Transaction = tx.into();
+        let (token0, token1) = sort_tokens(&event.token0, &event.token1);
+        let id = component_id(token0, token1);
+
+        // A re-registered pair already has a component; `map_protocol_changes` unpauses it instead.
+        if !first_registrations.contains(&id) {
+            return;
+        }
```
die-herdplatte: If a pair gets added, removed, and re-added within one block this emits the component twice

protocols/substreams/ethereum-tempest/src/modules.rs
die-herdplatte, line 1: Can we somehow handle `VaultUpdated` instead of hardcoding the vault?

protocols/substreams/ethereum-tempest/src/modules.rs
```diff
@@ -0,0 +1,753 @@
...
+        for (log, _) in raw_tx.logs_with_calls() {
+            if let Some(transfer) = erc20::events::Transfer::match_and_decode(log) {
+                if transfer.from.as_slice() == vault {
+                    vault_token_deltas.push((
+                        tycho_tx.clone(),
+                        log.ordinal,
+                        log.address.clone(),
+                        BigInt::zero() - transfer.value,
+                    ));
+                } else if transfer.to.as_slice() == vault {
```
die-herdplatte: A self-transfer would only debit the vault's balance. Using `tycho_substreams::balances::extract_balance_deltas_from_tx` would reduce LOC and fix this.

protocols/substreams/ethereum-tempest/src/modules.rs
```diff
@@ -0,0 +1,753 @@
...
+        for entity_change in &tx_changes.entity_changes {
+            builder.add_entity_change(entity_change);
+        }
+    }
+
+    for trx in block.transactions() {
+        let tx: Transaction = trx.into();
+        let builder = transaction_changes
+            .entry(tx.index)
+            .or_insert_with(|| TransactionChangesBuilder::new(&tx));
```
die-herdplatte, line 538: It would be more efficient if this were inserted lazily

protocols/substreams/ethereum-tempest/src/modules.rs
```diff
@@ -0,0 +1,753 @@
...
+                let id = component_id(&ev.token0, &ev.token1);
+                if pair_registered_store
+                    .get_last(&id)
+                    .is_none()
+                {
+                    continue;
+                }
+                if let Some(paused) = pause_transition(&mut pending_creation, &id, ev.registered) {
+                    builder.change_component_pause_state(&id, paused);
+                }
```
die-herdplatte, line 569: Could this allow an added or re-added component be marked as active while the router is paused? Is admin functionality also affected by the pause state?

die-herdplatte (changes requested):
Thanks for the PR!
I left some comments, please have a look.

protocols/adapter-integration/evm/src/tempest/TempestAdapter.sol
JonathanChow2000: Agreed — dropped in `0755b5f`. `PriceFunction` is gone from `getCapabilities` and `price` now reverts `NotImplemented`, so simulation falls back to the numerical derivative you describe.

protocols/substreams/ethereum-tempest/src/utils.rs
JonathanChow2000: Fixed in `48eedd99` — ids are now `keccak256(router, token0, token1)`.

You were right that this is exactly where it bites, and it turned out not to be hypothetical: `ethereum-fermiswap`'s integration fixture already claims the same WETH/USDT id `0x2b2f5776…` that this package was emitting.

I kept the id at 32 bytes rather than adopting the `pamm ++ token0 ++ token1` form the price-level-stream protocols use, because `string_to_bytes32` rejects anything longer and a VM adapter receives the id as a `bytes32` pool id. One consequence: a registry lane index can no longer be inverted to a component id, so `store_pair_registered` records a `lane:{laneFor}` → id mapping at registration and `add_override_block_timestamp` resolves through it. `TempestAdapter._poolId` mirrors the new scheme — `laneFor` stays what keys the registry lane, and the two are now deliberately distinct.

protocols/substreams/ethereum-tempest/src/modules.rs
JonathanChow2000: Confirmed and fixed in `0755b5f`.

Same block shape as the pause bug you found earlier, but in `get_new_pairs`, which that fix didn't cover — both `registered = true` events passed the `first_registrations` check, so the component was emitted twice. The creation entry is now consumed on the first one, so it is emitted exactly once and the third event falls through to the unpause path instead.

protocols/substreams/ethereum-tempest/src/modules.rs
JonathanChow2000: Done in `48eedd99`. A new `store_router_state` module sources the vault from `VaultUpdated`, and `vault_address` is gone from the module params entirely.

I checked the chain before wiring it: `VaultUpdated` has fired exactly once, at block 25637319, which is the package's `initialBlock`. So the key is populated from the first block the package sees and no bootstrap is needed.

protocols/substreams/ethereum-tempest/src/modules.rs
JonathanChow2000: Confirmed and fixed in `0755b5f`.

The old code chained `if from == vault` / `else if to == vault`, so a vault-to-vault transfer took the first branch and only ever debited. Switched to `extract_balance_deltas_from_tx`, which credits and debits independently and nets such a transfer to zero — and it did cut the block down, as you expected.

protocols/substreams/ethereum-tempest/src/modules.rs
JonathanChow2000: Done in `0755b5f` — though as a pre-filter rather than a deferred insertion: the loop now skips transactions that touch neither the router nor the registry, which is nearly all of them in a block.

protocols/substreams/ethereum-tempest/src/modules.rs
```diff
@@ -0,0 +1,827 @@
...
+    vault: &[u8],
+    block: &Block,
+    mut pending_creation: HashSet<String>,
+    changes: &mut Vec<TransactionEntityChanges>,
+) {
+    // The router's implementation is deliberately absent: it is an upgradeable ERC1967 proxy, and a
+    // component's contract set is frozen at creation, so a hardcoded implementation goes stale on
+    // the next `upgradeToAndCall` and leaves the proxy delegatecalling a codeless account. DCI
+    // discovers it instead — see `add_entrypoints`.
+    let contracts = [config.router_address.as_slice(), vault, config.registry_address.as_slice()];
```
die-herdplatte: The comment directly above excludes the router's implementation *because* "a component's contract set is frozen at creation, so a hardcoded implementation goes stale" — and then this line puts the vault into that same frozen set. `store_router_state`'s own doc (L187-188) describes the vault as movable for exactly the same reason: "the router can migrate it and a stale address would silently stop balance tracking."

So after a `VaultUpdated`, components created before the rotation keep the old vault in `contract_addresses` forever while later ones get the new address. That set becomes `involved_contracts` in simulation (`vm/decoder.rs:80-85`) and filters which contracts' token balances reach the pool (`vm/decoder.rs:136-137`), so the two cohorts diverge permanently.

The fix is the one this comment already argues for: don't list the vault, let DCI discover it. The obstacle is the choice of traced entrypoint — `isActive` was picked because its calldata is fully known, but its access list doesn't reach the vault, so DCI never snapshots it. Adding an entrypoint that does touch the vault (a `quote` with a fixed probe amount, say) puts it back under DCI's retrigger machinery, where a rotation is picked up without anything being frozen at creation.

protocols/substreams/ethereum-tempest/src/modules.rs
```diff
@@ -0,0 +1,827 @@
...
+            .as_swap_type("tempest_pair", ImplementationType::Vm);
+
+        // No `paused` attribute at creation: `addPair` is itself the activation, and an absent
+        // attribute already means "not paused".
+        changes.push(TransactionEntityChanges {
+            tx: Some(tycho_tx),
+            entity_changes: vec![EntityChanges {
+                component_id: id,
+                attributes: vec![Attribute {
+                    name: BALANCE_OWNER_ATTRIBUTE.to_string(),
```
die-herdplatte, line 174: `balance_owner` is written once, here, with `ChangeType::Creation`, and never again — this is the only write site in the package.

Unlike the contract set, attributes *are* mutable, so this half of the vault-rotation problem is cheaply fixable: on `VaultUpdated`, fan a `balance_owner` update over `known_component_ids` the way the `Paused`/`Unpaused` handlers already do.

Without it, after a rotation `get_balance_overwrites` writes every existing component's inventory to the old vault address. That failure is silent and total — the venue just stops quoting. Worth doing even if the contract-set question above gets deferred.

protocols/substreams/ethereum-tempest/src/modules.rs
```diff
@@ -0,0 +1,827 @@
...
+                store.set(log.ordinal, PAUSED_KEY, &"0".to_string());
+            }
+        }
+    }
+}
+
+/// The `TempestVault` address recorded by [`store_router_state`], if the package has seen it.
+fn vault_address(router_state_store: &StoreGetString) -> Option<Vec<u8>> {
+    let Some(raw) = router_state_store.get_last(VAULT_KEY) else {
+        substreams::log::info!("no VaultUpdated seen yet; vault-dependent output is skipped");
```
die-herdplatte: There is exactly one `VaultUpdated` on chain — block 25,637,319, `address(0)` -> `0xc9d748e6...`, the initialisation — and `initialBlock` is that same block, so a full sync is fine and the doc comment above is accurate.

But `vault_address()` returns `None` whenever that block isn't in the indexed range: any partial re-index, any shortened test range, any start after init. `map_protocol_components` then returns early at L56 and `map_vault_balance_deltas` at L291, so there are no components and no balances — and the venue is indistinguishable from one that never traded.

This log line is the only diagnostic, and it reaches nobody. The indexer requests `production_mode: true` (`crates/tycho-indexer/src/substreams/stream.rs:131`), which makes the server strip module logs; nothing reads `debug_info` even when present; and `substreams::log` is a no-op in native builds, so it prints nothing under `cargo test` either.

Cheap fix, and the value is already being passed in the params: keep `vault_address` in `Config` and resolve it as `store.get_last(VAULT_KEY).or(config.vault_address)`. Event-sourced rotation is preserved and late starts become correct.

protocols/substreams/ethereum-tempest/substreams.yaml
```diff
@@ -0,0 +1,133 @@
...
+
+binaries:
+  default:
+    type: wasm/rust-v1
+    file: ../target/wasm32-unknown-unknown/release/ethereum_tempest.wasm
+
+network: mainnet
+
+params:
+  map_protocol_components: &tempest_params "router_address=00000003f1ec2379e79f58e12ec6c4f51ee92149&vault_address=c9d748e601d9984a43da0b80e5b91dc28d31d9fb&registry_address=DA7AFeEd01fe625cF15D187A19F94B45F00b8C5f"
```
die-herdplatte, line 27: `vault_address` is still passed here but was dropped from `Config` (`src/utils.rs:18-25`), and `serde_qs` ignores unknown keys — so it is now silently inert.

Two problems: it reads as live configuration to whoever maintains this next, and it is precisely the fallback that would fix the start-block dependence I flagged in `modules.rs`. Either remove it from the params string or restore the field and use it as the fallback.

protocols/adapter-integration/evm/src/tempest/manifest.yaml
```diff
@@ -0,0 +1,26 @@
...
+# Protocol Constants
+constants:
+  # The expected average gas cost of a swap
+  protocol_gas: 130000
+  # Minimum capabilities we can expect, individual pools may extend these
+  # No ConstantPrice: the lane is a VWAP spread ladder, so larger sizes cross into wider bands.
+  capabilities:
+    - SellSide
+    - BuySide
+    - PriceFunction
```
die-herdplatte: `getCapabilities` no longer returns `PriceFunction` (`TempestAdapter.sol:183-185`, and `price()` now reverts `NotImplemented`), so the manifest and the contract disagree.

Nothing reads the manifest's `capabilities` block today, so this is documentation rather than behaviour — but it's exactly the manifest/contract drift we keep meaning to add an automated check for, so worth keeping honest.

protocols/substreams/ethereum-tempest/src/modules.rs
```diff
@@ -0,0 +1,827 @@
...
+        }
+        for entity_change in &tx_changes.entity_changes {
+            builder.add_entity_change(entity_change);
+        }
+    }
+
+    for trx in block.transactions() {
+        // Most transactions in a block touch neither the router nor the registry. Skipping them
+        // here avoids allocating a builder per transaction just to drop it again at `build()`.
+        let touches_router = trx
```
die-herdplatte: `logs_with_calls()` is the wrong tool for a boolean: it allocates a `Vec` of every log in the transaction and sorts it by ordinal (`substreams-ethereum-core/src/block_view.rs:129-153`) before you look at the first element. So this gate — whose stated purpose is to avoid a per-transaction allocation — performs one of its own, on every transaction in every block.

Collapsing both arms into the call iteration is allocation-free and equivalent:

```rust
let touches_tempest = trx.calls.iter().any(|call| {
    !call.state_reverted &&
        (call.address == config.router_address || call.address == config.registry_address)
});
if !touches_tempest {
    continue;
}
```

Equivalent for the router because the gate is an `any()` over *all* frames, not just the log-emitting one. The router is an ERC1967 proxy, so a `PairRegistered` log is emitted from the delegatecall frame, whose `call.address` is the *implementation* — but the entry frame that called the proxy still has `address == router`, so it matches. (That `address`-is-the-code semantics for delegate frames is exactly what `extract_contract_changes_generic` compensates for at `tycho-substreams/src/contract.rs:111-118`, falling back to `call.caller` for `Delegate`/`Callcode` — that's the authority for the claim.)

Two ways it is more permissive, both harmless: transactions that call the router without emitting an event (a `quote` staticcall) and reverted router calls now pass the gate, then get dropped a moment later when `filter_map(build())` returns `None` for the empty builder. Keeping `!call.state_reverted` costs nothing and preserves most of the tightness.

die-herdplatte (changes requested):
Thanks!
Could you please stop rebasing for future commits such that we can see the changes clearly?

protocols/substreams/ethereum-tempest/src/modules.rs
JonathanChow2000: Done in `b131383c`. `VaultUpdated` now fans a `balance_owner` update over `known_component_ids`, the way the `Paused`/`Unpaused` handlers already do.

protocols/substreams/ethereum-tempest/src/modules.rs
JonathanChow2000: Fixed in `b131383c`, as suggested: `vault_address` is back in `Config` and resolution is store-first with the config value as fallback, so event-sourced rotation is preserved and late starts stop being a silent hole.

protocols/substreams/ethereum-tempest/substreams.yaml
JonathanChow2000: Confirmed — it was still in the params string while gone from `Config`, so `serde_qs` was quietly dropping it.

Took the second of your two options in `b131383c`: restored the field and wired it up as the fallback, so it's live configuration again rather than decoration.

protocols/adapter-integration/evm/src/tempest/manifest.yaml
JonathanChow2000: Fixed in `b131383c`.

I also left a note beside the `capabilities` block pointing at `getCapabilities`, so whoever changes one next has a reason to look at the other — until that automated check exists.

protocols/substreams/ethereum-tempest/src/modules.rs
JonathanChow2000: Fixed in `b131383c`, using your version essentially verbatim, `!call.state_reverted` included.

die-herdplatte (approved):

protocols/substreams/ethereum-tempest/src/modules.rs
```diff
@@ -0,0 +1,847 @@
...
+                        .upsert_token_balance(&balance_change.token, &balance_change.balance);
+                    builder.add_balance_change(balance_change);
+                }
+            }
+            builder.add_contract_changes(&contract_change);
+        });
+
+    extract_contract_changes_builder(
+        &block,
+        |addr| addr == config.router_address || addr == vault || addr == config.registry_address,
```
zach030: The vault's been removed from the contract set, so we don't need to track vault contract changes anymore.

protocols/substreams/ethereum-tempest/src/modules.rs
```diff
@@ -0,0 +1,847 @@
...
+                BigInt::zero()
+            });
+        balance_deltas.push(BalanceDelta {
+            ord: tx.index,
+            tx: Some(tx.clone()),
+            token,
+            delta: balance.to_signed_bytes_be(),
+            component_id: vec![],
+        });
+    }
```
zach030: I'm a bit concerned here. We only consider new_tokens as a trigger for the balanceOf call, but what happens if the `vault` gets updated? We'd need to refresh balances for every token then.

zach030 (commented):
LGTM! Just flagged my concern about vault updates, we don't really handle that logic in substreams yet.

protocols/substreams/ethereum-tempest/src/modules.rs
```diff
@@ -817,6 +854,24 @@ mod tests {
 
     /// An existing component re-registered in a later block has no pending creation, so the very
     /// first registration it sees must lift the pause.
+    /// A rotation must be detected from the store delta so that every tracked token is
+    /// re-snapshotted against the new vault. The initialisation write is not a rotation.
+    #[test]
+    fn test_vault_rotation_detected_from_deltas() {
```
die-herdplatte, line 860: This just tests a closure only defined in the test itself.
Also, the two lines of comments before this insertion should be moved to the test after this one.

protocols/adapter-integration/evm/src/tempest/TempestAdapter.sol
```diff
@@ -0,0 +1,292 @@
...
+/// @title TempestAdapter
+/// @notice Adapter for swapping tokens on Tempest, Flowdesk's propAMM.
+/// @dev Tempest prices from a quote "lane" the maker commits to the shared
+/// `PrioUpdateRegistry`. The router only reads a lane whose committed
+/// `updateTimestamp` sits inside `laneWindow` of `block.timestamp`, and reverts
+/// `StaleUpdate` otherwise; simulation pins the timestamp through the
+/// `override_block_timestamp` attribute the substreams package emits.
+///
+/// The adapter is quote-only: it never calls a settlement entrypoint. Every
+/// `swap*` function on the router is gated on `allowedTaker[msg.sender]`, and
```
die-herdplatte, line 16: Is this still true? We blocked Tempest for the same reason on our generic Titan price level stream because of that but the implementation contract has since been replaced and the restriction lifted.

This would allow us to actually execute `swap` here and avoid the constant `SETTLEMENT_GAS`.

protocols/substreams/ethereum-tempest/src/modules.rs
die-herdplatte, line 363: The rotation re-snapshot double-counts. `map_vault_balance_deltas` pushes an **absolute** balance
as a delta (line 363), and `store_vault_token_balances` is `updatePolicy: add` (line 412). That was
safe while snapshots only ever happened for a token seen for the first time — store empty,
`0 + balance == balance`. On a rotation the store already holds ≈ `balanceOf(old_vault)` for every
tracked token, so it becomes `old + new`, and `map_balance_deltas` fans that same absolute value out
to every existing component, so `store_balances` inflates identically. The inventory doubles
instead of repointing, and nothing fails loudly — quotes just price against a reserve roughly twice
the real one. It is latent today (the only `VaultUpdated` so far is the `address(0)` initialisation
at block 25,637,319, which this PR correctly excludes), but rotation handling is the purpose of the
commit.

Rather than computing `new − old` at the call site, I would make the two kinds of change explicit
and let the store apply the right operation — which is already the pattern in `ethereum-ekubo-v2`:

```proto
// proto/tempest.proto
enum ChangeType { DELTA = 0; ABSOLUTE = 1; }
```

```rust
// src/store.rs  (cf. ekubo-v2/src/store.rs)
ChangeType::Delta    => StoreSetSum::sum,
ChangeType::Absolute => StoreSetSum::set,
```

with `store_vault_token_balances: StoreSetSumBigInt` and `updatePolicy: set_sum`. `StoreSetSum` is
available in the substreams version we pin. Snapshots then `set`, transfer deltas `sum`, and the
rotation stops being a special case: because writes apply in ordinal order and a `set` replaces
rather than adds, the double-count guard at line 384 is no longer required for correctness. It is
still worth keeping if we want clean per-transaction balances inside a rotation block — without it
the pre-`set` sums are recorded as transient values on top of the stale base, and only the final
`set` is right.

Only this package's own code writes and reads that store, so the change is contained; the
component-level `store_balances` (line 498) goes through `tycho-substreams`' shared balance
helpers, so the same shape there would mean set_sum-aware siblings in that crate. Which points
at a simplification worth considering instead: drop the per-`(component, token)` accumulation in
`store_balances` entirely. `map_balance_deltas` only copies one vault-wide token balance onto every
component that trades that token, so `store_balances` holds nothing that
`store_vault_token_balances` does not. `map_protocol_changes` can emit each `BalanceChange` —
absolute by definition in the tycho API — straight from `store_vault_token_balances`, taking it in
`deltas` mode to learn which tokens changed this block and in `get` mode for the values, including
the initial pair of a component created in a block where neither of its tokens moved. Only the
delta *values* carry the `set:`/`sum:` prefix (which is why ekubo has
`bigint_from_set_sum_store_delta_value`); `get` returns the resolved number, because the runtime
strips the prefix for `set_sum` stores — so taking keys from the deltas and values from `get` needs
no decoding. That removes `map_balance_deltas`, `store_balances`, the new-component seeding branch
and the double-count guard together.

One detail to carry into either version: **the snapshot ordinal has to change.** A `set` only
supersedes the `sum`s it is meant to replace if it sorts after them, and the snapshot currently
carries `ord = tx.index` — a transaction index, single digits — while the transfer deltas carry
`log.ordinal` (`extract_balance_deltas_from_tx`), which is block-scoped and far larger. Harmless
under `add`, but under `set_sum` the `set` would sort first and the sums would overwrite it.
`end_ordinal` of the block's last `TransactionTrace` is the value to use:
`begin_ordinal`/`end_ordinal` bracket every element ordinal of a transaction, so the last one's end
ordinal is the block maximum, it stays consistent with the `tx` the delta is attributed to, and
`store_balance_changes`' strictly-increasing-per-key rule still holds. Note the module currently
keeps only `Transaction::from(trace)`, which drops the ordinals — it needs the trace itself to read
`end_ordinal`. (`balanceOf` reads end-of-block state, which is what makes `set` correct here.)

die-herdplatte (changes requested):
Hey, I just noticed that the Tempest executor uses pull-payment (`swapWithAllowances`) instead of push-payment (`swap`).
This costs more gas when a Tempest leg is the first leg of a route since the funds would move `user -> Tycho router -> venue` instead of just `user -> venue`.

At this point, we could even reuse the `PropAMMExecutor` since it already calls `IPropAMM.swap` (push-payment).
What do you think?

protocols/substreams/ethereum-tempest/integration_test.tycho.yaml
```diff
@@ -0,0 +1,111 @@
...
+        skip_execution: true
+
+  - name: test_pair_simulation
+    start_block: 25637319
+    # The router only prices a lane whose committed `updateTimestamp` is inside `laneWindow`
+    # (currently 12s either side of the settling block). The stop block is the first `updateState`
+    # for the USDC/WETH lane, whose `updateTimestamp` equals this block's timestamp exactly — so
+    # the `override_block_timestamp` the package emits lands the simulation dead centre in the
+    # window. Stopping earlier reverts with `StaleUpdate`.
+    stop_block: 25673321
```
zach030, line 72: At this block, Tempest is still restricting `allowedTaker`, the test fails with `TakerNotAllowed()`.

crates/tycho-execution/src/encoding/evm/swap_encoder/swap_encoder_registry.rs
zach030, line 1: Tempest uses push payments, but it isn’t included in the optimizable transfer protocol list. [gas_estimator.rs](https://github.com/propeller-heads/tycho/blob/9dadcf3c1e9453b3d199b364394f076007949944/crates/tycho-execution/src/encoding/evm/gas_estimator.rs#L34-L51)

protocols/substreams/ethereum-tempest/integration_test.tycho.yaml
```diff
@@ -0,0 +1,111 @@
...
+    # The router and vault proxies are deployed at the start block; the three `addPair` calls land
+    # at 25637575 (WETH/USDT), 25637589 (USDC/WETH) and 25637601 (USDC/USDT).
+    stop_block: 25637601
+    expected_components:
+      - id: "0x2de07e5537a6bd50791ca38fd3dda74a0f7b8f196ecdb2f7f1e48ef06bfa4930"
+        tokens:
+          - "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2" # WETH
+          - "0xdAC17F958D2ee523a2206206994597C13D831ec7" # USDT
+        static_attributes:
+          self_contained_tokens: "0x5b22307863303261616133396232323366653864306130653563346632376561643930383363373536636332222c22307864616331376639353864326565353233613232303632303639393435393763313364383331656337225d" # ["0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2","0xdac17f958d2ee523a2206206994597c13d831ec7"]
```
zach030, line 35: The `pamm_address` attribute is missing here.

PR #1321, opened 2026-08-13
feat(sky): add Sky (ex-MakerDAO) Ethereum integration

die-herdplatte:
Adds the Sky stablecoin mint/redeem venues on Ethereum mainnet as protocol system `sky`, covering indexing, simulation, and execution.

### Components

| component_type | Contract | Pair | Address |
|---|---|---|---|
| `psm` | DssLitePsm (`MCD_LITE_PSM_USDC_A`) | DAI↔USDC | `0xf6e72Db5454dd049d0788e411b06CfAF16853042` |
| `psm_wrapper` | UsdsPsmWrapper | USDS↔USDC | `0xA188EEC8F81263234dA3622A406892F3D630f98c` |
| `converter` | DaiUsds | DAI↔USDS | `0x3225737a9Bbb6473CB4a45b7244ACa2BeFdB276A` |

All swaps are 1:1 (with 18↔6 decimal rescaling on USDC legs) subject to the PSM's governance-mutable `tin`/`tout` fees, both zero since deployment. The MKR→SKY converter is out of scope (fee ratchets +1pp per quarter, path being wound down by governance). sUSDS is already covered by the `erc4626` protocol; sDAI is configured there but excluded by `erc4626_filter` — its `maxDeposit` returns `type(uint256).max`, which `ERC4626State` takes at face value as the deposit limit, so deposits would quote unbounded, and the real limit is not exposed on-chain in any traceable form. This PR documents that rationale at the filter, which previously carried no comment.

### Indexing (`protocols/substreams/ethereum-sky`)

- Three hardcoded components (`ImplementationType::Custom`, singleton pattern, no factory). Params carry only contract addresses and creation anchors.
- One seeding rule for all state: at a component's creation block, balances, fees, and escrows are seeded from deterministic post-block `eth_call` snapshots; event-driven updates count strictly after that block.
- PSM balances (DAI buffer + pocket USDC) from ERC20 Transfer logs; the stateless wrapper mirrors the PSM's inventory (DAI side relabelled as USDS) so both entry points survive min-TVL filtering.
- The converter's balances and the wrapper's `dai_escrow`/`usds_escrow` attributes are the joins' internal dai escrows (`vat.dai[join]`), read as absolute values from Vat storage writes to the escrow slots. This bypasses delta aggregation and captures every Vat channel that can touch them (`join`/`exit` as well as `move`/`frob`/`suck`/`fold` donations). `tin`/`tout` update from LitePSM `File` events, mirrored onto the wrapper.

### Simulation (`tycho-simulation`)

Native `SkyState` for all three component kinds: PSM sell/buy math with fee handling and `HALTED` sentinels, converter burn bounded by the sell-side escrow, and the wrapper additionally bounded by both join escrows its in-flight DAI↔USDS conversion burns through (sellGem burns its DAI payout via DaiJoin, buyGem burns its full USDS input via UsdsJoin). `spot_price` follows the trait's buy-price convention; `query_pool_swap` is closed-form since prices are size-independent (the generic root-search assumes a falling price curve and would return an arbitrary interior amount on a flat one).

### Execution (`tycho-execution`)

`SkyExecutor` holds the three venues as constructor immutables and fetches the token wiring in the constructor (reverting on mismatch). Calldata is 2 bytes: target index + direction flag, matching the encoder's `gem` attribute-derived direction. `buyGem` floors the gem amount, so at most 1e12-1 wei of input dust stays at the router (tested). 19 Foundry fork tests including router integration.

### Testing

- Substreams unit tests pin the escrow slot derivation and `File` topic against mainnet.
- 34 `SkyState` unit tests; decoder tests for strict attribute/balance requirements.
- Protocol-testing harness: 3/3 range tests green (indexing + simulation; simulation/execution skipped where the range windows make limits legitimately zero or the venue is provably non-operational pre-launch — reasons documented in the yaml). The harness runs execution against a runtime-bytecode fixture, so no deployed executor is required.
- **Full test run (creation block → chain tip): passed.** Indexed blocks 20,283,666 → chain tip, then ran the live phase at the chain head. Across blocks 25,782,603–25,788,830 the stream processor sampled 2,650 block updates (it skips blocks while executing a batch and sleeping the mandated block time); 871 carried sky state changes, and 852 of those became simulation+execution batches (the other 19: 18 head-lag `Block not found` races where execution was skipped after successful simulation, 1 pre-existing native_wrapper harness error) — **~9,900 swaps executed, 0 failures** — covering all three components in both directions at three trade sizes each (PSM DAI↔USDC, wrapper USDS↔USDC with the 18↔6 rescaling, converter DAI↔USDS). Every executed amount matched the simulated quote **exactly, to the wei** (the harness tolerates 0.5% slippage; measured deviation was zero).
- The full run also surfaced a harness bug (not a Sky issue): the live phase's token-quality filter silently dropped all components on a local DB — fixed separately in #1338; this run was unblocked with the equivalent DB workaround.

### Not included (follow-ups)

- Production executor deployment (needs the three constructor args from `executor_deployments.json`), spkg release, and `extractors.yaml` entry.
- stUSDS via the `erc4626` protocol (requires a real `maxRedeem` read; it is deposit-capped with utilization-gated withdrawals).


protocols/substreams/ethereum-uniswap-v4/Cargo.lock
zach030, line 1: Let's split the uniswap-v4 and ramses-v3 upgrades into a separate PR?

protocols/substreams/ethereum-sky/src/modules/3_map_relative_balances.rs
```diff
@@ -0,0 +1,326 @@
...
+    }
+    if block.number == config.wrapper_creation_block {
+        balance_deltas.extend(seed_deltas(
+            &block,
+            &config.wrapper_creation_tx,
+            config.wrapper_component_id(),
+            [
+                (&config.usds, balance_of(&config.dai, &config.psm)),
+                (&config.usdc, balance_of(&config.usdc, &config.pocket)),
+            ],
```
zach030, line 63: Could you add some comments here to clarify why the DAI balance represents the USDS balance? Is it because they convert 1:1 in the converter?

crates/tycho-simulation/src/evm/protocol/sky/state.rs
```diff
@@ -0,0 +1,954 @@
...
+                // i.e. the opposite trade with the opposite fee.
+                let execution = if self.is_gem_to_stable(&token_in.address, &token_out.address)? {
+                    1.0 - Self::fee_f64(self.tin)
+                } else {
+                    1.0 / (1.0 + Self::fee_f64(self.tout))
+                };
+                let limit = price_f64(limit, token_in.decimals, token_out.decimals);
+                if execution < limit && !within_tolerance(execution, limit, *tolerance) {
+                    return Ok(zero_swap());
+                }
```
zach030: `within_tolerance` is symmetric, this basically says "only reject if execution is below the limit AND far off", so anything just below the limit still slips through. But the trait docs say tolerance shouldn't let you violate the limit, and the generic helper enforces a hard floor (actual >= target && ...).

crates/tycho-simulation/src/evm/protocol/sky/state.rs
```diff
@@ -0,0 +1,954 @@
...
+            if self.kind == SkyComponentKind::Converter {
+                return Ok((u256_to_biguint(self.gem_balance), u256_to_biguint(self.gem_balance)));
+            }
+            // The wrapper's stable payout is additionally bounded by the DAI join
+            // escrow its in-flight conversion burns through.
+            let max_out = match &self.escrows {
+                Some(escrows) => self.stable_balance.min(escrows.dai),
+                None => self.stable_balance,
+            };
+            let max_in = max_out * wad / (self.conversion_factor() * (wad - self.tin));
```
zach030: There are quite a few calculations like this, I think we can replace them with [safe_math](https://github.com/propeller-heads/tycho/blob/main/crates/tycho-simulation/src/evm/protocol/safe_math.rs).

zach030 (commented):
Looks good overall! Left some comments, please have a look!

protocols/substreams/ethereum-uniswap-v4/Cargo.lock
die-herdplatte: Removed from the branch entirely (reverted to main's versions in 636ddb2d7). They weren't intentional upgrades — just artifacts our substreams build regenerated as a side effect — so nothing to carry into a separate PR.

protocols/substreams/ethereum-sky/src/modules/3_map_relative_balances.rs
die-herdplatte: Added in 636ddb2d7. Exactly — the wrapper is stateless and its liquidity IS the PSM's: it converts DAI↔USDS 1:1 in-flight via the joins (sellGem's DAI payout → USDS, buyGem's USDS input → DAI), so the PSM's DAI balance is precisely the USDS the wrapper can pay out. The comment now says so at the seeding site.

crates/tycho-simulation/src/evm/protocol/sky/state.rs
die-herdplatte: Good catch — fixed in 636ddb2d7. The limit is now a hard floor: `execution < limit` returns the zero swap and tolerance can't rescue it. The target-price arm now uses the generic `is_within_tolerance` (made `pub(crate)` and shared instead of the local symmetric near-duplicate), and a new test case pins that a generous tolerance no longer admits below-limit executions.

crates/tycho-simulation/src/evm/protocol/sky/state.rs
die-herdplatte: Done in 636ddb2d7 — `stable_out`, `gem_out`, and both `get_limits` chains now use safe_math wherever operands are state-dependent (`tin`/`tout`/amounts), so absurd inputs error instead of wrapping, and `tin == WAD` errors instead of panicking on division by zero. Plain division remains only where the divisor is provably non-zero (constant WAD, or a checked power-of-ten product).

zach030 (approved):

PR #1323, opened 2026-08-13
fix: decouple partition retention from partition creation

zizou0x:
## Problem

pg_partman's `run_maintenance()` creates upcoming partitions and drops expired ones in one transaction per parent table. When it reaches the retention `DROP TABLE` it already holds locks on the default partition, and the drop needs an `AccessExclusiveLock` on the parent. A concurrent query that took the parent lock in between and then waits on the default partition closes a lock cycle; Postgres kills the maintenance run, rolling back the created partitions along with the failed drop.

On dev-bsc this lost the race every midnight from Jul 10 to Aug 12 (verified in RDS logs; the partner was the hourly price export, then the indexer's own queries after propeller-heads/helm-configuration#905 moved the export off the hour). Partition premake stayed frozen until the premade partitions ran out on Aug 9, at which point closed row versions could no longer leave the default partition and inserts crashed the indexer on `component_balance_default_unique_pk`.

## Change

New migration `2026-08-13_decouple_partition_retention`:

- Sets `retention = NULL` in `part_config` for `component_balance`, `contract_storage` and `protocol_state`. The nightly `run_maintenance_proc()` becomes creation-only and can no longer be rolled back by a retention failure.
- Adds a single-row `partition_retention_config` table (default `'1 month'`) as the one place the retention horizon is defined, with a `partition_retention()` reader function. `drop_expired_partitions()` resolves its retention from it on every run, and the `cleanup_orphaned_transactions` cron job is re-pointed to `p_min_age => partition_retention() + interval '4 days'` (its `part_config`-based default would otherwise pin the horizon at the 31+4-day fallback). Changing retention is an `UPDATE` of that row.
- Adds `drop_expired_partitions(p_retention default NULL, p_lock_timeout default '2s', p_max_attempts default 5)`: calls `partman.drop_partition_time()` per parent table, each attempt in its own transaction. A fresh transaction holds nothing when it requests the parent lock, so it can queue but not deadlock; `lock_timeout` bounds how long other queries can stall behind the queued request, with 30s-backoff retries. Retries fire on any error because partman re-raises internal errors with a generic SQLSTATE, so lock timeouts cannot be told apart from other failures; the call is idempotent and a persistent error still surfaces. If attempts run out, the run raises after committing its completed work, so the failure shows in `cron.job_run_details` and the next run picks up the remainder.
- Schedules it at 00:30, after the midnight creation run and clear of the 02:00 transaction cleanup.

`down.sql` restores the `'1 month'` `part_config` retention, restores the cleanup job's original command, and removes the job, procedure, function and config table.

## Testing

Against a from-migrations database (PG16, pg_partman with the same `drop_partition_time` signature as prod):
- planted expired partitions are dropped; `run_maintenance()` and `run_maintenance_proc()` succeed with `retention = NULL`
- under a held conflicting lock: each attempt fails with a caught lock timeout and is retried, the run fails visibly after max attempts, and the next run drops the survivor
- `diesel migration revert` + re-run both apply cleanly; both rescheduled cron commands execute manually
- new `test_drop_expired_partitions_serial_db` plants an expired partition, drops it through the procedure (retention resolved from the config table), and re-runs for no-work idempotency; verified it fails on a typo in the nested partman call
- `cargo test -p tycho-storage --lib`: 108 passed

Deploying this only fixes new databases' schedules going forward; existing environments get it at next migration run. Alerting on pg_cron failures / partition runway is a separate follow-up in the monitoring stack.


crates/tycho-storage/migrations/2026-08-13_decouple_partition_retention/up.sql
```diff
@@ -0,0 +1,101 @@
...
+
+UPDATE partman.part_config
+SET retention = NULL
+WHERE parent_table IN ('public.component_balance', 'public.contract_storage', 'public.protocol_state');
+
+-- Retention now lives in p_retention below instead of part_config (part_config.retention
+-- would put the drop back inside run_maintenance). To change it, alter this default; to
+-- change which tables are swept, update the array.
+CREATE OR REPLACE PROCEDURE drop_expired_partitions(
+    p_retention interval DEFAULT '1 month',
```
tvinagre: **Non-blocking:** Moving retention out of `part_config` silently decouples `cleanup_orphaned_transactions` from the actual retention horizon. That procedure still computes `coalesce(max(retention::interval), interval '31 days') + interval '4 days'`; after these rows are cleared it always uses 35 days. Today that roughly matches `1 month + 4 days`, and the `NOT EXISTS` gates keep it safe, but changing this default (as the comment suggests) will not propagate and can make cleanup scan a much larger band of non-orphan rows. Could we either document that coupling here or use a shared retention setting?

crates/tycho-storage/migrations/2026-08-13_decouple_partition_retention/up.sql
```diff
@@ -0,0 +1,101 @@
...
+        v_done := false;
+        v_attempt := 0;
+        WHILE NOT v_done AND v_attempt < p_max_attempts LOOP
+            v_attempt := v_attempt + 1;
+            BEGIN
+                -- Applies to the current transaction only. If the parent lock is not granted
+                -- within the timeout, give up and retry instead of stalling the queries that
+                -- queue behind the AccessExclusiveLock request.
+                PERFORM set_config('lock_timeout', p_lock_timeout, true);
+                PERFORM partman.drop_partition_time(
```
tvinagre, line 84: **Non-blocking:** Can we add a serial DB test that actually calls this procedure? `CREATE PROCEDURE` does not resolve the nested `partman.drop_partition_time(...)` call, so migration-only coverage will not catch a typo or future pg_partman signature change; the first failure would be the 00:30 cron job. The existing `test_cleanup_orphaned_transactions_serial_db` is a suitable template: create an expired partition, `CALL drop_expired_partitions()`, assert removal, then call it again to cover no-work idempotency.

tvinagre (commented):
Two non-blocking suggestions from the review.

tvinagre (approved):

PR #1324, opened 2026-08-13
feat(integration-test): bypass the executor activation timelock in execution simulations

die-herdplatte:
Lets the live integration test validate an executor that is deployed but not yet usable, instead of staying red for 3 days after every executor deployment. [ENG-6070](https://propeller-heads.atlassian.net/browse/ENG-6070).

`Dispatcher._validateExecutor` reverts with `Dispatcher__UnapprovedExecutor` when `executorsActivationTimestamp[executor] == 0` and with `Dispatcher__ExecutorIsTimelocked` while it is in the future, and `_setExecutor` always sets `block.timestamp + DELAY_EXECUTOR_ACTIVATION` (3 days). The offline `protocol-testing` harness already sidesteps this by planting `1` into the activation slot of its fake executor; the live `tycho-integration-test` passed no router overwrites at all, so every swap through a fresh executor reverted before reaching the protocol.

## What

- **`--bypass-executor-timelock`** on `tycho-integration-test` (default **off**): writes `executorsActivationTimestamp = 1` for every executor of the chain into the simulation's state overrides. The router keeps its deployed bytecode, only the `debug_traceCall` is affected, and no real transaction path changes. A `warn!` at start-up states that a passing run under the flag does not prove the executors are activated.
- **`RouterOverwritesData` generalized** (`crates/tycho-test/src/execution/models.rs`). Every field is now optional and the single executor bytecode became a map:
  ```rust
  pub struct RouterOverwritesData {
      pub router_bytecode: Option<Vec<u8>>,
      pub executors: HashMap<Address, Option<Vec<u8>>>,
      pub fee_calculator_bytecode: Option<Vec<u8>>,
  }
  ```
  Presence in `executors` activates that executor on the router; `Some(bytecode)` additionally plants code at its address. Values are `Option<Vec<u8>>` rather than `Vec<u8>` because the live path must activate executors while keeping their deployed code — `with_code(vec![])` would blank it.
- **`setup_router_overwrites(router_address, data)`** takes the struct directly and emits only the overwrites the data asks for, so `RouterOverwritesData::default()` simulates the deployed contracts unchanged — what the live test previously expressed by passing `None`. `simulate_swap_transaction` therefore loses its `Option`, and the helper loses an `async` it never awaited.
- **`SwapEncoderRegistry::executor_addresses()`** returns the executor address per protocol system, derived from the encoders the registry already holds. The registry had no way to enumerate them; this avoids re-parsing `executor_addresses.json` and keeps one source of truth.
- Each consumer builds its own overwrites: `create_router_overwrites_data(chain)` in `tycho-integration-test/src/main.rs`, mirroring `protocols/testing`'s `create_router_overwrites_data(protocol_system)`. The offline path is unchanged in behaviour — it still plants router, executor and fee calculator bytecode at the same addresses.

## Why it is off by default

The live test doubles as a canary for "an executor was never activated in production". With the bypass always on, that signal would disappear. Off by default keeps the revert visible; pass the flag when validating an executor inside its timelock window.

## Which slot, and how it was confirmed

`executorsActivationTimestamp` is a `mapping(address => uint256)` at base slot 6 of `TychoRouterV3`, so the value lives at `keccak256(abi.encodePacked(bytes32(executor), bytes32(uint256(6))))` — exactly what the existing `calculate_executor_storage_slot` computes. Confirmed against the **deployed** router (`0xea290cE3Eae57BdB37E57872a5a14DC0d2f6e614`), not just the local contracts: `cast storage <router> $(cast index address <executor> 6)` returns the same value as the `executorsActivationTimestamp(address)` getter.

## Testing

Unit:

- `cargo +nightly fmt --all` clean, `cargo clippy --workspace --all-targets --all-features` clean.
- New tests in `crates/tycho-test/src/execution/encoding.rs`: activation-only overwrites touch the router account alone and set no `code`; the full offline shape plants router/executor/fee-calculator bytecode, activates the executor and points the `_feeCalculator` slot at the overwritten address; `RouterOverwritesData::default()` produces an empty override map. New test in `swap_encoder_registry.rs`: `executor_addresses()` agrees with `get_encoder(protocol).executor_address()`.
- Both behavioural tests were checked against deliberate mutations (activation value → 0, override merge → clobber) and fail as expected.
- The 5 failures in `cargo test -p tycho-execution --lib` are the pre-existing `Missing RPC_URL in environment` permit2/approvals tests.

Live, on Ethereum against `tycho-beta`, using the **PropAMMExecutor** `0xc196f22d69d41A2B1DbF6ef8d28C09bc2CA4E4fa` as the subject — 803 bytes of code deployed, activation timestamp `1786872335` vs block timestamp `1786635407`, i.e. ~66 h of timelock remaining. Run on a temporary merge of this branch onto the pAMM branch that registers that executor, with `--disable-onchain --disable-rfq` so every simulation routes through it:

| | baseline (no flag) | `--bypass-executor-timelock` |
|---|---|---|
| `0x45d4ba96` = `Dispatcher__ExecutorIsTimelocked` | **1211** | **0** |
| `simulation_execution_success` | 0 | 10 (with slippage comparisons) |
| expected stale-oracle skips | 0 | 181 |

## Not covered

An executor that is not deployed at all still cannot be simulated: with no bytecode the call returns empty returndata and the `amount_out` decode fails. Lifting the activation check does not change that — testing a genuinely undeployed executor needs a bytecode override, i.e. the `protocol-testing`-style path.

[ENG-6070]: https://propeller-heads.atlassian.net/browse/ENG-6070?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/tycho-integration-test/src/main.rs
```diff
@@ -991,13 +1042,25 @@ async fn process_update(
...
-            Ok(results) => results,
-            Err((e, _, _)) => return Err(e),
-        };
+    // Everything is simulated against the deployed contracts; the executors are only activated
+    // when asked for, so that a missing activation still surfaces as a revert.
+    let router_overwrites_data = if cli.bypass_executor_timelock {
+        create_router_overwrites_data(chain)?
+    } else {
+        RouterOverwritesData::default()
+    };
```
zach030: `create_router_overwrites_data` function gets called on every update to rebuild the overwrite data from the json file. Could we turn it into a globe shared config, loaded just once at startup?

crates/tycho-test/src/execution/encoding.rs
```diff
@@ -630,74 +633,66 @@ fn overwrite_bopamm_lane_timestamp(stored_value: B256, timestamp: u32) -> B256 {
...
-    fee_calculator_bytecode: Vec<u8>,
+    data: RouterOverwritesData,
 ) -> miette::Result<AddressHashMap<AccountOverride>> {
     let mut state_overwrites = AddressHashMap::default();
+    if data.router_bytecode.is_none() &&
+        data.executors.is_empty() &&
+        data.fee_calculator_bytecode.is_none()
+    {
+        return Ok(state_overwrites);
+    }
```
zach030: Maybe add an is_empty method to `RouterOverwritesData`? That way we can drop the three-condition guard and make it easier to maintain.

zach030 (commented):
Thanks! Left some small comments, otherwise LGTM.

crates/tycho-integration-test/src/main.rs
die-herdplatte: Done. If you meant a struct that contains all the other fields that `process_update` receives: I left that out here, but it would be a good cleanup.

crates/tycho-test/src/execution/encoding.rs
die-herdplatte: Done

zach030 (approved):

PR #1328, opened 2026-08-14
feat(execution)!: require explicit native wrap swaps in solutions

tamaralipows:
[Task](https://propeller-heads.atlassian.net/browse/ENG-6255)

Supersedes #1279.

## Why remove instead of fix

Fixing the edge cases raised on #1279 made the insertion logic more and more complex until a red flag went off. The encoding-side insertion predates wrap/unwrap injection into the Tycho stream — now that a `native_wrapper` component is auto-injected, there is no reason to expect the encoder to fill ETH/WETH gaps.

The failing solutions were already complete, like this valid split:

```
        ┌──[70%]── WETH ──┐
USDT ───┤                 ├── USDC
        └──[rem]── ETH ───┘
```

In the flattened swap list the two branches sit next to each other, and the adjacency check injected a spurious 0% ETH→WETH wrap between them — failing split validation, or silently re-routing the payout when the wrapped token was the output.

## What the encoder does now

Solutions encode as given; wrap swaps are regular swap steps on the `native_wrapper` component. A solution with an ETH/WETH gap previously encoded without error as a single swap and reverted on-chain — the single-swap strategy now runs the same swap-path validation as the sequential and split strategies. Docs and migration guide updated. Breaking for any solver that relied on auto-insertion.


crates/tycho-execution/src/encoding/evm/tycho_encoders.rs
```diff
@@ -432,154 +368,160 @@ mod tests {
...
-            assert_eq!(solution.swaps().len(), 2);
-            assert_eq!(last_swap.token_in().address, eth());
-            assert_eq!(last_swap.token_out().address, weth());
-            assert_eq!(last_swap.component().protocol_system, "native_wrapper");
+            let encoded_solution = encoder
+                .encode_solution(&solution)
+                .unwrap();
+            assert!(encoded_solution
+                .function_signature()
+                .contains("splitSwap"));
```
Troshchk, line 488: Is it enough -- other asserts?

crates/tycho-execution/src/encoding/evm/tycho_encoders.rs
```diff
@@ -432,154 +368,160 @@ mod tests {
...
+        //
+        //       ┌──[70%]── WETH ──┐
+        // DAI ──┤                 ├── USDC
+        //       └──[rem]── ETH ───┘
         #[test]
-        fn test_add_missing_wrapped_eth_swap_in_the_end() {
-            // before adding swap: USDC -> ETH, WETH is the solution token_out
-            // after adding swap:  USDC -> ETH -> WETH
-
+        fn test_encode_split_solution_with_native_and_wrapped_branches() {
```
Troshchk, line 465: This is the only test which fails before the fix

crates/tycho-execution/src/encoding/evm/tycho_encoders.rs
Troshchk: Before this fix, this test fails with the current asserts. This assert is enough to test the fixed behaviour

docs/for-solvers/execution/encoding/README.md
```diff
@@ -296,7 +296,7 @@ that surplus beyond your quote reaches the receiver. Amounts between `minAmountO
 
 #### Native Tokens <a href="#native-tokens" id="native-tokens"></a>
 
-The encoder automatically bridges ETH↔WETH gaps anywhere in the swap path — at the start, end, or between swaps — using a dedicated WETH executor. Set `token_in` and `token_out` to the tokens the user actually holds and expects to receive, and the encoder inserts wrap/unwrap steps as needed. This works with protocols like Uniswap V4 that accept native ETH directly, with no extra configuration required.
+Wrapping and unwrapping are regular swap steps on the `native_wrapper` protocol, executed by a dedicated WETH executor. The Tycho stream injects a `native_wrapper` component on every chain, so routes cross between ETH and WETH through it like through any other pool — including with a split, to convert only part of a balance. The encoder does not insert these swaps for you: a solution that needs ETH or WETH where it only holds the counterpart, and carries no wrap swap, is rejected at validation.
```
tamaralipows: I think this doc can be clearer... it's putting a lot of focus on splits and using words like "counterpart". Not sure we need such detail, just say to make sure all swaps are connected as we expect...

Troshchk (approved):

PR #1329, opened 2026-08-15
fix: encode Bebop original taker amount

pedrobergamini:
## What

Fixes Bebop partial fills when the router executes less input than the signed quote.

- Encodes the signed quote's `amount_in` as the original filled taker amount. `BebopExecutor` compares this value with runtime `amountIn`, so using `amount_out` mixed input and output token units.
- Parses the provider's full big endian `u64` offset and rejects malformed values or values that do not fit the executor's `u8` field.

## Testing

- Adds focused encoder coverage with different estimated input, signed input, and output amounts.
- Adds a TychoRouterV3 regression that routes a 10 WETH runtime fill from a 20 WETH signed quote through Dispatcher and BebopExecutor.


die-herdplatte (approved):
Thank you very much!

zach030 (commented):
LGTM. Could you please resolve the conflicts?

PR #1330, opened 2026-08-17
fix: survive reverts targeting blocks not sealed in the reorg buffer

brunoguerios:
## Problem

The indexer process exited three times in one week, all on Base, killing every extractor on every
chain in the process and dropping all websocket subscriptions:

```
08-05 14:31:54 base-tycho-indexer-fynd | Error: Reorg buffer error: Could not find block with id `0x8d0fd3ee…`!
07-30 07:54:34 base-tycho-indexer-fynd | Error: Reorg buffer error: Could not find block with id `0x2e37c02d…`!
07-30 07:36:28 base-tycho-indexer-fynd | Error: Reorg buffer error: Could not find block with id `0x4f5eae77…`!
```

All three target hashes resolve to real, canonical, sealed Base blocks 5-6 blocks below the chain
head, against a `ReorgBuffer` holding ~225 blocks. This is not a reverts-deeper-than-the-buffer
problem.

## Root cause

`ReorgBuffer::purge` matched on hash only. Base runs `--enable-partial-blocks`, so a block enters
the `ReorgBuffer` only when its last flashblock partial arrives — the block currently streaming
lives only in the `PartialBlockBuffer`. During a revert burst, undo #1 purges sealed blocks,
Substreams re-streams the new fork as flashblocks, and undo #2 names the block still being
re-streamed. The buffer has no entry to match and `handle_revert` returns a fatal error. A second
shape fits the same evidence: the buffer holds the right height under a stale last-partial hash.

## Fix

`ReorgBuffer::purge_to` makes the extractor revert path height-aware. Hash stays authoritative:

- Hash match → purge strictly after it (unchanged).
- No hash match, target height buffered → our copy is stale; purge from that height inclusive.
- No hash match, target above the buffer → purge nothing.
- No hash match, target below the buffer's oldest block → still fatal, with the identical
  `NotFound` string alerting greps.

Pending partials are now dropped only above the target height; partials at the target height are the
valid prefix of the last valid block. An undo that invalidates nothing advances the cursor and
returns no message — that case used to die with "Reorg buffer is empty after purge".

Both fallbacks log a `warn!` and increment `extractor_revert_hash_miss` with a `resolution` label
(`height_purge` / `target_ahead`).

`purge` keeps strict hash-only semantics; `PendingDeltasBuffer` still uses it.

## Notes for reviewers

An inclusive purge at a hash-mismatched height relies on Substreams re-sending the canonical block
from `last_valid_cursor`. That semantic is unverified against the live endpoint. If it does not
hold, the next `insert_block` fails the parent-hash check and the process dies — the same outcome as
today's immediate crash, so the change is strictly no worse. The new metric confirms the semantics
in production.

DCI is untouched. Its `revert_to` is also hash-only, but DCI cannot run on partial-block chains
today (`validate_and_ensure_block_layer_internal` needs parent-hash linkage, which ephemeral partial
hashes break), so DCI extractors cannot reach the new fallback paths.

## Follow-ups

- Watch `extractor_revert_hash_miss` on Base to confirm the cursor re-send semantics.


tvinagre:
## Revert-buffer boundary follow-up

I agree the DCI concern is not blocking here: Flashblocks currently exists only on Base, and Base cannot run DCI. I'm fine with the decision, but please document this clearly in the code so we can pick-this up once we have DCI in Base in the future.

The remaining design question is at the **oldest buffered height**, and is distinct from the Flashblocks `TargetAhead` case.

`BlockUndoSignal.last_valid_block` is the keep-anchor: Substreams specifies that data *above* it must be deleted and processing restarts from `last_valid_cursor`.

Example:

```text
DB/finalized: … 99
ReorgBuffer:  100 (stale hash A), 101 (B), 102 (C)
Undo anchor:  100 (canonical hash A')
```

Here the hash is unknown but the height is buffered. The intended `HeightMatch` behavior correctly purges `100..=102`: Tycho's local block 100 is stale. But because it is the first buffer entry, the inclusive `split_off(0)` empties `ReorgBuffer`. `handle_revert` then requires `get_most_recent_block()` to anchor the emitted revert message and returns `Reorg buffer is empty after purge` instead. The buffer has already been mutated, while the cursor is not advanced and neither WebSocket clients nor `PendingDeltas` receive a reconciliation event.

Simply returning `Ok(None)` after the empty purge would not be correct: consumers could retain the stale full blocks.

This is **not** the intended Base Flashblocks case:

```text
ReorgBuffer:          … 100, 101
PartialBlockBuffer:              102 (currently streaming)
Undo anchor:                      102
```

That anchor is numerically ahead of the *sealed* buffer because it only exists as a partial. It should preserve that partial, update the cursor, and emit no revert.

### Suggested policy

1. Keep `TargetAhead` nonfatal **only** when `PartialBlockBuffer` contains the undo target's exact height. Without that matching partial, an undo anchor above the sealed buffer violates the existing history/finality invariant and should remain fatal rather than silently becoming a no-op.
2. Choose explicitly how to handle a hash miss at the oldest buffered height:
   - **Conservative option (my recommendation unless we have production evidence):** treat it as fatal. Permit inclusive `HeightMatch` only when there is still an in-buffer predecessor to anchor the revert event. This preserves the finality/history boundary.
   - **Fully recoverable option:** fetch the stale oldest block's parent from storage, emit the inverse revert anchored at that parent, and extend `PendingDeltas` so it can clear its buffer when that anchor is the parent of its oldest pending block. This needs end-to-end coverage because its current strict `purge(anchor_hash)` would fail when the anchor is finalized and absent from `PendingDeltas`.

At minimum, please add tests for: (a) `TargetAhead` without a matching partial is fatal, (b) the valid Flashblocks partial case is nonfatal, and (c) the chosen oldest-height hash-miss behavior.

brunoguerios:
Thanks for the feedback. I applied all of them:
1. `TargetAhead` fatal _unless_ the `PartialBlockBuffer` holds the undo target's exact height.
2. `HeightMatch` at the oldest buffered height → fatal
3. Note on revisiting DCI in the future: a code comment documenting that DCI's `revert_to` is still hash-only and needs revisiting once DCI can run on Base.

crates/tycho-indexer/src/extractor/reorg_buffer.rs
```diff
@@ -178,6 +191,60 @@ where
...
+            // Neither hash nor height found: the target is below the oldest buffered
+            // block. Keep the exact error `purge` raises today — alerting greps it.
+            return Err(StorageError::NotFound("block".into(), target_hash.to_string()));
+        };
+        if idx == 0 {
+            // Purging inclusively from the oldest buffered block would empty the buffer,
+            // leaving no in-buffer predecessor to anchor the revert message. Fatal, same
+            // as a below-buffer target.
+            return Err(StorageError::NotFound("block".into(), target_hash.to_string()));
+        }
```
louise-poole: We emit the same error for 2 causes - I wonder if there is a clean way to differentiate them? (not important, just a thought)

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -1126,19 +1131,74 @@ where
...
+                    .partial_block_buffer
+                    .lock()
+                    .await
+                    .as_ref()
+                    .is_some_and(|partial| partial.block.number == block_ref.number);
+                if !partial_at_target {
+                    return Err(ExtractionError::ReorgBufferError(format!(
+                        "Revert target {} (hash {}) is above the sealed buffer with no pending partial at that height",
+                        block_ref.number, block_hash,
+                    )));
```
louise-poole: Elsewhere in this PR you mentioned that we grep logs to get these revert error alerts. Should we not model this error message after that text too so that it gets included in the alerts?

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -1126,19 +1131,74 @@ where
...
+                }
+                warn!(
+                    target_number = block_ref.number,
+                    "Revert target hash not in reorg buffer; target is streaming as partials, nothing to purge"
+                );
```
louise-poole, line 1179: Maybe we should include the partial index here? For debugging purposes?

crates/tycho-indexer/src/extractor/reorg_buffer.rs
```diff
@@ -64,6 +64,19 @@ where
     strict: bool,
 }
 
+/// Result of a height-aware purge. Hash match is authoritative; height is a fallback.
+#[derive(Debug)]
+pub enum PurgeOutcome<B> {
```
louise-poole: I think this reorg_buffer mod is `pub(crate)`. We should make this pub(crate) too then.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
louise-poole: Elsewhere in this PR you mentioned that we grep logs to get these revert error alerts. Should we not model this error message after that text too so that it gets included in the alerts?

louise-poole (approved):
Just some minor comments, otherwise approved! 
Small nit pick: in your PR description you reference a file reviewer don't have access to (`.claude/plans/2026-08-13-tycho-indexer-revert-purge-miss-design.md`).

crates/tycho-indexer/src/extractor/reorg_buffer.rs
brunoguerios: I don't think these cases deserve 2 separate errors.
The idx = 0 is more like a safety net. It's logically unfeasible, because it would mean a revert of dozens of blocks.
I'll refactor it a bit to make it easier to read and reason about the single error over here.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: I double checked and we don't grep these error logs to generate alerts. That was an oversight from a claim made by Claude. I'll update the error message so it's at least manually greppable if needed.

PR #1331, opened 2026-08-17
fix: survive reverts touching components with no persisted state

brunoguerios:
## Problem

The indexer process exited six times in one week, on Base and on Ethereum (07-31, 08-01, 08-03, 08-06). Each exit logged this line:

```
Could not find Component with id `0xee537c8d…`
```

`prod-tycho` and `prod-tycho-fynd` exited in the same second. They use separate databases, but read the same chain data. Each exit stops all extractors, so all websocket clients disconnect.

## Root cause

A revert restores the previous value of each changed attribute. `handle_revert` reads the reorg buffer, then Postgres. Three defects made this read fail.

1. The rule that excludes attributes created inside the reverted range applied to updated attributes only, not to deleted ones. A JIT liquidity provider creates a tick and deletes it in one block. That attribute has no previous value anywhere.
2. A component with no rows in Postgres was fatal. This state is normal: Postgres receives a block only after finality. On Base, finality is about 225 blocks behind the tip. The pool in the 08-06 exit was 212 blocks old.
3. The DB read can race the write path. The commit task drains blocks from the buffer before its write lands, and a small batch stays staged in the write cache.

## Fix

One chronological pass classifies the reverted attributes. The first event for each (component, attribute) decides: Creation reverts as a deletion, anything else looks up the prior value. An attribute born inside the range reverts as a deletion only when it survives the range — born and deleted inside it reverts to nothing. An attribute deleted and recreated inside the range gets its pre-range value back.

A miss no longer stops the extractor. The revert deletes the attribute, logs one summary warning, and increments `extractor_revert_attr_miss`. Its `component_state_found` label says whether the DB returned state rows for that component.

The reverted lookups never wait on the DB. `CachedGateway` records the height it flushed to Postgres. `ReorgBuffer` keeps drained blocks in a second, non-revertable section until that height passes them, and the state lookups read both sections. A buffer miss again proves no prior state exists.

## Notes for reviewers

Some integrations register an Update instead of a Creation when an attribute first appears (for example `ethereum-uniswap-v3/src/storage/pool_storage.rs:81`). Every counter hit names such a case.


tvinagre:
I agree with the removal of the fatal error for the known causes. In those causes, no prior state exists. A deletion is the correct revert result.

But there is one case where the new code can emit incorrect data:

1. `drain_blocks_until` removes finalized blocks from the reorg buffer immediately ([reorg_buffer.rs#L131-L136](https://github.com/propeller-heads/tycho/blob/58bac12642ac04c42cf0a760dcb92932824424eb/crates/tycho-indexer/src/extractor/reorg_buffer.rs#L131-L136)).
2. A background task then commits these blocks to the database ([protocol_extractor.rs#L726-L795](https://github.com/propeller-heads/tycho/blob/58bac12642ac04c42cf0a760dcb92932824424eb/crates/tycho-indexer/src/extractor/protocol_extractor.rs#L726-L795)).
3. The block-processing path waits for the previous commit task. But `handle_revert` does not wait for it (`commit_handle`).
4. The `CachedGateway` reads go directly to the database. They do not see writes that are not committed ([cache.rs#L993-L1017](https://github.com/propeller-heads/tycho/blob/58bac12642ac04c42cf0a760dcb92932824424eb/crates/tycho-storage/src/postgres/cache.rs#L993-L1017)).
5. Thus, while a commit is in progress, the drained state is not in the buffer and is not visible in the database. If a revert occurs in this window, the new code deletes state that exists ([protocol_extractor.rs#L1483-L1520](https://github.com/propeller-heads/tycho/blob/58bac12642ac04c42cf0a760dcb92932824424eb/crates/tycho-indexer/src/extractor/protocol_extractor.rs#L1483-L1520)). This is incorrect data.

Suggestion: make `handle_revert` wait for the pending commit task before it does the database lookup. The block-processing path already does this. Then "not found in the buffer or the database" always means "no prior state exists". After that change, the non-fatal path is safe.

Two more requests:

- Add an alert for the new counter. The counter is now the only signal for integrations that mark a `Creation` as an `Update`. Before this PR, a crash made these bugs visible.
- Add a label to the counter. The label must separate the expected cause (a component younger than the finality horizon) from the unexpected causes. PR #1330 used the same pattern: the `resolution` label on `extractor_revert_hash_miss`.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -1481,31 +1491,36 @@ where
...
+        for (component_id, keys) in missing_map {
+            let state = missing_components_states
+                .iter()
+                .find(|comp| comp.component_id == component_id);
+            if state.is_none() {
+                // No rows in buffer or DB: the component is younger than the finality
+                // horizon, or an upstream module mis-marked its attributes. Either way no
+                // previous value exists, so the revert deletes the attributes. Substreams
+                // output is external input — this must never be fatal.
+                warn!(
```
tvinagre: maybe even log as an error, now warning - so we can easily filter on the logs

brunoguerios: since we're also warning on the JIT liquidity scenario (where the component creation is still within the revert range), I don't think it should be an error - latest changes added a `cause` to the warning, so we can differentiate between - let me know your thoughts 😉

brunoguerios: here's the alert for the legit `unknown_component` scenario -> https://github.com/propeller-heads/helm-configuration/pull/935
this should be enough visibility to handle errors like these

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -854,6 +854,38 @@ where
...
+                        tx.protocol_components
+                            .get(component_id)
+                            .is_some_and(|component| component.change == ChangeType::Creation)
+                    })
+            })
+        })
+        .unwrap_or(false);
+
+    if creation_is_buffered {
+        "young_component"
```
tvinagre: I'd rather have this as an Enum instead of string.

crates/tycho-indexer/CLAUDE.md
```diff
@@ -95,17 +95,21 @@ On `BlockUndoSignal(target_hash, target_number)` from Substreams:
...
+2. If nothing was invalidated, only the cursor advances — no message is emitted. Otherwise
    a `BlockAggregatedChanges` with `revert = true` is broadcast.
-4. **No DB rollback is needed** — only finalized blocks ever reach the DB, so the persisted
-   state is always on the canonical chain.
+3. `handle_revert` first waits for any in-flight DB commit task, so a miss in the buffer and
+   the DB proves that no prior state exists. Previous attribute values are then restored from
+   the buffer, then the DB. Attributes created inside the reverted range, and attributes of
+   components with no state in either place, revert as deletions instead — with a warning and
+   the `extractor_revert_component_not_found` counter. Its `cause` label separates
+   `young_component` (creation still buffered — expected below the finality horizon) from
```
tvinagre: Valid, but this blocking await will be removed once the Blocker is addressed.

---

**[BIZ `#2`]**

**What changed**: The PR description covers only the two defects, but commits `ad8f962dc`/`8cc7e7e2b` add a blocking await of the pending commit task at the top of `handle_revert` and a `cause` label on the counter. Separately, the CLAUDE.md hunk deletes "above the buffer without a pending partial at exactly that height (the flashblocks case)" and the "partials at the target height are kept" step, although the `PurgeOutcome::TargetAhead` branch and the drop-partials-only-above-target logic are unchanged on this branch.

**Where**: `crates/tycho-indexer/CLAUDE.md:97–108`; `crates/tycho-indexer/src/extractor/protocol_extractor.rs:1150–1169`, `1208–1255`

- Compare CLAUDE.md steps 1–2 against `handle_revert` lines 1208–1255: the doc now describes a stricter rule than the code implements.
- Confirm the description is meant to omit the commit-wait — it is the change with the widest reach (every revert, not just the crash scenario).

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -1115,6 +1147,26 @@ where
...
+                    Ok(Err(err)) => return Err(err),
+                    Err(join_err) => {
+                        return Err(ExtractionError::Storage(StorageError::Unexpected(format!(
+                            "Failed to join database commit task: {join_err}"
+                        ))))
+                    }
+                }
+            }
+        }
+
```
tvinagre: **[BIZ `#3`]**

**What changed**: `handle_revert` takes `commit_handle` and awaits the previous commit task before the DCI revert, the buffer purge, and the lookups. Previously reverts ran concurrently with the commit; a commit `Err` was only observed at the next `handle_tick_scoped_data`.

**Where**: `crates/tycho-indexer/src/extractor/protocol_extractor.rs:1150–1169` (new) vs `745–784` (existing tick-side await)

- On Base/flashblocks, reverts are frequent and commit batches can be large (`database_commit_duration_ms`); check whether the added stall on the revert path is acceptable for websocket consumers waiting on the revert broadcast.
- The tick-side await emits `CommitTaskAwaited` with wait time; the revert-side one emits no metric/log, so the added latency is invisible to operators.
- A commit failure is now returned from `handle_revert` — confirm the runner treats that identically to the same error from the tick path (extractor exit), so the "never fatal" intent of the component-miss fix is not undercut by a different failure mode on the same call.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -854,6 +854,38 @@ where
...
+            })
+        })
+        .unwrap_or(false);
+
+    if creation_is_buffered {
+        "young_component"
+    } else {
+        "unknown_component"
+    }
+}
```
tvinagre: Confirmed: static attributes live in protocol_component.attributes (jsonb); protocol_state rows exist only for dynamic attributes, and get_protocol_states starts from protocol_state with an inner join. So a finalized component with zero dynamic rows (VM pools, or a pool whose first dynamic attr arrives later) returns nothing and is labelled unknown_component; once it has other rows, a mis-marked attr is not counted at all (silent not_found path).

---

**[BIZ `#4`]**

**What changed**: `revert_miss_cause` classifies solely by whether a `ChangeType::Creation` for the component is still in the (already purged) reorg buffer. A young component whose first attribute touch was mis-marked as `Update` (the exact `test_revert_update_attr_without_history_deletes` scenario) is labeled `young_component`; a correctly created component that is already finalized but has zero attribute rows is labeled `unknown_component`, because the DB query keys on attribute rows, not component existence.

**Where**: `crates/tycho-indexer/src/extractor/protocol_extractor.rs:861–887`, `1550–1571`; DB side `crates/tycho-storage/src/postgres/orm.rs:747–752`

- `get_protocol_states` starts from `protocol_state` with an inner join, so a component in `protocol_component` with no `protocol_state` rows yields nothing; the `unknown_component` label is then a proxy for "finalized component with a mis-marked attr", not "component never created", contradicting the CLAUDE.md text ("no creation anywhere").
- Alert/dashboard consumers of this counter should not treat `young_component` as ignorable if the goal (per PR notes) is to find integrations emitting `Update` for first-time attributes.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -1115,6 +1147,26 @@ where
...
+                    Ok(Ok(())) => {}
+                    Ok(Err(err)) => return Err(err),
+                    Err(join_err) => {
+                        return Err(ExtractionError::Storage(StorageError::Unexpected(format!(
+                            "Failed to join database commit task: {join_err}"
+                        ))))
+                    }
+                }
+            }
+        }
```
tvinagre: **[DESIGN `#1`]**

**Description**: The runner handles stream messages serially (`runner.rs:232–310`), so while `handle_revert` awaits the in-flight commit task no later block is processed and nothing is broadcast for that extractor. The wait is unconditional: every revert arriving while a commit is running stalls for the remaining `database_commit_duration_ms`, even when the revert could be resolved entirely from the reorg buffer. On flashblocks chains reverts against the streaming partial are routine (the `TargetAhead` branch at `:1208–1240` exists for exactly that), so on Base this puts DB latency on the block-processing path many times per block — contrary to the design rule that block processing never blocks on DB work (the tick path only awaits when the *next* batch is ready, allowing one batch of overlap). Per IMPL `#1` the wait still does not close the window it was added for (non-forced batches stay in the write cache), so the latency is paid without the guarantee.

**Suggestion**: Preferred: retain-until-committed — keep drained blocks in the reorg buffer (a committing segment or a `committed_up_to` height) until the commit task completes, so buffer lookups always cover the in-flight batch and no await is needed on the revert path; this also covers the write-cache window. Fallback if it must ship now: make the wait conditional — run `lookup_protocol_state` first and await the commit handle only when `missing` is non-empty, immediately before `get_protocol_states` (`:1531`), so shallow reverts never block.

**Scope**: Every revert on every chain; most visible on Base/flashblocks where partial-block consumers wait on the stalled revert broadcast.

tvinagre: **[DESIGN `#2`]**

**Description**: The new block in `handle_revert` (lock `commit_handle`, `take()`, `.instrument(info_span!("await_previous_commit"))`, three-arm match mapping `JoinError` to `StorageError::Unexpected("Failed to join database commit task: …")`) is a verbatim copy of `protocol_extractor.rs:755–775`. The two copies already diverge (the tick path records `needs_await` and emits `CommitTaskAwaited`; the revert path does not, and names the taken handle `pending_commit` where the tick path says `db_commit_handle_to_join`), so the join-error mapping and span name are now encoded in two places and will drift further. `GatewayInner` owns the handle, so it is the natural home for this behaviour.

**Suggestion**: Add `async fn join_pending_commit(&self) -> Result<bool, ExtractionError>` (or return the wait `Duration`) on `GatewayInner<G>` that takes the handle, awaits it under the shared span and maps errors once; `handle_tick_scoped_data` uses the returned flag for its `CommitTaskAwaited` trace, `handle_revert` ignores it. Pros: one error mapping, one span name, revert path gets the wait-time trace for free. Cons: none beyond a small refactor of existing code.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
tvinagre: **[DESIGN `#3`]**

**Root cause**: The function bundles a `ReorgBuffer` question ("is a Creation for this id buffered?") with an extractor/metrics concern (map a bool to one of two `&'static str` labels), and asks the buffer question through `get_block_range(None, None)` per missing component.

- **Architecture** *(was DESIGN `#3`)*: the strings `"young_component"` / `"unknown_component"` are the sole encoding of the cause and are consumed by `warn!`, `counter!` (`:1560–1571`) and the test (`:4858`); the conventions doc asks for enums over primitives. The `&ReorgBuffer<BlockUpdateWithCursor<BlockChanges>>` parameter pins the helper to one concrete instantiation.
- **Complexity** *(was SIMPLIFY `#3`)*: `get_block_range(None, None)` cannot fail (both index lookups are skipped), so `.map(...).unwrap_or(false)` is dead error-swallowing; the call also records a span field and formats a `buffered_range` string on every invocation.
- **Performance** *(was IMPL `#4`)*: for each component in `missing_map` with no DB rows the helper iterates every buffered block and every tx. On Base the buffer holds ~225 blocks; a revert that misses on many components (a batch of new pools from one factory, or the unflushed-batch window in IMPL `#1`) turns this into blocks × txs × components.

**Proposed fix**: Move the scan into `ReorgBuffer` as `fn buffered_component_creations(&self) -> HashSet<&str>` (iterate `self.block_messages` directly — no `Result`, no span), computed lazily once on the first miss, and define `enum RevertMissCause { YoungComponent, UnknownComponent }` with `fn label(self) -> &'static str` used by both the log field and the metric label. One type owns the label vocabulary (architecture), the impossible-error path and per-call span overhead disappear (complexity), and each miss becomes a set lookup (performance).

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -854,6 +854,38 @@ where
...
+        })
+        .unwrap_or(false);
+
+    if creation_is_buffered {
+        "young_component"
+    } else {
+        "unknown_component"
+    }
+}
+
```
tvinagre: **[DESIGN `#4`]**

**Root cause**: `revert_miss_cause` classifies by "is a Creation still buffered", but the doc-comment, the `warn!`, and CLAUDE.md present `young_component` as "expected — no prior state can exist" and `unknown_component` as "no creation anywhere — an upstream bug". Neither claim follows from the check.

- **Architecture** *(was DESIGN `#4`)*: by the time this branch runs, the commit was awaited (`:1152`), the buffer was consulted, and components created inside the reverted range were already filtered out via `reverted_components_creations` (`:1486`). A `young_component` miss therefore means an attribute reached the DB lookup with no prior value anywhere — an `Update`/`Deletion` on an attribute never `Creation`-marked, i.e. the same integration defect the PR attributes to `unknown_component` (`pool_storage.rs:81`), merely on a younger component (the 08-06 crash pool was 212 blocks old, i.e. young). Labelling it "expected" under-counts the very thing the counter was added to find. The attribute-level miss on a component that *does* exist (`not_found` at `:1580–1584`) is the identical defect and is neither logged nor counted.
- **Behavior** *(was IMPL `#3`)*: a component whose creation was drained and committed but that has no attribute rows yet is reported as `unknown_component`: `orm::ProtocolState::by_id` starts from `protocol_state` with an inner join (`crates/tycho-storage/src/postgres/orm.rs:747–752`), so a component with zero attributes yields no `ProtocolComponentState` even though it is fully known to the DB.
- **Docs** *(was DOCS `#2`, second half)*: CLAUDE.md says `unknown_component` means "no creation anywhere", but the classifier only scans the reorg buffer; it never consults the DB for a creation row.

**Proposed fix**: Either drop the split and document that every hit is an upstream marking bug (simplest, and matches reality), or make the labels factual (`creation_buffered` / `creation_not_buffered`) and rewrite the docstring and CLAUDE.md to match. If the goal is a direct measure of mis-marked integrations irrespective of finality, count at the attribute level instead — one increment per key that ends up in `not_found`, with a label for whether the component row was found — which also covers the uncounted found-component case. If detecting truly unknown components matters, check `protocol_component` existence rather than presence of state rows.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -1115,6 +1147,26 @@ where
             return Ok(None);
         }
 
+        // The commit task drains blocks from the buffer before its DB write lands. Wait for
```
tvinagre: **[DESIGN `#5`]**

The comment justifies the wait only for the protocol-state lookup, but the earlier `get_contracts` call at `:1377` has the same drain-before-write window; say that the wait covers every DB read in the revert so nobody later moves the block down.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -1481,31 +1543,39 @@ where
...
-
-        let mut not_found: HashMap<_, HashSet<_>> = HashMap::new();
+        let mut not_found: HashMap<String, HashSet<String>> = HashMap::new();
         let mut db_states: HashMap<(String, String), Bytes> = HashMap::new();
 
-        for (state, keys) in missing_components_states_map {
+        for (component_id, keys) in missing_map {
+            let state = missing_components_states
+                .iter()
+                .find(|comp| comp.component_id == component_id);
```
tvinagre: **[DESIGN `#6`]**

`missing_components_states.iter().find(...)` inside the `missing_map` loop is O(components²); index `missing_components_states` into a `HashMap<&str, &ProtocolComponentState>` once before the loop.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
tvinagre: The window is post-sync, not during sync: the last syncing batch is committed with force_db_commit=false and stays in open_tx (drained from the buffer), sync flips off, and the next forcing commit only happens once finality advances past the first post-sync block (~225 blocks on Base). Reverts in that window miss both places. Retain-until-committed (DESIGN `#1`) closes it; the await does not.

---

**[IMPL `#1`]**

**Category**: bug

**Description**: The new block awaits the pending commit task and the comment claims that afterwards "a miss in the buffer and the DB proves that no prior state exists". That only holds when the awaited batch was flushed. `process_full_block_message` (line 790) passes `force_db_commit = false` for every block while `is_syncing`, and `ExtractorPgGateway::advance` (line 2118) then calls `commit_transaction(self.db_tx_batch_size)`, which — when the op count is below the threshold — puts the transaction back into `CachedGateway::open_tx` without writing to Postgres (`crates/tycho-storage/src/postgres/cache.rs:633–670`). `get_protocol_states` reads straight from the connection pool (`cache.rs:993`), not through `open_tx`. So in the window between the last syncing-era drain and the first non-syncing drain, the drained blocks' attribute values are neither in the buffer nor visible to the DB query. A revert in that window now deletes those attributes (with cause `unknown_component`) instead of restoring them. `test_handle_revert_waits_for_pending_commit` cannot see this because the mock gateway has no write cache.

**Plain English**: When the indexer has just caught up with the chain tip and a reorg arrives shortly after, some recently finalized values may still be sitting in an in-memory write queue rather than in the database. The revert will then not find the old value and will wipe the attribute from every consumer's view instead of restoring it, while the metric blames an "unknown component". The revert should either force the queued writes to the database before looking up prior values, or read through the queue.

**Suggestion**: Add a gateway method that flushes the open transaction (`commit_transaction(0)` semantics, e.g. `ExtractorGateway::flush`) and call it in `handle_revert` after awaiting the commit handle; alternatively have the commit task always force-commit the last block of a batch once `final_block_height` is within one batch of the head. At minimum, soften the comment and CLAUDE.md wording so operators do not read the `unknown_component` label as proof of an upstream bug.

**Scope**: The unflushed-batch window predates this branch (before it, the same window was fatal or a silent `not_found`). In scope because the branch's comment, CLAUDE.md text and counter semantics are built on a guarantee the wait does not provide.

tvinagre (commented):
Review of the branch at c5b2149 (reviewed at 18e1b8d; rebase since is content-identical for both files). Two blockers: the unconditional commit-await on the revert path (DESIGN ``#1``) and delete-then-recreate in range losing the prior value (IMPL ``#2``). Details inline.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -1418,7 +1470,8 @@ where
                 acc
```
tvinagre: Blocker as this would lead to wrong state being served.

---

**[IMPL `#2`]**

**Category**: bug

**Root cause**: `reverted_created_attrs` collects every `(c_id, attr)` that appears in any `created_attributes` set anywhere in the reverted range, regardless of whether that creation was the attribute's *first* event in the range; a second pass then subtracts that set from the lookup keys through a closure that names the wrong predicate.

**Plain English**: If a liquidity tick that already existed before the reorg is removed and then added back within the reverted blocks, the revert will erase it from consumers' state instead of putting back its original value. Consumers would then see a pool with a missing tick until the next update touches it. The revert should restore the original value whenever the attribute existed before the reverted range, regardless of what happened to it inside the range.

- **Behavior** *(was IMPL `#2`)*: for an attribute with real prior value V, block N deletes it and block N+1 re-creates it (a burned then re-minted tick). The attr lands in `reverted_created_attrs`, so `has_prior_state` filters it out of the lookup for both the update path and — new in this PR — the deletion path; the created-attrs loop at `:1616–1624` then emits it as a deletion. Before this branch the deletion was still looked up, so the delta carried both `updated_attributes[attr] = V` and `deleted_attributes ∋ attr` (contradictory, but V was present); now it is deterministically a deletion. `test_revert_deleted_attr_with_history_restores_value` covers delete-without-recreate only.
- **Architecture** *(was DESIGN `#5`)*: the second pass (`:1481–1512`) re-walks the same `reverted_state` / `txs_with_update` / `state_updates` / `reverted_components_creations` chain as the first pass only to subtract its result, so the rule "attrs created in range have no history" is stated in the first-pass comment, the closure, the second-pass comment and the created-attrs loop.
- **Naming** *(was DOCS `#7`)*: `has_prior_state` returns true for anything not in `reverted_created_attrs`, including mis-marked Updates that provably have no prior state; the real prior-state check is the buffer+DB lookup that follows. `created_in_range` (negated) states the actual predicate.
- **Complexity** *(was SIMPLIFY `#4`)*: the three-line comment above the closure (`:1488–1490`) repeats the "First pass" comment at `:1443` and the "Second pass" comment at `:1481`.

**Proposed fix**: One chronological pass over `reverted_state` (oldest → newest, `txs_with_update` in order) recording the first `ChangeType` seen per `(c_id, attr)`; keys whose first event is `Creation` go to a `created_in_range` set, all others go to the lookup set. This restores V for delete-then-recreate on both the update path (pre-existing) and the deletion path (extended here), collapses the two passes and their four restatements of the rule into one, and removes the need for the mis-named closure and its comment.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -1481,31 +1543,39 @@ where
...
+                    "Component not found in buffer or DB during revert; \
+                     reverting its attributes as deletions"
+                );
+                counter!(
+                    "extractor_revert_component_not_found",
+                    "extractor" => self.name.clone(),
+                    "chain" => self.chain.to_string(),
+                    "cause" => cause,
+                )
+                .increment(1);
```
tvinagre: **[IMPL `#3`]**

Consistent with the other counters in this function; if DESIGN `#3` lands, hoist the labels out of the loop at the same time.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -1481,31 +1543,39 @@ where
...
+                // horizon, or an upstream module mis-marked its attributes. Either way no
+                // previous value exists, so the revert deletes the attributes. Substreams
+                // output is external input — this must never be fatal.
+                let cause = revert_miss_cause(&reorg_buffer, &component_id);
+                warn!(
+                    component_id = component_id.as_str(),
+                    cause,
+                    "Component not found in buffer or DB during revert; \
+                     reverting its attributes as deletions"
+                );
```
tvinagre: **[IMPL `#4`]**

A revert that misses on many components emits one warning line each; a single summary line with the count and the list of ids would be easier to correlate with the counter.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -3970,6 +4040,848 @@ mod test {
...
+            .lock()
+            .await
+            .replace(fake_commit)
+        {
+            real_commit.await.unwrap().unwrap();
+        }
+
+        // Revert to block 2. `tick`'s prior value only exists in the pending commit.
+        let revert_msg = extractor
+            .handle_revert(BlockUndoSignal {
```
tvinagre: **[SIMPLIFY `#1`]**

There is precedent for touching `gateway.commit_handle` in tests (lines 2347, 5494), so this is acceptable, but if TESTS `#2` lands, a `fn replace_commit_handle(ex, fut)` helper that also joins the displaced real handle would turn these 14 lines into one call and keep the private-field poke in a single place.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -1115,6 +1147,26 @@ where
...
+                {
+                    Ok(Ok(())) => {}
+                    Ok(Err(err)) => return Err(err),
+                    Err(join_err) => {
+                        return Err(ExtractionError::Storage(StorageError::Unexpected(format!(
+                            "Failed to join database commit task: {join_err}"
+                        ))))
+                    }
+                }
+            }
```
tvinagre: **[TESTS `#1`]**

**Description**: The new block awaits the pending commit and has three arms: `Ok(Ok(()))`, `Ok(Err(err)) => return Err(err)` (line 1160), and `Err(join_err)` mapped to `StorageError::Unexpected` (lines 1161–1165). Only the success arm is exercised (`test_handle_revert_waits_for_pending_commit`, line 4555). A failed or panicked commit turning a revert into a fatal `Err` is exactly the kind of behaviour the PR is about (process exits on revert), and nothing pins it. The rust conventions doc asks to "test edges and errors".

**Suggestion**: Add one test (or two) that replaces `commit_handle` with a fake task the same way the existing test does (lines 4677–4692), but returning `Err(ExtractionError::Storage(...))` in one case and `panic!()` in the other, then assert `handle_revert` returns `Err` and that `commit_handle` is `None` afterwards (the handle must be `take()`n even on failure so a later tick does not re-await it). Both fit in ~20 lines each if they reuse `revert_test_gateway()`.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -3970,6 +4040,848 @@ mod test {
...
+        gw.expect_advance()
+            .times(0)
+            .returning(|_, _, _| Ok(()));
+        gw.expect_get_contracts()
+            .returning(|_| Ok(Vec::new()));
+        // Component is non-finalized: not in DB.
+        gw.expect_get_protocol_states()
+            .returning(|_| Ok(Vec::new()));
+        gw.expect_get_components_balances()
+            .returning(|_| Ok(HashMap::new()));
```
tvinagre: **[TESTS `#2`]**

**Description**: `revert_test_gateway()` (line 3439) is byte-for-byte the 20-line mock block that `test_revert_attr_created_and_deleted_in_range` (4051–4070), `test_revert_deleted_attr_with_history_restores_value` (4198–4217), `test_revert_update_attr_without_history_deletes` (4337–4356), `test_revert_attr_on_component_absent_everywhere_deletes` (4467–4486) and `test_revert_component_not_found_counter_cause_label` (4737–4757) inline. Only `test_handle_revert_waits_for_pending_commit` genuinely needs a custom `get_protocol_states`/`advance`. The pre-existing `test_revert_new_attribute_on_non_finalized_component` (3892) has the same copy, which is presumably where the new tests were cloned from, but the branch adds five more copies on top. Likewise the "Block 1: empty anchor" tick (e.g. 4076–4085) and the `BlockUndoSignal { last_valid_block: Some(BlockRef { id: format!("0x{:0>64x}", n), number: n }), .. }` literal are repeated in every test.

**Suggestion**: Use `create_extractor(revert_test_gateway()).await` in the five tests with the default mock. Add two tiny test-module helpers so the scenario reads as data instead of 60 lines of protobuf: e.g. `fn component_creation_tx(block: u64, id: &str, attrs: &[(&str, u64, PbChangeType)]) -> TransactionChanges` and `fn entity_change_tx(block: u64, idx: u64, component: &str, attr: &str, value: Vec<u8>, change: PbChangeType) -> TransactionChanges`, plus `fn undo_to(n: u64) -> BlockUndoSignal`. Each new test then shrinks to roughly 30–40 lines (from 130–165), saving on the order of 600 of the ~840 added test lines, and the per-test `use` blocks disappear once the fixtures own the protobuf imports; the differences between the six scenarios (the actual subject of review) become visible at a glance.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -3970,6 +4040,848 @@ mod test {
...
+            cause_counts.get("young_component"),
+            Some(&1),
+            "pool_y creation is buffered, its miss is the expected cause"
+        );
+        assert_eq!(
+            cause_counts.get("unknown_component"),
+            Some(&1),
+            "pool_z has no creation anywhere, its miss is the unexpected cause"
+        );
+    }
```
tvinagre: **[TESTS `#3`]**

**Description**: The production counter is keyed by `extractor`, `chain` and `cause` (lines 1565–1571). The test's fold keeps only `cause`, uses `unwrap_or_default()` (4865) for a missing label and `_ => 0` (4868) for a non-counter value, so an accidental rename of the label to e.g. `reason`, or dropping the `extractor`/`chain` labels, is not caught — the map would just contain an empty-string key and the two `assert_eq!`s would still fail, but for a reason the message does not explain. `services/middleware/metrics.rs` already has `snapshot_to_map` + `counter_value(map, name, &[(k, v), ..])` (lines 141–171) that match on the full label set.

**Suggestion**: Assert on the full key: `counter_value(&map, "extractor_revert_component_not_found", &[("extractor", EXTRACTOR_NAME), ("chain", "ethereum"), ("cause", "young_component")]) == 1` (and the same for `unknown_component`). Either lift `snapshot_to_map`/`counter_value` into `testing.rs` or reimplement the 10 lines here; either way drop the lossy `unwrap_or_default`/`_ => 0`.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -3970,6 +4040,848 @@ mod test {
...
+            .unwrap()
+            .unwrap();
+
+        // Replace the finished commit handle with a slow in-flight commit.
+        let fake_flag = committed.clone();
+        let fake_commit = tokio::spawn(async move {
+            tokio::time::sleep(std::time::Duration::from_millis(250)).await;
+            fake_flag.store(true, Ordering::SeqCst);
+            Ok(())
+        })
```
tvinagre: **[TESTS `#4`]**

**Description**: The fake commit sleeps 250 ms before flipping `committed`. The assertion does not race (the flag can only flip after the task runs), so the test is not flaky, but it costs a quarter second every run and its ability to catch a regression is timing-dependent: if `handle_revert` were to skip the await and take longer than 250 ms to reach `get_protocol_states` on a loaded CI box, the flag would already be set and the test would pass. The existing `test_handle_tick_scoped_respects_batch_async_commit` (line 2410) also uses a sleep, so there is precedent, but that one at least gates the mock on a channel.

**Suggestion**: Mark the test `#[tokio::test(start_paused = true)]`. On the current-thread runtime, paused time auto-advances only when every task is idle, i.e. only while `handle_revert` is actually awaiting the handle; the test becomes instant and a missing await can no longer pass by accident. Alternatively gate the fake commit on a `tokio::sync::oneshot` that the mocked `get_protocol_states` never releases, so a missing wait deadlocks visibly instead of passing.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -854,6 +854,38 @@ where
...
+                    })
+            })
+        })
+        .unwrap_or(false);
+
+    if creation_is_buffered {
+        "young_component"
+    } else {
+        "unknown_component"
+    }
```
tvinagre: **[TESTS `#5`]**

**Description**: `revert_miss_cause` reports `unknown_component` for any component without a buffered creation. The counter test covers "creation in buffer" and "no creation anywhere", but not the third real-world shape: the creation was finalized and drained, the DB has no attribute rows for the component (created with no attributes, then an `Update` in the reverted range), so `get_protocol_states` returns nothing. That case produces the `unknown_component` label for a perfectly legitimate component, which is the opposite of what the CLAUDE.md text ("no creation anywhere — an upstream bug") promises. Whether that is the intended semantics is a design call (see DESIGN `#4`), but a test should pin whichever answer is chosen; today the behaviour is untested and undocumented.

**Suggestion**: Add a case to the counter test (block 1 creates `pool_w` with no attributes at finality 1, block 2 with finality 2 drains it, block 3 updates `pool_w`, revert to 2) and assert the label the team actually wants. If the answer is "still `unknown_component`", say so in the docstring of `revert_miss_cause`.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -3970,6 +4040,848 @@ mod test {
...
+        .instrument(tracing::info_span!("fake_commit"));
+        if let Some(real_commit) = extractor
+            .gateway
+            .commit_handle
+            .lock()
+            .await
+            .replace(fake_commit)
+        {
+            real_commit.await.unwrap().unwrap();
+        }
```
tvinagre: **[TESTS `#6`]**

The comment says block 2 "spawns the commit task", but if it did not, the `if let` silently skips and the test still passes because the fake commit is what is awaited; `.expect("block 2 must have spawned a commit task")` would make the setup self-checking.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -3970,6 +4040,848 @@ mod test {
...
+            .get("pool_ghost")
+            .expect("state_deltas should contain pool_ghost");
+        assert!(
+            ghost_delta
+                .deleted_attributes
+                .contains("reserve"),
+            "Expected reserve in deleted_attributes, got: {:?}",
+            ghost_delta.deleted_attributes
+        );
+    }
```
tvinagre, line 4557: **[TESTS `#7`]**

The other three deletion tests assert both `deleted_attributes.contains(..)` and `updated_attributes.is_empty()`; this one asserts only the former, so a regression that both deleted and "restored" `reserve` would slip through.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -3970,6 +4040,848 @@ mod test {
...
+        use metrics_util::debugging::{DebugValue, DebuggingRecorder};
+
+        let recorder = DebuggingRecorder::new();
+        let snapshotter = recorder.snapshotter();
+        let rt = tokio::runtime::Builder::new_current_thread()
+            .enable_all()
+            .build()
+            .unwrap();
+
+        metrics::with_local_recorder(&recorder, || {
```
tvinagre: **[TESTS `#8`]**

`metrics::with_local_recorder` takes a sync closure, so `#[tokio::test]` cannot be used and a hand-built `current_thread` runtime is the right call (and better isolated than the global `install()` used in `services/middleware/metrics.rs:126–138`), but the next reader will wonder why this is the only non-`#[tokio::test]` test in the module; a short comment saves them the trip.

crates/tycho-indexer/CLAUDE.md
```diff
@@ -95,17 +95,21 @@ On `BlockUndoSignal(target_hash, target_number)` from Substreams:
...
-   fallbacks log a warning and increment `extractor_revert_hash_miss`.
-2. Pending partials are dropped only when above the target height; partials at the target
-   height are the valid prefix of the last valid block and are kept.
-3. If nothing was invalidated, only the cursor advances — no message is emitted. Otherwise
+   the revert), or above the buffer. Nonfatal hash-miss fallbacks log a warning and increment `extractor_revert_hash_miss`.
+2. If nothing was invalidated, only the cursor advances — no message is emitted. Otherwise
    a `BlockAggregatedChanges` with `revert = true` is broadcast.
-4. **No DB rollback is needed** — only finalized blocks ever reach the DB, so the persisted
-   state is always on the canonical chain.
+3. `handle_revert` first waits for any in-flight DB commit task, so a miss in the buffer and
```
tvinagre: **[DOCS `#1`]**

**Description**: The hunk deletes two statements that the code on this branch still implements. Old bullet 1 said a hash miss above the buffer is fatal "without a pending partial at exactly that height (the flashblocks case — the only legitimate target-ahead shape)"; the new text says simply "or above the buffer", which is wrong: `PurgeOutcome::TargetAhead` at `protocol_extractor.rs:1208–1240` is non-fatal when the pending partial's height equals the target and increments `extractor_revert_hash_miss{resolution="target_ahead"}`. Old bullet 2 ("Pending partials are dropped only when above the target height; partials at the target height are ... kept") was deleted outright, but `protocol_extractor.rs:1243–1255` still does exactly that. No commit on the branch touches this logic, so this is a doc regression relative to main, not a description of a behaviour change.

**Suggestion**: Restore the "without a pending partial at exactly that height (the flashblocks case)" clause in bullet 1 and the "pending partials are dropped only above the target height" bullet, then renumber. Keep the new commit-wait/delete-on-missing bullet as an addition, not a replacement.

crates/tycho-indexer/CLAUDE.md
```diff
@@ -95,17 +95,21 @@ On `BlockUndoSignal(target_hash, target_number)` from Substreams:
...
+   the DB proves that no prior state exists. Previous attribute values are then restored from
+   the buffer, then the DB. Attributes created inside the reverted range, and attributes of
+   components with no state in either place, revert as deletions instead — with a warning and
+   the `extractor_revert_component_not_found` counter. Its `cause` label separates
+   `young_component` (creation still buffered — expected below the finality horizon) from
+   `unknown_component` (no creation anywhere — an upstream bug, e.g. a Creation marked as an
+   Update).
+4. **No DB rollback is needed** — only finalized blocks ever reach the DB, so the persisted state
+   is always on the canonical chain.
```
tvinagre: **[DOCS `#2`]**

**Description**: The list is presented as a sequence ("1. purge_to ... 2. ... broadcast ... 3. `handle_revert` first waits for any in-flight DB commit task"). In code the wait is at `protocol_extractor.rs:1150–1167`, before `purge_to` at line 1190 and long before the broadcast. Placing it third with the word "first" is self-contradictory and implies the buffer/DB lookup happens after the message is emitted.

**Suggestion**: Move the commit-wait sentence to a new bullet 1 (or the start of bullet 1) and describe the lookup/deletion behaviour as part of building the broadcast message, before the "broadcast" bullet. The "no creation anywhere" wording is covered by DESIGN `#4`.

crates/tycho-indexer/CLAUDE.md
```diff
@@ -95,17 +95,21 @@ On `BlockUndoSignal(target_hash, target_number)` from Substreams:
...
-   height purges from that height inclusive (stale copy). A hash miss is fatal when the
+   height purges from that height inclusive (stale copy). A hash miss is fatal when the height
    target is below the buffer, at the buffer's oldest block (no predecessor left to anchor
-   the revert), or above the buffer without a pending partial at exactly that height (the
-   flashblocks case — the only legitimate target-ahead shape). Nonfatal hash-miss
-   fallbacks log a warning and increment `extractor_revert_hash_miss`.
-2. Pending partials are dropped only when above the target height; partials at the target
-   height are the valid prefix of the last valid block and are kept.
-3. If nothing was invalidated, only the cursor advances — no message is emitted. Otherwise
+   the revert), or above the buffer. Nonfatal hash-miss fallbacks log a warning and increment `extractor_revert_hash_miss`.
```
tvinagre: **[DOCS `#3`]**

Line 100 is 123 characters (the file otherwise wraps near 90), and the re-wrap moved "height" to the end of line 97 so the sentence reads "fatal when the height / target is below the buffer"; rewrap the paragraph so "height target" stays together and no line exceeds 100.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -854,6 +854,38 @@ where
     }
 }
 
+/// Classifies a revert lookup miss. "young_component" means the component's creation is still
+/// in the reorg buffer, so it is younger than the finality horizon — expected, no prior state
+/// can exist. "unknown_component" means no creation explains the miss — unexpected, e.g. an
+/// integration marked a Creation as an Update.
```
tvinagre: **[DOCS `#4`]**

**Description**: The first line "Classifies a revert lookup miss." does not tell the reader the function returns a `&'static str` intended as a metric label value, nor that it inspects only the reorg buffer. "Finality horizon" appears nowhere else in the crate; the code and existing tests use `final_block_height` / "finality height" / "non-finalized" (`protocol_extractor.rs:711–737`, `3887`, `4066`). The comment also explains the label semantics inline with quoted strings, which will drift if the values change.

**Suggestion**: Something like: "Returns the `cause` label for `extractor_revert_component_not_found`: `young_component` when the component's Creation is still in the reorg buffer (not yet finalized, so no prior state can exist), otherwise `unknown_component` (no buffered creation explains the miss, e.g. an integration emitted an Update for a first write). Only the buffer is inspected." Use the crate's existing "finalized"/`final_block_height` vocabulary instead of "finality horizon".

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -854,6 +854,38 @@ where
...
+                            .is_some_and(|component| component.change == ChangeType::Creation)
+                    })
+            })
+        })
+        .unwrap_or(false);
+
+    if creation_is_buffered {
+        "young_component"
+    } else {
+        "unknown_component"
```
tvinagre: **[DOCS `#5`]**

"Young" does not appear elsewhere in the crate whereas "non-finalized component" is the established phrase (`test_revert_new_attribute_on_non_finalized_component`, test comments at lines 3887/4066/4352); `non_finalized_component` (or `unfinalized_component`) would read directly off the existing test names and the CLAUDE.md "only finalized blocks ever reach the DB" bullet.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -1481,31 +1543,39 @@ where
...
+                // output is external input — this must never be fatal.
+                let cause = revert_miss_cause(&reorg_buffer, &component_id);
+                warn!(
+                    component_id = component_id.as_str(),
+                    cause,
+                    "Component not found in buffer or DB during revert; \
+                     reverting its attributes as deletions"
+                );
+                counter!(
+                    "extractor_revert_component_not_found",
```
tvinagre: **[DOCS `#6`]**

The only other revert metric is `extractor_revert_hash_miss` (lines 1200, 1233) and the classifier is documented as classifying a "lookup miss", so `extractor_revert_component_miss` would keep the family consistent; if `not_found` is kept, consider aligning the doc-comment wording instead.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -1481,31 +1543,39 @@ where
...
 
-        for (state, keys) in missing_components_states_map {
+        for (component_id, keys) in missing_map {
+            let state = missing_components_states
+                .iter()
+                .find(|comp| comp.component_id == component_id);
+            if state.is_none() {
+                // No rows in buffer or DB: the component is younger than the finality
+                // horizon, or an upstream module mis-marked its attributes. Either way no
+                // previous value exists, so the revert deletes the attributes. Substreams
```
tvinagre: **[DOCS `#7`]**

The same function deliberately keeps several Substreams-driven shapes fatal ("must stay fatal", lines 1210–1211, 1221–1224), so the blanket "must never be fatal" contradicts the file's own policy; say why this particular miss is benign (no prior value can exist, so deletion is the correct revert) and drop the generalisation.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -3970,6 +4040,848 @@ mod test {
...
+            ghost_delta
+                .deleted_attributes
+                .contains("reserve"),
+            "Expected reserve in deleted_attributes, got: {:?}",
+            ghost_delta.deleted_attributes
+        );
+    }
+
+    #[tokio::test]
+    async fn test_handle_revert_waits_for_pending_commit() {
```
tvinagre: **[DOCS `#8`]**

Every other revert test in this module is `test_revert_*` (lines 3208–4721), so `test_revert_waits_for_pending_commit` would sort alongside them; `pool_ghost` and `pool_z` for the never-created component are fine, but consider one id for both tests since they exercise the same case.

crates/tycho-indexer/CLAUDE.md
```diff
@@ -95,17 +95,21 @@ On `BlockUndoSignal(target_hash, target_number)` from Substreams:
...
    a `BlockAggregatedChanges` with `revert = true` is broadcast.
-4. **No DB rollback is needed** — only finalized blocks ever reach the DB, so the persisted
-   state is always on the canonical chain.
+3. `handle_revert` first waits for any in-flight DB commit task, so a miss in the buffer and
+   the DB proves that no prior state exists. Previous attribute values are then restored from
+   the buffer, then the DB. Attributes created inside the reverted range, and attributes of
+   components with no state in either place, revert as deletions instead — with a warning and
+   the `extractor_revert_component_not_found` counter. Its `cause` label separates
+   `young_component` (creation still buffered — expected below the finality horizon) from
+   `unknown_component` (no creation anywhere — an upstream bug, e.g. a Creation marked as an
```
tvinagre: **[DOCS `#9`]**

A found component with a missing attribute is also deleted (line 1577–1581) but silently, without the warning or counter; the sentence bundles both under "with a warning and the counter", so either qualify it ("components absent from both — logged and counted — and, silently, attributes absent on a found component") or drop the attribute clause.

crates/tycho-indexer/CLAUDE.md
brunoguerios: PR description updated: it now covers the commit-wait (conditional since 5c795e03f) and the new counter. The deleted CLAUDE.md statements are restored in 9ec6cc65e.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: The wait is now conditional (see DESIGN #1) and emits a `RevertCommitAwaited` trace with the wait time. A commit failure surfacing from `handle_revert` exits the extractor exactly like the tick path; the miss itself stays non-fatal. 5c795e03f.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: The split is gone. The counter is attribute-level with a factual `component_found` label (whether the DB returned state rows), so neither label claims "expected" or "upstream bug" anymore. 4772f8bb6.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: Took the conditional shape, extended: the revert settles (joins the commit task, then flushes the write cache) at most once, and only before a DB read whose buffer lookup missed. Shallow reverts never block. 5c795e03f, flush in 331d8cee3. On retain-until-committed: keyed on task completion it does not cover the write-cache window — a below-threshold batch stays in `open_tx` after the task completes Ok — and keyed on real flushes it needs flush-height plumbing through `CachedGateway`. Kept as a follow-up.

brunoguerios: Added `GatewayInner::join_pending_commit`; it takes the handle even on failure, maps errors once, and returns whether an unfinished task was awaited. Both paths use it. a0897399c.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: `revert_miss_cause` is deleted together with the split (see DESIGN #4), so the buffer scan, the dead error path, and the per-call span are all gone. 4772f8bb6.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: Took the attribute-level option: `extractor_revert_attr_miss`, one increment per key that ends in `not_found`, labeled `component_found` = DB returned state rows (documented as rows, not component existence). The found-component miss is now counted too. 4772f8bb6.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: The settle is a helper invoked before both `get_contracts` and `get_protocol_states`; its comment states it covers every DB read in the revert. 5c795e03f.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: Indexed into a `HashMap` (`states_by_id`) before the loop. 4772f8bb6.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: Confirmed. Added `CachedGateway::flush` / `ExtractorGateway::flush` (331d8cee3); the revert flushes after joining the commit task, before any DB read that missed (5c795e03f), so the "miss proves no prior state" claim now holds.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: Replaced the two passes with one chronological pass keyed on the first event per (component, attribute): first event Creation → deletion, anything else → lookup. Delete-then-recreate now restores the prior value; pinned by `test_revert_recreated_attr_restores_prior_value`. 3fd6e6393.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: Counts are accumulated in the loop; at most two counter emissions per revert. 4772f8bb6.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: One summary warning with per-component miss counts and the total. 4772f8bb6.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: Added `replace_commit_handle`; it joins any displaced real handle and returns whether one existed. All tests go through it. 5c795e03f.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: Added `test_revert_pending_commit_error_is_fatal` and `test_revert_pending_commit_panic_is_fatal`; both assert the error surfaces and the handle is `None` afterwards. 5c795e03f.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: Added `tick` / `creation_tx` / `entity_change_tx` / `undo_to` (3fd6e6393) and converted the duplicated tests to `revert_test_gateway()` + helpers (3681df50c).

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: Reimplemented `snapshot_to_map` / `counter_value` locally; the test asserts on the full label set (`extractor`, `chain`, `component_found`). 4772f8bb6.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: `#[tokio::test(start_paused = true)]`: the test is instant and a skipped await fails deterministically. 5c795e03f.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: The causal split is gone; the rewritten counter test pins the zero-attribute shape (`pool_y` → `component_found="false"`) and the rows-without-the-attr shape (`pool_w` → `"true"`). 4772f8bb6.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: The swap now asserts a real handle was displaced ("block 2 must have spawned a commit task"). 5c795e03f.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: Added the `updated_attributes.is_empty()` assertion. 3681df50c.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: Added the comment: `with_local_recorder` takes a sync closure, so the test drives its own current-thread runtime. 4772f8bb6.

crates/tycho-indexer/CLAUDE.md
brunoguerios: Restored both statements — the flashblocks target-ahead clause and the partials-kept bullet — and renumbered. 9ec6cc65e.

crates/tycho-indexer/CLAUDE.md
brunoguerios: The wait is now described inside the restore step, before the broadcast bullet. 9ec6cc65e.

crates/tycho-indexer/CLAUDE.md
brunoguerios: Rewrapped to ≤100 chars with "height target" on one line. 9ec6cc65e.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: Moot: `revert_miss_cause` is deleted. The counter semantics are documented at the emission site and in CLAUDE.md. 4772f8bb6.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: Moot: the `young_component` label is gone with the attribute-level counter. 4772f8bb6.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: Renamed to `extractor_revert_attr_miss`, matching `extractor_revert_hash_miss`. 4772f8bb6.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: Rewritten: the comment says why this miss is benign (no prior value can exist, so deletion is the correct revert), without the blanket claim. 4772f8bb6.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: Renamed to `test_revert_waits_for_pending_commit` (5c795e03f); never-created component unified on `pool_ghost` (3681df50c).

crates/tycho-indexer/CLAUDE.md
brunoguerios: No longer silent: a found-component attribute miss is counted (`component_found="true"`) and covered by the summary warning; the CLAUDE.md sentence now matches. 4772f8bb6, 9ec6cc65e.

brunoguerios:
Both blockers and all other items from review 5020999719 are addressed in 3fd6e6393..9ec6cc65e. One deviation: for DESIGN #1 we took the conditional wait, extended with a write-cache flush, instead of retain-until-committed — reasoning in that thread. Retain-until-flushed is noted as a follow-up.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -1386,66 +1428,46 @@ where
...
-                                    .chain(delta.deleted_attributes.iter())
-                                    .map(move |key| (c_id, key))
-                            })
-                    })
-            })
-            .collect();
+        // Handle reverted protocol state.
+        // One chronological pass (blocks oldest → newest, txs in index order): the first
+        // event seen for a (component, attribute) pair decides its revert. First event
+        // Creation → the attr did not exist before the range, revert it as a deletion;
```
zizou0x: What if it was created and then deleted in the range? In that case we shouldn't emit a deletion for it

Might be an already existing bug, I'm not sure. If it is then it means it's handled properly right now but we should still add a todo to address this in a later PR.

crates/tycho-storage/src/postgres/cache.rs
```diff
@@ -673,6 +681,20 @@ impl CachedGateway {
         }
     }
 
+    /// Forces any staged operations in the open transaction to the store. No-op when no
+    /// transaction is open or it holds no operations.
+    pub async fn flush(&self) -> Result<(), StorageError> {
```
zizou0x: I'm not a big fan of this refactor, we already have this case and we handle it like this

https://github.com/propeller-heads/tycho/blob/0a5ea47841e816b7b960fe4bd656641dafe06af3/crates/tycho-indexer/src/extractor/protocol_extractor.rs#L2048

I'd prefer to keep it that way to keep the changes small (unless we have a reason not to?)

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -1469,6 +1491,11 @@ where
 
         trace!("Missing state keys after buffer lookup {:?}", &missing_map);
 
+        if !missing_map.is_empty() {
+            self.settle_pending_commit(&mut db_settled)
```
zizou0x: So this is in the hot path? I think it could hurt us because we are adding db commits back in the extractor processing hot path

Have you considered keeping blocks that are being commited in the buffer until they are committed? Like have a "pending_commit" section in memory for those blocks. Then here you would just get all state from memory (both buffered and pendin_commit) and look for the missing things in the db as usual. This should be safe because we know that the buffer is always more recent than the db.

crates/tycho-storage/src/postgres/cache.rs
zizou0x: But it shouldn't be needed at all if we do the in-memory suggestion

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: Nice catch. It's quite odd that we ended up in the current state, because the scenario you described (i.e. creation + deletion in the range) is exactly the problem this PR is intended to fix. Before it, it was reverting only the deletion and now it's applying it twice 😬. It was an oversight from the fixes to the edge cases. Even though this double deletion seems to be a no-op on most cases, apparently it's an issue with uniswap ticks. I'll fix this.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: Yeap, that was marked as a follow-up, but I'm ok on adding it to this PR if you think that's best. I'll fix it.

crates/tycho-indexer/CLAUDE.md
```diff
@@ -94,17 +94,30 @@ When no DCI is configured the call is a no-op.
...
 2. Pending partials are dropped only when above the target height; partials at the target
-   height are the valid prefix of the last valid block and are kept.
-3. If nothing was invalidated, only the cursor advances — no message is emitted. Otherwise
+   height are the still-valid prefix of the last valid block and are kept.
+3. Previous attribute values are restored from the buffer, then the DB. The buffer keeps
+   drained blocks in a committing section until `CachedGateway` reports their write
+   flushed, so lookups cover every block whose write has not landed and a DB miss proves
+   that no prior state exists — the revert never waits on the commit task. Attributes
+   first created inside the reverted range revert as deletions when they survive the
+   range, and revert to nothing otherwise. A same-tx create+delete merges into one
```
zizou0x: > A same-tx create+delete merges into one ambiguous delta

Why? Substreams aggregate state by tx, if the attribute is created and deleted in the same transaction we shouldn't even see it in Substreams message AFAIK

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -854,6 +863,37 @@ where
...
+/// How an attribute first appeared inside a reverted range. The first event decides the
+/// revert: `Born` attrs have no prior value, `PreExisting` attrs restore one, and
+/// `Ambiguous` attrs go through the lookup because the truth is unrecoverable.
+enum FirstSeen {
+    Born,
+    PreExisting,
+    /// The first delta held the attr in both `created_attributes` and
+    /// `deleted_attributes`: create-then-delete and delete-then-recreate inside one tx
+    /// merge to the same shape.
+    Ambiguous,
```
zizou0x: This shouldn't be possible, Substreams emits the aggregated latest state per transaction. The extractor wouldn't know how to handle this same case in many places, so this can be considered as a bug and not worth handling correctly here (we can just error if that happens).

zizou0x: Given that, I wonder if these enums, structs are still relevant? I'll let you decide when addressing this

crates/tycho-storage/src/postgres/cache.rs
```diff
@@ -586,6 +588,8 @@ impl Clone for CachedGateway {
         Self {
             // create a separate open tx state for new instances
             open_tx: Arc::new(Mutex::new(None)),
+            // each instance tracks the flush progress of its own transactions
+            flushed_block_height: Mutex::new(None),
```
zizou0x: Have you considered `AtomicU64` here? I think it would be better

zizou0x (commented):
Looks good, just have a simplification suggestion

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -1481,34 +1524,64 @@ where
...
+            .map(|state| (state.component_id.as_str(), state))
+            .collect();
+
+        // Misses below mean an upstream module emitted an Update/Deletion for an
+        // attribute that never had a Creation, or a malformed delta held the attribute in
+        // both sets and it never existed before the range. Either way no prior value
+        // exists and the revert deletes it. Substreams output is external input — a miss
+        // is a data-quality signal, not a reason to kill the extractor.
+        let mut attr_misses_state_found = 0_u64;
+        let mut attr_misses_state_missing = 0_u64;
```
zizou0x: I find these names hard to understand. 

One is when we have state but not the specific attribute and one if for when we don't have state at all, correct? I think we could either name them better or add comments explaining each case

zizou0x (approved):

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: Going to change those to `attr_misses` and `component_misses`. It should make things easier to read/understand.

PR #1333, opened 2026-08-18
chore(substreams): stop polygon-ramses-v3 ABI regeneration from dirtying every worktree

die-herdplatte:
## Problem

`protocols/substreams/polygon-ramses-v3/src/abi/{factory,pool}.rs` show up as modified in
essentially every worktree, permanently. Reproduce on `main`:

```console
$ git checkout -- protocols/substreams/polygon-ramses-v3/src/abi
$ cargo check -p polygon-ramses-v3
$ git status --short
 M protocols/substreams/polygon-ramses-v3/src/abi/factory.rs
 M protocols/substreams/polygon-ramses-v3/src/abi/pool.rs
```

It is not a formatting command — it's `build.rs`, which runs
`substreams_ethereum::Abigen(...).write_to_file("src/abi/…")` on every `cargo check`/`build`/`test`
of the package.

## Why only this package

The committed ABI came from #1208 as *nightly-rustfmt-formatted* Abigen output (packed byte
arrays). Until #1264 that was self-healing: the build wrote raw Abigen output, and the next
`cargo +nightly fmt` silently re-compacted it back to the committed shape.

#1264 (`9e0fe4924`) then added `polygon-ramses-v3/src/abi` to the `ignore` list in
`protocols/substreams/rustfmt.toml` — a reasonable response to the diffs, but it removed the
repair step, so the raw regenerated output now stays as a permanent diff. The other ignored
`*/src/abi` dirs don't drift because their committed content already equals raw Abigen output.

## Fix

Commit exactly what Abigen emits, so regeneration is a byte-for-byte no-op — bringing ramses in
line with every other ignored ABI dir. No behavioural change: pure formatting of generated code
that the build overwrites anyway.

## Verification

- `cargo check -p polygon-ramses-v3` and `cargo build -p polygon-ramses-v3
  --target wasm32-unknown-unknown` both leave the working tree clean.
- `cargo +nightly fmt --all --check` in `protocols/substreams`: 0 diffs.

### DIFF
--- a/protocols/substreams/polygon-ramses-v3/src/abi/factory.rs
+++ b/protocols/substreams/polygon-ramses-v3/src/abi/factory.rs
@@ -18,9 +18,38 @@ pub mod events {
     }
     impl PoolCreated {
         const TOPIC_ID: [u8; 32] = [
-            120u8, 60u8, 202u8, 28u8, 4u8, 18u8, 221u8, 13u8, 105u8, 94u8, 120u8, 69u8, 104u8,
-            201u8, 109u8, 162u8, 233u8, 194u8, 47u8, 249u8, 137u8, 53u8, 122u8, 46u8, 139u8, 29u8,
-            155u8, 43u8, 78u8, 107u8, 113u8, 24u8,
+            120u8,
+            60u8,
+            202u8,
+            28u8,
+            4u8,
+            18u8,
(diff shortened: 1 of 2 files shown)


brunoguerios (approved):
Thanks!

PR #1335, opened 2026-08-18
feat(simulation): select venues that execute via the PropAMMRouter

tamaralipows:
Stacked on #1278. Turns the pAMM fallback on: nothing emitted `propammfallback:` before this.

## Change

Venue selection for the pAMM fallback. The price level stream builder emits whitelisted venues' components under the `propammfallback:` family instead of `pricelevelstream:`, so tycho-execution routes their swaps through Titan's PropAMMRouter and a stale maker quote falls back to a single-hop Uniswap V3 pool instead of reverting the route.

This is the default. `without_fallback_router()` is the opt-out and keeps every venue on the direct path.

The stream reads the router's on-chain whitelist itself, through the node at `RPC_URL`, once on the first poll: the set is governance-gated and changes rarely, and renaming a running component's protocol system would churn every consumer's component set. Without `RPC_URL`, or when the read fails, it warns and keeps every venue on the direct path.

`build()` stays sync. The read happens in the stream, next to the lazy WebSocket connect, so no call site changes.

The check is by address, so it also covers auto-detected venues. Component ids and static attributes are unchanged.

The `price-level-stream` feature pulls in alloy and reqwest, so `fallback_router` no longer needs the `evm` feature.

## Why the whitelist must gate the family

A venue outside the whitelist reverts `UnknownVenue` on the router, which fires the catch arm: every swap executes on the Uniswap V3 fallback at a worse price than the venue gives.

## tycho-integration-test opts out

That test measures how well a venue's own quote matches its own fill. Through the router a stale quote executes on the Uniswap V3 fallback instead of reverting, which reads as a quote mismatch and hides the staleness its revert classification records.

## Client impact

No wire or schema change, no new endpoint, no change to how the stream is constructed, and the Titan stream connection is unchanged.

Whitelisted venues now arrive as `propammfallback:`. A consumer must run tycho-execution with #1278 (encoding the family fails otherwise) and match on the new prefix wherever it matches on `pricelevelstream:` names (Fynd does in propeller-heads/fynd#402). `without_fallback_router()` emits the same components as before, byte for byte.

Sequencing: land after or with #1278 and Fynd #402.

## Open

- Gas: the router path costs more than the direct venue call. Encoding accounts for it, but the stream's per-venue gas estimate is unchanged here, so route ranking still uses the direct-call estimate.
- `vm:fermiswap` has the same stale-quote problem and does not pass through this check. Execution-side follow-up.
- GitBook docs for the price level stream and this family: separate PR.


crates/tycho-simulation/src/price_level_stream/stream.rs
```diff
@@ -121,6 +122,47 @@ impl PriceLevelStreamBuilder {
...
+    ///
+    /// Reads the router's on-chain venue whitelist once, via the node at `RPC_URL` (from the
+    /// environment, falling back to `.env`). Only whitelisted venues may use the family — the
+    /// router reverts `UnknownVenue` for others, so every swap would execute on the Uniswap V3
+    /// fallback at a worse price than the venue gives.
+    ///
+    /// Degrades instead of failing: without `RPC_URL`, or when the whitelist read fails, a
+    /// warning is logged and every venue stays on the direct `pricelevelstream:` path.
+    #[cfg(feature = "evm")]
+    pub async fn via_fallback_router(mut self) -> Self {
```
tamaralipows: Should this actually be opt-in? Or just default behaviour?

tvinagre: default. Avoiding the fallback should be opt-out

crates/tycho-simulation/src/price_level_stream/stream.rs
```diff
@@ -249,7 +303,11 @@ impl SnapshotTracker {
                 let component = previous
                     .remove(&id_string)
                     .unwrap_or_else(|| {
-                        let component = build_component(&self.tokens, config, id, &token0, &token1);
+                        let via_router = self
```
tvinagre, line 436: This check covers only components this stream emits — the `pricelevelstream:`/`propammfallback:` families. `vm:fermiswap`, the separate indexed-path implementation of the same venue (executed via `FermiSwapExecutor` → the FermiSwapper at `0xb1076fE3…`), has the exact same stale-maker-quote problem: a swap simulated against a mined block reverts because the venue only executes in blocks Titan built. It never passes through this check, so it will never get the PropAMMRouter fallback.

Capturing the follow-up here since it's execution-side work, not this PR: covering `vm:fermiswap` needs (1) venue-aware calldata — `FermiSwapEncoder` currently emits 40 bytes `(tokenIn, tokenOut)` while `PropAMMFallbackExecutor._decodeData` requires 60 bytes `(venue, tokenIn, tokenOut)`; (2) routing the family to `PropAMMFallbackExecutor` in the execution config; (3) the venue to route to is the pAMM itself (`0x59794589…`), which is already on the PropAMMRouter whitelist — note it is NOT the FermiSwapper address that `FermiSwapExecutor` calls. Worth a follow-up issue so it doesn't get lost.

crates/tycho-simulation/src/price_level_stream/stream.rs
tvinagre: To make the default flip concrete: attempt the whitelist read from `build()` (not only when a consumer calls `via_fallback_router()`), and add an explicit opt-out (e.g. `without_fallback_router()`) that skips the read and keeps every venue on the direct path. The existing degrade-on-missing-`RPC_URL` behavior makes this safe as a default — consumers without `RPC_URL` just get the warn + direct path. Sequencing note: the flip should land together with or after the execution side (#1278 + Fynd #402), since consumers will start receiving `propammfallback:` components they cannot encode without it.

crates/tycho-simulation/CLAUDE.md
```diff
@@ -27,9 +27,12 @@ for any protocol indexed by Tycho.
   identified as `pricelevelstream:{pamm}`. A new builder serves nothing: `with_default_pamms`
   registers the known venues, `add_pamm` individual ones (overriding per address), and opt-in
   auto-detection additionally serves unknown venues under their address
-  (`pricelevelstream:{0xaddress}`). Venues may overlap with other integration
-  paths of the same liquidity (e.g. `vm:fermiswap`) — consumers must deduplicate by venue where
-  double-counting matters
+  (`pricelevelstream:{0xaddress}`). Opting in with `via_fallback_router` (`evm` feature) emits
```
tvinagre: Docs should go to the GitBook docs, not CLAUDE.md — CLAUDE.md is contributor/agent context, while docs/for-solvers/ is the user-facing surface. Neither `docs/for-solvers/simulation.md` nor `supported-protocols.md` currently mentions the price level stream or the `propammfallback:` family at all. With the default flip, the GitBook docs should describe the fallback as default-on with the opt-out escape hatch explicit (the mechanism reads the whitelist at startup and stays on the direct path when `RPC_URL` is missing or the read fails), state that whitelist coverage is automatic by address (Kipseli and FermiSwap are already whitelisted on the PropAMMRouter; newly whitelisted venues are picked up at startup), and note that consumers must run tycho-execution with #1278 (and Fynd #402) to encode `propammfallback:` components.

crates/tycho-simulation/src/price_level_stream/config.rs
```diff
@@ -11,6 +11,13 @@ use tycho_common::Bytes;
 /// details.
 pub const PRICE_LEVEL_STREAM_FAMILY: &str = "pricelevelstream";
 
+/// Protocol system family of components executed through Titan's PropAMMRouter instead of the
+/// venue directly, so a stale maker quote falls back to a single-hop Uniswap V3 pool instead of
+/// reverting the route.
+///
+/// Must match `tycho-execution`'s `PROPAMM_FALLBACK_KEY`.
+pub const PROPAMM_FALLBACK_FAMILY: &str = "propammfallback";
```
tvinagre, line 19: Duplicated protocol-family and router-address literals across crates/domains with no enforcement. The family key (`propammfallback`) now exists in three places (this const, tycho-execution's `PROPAMM_FALLBACK_KEY`, and the executor deployment config keys) and the router address in two (here and `PropAMMFallbackExecutor.PROPAMM_ROUTER` in Solidity). If either drifts: family mismatch makes tycho-execution's `get_encoder` fail to resolve `propammfallback:{venue}` (every route through that venue fails to encode), and an address mismatch reads the whitelist from a different contract than the executor calls (silent disagreement between the emitted family and the executed router). No test asserts equality between the crates. A compile-time or unit cross-check (e.g. a tycho-execution test asserting `PROPAMM_FALLBACK_KEY` equals the value the simulation emits, or a single shared source for the family name) would close this.

crates/tycho-simulation/src/price_level_stream/stream.rs
```diff
@@ -249,7 +303,11 @@ impl SnapshotTracker {
...
+                            .router_venues
```
tvinagre, line 437: Read-once whitelist with no staleness handling: a venue removed from the on-chain whitelist after startup keeps being emitted under `propammfallback:` until process restart, and every swap on it then reverts `UnknownVenue` — the exact failure mode this feature exists to prevent, applied in the opposite direction. There is no re-read, TTL, or error path. The code documents this as a deliberate tradeoff (whitelist is governance-gated, changes rarely), but the removal direction is strictly worse than the direct path (which at least executes when quotes are fresh) and is untested. Related: component ids are name-independent (component_id = pamm++token0++token1), so across a restart a venue flipped between families keeps the same id with a different protocol_system; consumers that cache components by id keep serving the stale family unless the re-announcement is honored. At minimum add a test of tracker behavior when membership excludes a previously-routed venue, and consider re-reading the whitelist periodically or documenting the read-once caveat for operators.

crates/tycho-simulation/src/price_level_stream/stream.rs
```diff
@@ -121,6 +122,47 @@ impl PriceLevelStreamBuilder {
...
+            });
+        let Some(rpc_url) = rpc_url else {
+            tracing::warn!(
+                "RPC_URL is not set; pAMM swaps execute on the venues directly, without the \
+                 PropAMMRouter's Uniswap V3 fallback"
+            );
+            return self;
+        };
+
+        match fetch_fallback_router_venues(&rpc_url).await {
```
tvinagre: The builder's degrade paths are untested. `via_fallback_router` has three branches (no RPC_URL -> warn + direct path; fetch error -> warn + direct path; success -> populate) and none is exercised: the fetch function only has the invalid-URL unit test and an ignored network test. The warn-and-continue behavior is the feature's safety guarantee per the docs, so the untested branches are exactly the ones that run in any misconfigured deployment. All of them are testable without a network (env manipulation + injected failure) — worth covering.

crates/tycho-simulation/src/price_level_stream/stream.rs
tvinagre: The selection feature is inert: nothing in the repo calls `via_fallback_router` (search returns only the definition, its docs, and CLAUDE.md; the integration-test processor builds the stream with `with_default_pamms().auto_detect(true)` only). The execution side (base) is fully wired, but nothing enables the simulation side, so the PR's stated behavior cannot be exercised end-to-end yet. Additionally, when the family is eventually enabled, tycho-integration-test's stale-quote classification (`starts_with(PRICE_LEVEL_STREAM_PREFIX)`) will not classify `propammfallback:*` executions: with the router, a stale maker quote no longer reverts (it falls back to the V3 leg), so revert-based staleness metrics and the CLAUDE.md claim that stale-update reverts are recorded silently stop meaning what the docs say for those venues. Worth a deliberate decision in the wiring change that enables the opt-in: extend the classification or document the semantic change.

tamaralipows: Changed in c7a837c7e. `build` reads the whitelist and serves the venues on it under `propammfallback:`; `without_fallback_router()` skips the read and keeps every venue on the direct path.

`build` stays sync: 784d2fa63 moves the read into the stream, next to the lazy WebSocket connect, so it happens on the first poll and no call site changes.

crates/tycho-simulation/src/price_level_stream/stream.rs
tamaralipows: Agreed, and out of scope here — the three items you list are all execution-side. I will file a follow-up issue for `vm:fermiswap`.

crates/tycho-simulation/CLAUDE.md
tamaralipows: Leaving the GitBook docs out of this PR — I want to see the fallback work end to end first. Separate PR, separate author.

CLAUDE.md still describes the new default, since that is contributor context.

crates/tycho-simulation/src/price_level_stream/config.rs
tamaralipows: Added two checks in c7a837c7e:

- `families_match_the_execution_side_prefixes` (tycho-simulation) asserts `PRICE_LEVEL_STREAM_FAMILY` and `PROPAMM_FALLBACK_FAMILY` equal tycho-execution's exported prefixes. tycho-execution is already a dev-dependency, so no new coupling. `PROPAMM_FALLBACK_PREFIX` is now re-exported from `encoding::evm` next to `PRICE_LEVEL_STREAM_PREFIX`.
- `test_family_keys_and_prefixes_agree` (tycho-execution) asserts `{KEY}:` equals the matching prefix, which closes the KEY-to-PREFIX half.
- `test_router_address_matches_the_executor` reads `PropAMMFallbackExecutor.sol` and asserts it contains `FALLBACK_ROUTER_ADDRESS`.

crates/tycho-simulation/src/price_level_stream/stream.rs
tamaralipows: The whitelist is still read once and never re-read, and `build`'s rustdoc now says so: restart to pick up a change. Operators get that from the rustdoc.

Added `unwhitelisted_venue_keeps_the_direct_family` to cover the membership-excludes direction.

On the in-process transition: the set is fixed for the lifetime of the stream, so a venue cannot change family while running. A venue that vanishes and reappears rebuilds against the same set. The removal risk is real but bounded to a governance action plus no restart, and a re-read would trade it for churn on every consumer's component set. Happy to add a TTL if you think the tradeoff is wrong.

crates/tycho-simulation/src/price_level_stream/stream.rs
tamaralipows: Covered in c7a837c7e. Split the env lookup out of the fetch so both degrade paths test without a network or env manipulation:

- `missing_rpc_url_leaves_every_venue_on_the_direct_path` — `fetch_router_venues(None)`
- `failed_whitelist_read_leaves_every_venue_on_the_direct_path` — invalid URL
- `fallback_router_is_on_unless_opted_out` — the new default and the opt-out

crates/tycho-simulation/src/price_level_stream/stream.rs
tamaralipows: `build()` now serves whitelisted venues under `propammfallback:` by default, so there is nothing left to enable.

On the classification: tycho-integration-test calls `without_fallback_router()`. That test measures how well a venue's own quote matches its own fill, and through the router a stale quote executes on the Uniswap V3 fallback instead of reverting — it would read as a quote mismatch and hide the staleness the revert classification records. Comment in the processor says so.

crates/tycho-simulation/src/price_level_stream/stream.rs
tamaralipows: Filed as ENG-6335 under Tycho Maintenance Triage: https://propeller-heads.atlassian.net/browse/ENG-6335

tamaralipows: Implemented in #1344 (stacked on #1278), so ENG-6335 is covered.

Your three items, in order:
1. Venue-aware calldata — `PropAMMSwapEncoder` falls back to a configured venue address when a component carries no `pamm_address` attribute.
2. Routing — the caller labels the swap `propammfallback:fermiswap`; `vm:fermiswap` itself is untouched and still executes on `FermiSwapExecutor`. Opt-in rather than a silent reroute: `vm:fermiswap` is a separate integration with its own executor and component ids, so moving its swaps to the router path, and paying the extra hop, is the caller's call per swap.
3. Venue — `0x5979458912F80B96d30D4220af8E2e4925A33320` in `protocol_specific_addresses.json`, not the FermiSwapper. A fork test pins it: a wrong address reverts `UnknownVenue`.

Also found while testing: `test_executor_addresses.json` named the wrong address for `propammfallback`, so Rust-encoded calldata for the family reverted `Dispatcher__UnapprovedExecutor`. Fixed in #1278 (437d3513c).

crates/tycho-integration-test/src/stream_processor/price_level_stream_processor.rs
```diff
@@ -58,10 +58,16 @@ impl PriceLevelStreamProcessor {
...
+        //
+        // The PropAMMRouter path is off here: this test measures how well a venue's own quote
+        // matches its own fill. Through the router a stale quote silently executes on the
+        // Uniswap V3 fallback instead of reverting, which reads as a quote mismatch and hides
+        // the staleness the revert classification below records.
         let stream = PriceLevelStreamBuilder::new()
             .with_known_pamms()
             .auto_detect(true)
             .with_tokens(all_tokens.clone())
+            .without_fallback_router()
```
tvinagre, line 70: Nice, but we need to add the overrides to the integration test simulation, if we haven't done it yet

tvinagre (approved):

PR #1337, opened 2026-08-18
feat: support Ekubo V3 Ve33 pools

die-herdplatte:
Adds support for Ekubo V3 Ve33 pools, which are live on Robinhood Chain, across indexing, simulation, and execution. Supersedes #1223, which could not be updated: its head branch lives in an org-owned fork, and GitHub does not allow maintainer pushes to those. Includes all commits from that PR unchanged, plus review fixes and the Robinhood rollout pieces.

Robinhood CI access is in place: #1410 wires `ROBINHOOD_RPC_URL` into `ci-foundry.yaml` and the substreams integration harness from the `ROBINHOOD_RPC_URL_ARCHIVE` secret, which is configured. It has to stay an **archive** endpoint — the public `rpc.mainnet.chain.robinhood.com` prunes state and can serve neither `testVe33Swap`'s pinned fork block nor the integration test's block range.

## Substreams (`ethereum-ekubo-v3` v0.2.3)

- Track Ve33 pool swap-fee changes so off-chain quotes stay accurate. Ve33 pools emit a `swap_fee` entity attribute; fee updates come from the extension's events.
- The Ve33 extension address is chain-specific (its constructor takes a chain-specific stake token, so the CREATE2 address differs per chain). It is passed via module params as a query string (`ve33_address=0x...`, serde_qs like the other packages); empty params disable Ve33 handling.
- New `robinhood-ekubo-v3.yaml` manifest with the deployed Robinhood Ve33 address in the params, starting at the chain's first pool creation (block 42961), plus `integration_test_robinhood_ekubo_v3.tycho.yaml` covering the first Ve33 pool (run with `--chain robinhood`).

## Simulation

- New Ve33 pool type wrapping the underlying concentrated/full-range pool via `enum_delegate`, applying the voted swap fee on quotes.
- Ve33 extension detection is gated on `Chain::Robinhood` and the deployed address. SignedExclusiveSwap keeps a single ungated constant: it was deployed through the canonical cross-chain CREATE2 deployer with chain-invariant constructor args, so any future deployment lands at the same address.

## Execution

- `EkuboV3Executor` handles the address-deterministic extensions (MEVCapture, SignedExclusiveSwap) on any chain and exposes virtual `_swapHop`/`_hopEnd` hooks; `EkuboV3RobinhoodExecutor` overrides them to add the deployment-specific Ve33 extension via `Core.forward`. Both live under `src/executors/ekubo_v3/`. A Robinhood executor therefore also serves SES pools if that extension is later deployed there.
- `executor_deployments.json` and the deploy script gain the `robinhood` → `EkuboV3RobinhoodExecutor` entry.
- `testVe33Swap` runs on a Robinhood Chain fork pinned to block 39875978, swapping through the deployed Ve33 extension with the WETH/USDG pair of a live Ve33 pool. The `robinhood` rpc endpoint is wired in `foundry.toml`.
- The signed-swap tests are part of `EkuboV3ExecutorTest`; `EkuboV3RobinhoodExecutorTest` covers the Ve33 path on a live Robinhood fork.

## Merge with main

`origin/main` is merged in: the deployed SignedExclusiveSwap address replaces the placeholder in the split executors, the simulation keeps main's TWAMM V1/V2 split alongside the Ve33 address, and `calldata.txt` plus the `EkuboV3.runtime.json` fixture are regenerated from the merged code. A later merge resolves the `CLAUDE.md` conflicts against #1407's documentation resync, keeping main's version throughout — it already covers Ekubo V3 — and restoring the runtime-bytecode regeneration notes that resync did not carry.

## Harness (`protocols/testing`)

The Robinhood integration test could not actually run, and the Ethereum one had been validating far less than it looked like. Four fixes, each standing on its own:

| Commit | Problem | Fix |
|---|---|---|
| `fix(testing): keep the generic RPC URL across multi-package runs` | The per-protocol loop exports `RPC_URL`, and `get_rpc_url`'s fallback arms read it back, so `PROTOCOLS="robinhood-ekubo-v3 ethereum-ekubo-v3"` ran the Ethereum package against the Robinhood RPC | Capture the generic endpoint once, before the loop, and resolve fallbacks against that copy |
| `feat(testing): decode ekubo_v3 with its native state` | No `ekubo_v3` arm in `register_protocol`, so it fell through to the `EVMPoolState` default and failed with `Adapter for protocol ekubo_v3 not found`. `skip_state_decode_failures` swallowed that, leaving zero states — simulation and execution silently iterated over nothing and the test still passed | Register `EkuboV3State` with the default extension filter (SES pools excluded: their swaps need a per-swap signature the harness cannot produce) |
| `feat(testing): pick the executor for the chain under test` | `EXECUTOR_MAPPING` is keyed by protocol system alone, so Robinhood got the chain-agnostic `EkuboV3Executor`, which calls `Core.swap` directly — every Ve33 swap reverted with `SwapMustHappenThroughForward()` | Look up `(chain, protocol system)` first, falling back to the protocol-only map; add the `EkuboV3Robinhood` fixture |
| `feat(testing): make robinhood-ekubo-v3 testable` | The Robinhood manifest lives inside the `ethereum-ekubo-v3` package, so the runner could not find its substreams directory | Clone-to-base mapping in `test_runner.rs` and both `run.Dockerfile` tables |

The decoder registration is why `ethereum-ekubo-v3` gains seven executing tests here: its execution path had never run.

## Testing

- All 9 Foundry `EkuboV3.t.sol` tests pass, including the Ve33 swap against the pinned Robinhood fork; 555 in the full Foundry suite.
- Full `tycho-execution` test suite (223 tests) and the simulation `ekubo_v3` tests pass; clippy, nightly fmt and `update_runtime_bytecode.sh --check` clean.
- `protocol-testing` harness, both packages in one run: `robinhood-ekubo-v3` `Passed 1/1` and `ethereum-ekubo-v3` `Passed 8/8`, every execution batch 6 successes / 0 failures. The Robinhood test validates indexing, simulation and execution.
- Substreams verified against live endpoints: the Robinhood manifest emits the first Ve33 pool with the params-gated `swap_fee` attribute; the Ethereum manifest (empty Ve33 params, SES tagging) still emits `is_exclusive` for signed pools.


zach030 (approved):
LGTM! Can you add this to the protocol tests so we can test?

PR #1339, opened 2026-08-19
fix: refresh token cache last_traded from balance deltas

brunoguerios:
## Problem

The `rpc` command serves `/v1/tokens` from the in-memory token cache (#1302), but its `last_traded` timestamps freeze at the startup snapshot. Tokens first traded after boot stay excluded from every "traded within N days" filter, and the gap grows with pod uptime. Raised three times in the #1302 review. ENG-6330.

## Root cause

`last_traded` converges through write-through on balance writes only. The delta refresh polls `token.modified_ts`, and a balance write never touches the token row, so no balance signal crosses process boundaries. A process that serves queries without indexing balances gets no updates at all.

## Fix

`refresh` now runs two polls per tick behind two sync markers:

1. `refresh_tokens` — the existing `token.modified_ts` poll.
2. `refresh_balances` — newest `component_balance_default.valid_from` per token, chain-filtered, fed into the idempotent `update_last_traded`.

Both use the existing 600s overlap window and advance their marker only on success. The token poll runs first, so a token and its first trade arriving in the same window apply in order.

A new migration indexes `component_balance_default (valid_from)`, created `CONCURRENTLY` — the table is large on Base and on the extractor's hot write path.

## Notes for reviewers

- The index migration is the repo's first non-transactional migration: `metadata.toml` with `run_in_transaction = false`. `embed_migrations!` and the diesel CLI both honor it.
- Ship the migration before the poll; without the index the poll scans the default partition.
- The module docs lose the `last_traded` known-limit paragraph. `enable_token_cache()` stays on  for `rpc`.

### DIFF
--- a/crates/tycho-storage/CLAUDE.md
+++ b/crates/tycho-storage/CLAUDE.md
@@ -39,8 +39,9 @@ and `apply_versioning()` utility consumed by `contract` and `protocol`.
 
 `token_cache` holds an in-memory copy of the token tables so `get_tokens` never touches SQL.
 Opt-in via `GatewayBuilder::enable_token_cache()` (the `index` and `rpc` commands enable it;
-the token-analysis job does not). Kept fresh by write-through on token/balance writes plus a
-periodic `modified_ts` delta poll for out-of-process writers. See the module docs for design.
+the token-analysis job does not). Kept fresh by write-through on token/balance writes plus
+periodic delta polls (`token.modified_ts` and `component_balance_default.valid_from`) for
+out-of-process writers. See the module docs for design.
 
 ## Write Order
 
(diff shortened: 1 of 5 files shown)


zizou0x:
I didn't get it, isn't it updated live by https://github.com/propeller-heads/tycho/pull/1302/changes#diff-d0c7eb13bbba8004947c381f95fb16086ea66bbbe90004d63815797899bf7e7eR1426 ?

From my review https://github.com/propeller-heads/tycho/pull/1302#discussion_r3757240335 it's only affecting when we use RPC only mode, which we almost never use

brunoguerios:
> I didn't get it, isn't it updated live by https://github.com/propeller-heads/tycho/pull/1302/changes#diff-d0c7eb13bbba8004947c381f95fb16086ea66bbbe90004d63815797899bf7e7eR1426 ?
> 
> From my review [#1302 (comment)](https://github.com/propeller-heads/tycho/pull/1302#discussion_r3757240335) it's only affecting when we use RPC only mode, which we almost never use

Yes, you're right that this affects RPC only. Problem here is that cache was enabled in the RPC endpoint. We must either revert that or make sure the cache is serving correct/up-to-date data (which is the fix from this PR).
Do we have an expectation on the future usage of this? Intuitively I'd say it's more future proof to have this fixed and serving cached data. And the cost doesn't seem to be significant (i.e. one added index).
Let me know if you disagree.

PR #1341, opened 2026-08-19
fix: apply token cache write-through only after DB commit

brunoguerios:
## Problem

A write transaction that rolls back left the token cache ahead of the database. The cache then served phantom tokens that were never persisted, and `last_traded` timestamps advanced past any trade the database records. No refresh cycle removes either: the delta poll only ever inserts tokens and moves timestamps forward.

Found in review of #1302, not from a production incident. Today a write failure kills the process, so a restart reloads the cache and masks the divergence.

## Root cause

The three write-through sites in `protocol.rs` (`add_tokens`, `update_tokens`, `add_component_balances`) mutate the cache inline, while `DBCacheWriteExecutor::write` still holds the enclosing DB transaction open. The cache mutation is not part of that transaction, so a rollback undoes the SQL and keeps the cache mutation.

The extractor supervisor (#1026) removes the crash that hid this. Once the process survives a failed write, the divergence persists for the lifetime of the process.

## Fix

`TokenCache` gains a `CacheOp` staging buffer. The write-through sites now call `write_through_*` entry points, which buffer while a staging session is open and apply immediately otherwise. `DBCacheWriteExecutor::write` opens a session per transaction attempt and commits or discards the buffer based on the result, so a discarded retry cannot apply twice.

`add_tokens`, `upsert_tokens`, and `update_last_traded` are now private to `token_cache.rs`. The compiler rejects any future external call site that would bypass staging.

## Notes for reviewers

- Staged ops apply in push order on commit. `test_retry_applies_staged_ops_exactly_once_and_in_order` pins this.
- The refresh polls keep calling the direct methods on purpose: they read committed rows, so staging their results would be wrong.
- Without a staging session, mutations apply at once. This preserves `DirectGateway` semantics, guarded by `test_serial_db_cache_write_through`.

ENG-6331

### DIFF
--- a/crates/tycho-storage/src/postgres/cache.rs
+++ b/crates/tycho-storage/src/postgres/cache.rs
@@ -371,7 +371,14 @@ impl DBCacheWriteExecutor {
         let mut res =
             Err(PostgresError(StorageError::Unexpected("default response error".to_string())));
 
+        let token_cache = self.state_gateway.token_cache.clone();
+
         while retry_count < max_retries {
+            // Each attempt gets a fresh staging session; a failed attempt's
+            // leftovers are replaced by the next begin.
+            if let Some(cache) = &token_cache {
+                cache.begin_staging();
+            }
             res = conn
                 .build_transaction()
                 .repeatable_read()
@@ -419,6 +426,13 @@ impl DBCacheWriteExecutor {
                     }
                 }
                 _ => break,
+            }
+        }
+
+        if let Some(cache) = &token_cache {
+            match &res {
+                Ok(_) => cache.commit_staged(),
+                Err(_) => cache.discard_staged(),
             }
         }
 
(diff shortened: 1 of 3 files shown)


zizou0x:
I don't think this is really relevant, have you seen https://github.com/propeller-heads/tycho/pull/1302#discussion_r3758617208?

I think worst case that can happen is the cache contains a token that is real but not on the db yet (basically the cache would be ahead of the db but still correct) which is fine IMO.

brunoguerios:
> I don't think this is really relevant, have you seen [#1302 (comment)](https://github.com/propeller-heads/tycho/pull/1302#discussion_r3758617208)?
> 
> I think worst case that can happen is the cache contains a token that is real but not on the db yet (basically the cache would be ahead of the db but still correct) which is fine IMO.

Agreed. Closing as not needed.

PR #1342, opened 2026-08-19
fix: harden the token cache against bulk backfills and lock poisoning

brunoguerios:
Stacked on #1339 (ENG-6330); merge that first. Retarget to `main` after it lands.

Five hardening items on the token cache, all follow-ups from #1339:

1. **Pagination in the delta polls.** Both refresh polls page along an id cursor at 50k rows per batch and apply each batch before they fetch the next one. Before, one tick read the whole lookback window into one `Vec` under one write lock. Behavior at the default batch size is unchanged.
2. **Recovery from poisoned locks.** A panic inside a write guard no longer kills the process. `ChainTokenStore::upsert` is reordered so an interrupted writer cannot leave a dangling position, and every lock site then recovers the guard instead of calling `.expect()`.
3. **Retries on the initial load.** Three attempts with exponential backoff, per batch on the same connection, plus one more for the whole load on a fresh pooled connection.
4. **`WithTotal.total` is now `Option<u64>`.** It is a row count. `PaginationResponse.total` stays `i64` because it is the wire type, and converts once at each `PaginationResponse::new` call in `services/rpc.rs`.
5. **Migration rule documented.** An index on a hot-path table needs `CREATE INDEX CONCURRENTLY` and `run_in_transaction = false`.

Item 3 is verified by inspection, not by a test. Failure injection needs a mockable async-diesel connection, which is more scaffolding than the 20 lines it covers.

## Follow-ups

- ENG-6331 (staged write-through): its `staged: Mutex<...>` needs the same poison recovery when it lands.
- ENG-6332 (single-chain refactor / loader split): per-batch reconnection lands naturally there if the whole-load retry proves too coarse.
- `PaginationResponse.total` → `u64` at the next breaking DTO release.

ENG-6333

### DIFF
--- a/crates/tycho-common/src/storage.rs
+++ b/crates/tycho-common/src/storage.rs
@@ -262,7 +262,7 @@ impl Version {
 #[derive(Debug)]
 pub struct WithTotal<T> {
     pub entity: T,
-    pub total: Option<i64>,
+    pub total: Option<u64>,
 }
 
 /// Store and retrieve protocol related structs.
--- a/crates/tycho-storage/CLAUDE.md
+++ b/crates/tycho-storage/CLAUDE.md
@@ -58,3 +58,13 @@ out-of-process writers. See the module docs for design.
 Every mutable entity carries `valid_from` / `valid_to` timestamps enabling time-travel
 queries. `versioning::apply_versioning()` sets `valid_to` on the previous row when a new
 version is inserted. Historical rows are never mutated.
+
+## Migrations
+
+An index on a table the extractor writes on the hot path (`token`, the
+`component_balance` partitions, `protocol_state`) must be created with
+`CREATE INDEX CONCURRENTLY`, in a migration whose `metadata.toml` sets
+`run_in_transaction = false` — a plain `CREATE INDEX` takes a SHARE lock and
+stalls indexing for the whole build. See
+`migrations/2026-08-19_component_balance_default_valid_from_index/` for the
+pattern.
--- a/crates/tycho-storage/src/postgres/token_cache.rs
+++ b/crates/tycho-storage/src/postgres/token_cache.rs
@@ -108,6 +108,10 @@ const LOAD_BATCH_SIZE: i64 = 500_000;
 /// load batch so readers interleave with the apply loop during a bulk backfill.
 const REFRESH_BATCH_SIZE: i64 = 50_000;
 
+/// How many times the initial load tries a query before giving up. Applies both
+/// per batch and to the whole load, with exponential backoff between attempts.
+const LOAD_RETRY_ATTEMPTS: u32 = 3;
+
 /// Timestamp value for tokens that never appeared in a component balance.
 /// `i64::MIN` sorts below any real threshold, matching the SQL `EXISTS` filter
 /// which excludes such tokens.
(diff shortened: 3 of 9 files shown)


crates/tycho-common/src/storage.rs
```diff
@@ -262,7 +262,7 @@ impl Version {
 #[derive(Debug)]
 pub struct WithTotal<T> {
     pub entity: T,
-    pub total: Option<i64>,
+    pub total: Option<u64>,
```
zizou0x, line 265: What's the motivation here? Isn't this a RPC interface breaking change that would affect clients?

crates/tycho-storage/CLAUDE.md
```diff
@@ -58,3 +58,13 @@ out-of-process writers. See the module docs for design.
 Every mutable entity carries `valid_from` / `valid_to` timestamps enabling time-travel
 queries. `versioning::apply_versioning()` sets `valid_to` on the previous row when a new
 version is inserted. Historical rows are never mutated.
+
+## Migrations
+
+An index on a table the extractor writes on the hot path (`token`, the
+`component_balance` partitions, `protocol_state`) must be created with
+`CREATE INDEX CONCURRENTLY`, in a migration whose `metadata.toml` sets
+`run_in_transaction = false` — a plain `CREATE INDEX` takes a SHARE lock and
```
zizou0x, line 67: Don't you think this is unsafe? what if the migration fails for a reason? Wouldn't we end with with an unnoticed non-fully applied migrations on the db?

crates/tycho-storage/src/postgres/token_cache.rs
```diff
@@ -1031,7 +1163,7 @@ mod benchmark {
             })
             .expect("query failed")
             .total
-            .unwrap();
+            .unwrap() as i64;
```
zizou0x, line 1166: I really dislike all these back and forth between u64 and i64

crates/tycho-storage/src/postgres/token_cache.rs
```diff
@@ -700,6 +788,25 @@ impl TokenCache {
...
+    })
+}
+
+/// See [`read_recovered`].
+fn write_recovered<T>(lock: &RwLock<T>) -> RwLockWriteGuard<'_, T> {
+    lock.write().unwrap_or_else(|poisoned| {
+        error!("Token cache lock poisoned; recovering");
+        poisoned.into_inner()
+    })
+}
```
zizou0x, line 808: What's the motivation here? Have we seen some poisoned locks?

crates/tycho-common/src/storage.rs
brunoguerios: The goal here is to be coherent with the fact that this count can't be negative, so it should be an unsigned type.
`WithTotal` is not on the wire — it's the storage trait return type. The wire type `PaginationResponse.total` stays `i64`, so RPC clients see no change. It does break the Rust API of the storage traits though.

crates/tycho-storage/src/postgres/token_cache.rs
brunoguerios: This is more like a preventive fix suggested by Claude. It's not something we've seen no.
Felt like a small improvement that would make the code more robust though.

For context, the failure mode was: one panic inside a write guard poisons the lock, then every token query panics until restart, because all sites used .expect().

If you still think it's better to not have it, I can revert the change.

crates/tycho-storage/src/postgres/token_cache.rs
brunoguerios: Yeah, I don't particularly like this as well. I'll push a fix for that reducing the cast to a single place so it's not spread out through the code.

crates/tycho-storage/CLAUDE.md
brunoguerios: Yeah, this does come with a trade-off. An intermediary invalid index can happen indeed. Here's the full context for us to decide how to move forward:

**Why `CONCURRENTLY`:** a plain `CREATE INDEX` holds a SHARE lock for the whole build, which blocks all writes to the table. `component_balance_default` is on the extractor's hot write path, so every such migration would stall indexing for the full build. That lock is the entire motivation.

**Cost:** `CONCURRENTLY` can't run in a transaction (Postgres rejects it — `run_in_transaction = false` is forced, not optional), so there's no rollback. A mid-build failure leaves an INVALID index shell, and because the migration uses `IF NOT EXISTS`, the rerun skips the shell and records the migration as applied. So yes — a partially-applied end state is reachable.

**Scoping the risk:** it's never a correctness issue. The planner ignores INVALID indexes; queries stay correct, data is untouched. The damage is the missing index (slow poll) plus write overhead maintaining the shell. And the first failure is loud — the migration errors and the deploy fails; only the rerun is silent.

**`CREATE INDEX`** can never leave an invalid state (full rollback), but buys that with the write stall above. For small/cold tables it remains the right default — the rule is only for hot-path tables.

**Mitigation:** detection is a one-liner against `pg_index.indisvalid`; recovery is `REINDEX INDEX CONCURRENTLY` (or drop + re-create). We can't self-heal inside the migration file — diesel batches the file, and `CREATE INDEX CONCURRENTLY` must be the sole statement.

**Proposal:** document the runbook now, and add a post-migration startup check that logs an error when any invalid index exists.

brunoguerios (commented):
As a general statement, these were some non-critical improvement suggestions from the original PR.
I followed-up on your comments so we can decide on how to move forward on each one.

PR #1344, opened 2026-08-20
feat(execution): resolve vm:fermiswap to PropAMMFallbackExecutor

tamaralipows:
Stacked on #1278. Follow-up to https://github.com/propeller-heads/tycho/pull/1335#discussion_r3815579675, filed as ENG-6335.

## Problem

The indexed `vm:fermiswap` path has the same stale-quote problem #1278 fixes for the price level stream: a swap simulated against a mined block reverts `StaleUpdate`. The venue needs the maker's oracle update for the block it executes in to be on-chain, and a mined block we simulate against carries none.

That path could not use the PropAMMRouter at all, because a `vm:fermiswap` component names only the FermiSwapper contract (`0xb1076fE3AB5e28005C7c323Bac5AC06a680d452e`), never the pAMM the router whitelists (`0x5979458912F80B96d30D4220af8E2e4925A33320`), and the executor needs the venue in its swap data.

## Change

`vm:fermiswap` swaps reach the venue through the PropAMMRouter, so a stale quote retries on Uniswap V3. The venue address comes from configuration, since the component does not carry one.

Merge after #1278, which deploys `PropAMMFallbackExecutor` and gives it a production address.

## Tests

`FermiSwapRouterTest` moves to block 25682938, where the venue is stale, and executes Rust-encoded calldata end to end: 1 WETH in, USDC out of the retry. It was on 25143884, where the PropAMMRouter is not deployed. `PropAMMFallbackRouterTest.testIndexedFermiswapIntegration` covers the same route under a `propammfallback:fermiswap` label. A wrong venue address reverts `UnknownVenue`, so both tests pin it.


tamaralipows:
`PropAMMFallbackExecutor` is deployed at `0xA2F9f6f4a3F64EDa6014ce71A8ec8A874d690ADF`, and #1278 now carries that address in `config/executor_addresses.json` (d0c056aa4).

Verified on-chain: the contract's `PROPAMM_ROUTER()` returns `0x4DdF368080CD7946db5b459aD591c350158175e1`, and the TychoRouter at `0xea290cE3Eae57BdB37E57872a5a14DC0d2f6e614` holds an activation timestamp of `1787424959` for it. That is about 23 hours out, so until it passes every swap on this executor reverts `Dispatcher__ExecutorIsTimelocked`.

With that address in place, the production config satisfies both conditions this PR needs, so `vm:fermiswap` resolves to the router as soon as both PRs merge.

tvinagre (approved):

PR #1347, opened 2026-08-21
feat(simulation): Groundwork for VM protocols as native mid-block processors

PierreMkt:
Groundwork for [VM protocols as native mid-block processors](https://propeller-heads.atlassian.net/wiki/spaces/GEN/pages/3911974935/VM+protocols+as+native+mid-block+processors). Fluid prices as before — nothing writes `pool_reserves_adjusted` yet.

| Commit | Change |
|---|---|
| `refactor(simulation)` | `SimulationParameters` derives `Default` |
| `feat(common)!` | `generate_deltas` takes a `PendingBlock`: target block, txs, account state |
| `feat(simulation)` | per-call native balance overrides, applied in `OverriddenSimulationDB::basic_ref` |
| `refactor(simulation)` | fluid splits into `call_resolver` / `decode_reserves`; `delta_transition` reads the `pool_reserves_adjusted` attribute |

`PendingBlock` carries the target block because an indexer that derives number and timestamp from the confirmed parent misprices fluid's time-dependent limits whenever a slot is missed.


crates/tycho-simulation/src/evm/protocol/fluid/vm.rs
```diff
@@ -79,24 +130,34 @@ where
...
         data,
         value: U256::ZERO,
-        overrides: None,
+        overrides,
         gas_limit: None,
         transient_storage: None,
-        block_overrides: None,
+        block_overrides: block_timestamp
+            .map(|timestamp| BlockEnvOverrides { number: None, timestamp: Some(timestamp) }),
+        balance_overrides: None,
```
PierreMkt: This is wrong. Here the indexer will need to pass the native balance override!

crates/tycho-simulation/src/evm/protocol/fluid/v1.rs
```diff
@@ -364,28 +364,61 @@ impl ProtocolSim for FluidV1 {
...
+                    })?;
+                let sync_time = u64::from_be_bytes(timestamp.as_ref().try_into().map_err(
+                    |_| {
+                        TransitionError::DecodeError(format!(
+                            "block_timestamp must be 8 bytes, got {}",
+                            timestamp.len()
+                        ))
+                    },
+                )?);
+                vm::decode_reserves(reserves, sync_time)?
```
PierreMkt: make sure the `FluidProcessor` in `builder-integration` adds `pool_reserves_adjusted` and `block_timestamp` to `delta.updated_attributes`

crates/tycho-simulation/src/evm/engine_db/simulation_db.rs
```diff
@@ -42,20 +45,38 @@ impl<'a, DB: DatabaseRef> OverriddenSimulationDB<'a, DB> {
...
+        else {
+            return Ok(info);
+        };
+        Ok(Some(match info {
+            Some(mut account_info) => {
+                account_info.balance = *balance;
+                account_info
+            }
+            None => AccountInfo { balance: *balance, ..Default::default() },
+        }))
```
PierreMkt: Yeah this doesn't look right. We should return `None` if info is `None`, not `AccountInfo { balance: *balance, ..Default::default() }`

crates/tycho-simulation/src/evm/engine_db/simulation_db.rs
```diff
@@ -27,12 +27,15 @@ use super::{
 };
 
 /// A wrapper over an actual SimulationDB that allows overriding specific storage slots
+/// and native balances
 pub struct OverriddenSimulationDB<'a, DB: DatabaseRef> {
     /// Wrapped database. Will be queried if a requested item is not found in the overrides.
     pub inner_db: &'a DB,
     /// A mapping from account address to storage.
     /// Storage is a mapping from slot index to slot value.
     pub overrides: &'a HashMap<Address, HashMap<U256, U256>>,
```
michcio1234, line 36: Maybe rename to `storage_overrides` here and in other places? Because now `overrides` is no longer unambiguous - it could be storage or native balance overrides. 
Just a suggestion.

crates/tycho-simulation/src/evm/protocol/fluid/v1.rs
```diff
@@ -364,28 +364,61 @@ impl ProtocolSim for FluidV1 {
 
     fn delta_transition(
```
michcio1234: Do I understand correctly: this function infers the protocol state from passed deltas, and if they are missing, it falls back to reading the vm state?

michcio1234 (commented):
Reviewed, but I don't know the codebase well enough to approve.

crates/tycho-simulation/src/evm/engine_db/simulation_db.rs
PierreMkt: yeah i thought about that but i don't want to break existing interfaces here 😬

crates/tycho-simulation/src/evm/engine_db/simulation_db.rs
```diff
@@ -27,35 +27,42 @@ use super::{
...
-        self.inner_db.basic_ref(address)
+        let info = self.inner_db.basic_ref(address)?;
+        let Some(balance) = self
+            .native_balance_overrides
+            .get(&address)
+        else {
+            return Ok(info);
+        };
+        Ok(Some(AccountInfo { balance: *balance, ..info.unwrap_or_default() }))
     }
```
PierreMkt, line 70: taking into account native_balance_overrides is the main change of commit 3

crates/tycho-simulation/src/evm/pending.rs
```diff
@@ -139,34 +138,43 @@ impl PendingBlockProcessor {
...
-    /// * `txs` — candidate bundle in execution order; failed transactions are skipped.
-    /// * `target_header` — header of the block being built. Its `number` is used for the
-    ///   parent-block guard; the full header is forwarded to `apply_deltas_ephemeral` so that block
-    ///   number and timestamp are injected into each state delta.
+    /// * `pending` — the in-flight block: the block being built, the candidate bundle in execution
+    ///   order (failed transactions are skipped), and post-execution account state for the accounts
+    ///   it touched. Its [`block`][PendingBlock::block] is the single source of the target block:
+    ///   its number drives the parent-block guard, and it is the header forwarded to
+    ///   `apply_deltas_ephemeral`, so the clock an indexer reads state under is the same one each
+    ///   state delta is stamped with.
```
PierreMkt: too verbose

brunoguerios (commented):
**21 comments** — 0 Blocker · 5 Important · 8 Suggestion · 7 Nit · 1 Context.

### Important
- **DESIGN `#1`** `call_resolver` re-clones the shared override maps per pool, defeating the borrow-not-clone design it sits on — `vm.rs:105`
- **IMPL `#1`** Balance override unreachable when the inner DB errors on missing accounts — `simulation_db.rs:57`
- **IMPL `#2`** `delta_transition` panics on a `block_timestamp` attribute longer than 8 bytes — `v1.rs:379`
- **TESTS `#1`** Resolver tests need RPC_URL but lack the mandated `#[ignore]` marker — `vm.rs:468`
- **DOCS `#1`** `delta_transition` doc promises a fallback the code does not have — `v1.rs:365`

### Suggestions
- **DESIGN `#2`** Block→BlockHeader field mapping duplicated inline instead of living on a type — `pending.rs:167`
- **DESIGN `#4`** Fluid's delta-attribute wire contract is scattered string literals, decoupled from the ABI knowledge that fulfills it — `v1.rs:365`
- **TESTS `#3`** `ResolverOverrides.storage` is never exercised — `vm.rs:468`
- **TESTS `#4`** `simulate()`'s balance-override pass-through has no test at its own layer — `simulation.rs:617`
- **DOCS `#2`** `PendingBlock` struct doc defines the type by its consumer — `blockchain.rs:153`
- **DOCS `#3`** `block()` getter doc carries caller narration and duplicated rationale — `blockchain.rs:166`
- **DOCS `#4`** `accounts()` doc conflates producer and consumer — `blockchain.rs:178`
- **DOCS `#5`** `generate_pending_update` doc references a `target_block` parameter that no longer exists — `pending.rs:128`

### Nits
- **DESIGN `#7`** `generate_deltas` takes `&mut self` while its docs guarantee no mutation — `traits.rs:65`
- **SIMPLIFY `#5`** `&Address` resolver params force an allocation — `vm.rs:124`
- **TESTS `#5`** Known-account override fixture has no code — `simulation_db.rs:777`
- **DOCS `#6`** Param doc narrates the internal forwarding path — `pending.rs:143`
- **DOCS `#7`** Doc example binds the processor to `pending` — `pending.rs:60`
- **DOCS `#8`** `ResolverOverrides` public fields lack docs — `vm.rs:99`
- **DOCS `#10`** `fields` local for a `FluidPoolState` — `v1.rs:374`

### Context
- **BIZ `#2`** The stream decoder always injects and overwrites `block_timestamp`, constraining the future `pool_reserves_adjusted` producer — `decoder.rs:1057`

crates/tycho-simulation/src/evm/protocol/fluid/v1.rs
```diff
@@ -362,30 +362,43 @@ impl ProtocolSim for FluidV1 {
...
+                    .ok_or_else(|| {
+                        TransitionError::MissingAttribute("block_timestamp".to_string())
+                    })?;
+                vm::decode_reserves(reserves, sync_time)?
+            }
+            None => {
+                let engine = create_engine(SHARED_TYCHO_DB.clone(), false).expect("Infallible");
+                vm::fetch_pool_state(&self.pool_address, &Bytes::from(RESERVES_RESOLVER), &engine)?
+            }
+        };
```
brunoguerios: **[BIZ `#2`]**

**What changed**: The decoder unconditionally inserts an 8-byte `block_timestamp` into every delta before `delta_transition` runs (confirmed and ephemeral paths alike), and that insert silently replaces any `block_timestamp` an extractor emitted. Two consequences for the future producer of `pool_reserves_adjusted`: the `MissingAttribute` error is reachable only for direct `delta_transition` callers, and the `sync_time` fluid stamps always comes from the decoder’s header, never from the delta producer.

**Where**: `crates/tycho-simulation/src/evm/decoder.rs:852–855`, `crates/tycho-simulation/src/evm/decoder.rs:1037–1040`, `crates/tycho-simulation/src/evm/decoder.rs:1057–1072`, `crates/tycho-simulation/src/evm/protocol/fluid/v1.rs:374–392`

- Whatever future component writes `pool_reserves_adjusted` must accept that the paired timestamp fluid reads is the decoder-injected block timestamp, not one it emits itself.

crates/tycho-simulation/src/evm/protocol/fluid/vm.rs
```diff
@@ -73,30 +121,62 @@ where
...
-        value: U256::ZERO,
-        overrides: None,
-        gas_limit: None,
-        transient_storage: None,
-        block_overrides: None,
+        overrides: overrides.storage.clone(),
+        block_overrides: overrides.block.clone(),
+        native_balance_overrides: overrides.native_balances.clone(),
+        ..Default::default()
     };
```
brunoguerios, line 150: **[DESIGN `#1` · Important]**

**Description**: The docstring (vm.rs:107–108) says `overrides` is taken by reference "so one override set can be built once and reused across every pool priced against the same pending block", but the body clones all three maps into the owned `SimulationParameters` fields on every call (vm.rs:129–131). The same PR adds a comment in `simulate` (crates/tycho-simulation/src/evm/simulation.rs:100–101) justifying borrowing precisely because "an override map can hold every storage slot a pending block touched, and this runs once per pool" — so the O(map-size)-per-pool copy eliminated at the engine layer is reintroduced one layer up, on the exact hot path this groundwork targets.

**Symptoms** (one fix resolves all):
- **Performance**: the future mid-block flow pays an O(touched slots) clone of `storage`, `block`, and `native_balances` per pool per bundle, `crates/tycho-simulation/src/evm/protocol/fluid/vm.rs:129–131`
- **Misleading doc**: the "built once and reused" paragraph promises sharing the shape does not deliver, and is caller-side narration in the fluid layer, `crates/tycho-simulation/src/evm/protocol/fluid/vm.rs:105–108`
- **Hidden cost**: the by-reference signature saves no work over taking `ResolverOverrides` by value — it only hides the triple clone from the caller, `crates/tycho-simulation/src/evm/protocol/fluid/vm.rs:107–108`

**Suggestion**: Either make the sharing structural — store the maps as `Arc<HashMap<...>>` in `ResolverOverrides` and change `SimulationParameters.overrides` / `native_balance_overrides` (and the doc paragraph) to match, so per-pool reuse is a refcount bump — or take `ResolverOverrides` by value and move the fields into `SimulationParameters` (today's only callers pass a fresh value; a future reuse-across-pools caller writes an explicit clone). Both variants remove the misleading doc paragraph; deciding now is cheap, and once producers code against these public shapes it becomes a breaking change.

crates/tycho-simulation/src/evm/pending.rs
```diff
@@ -139,34 +138,43 @@ impl PendingBlockProcessor {
...
                 current: self.current_confirmed_block,
             });
         }
+        let target_header = BlockHeader {
+            hash: target_block.hash.clone(),
+            number: target_block.number,
+            parent_hash: target_block.parent_hash.clone(),
+            timestamp: target_block.ts.and_utc().timestamp() as u64,
+            ..Default::default()
+        };
```
brunoguerios: **[DESIGN `#2` · Suggestion]**

**Description**: `generate_pending_update` hand-builds a `BlockHeader` from `pending.block()` (including the `ts.and_utc().timestamp() as u64` conversion), duplicating the mapping that already exists in `From<&BlockAggregatedChanges> for BlockHeader` (crates/tycho-client/src/feed/mod.rs:84–96). Two copies of the same field-mapping knowledge can drift — e.g. if `BlockHeader` gains a field derivable from `Block`, only one site would be updated.

**Suggestion**: Add `impl From<&Block> for BlockHeader` in tycho-client (where `BlockHeader` lives and `Block` is a dependency), use it here, and let the existing `From<&BlockAggregatedChanges>` impl delegate to it for the block-derived fields.

crates/tycho-simulation/src/evm/protocol/fluid/v1.rs
```diff
@@ -362,30 +362,43 @@ impl ProtocolSim for FluidV1 {
...
+        {
+            Some(reserves) => {
+                let sync_time = delta
+                    .updated_attributes
+                    .get("block_timestamp")
+                    .map(|timestamp| u64::from(timestamp.clone()))
+                    .ok_or_else(|| {
+                        TransitionError::MissingAttribute("block_timestamp".to_string())
+                    })?;
+                vm::decode_reserves(reserves, sync_time)?
```
brunoguerios: **[DESIGN `#4` · Suggestion]**

**Description**: The consumer side of the new wire contract — attribute names `"pool_reserves_adjusted"` / `"block_timestamp"`, ABI-encoded `PoolWithReserves` bytes, big-endian `Bytes` timestamp — exists only as literals and a docstring in v1.rs, while the payload knowledge (the `sol!` types, `call_resolver` producing the bytes) lives in vm.rs. The future producer this PR is groundwork for must re-derive names and encodings by reading both files, and a typo compiles fine and silently falls through to the VM path.

**Suggestion**: Define named consts for the attribute keys in vm.rs next to the `sol!` block, and pair the split with an extraction/encoding helper there (e.g. `fn pending_state_attributes(bytes, ts) -> [(name, Bytes); 2]` / `fn from_attributes(&updated_attributes) -> Option<Result<FluidPoolState>>`) so producer and consumer share one definition of the contract.

crates/tycho-common/src/traits.rs
```diff
@@ -52,26 +51,28 @@ pub trait TxDeltaIndexer: Send {
...
     ///
     /// # Parameters
     ///
-    /// * `txs` — ordered slice of in-flight transactions, typically a builder's candidate bundle or
-    ///   the full mempool selection for one block.
-    fn generate_deltas(&mut self, txs: &[TxInput]) -> BlockAggregatedChanges;
+    /// * `pending` — the in-flight block: the block being assembled, transactions in execution
+    ///   order (typically a builder's candidate bundle or the full mempool selection for one
+    ///   block), and post-execution account state for the accounts those transactions touched.
+    fn generate_deltas(&mut self, pending: &PendingBlock) -> BlockAggregatedChanges;
```
brunoguerios: **[DESIGN `#7` · Nit]**

The rewritten docstring promises "Internal state is **not** modified" and idempotent results, but the `&mut self` receiver leaves that as a convention; `&self` would enforce it structurally (or drop the guarantee if implementors need caches).

crates/tycho-simulation/src/evm/engine_db/simulation_db.rs
brunoguerios, line 70: **[IMPL `#1` · Important]**

**Category**: bug

**Description**: `basic_ref` runs `self.inner_db.basic_ref(address)?` before consulting `native_balance_overrides`, so the documented branch (lines 37–40: "An override for an account absent from the inner database materializes it as an empty account") only works when the inner DB returns `Ok(None)`. `PreCachedDB::basic_ref` returns `Err(MissingAccount)` for untracked accounts (crates/tycho-simulation/src/evm/engine_db/tycho_db.rs:357–363) and `SimulationDB::basic_ref` RPC-fetches misses (simulation_db.rs:386–397), so on both real backends the promised branch is dead — the `?` fails the call first, or a needless RPC fetch happens for an account the caller wants synthesized.

**Plain English**: A caller sets a balance override for an account the database does not track, expecting the simulation to see a fresh account holding that balance, as the documentation promises. On the cached backend the whole simulation instead fails with a missing-account error, because the database is asked about the account before the override is consulted. On the RPC backend it works, but only after a wasted network fetch of an account the caller meant to synthesize. The override should apply even when the underlying database has no such account.

**Suggestion**: Check the override map first; only when no override exists, propagate the inner result unchanged. When an override exists, treat an inner miss (`Err` or `Ok(None)`) as an empty account and stamp the balance onto it, mirroring how `storage_ref` shields overridden slots from the inner DB.

crates/tycho-simulation/src/evm/protocol/fluid/v1.rs
```diff
@@ -362,30 +362,43 @@ impl ProtocolSim for FluidV1 {
...
+            .get("pool_reserves_adjusted")
+        {
+            Some(reserves) => {
+                let sync_time = delta
+                    .updated_attributes
+                    .get("block_timestamp")
+                    .map(|timestamp| u64::from(timestamp.clone()))
+                    .ok_or_else(|| {
+                        TransitionError::MissingAttribute("block_timestamp".to_string())
+                    })?;
```
brunoguerios: **[IMPL `#2` · Important]**

**Category**: bug

**Description**: `u64::from(timestamp.clone())` uses `impl From<Bytes> for u64` (crates/tycho-common/src/hex_bytes.rs:379–391), whose `buf[size_of::<u64>() - bytes_slice.len()..]` underflows and panics whenever the attribute holds more than 8 bytes. The decoder stamps exactly 8 bytes, but `delta_transition` is a public `ProtocolSim` method and `updated_attributes` is externally produced data, so a differently encoded (e.g. 32-byte) timestamp crashes instead of erroring. The new tests cover only the present and absent cases — the wrong-length edge is untested.

**Plain English**: A state update whose timestamp field is encoded wider than eight bytes crashes the whole process instead of failing that one pool's update. The update content comes from outside the library, so one malformed or differently-encoded field can take down every consumer of the stream. A bad field should surface as a per-pool transition error, and other pools should keep updating.

**Suggestion**: Reject attributes longer than 8 bytes with a `TransitionError` (e.g. via `<[u8; 8]>::try_from(...)` and `u64::from_be_bytes`, or a length check before the conversion) instead of the panicking `From` impl, and pin the hardened behavior with a test that feeds a 32-byte `block_timestamp` (next to `test_delta_transition_attribute_without_timestamp`) and asserts the error.

crates/tycho-simulation/src/evm/protocol/fluid/vm.rs
```diff
@@ -73,30 +121,62 @@ where
...
     let data = reserves_call.abi_encode();
 
-    let to = alloy::primitives::Address::from_slice(resolver_address);
+    let to = alloy::primitives::Address::from_slice(resolver_address.as_ref());
```
brunoguerios: Decision: revert the parameter to the pre-PR `&[u8]`.

---

**[SIMPLIFY `#5` · Nit]**

`fetch_pool_state`'s `&Address` forces `&Bytes::from(RESERVES_RESOLVER)` to heap-allocate only for `call_resolver` to view it as a slice again; the pre-PR `&[u8]` shape serves both callers (decoder.rs passes `.as_ref()`).

crates/tycho-simulation/src/evm/protocol/fluid/vm.rs
```diff
@@ -256,8 +423,148 @@ mod test {
...
+            &ResolverOverrides::default(),
+        )
+        .expect("resolver call failed");
+
+        assert_eq!(
+            U256::from_be_slice(&number),
+            U256::ONE,
+            "Default overrides must leave the engine's confirmed block in place."
+        );
+    }
```
brunoguerios, line 603: **[TESTS `#1` · Important]**

**Description**: `test_call_resolver_applies_native_balance_override`, `test_call_resolver_applies_block_env_overrides`, and `test_call_resolver_without_overrides_reads_confirmed_block` build their engine via `opcode_reporting_engine`, which calls `get_client(None).unwrap()` (vm.rs:454) — that panics without `RPC_URL` (crates/tycho-simulation/src/evm/engine_db/utils.rs:24–39). crates/tycho-simulation/CLAUDE.md:67–69 mandates `#[ignore = "Requires RPC_URL ..."]` on such tests; the neighboring `test_decode_simulation_db` (vm.rs:387) carries it.

**Suggestion**: Add `#[ignore = "Requires RPC_URL to be set in environment variables or .env file"]` to all three tests, matching `test_decode_simulation_db`.

brunoguerios, line 603: **[TESTS `#3` · Suggestion]**

**Description**: `call_resolver` wires all three override kinds into `SimulationParameters`, but the three resolver tests cover only `native_balances` and `block`; the `storage` field's plumbing (vm.rs:129) has no test, so dropping that line would pass the suite.

**Suggestion**: Add a fourth resolver test using the existing `opcode_reporting_engine` pattern with SLOAD bytecode (`PUSH1 0; SLOAD; PUSH1 0; MSTORE; PUSH1 32; PUSH1 0; RETURN`) and a storage override on slot 0.

crates/tycho-simulation/src/evm/simulation.rs
```diff
@@ -636,10 +646,8 @@ mod tests {
...
             data: Vec::new(),
             value: U256::ZERO,
-            overrides: None,
-            gas_limit: None,
-            transient_storage: None,
             block_overrides: Some(BlockEnvOverrides { number: Some(123), timestamp: Some(456) }),
+            ..Default::default()
         };
 
         let engine = SimulationEngine::new(state, false);
```
brunoguerios, line 657: **[TESTS `#4` · Suggestion]**

**Description**: The only test that exercises `params.native_balance_overrides` flowing through `simulate()` is fluid's `test_call_resolver_applies_native_balance_override`, which needs `RPC_URL` and per TESTS `#1` should be `#[ignore]`d — after which a default `cargo test` run no longer pins this wiring (the simulation_db.rs unit tests cover `basic_ref` only, not the engine passing the map in).

**Suggestion**: Add `test_simulate_applies_native_balance_overrides` beside `test_simulate_applies_block_env_overrides`, reusing its opcode-contract pattern with SELFBALANCE (0x47) and an override on the contract address.

crates/tycho-simulation/src/evm/engine_db/simulation_db.rs
```diff
@@ -739,4 +747,93 @@ mod tests {
...
+        let overridden_db = OverriddenSimulationDB::new(&db, &overrides, &native_balance_overrides);
+
+        let info = overridden_db
+            .basic_ref(address)
+            .expect("basic_ref should succeed")
+            .expect("account should exist");
+
+        assert_eq!(info.balance, U256::from(999), "Overridden balance should be returned.");
+        assert_eq!(info.nonce, 7, "Other account fields should be preserved.");
+    }
```
brunoguerios, line 802: **[TESTS `#5` · Nit]**

`test_balance_override_known_account` pins nonce survival but uses a code-less account; giving the fixture bytecode and asserting it survives would pin that `..info.unwrap_or_default()` preserves contract code — the case that matters for overriding a contract's balance.

crates/tycho-simulation/src/evm/protocol/fluid/v1.rs
```diff
@@ -362,30 +362,43 @@ impl ProtocolSim for FluidV1 {
         ))
     }
 
+    /// When `updated_attributes` carries `pool_reserves_adjusted` (the resolver's ABI-encoded
+    /// `PoolWithReserves` bytes) and `block_timestamp` (as `Bytes`), pool state is decoded from
+    /// them. Otherwise the reserves resolver is called against confirmed state.
```
brunoguerios: **[DOCS `#1` · Important]**

**Description**: The doc reads "When `updated_attributes` carries `pool_reserves_adjusted` ... and `block_timestamp` ..., pool state is decoded from them. Otherwise the reserves resolver is called." "Otherwise" covers the case where `pool_reserves_adjusted` is present but `block_timestamp` is missing — the code (v1.rs:378–386) errors with `TransitionError::MissingAttribute` there and never falls back to the resolver.

**Suggestion**: State the real contract: `pool_reserves_adjusted` alone selects the decode path, `block_timestamp` must accompany it or the transition fails with `MissingAttribute`; only when `pool_reserves_adjusted` is absent is the resolver called against confirmed state. While rewriting, say what `block_timestamp` encodes — "(as `Bytes`)" says nothing; it is a big-endian u64 of unix seconds.

crates/tycho-common/src/models/blockchain.rs
```diff
@@ -150,6 +150,38 @@ impl TxInput {
     }
 }
 
+/// The in-flight block a delta generation runs against.
```
brunoguerios: **[DOCS `#2` · Suggestion]**

**Description**: "The in-flight block a delta generation runs against" describes the type through `TxDeltaIndexer::generate_deltas`, its consumer. This is a domain model in tycho-common; per the project convention docs should not describe where the item is used, and a reader outside the indexer context learns nothing about what the value holds.

**Suggestion**: Define it by content: e.g. "A block under construction: the target block, its transactions in execution order, and post-execution state for the accounts they touched."

crates/tycho-common/src/models/blockchain.rs
```diff
@@ -150,6 +150,38 @@ impl TxInput {
...
+}
+
+impl PendingBlock {
+    pub fn new(block: Block, txs: Vec<TxInput>, accounts: HashMap<Address, AccountDelta>) -> Self {
+        Self { block, txs, accounts }
+    }
+
+    /// The block being assembled, as the caller stamped it. Its number and timestamp are the
+    /// clock the transactions execute under; deriving either from the confirmed parent instead
+    /// misprices any protocol whose state depends on elapsed time.
```
brunoguerios: **[DOCS `#3` · Suggestion]**

**Description**: "as the caller stamped it" and the mispricing sentence are written from the producer/consumer side of the interface, and the "deriving from the confirmed parent misprices..." rationale already lives on the `generate_deltas` contract (traits.rs:59–61) and again in the `generate_pending_update` param doc (pending.rs:141–146) — three copies of one design argument.

**Suggestion**: Keep one sentence on the getter — "The block being assembled; its number and timestamp are the clock the transactions execute under." — and let the trait doc own the never-derive-from-the-parent rule.

crates/tycho-common/src/models/blockchain.rs
```diff
@@ -150,6 +150,38 @@ impl TxInput {
...
+        &self.block
+    }
+
+    /// Transactions in execution order.
+    pub fn txs(&self) -> &[TxInput] {
+        &self.txs
+    }
+
+    /// Post-execution account state for the accounts the transactions touched.
+    /// Empty for protocols whose state is fully described by logs.
```
brunoguerios: **[DOCS `#4` · Suggestion]**

**Description**: "Empty for protocols whose state is fully described by logs" reads as if emptiness were keyed per protocol, but one `PendingBlock` is handed to every indexer and the producer decides what to populate. A producer could read this as licence to omit `accounts` whenever *some* protocol is log-based.

**Suggestion**: Describe it from the data's side, e.g. "May be empty when no consuming indexer needs account state — for example when every registered protocol derives its state from logs."

crates/tycho-simulation/src/evm/pending.rs
```diff
@@ -126,7 +125,7 @@ impl PendingBlockProcessor {
         self.advance_inner(msg)
     }
 
-    /// Simulates `txs` against the confirmed parent state of `target_block` and returns an
+    /// Simulates `pending` against the confirmed parent state of `target_block` and returns an
     /// ephemeral [`Update`].
     ///
     /// Drains any confirmed blocks that have arrived since the last call, then immediately
```
brunoguerios, line 129: **[DOCS `#5` · Suggestion]**

**Description**: The doc says "Simulates `pending` against the confirmed parent state of `target_block`" and "checks whether the parent block (`target_block - 1`) is available", but `target_block` is no longer a parameter — it is now an internal local (line 159). A reader scanning the signature finds no such name.

**Suggestion**: Refer to the pending block's own header: "against the confirmed parent state of `pending.block()`" and "(`pending.block().number - 1`)".

**Scope**: The `PendingError::ParentNotYetConfirmed` doc (pending.rs:36–37) uses the same phantom `target_block`; same fix applies.

crates/tycho-simulation/src/evm/pending.rs
brunoguerios: **[DOCS `#6` · Nit]**

"it is the header forwarded to `apply_deltas_ephemeral`" describes plumbing, not what the caller can expect; the clause after it already states the guarantee. Say the observable contract only — the returned update's deltas are stamped with this block's number and timestamp.

crates/tycho-simulation/src/evm/pending.rs
```diff
@@ -58,16 +58,15 @@ pub enum PendingError {
...
 /// # ) {
 /// pending
 ///     .subscribe_confirmed_block()
-///     .wait_for(|&n| n >= target_header.number - 1)
+///     .wait_for(|&n| n >= pending_block.block().number - 1)
 ///     .await
 ///     .expect("stream closed");
 /// let update = pending
-///     .generate_pending_update(txs, target_header, "bundle-1".to_string())
+///     .generate_pending_update(pending_block, "bundle-1".to_string())
```
brunoguerios: **[DOCS `#7` · Nit]**

In the example `pending` is the `PendingBlockProcessor` and `pending_block` is the `PendingBlock`, while the method's own `pending` parameter is the block — the same word names both concepts. Rename the example binding to `processor`.

crates/tycho-simulation/src/evm/protocol/fluid/vm.rs
```diff
@@ -57,13 +57,61 @@ sol! {
...
+}
+
+/// State the resolver call runs against, overriding what the engine's database holds.
+/// `Default` overrides nothing and reads confirmed state.
+#[derive(Debug, Clone, Default)]
+pub struct ResolverOverrides {
+    pub storage: Option<HashMap<alloy::primitives::Address, HashMap<U256, U256>>>,
+    pub native_balances: Option<HashMap<alloy::primitives::Address, U256>>,
+    pub block: Option<BlockEnvOverrides>,
+}
```
brunoguerios, line 120: **[DOCS `#8` · Nit]**

`storage`, `native_balances`, and `block` are public API (re-exported via fluid/mod.rs) with no field docs, and `block` is a block-environment override rather than the "state" the struct doc promises. One line each — storage slots, native balances, block number/timestamp — settles what maps onto `SimulationParameters.overrides` / `native_balance_overrides` / `block_overrides`.

crates/tycho-simulation/src/evm/protocol/fluid/v1.rs
```diff
@@ -362,30 +362,43 @@ impl ProtocolSim for FluidV1 {
...
+                vm::decode_reserves(reserves, sync_time)?
+            }
+            None => {
+                let engine = create_engine(SHARED_TYCHO_DB.clone(), false).expect("Infallible");
+                vm::fetch_pool_state(&self.pool_address, &Bytes::from(RESERVES_RESOLVER), &engine)?
+            }
+        };
 
-        trace!(?state, "Calling delta transition for {}", &self.pool_address);
+        trace!(?fields, "Calling delta transition for {}", &self.pool_address);
```
brunoguerios: I'd stick with the previous `state`

---

**[DOCS `#10` · Nit]**

`fields` is a generic word for a domain value; `state` says what it is (also `trace!(?fields, ...)` at v1.rs:394 and the local in vm.rs:82).

crates/tycho-simulation/src/evm/protocol/fluid/v1.rs
PierreMkt: yes this is correct. the processor will always injects ts of the pending block

crates/tycho-simulation/src/evm/protocol/fluid/vm.rs
PierreMkt: sure, but for a separate PR. i don't want to modify existing params/interfaces here

crates/tycho-simulation/src/evm/pending.rs
PierreMkt: sure we can do that 👍

crates/tycho-simulation/src/evm/protocol/fluid/v1.rs
PierreMkt: sure 👍

brunoguerios (approved):

PR #1350, opened 2026-08-24
feat(sushiswap_v2): enable execution and integration test on Base

zach030:
Follow-up to #1301 (Base indexing for `sushiswap_v2`), which added the
substreams manifest but not the execution/testing wiring.

- `crates/tycho-execution/config/executor_addresses.json`: register
  `sushiswap_v2` on Base pointing at the `UniswapV2Executor` already deployed
  there for `uniswap_v2` (`0x321C00604355DD20A2CB472D44fe8f37d821A82c`). No new
  deployment. Same reuse pattern as Ethereum, where `uniswap_v2` and
  `sushiswap_v2` share one executor.
- `crates/tycho-integration-test`: add `sushiswap_v2` to the Base default
  protocol list.

Why the executor can be reused:

- `UniswapV2Executor` takes the pool address from the swap calldata, has no
  factory address or init-code-hash check, and no callback path — nothing in it
  is Uniswap-specific.
- Its only parameter is `feeBps`; the deployed Base instance returns
  `feeBps() == 30`, which matches SushiSwap V2's 0.30% fee.
- The runtime bytecode of a Base SushiSwap V2 pair is byte-for-byte identical to
  a mainnet SushiSwap V2 pair, so the swap math the executor and `UniswapV2State`
  assume holds on Base too.


zach030:
Duplicate of #1305, which already carries these changes plus the price_printer/quickstart examples. Closing in favour of that PR.

PR #1351, opened 2026-08-24
feat(slipstreams): index the third Slipstream factory

zach030:
[DXI-41](https://propeller-heads.atlassian.net/browse/DXI-41)

Adds Aerodrome Slipstream's third factory `0xf8f2eB4940CFE7d13603DDDD87f123820Fc061Ef` (deployed at block 44,394,724) and its dynamic swap fee module `0x87D8f999BBa9343E8099552426775B51C338E8CB` (block 44,394,736) to the `aerodrome_slipstreams` package parameters, and adds the module to the simulation's supported list.

The factory is compatible with the existing implementation. Its runtime bytecode matches the second-generation factory except for the `poolImplementation` immutable; its pool implementation matches byte for byte except for the trailing metadata hash; and the fee module matches the two configured modules except for the `factory` immutable. Pool discovery, storage slot decoding, balance tracking, and the fee module ABI are therefore unchanged. The factory has created 783 pools holding roughly $40M, and 23 of the 46 pools above $100k are already configured through the new fee module.

## Changes

- Package parameters gain the factory and the fee module; package version `v0.1.3` → `v0.1.4`.
- `TickSpacingFee` carries the emitting factory, and the tick spacing fee store is keyed `{factory}:tick_spacing_{tick_spacing}`. Under the previous global key a `TickSpacingEnabled` event from one factory overwrote another factory's fee for the same tick spacing. The three deployed factories agree on every enabled tick spacing, so no indexed `default_fee` changes. `DynamicSwapFeeModule.getFee` resolves an unset base fee through `factory.tickSpacingToFee(pool.tickSpacing())`, so `default_fee` has to come from the pool's own factory.
- `map_pools_created` skips a pool whose factory has no stored fee for its tick spacing instead of panicking. A factory enables a tick spacing before it can create a pool on it, so a module started at the package's initial block always has the fee; a module started later — an initial-block override, as the range test runner uses — previously killed the stream with a deterministic wasm panic.
- `SUPPORTED_DYNAMIC_FEE_MODULES` in `tycho-simulation` gains the new module, with a matching decoder test case.
- New range integration test covering the factory's WETH/USDC pool.

## Dynamic fee logic is reused unchanged

Adding the module address to the allowlist is the only simulation change; `get_dynamic_fee` and its constants are untouched. The new module's runtime bytecode is byte-identical to the two configured modules apart from the embedded `factory` immutable, so its events, `dynamicFeeConfig` layout, and fee math are the same. The constants hardcoded in `dynamic_fee_module.rs` match the new deployment on chain: `secondsAgo` 600, `defaultFeeCap` 30,000, `defaultScalingFactor` 0, `SCALING_PRECISION` 1e6, `ZERO_FEE_INDICATOR` 420. Checked against a live pool, `0xa135b59fe221c0c8d441294f97f96fbc37bc9fbe` reports `dynamicFeeConfig` (baseFee 2700, feeCap 3600, scalingFactor 3e6, initialFeeEnabled false, initialFee 0) and `getFee` 2703, equal to `pool.fee()` and to what the Rust implementation computes.

Two things the module does that the simulation still does not model, unchanged by this PR:

- `DefaultFeeCapSet` / `DefaultScalingFactorSet` set module-global values that `dynamic_fee_module.rs` hardcodes. All three deployed modules currently match the hardcoded values.
- `registerDiscounted` grants a proportional fee discount, applied at the end of `getFee` after the fee cap. It is keyed by `tx.origin`, not by pool, so it cannot be expressed as component state; a discount would have to be applied per sender on the consumer side. No `DiscountedRegistered` event has ever been emitted by any of the three modules.

## Deployment notes

- The added parameters change every module hash, including modules that take no parameters, so the package back-processes from initial block 13,843,704.
- Components are only emitted on their `PoolCreated` block. An extractor whose cursor is already past block 44,394,724 picks up new pools of this factory but not the ones it created before that cursor.

## Testing

- `cargo +nightly fmt --check`, `cargo clippy -- -D warnings`, and `cargo test` for `base-aerodrome-slipstreams` (13 tests) and `tycho-simulation` slipstreams (34 tests).
- `cargo build --locked --release` plus `substreams pack`: the v0.1.4 spkg builds and every module reports initial block 13,843,704.
- `protocol-testing range --package base-aerodrome-slipstreams --chain base`: 1/1 passed, including the simulation and execution legs. All six encoded swaps return exactly the simulated amount, wei for wei, in both directions at 0.1%, 1%, and 10% of max — for example 58808621032895038 WETH in returns 130279794 USDC both simulated and executed. Execution needs a Base RPC with `debug_traceCall` for the harness's balance slot detection; no free public endpoint has it. `state_registry.rs` had no arm for `aerodrome_slipstreams`, and its fallback sends any unlisted protocol system to `EVMPoolState`, so the harness looked for an `ISwapAdapter` the native implementation does not have, logged `StateDecodingFailure`, and passed without simulating. Registering `AerodromeSlipstreamsState` fixes that. Cross-checked against chain state at the stop block: `sqrtPriceX96` implies a 2231.59 USDC/WETH spot, the smallest simulated trade returns 2230.32 (ratio 0.999431 against a 0.05% fee, the rest price impact), and the 10% trade returns 2215.32.

  The execution leg was blocked the same way: `EXECUTOR_MAPPING` in `execution.rs` had no `aerodrome_slipstreams` entry and no Slipstreams runtime bytecode fixture existed, so the harness failed with "Unknown protocol system". Both are added here. The fixture is the `SlipstreamsExecutor` `deployedBytecode` built with `profile.default`, whose 31 recorded sources all hash-match this branch, and it carries the same metadata shape as the existing fixtures — that profile pins solc and sets `bytecode_hash = 'none'` precisely so fixtures stay reproducible. It is 1017 bytes against the 1116 bytes deployed at the registered Base address; the deployed contract is verified and its source is identical to this branch, the size gap comes from `profile.production` compiling at `optimizer_runs = 1000` instead of 200. Same semantics, so the execution check is valid; only gas figures are approximate.

  Other native protocols still hit that fallback — `aerodrome_v1` among them, whose tests also set `skip_simulation: false` — so their simulation legs are silently skipped too. Left for a separate change, since enabling validation there may surface accuracy failures that are currently hidden.

[DXI-41]: https://propeller-heads.atlassian.net/browse/DXI-41?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


protocols/substreams/base-aerodrome-slipstreams/integration_test.tycho.yaml
```diff
@@ -5,7 +5,27 @@ protocol_type_names:
...
+      - id: "0x493e74eda2720e127baccc1a19b2d567bc14ab43"
+        tokens:
+          - "0x4200000000000000000000000000000000000006" # WETH
+          - "0x833589fcd6edb6e08f4c7c32d4f71b54bda02913" # USDC
+        static_attributes:
+          default_fee: "0x01f4"
+          tick_spacing: "0x0a"
+        creation_tx: "0x107e8160b39e8a24feec22ea937f191413b31a8c16250775473992767edf0218"
+        skip_simulation: false
+        skip_execution: true
```
die-herdplatte: We should be able to test execution by adding the protocol to `EXECUTOR_MAPPING`, no?

die-herdplatte (approved):
> * `DefaultFeeCapSet` / `DefaultScalingFactorSet` set module-global values that `dynamic_fee_module.rs` hardcodes. All three deployed modules currently match the hardcoded values.

Wouldn't it be good if we included this in the same resync (different PR)?

zach030:
> > * `DefaultFeeCapSet` / `DefaultScalingFactorSet` set module-global values that `dynamic_fee_module.rs` hardcodes. All three deployed modules currently match the hardcoded values.
> 
> Wouldn't it be good if we included this in the same resync (different PR)?

This was discussed before. Both `DEFAULT_SCALING_FACTOR` and `DEFAULT_FEE_CAP` are hardcoded in tycho-simulation. They could be tracked in substreams, but they'd need to be populated to every component's entity change, since we don't have anything like shared state. Flo had a plan for this, but it's currently pending on priority. Given that these two values update infrequently, we're currently leaning towards keeping them hardcoded plus adding monitoring. If they do change, we'd update tycho-simulation and then notify clients to update.

PR #1352, opened 2026-08-24
test(integration): override pAMM oracle state at the quoted block

tamaralipows:
## What changed

The integration test applies Titan's per-block registry slot writes as state overrides when it
simulates a pAMM swap, and serves whitelisted venues under `propammfallback:*` so their swaps go
through the PropAMMRouter. A new counter
`tycho_integration_price_level_oracle_override_misses_total` counts blocks that arrive with no
overrides.

## Why

A pAMM reads its price from the priority-update registry. The registry reverts `StaleUpdate()`
unless the venue's slot holds the current `block.timestamp`. Titan's `pamm_quote_stream` publishes
those slot writes for the block it builds, so applying them at that block lets the venue price the
swap.

Without them the venue's quote is never compared against its own fill: called directly it reverts,
and through the PropAMMRouter it swaps on Uniswap V3 instead.

For a captured FermiSwap frame, all 7 published slots carried the mined block's timestamp, while
the on-chain values were 9 to 81 minutes old.

The binary has not been run end to end yet; that needs a `TYCHO_URL`.

### DIFF
--- a/crates/tycho-integration-test/src/main.rs
+++ b/crates/tycho-integration-test/src/main.rs
@@ -56,7 +56,7 @@ use tycho_test::{
 
 use crate::{
     fee_fetcher::{fetch_router_fee_on_output, RouterFeeOnOutput},
-    oracle_overrides::OracleOverrides,
+    oracle_overrides::{override_protocol, BlockOverrides, OracleOverrides},
     statistics::TestStatistics,
     stream_processor::{
         price_level_stream_processor::PriceLevelStreamProcessor,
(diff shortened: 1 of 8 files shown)


crates/tycho-integration-test/src/main.rs
```diff
@@ -1179,13 +1191,37 @@ async fn process_update(
...
     }
 
-    let results =
-        match simulate_swap_transaction(&rpc_tools, block_execution_info.clone(), &block, None)
-            .await
-        {
-            Ok(results) => results,
-            Err((e, _, _)) => return Err(e),
-        };
+    // Titan publishes overrides per block, so only price level stream updates take them.
```
Troshchk, line 1258: Can there be a case, that we want to simulate for block, but the overrides for that block did not arrive yet?

Troshchk (commented):
Thank you @tamaralipows!
Left a q

crates/tycho-integration-test/src/main.rs
tamaralipows: Yes. The overrides are read once, at the moment the swap is simulated, and nothing waits for them.

In practice the test first waits for the chain to reach the quoted block, which usually gives the override stream enough time to publish. They can still be missing - at startup, after a Titan reconnect, or when the override stream lags the quote stream.

That case is counted, not silent, and it does not produce a false failure. The miss counter increments, the swap simulates on the indexed state, and the stale-quote revert that follows is recorded separately from the revert and validation-failure metrics.

If the miss counter is non-zero once this runs against a real Tycho instance, I would add a short wait for the block's overrides. I would rather measure that first.

crates/tycho-integration-test/src/main.rs
```diff
@@ -1179,13 +1191,37 @@ async fn process_update(
...
+    let oracle_overwrites = match update.update_type {
```
tvinagre, line 1259: We also need to include vm PropAmms, like vm:fermi

tvinagre (approved):

PR #1353, opened 2026-08-25
feat(indexing): Pendle protocol

kirsanium:
Substreams package for every Pendle V2 market on Ethereum, discovered from all five factories. Two component types: `pendle_market` (the SY↔PT AMM) and `pendle_sy` (the ERC-5115 wrapper and its quotable entry/exit tokens).

First of three PRs. No stacked PRs off of a fork so there are [simulation](https://github.com/kirsanium/tycho/pull/7) and [execution](https://github.com/kirsanium/tycho/pull/8) following.

## What this PR adds

`protocols/substreams/ethereum-pendle/` — 11 modules over 7 DAG stages:

| Stage | Modules | What it does |
|---|---|---|
| Discovery | `1_map_market_components`, `2_store_sy_seen`, `3_map_protocol_components`, `4_store_protocol_components` | Decode `CreateNewMarket` from both factory ABIs; batch `readTokens`/`expiry`/SY profile at creation; emit `pendle_market` + a deduped `pendle_sy` |
| Registry | `4_store_market_registry` (`src/registry.rs`) | One append-store key holding every market — id, SY, expiry, factory, creation fee — so "every live market" is enumerable without a store iterator |
| Balances | `5_map_relative_component_balance`, `6_store_balances` | `Transfer`-derived for markets, `balanceOf`-read for SYs |
| Reserves | `5_map_reserve_deltas`, `6_store_market_reserves` (`src/market_state.rs`) | `totalPt` / `totalSy` / `lastLnImpliedRate` replayed from `Swap`/`Mint`/`Burn` |
| PY index | `5_store_py_index` (`src/sy_rates.rs`) | `pyIndexStored` from `NewInterestIndex`, seeded at creation; `SY.exchangeRate()` batched per refresh block |
| Fees | `src/fees.rs` | Per-generation `lnFeeRateRoot` / `reserveFeePercent`, read back from `getMarketConfig`, keyed to Router V4 |
| Output | `7_map_protocol_changes` | Components, balances, reserves, fees, the two clocks |

Supporting: `src/consts.rs` (five factory addresses, two ABIs), `src/sy.rs` (per-`(SY, token)` deposit/redeem classification), `src/keys.rs`, `src/abi/`. 48 unit tests, all fixtures real mainnet logs or `eth_call` responses named with the block they were read at.

`integration_test.tycho.yaml` — three ranges, `skip_balance_check: false` throughout:

| Range | Blocks | What it pins |
|---|---|---|
| `test_market_creation_and_swaps` | 16032059–16072700 | Original factory: four-parameter `CreateNewMarket`, two markets + two SYs, the stETH rebase that forced `balanceOf` balances |
| `test_v6_market_creation_and_swaps` | 25722400–25755000 | Current generation: five-parameter event with a real `ln_fee_rate_root_at_creation`, and an SY quoting 6-decimal USDC against an 18-decimal asset |
| `test_reference_market_creation_and_swaps` | 17363200–17387000 | The brief's reference market, so `scalar_root` and `expiry` are checked against numbers from outside this integration |

## How to run it

```bash
cd protocols/substreams
cargo +nightly fmt -p ethereum-pendle -- --check
cargo +nightly clippy -p ethereum-pendle --all-features --all-targets -- -D warnings
cargo build -p ethereum-pendle --target wasm32-unknown-unknown --all-features
cargo test -p ethereum-pendle
```

The range test shells out to two binaries a plain checkout does not provide:

```bash
curl -sL https://github.com/streamingfast/substreams/releases/latest/download/substreams_linux_x86_64.tar.gz \
  | tar xz -C ~/.local/bin substreams
cargo build --release --bin tycho-indexer && export PATH="$PWD/target/release:$PATH"
cd docker && docker compose up db -d

cd protocols/testing
export RPC_URL=<archive node>
export SUBSTREAMS_API_TOKEN=<JWT API key>
cargo run -- range --package "ethereum-pendle"
```

## Design notes

**Two factory ABIs, not one.** The brief says all five factories emit the same `CreateNewMarket`. They do not: the original `0x27b1dAcd…` emits a four-parameter version with no `lnFeeRateRoot` (topic0 `0x166ae5f5…`), V3–V6 the five-parameter one (`0xae811fae…`). Both confirmed by decoding real logs. The original factory is the generation of the brief's reference market, so it cannot be cut — hence two decode paths.

**Two component types.** The market custodies SY and PT; the SY custodies the underlying. Merging would report a balance `Component.id` does not hold, and one SY backs every market on that asset across expiries — merging would duplicate the wrap edge per expiry and kill it whenever one market expired.

**Reserves ≠ balances.** `totalSy` excludes donated excess that `skim()` sweeps, so the balance channel carries real ERC-20 balances and reserves get their own pipeline. Reserves are replayed from events rather than read from storage: every write path emits one, the events are identical across all five generations, `MarketStorage`'s slot layout is not, and markets are created empty so a replay from creation is exact. `netSyToReserve` must be subtracted alongside `netSyOut` — a unit test fails if that term is dropped.

**SY balances are read, not accumulated.** A rebasing token moves a holder's balance without a `Transfer`; the wstETH SY holds stETH, and at block 16032658 transfer-derived accounting said `0` where the node said `4` wei. Markets stay on the transfer path — SY shares and PT do not rebase.

**The PY index.** `max(pyIndexStored, SY.exchangeRate())`. The stored half comes from `NewInterestIndex`, whose value is **indexed** — `topics[1]`, empty data, so a decoder reading data emits a silent zero. The rate half has no event stream at all and is batched per refresh block. `pendle_sy` carries `sy_exchange_rate` rather than `py_index_current`: `pyIndexStored` belongs to a YT, and one SY backs many YTs.

**Two clocks.** `rate_sampled_at` dates the rate and is emitted only when `exchangeRate()` resolved; `block_timestamp` dates the look and is emitted regardless. A failed read leaves the rate where it was and simply stops advancing its clock, which lets the consumer refuse to quote without a staleness flag. A market's clock is published for `EXPIRY_GRACE_SECONDS` (24h) past expiry so the component can be seen to die — emission stopping *before* expiry would freeze it live forever.

**Fees are state, not static attributes.** `readState()` fetches both from the factory per call and the override is keyed by the **calling router**, so an effective fee is only defined relative to one — indexed against Router V4. Only the *scope* is decoded from events; the value is read back from `getMarketConfig`, the same source the market uses. Creation-time-only reads would be wrong: the original factory's last `NewMarketConfig` fired at block 19419012, after its last market was created, halving `lnFeeRateRoot` to `499875041000000` — exactly the value the brief quotes for the wstETH market created at 17363218.

**Refresh anchoring.** Every `EntityChanges` must ride a `TransactionChanges`, and a per-block refresh has no causing transaction. It is anchored to the block's **last real transaction** — a fabricated hash would decode and then be persisted as real. Blocks with no transactions skip the refresh and log it.

**Staleness knob.** `sy_rate_refresh_blocks` defaults to `1`. ~112 B egress per live component per refresh block, linear across intervals 1/2/5/10/50 to within 3% — ~15 KB/block steady state (~109 MB/day) and ~147 GB for a full backfill. Wall-clock could not distinguish the settings at all (29–76 s per 1000 blocks at *every* interval, including none). So `1` ships for live indexing and a historical sync should raise it.

## Open decision — how long an SY keeps being refreshed

An SY has no expiry: it is an ERC-5115 wrapper whose deposit/redeem edges route through no market. Committed answer (**Option A**): refresh every SY component for as long as it exists, so an orphaned wrapper stays quotable at a live rate. The cost grows with the protocol's *history* rather than its live liquidity — one `exchangeRate()` plus one `balanceOf` per declared token per refresh block, forever, and through the whole backfill.

**Option B** — keep `rate_targets()` on the live set and bury an orphaned SY the way a market is buried: one clock-only emission, so the consumer sees a stale sample and refuses. Cost capped at the live set; deprecated wrappers refused rather than served.

Neither serves a stale rate as current; they differ on whether a wrapper nobody may be trading is worth a per-block call forever. **Recommendation: B**, unless someone routes through an SY with no live market — settled by counting orphaned SY components and checking whether their entry tokens appear in a router's token set at all. Reverting is a one-function change.

## What is missing

- **The V6 factory is an upgradeable proxy and the decoders are not version-guarded.** `0x6d247b1c…` is a `TransparentUpgradeableProxy` over `PendleMarketFactoryV6Upg`; the other four are immutable, and `0x1A6f…`/`0x3d75…`/`0x6fcf…` all verify as the same `PendleMarketFactoryV3.sol` source, which is why one ABI covers four generations. V6 holds **50 of 52 live markets**. Its surface is V3-compatible today, but an implementation swap changing a topic0 would make `decode_creation` and `fee_scope` stop matching **silently** — no new markets, frozen fees, no error. Cheapest guard not taken: log the topic0 of any unmatched log from a known factory address, in a loop that already runs. Alternatives: a `VERSION()` check in the once-per-market `RpcBatch`, a hard `bail!` on the proxy's `Upgraded` event, or an off-chain pin of the EIP-1967 slot.
- **Creation depends on creation-time `eth_call`s** (`readTokens`, `expiry`, the SY profile) with no retry: a node that fails them at the creation block yields a silently absent component.
- **Unchanged refresh values are re-emitted.** At `=1` most messages repeat the previous value; suppressing them needs a last-emitted store. An optimisation, not a correctness issue.
- **10 of 91 `(SY, token)` pairs classify as neither closed form** and are left out of the token list rather than flagged — 89% quotable, and the remainder is absent rather than wrong.
- **Native-ETH entry tokens are dropped** (no `decimals()` to classify): one token across all live markets.
- **Balances are tracked only from a component's creation block**, so anything an SY held before its first market existed is invisible.
- **The v1 fee fan-out is unit-tested only.** The one `NewMarketConfig` landing after a market exists sits ~3.4M blocks past `initialBlock`, ~10M blocks of store preparation away. The selection logic is a pure function with tests that fail if the fan-out is truncated.
- **`UnsetOverriddenFee` is carried but never exercised** — the original factory's entire history is 42 events with zero overrides ever.


kirsanium:
Didn't mention following PRs in the original post, so here they are: [simulation](https://github.com/kirsanium/tycho/pull/7) & [execution](https://github.com/kirsanium/tycho/pull/8.) (no stacked PRs off of a fork so they are not in the current repo)

kirsanium:
Realized I forgot to put out the PR which makes pendle's interation testing more strict: #1361 . This PR's correctness depends on it, since it also fixes the storage correctness.

PR #1359, opened 2026-08-26
feat(simulation): quote slipstream fees at the execution block

zach030:
Closes DXI-46. Design discussion: [DXI-46](https://propeller-heads.atlassian.net/browse/DXI-46).

## Problem

The `DynamicSwapFeeModule` charges the initial fee to the first tick-moving operation of a block and the dynamic fee to every later swap, keyed on `last observation ts == block.timestamp`. In simulation, the state's `block_timestamp` was injected by the decoder in the same delta that carried the observation update, so the two values always moved together and the check never selected the initial-fee branch — affected pools always quoted the dynamic fee (e.g. 2700 pips on `0xdFe5F275020def30993f042174Fc2D335678b626` where the first swap of a block pays 750).

Second mismatch: chained simulations of several swaps on one pool within a block all resolved the same fee, because no observation was recorded between them.

## Change

The stored observed-block timestamp is replaced by an **execution-block** timestamp maintained by the stream decoder, so the fee check asks the question the contract asks: has an observation been written in the block this quote will execute in?

- **`ProtocolSim::apply_block`** (defaulted, non-breaking): advances a state to its execution block and returns whether quoting behavior changed.
- **Decoder**: derives the execution block from each message header — a partial (flashblock) header targets the still-open block, a confirmed header the next one — and advances every stored state on each message, re-emitting only those whose fee actually changed. Idle pools cost one virtual call, no clones.
- **Chained same-block simulations**: `get_amount_out` records the observation the pool would write, mirroring `CLPool.swap` (only on tick change, pre-swap tick and liquidity, same-block repeat is a no-op). Mint/burn writes arrive via substreams storage diffs and need no simulation-side handling.
- **Gas**: the TWAP observe overhead is charged whenever the module reaches the oracle. `debug_traceCall` shows it runs `observe()` even at zero scaling factor, so only the cardinality guard skips it — this also corrects a pre-existing miss.

`velodrome_slipstreams` has no dynamic fee module and is unchanged.

## The first-in-block bet is opt-in

Before an observation exists in the execution block, the swap's position within it is unknowable from a local simulation. Per the Slack discussion that choice belongs to the client:

- **Default** (`BlockPositionAssumption::WorstCase`): quote `max(initial, dynamic)` — nothing stops a pool from configuring `initialFee` above its dynamic fee. Never over-quotes the output, and matches today's behavior for these pools, so upgrading without config changes is a pure bug fix.
- **Opt-in** (`BlockPositionAssumption::First`, per protocol registration): quote the initial fee, betting the swap lands first. Better quotes where the protocol discounts the first swap, at the cost of reverts when someone lands first — enable only if the submission path can win that race.

Once the pool has been touched in the execution block (visible to flashblock consumers), position is a known fact and the setting has no effect.

Opting in means registering the exchange with a decoder context instead of the plain `exchange` call:

```rust
use tycho_simulation::protocol::models::{BlockPositionAssumption, DecoderContext};

ProtocolStreamBuilder::new(&tycho_url, Chain::Base)
    .exchange_with_decoder_context::<AerodromeSlipstreamsState>(
        "aerodrome_slipstreams",
        filter,
        None,
        DecoderContext::new()
            .block_position_assumption(BlockPositionAssumption::First),
    )
```

Flashblock consumers benefit most from the opt-in: each partial update shrinks the unknown window, and once someone trades in the block the quote flips to the dynamic fee on its own.

## Migration

**Upgrading without config changes is behavior-neutral**: the default assumption reproduces today's quotes for every affected pool (verified over 2835 samples on dev, zero deviations). The two items below are the only things a consumer may need to touch.

- **`TychoStreamDecoder::new(chain)`** now requires the chain and the `Default` impl is gone, so a missing chain is a compile error instead of a silently wrong block time. Consumers using `ProtocolStreamBuilder` are unaffected; direct constructors change one line.
- The slipstream state's serde field was renamed `block_timestamp` → `execution_block_timestamp` with no alias: the two mean different things (block last observed vs block the quote executes in), so a stale dump fails loudly instead of loading a wrong value.

## Testing

- **Wei-exact replay** of the reported swap pair (Base block 50166683): both fee branches, both outputs and the oracle-index advance match the chain. [Details](https://github.com/propeller-heads/tycho/pull/1359#issuecomment-5423903774).
- **Live on dev**, both consumer modes: flashblock 2806/2806 sealed-block fee checks matched with 443 intra-block flips tracked; full-block 883/883 against `eth_call` with `blockOverrides` (state at N, block env of N+1), which asks for the first-in-block fee rather than the end-of-block fee.
- **A/B run** of the two assumptions over 2202 shared samples: the reported pool quotes only 2700 under the default and 750/2700 under the opt-in; `First` is never more expensive than `WorstCase`; pools whose initial fee never applies are bit-identical in both modes.
- Unit tests cover execution-block derivation, refresh re-emission and failed-component exclusion, both assumptions (including `initialFee > dynamic`), the observation write, the gas branch and `transform` underflow.

[DXI-46]: https://propeller-heads.atlassian.net/browse/DXI-46?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


zach030:
## Test report

Three layers of validation, all against the originally reported pool plus the other active `initialFeeEnabled` pools on Base.

### 1. Wei-exact replay of the reported case (unit test, in this PR)

`replays_base_block_50166683_swap_pair_wei_exact` reconstructs pool `0xdFe5F275…` (AERO/cbBTC) from archive RPC at the parent block and replays the two swaps from the report:

| | on-chain | simulated |
|---|---|---|
| swap 1 (tx index 67): 0.01688626 cbBTC in | 2702.489253591513843346 AERO, **750 pips** | same, wei-exact |
| swap 2 (tx index 154), chained on swap 1's `new_state` | 719.894300964297656776 AERO, **2700 pips** | same, wei-exact |
| oracle index after the block | 214 (slot0) | 214 |

Fees independently cross-checked by back-solving the sqrtPrice displacement: swap 1 is only reproducible with 750, swap 2 only with 2700. The fee module config (`baseFee=2700, initialFee=750`) was confirmed from the `dynamicFeeConfig` getter and the `InitialFeeSet` event (block 45962238).

### 2. Live flashblock run against dev (~25 min, blocks 50476357–50477117)

Ran the branch as a `--partial-blocks` consumer on `tycho-base-dev`, streaming the 4 most active `initialFeeEnabled` pools. For every sealed block N, the fee the state quoted while N was the pending target was compared against the chain's own `pool.fee()` eth_call at N.

**3021 / 3021 checks matched, zero mismatches.**

| pool | checks | chain fee values observed | fee transitions tracked |
|---|---|---|---|
| `0xdFe5F275…` (reported) | 756 | 750 ×641, 2700 ×115 | 137 |
| `0x3f0296bf…` | 751 | 100 ×654, 400–499 (TWAP drift) ×97 | 173 |
| `0x4e962bb3…` | 753 | 100–156 in 8-pip TWAP steps | 149 |
| `0xb2cc224c…` | 761 | 305–439 in 15-pip TWAP steps | 95 |

The reported pool spent 641 blocks on the 750 initial fee and flipped to 2700 in the 115 blocks where someone got in first — under the old code the sim would have answered 2700 for all 756 (~85% of blocks misquoted on this pool). Two pools run a live scaling factor, so the match covers `observe` at execution-block timestamps pip-by-pip, not just the branch select. Many flips land on messages where the pool itself had no delta, which exercises the every-message refresh.

Emission volume (for the base-load discussion): 776 blocks, 10.9 messages/block, **3.62 states emitted per block across the 4 pools** (74% of messages carried zero states); refresh-driven re-emissions are ~0.8/block, the rest are real deltas.

### 3. Intra-block flips (follow-up instrumentation, 10 min)

1089 sealed-block checks (0 mismatches), **229 intra-block flips tracked** — a quote for the still-pending block changing the moment an earlier flashblock disclosed activity. On the reported pool these are the textbook 750 → 2700. Cross-checking flip blocks on chain surfaced a subtlety: several (e.g. 50479570/50479571/50479583) contain **no swaps at all** — the oracle write came from in-range **mint/burn** (`CLPool._modifyPosition` also checkpoints). The chain's fee flips there and so does ours, because the check keys on the observation timestamp from the substreams storage diffs rather than on trade events; a "did the pool trade this block" flag would have answered 750 and been wrong. Also caught a one-step flip to dynamic-with-TWAP (100 → 430).

### Coverage note

On dev in flashblock mode the client never received an `is_partial = false` message, so the confirmed-header projection (`number + 1`) is exercised by the replay test but not by the live run; a short full-block-mode validation run is a worthwhile follow-up.

zach030:
Did another review pass focused on what the sweep changes for downstream clients, cross-checked against the live-run logs. Pushed fixes for the findings (each with a regression test) and merged main (picks up the third fee module, `0x87D8f999…`):

1. **Zombie re-emission** — the sweep ran after the failed-components retain, so a lingering stored state for a component consumers were told was removed could be re-emitted on every fee flip. The sweep now skips failed components, and `removed_pairs` are purged from stored states too (the decode-failure removal path never cleared them — pre-existing leak that only became visible once something re-emits).
2. **Mid-stream panic for custom chains** — `Chain::block_time_secs()` panics for an unregistered custom chain and was called per confirmed message. Block time is now resolved once in `set_chain`, so a misconfigured consumer fails at construction with the registry error instead of on the first decoded block. (The RFQ stream's decoder never has a block header, so the sweep never runs there — verified.)
3. **Serde compatibility** — states serialized before the `block_timestamp` → `execution_block_timestamp` rename would fail to deserialize; added `#[serde(alias = "block_timestamp")]`.
4. **TicksExceeded partial result** — it skipped the observation write, so chaining a simulation on it would re-quote the initial fee.

Emission-volume numbers from the 25-min live run, for the base-load discussion: 776 blocks, 10.9 messages/block, **3.62 states emitted per block across the 4 pools** (74% of messages carried zero states) — refresh-driven re-emissions are ~0.8/block, the rest are real deltas.

One remaining coverage note: on dev in flashblock mode the client never received a single `is_partial = false` message, so the confirmed-header projection (`number + 1`) is exercised by the wei-exact replay test but not by the live run. Worth a short full-block-mode validation run before or after merge.

crates/tycho-common/src/simulation/swap.rs
```diff
@@ -70,15 +70,22 @@ macro_rules! params_with_context {
...
+    block: Option<BlockContext>,
+}
 
-impl Default for Context {
-    /// Creates a new `Context` with no future block information.
-    fn default() -> Self {
-        Self {}
+impl Context {
+    /// Creates a context targeting `block`.
+    pub fn at_block(block: BlockContext) -> Self {
```
die-herdplatte: I'm not sure whether this is the right method:
`Context` is a product type and this method would be more fitting on a sum type like `enum Context { Block(BlockContext) }`.
What if `Context` grows?
I think a builder-style `with_block(self, block: BlockContext) -> Self` would fit better

crates/tycho-simulation/src/evm/protocol/aerodrome_slipstreams/state.rs
```diff
@@ -64,7 +64,13 @@ const MAX_TICKS_CROSSED: u64 =
...
 pub struct AerodromeSlipstreamsState {
     id: String,
-    block_timestamp: u64,
+    /// Timestamp of the block a quote against this state is expected to execute in.
+    ///
+    /// Maintained by the stream decoder via [`ProtocolSim::set_execution_block`], not decoded from
+    /// the pool: the fee module's initial-vs-dynamic branch keys on the *execution* block, which
+    /// is the next block for a confirmed update and the still-open block for a flashblock
+    /// update.
+    execution_block_timestamp: u64,
```
die-herdplatte, line 76: The `#[serde(default)]` below makes me think that serde backwards-compatibility is important (idk?).
If so, this should be aliased.

crates/tycho-simulation/src/evm/protocol/aerodrome_slipstreams/state.rs
```diff
@@ -75,14 +81,18 @@ pub struct AerodromeSlipstreamsState {
     ticks: TickList,
     observations: Observations,
     dfc: DynamicFeeConfig,
+    /// Quote as if the swap lands first in its execution block (see
+    /// [`DecoderContext::assume_first_in_block`]); off = the worse of the two fee branches.
```
die-herdplatte: > off = the worse of the two fee branches

Hm, is `assume_first_in_block` the correct name then?
What if we would instead have
```rust
#[derive(Default)]
pub enum BlockPositionAssumption {
    /// No assumption: quote the worse of the two fee branches.
    #[default]
    WorstCase,
    /// Bet the swap lands before any other on the pool.
    First,
    // room for: AfterOthers
}
```

crates/tycho-simulation/src/evm/protocol/aerodrome_slipstreams/state.rs
```diff
@@ -123,10 +133,17 @@ impl AerodromeSlipstreamsState {
             ticks: tick_list,
             observations: Observations::new(observations),
             dfc,
+            assume_first_in_block: false,
         })
     }
 
-    fn get_fee(&self) -> Result<u32, SimulationError> {
+    /// Sets whether quotes assume the swap lands first in its execution block.
+    pub fn with_first_in_block_assumption(mut self, assume: bool) -> Self {
```
die-herdplatte: Why is this a builder method if everything else is passed into `new`? Backward compatibility?

crates/tycho-simulation/src/evm/protocol/utils/slipstreams/dynamic_fee_module.rs
```diff
@@ -138,39 +163,86 @@ pub(crate) fn get_dynamic_fee(
...
+    observations: &Observations,
+    execution_timestamp: u32,
+) -> Result<ResolvedFee, SimulationError> {
     let (scaling_factor, fee_cap) = if dfc.scaling_factor != 0 {
         (dfc.scaling_factor, dfc.fee_cap)
     } else {
         (DEFAULT_SCALING_FACTOR, DEFAULT_FEE_CAP)
     };
+    if scaling_factor == 0 {
+        return Ok(ResolvedFee::flat(base_fee.min(fee_cap)));
```
die-herdplatte: Shouldn't this be `observed_twap: true`?

crates/tycho-simulation/src/evm/decoder.rs
```diff
@@ -136,6 +140,56 @@ where
             registry: HashMap::new(),
             inclusion_filters: HashMap::new(),
             override_providers: HashMap::new(),
+            block_time_secs: Chain::Ethereum.block_time_secs(),
```
die-herdplatte: Hm, compared to the other fields I don't think this is a safe default, should we make `chain` or `block_time_secs` a required parameter to `new`?

crates/tycho-common/src/simulation/swap.rs
zach030: Agreed, replaced with builder-style `with_block(self, block) -> Self`.

crates/tycho-simulation/src/evm/protocol/aerodrome_slipstreams/state.rs
zach030: I don't see any serialize/deserialize round-trip of state in the repo. But even if someone does it, I'd rather not stay compatible here: `block_timestamp` and `execution_block_timestamp` mean different things, so an alias would silently swallow a wrong value. wdyt?

crates/tycho-simulation/src/evm/protocol/aerodrome_slipstreams/state.rs
zach030: Good point, took your shape as-is in 321a81768.

crates/tycho-simulation/src/evm/protocol/aerodrome_slipstreams/state.rs
zach030: Not backward compat, `new` takes the pool's on-chain facts, the `block_assumption` is a consumer-side preference with a sensible default. Didn't want to mix the two.

crates/tycho-simulation/src/evm/protocol/utils/slipstreams/dynamic_fee_module.rs
zach030: Yes, and worse.I traced `fee()` on a pool which scaling_factor is 0, the module calls `observe()` anyway (~50k gas), the factor just multiplies the deviation away. Fixed in 321a81768. Pre-existing miss too, the old gas line keyed on `scaling_factor != 0` as well.

crates/tycho-simulation/src/evm/decoder.rs
zach030: Agreed. Made it `Option<u64>`, a required param doesn't work since the RFQ decoder is headerless and has no chain. Unset now panics on the first block header instead of silently using 12s.

crates/tycho-simulation/src/evm/protocol/aerodrome_slipstreams/state.rs
die-herdplatte: It's used in fynd, but nothing really crucial:

1. Market recordings — test-fixtures/src/recording.rs: MarketRecording { metadata, updates: Vec<Update> } is written as zstd-compressed JSON by write_recording and read back by read_recording. tools/record-market captures live stream updates into these files; fynd-core/benches/common/mod.rs and fynd-core/tests/integration/harness.rs replay them. This is the one place the typetag tag/content encoding round-trips for real, and it depends on the skip-on-error behaviour (VM states are absent after a roundtrip).
2. types::Swap (fynd-core/src/types/quote.rs:1629) derives Serialize/Deserialize with protocol_state: Box<dyn ProtocolSim> inline — so the bound is required to compile. But production code deliberately avoids serializing it: tools/hindsight projects to slim_quote() precisely because the embedded state "dominates size and fails to serialize for vm pools (e.g. Curve)" (resolve/monitor.rs:260). The RPC layer (fynd-rpc, fynd-rpc-types) never mentions ProtocolSim, so nothing crosses the wire. Actual serde_json on Swap only happens in quote.rs unit tests.

It's fine to be breaking here imo

crates/tycho-simulation/src/evm/decoder.rs
die-herdplatte: I'm working on an RFQ architecture rework which also removes the co-opting of the `TychoStreamDecoder` for decoding RFQ states (so after that, the RFQ construction site issue doesn't matter anymore).

Given that `RFQStreamBuilder` already fills in dummy values to pass into `TychoStreamDecoder` (e.g. empty `sync_states`), I'd say passing `Chain::Ethereum` to `TychoStreamDecoder::new` wouldn't hurt: for this decoder `execution_block` is unreachable by type (`TimestampHeader::block()` is `None`), so the value can never be read.

In exchange, every block-based consumer gets the misconfiguration caught at compile time instead of on the first header, and `set_chain` + the panic path go away entirely.

crates/tycho-simulation/src/protocol/models.rs
```diff
@@ -38,21 +38,33 @@ use tycho_common::{
 
 use crate::evm::override_stream::OverrideSnapshot;
 
+/// What a quote may assume about the swap's position within its execution block, for protocols
+/// that price the first swap of a block differently.
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub enum BlockPositionAssumption {
+    /// No assumption: quote the worse of the two fee branches, so the output is never
```
die-herdplatte: I think this came from my comment, but it should be more general, not "the worse of the two fee branches"

crates/tycho-simulation/src/evm/protocol/utils/slipstreams/dynamic_fee_module.rs
```diff
@@ -138,39 +163,86 @@ pub(crate) fn get_dynamic_fee(
...
-            current_tick,
-            liquidity,
-            observation_index,
-            observation_cardinality,
-            observations,
-            blocktime,
-            scaling_factor,
-        )?;
-    Ok(total_fee.min(fee_cap))
+    // No scaling shortcut: the module runs the TWAP computation regardless and a zero factor
```
die-herdplatte: "No scaling shortcut" refers to the codebase's history which we only have context for in this PR

crates/tycho-simulation/src/evm/decoder.rs
zach030: `new(chain)` is required now. `set_chain`, the Option, the panic path, and the Default impl are all gone. Left a comment there saying so, worth clearing out when your rework lands.

crates/tycho-simulation/src/evm/decoder.rs
```diff
@@ -135,31 +123,24 @@ impl<H> TychoStreamDecoder<H>
 where
     H: HeaderLike + Clone + Sync + Send + 'static + std::fmt::Debug,
 {
-    pub fn new() -> Self {
+    /// Creates a decoder for `chain`.
+    ///
+    /// # Panics
+    ///
+    /// Panics for a custom chain with no registered config, at construction rather than on the
+    /// first decoded block.
```
die-herdplatte, line 131: This is the same "reference to an earlier state" issue as before

die-herdplatte (approved):

PR #1362, opened 2026-08-27
feat(simulation): pending-block state for Curve pools

michcio1234:
*🤖 Generated by AI.*

Adds the `tycho-simulation` support for pricing Curve pools against a pending block, and the public
surface an out-of-tree `TxDeltaIndexer` needs to drive it. Follows
[VM protocols as native mid-block processors](https://propeller-heads.atlassian.net/wiki/spaces/GEN/pages/3911974935/VM+protocols+as+native+mid-block+processors).

The Curve processor itself is not in this repo — it lives in the consuming repo. The substreams
package is untouched.

## Why Curve needs this

UniswapV2/V3/V4 describe their state in their logs. Curve is a VM protocol: its state lives in
contract storage, and `apply_deltas_ephemeral` applies only `state_deltas`, so nothing on the
pending path writes to the VM database. `CurveState::delta_transition` re-read that database and
ignored its delta, so a pending quote was one block stale.

Curve closes the gap the same way Fluid does: the indexer reads the pool's view getters under the
pending block's overrides and passes the readings across as a state-delta attribute.

## What changed

**`evm::simulation`** — `PendingOverrides` bundles storage, native-balance and block-environment
overrides, and `view_call` builds the `SimulationParameters` for a view call under them.
`PendingOverrides::default()` overrides nothing and reads confirmed state.

**`curve::vm`** — `decode_from_vm` splits into `read_raw_pool_state` + `build_pool`. The reading
half returns the adapter's existing `RawPoolState`, which now derives `Serialize`/`Deserialize`
and `PartialEq`. It takes the coin decimals as an argument rather than leaving the caller to patch
them in, so it never returns a `RawPoolState` that misdescribes the pool. Reading is done by a
`PoolReader` that holds the engine and a `&PendingOverrides` together — a getter that lost the
overrides would silently read confirmed state. The block environment travels with them, because a
ramping `A()` and the rate providers behind `stored_rates()` interpolate against the block's own
timestamp. Variant probing keeps reading confirmed state — which getters a pool exposes is fixed at
deployment.

**The attribute** — `pool_state_adjusted` (`POOL_STATE_ADJUSTED`), carrying a JSON-encoded
`RawPoolState`, with `encode_raw_state` / `decode_raw_state` for its payload.

**`CurveState::delta_transition`** — rebuilds the pool from that attribute when the delta carries
one, and falls back to the VM read when it does not. It checks the attribute's variant and coin
decimals against its own and fails on a mismatch. A malformed attribute fails too, rather than
falling back, which would silently price a pending block against confirmed state.

**`with_pending_indexer`** — no longer rejects extractors prefixed `vm:`. Curve is served by
extractor `vm:curve` but decodes into a hybrid state, so the prefix was the wrong test. The
requirement is now stated in the doc: the exchange must decode into a state whose `delta_transition`
can rebuild it from `state_deltas`. The generic VM adapter still does not qualify.

**Registration** — `vm:curve` joins `EXCHANGES_REQUIRING_FILTER`, so registering it without a
client-side filter warns.

**Public surface** — for an out-of-repo indexer to build against: `revm::DatabaseRef` from the
crate root under the `evm` feature; `curve::adapter` (`RawPoolState`, `build_pool`, `CurveVariant`)
as a public module; `CurveState::new`; and `read_raw_pool_state`, `encode_raw_state`,
`decode_raw_state`, `POOL_STATE_ADJUSTED`, `resolve_variant` from `evm::protocol::curve`.

**Docs** — `crates/tycho-simulation/CLAUDE.md` describes the pending path Fluid and Curve share.
`variant.rs` records that the `META_POOL_FACTORY` entry only holds while the substreams index that
factory's plain-pool branch alone.

## Testing

`cargo test -p tycho-simulation --lib --features evm curve`: 204 passed, 2 ignored (both RPC-backed
and unchanged by this PR). The new tests need no RPC:

- `delta_transition` rebuilds from the attribute, and rejects a malformed one, a variant mismatch
  and a decimals mismatch instead of falling back to the VM.
- `RawPoolState` survives the attribute encode/decode round-trip, including the `None`s inside
  `dynamic_rates`.
- `view_call` carries every override into the simulation parameters, and `default()` carries none.

Validation of the pending path against the chain lives with the processor, in the consuming repo.

## Known gaps

- Nothing in this repo writes `pool_state_adjusted`, so the new `delta_transition` branch is
  reachable only from its tests until the processor lands.
- `PendingOverrides` and `fluid::vm::ResolverOverrides` are now field-identical duplicates.
  Unifying them belongs in a change that is not also adding a protocol.


crates/tycho-simulation/src/evm/protocol/curve/state.rs
```diff
@@ -206,16 +206,33 @@ impl ProtocolSim for CurveState {
         Ok((u256_to_biguint(balance_in), u256_to_biguint(max_out)))
     }
 
+    /// When `updated_attributes` carries [`vm::POOL_STATE_ADJUSTED`], the pool is rebuilt from
+    /// those readings. Otherwise the view getters are read from the indexed VM storage.
+    ///
+    /// The attribute exists for pending blocks, whose state never reaches that storage: an
+    /// indexer that has already read the pool under the pending block's overrides passes the
+    /// readings through instead. Only the readings travel — the variant and the coin decimals are
+    /// static, so this state supplies them.
```
michcio1234:
```suggestion
    /// When `delta.updated_attributes` carries [`vm::POOL_STATE_ADJUSTED`], the pool is rebuilt from
    /// those readings. Otherwise the view getters are read from the indexed VM storage.
    ///
    /// The attribute exists for pending blocks, whose state never reaches that storage: an
    /// indexer that has already read the pool under the pending block's overrides passes the
    /// readings through instead.
```

crates/tycho-simulation/src/evm/protocol/curve/vm.rs
```diff
@@ -89,75 +131,129 @@ where
...
+
+/// Read the view getters `variant` needs from `engine`, for a pool with `n_coins` coins.
+///
+/// `overrides` is the state the getters run against; [`PendingOverrides::default`] reads the
+/// engine's confirmed state. Pass a pending block's storage, native balances and block
+/// environment to read the state that block would leave behind — the block environment matters
+/// because a ramping `A()` and rate providers interpolate against the block's own timestamp.
+///
+/// Returns a [`SimulationError`] if a getter required by the variant reverts. Getters that only
+/// some deployments of a variant expose are recorded as `None` instead.
```
michcio1234:
```suggestion
/// `overrides` allows for overriding the state the getters run against; empty [`PendingOverrides`]
/// means normal reading from engine's confirmed state. Pass a pending block's storage, native balances and block
/// environment to read the state that block would leave behind (block timestamp matters!).
///
/// Returns a [`SimulationError`] if a getter required by the variant reverts. Getters that only
/// some deployments of a variant expose are recorded as `None` instead.
```

crates/tycho-simulation/src/evm/decoder.rs
```diff
@@ -1058,9 +1058,10 @@ where
     /// supplied `pending_deltas`, and returns the result — **without writing back** to
     /// `DecoderState`. Calling this method twice with the same input produces identical results.
     ///
-    /// Only native protocols are supported. VM protocols (extractor prefix `"vm:"`) are rejected
-    /// at registration time in
-    /// [`with_pending_indexer`](crate::evm::stream::ProtocolStreamBuilder::with_pending_indexer).
+    /// Every state is rebuilt from `state_deltas` alone; nothing here writes to the VM database.
+    /// A protocol decoding into the generic VM adapter therefore cannot take part: it re-reads
+    /// pool state from that database, so its storage-derived values stay at the confirmed block —
+    /// even though the delta's balance and block-environment attributes do get applied.
```
michcio1234, line 1064: For reviewer's information: This will be changed in a future PR, where fully-vm protocols will also get support for pending blocks. For now, `apply_deltas_ephemeral` doesn't even _read_ from VM.

crates/tycho-simulation/src/evm/simulation.rs
```diff
@@ -417,6 +417,77 @@ pub struct BlockEnvOverrides {
     pub timestamp: Option<u64>,
 }
 
+/// State a view call runs against, overriding what the engine's database holds.
+///
+/// Built once per pending block and reused across every pool priced against it. `Default`
+/// overrides nothing, so a call reads the engine's confirmed state.
```
michcio1234: Remove the first sentence.

crates/tycho-simulation/CLAUDE.md
```diff
@@ -64,6 +64,24 @@ fallback for protocols too complex to port, not a default.
    adapter contract in `protocols/adapter-integration/`. Use only when native is not feasible.
 4. **RFQ** — off-chain quotes via API; for protocols that cannot be simulated on-chain at all.
 
+## Pending-block state for hybrid/VM protocols
+
+`apply_deltas_ephemeral` applies only `state_deltas`, so nothing on the pending path writes to the
+VM database. A protocol whose `delta_transition` re-reads the VM would therefore quote a pending
+block against confirmed state. Fluid and Curve close that gap the same way:
```
michcio1234:
```suggestion
`apply_deltas_ephemeral` applies only `state_deltas`; nothing on the pending path writes to the
VM database, so `apply_deltas_ephemeral` can't read the pending state from there. A protocol whose `delta_transition` re-reads the VM would therefore quote a pending
block against confirmed state. Fluid and Curve close that gap the same way:
```

crates/tycho-simulation/src/evm/simulation.rs
```diff
@@ -417,6 +417,76 @@ pub struct BlockEnvOverrides {
...
 
+/// State a view call runs against, overriding what the engine's database holds.
+///
+/// `Default` means no overrides, so a call reads the engine's confirmed state.
+#[derive(Debug, Clone, Default)]
+pub struct PendingOverrides {
+    pub storage: Option<HashMap<Address, HashMap<U256, U256>>>,
+    pub native_balances: Option<HashMap<Address, U256>>,
+    pub block: Option<BlockEnvOverrides>,
+}
```
PierreMkt, line 428: mmh this duplicates the ResolverOverrides in `crates/tycho-simulation/src/evm/protocol/fluid/vm.rs:113–120`

we should use either one for both protocols

crates/tycho-simulation/src/evm/protocol/curve/vm.rs
```diff
@@ -74,6 +75,47 @@ const A_PRECISION: u64 = 100;
...
+    pub out_fee: Option<U256>,
+    pub fee_gamma: Option<U256>,
+    pub offpeg_fee_multiplier: Option<U256>,
+    pub price_scale: Option<Vec<U256>>,
+    pub d: Option<U256>,
+    pub gamma: Option<U256>,
+    pub dynamic_rates: Option<Vec<Option<U256>>>,
+    pub precisions: Option<Vec<U256>>,
+    pub eth_variant: Option<bool>,
+}
```
PierreMkt: This also seems valid. Why not use `RawPoolState` instead of making a new struct ?

__________

The thirteen fields duplicate RawPoolState (crates/tycho-simulation/src/evm/protocol/curve/adapter/build.rs:94) minus variant and token_decimals, and build_from_readings copies them straight back. Two types and a copy exist to hide two static fields.

Proposed fix

Derive Serialize, Deserialize, PartialEq and Eq on RawPoolState, carry it as the attribute payload, and delete both CurvePoolReadings and build_from_readings, so delta_transition calls build_pool on the decoded state. One type, one name, no copy, and the field docs are already written. What it costs: the payload repeats variant and token_decimals that the consumer already holds, and the crate exports a vendored type whose derives then diverge from upstream. If the duplicate stays, move RawPoolState's field docs onto it and rename the attribute key to match the type family.

crates/tycho-simulation/src/evm/simulation.rs
michcio1234: Correct! I want to unify these types in a separate PR.

crates/tycho-simulation/src/evm/protocol/curve/vm.rs
michcio1234: Yeah, I even had a commit that removed this duplication (another way) and removed it to keep things simple. But okay, I can do what you're proposing, it is better than what I tried.

crates/tycho-simulation/src/evm/stream.rs
```diff
@@ -588,18 +589,18 @@ impl ProtocolStreamBuilder {
...
     pub fn with_pending_indexer(
         mut self,
         extractor: &str,
         indexer: Box<dyn TxDeltaIndexer>,
     ) -> Result<Self, StreamError> {
-        if extractor.starts_with("vm:") {
-            return Err(StreamError::SetUpError(format!(
-                "extractor '{extractor}' is a VM protocol; TxDeltaIndexer only supports native protocols"
-            )));
-        }
```
zach030, line 602: Not all VM protocols support this, right? So maybe we should use a whitelist instead of enabling it for all of them.

crates/tycho-simulation/src/evm/stream.rs
```diff
@@ -142,7 +142,8 @@ use crate::{
     utils::default_blocklist,
 };
 
-const EXCHANGES_REQUIRING_FILTER: [&str; 4] = ["vm:balancer_v2", "fluid_v1", "erc4626", "ekubo_v3"];
+const EXCHANGES_REQUIRING_FILTER: [&str; 5] =
+    ["vm:balancer_v2", "fluid_v1", "erc4626", "ekubo_v3", "vm:curve"];
```
zach030, line 146: Why do we need to add `vm:curve` here? Is it related to this PR?

crates/tycho-simulation/src/lib.rs
```diff
@@ -18,6 +18,8 @@ pub use tycho_common;
 pub mod tycho_core {
     pub use tycho_common::*;
 }
+#[cfg(feature = "evm")]
+pub use revm;
```
zizou0x: What's the motivation behind re-exporting this? I'm concerned that this makes bumping revm version much harder

crates/tycho-simulation/src/evm/stream.rs
```diff
@@ -588,18 +589,18 @@ impl ProtocolStreamBuilder {
...
     ///
-    /// Returns an error if `extractor` names a VM protocol (prefix `"vm:"`), which requires
-    /// `update_engine()` and cannot be simulated natively.
+    /// The exchange must decode into a state whose `delta_transition` can rebuild it from the
+    /// `state_deltas` the indexer produces, because that is all
+    /// [`apply_deltas_ephemeral`](crate::evm::decoder::TychoStreamDecoder::apply_deltas_ephemeral)
+    /// applies. Native and hybrid states qualify; the generic VM adapter does not, because it
+    /// re-reads pool state from the VM database — an indexer registered for one still gets its
+    /// balance and block-environment attributes applied, but every storage-derived value stays at
+    /// the confirmed block, with no error.
```
zizou0x, line 598: Should we keep a list of compatible extractors then (the one we have hybrid for)?

What happens if someone tries this with something that isn't compatible, like balancer, maverick and other vms? It silently ignores the delta?

zizou0x (commented):
Nice, just small comments/questions

crates/tycho-simulation/src/evm/stream.rs
michcio1234: There's another mechanism, added in a PR stacked on top of this one, that replaces the removed gate.

crates/tycho-simulation/src/evm/stream.rs
michcio1234: There's a mechanism in a PR stacked on top of this one, that adds a gate that prevents using incompatible protocols.

Anyway, soon we'll have to extend `apply_deltas_ephemeral` to also support state overrides, so that generic VM protocols are also supported.

crates/tycho-simulation/src/lib.rs
michcio1234: Without this re-export, builder-integration needs to pin revm version to exactly the one used by Tycho, to have compatible types. If this is an issue for Tycho, I can revert it and continue pinning the version in builder-integration repo.

crates/tycho-simulation/src/evm/stream.rs
michcio1234: Not directly, no.
But doesn't curve need a filter? https://github.com/propeller-heads/tycho/blob/229c83437b06f02bd3e5ad5649b6b719e7ded0bc/crates/tycho-simulation/src/evm/protocol/filters.rs#L115-L138

crates/tycho-simulation/src/lib.rs
zizou0x: Yes IMO this needs to be carefully considered

I think if we do this then bumping revm version could become a breaking change for us too as their whole interface will also be part of our interface. We need to assess wether we're fine with this or maybe only re-exporting some parts of their interface instead of the whole crate

crates/tycho-simulation/src/evm/stream.rs
zizou0x: I think this is correct, thanks for flagging it!

crates/tycho-simulation/src/evm/protocol/curve/state.rs
```diff
@@ -223,8 +227,14 @@ impl ProtocolSim for CurveState {
             .get(vm::POOL_STATE_ADJUSTED)
         {
             Some(encoded) => {
-                let readings = vm::decode_readings(encoded)?;
-                vm::build_from_readings(&readings, self.variant, &self.decimals)?
+                let mut state = vm::decode_readings(encoded)?;
```
michcio1234: It's no longer `readings` that are encoded, so the function should be named `decode_state`.

crates/tycho-simulation/src/evm/protocol/curve/state.rs
```diff
@@ -223,8 +227,14 @@ impl ProtocolSim for CurveState {
...
+                state.variant = self.variant;
+                state
+                    .token_decimals
+                    .clone_from(&self.decimals);
```
michcio1234: I'd rather check if they are the same, and error if they are not.

crates/tycho-simulation/src/evm/protocol/curve/state.rs
```diff
@@ -272,9 +280,11 @@ mod tests {
     }
 
     /// The TriCryptoNG USDC/WBTC/WETH pool (0x7f86bf…), balances aside.
-    fn readings(balances: Vec<U256>) -> CurvePoolReadings {
-        CurvePoolReadings {
+    fn readings(balances: Vec<U256>) -> RawPoolState {
```
michcio1234: Change the name to raw_pool_state

crates/tycho-simulation/src/evm/protocol/curve/state.rs
```diff
@@ -287,7 +297,10 @@ mod tests {
     }
 
     fn state(balances: Vec<U256>) -> CurveState {
-        let pool = build_from_readings(&readings(balances), VARIANT, &DECIMALS).expect("build");
+        let mut readings = readings(balances);
+        readings.variant = VARIANT;
+        readings.token_decimals = DECIMALS.to_vec();
+        let pool = build_pool(&readings).expect("build");
```
michcio1234: readings -> raw_state

crates/tycho-simulation/src/evm/protocol/curve/vm.rs
```diff
@@ -75,45 +74,23 @@ const A_PRECISION: u64 = 100;
...
 ///
 /// A pending block's state is not in the indexed VM storage, so an indexer that has already read
 /// the pool under that block's overrides passes the readings through this attribute instead.
 pub const POOL_STATE_ADJUSTED: &str = "pool_state_adjusted";
 
-/// Encode `readings` for the [`POOL_STATE_ADJUSTED`] attribute.
-pub fn encode_readings(readings: &CurvePoolReadings) -> Result<Bytes, SimulationError> {
-    serde_json::to_vec(readings)
+/// Encode `state` for the [`POOL_STATE_ADJUSTED`] attribute.
+pub fn encode_readings(state: &RawPoolState) -> Result<Bytes, SimulationError> {
```
michcio1234: Change the name.

zizou0x (approved):

PR #1363, opened 2026-08-27
feat(execution): fetch 10 blocks of Angstrom attestations

tamaralipows:
## What changed

`ANGSTROM_DEFAULT_BLOCKS_IN_FUTURE` goes from 5 to 10. A fetched attestation window now covers the current block plus 10 future ones, so it holds 11 entries instead of 6.


Troshchk (approved):
Thank you!

PR #1364, opened 2026-08-28
feat: integrate Tessera on Base

zach030:
Integrates Tessera V (Wintermute's propAMM on Base) as `vm:tessera`: substreams indexing, VM simulation adapter, and execution encoder + executor. Assessment measured $432M of 30-day volume (1.9% of Base DEX volume, reconciles with DeFiLlama within 0.7%) at sub-1 bp round-trip spread on the major books.

Plan: [Tessera Integration Plan](https://propeller-heads.atlassian.net/wiki/spaces/~71202094d3d73818b5447b85231f4bd027b78c/pages/3930849284/Tessera+Integration+Plan). On-chain research: `protocols/substreams/base-tessera/HANDOVER.md` (authoritative where it revises the plan).

## What Tessera is

No pools, no curves: one component per `base/USDC` book. Each book has its own price store (EIP-1967 proxy behind a shared implementation) into which an operator posts prices **every block**; a single treasury holds all inventory and settles by allowance through the verified `TesseraSwap` entrypoint. Quotes are a pure function of tracked storage plus a freshness gate on the age of the last post (full price ≤4 blocks, dead ≥~20) — simulating with the indexed block's environment always passes the gate. A delisted or stale book quotes zero and drops out of routing via limits.

## Indexing — `protocols/substreams/base-tessera`

- Discovery without events: a book is recognized when one transaction contains the store proxy's init writes (EIP-1967 impl slot + base-token slot 48 + quote-token slot 49 = USDC) plus the engine's mapping write of that store address. Verified against the WETH book creation (block 37,518,780) and the NVDAc book creation (50,526,653) in live runs.
- Dynamic tracked-contract set, no DCI: stable addresses and code-only satellites (3 impl generations, 2 pricing libs, write-path contract) via `params`; per-book stores via the components store. Components reference only contracts that exist at creation time.
- Treasury balance model (BopAMM pattern): `balanceOf` seeding on book creation, exact re-seed on treasury rotation (rotated once on Base, block 37,737,344), ERC20 `Transfer` + WETH `Deposit`/`Withdrawal` deltas. USDC is duplicated under every book.
- Admin mutations that can silently break simulation are surfaced as monitoring attributes: `engine` (TesseraSwap slot0 hot-swap), `store_impl` (EIP-1967 upgrade), `book_lib` (pricing-lib assignment). Runbook: append the new address to `params` and re-release.
- 20 unit tests, including synthetic-block discovery tests modeled on the NVDAc creation.

## Simulation — `TesseraSwapAdapter`

`EVMPoolState` adapter over `tesseraSwapViewAmounts` (byte-identical to the executable path at a pinned block): exact-in and exact-out natively, quote-ladder limit bisection (oversize returns 0, never reverts). Runtime asset + `constants.rs` wiring; no new Rust state. 12 Base fork tests, including settlement balance-diff == view quote in both directions and both order sides.

## Execution — `TesseraSwapEncoder` + `TesseraExecutor`

Protocol data is the packed token pair (40 bytes); the executor holds `TesseraSwap` as an immutable and calls `tesseraSwapWithAllowances(tokenIn, tokenOut, int256(amountIn), 0, receiver, "")`. `ProtocolWillDebit` with `receiver = TesseraSwap` (the venue pulls the input via `transferFrom`), `outputToRouter = false`. All four test layers pass: encoder unit tests, executor fork tests, cross-language calldata decode, and a TychoRouter single-swap integration test asserting the exact on-chain quote.

## Review

An adversarial review pass found three substreams defects (components referencing not-yet-deployed impl generations, token-global seed suppression, rotation-block double-count under in-block migration) — all fixed in the last commit and recorded in HANDOVER §9.5; the execution, encoder, and adapter layers were clean.

## Not in this PR (rollout phase)

- spkg release to S3 and extractor registration
- supported-protocols documentation row and register_exchanges snippet (docs are updated separately)
- Base executor deployment + `config/executor_addresses.json` entry (the integration test yaml sets `skip_execution` until then)
- week-long quote-vs-fill markout re-run gating consumer go-live; majors first (WETH, cbBTC, EURC = 90.9% of volume) — the long tail overquotes a median 4.5–5.7 bps

## Verification status

- [x] `cargo test` (substreams package, encoder, simulation wiring), clippy clean, nightly fmt
- [x] wasm build + `substreams pack`
- [x] Foundry: 12 adapter tests, 10 executor/router tests (Base forks)
- [x] Live substreams run over the WETH-book creation range
- [ ] Live substreams runs over the treasury-rotation and NVDAc ranges (running — server-side store prep)
- [ ] `protocols/testing` harness over the NVDAc range


zach030:
## Verification results

### Tests

| Layer | Suite | Result |
|---|---|---|
| Substreams | `base-tessera` unit tests (discovery, balances, config) | 21 passed |
| Simulation | `TesseraSwapAdapterTest` — Base fork @ 50,548,423 | 12 passed — settled amounts equal the view quote exactly, both directions and both order sides |
| Execution | encoder unit tests + cross-language calldata decode | 3 + 1 passed |
| Execution | `TesseraExecutorTest` + `TesseraRouterTest` — Base forks | 10 passed — full TychoRouter swap delivers exactly the on-chain quote |
| Repo | clippy (all targets), nightly fmt, wasm build, `substreams pack`, CI | clean / green |

### Live substreams runs (Base mainnet)

| Range | Covers | Result |
|---|---|---|
| 37,518,600–800 | WETH book creation: discovery, attributes, seeding, store code capture | pass |
| 37,737,340–350 | treasury rotation: `balance_owner` fan-out, re-seed, USDC duplicated identically per book | pass |
| 50,526,640–50,528,260 | full book lifecycle — the creation tx made **4 books at once** (NVDAc + 3 more tokenized equities); `book_lib` alerts on lib assignment; re-simulation marked 21/21 blocks after the first price post; **final indexed balances match on-chain `balanceOf` to the wei** | pass |

The lifecycle run uses an initial-block-patched manifest (the same `modify_initial_block` mechanism the testing harness applies) — full-history store prep on the shared StreamingFast tier was impractical.

### Bugs the live runs caught (both fixed)

- **Self-transfer double-count** (`fa84a31b4`): `Transfer(from=treasury, to=treasury)` was credited as an inflow. Tessera's first test swap self-transferred 3,444,538 USDC-wei (block 37,519,381), matching the observed drift exactly; with the fix, a per-block log reconciliation of the whole pre-rotation epoch matches on-chain balances to the wei. Same bug exists upstream in `ethereum-bopamm` (separate fix flagged).
- **Empty treasury store under a patched initial block** (`eecbd67de`): the constructor-era treasury write is never seen, so no balances / no `balance_owner` were emitted — the harness would hit this too. Fixed with a `treasury` params fallback; a production sync never reads it.

Also: an adversarial review pass found three substreams defects (fixed in `ce1fa9c58`, documented in HANDOVER §9.5); execution/encoder/adapter layers were clean.

### Outstanding before undrafting

- [ ] `protocols/testing` harness end-to-end (indexing → simulation) over the lifecycle range
- [ ] rollout items in the PR description (spkg release, executor deployment, markout gate)

zach030:
## NVDAc-range live run: pass (final code, `eecbd67de`)

The deep-history store prep on the shared StreamingFast tier was impractical (~3.8M of 39M stage-blocks cached after 3 days), so the run uses an initial-block-patched manifest — the same mechanism the `protocols/testing` harness applies (`modify_initial_block`). Range 50,526,640 → 50,528,260, covering the full lifecycle of the books created at block 50,526,653.

**Results**

- **Discovery**: the creation tx actually created **4 books at once** — NVDAc plus three more tokenized-equity books (`0xb2…d1fb`, `0xb2…58b7`, `0xb2…707c`; the pair-list helper only shows books once they start quoting). All 4 components emitted with correct ids, sorted tokens, `contracts = [TesseraSwap, engine, own store]`, and static attributes.
- **Monitoring attributes**: `book_lib` emitted at each book's pricing-lib assignment (staggered, 30-block intervals) — the alert surface for a future lib generation works.
- **Re-simulation cadence**: store changes marked on 21/21 blocks after the first price post — every-block repricing as designed.
- **Balances**: 2,358 deltas over the range; USDC seeded identically under all 4 books (duplication invariant); NVDAc inventory picked up via Transfer deltas when the treasury was funded pre-first-post. Final indexed balances match on-chain `balanceOf(treasury)` **to the wei** (USDC 422,020,567,818 @ 50,528,252; NVDAc 6,698,277,311 @ 50,528,247).

**Fix that fell out of it** (`eecbd67de`): with a patched initial block, TesseraSwap's constructor-era treasury write is never seen, so the treasury store was empty → no balances, no `balance_owner`. The harness would have hit the same. A `treasury` params fallback covers patched runs; a production sync from the real initial block witnesses every treasury write and never reads it.

Remaining before undrafting: `protocols/testing` harness end-to-end (indexing → simulation) over this range, then the rollout items in the description.

PR #1365, opened 2026-08-28
refactor(execution): encode swap groups and solutions in parallel

tamaralipows:
## What

RFQ swap encoders (`rfq:bebop`, `rfq:hashflow`, `rfq:liquorice`, `rfq:metric`) block on a network round trip to fetch the signed quote. Before this change a sequential route with several RFQ hops paid that round trip once per hop, one after the other, and a request with several solutions paid it once per solution.

Now:

- `encode_swap_groups` encodes every swap group of a solution on its own OS thread. `SequentialSwapStrategyEncoder` and `SplitSwapStrategyEncoder` use it.
- `TychoRouterEncoder::encode_solutions` encodes every solution on its own OS thread.
- Both use the new `map_on_threads` helper and return results in input order. A single group or solution runs on the calling thread.

Total encoding time for a route is bounded by its slowest RFQ hop instead of the sum of all hops.

## Public API

Unchanged. `TychoEncoder::encode_solutions` and `SwapEncoder::encode_swap` stay sync `fn`. Consumers such as Fynd need no changes.


crates/tycho-execution/src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -8,18 +8,83 @@ use crate::encoding::{
...
+/// bounds the total wait by the slowest group instead of the sum of all groups.
+fn encode_swap_groups(
+    swap_encoder_registry: &SwapEncoderRegistry,
+    grouped_swaps: &[SwapGroup],
+    router_address: &Bytes,
+) -> Result<Vec<EncodedSwapGroup>, EncodingError> {
+    map_on_threads(grouped_swaps, |grouped_swap| {
+        encode_swap_group(swap_encoder_registry, grouped_swap, router_address)
+    })
+}
```
tamaralipows, line 101: I removed the `get_swap_encoder` trait method from `StrategyEncoder` because I found it simpler to just pass the registry in here. Else this method would have looked like
```
encode_swap_groups<E: StrategyEncoder + Sync>(encoder: &E, groups: &[SwapGroup], ...) {
      map_on_threads(groups, |g| {
          let swap_encoder = encoder.get_swap_encoder(&g.protocol_system)?;
          ...
      })
  }
```

crates/tycho-execution/src/encoding/strategy_encoder.rs
```diff
@@ -17,14 +16,4 @@ pub(crate) trait StrategyEncoder {
...
-
-    /// Retrieves the swap encoder for a specific protocol system.
-    ///
-    /// # Arguments
-    /// * `protocol_system` - The identifier of the protocol system (e.g., "uniswap_v2")
-    ///
-    /// # Returns
-    /// * `Option<&Box<dyn SwapEncoder>>` - The swap encoder for the protocol if available
-    #[allow(clippy::borrowed_box)]
-    fn get_swap_encoder(&self, protocol_system: &str) -> Option<&Box<dyn SwapEncoder>>;
```
carloszanella, line 29: not sure why this was removed

crates/tycho-execution/src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -8,18 +8,83 @@ use crate::encoding::{
...
+    executor_address: Bytes,
+    protocol_data: Vec<u8>,
+}
+
+/// Encodes the swaps of one group into the protocol data its executor expects.
+fn encode_swap_group(
+    swap_encoder_registry: &SwapEncoderRegistry,
+    grouped_swap: &SwapGroup,
+    router_address: &Bytes,
+) -> Result<EncodedSwapGroup, EncodingError> {
```
carloszanella, line 32: was this a pure refactor into a function?

carloszanella (approved):
adding some more comments next but already approving. I checked on Fynd side and this should work but probably we should limit the amount of OS threads this can use

crates/tycho-execution/src/encoding/evm/tycho_encoders.rs
```diff
@@ -78,16 +78,13 @@ impl TychoRouterEncoder {
...
         &self,
         solutions: Vec<Solution>,
     ) -> Result<Vec<EncodedSolution>, EncodingError> {
-        let mut result: Vec<EncodedSolution> = Vec::new();
-        for solution in solutions.iter() {
-            let encoded_solution = self.encode_solution(solution)?;
-            result.push(encoded_solution);
-        }
-        Ok(result)
+        map_on_threads(&solutions, |solution| self.encode_solution(solution))
```
carloszanella, line 116: should we maybe limit threads here somehow? maybe have a conservative default and allow for overriding it via env vars?

---

**[BIZ `#4`]**

**What changed**: The two thread scopes nest, and each RFQ swap already runs its `block_on` on a further fresh thread through `on_blocking_thread`. One `encode_solutions` call can now hold roughly solutions x groups x RFQ swaps OS threads at once, with no cap tied to batch size or core count.

**Where**: `crates/tycho-execution/src/encoding/evm/tycho_encoders.rs:87`, `crates/tycho-execution/src/encoding/evm/strategy_encoder/strategy_encoders.rs:78–86`, `crates/tycho-execution/src/encoding/evm/utils.rs:153–160`

- The largest batch a caller such as Fynd passes to `encode_solutions`, and what that batch size costs in live threads.
- Each RFQ swap encoder holds its own single-worker tokio runtime created once in `SwapEncoderRegistry`; confirm concurrent `Handle::block_on` calls from many scope threads against that one runtime do not serialize the quote requests.

crates/tycho-execution/src/encoding/evm/strategy_encoder/strategy_encoders.rs
carloszanella, line 101: this sounds better IMO. related to the previous comment

---

**[DESIGN `#1`]**

**Root cause**: `map_on_threads` spawns one raw OS thread per item with no cap and no pool, and the diff applies it at two nesting levels. `encode_solutions` spawns a thread per solution; each of those spawns a thread per swap group; each RFQ group then calls `on_blocking_thread` (`utils.rs:153`), which spawns a third. `encode_solutions` takes a caller-supplied `Vec<Solution>` of any length, so nothing bounds the product.

- **Architecture**: a batch of 200 solutions with 4 groups each holds roughly 1600 live threads at ~2 MB of reserved stack apiece, `crates/tycho-execution/src/encoding/evm/utils.rs:177–181`
- **Behavior**: the common non-RFQ path pays the tax too — a split route of N UniswapV2/V3 groups does no network I/O at all, so the spawn-and-join per group now dominates what used to be microseconds of pure CPU encoding, `crates/tycho-execution/src/encoding/evm/strategy_encoder/strategy_encoders.rs:78–86`
- **Design**: parallelisation is unconditional, but only the four `rfq:` encoders block on I/O, so the strategy layer spends threads without knowing whether any group needs a quote, `crates/tycho-execution/src/encoding/evm/strategy_encoder/strategy_encoders.rs:78–86`

**Proposed fix**: Take the threaded path only when at least one item actually blocks, and bound it when it is taken. Give `SwapEncoder` a default-`false` predicate such as `fn blocks_on_quote(&self) -> bool`, overridden by the four RFQ encoders, so `encode_swap_groups` runs the plain loop for a pure on-chain route — that removes the common-path tax and keeps the "which protocols do network I/O" knowledge on the encoders instead of re-listing `rfq:` prefixes in the strategy layer. For the blocking case, cap the fan-out with a small reusable pool or a counting semaphore that runs items above the cap on the calling thread. Rayon is the wrong primitive: its pool is sized to CPU count and these threads block on network I/O, so it would re-serialise the quote waits. A further alternative worth weighing — every RFQ encoder already owns a tokio runtime and only calls `block_on`, so joining the futures on that one runtime would give the same overlap with no thread stacks at all.

crates/tycho-execution/src/encoding/evm/testing_utils.rs
```diff
@@ -96,3 +101,111 @@ impl IndicativelyPriced for MockRFQState {
...
+    async fn request_signed_quote(
+        &self,
+        params: GetAmountOutParams,
+    ) -> Result<SignedQuote, SimulationError> {
+        tokio::time::sleep(self.delay).await;
+        self.quote
+            .request_signed_quote(params)
+            .await
+    }
+}
```
carloszanella: less code better code

---

**[DESIGN `#2`]**

**Root cause**: The new decorator exists only to add `tokio::time::sleep(self.delay)` before `MockRFQState::request_signed_quote`, but to get there it re-implements the whole `ProtocolSim` trait as pass-throughs. Nine of the ten forwarded methods land on `MockRFQState` bodies that just `panic!` (`testing_utils.rs:31–79`), so the delegation carries no behavior — and two of them are wrong if ever reached.

- **Architecture**: a decorator whose nine forwarding methods are unreachable dead code, `crates/tycho-execution/src/encoding/evm/testing_utils.rs:110–165`
- **Complexity**: ~78 lines and one extra typetag'd trait impl to express one field and one `sleep`, `crates/tycho-execution/src/encoding/evm/testing_utils.rs:105–183`
- **Behavior**: `clone_box`, `as_any`, `as_any_mut` and `eq` all forward to the inner `MockRFQState`, so a clone silently loses the delay, a downcast to `DelayedRFQState` never succeeds, and the type compares equal to an undelayed mock. If any encoding path clones the state before requesting the quote, both new tests pass instantly and prove nothing, `crates/tycho-execution/src/encoding/evm/testing_utils.rs:151–165`

**Proposed fix**: Delete `DelayedRFQState` and add `#[serde(default)] pub delay: Duration` to `MockRFQState`, with its own `request_signed_quote` sleeping for it. A zero default keeps every existing call site behaving exactly as today; the cost is adding `delay: Duration::ZERO` (or `..Default::default()`) to the ~7 existing `MockRFQState` literals in `protocol_integration_tests.rs` and `liquorice.rs`. This resolves all three symptoms at once: there is no decorator left to be dead code, no second trait impl, and no forwarding methods that can drop the delay on clone or downcast.

crates/tycho-execution/src/encoding/strategy_encoder.rs
```diff
@@ -17,14 +16,4 @@ pub(crate) trait StrategyEncoder {
...
 }
```
carloszanella: Is the trait still justified?

---

**[DESIGN `#5`]**

Removing `get_swap_encoder` is right — all three impls were the same registry delegation and nothing outside the module called it — but the trait is left with one method and three concrete implementors held in named `TychoRouterEncoder` fields, so it is worth asking whether it still earns its keep.

crates/tycho-execution/src/encoding/evm/utils.rs
```diff
@@ -162,6 +162,34 @@ where
...
+        let mut results = Vec::with_capacity(handles.len());
+        for handle in handles {
+            let result = handle
+                .join()
+                .map_err(|_| EncodingError::FatalError("encoding thread panicked".to_string()))??;
+            results.push(result);
+        }
+        Ok(results)
+    })
+}
```
carloszanella, line 255: sounds better to be consistent with errors

---

**[IMPL `#1`]**

**Category**: bug

**Root cause**: `map_on_threads` short-circuits `[item]` to run `f` on the calling thread. That makes one input take a different code path from two, and the two paths do not behave the same: a panic on the fast path unwinds into the caller, while the threaded path converts it to `EncodingError::FatalError("encoding thread panicked")` and discards the panic payload.

**Plain English**: An integrator encodes one route and a bug inside an encoder makes the process abort with a stack trace and a readable message. The same integrator encodes two routes in one call, the same bug fires, and instead they get a generic error saying a thread panicked, with no indication of which route failed or why. Two identical faults should surface the same way, and the reported error should carry the original message.

- **Behavior**: panic reporting depends on item count, and the payload is thrown away in the threaded case, `crates/tycho-execution/src/encoding/evm/utils.rs:184–186`
- **Complexity**: three extra lines plus a documented rule the general `thread::scope` path already handles correctly for one item, `crates/tycho-execution/src/encoding/evm/utils.rs:174–176`
- **Documentation**: the two call-site docs promise "one thread per group" and "one thread each", which the fast path contradicts, `crates/tycho-execution/src/encoding/evm/strategy_encoder/strategy_encoders.rs:74`, `crates/tycho-execution/src/encoding/evm/tycho_encoders.rs:81`

**Suggestion**: Drop the fast path so every non-empty input takes one code path, and format the joined `Box<dyn Any>` payload into the `FatalError` message when it is a `&str` or `String`. The cost is one thread spawn per single-swap encode; every RFQ encoder already spawns its own thread via `on_blocking_thread` (`swap_encoder/bebop.rs:99`, `hashflow.rs:60`, `liquorice.rs:80`), so no tokio-context assumption depends on `f` running on the caller. Confirm no caller relies on thread-local state during encoding. Removing the branch also makes the two call-site docs true as written.

crates/tycho-execution/src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -364,50 +351,17 @@ impl StrategyEncoder for SplitSwapStrategyEncoder {
...
-                }
-            }
+        let encoded_groups =
+            encode_swap_groups(&self.swap_encoder_registry, &grouped_swaps, &self.router_address)?;
 
+        let mut swaps = vec![];
+        for (grouped_swap, encoded_group) in grouped_swaps.iter().zip(encoded_groups) {
             let swap_data = self.encode_swap_header(
                 get_token_position(&tokens, &grouped_swap.token_in)?,
                 get_token_position(&tokens, &grouped_swap.token_out)?,
```
carloszanella: **[IMPL `#2`]**

**Category**: perf

**Description**: `encode_swap_groups` runs to completion before `get_token_position` is checked, and `encode_solutions` encodes all solutions before returning the first failure, so a solution that fails validation or token lookup no longer stops the sibling groups' and solutions' RFQ round trips. Those are real HTTP requests to market makers that count against rate limits and produce signed quotes nobody will use.

**Suggestion**: Run the cheap, non-blocking checks — `get_token_position` for every group, and `validate_solution` for every solution — before calling `map_on_threads`, so a bad input is rejected without requesting any quote.

crates/tycho-execution/src/encoding/strategy_encoder.rs
carloszanella: yeah I think it could've been better structured in commits

---

**[PERSONAL `#2`]**

**Description**: The commit says `perf`, the PR says `refactor`, and it is in fact both. On top of that the commit removes `StrategyEncoder::get_swap_encoder` and its three impls, which the message does not mention at all — a trait shrinking is exactly the kind of change I need called out.

**Suggestion**: Pick one honest type per commit (see the split above) and name the trait method removal in the body, with the reason: the free function takes the registry directly, so no impl needs it any more.

crates/tycho-execution/src/encoding/evm/testing_utils.rs
```diff
@@ -96,3 +101,111 @@ impl IndicativelyPriced for MockRFQState {
...
+            protocol_system: "rfq:bebop".to_string(),
+            ..Default::default()
+        },
+        default_token(token_in),
+        default_token(token_out),
+        BigUint::ZERO,
+    )
+    .with_estimated_amount_in(BigUint::from(1_000u64))
+    .with_protocol_state(Arc::new(state))
+}
```
carloszanella: Agreed but maybe we dont need this trait at all

---

**[PERSONAL `#4`]**

`testing_utils.rs` exists for what integration tests import — its own first line says so, and it is gated behind the non-default `test-utils` feature — but `DelayedRFQState` and `delayed_bebop_swap` are used only by unit tests inside the crate, so put them in a `#[cfg(test)]` helper beside those tests.

crates/tycho-execution/src/encoding/evm/utils.rs
```diff
@@ -162,6 +162,34 @@ where
     })
 }
 
+/// Runs `f` on every item, each on its own OS thread, and returns the results in input order.
+///
+/// A single item runs on the calling thread. Any item's error is returned; earlier items win.
+pub(crate) fn map_on_threads<T, R, F>(items: &[T], f: F) -> Result<Vec<R>, EncodingError>
```
carloszanella, line 195: I was also curious about the placement here but up to you, I dont know the codebase enough

---

**[PERSONAL `#5`]**

It is a generic threading primitive with nothing EVM about it, living beside `bytes_to_address` and `EVMProvider`; move it to a shared helper module.

crates/tycho-execution/CLAUDE.md
```diff
@@ -229,6 +229,12 @@ group is PLE-encoded (`[len: u16][data]...`); Ekubo uses concatenation instead (
 | `SequentialSwapStrategyEncoder` | `sequentialSwap` / `Permit2` / `UsingVault` | Validates path connectivity, groups by protocol, PLE-encodes each group with executor header                                       |
 | `SplitSwapStrategyEncoder`      | `splitSwap` / `Permit2` / `UsingVault`      | Builds token array [tokenIn, intermediaries, tokenOut], encodes token indices + split percentages (U24) + executor + protocol data |
 
+**Parallel encoding**: `encode_swap_groups` (`strategy_encoders.rs`) encodes every swap group of a solution on its
+own OS thread, and `TychoRouterEncoder::encode_solutions` does the same for every solution. Both use
+`map_on_threads` (`evm/utils.rs`) and keep input order. This bounds a route with several RFQ hops (each a network
+round trip for the signed quote) by its slowest hop instead of the sum. A single group or solution runs on the
+calling thread.
```
carloszanella: **[DOCS `#1`]**

**Description**: The "RFQ hops are a network round trip, so encode them at the same time" explanation appears in the CLAUDE.md paragraph, the `encode_swap_groups` docstring, the `encode_solutions` docstring, and both test docstrings — each a slightly different wording of the same thing. Five copies drift. The CLAUDE.md paragraph is the worst of them: it repeats the rationale, ends with an implementation detail of `map_on_threads` ("A single group or solution runs on the calling thread") that belongs next to the helper, and documents a threading detail at a level the rest of that file does not. It reads as a summary of why the PR was made rather than a description of the architecture.

**Suggestion**: Keep the rationale in one place — the `map_on_threads` docstring — and cut the CLAUDE.md paragraph to the structural fact and the pointer, e.g. "**Parallel encoding**: `encode_swap_groups` (`strategy_encoders.rs`) and `TychoRouterEncoder::encode_solutions` encode groups and solutions in parallel via `map_on_threads` (`evm/utils.rs`), preserving input order." Drop the rationale sentence and the calling-thread sentence.

carloszanella (commented):
Review comments below.

---

up to you to decide whether the two are worth keeping

---

**[DESIGN `#4`]**

**Description**: The diff collapses three copies of the "encode each swap, split off the initial one, PLE-encode the rest" loop into `encode_swap_group`, but leaves the fourth copy in `encode_executor_calldata` untouched. The same knowledge is now encoded in two modules, so a future change to group encoding has to be applied twice.

**Suggestion**: Widen `encode_swap_group` to take `router_address: Option<&Bytes>` and call it from `encode_executor_calldata` too. The two bodies are not identical today — the executor copy always uses `ple_encode` and skips the `NON_PLE_ENCODED_PROTOCOLS` branch — so decide deliberately whether that difference is intended before merging them.

crates/tycho-execution/src/encoding/evm/tycho_encoders.rs
```diff
@@ -343,7 +340,47 @@ mod tests {
...
+        use alloy::hex::encode;
+
         use super::*;
+        use crate::encoding::evm::testing_utils::delayed_bebop_swap;
+
+        /// Two solutions each hold one RFQ swap that waits 300ms for its quote. Encoded together
+        /// they finish well before the 600ms a one-after-the-other encoding needs, and stay in
+        /// input order.
+        #[test]
+        fn test_encode_solutions_encodes_rfq_solutions_at_the_same_time() {
```
carloszanella: fucking robot...

---

**[DOCS `#3`]**

The crate already says "parallel" (`tycho_encoders.rs:458`, `model/src/lib.rs:118`); `..._at_the_same_time` is a longer synonym for it. Rename to `test_sequential_swap_encodes_rfq_hops_in_parallel` and `test_encode_solutions_encodes_rfq_solutions_in_parallel`.

crates/tycho-execution/src/encoding/evm/strategy_encoder/strategy_encoders.rs
carloszanella: Probably not a huge impact but sounds like an easy - even if small - win

crates/tycho-execution/CLAUDE.md
carloszanella: Makes sense to avoid so much copies of the same content

crates/tycho-execution/src/encoding/strategy_encoder.rs
carloszanella: I saw it later you added a comment in the PR, my bad

crates/tycho-execution/src/encoding/evm/testing_utils.rs
tamaralipows: Done in 26545d83e. `DelayedRFQState` and its two trait impls are gone. `MockRFQState` now has a `#[serde(default)] pub delay: Duration` and sleeps for it in `request_signed_quote`. The 11 existing literals gained `..Default::default()`, so they all keep a zero delay and behave as before.

One correction on the behaviour symptom: the forwarded `clone_box`, `as_any`, `as_any_mut` and `eq` did not drop the delay silently. `MockRFQState` panics in all four, so any of those calls would have aborted the test rather than passed it. The dead-code and complexity points stand on their own.

crates/tycho-execution/src/encoding/evm/strategy_encoder/strategy_encoders.rs
tamaralipows: Done in 5420627af. `encode_solutions` runs `validate_solution` over the whole batch before `map_on_threads`, and `SplitSwapStrategyEncoder` looks the token indices up before `encode_swap_groups`.

New test `test_encode_solutions_rejects_an_invalid_solution_before_requesting_quotes` asserts that a batch holding one invalid solution fails in less than the 300ms RFQ delay. I removed the fix and confirmed the test fails, then restored it.

One note on the `get_token_position` half: it cannot fail today. `tokens` is built from the grouped swaps' own `token_in` and `token_out` plus the solution's, so every lookup hits. The reorder is still the correct order, but the saved quote requests come from `validate_solution`.

crates/tycho-execution/CLAUDE.md
tamaralipows: Done in bdf847598. The rationale now sits only on the `map_on_threads` docstring. `encode_swap_groups` and `TychoRouterEncoder::encode_solutions` point at it, and the CLAUDE.md paragraph is your two-line version.

crates/tycho-execution/src/encoding/evm/tycho_encoders.rs
tamaralipows: Done in d741a6828. Renamed to `test_sequential_swap_encodes_rfq_hops_in_parallel` and `test_encode_solutions_encodes_rfq_solutions_in_parallel`.

crates/tycho-execution/src/encoding/evm/tycho_encoders.rs
tamaralipows: Done in 5dcd2dc76, with a constant instead of an env var. `map_on_threads` now runs at most `MAX_ENCODING_THREADS` (32) worker threads that pull items from a shared index, so a batch of 200 solutions holds 32 encoding threads, not 200. We can make it configurable when someone needs to tune it.

On the second point: concurrent `Handle::block_on` calls against the one single-worker runtime do not serialize the quote requests. `block_on` polls the future on the calling thread; the runtime's worker only drives the I/O and timer drivers. `test_encode_solutions_encodes_rfq_solutions_in_parallel` runs two 300ms quotes in ~0.3s, which fails if they serialize.

crates/tycho-execution/src/encoding/evm/strategy_encoder/strategy_encoders.rs
tamaralipows: Done in 5dcd2dc76.

- `SwapEncoder` has `blocks_on_quote()`, default `false`; Bebop, Hashflow, Liquorice and Metric return `true` (the four that call `on_blocking_thread` inside `encode_swap`).
- `encode_swap_groups` and `encode_solutions` run a plain loop when no group or solution blocks, so a pure UniswapV2/V3 route spawns no threads.
- When the threaded path is taken, `map_on_threads` caps at 32 workers pulling items from a shared index. Order and earliest-error-wins behavior are unchanged.

Mutation-checked: forcing Bebop's `blocks_on_quote` to `false` fails both `_in_parallel` timing tests. I stayed with OS threads over joining futures on the encoder's runtime — the four encoders each own a separate runtime, so there is no single place to join across protocols.

crates/tycho-execution/src/encoding/evm/utils.rs
tamaralipows: Done in f60ed7baa. `map_on_threads` and `on_blocking_thread` format the panic payload's `&str`/`String` message into the `FatalError`, so both paths report the original message; a new test asserts it survives.

I kept the single-item fast path: it is the cheap common case, and skipping thread spawns is the direction of the fan-out fix. The docstrings that promised "one thread per group" were already rewritten in bdf847598.

crates/tycho-execution/src/encoding/strategy_encoder.rs
tamaralipows: yeah agreed with this :[ will be betta

crates/tycho-execution/src/encoding/evm/testing_utils.rs
tamaralipows: Dissolved by the DESIGN 2 fix (26545d83e): `DelayedRFQState` is gone and the delay is a `MockRFQState` field, and `MockRFQState` must stay in `testing_utils.rs` for the integration tests. `delayed_bebop_swap` is the only unit-test-only item left, and it's 25 lines beside the struct it configures — not worth its own `#[cfg(test)]` module.

crates/tycho-execution/src/encoding/evm/utils.rs
tamaralipows: Leaving it. `on_blocking_thread` and `create_encoding_runtime` sit in the same file and are equally non-EVM; moving one of the three threading helpers away from the other two makes the grouping worse, and nothing outside `encoding/evm` uses any of them. If a second module ever needs them, all three can move together.

crates/tycho-execution/src/encoding/strategy_encoder.rs
tamaralipows: It wasn't — removed in abc01f70a. `encode_strategy` is now an inherent method on the three concrete encoders; the trait was `pub(crate)`, so the crate's public API is unchanged. Updated the crate CLAUDE.md diagram and the two GitBook mentions.

tamaralipows:
Re **[DESIGN `#4`]** (the duplicated group-encoding loop in `encode_executor_calldata`):

Resolved by deletion instead of merging: fd14a7b9d removes `TychoExecutorEncoder`, `TychoExecutorEncoderBuilder`, and the `tycho-encode` CLI's `tycho-executor` subcommand. We no longer support executor-direct encoding — Turbine was our last client using it.

The Ekubo difference the comment flagged was a defect only in the deleted copy: it always PLE-encoded, while the `EkuboExecutor` parses hops by a fixed byte stride. The surviving `encode_swap_group` (the router path) already handles `NON_PLE_ENCODED_PROTOCOLS`, so nothing remains to fix. The user-facing docs never documented the executor-direct path, so no doc changes were needed beyond the crate CLAUDE.md.

PR #1371, opened 2026-09-01
feat(execution): add per-client positive slippage exemptions

tamaralipows:
## What

Lets the FeeCalculator skip positive slippage capture for selected clients while capture stays enabled globally.

## How it works

The FeeCalculator already resolves a client for every swap: the signed `clientFeeReceiver` from the client fee params, or `tx.origin` when none is supplied. This change adds a `mapping(address => bool)` of exemptions on top of that resolution, settable per client under `ROUTER_FEE_SETTER_ROLE`.

For an exempt client:

1. The surplus above `expectedAmountOut` stays in the swap output instead of going to the router.
2. Fees compute on `actualAmountOut` (the full output), exactly as if capture were disabled.
3. With no fees to take, the output no longer has to route through the router after the final swap, saving gas.

Everyone else is unaffected. The exemption is FeeCalculator state, so activating it is a FeeCalculator redeploy behind the router's existing timelock — the router itself is untouched.

This pairs with propeller-heads/fynd#491, which encodes transactions carrying server-signed zero-fee client params so a designated request flow resolves to an exempt client address.


Troshchk (approved):
Looks good to me, thanks!

Lets keep in mind to update the model too?

louise-poole (approved):
Looks good to me! I agree with Ksenia that we should update the model too though.

PR #1378, opened 2026-09-02
chore(execution): verify contracts on Blockscout chains

louise-poole:
Verification on Robinhood Chain went through hardhat-verify, which cannot set a browser `User-Agent` — Cloudflare answers 403 to everything else.

  Blockscout chains now verify through the explorer's native v2 standard-input API, with a browser `User-Agent` and no API key:

  - Standard JSON input comes from the contract's own `.dbg.json` build-info reference, not a filename-suffix scan over build-info (which also matched `IFeeCalculator.sol` and stale compiles).
  - Constructor arguments encode from the artifact ABI, so one path serves the router's six arguments and the fee calculator's one.
  - Blockscout returns HTTP 500 for an address it has not indexed as a contract; a pre-flight check reports that separately from missing bytecode.
  - Both deploy scripts skip a CREATE2 deployment that already exists, so retrying verification is just a re-run.

### DIFF
--- a/crates/tycho-execution/contracts/scripts/utils.js
+++ b/crates/tycho-execution/contracts/scripts/utils.js
@@ -91,8 +91,12 @@ function getHardhatStandardJsonInput(contractsDir, contractFqn) {
 /**
  * Blockscout's verification endpoints answer HTTP 500 for any address it does
  * not hold as a contract, so check the two reasons that happens up front.
+ *
+ * @returns {boolean} whether Blockscout already holds the contract as verified.
+ *     Submitting again for such an address is rejected, and deploy scripts are
+ *     meant to be re-runnable.
  */
-async function assertVerifiable(config, address) {
+async function checkVerifiable(config, address) {
     if ((await ethers.provider.getCode(address)) === "0x") {
         throw new Error(
             `No contract deployed at ${address}. The address is derived from ` +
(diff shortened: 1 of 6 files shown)


Troshchk (commented):
Thank you!
Is this still needed: https://github.com/propeller-heads/tycho/blob/6ad86479f81925f4526b5e70ee2ceb34ac919ef3/crates/tycho-execution/contracts/hardhat.config.js#L8 ?

crates/tycho-execution/contracts/scripts/utils.js
```diff
@@ -1,9 +1,224 @@
...
+        chainId: 4663,
+        apiBase: "https://robinhoodchain.blockscout.com/api/v2",
+        browserUrl: "https://robinhoodchain.blockscout.com/",
+    },
+};
+
+// Cloudflare answers 403 to requests carrying a curl or node User-Agent.
+const USER_AGENT =
+    "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 " +
+    "(KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36";
```
tamaralipows, line 25: naughty naughty

tamaralipows (approved):
Thank you!

PR #1379, opened 2026-09-02
feat(simulation): gate pending indexers on a state capability

michcio1234:
*🤖 Generated by AI.*

`with_pending_indexer` used to refuse any extractor named `vm:*`. The name is not the property that matters: `vm:curve` decodes into the native `CurveState` and qualifies, while `erc4626`, `fluid_v1` and `uniswap_v4_hooks` read the shared VM database despite native-looking names. The base branch dropped that gate and put nothing in its place, so an indexer registered for a state that cannot absorb its deltas had them applied to nothing, and the pool kept quoting confirmed prices with no error.

## What this adds

`ProtocolSim::transitions_from_delta_alone(delta)`. A state reports whether its `delta_transition` — and the quote taken after it — depend on nothing but the delta. `apply_deltas_ephemeral` asks per pool and fails the update when the answer is no, naming the pool and the extractor.

The property is per state and per delta, not per protocol name. `uniswap_v4` and `uniswap_v4_hooks` decode into the same type and differ only in whether a VM-backed hook is attached. Fluid and Curve qualify only while the indexer supplies the attribute they rebuild from (`pool_reserves_adjusted`, `pool_state_adjusted`); without it both fall back to a VM read that only holds confirmed state.

| Reports | States |
|---|---|
| `true` | 15 native states (Uniswap V2/V3, Ekubo, Aerodrome, Rocketpool, …) |
| `true` only with its attribute | `FluidV1`, `CurveState` |
| delegates to its hook | `UniswapV4State` |
| `false` (default) | `EVMPoolState`, `ERC4626State`, `BalancerV3State`, `GenericVMHookHandler`, RFQ states |

The default is `false`, so a protocol is refused from the pending path until its author has checked both the transition and the quote. That costs one opt-in per native state and makes the failure loud instead of silent.

`HookHandler` gets the same method so a V4 pool can defer to its hook. `SwapQuoter` gets it too, and the blanket `ProtocolSim` impl forwards, so protocols moving to that trait are not silently exempt.

`build_with_pending` separately rejects an indexer registered for an exchange no `exchange()` call added — the other way deltas reach no state.

## Breaking change

`ProtocolSim` and `SwapQuoter` gain a defaulted method. Downstream native states must return `true` to take part in pending-block simulation; nothing else breaks at compile time.

## Verification

`cargo +nightly fmt` and `cargo clippy --workspace --all-targets --all-features` clean. `cargo nextest run -p tycho-simulation --all-features`: 1121 passed, 1 failed. That failure, `vm::state::tests::test_engine_block_advance_invalidates_cached_limits`, reproduces on `feat/curve-core-crate` with this branch's changes stashed. `cargo nextest run -p tycho-common`: 100 passed.

`tycho-storage`, `tycho-indexer` and `tycho-execution` tests did not run locally — the test link needs `libpq`, which is not installed on this machine. They compile under `clippy --all-targets`. `serial_db` and `#[ignore]`d tests were skipped (no DB, no `RPC_URL`).

Six new tests: the two `apply_deltas_ephemeral` outcomes plus the missing-state skip, the Fluid and Curve attribute conditions, and the `build_with_pending` rejection.

### DIFF
--- a/crates/tycho-simulation/src/evm/protocol/uniswap_v4/hooks/hook_handler.rs
+++ b/crates/tycho-simulation/src/evm/protocol/uniswap_v4/hooks/hook_handler.rs
@@ -65,6 +65,20 @@ pub trait HookHandler: Debug + Send + Sync + 'static {
         tokens: &HashMap<Bytes, Token>,
         balances: &Balances,
     ) -> Result<(), TransitionError>;
+
+    /// Whether this handler can serve the block `delta` describes after
+    /// [`delta_transition`](Self::delta_transition), reading nothing beyond its arguments.
+    ///
+    /// Mirrors
+    /// [`ProtocolSim::transitions_from_delta_alone`](tycho_common::simulation::protocol_sim::ProtocolSim::transitions_from_delta_alone)
+    /// for the hook half of a pool: a
+    /// handler that simulates against the shared VM database returns `false`, because that
+    /// database only ever holds confirmed state.
+    #[allow(unused)]
+    fn transitions_from_delta_alone(&self, delta: &ProtocolStateDelta) -> bool {
+        false
+    }
+
     fn clone_box(&self) -> Box<dyn HookHandler>;
 
     fn as_any(&self) -> &dyn std::any::Any;
--- a/crates/tycho-common/src/simulation/protocol_sim.rs
+++ b/crates/tycho-common/src/simulation/protocol_sim.rs
@@ -376,6 +376,23 @@ pub trait ProtocolSim: fmt::Debug + Send + Sync + 'static {
         balances: &Balances,
     ) -> Result<(), TransitionError>;
 
+    /// Whether [`delta_transition`](Self::delta_transition) can bring this state to the block
+    /// `delta` describes, and quote it afterwards, without reading anything beyond its own
+    /// arguments.
+    ///
+    /// `false` means the transition or a quote taken after it falls back to a source the caller
+    /// does not supply — for EVM protocols, contract storage held in a database that only the
+    /// confirmed indexing pipeline updates. Applying `delta` to such a state layers new
+    /// attributes over stale storage and reports no error, so callers that transition a state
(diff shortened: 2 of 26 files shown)


crates/tycho-simulation/src/evm/protocol/uniswap_v4/hooks/hook_handler.rs
```diff
@@ -65,6 +65,20 @@ pub trait HookHandler: Debug + Send + Sync + 'static {
...
+    ///
+    /// Mirrors
+    /// [`ProtocolSim::transitions_from_delta_alone`](tycho_common::simulation::protocol_sim::ProtocolSim::transitions_from_delta_alone)
+    /// for the hook half of a pool: a
+    /// handler that simulates against the shared VM database returns `false`, because that
+    /// database only ever holds confirmed state.
+    #[allow(unused)]
+    fn transitions_from_delta_alone(&self, delta: &ProtocolStateDelta) -> bool {
+        false
+    }
```
PierreMkt, line 80: AI review pointed out that the angstrom hook is natively implemented so it should have the `true` override.

I checked locally and if you add a `test_apply_deltas_ephemeral_angstrom_hooked_v4_pool` test in `tycho-simulation/src/evm/decoder.rs`, it will pass before your PR and fail after. 
Roughly:

```
let mut pool = /* the UniswapV4State the state.rs test at :1654 builds */;
pool.set_hook_handler(Box::new(AngstromHookHandler::new(/* … */)));
let decoder = decoder_holding(Box::new(pool)).await;

decoder
    .apply_deltas_ephemeral(&reserve_deltas(), BlockHeader::default())
    .await
    .expect("an Angstrom pool rebuilds from its delta");
```

I have the test and the fix locally. I can commit and push it to your branch if you want

michcio1234: Yes please 🙏🏻

PR #1381, opened 2026-09-02
fix: share one protocol cache across all extractors

zizou0x:
## What

Creates and populates the `ProtocolMemoryCache` once in `build_all_extractors` and passes it into every `ExtractorFactory`, instead of each factory building and populating its own instance.

## Why

The cache holds the chain's full token and component sets (~10 GB on base dev). Since #1026, each extractor populated its own copy, so memory and the bulk startup DB read were multiplied by the extractor count. On rollout this OOM-killed the dev indexer pods on base, ethereum, and bsc (base: 8 extractors against a 14Gi limit; the pod died partway through building the second factory).

## Why sharing is safe

- The cache is insert-only and its entries are chain-level facts: token metadata comes from the contract itself, and components are static definitions keyed by id. An entry left over from a block that was never committed (or later reverted) is never queried again; it cannot produce a wrong answer.
- Component entries are namespaced by protocol system and reads are system-scoped, so one extractor cannot read another's components. Tokens are the only shared namespace, and those are chain-global by design — a token discovered by one extractor is the same token every other extractor needs.
- After a restart, the extractor replays from the last committed block and re-inserts the same entries idempotently, so reused entries match what replay would produce.
- All maps are `Arc<RwLock>`-protected, and one shared instance across all of a chain's extractors is the configuration that ran in production until #1026 merged; restarts reusing the live cache (unchanged here) already rely on the same properties.

### DIFF
--- a/crates/tycho-indexer/src/extractor/factory.rs
+++ b/crates/tycho-indexer/src/extractor/factory.rs
@@ -152,8 +152,9 @@ where
 /// with a new `ReorgBuffer` and DCI plugin — suitable for restart after failure.
 ///
 /// Reused across restarts:
-/// - `protocol_cache`: populated once at construction; Arc-based so cloning is cheap and all runs
-///   share the same live cache. The TTL mechanism refreshes stale entries automatically.
+/// - `protocol_cache`: one chain-wide instance shared by every extractor, populated once by the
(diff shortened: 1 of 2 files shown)


brunoguerios (approved):
LGTM ✅

PR #1382, opened 2026-09-02
fix: retry substreams auth failures once credential is proven

zizou0x:
## What this changes

`Unauthenticated` from the Substreams endpoint was the one gRPC status the stream never retried: it was returned to the consumer, which ends the extractor and exits the process. Every other status goes through the existing exponential backoff.

The stream now decides based on whether the endpoint has ever accepted the credential in this process:

- **No block received yet** — the status ends the stream, as before. A credential that has never delivered a block cannot be told apart from a misconfigured one, and the token is read once at startup, so retrying cannot fix it.
- **A block has been received** — the credential is proven, so the rejection is the endpoint's. The stream reconnects from its last cursor, up to `MAX_UNAUTHENTICATED_RETRIES` (5) consecutive attempts, which spans about three minutes on the existing backoff. The counter resets on the next block.

## Behaviour change to note

An `Unauthenticated` raised before the stream opens surfaces as an error from the initial call rather than as a stream error, so it previously fell into the connection-error branch and was retried indefinitely. It now ends the stream when no block has been received, which makes the fail-fast-on-misconfiguration path apply to both shapes of rejection instead of only one.

## Why

On 2026-09-02 the prod robinhood indexer exited twice, at 14:57:35 and 14:59:55 UTC, both times on `Stream terminated with error. error=status: Unauthenticated` from the `uniswap_v4` extractor. The endpoint accepted the same credential again on the third start, so the rejection lasted about two and a half minutes. The two exits cost roughly five and a half minutes of API downtime, since each restart reloads the token and protocol caches before the server binds. The `uniswap_v2` and `uniswap_v3` extractors hit a broken pipe in the same 5ms window and reconnected on their own.

Chains on other endpoints logged no stream errors in that window, and the same credential kept working for them throughout.

As a side effect the `substreams_failure{cause="unauthenticated"}` counter becomes observable — the process previously exited before the next scrape, so it has never been recorded.

## Tests

- `test_unauthenticated_before_first_block_is_fatal` — the stream errors and does not dial the endpoint again.
- `test_unauthenticated_after_first_block_reconnects` — the stream reconnects and resumes from the last cursor.
- Three cases covering the retry bound.

The mock Substreams server gained a per-request script and a response body that emits one block and then error trailers, which is the shape needed to reject a stream that was already delivering blocks.

### DIFF
--- a/crates/tycho-indexer/src/substreams/mock.rs
+++ b/crates/tycho-indexer/src/substreams/mock.rs
@@ -1,9 +1,11 @@
 //! Generic mock for the Substreams Stream/Blocks gRPC service.
 //!
-//! Captures every `Request` protobuf sent by the client and returns an empty
-//! stream (trailers-only `grpc-status: 0`), which makes `stream_blocks` yield
-//! `BlockResponse::Ended` and the runner exit cleanly.
+//! Captures every `Request` protobuf sent by the client and answers it from a
+//! script of [`MockResponse`]s, one per request. Requests beyond the end of the
+//! script get an empty stream (trailers-only `grpc-status: 0`), which makes
+//! `stream_blocks` yield `BlockResponse::Ended` and the runner exit cleanly.
 use std::{
+    collections::VecDeque,
     convert::Infallible,
     future::Future,
     net::SocketAddr,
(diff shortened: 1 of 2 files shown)


brunoguerios (approved):
LGTM ✅

PR #1400, opened 2026-09-03
fix: join Uniswap V4 tick store deltas by key instead of position

brunoguerios:
## Problem

On Base dev (2026-09-02/03) the indexer fires `extractor_revert_attr_miss{component_state_found="true"}`
about 3 times per 6 hours. Uniswap V4 `ticks/<idx>/net-liquidity` attributes carry the value of the
wrong tick, and a tick's first write gets the label `Update` instead of `Creation`. On revert the
indexer then finds no prior value. The wrong net-liquidity value also feeds simulation.

## Root cause

`map_ticks_changes` emits two tick deltas per `ModifyLiquidity` event, one for `tick_lower` and one
for `tick_upper`. Both deltas carry the same log ordinal. `store_ticks_liquidity` sorts them with
`sort_unstable_by_key` on that ordinal, so it can swap the tied pair. `collect_transaction_changes`
then paired store deltas with tick deltas by position. After a swap, the attribute name of
`tick_lower` received the store delta of `tick_upper`: wrong value and wrong ChangeType.

## Fix

`collect_transaction_changes` joins store deltas to tick deltas by `(store key, ordinal)`. The store
key `pool:<pool>:tick:<index>` with the ordinal is unique: one event cannot write the same tick
twice, and different events have different log ordinals. The attribute body is unchanged.

## Rollout

Both variants embed the corrected shared code, so both need new spkgs: `no-hooks v0.4.3` and
`with-hooks v0.7.1`.

1. Build and publish the two spkgs through the usual release flow.
2. Pin the new `no-hooks` spkg on Base dev first.
3. Watch `extractor_revert_attr_miss{namespace="dev-tycho", chain="base", component_state_found="true"}`
   stay flat for a few days. Baseline before the fix is about 3 per 6 hours. The "Revert Attribute
   Miss" alert covers a regression.
4. Roll out the remaining pins in `helm-configuration`, dev before prod.

Current `no-hooks` pins to update (21 files, 7 chains):

| Chain | dev | prod |
|---|---|---|
| base | `base-tycho-indexer` v0.4.1, `base-new-tycho-indexer` v0.4.1 | `tycho/base-tycho-indexer-1` v0.4.1, `-2` v0.4.1, `tycho-fynd/base-tycho-indexer` v0.4.1 |
| ethereum | `ethereum-tycho-indexer` v0.4.0, `-new-protocol` v0.4.0, `ethereum-staging` v0.4.0 | `tycho/ethereum-tycho-indexer-1` v0.4.0, `-2` v0.4.0, `tycho-fynd/ethereum-tycho-indexer` v0.4.0 |
| arbitrum | v0.4.2 | v0.4.2 |
| bsc | v0.4.1 | v0.4.1 |
| polygon | v0.4.2 | v0.4.2 |
| robinhood | v0.4.2 | v0.4.2 |
| unichain | v0.4.1 | v0.4.1 |

Current `with-hooks` pins to update (6 files): dev `ethereum` v0.7.0, `ethereum-new-protocol` v0.6.0,
`ethereum-staging` v0.7.0; prod `tycho/ethereum-tycho-indexer-1` and `-2` v0.7.0,
`tycho-fynd/ethereum` v0.7.0. The commented-out unichain `with-hooks` pins need no change.

Note that the arbitrum spkg paths keep a `no-hooks/` sub-directory, unlike the other chains.

## Notes for reviewers

The pool-liquidity zip in the same function stays positional. It is sound today, because
`PoolLiquidityChanged` produces one change per event with a unique ordinal.

The join uses `remove`, and panics if a tick delta has no matching store delta. Every
`store.add` call emits one delta, so the count and the keys always match. The panic marks a broken
invariant instead of dropping changes, which the previous `zip` did without a signal.

### DIFF
--- a/protocols/substreams/ethereum-uniswap-v4/shared/src/utils/protocol_changes.rs
+++ b/protocols/substreams/ethereum-uniswap-v4/shared/src/utils/protocol_changes.rs
@@ -4,7 +4,10 @@ use crate::pb::uniswap::v4::{
 };
 use itertools::Itertools;
 use std::{collections::HashMap, str::FromStr, vec};
-use substreams::{pb::substreams::StoreDeltas, scalar::BigInt};
+use substreams::{
+    pb::substreams::{StoreDelta, StoreDeltas},
+    scalar::BigInt,
+};
 use substreams_helper::hex::Hexable;
 use tycho_substreams::{balances::aggregate_balances_changes, prelude::*};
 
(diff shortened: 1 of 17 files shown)


protocols/substreams/ethereum-uniswap-v4/shared/src/utils/protocol_changes.rs
```diff
@@ -72,12 +75,33 @@ pub fn collect_transaction_changes(
...
+                tick_delta.pool_address.to_hex(),
+                tick_delta.tick_index
+            );
+            let store_delta = ticks_store_deltas_by_key
+                .remove(&(store_key.clone(), tick_delta.ordinal))
+                .unwrap_or_else(|| {
+                    panic!(
+                        "no tick store delta for key {store_key} at ordinal {}",
+                        tick_delta.ordinal
+                    )
```
brunoguerios, line 102: I'm not super comfortable with this `panic` here, so let me know if you have any suggestions of how we could handle this better 🙏

zach030 (approved):
LGTM! Can you bump all the substreams manifest versions?

protocols/substreams/ethereum-uniswap-v4/shared/src/utils/protocol_changes.rs
zach030: I'm fine with `panic`. It shouldn't error in theory, right? If it does, we should panic instead of failing silently.

PR #1401, opened 2026-09-03
feat(router-trades): release the spkg packages apart from the image

kayibal:
## Problem

The image packed a `.spkg` per chain, so **every chain's module hash was a property of the image** — and the image is rebuilt on every monorepo release (`build-and-push-tycho-router-trades` is `needs: release`, no path filter) from a floating base tag (`FROM rust:1.91-bookworm`). Consequences:

- An edit under `src/` gives all seven chains a new hash at once.
- So does a base-image refresh, with **nobody having touched this directory**.
- Recovery is not cheap: with `--on-module-hash-mismatch=error` every sink exits, and because the sink writes plain `INSERT`s each chain needs its rows deleted and a full re-sync from its deployment block (ethereum ~3.3M blocks, base ~20M, robinhood ~29M).
- There was also no way to run different versions per chain.

## Approach

Publish the packages separately and let each container fetch the one it is pinned to — the mechanism `protocols/substreams` already uses.

- **`release.sh`** (forked from `protocols/substreams/release.sh`, same rules) builds the wasm once with `--locked`, packs one `.spkg` per chain and uploads to `s3://repo.propellerheads-propellerheads/substreams/tycho-router-trades/<chain>-<version>.spkg`. The version comes from `tycho-router-trades/Cargo.toml`; a release needs a matching `tycho-router-trades-<semver>` tag on HEAD and a clean tree; **releases are immutable** (`put-object --if-none-match '*'`); without a tag it publishes an overwritable `pre.<sha>`. It prints each `db_out` module hash.
- **`release-router-trades.yaml`** — `workflow_dispatch` with an optional space-separated `chains` input, mirroring the existing release workflow.
- **The Dockerfile loses its Rust stage entirely.** The image is now the sink binary plus `psql`, the AWS CLI and the SQL, so it carries no module hash and rebuilding it cannot invalidate a cursor.
- **The entrypoint resolves `SPKG`**: a local path when that file exists, otherwise a key in `TYCHO_S3_BUCKET` (default `repo.propellerheads-propellerheads`) cached under `SPKG_CACHE_DIR`. Same rule as the indexer's `ensure_spkg`, so a pinned release and a bind-mounted local build both work.
- **Local runs**: `make pack-all` writes packages to `target/spkg/`, which `docker-compose.yaml` mounts.

Pinning lives in helm, per chain and per environment:

```yaml
spkgs:
  ethereum: substreams/tycho-router-trades/ethereum-v0.2.0.spkg
  bsc:      substreams/tycho-router-trades/bsc-v0.3.0.spkg
```

A chain's hash now moves only when someone changes that one line, and only for that chain.

## Verification

- `release.sh` passes `shellcheck` and `shfmt`; the workflow passes `actionlint` and `zizmor` (the first draft had a real `template-injection` on the `chains` input, now passed through `env:`).
- Image builds; `aws --version` reports 2.9.19 inside it.
- End-to-end run of the new resolution against a locally packed package: no S3 fetch (local path taken), `setup completed successfully`, and `output_module_hash: 299766f1…` read from the mounted `.spkg`.
- `docker compose config` valid.

## Migration — needs care exactly once

CI-published packages will not carry the module hashes the running image produced, because a wasm build is not byte-identical across environments. Pinning freshly built packages would therefore invalidate all seven cursors.

To migrate without a re-sync: extract the `.spkg` files from the **currently deployed image**, upload those exact bytes as the first pinned version, and pin them. The hashes then match the live `cursors_<chain>.id` and every chain resumes untouched.

## Follow-up

The helm side (the `spkgs` map, the `SPKG` env per sink container, and the `eks.amazonaws.com/role-arn` annotation the pod needs for S3) is a separate PR in `helm-configuration`, which must land together with this one.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -16,6 +16,7 @@ Reusable build/deploy steps are delegated to [propeller-heads/ci-cd-templates](h
 | `ci-substreams.yaml` | PR (`substreams/`), manual | Lint and unit tests for changed substreams packages |
 | `ci-substreams-integration.yaml` | PR (`protocols/substreams/`, `protocols/testing/`), manual | Full protocol integration tests against a live DB |
 | `ci-router-trades.yaml` | PR/push (`crates/tycho-execution/substreams/`), manual | Router-trades lint, WASM build, unit tests, and PostgreSQL pricing integration |
+| `release-router-trades.yaml` | Manual | Build and publish one router-trades `.spkg` per chain to S3, where the sink containers fetch the version pinned in helm |
 | `cd-deploy-dev.yaml` | Manual | Build a branch image and deploy it to dev |
 | `promote-to-prod.yaml` | Manual | Promote a tagged dev image to production |
 | `release.yaml` | GitHub Release published, manual | Upload binaries to GitHub + S3, publish crates to crates.io |
--- a/.github/workflows/release-router-trades.yaml
+++ b/.github/workflows/release-router-trades.yaml
@@ -1,0 +1,71 @@
+name: Release Router Trades Substreams
+
+# Builds one .spkg per chain and uploads it to S3. The sink containers fetch the package they are
+# pinned to, so this workflow — not the image build — is what changes what runs in production.
+#
+# A release needs the version in tycho-router-trades/Cargo.toml bumped, merged, and a matching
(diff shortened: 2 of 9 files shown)


kayibal (commented):
lgtm

carloszanella (approved):

carloszanella (approved):

PR #1402, opened 2026-09-03
feat(integration-test): allow --test-every-n-blocks with --partial-blocks

brunoguerios:
## Problem

On 2026-08-28 the three Base integration-test runners exhausted the Chainstack RPC plan (`-32005 RPS limit`). Since then every trace call fails, counts as a revert, and "Execution success rate low" fires for every Base protocol. With `--partial-blocks` a runner runs ~120–200 simulations per second.

## Root cause

`--test-every-n-blocks` caps the test rate, but clap rejected it with `--partial-blocks`: sampled mode fetched the target block by number, and a flashblock's pending state cannot be fetched by number.

## Fix

`fetch_sampled_block_by_number` picks the fetch path. Sampled runs without `--partial-blocks` (Robinhood) keep fetch-by-number. With `--partial-blocks` the runner keeps the pending/latest polling path, so sampling only gates which updates are tested. The clap conflict is removed.

## Notes for reviewers

Sampling is by block number, so every partial update of a sampled block is tested. That is bursty by design and stays representative of flashblock execution. Default `--test-every-n-blocks 1` is unchanged.

## Follow-ups

- Roll out on Base: propeller-heads/helm-configuration#1014 (dev), then prod.
- RPC transport errors still count as reverts.

### DIFF
--- a/crates/tycho-integration-test/src/main.rs
+++ b/crates/tycho-integration-test/src/main.rs
@@ -1021,13 +1021,11 @@ async fn process_update(
             let poll_interval = Duration::from_millis(cli.rpc_poll_interval_ms);
 
             let by_number =
-                fetch_sampled_block_by_number(cli.test_every_n_blocks, cli.partial_blocks);
+                should_fetch_block_by_number(cli.test_every_n_blocks, cli.partial_blocks);
 
             let block = if by_number {
-                // Sampled full-block mode: on fast chains the head is expected to be past the
-                // update, so the target block is fetched by number instead of racing the head.
-                // Sampled runs with --partial-blocks never take this path — a flashblock's
-                // pending state cannot be fetched by number — and use the polling path below.
+                // On fast chains the head is expected to be past the update, so fetch the target
+                // block by number instead of racing the head.
                 match await_target_block(
                     &rpc_tools,
                     update_block_number,
@@ -2035,12 +2033,12 @@ fn is_sampled_block(block_number: u64, interval: u64) -> bool {
     block_number.is_multiple_of(interval)
 }
 
-/// True when a sampled update's target block should be fetched by number (sampled full-block
-/// mode, e.g. Robinhood where the head races past the update). With --partial-blocks, sampling
-/// only gates WHICH updates are tested; block fetching stays on the pending/latest polling path
-/// because a flashblock's pending state cannot be fetched by number after the fact.
-fn fetch_sampled_block_by_number(test_every_n_blocks: u64, partial_blocks: bool) -> bool {
-    test_every_n_blocks > 1 && !partial_blocks
+/// True when a sampled update's target block should be fetched by number rather than polled for.
+///
+/// A flashblock's pending state cannot be fetched by number after the fact, so under
+/// `--partial-blocks` sampling only gates which updates are tested.
+fn should_fetch_block_by_number(interval: u64, partial_blocks: bool) -> bool {
+    interval > 1 && !partial_blocks
 }
 
 /// Selector of the priority-update-registry's `StaleUpdate()` error, the freshness guard of the
@@ -2172,7 +2170,7 @@ mod tests {
     use rstest::rstest;
 
     use super::{
-        fetch_sampled_block_by_number, is_oracle_stale_revert, is_sampled_block, pamm_venue, Cli,
+        is_oracle_stale_revert, is_sampled_block, pamm_venue, should_fetch_block_by_number, Cli,
     };
 
     #[rstest]
@@ -2231,17 +2229,9 @@ mod tests {
         assert_eq!(is_sampled_block(block, interval), expected);
     }
 
-    #[rstest]
-    #[case::sampled_full_block_mode(10, false, true)]
-    #[case::sampled_with_partial_blocks_polls(10, true, false)]
-    #[case::unsampled_polls(1, false, false)]
-    #[case::unsampled_with_partial_blocks_polls(1, true, false)]
-    fn sampled_fetch_by_number_only_without_partial_blocks(
-        #[case] test_every_n_blocks: u64,
-        #[case] partial_blocks: bool,
-        #[case] expected: bool,
-    ) {
-        assert_eq!(fetch_sampled_block_by_number(test_every_n_blocks, partial_blocks), expected);
+    #[test]
+    fn sampled_with_partial_blocks_polls_instead_of_fetching_by_number() {
+        assert!(!should_fetch_block_by_number(10, true));
     }
 
     #[test]
--- a/crates/tycho-integration-test/src/main.rs
+++ b/crates/tycho-integration-test/src/main.rs
@@ -1020,8 +1020,10 @@ async fn process_update(
 
             let poll_interval = Duration::from_millis(cli.rpc_poll_interval_ms);
 
-            let block = if fetch_sampled_block_by_number(cli.test_every_n_blocks, cli.partial_blocks)
-            {
+            let by_number =
+                fetch_sampled_block_by_number(cli.test_every_n_blocks, cli.partial_blocks);
+
+            let block = if by_number {
                 // Sampled full-block mode: on fast chains the head is expected to be past the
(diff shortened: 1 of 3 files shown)


crates/tycho-integration-test/src/main.rs
```diff
@@ -2216,9 +2231,22 @@ mod tests {
...
+    #[case::sampled_with_partial_blocks_polls(10, true, false)]
+    #[case::unsampled_polls(1, false, false)]
+    #[case::unsampled_with_partial_blocks_polls(1, true, false)]
+    fn sampled_fetch_by_number_only_without_partial_blocks(
+        #[case] test_every_n_blocks: u64,
+        #[case] partial_blocks: bool,
+        #[case] expected: bool,
+    ) {
+        assert_eq!(fetch_sampled_block_by_number(test_every_n_blocks, partial_blocks), expected);
+    }
```
louise-poole: 🤖 The four cases are the full truth table of `n > 1 && !partial`, so the test asserts the expression back at itself — it can only fail if someone edits the expression, and they'd edit the table with it. The behavior the PR changes, which of the two fetch paths `process_update` takes, has no test.

Cut the table to the one case that encodes the new rule — sampled plus `--partial-blocks` stays on the polling path.

crates/tycho-integration-test/src/main.rs
```diff
@@ -2227,7 +2255,9 @@ mod tests {
             "--partial-blocks",
             "--test-every-n-blocks",
             "10",
-        ]);
-        assert!(result.is_err());
+        ])
+        .expect("--test-every-n-blocks must be accepted alongside --partial-blocks");
+        assert!(cli.partial_blocks);
+        assert_eq!(cli.test_every_n_blocks, 10);
```
louise-poole: Up to you...

---

🤖 The `expect` is the assertion; `assert!(cli.partial_blocks)` and `assert_eq!(cli.test_every_n_blocks, 10)` only check that clap stored the two values it was handed. Drop both.

crates/tycho-integration-test/src/main.rs
```diff
@@ -2030,6 +2035,14 @@ fn is_sampled_block(block_number: u64, interval: u64) -> bool {
     block_number.is_multiple_of(interval)
 }
 
+/// True when a sampled update's target block should be fetched by number (sampled full-block
+/// mode, e.g. Robinhood where the head races past the update). With --partial-blocks, sampling
+/// only gates WHICH updates are tested; block fetching stays on the pending/latest polling path
+/// because a flashblock's pending state cannot be fetched by number after the fact.
+fn fetch_sampled_block_by_number(test_every_n_blocks: u64, partial_blocks: bool) -> bool {
```
louise-poole: Fully agree

---

🤖 It's a predicate returning `bool`, but the name is an imperative verb phrase, so `if fetch_sampled_block_by_number(...)` reads as though the block is fetched on that line. Its one-line sibling directly above is `is_sampled_block`, and the crate's other predicates are `is_oracle_stale_revert` and `reached_max_blocks` — none of them lead with an action verb.

`should_fetch_block_by_number`. Not a hill to die on — I'd just like the two predicates sitting next to each other to both read as predicates.

crates/tycho-integration-test/src/main.rs
```diff
@@ -2030,6 +2035,14 @@ fn is_sampled_block(block_number: u64, interval: u64) -> bool {
     block_number.is_multiple_of(interval)
 }
 
+/// True when a sampled update's target block should be fetched by number (sampled full-block
+/// mode, e.g. Robinhood where the head races past the update). With --partial-blocks, sampling
+/// only gates WHICH updates are tested; block fetching stays on the pending/latest polling path
+/// because a flashblock's pending state cannot be fetched by number after the fact.
```
louise-poole: 🤖 The same two-condition rule now lives in the CLI help (190–195), the call-site comment (1027–1030), this docstring, and the `CLAUDE.md` module bullet (43–46). Next time we tweak it that's four edits and three of them get forgotten. The docstring is also mostly about the call sites — which chain hits this mode, that the head races the update, which path callers take instead — and the Rust conventions ask for what the function does and what the caller can expect.

Split the reasoning instead of repeating it. Keep "the head is expected to be past the update on fast chains, so fetch the target block by number" as a one-line comment at the call site, delete the rest of that comment, and leave the predicate with only the `--partial-blocks` half:
```suggestion
/// True when a sampled update's target block should be fetched by number rather than polled for.
///
/// A flashblock's pending state cannot be fetched by number after the fact, so under
/// `--partial-blocks` sampling only gates which updates are tested.
```

crates/tycho-integration-test/src/main.rs
```diff
@@ -189,13 +189,14 @@ struct Cli {
 
     /// Run the protocol-stream test pipeline only on blocks whose number is a multiple of this
     /// value. 1 tests every block (current behavior). Use a higher value on fast chains (e.g.
-    /// Robinhood) where the harness cannot keep up with the head. State from every block is
-    /// still ingested. Incompatible with --partial-blocks.
+    /// Robinhood) where the harness cannot keep up with the head, or with --partial-blocks to
+    /// cap RPC usage (e.g. Base flashblocks). With --partial-blocks, every partial update of a
+    /// sampled block is tested (bursty by design) and blocks are still fetched via the
+    /// pending/latest polling path. State from every block is ingested either way.
```
louise-poole: 🤖 "blocks are still fetched via the pending/latest polling path" means nothing to whoever runs `--help` — it's an internal detail an operator can't act on. "(bursty by design)" editorialises, and "(current behavior)" is narration left from when 1 was the only behavior.

Keep the half an operator can act on: "Run the protocol-stream test pipeline only on blocks whose number is a multiple of this value. 1 tests every block. Use a higher value on fast chains where the harness cannot keep up with the head, or with --partial-blocks to reduce total RPC volume. Every partial update of a sampled block is tested, so the load is bursty. State from every block is ingested either way."

louise-poole (commented):
The fetch-path split is the right shape and I'm glad this is unblocked. One thing I want to settle before we roll it out on dev: whether block-number sampling actually addresses the limit we tripped, or only the total volume. Rest is naming and docs.

  ---

🤖   **`Timeout` arm skips the per-protocol skip counter** — `crates/tycho-integration-test/src/main.rs:1099–1106`
The `Stale` arm above and both failure arms of the by-number path record `record_protocol_sync_state_skipped` per protocol; `Timeout` records only the aggregate. Pre-existing, but sampled Base runs now route through this arm.

 ->  Could you correct this for us please?

crates/tycho-integration-test/src/main.rs
```diff
@@ -1019,10 +1020,14 @@ async fn process_update(
 
             let poll_interval = Duration::from_millis(cli.rpc_poll_interval_ms);
 
-            let block = if cli.test_every_n_blocks > 1 {
-                // Sampled mode: on fast chains the head is expected to be past the update, so the
-                // target block is fetched by number instead of racing the head. clap rejects
-                // --partial-blocks in this mode, so the block is always full.
+            let by_number =
+                fetch_sampled_block_by_number(cli.test_every_n_blocks, cli.partial_blocks);
```
louise-poole: Agreed - we still need to address the burst rate issue.

---

🤖 Sampling gates on the block number, so a sampled block still runs every one of its partial updates back to back — `parallel_updates × parallel_simulations` traces per flashblock, unchanged. That divides total volume by `test_every_n_blocks` but leaves the instantaneous rate exactly where it was. If the `-32005` we hit was a per-second ceiling rather than the plan quota, the bursts will still trip it.

Which was it? If it was the quota, this is fine as it stands. If it was RPS, gate on the updates instead of the block — test every Nth update, so the rate is capped rather than duty-cycled — and drop "cap RPC usage" from the flag help.

brunoguerios: It was RPS. I followed the suggestion of gating on updates instead of gating on blocks. Should properly solve the issue 👍

tamaralipows (approved):
Thank you!

PR #1404, opened 2026-09-03
feat(router-trades): name executors from a table and settle on one flag

kayibal:
Rewritten. The first version kept out of the substreams package so it could ship without a re-sync; #1405 forces one anyway, so this now removes the duplicates that constraint left behind.

## Two facts were recorded in the wrong place

**An executor's protocol names came from a table compiled into the wasm.** A newly deployed executor needed a package release, a new module hash and a cursor reset before any hop carried its name — and the trades already stored never got it. The address is in the row already, so the name can be resolved when the row is read.

**A trade was settled on `call_success`,** which only says the router call returned normally. A caller quoting the router on chain calls it, reads the return value, then reverts to throw the result away; such a probe carries `call_success` exactly like a fill.

| chain | probe rows | of stored trades | of reported USD volume |
| --- | --- | --- | --- |
| robinhood | 19,904 | 8.9% | **15.6%** ($111.4M → $94.0M) |
| ethereum | 289 | 0.22% | 2.0% |
| base | 6 | — | 0.12% |

Plus 31,556 phantom hops and 39,694 phantom `fees_taken` rows on robinhood. Verified against chain: two rows we stored as identical fills correspond to a single transaction with 11 ERC20 transfers and **no transfer repeated**. Arbitrum's 4,363 same-transaction pairs are all different token pairs, so they are genuine multi-leg trades.

## Executor names

`executors.sql` carries the `executors` table keyed by `(chain, address)`, its rows, and two views. Every container applies it on start.

| View | One row per | Notes |
| --- | --- | --- |
| `trade_hop_protocols` | hop | empty array for an executor no row knows |
| `trade_protocol_systems` | trade | no row when every executor is unknown |

The hop's own chain wins; an address with no row for that chain falls back to the names it carries on chains that do. That fallback is not cosmetic — two executors used on unichain are documented only under base, and without it 1,850 live hops would lose their names.

**The file is kept by hand.** A new executor is a row appended to it, which needs no release. The INSERT ends in `ON CONFLICT DO NOTHING`, so a row added straight to a database survives and re-applying never overwrites one. That has a consequence worth stating: **editing a row in the file does not change the database** — to correct a name, `DELETE` the row and apply again. The integration test asserts exactly that.

`scripts/gen_tables.py` mined the git history of the executor docs and `config/executor_addresses.json` to produce those rows. Its executor half is gone; what remains generates the error table alone, so it is now `gen_error_table.py` (the regenerated `error_table.rs` is byte-identical apart from the header).

**The wasm table and the frozen columns are deleted** — `src/executors_table.rs`, `src/executors.rs`, and the `protocol_systems` columns on `trades` and `trade_hops`. Keeping them beside the views would leave two sources for one fact, and the wasm one goes stale the day an executor is deployed. `Hop.protocol_systems` becomes a reserved field number.

Verified against the live database before deleting them: **329,930 hops resolve identically, 0 lose a name, 213 gain a repaired spelling** (the generator's acronym handling produced `prop_a_m_m_fallback` and `e_r_c4626`). 8,036 hops across 10 addresses stay unnamed — they are in neither the docs nor the config, and each is one row away.

## Settling on one flag

`state_committed` is `state_reverted` on the call, inverted. Firehose sets that flag when the chain threw away a call's state changes because an ancestor reverted, which is what a probe does. The Firehose docs are explicit: *"even if a call has a status of SUCCESS, the chain might have reverted all the state changes it performed."*

```sql
CREATE VIEW trades_settled AS
SELECT t.* FROM trades t
WHERE t.tx_success AND t.call_success AND t.state_committed;
```

**Sum volume over `trades_settled`, not `trades`.** The pricing pass filters on the same flag, so it does no work on a row the view will not carry.

`call_success` stays as it is: it says whether the call returned normally, which is what decides where `amount_out` comes from, and it stays true on a probe.

The first version of this PR also carried a dedup heuristic — "of an otherwise identical swap in one transaction, keep only the last call" — to cover rows indexed before the flag existed. It is gone. It had a real false positive: a route that legitimately hits one pair twice with the same input amount lost a leg. With every chain re-indexed in this release, the flag alone is exact.

`tx_success` and `call_success` look redundant next to the flag, since a reverted call discards its own state and a failed transaction discards all of it. They stay because nothing in the Firehose contract promises a call cannot report `status_reverted` with its state kept. The fixtures set those rows to `state_committed = TRUE` on purpose, so each filter is tested on its own — without that, two mutations survived.

## Tests

Three Postgres integration tests in CI: `make test-executors` (the join, the cross-chain fallback, the unknown executor, the re-apply rule), `make test-views` (`trades_settled` across kept, discarded, reverted, failed and pre-flag rows, plus the identical twins the old heuristic dropped), and the extended `make test-pricing` (a discarded probe stays unpriced).

Mutation tested: four mutations of `trades_settled`, one of the conflict rule, one of the pricing filter — each fails its test.

## Deployment

Every chain is re-indexed in this release, which is what lets the flag be exact and the frozen columns go. A row indexed before the column existed carries NULL and stays out of the view until then; nothing in it says the chain kept the call.

https://claude.ai/code/session_01UDmQJ8ED1jjCXzthvBGPo1


crates/tycho-execution/substreams/pricing/price_trades.sql
```diff
@@ -98,6 +98,9 @@ sided AS (
       AND t.chain = :'chain'
       AND t.tx_success
       AND t.call_success
+      -- A quote probe returns normally but the chain throws its state away. This is the filter
+      -- trades_settled uses, so pricing does no work on a row that view would not carry.
+      AND t.state_committed
```
kayibal: let's price not committed trades as well unless there is a big downside to it.

carloszanella (approved):

crates/tycho-execution/substreams/pricing/price_trades.sql
kayibal: Done — the filter is gone, so a discarded call is priced like any other call that returned.

No real downside. The hazard would be someone summing `volume_usd` over `trades`, but that is already wrong for `amount_in`/`amount_out`; pricing does not create it, and `trades_settled` is the guard. The extra work is small: probes are 8.8% of robinhood's rows, 289 on ethereum, 6 on base.

The upside is real — a quote is market activity, and now readable in USD.

Also dropped `state_committed` from the `trades_unpriced_idx` predicate, otherwise the partial index stops matching the query and the pricing pass loses it.

The fixture flipped from an ineligible row to a positive assertion, so re-adding the filter now fails the test:
```
ERROR:  a discarded call was not priced
```
One place still excludes a discarded call on purpose: the `vault_balances` fee source in #1411. A fee credit on a call the chain threw away never happened on chain, so counting it would make a balance disagree with `balanceOf`.

PR #1405, opened 2026-09-04
feat(router-trades): add a block index and the slippage exemption

kayibal:
Stacked on #1404. Review that one first. The discarded-call flag that used to be part of this PR moved down into #1404, where `trades_settled` needs it; this diff is the block index and the positive slippage exemption.

## 1. Skip a block that touched no router

Every block goes through three modules today. `map_fee_config_events` and `map_trades` each decode the whole block and `store_fee_config` runs on each one, but almost no block holds anything the package keeps.

`index_router_activity` is a `blockIndex` module writing two kinds of key per block: `addr:0x<address>` for every call frame, and `fee_cfg` for a block carrying a log the fee-config decoders recognise. `map_trades` and `map_fee_config_events` filter on those, so the server skips a block with neither — and everything downstream of a filtered module, `store_fee_config` included, only sees blocks that pass.

Measured on 50,000 cold base blocks from `initialBlock`, run from inside the cluster:

| | processed blocks | wall clock |
| --- | --- | --- |
| main | 150,460 | 1,646s |
| this PR | **50,241** | **916s** |
| | **3.0× fewer** | **1.80× faster** |

`db_out` output is byte-identical. Processed blocks are the billed metric. Warm re-runs are 3s either way, so the gain is on cold passes: backfills, a new chain, and any re-index.

**The fee-config filter is the `fee_cfg` key, not an address list.** `map_fee_config_events` decodes a FeeCalculator event from any emitter, and an address filter would have dropped 76 of the 162 stored events — nearly half come from an emitter no manifest configures. The index sets the key with the same two decoders the module uses, so the two cannot drift.

`map_trades` filters on the routers in its params. A filter that is too narrow drops trades and reports nothing, so `scripts/check_manifests.py` compares the two and runs in CI.

The index takes no params and knows no router, so adding a router changes the consumer's filter and leaves the index alone: an index already built stays valid.

## 2. Index the per-client positive slippage exemption

The new FeeCalculator can exempt a client from positive slippage capture. It is contract storage — `mapping(address => bool)` behind `ROUTER_FEE_SETTER_ROLE` — with an event, so it is indexable:

```solidity
event PositiveSlippageExemptionSet(address indexed client, bool exempt);
```

- `abi/FeeCalculatorExemption.json` holds just that event. `FeeCalculator.json` comes from the earlier deployment on Sourcify and every other event is byte-identical in the new contract, so a second file keeps each one true to a real deployment — the same shape `FeeCalculatorV3_0.json` already uses.
- It decodes into the existing `FeeConfigEvent`, which already carries a `client`, so `fee_config_events` needs no new column.
- Stored as `fc:<calculator>:ps_exempt:<client>`, keyed by calculator, so an exemption does not follow a rotation onto a different one.
- `resolve_fee_config` already resolved the client the way `FeeCalculator._resolveClient` does (signed receiver, else `tx.origin`), so the lookup reuses that address unchanged.
- New column `trades.positive_slippage_exempt`, and a schema comment saying who keeps the surplus:

| `positive_slippage_enabled` | `positive_slippage_exempt` | who keeps it |
| --- | --- | --- |
| false | — | the receiver |
| true | false | the router, in `router_fee_amount` |
| true | true | the receiver |

`positive_slippage` is the surplus the swaps *produced*, not what was captured. It had no comment at all, which is how this stays invisible.

**Nothing is wrong in the data today, and the timing is why.** All 251,415 v3_1 trades have capture enabled and **zero** show a surplus with capture off, so `positive_slippage > 0` currently does imply the router took it. The first trade through an exempt client would have broken that silently: flag still true, surplus still computed, receiver keeping it. No traffic has reached the new calculators yet.

Verified `0xdA7da7ACd714c407A37C635a42D5354d0C5B9136` is registered exempt on all seven reachable chains with `isPositiveSlippageExempt` (plasma has no indexer pod to call from).

## Tests

45 unit tests. New: five for the index module (one key per call target, dedup, empty block, a fee-config log setting the key, an unrelated log setting none) and three for the exemption. Both use topic hashes computed with `cast keccak`, because abigen's `TOPIC_ID` is private.

Mutation tested: a global exemption key instead of a per-client one, an inverted flag, and the wrong event name each fail the suite.

## Deployment

New module and a changed wasm, so the module hash changes on every chain and each needs its cursor cleared and its rows deleted. #1404, #1411 and #1412 share that one re-index.

Worth confirming before the rollout: `substreams-sink-sql` v4.13.1 advances cursors correctly over a filtered stream. Its changelog does not mention block filters, so run one chain against a scratch database first.

https://claude.ai/code/session_01UDmQJ8ED1jjCXzthvBGPo1


crates/tycho-execution/substreams/tycho-router-trades/src/modules/index.rs
```diff
@@ -0,0 +1,121 @@
...
+            .receipt
+            .as_ref()
+            .map(|r| r.logs.as_slice())
+            .unwrap_or_default()
+        {
+            if is_fee_config_log(log) {
+                keys.insert(FEE_CONFIG_KEY.to_string());
+                break;
+            }
+        }
```
kayibal, line 47: Do we ever filter by this?  In the manifests I only saw addresses and I hiope those addresses also include the FeeCalculator contracts, in which case this seems superfluous

crates/tycho-execution/substreams/tycho-router-trades/src/modules/index.rs
```diff
@@ -0,0 +1,121 @@
...
+/// Both are supersets of what the consumers select on. `map_trades` matches a call whose target
+/// is a router. `map_fee_config_events` decodes a fee-config log from any emitter, so the key it
+/// filters on comes from those same decoders rather than from a list of addresses; nearly half
+/// of the fee-config events stored so far come from an emitter no manifest configures.
+#[substreams::handlers::map]
+pub fn index_router_activity(block: Block) -> Result<Keys> {
+    Ok(Keys { keys: router_activity_keys(&block) })
+}
+
+fn router_activity_keys(block: &Block) -> Vec<String> {
```
kayibal, line 31: Are you sure the cardinality of this can be this high? I expect many thousand contracts per block.

carloszanella (approved):

crates/tycho-execution/substreams/tycho-router-trades/src/modules/index.rs
kayibal: Yes, and the hoped-for part does not hold — I checked rather than assume.

`map_fee_config_events` filters on this key alone, no addresses:
```yaml
  - name: map_fee_config_events
    blockFilter:
      module: index_router_activity
      query:
        string: "fee_cfg"
```

And across all eight manifests the `addr:` filters name **the 29 routers and none of the 23 FeeCalculators**:
```
routers configured:     29 | in addr: filters: 29
calculators configured: 23 | in addr: filters:  0
```
So without `fee_cfg` those blocks are skipped and every fee-config event is lost.

Adding the calculators to an address filter would not fix it either. `map_fee_config_events` matches a FeeCalculator event **by topic from any emitter**, on purpose: the constructor emits nothing, so a calculator only becomes known through a router rotation event or a manifest parameter — and nearly half the stored events come from an emitter no manifest configures. An address list would silently drop those, which is the failure mode the module was written to avoid.

The key is derived from the same two decoders `map_fee_config_events` uses (`is_fee_config_log`), so the index and the consumer cannot drift.

(I would have re-run the exact emitter count, but the AWS SSO token expired mid-session — the two numbers above are from the manifests, no database needed.)

crates/tycho-execution/substreams/tycho-router-trades/src/modules/index.rs
kayibal: Fair challenge — I had not measured it. I have now, with `debug_traceBlockByNumber` + `callTracer`, counting distinct call targets over the whole trace tree of three blocks per chain:

| chain | txs | distinct call targets |
| --- | --- | --- |
| ethereum | 216–306 | 402 / 537 / **917** |
| base | 268–392 | 337 / 450 / 550 |
| bsc | 51–97 | 254 / 319 / 358 |
| polygon | 73–112 | 148 / 166 / 170 |
| arbitrum | 3–4 | 12 / 12 / 28 |

Hundreds, not thousands. Worst sample is ethereum at 917, and ethereum is the chain where a block index matters least because it has the fewest blocks.

The measurement that already answers the cost question is the benchmark, though: **the 1.80× speedup and 3.0× fewer processed blocks are net of building the index**, because that run started at `initialBlock` with the index module in the graph and had to build it over the same 50,000 base blocks. Whatever the keys cost, the end-to-end number already pays it.

If the cardinality ever does become the problem, the fix is to take the router list as a parameter and emit `addr:` only for a configured address — one or two keys per block instead of hundreds. I deliberately did not: an index that takes no params never has to be rebuilt when a router is added, and a rebuild is the expensive event here. Happy to flip that trade if you would rather bound the index size.

PR #1408, opened 2026-09-04
fix(indexer): instrument commit task body, not its JoinHandle

zizou0x:
## What this changes

Three changes to indexer trace export.

**1. `commit_blocks_task` span wraps the task body** (`protocol_extractor.rs`)

`Instrument` on a `JoinHandle` enters the span only while the handle is polled. The spawned commit task therefore ran with no span context, and every `gateway.advance()` call inside it was exported as its own parentless trace. The span now wraps the task body, so a batch's DB writes nest under one trace.

`BatchCommitHandle` is now a plain `JoinHandle`, which removes the `inner()` hops at the await site and in the test helper. The `parent: None` + `follows_from` link is unchanged — the commit outlives the block that queued it, so it stays a linked root rather than a child.

**2. Constant span attributes are no longer exported** (`ot.rs`)

`code.filepath`, `code.namespace`, `code.lineno`, `thread.id` and `thread.name` are identical on every span of a given name and were re-sent with each one. Busy/idle timings are kept; they vary per span.

**3. `WsActor::handle` span is at debug level** (`ws.rs`)

The actor mailbox separates this handler from the block that produced the message, so the span has no parent and was exported as a single-span trace per message per subscriber. At `RUST_LOG=INFO` it is now filtered out.

## Why

Tempo caps live traces per tenant (default 10,000). The indexers were creating ~1,000 traces/sec and Tempo was discarding roughly 60% of spans, so traces were incomplete regardless of load.

Sampling 40 traces from the cluster showed the split:

| shape | traces | spans each |
|---|---|---|
| `extractor` (per-block, well formed) | 4 | 11 |
| `handle` | 14 | 1 |
| `upsert_block` | 10 | 1 |
| `save_cursor` | 10 | 2 |
| `add_component_balances` | 2 | 1 |

90% of traces were 1–2 span fragments. No child span had a missing parent, so the fragments were real rather than an artifact of the discards. One sampled `upsert_block` fragment reported `busy_ns = 898` — `CachedGateway::upsert_block` is a channel send, and it was producing a standalone trace.

Change 1 folds the `upsert_block` / `save_cursor` / `add_component_balances` fragments into their batch trace with no spans lost. Change 3 removes the `handle` fragments. Together that is most of the trace count.

## What is lost

- The `WsActor::handle` spans, unless `RUST_LOG` is raised to debug.
- Source location and thread name on spans. Where a span name is ambiguous (`upsert_block` exists in `cache.rs`, `direct.rs` and `chain.rs`) the parent chain now disambiguates it. Reverting is one line if the attributes turn out to be wanted.

No change to indexing behaviour.

## Testing

- `cargo test -p tycho-indexer --lib` — 394 pass, 0 fail (the 5 `serial_db` tests run against a local Postgres).
- `cargo +nightly-2026-06-28 fmt --all --check` clean.
- `cargo +nightly-2026-06-28 clippy -p tycho-indexer --all-targets --all-features -- -D warnings` clean.

Span parenting itself is not covered by a new test. Asserting it needs a subscriber that observes a spawned task, which means a global default subscriber and would be unreliable in a parallel test run. `test_handle_tick_scoped_respects_batch_async_commit` covers the commit-task lifecycle and passes.

### DIFF
--- a/crates/tycho-indexer/src/extractor/protocol_extractor.rs
+++ b/crates/tycho-indexer/src/extractor/protocol_extractor.rs
@@ -60,7 +60,7 @@ pub struct Inner {
     first_message_processed: bool,
 }
 
-type BatchCommitHandle = tracing::instrument::Instrumented<JoinHandle<Result<(), ExtractionError>>>;
+type BatchCommitHandle = JoinHandle<Result<(), ExtractionError>>;
 
 #[derive(Default)]
 struct GatewayInner<G> {
@@ -774,9 +774,7 @@ where
             let (extractor_name, chain) = (self.name.clone(), self.chain);
 
             if let Some(db_commit_handle_to_join) = commit_handle_guard.take() {
-                let needs_await = !db_commit_handle_to_join
-                    .inner()
-                    .is_finished();
+                let needs_await = !db_commit_handle_to_join.is_finished();
                 let now = chrono::Utc::now().naive_utc();
 
                 let result = db_commit_handle_to_join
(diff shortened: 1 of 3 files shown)


brunoguerios (approved):
LGTM ✅

PR #1420, opened 2026-09-07
fix: stop Hashflow quotes from invalidating each other

Troshchk:
Problem

Hashflow rejects a quote executed after a later-issued one from the same (pool, effectiveTrader) — nonces are maker timestamps and must strictly increase per slot. We never set an effectiveTrader, so every quote we requested was attributed to the router: all our trades through one maker pool shared a single nonce sequence. Quotes fetched in parallel (hops of one route, or concurrent solutions) could then execute out of nonce order, and the later-executing swap reverted on-chain with RFQT: nonce used.

Fix

Send a fresh random effectiveTrader with every quote request. Each quote gets its own nonce sequence, so quotes can never invalidate each other — no fetch ordering needed, all quotes request in parallel.

Gas

The nonce slot is keyed by (pool, effectiveTrader), so a fresh address pays a cold first write every swap. A shared per solution address costs the same — each hop hits a different pool, so its slots are all first writes too. Only reusing one stable address across many trades gets warm slots, and that reintroduces shared nonce sequences and sequential fetching.


crates/tycho-execution/src/encoding/evm/swap_encoder/hashflow.rs
```diff
@@ -51,12 +51,16 @@ impl SwapEncoder for HashflowSwapEncoder {
...
-        let sender = encoding_context
+        let router_address = encoding_context
             .router_address
             .clone()
             .ok_or(EncodingError::FatalError(
                 "The router address is needed to perform a Hashflow swap".to_string(),
             ))?;
+        // A fresh random address becomes the quote's effective trader, so every quote has its
+        // own nonce sequence: quotes never invalidate each other and encode in any order. The
+        // cost is a cold nonce storage slot on Hashflow's router (~15k extra gas per swap).
```
tamaralipows: So the effective trader is essentially just a random number and was meaningless on the hashflow contract side? This is not used for anything at all?

crates/tycho-execution/contracts/test/protocols/Hashflow.t.sol
```diff
@@ -53,7 +54,29 @@ contract HashflowExecutorECR20Test is Constants, TestUtils, HashflowUtils {
     }
 
     function testDecodeParams() public view {
-        IHashflowRouter.RFQTQuote memory expected_quote = rfqtQuote();
+        // Every field is distinct so a decoding offset mixup — like reading the
+        // trader slice as the effectiveTrader — fails the assertions. Decoding
+        // needs no valid signature.
```
tamaralipows: I don't think this comment is saying much? Fields are distinct so mismatched fields will fail assertion?

crates/tycho-execution/contracts/test/protocols/Hashflow.t.sol
tamaralipows: I don't remember what's the point of this passthrough method? I know it was here before... can we just remove it and put a comment in the original one that it defaults to empty fields

tamaralipows, line 153: It would be nice if we had a non-integration test confirming that setting the effective trader and sending from a random address that is not the effective trader, does nothing

crates/tycho-execution/tests/protocol_integration_tests.rs
```diff
@@ -1249,6 +1249,12 @@ fn test_single_encoding_strategy_hashflow() {
                 "trader".to_string(),
                 Bytes::from_str("0x6bc529dc7b81a031828ddce2bc419d01ff268c66").unwrap(),
             ),
+            (
+                // The market maker signed this quote with the effective trader equal to the
+                // trader.
```
tamaralipows: Would like a test where the effective trader is just a random address like we are doing in the encoding

crates/tycho-simulation/src/rfq/protocols/hashflow/models.rs
```diff
@@ -416,6 +427,27 @@ mod tests {
...
         }
 
+        #[test]
+        fn test_validate_effective_trader_mismatch() {
+            let mut quote = quote();
+            quote.quote_data.effective_trader =
+                Some(hex_to_bytes("0xabcdefabcdefabcdefabcdefabcdefabcdefabcd"));
+            let params = params();
+            let err = quote.validate(&params).unwrap_err();
+            assert!(format!("{err:?}").contains("Effective trader mismatch"));
```
tamaralipows, line 454: What exactly is this testing?

crates/tycho-simulation/src/rfq/protocols/hashflow/models.rs
```diff
@@ -416,6 +427,27 @@ mod tests {
...
+        }
+
+        #[test]
+        fn test_validate_effective_trader_absent() {
+            // Without an effective trader the quote defaults to the trader's scope, which only
+            // passes when that is what was requested.
```
tamaralipows: What does it mean by "only what that is what was requested"?

tamaralipows (commented):
Thanks Ksenia. Please remember to put a concise description of the original problem in a `fix` PR, else we won't know clearly without going to the commits what is being fixed.

Here are some questions/requests

crates/tycho-execution/src/encoding/evm/swap_encoder/hashflow.rs
```diff
@@ -65,8 +69,8 @@ impl SwapEncoder for HashflowSwapEncoder {
                         amount_in,
                         token_in: swap.token_in().address.clone(),
                         token_out: swap.token_out().address.clone(),
-                        sender: sender.clone(),
-                        receiver: sender,
+                        sender: effective_trader,
+                        receiver: router_address,
```
louise-poole: Agreed - we've sometimes toyed with an idea to use the `sender` field for various checks etc. Though there are no important ones existing just yet, I think we should try to keep `sender` true to what it should represent and rather generate the random address internally or pass it another way.

---

🤖 The encoder mints a random address and passes it as `GetAmountOutParams::sender`; the Hashflow client then maps `sender` onto `effectiveTrader`. Bebop, Liquorice and Metric all pass the router as `sender`, so the field means one thing for three protocols and another for Hashflow. Nothing in the shared struct says so, and the names now invert: the random address never sends anything, while `receiver` holds the address that does. Anyone who later puts a real payer back into `sender` reinstates the shared-nonce collision silently.

Generate the `effectiveTrader` inside `HashflowClient::request_binding_quote` and emit it as a quote attribute, leaving the encoder to pass the router as both `sender` and `receiver` like every other RFQ encoder. Pros: one module owns the nonce rule, encoders stay uniform, tycho-execution stops needing alloy's `rand` feature. Cons: a caller can no longer pin the effectiveTrader, and `validate` then checks a value the client itself produced. If you keep the split, document both fields on `GetAmountOutParams` in `tycho-common` and name the Hashflow mapping there.

crates/tycho-execution/contracts/src/executors/HashflowExecutor.sol
```diff
@@ -84,25 +84,24 @@ contract HashflowExecutor is IExecutor {
...
-        quote.quoteExpiry = uint256(bytes32(data[164:196]));
-        quote.nonce = uint256(bytes32(data[196:228]));
-        quote.txid = bytes32(data[228:260]);
-        quote.signature = data[260:325];
+        quote.baseTokenAmount = uint256(bytes32(data[120:152]));
+        quote.quoteTokenAmount = uint256(bytes32(data[152:184]));
+        quote.quoteExpiry = uint256(bytes32(data[184:216]));
+        quote.nonce = uint256(bytes32(data[216:248]));
+        quote.txid = bytes32(data[248:280]);
+        quote.signature = data[280:345];
```
louise-poole, line 104: Good suggestion - up to you! Does make it feel cleaner

---

🤖 Adding one 20-byte field took ten offset edits in the Solidity decoder, one new entry in the Rust field list, and a matching line in the Solidity test encoder. Nothing links them. Reorder a field on one side only and the payload is still 345 bytes, so the length check passes and the executor hands Hashflow a garbage quote — caught only by the mainnet-fork integration test, the slowest guard in the repo.

Replace the ten literal slice bounds in `_decodeData` with named cumulative offset constants, so inserting a field is one edit rather than ten. Note on `hashflow_fields` that its order and widths define that layout, and note the reverse on `_decodeData`. Compile-checked agreement instead would mean `abi.encode`/`abi.decode` of `RFQTQuote` on both sides, which costs roughly 230 extra calldata bytes per swap.

crates/tycho-simulation/src/rfq/protocols/hashflow/client.rs
```diff
@@ -524,7 +527,13 @@ impl RFQClient for HashflowClient {
                                 })?,
                             );
                         }
+                        let effective_trader = quote
+                            .quote_data
+                            .effective_trader
+                            .clone()
+                            .unwrap_or_else(|| quote.quote_data.trader.clone());
                         quote_attributes.insert("trader".to_string(), quote.quote_data.trader);
+                        quote_attributes.insert("effective_trader".to_string(), effective_trader);
```
louise-poole: Small suggestion

---

🤖 `validate` runs first and rejects any quote whose effective trader — or trader, when the API omits it — differs from `params.sender`, so the `unwrap_or_else` below always yields `params.sender`. The same default now exists in two files and can drift apart.

Insert `params.sender.clone()` and drop the local fallback, leaving the one in `HashflowQuote::validate`. If you would rather keep both call sites, put the fallback on `HashflowQuoteData` as an accessor so the rule lives in one place.

crates/tycho-execution/src/encoding/evm/swap_encoder/hashflow.rs
```diff
@@ -78,6 +82,7 @@ impl SwapEncoder for HashflowSwapEncoder {
             "pool",
             "external_account",
             "trader",
+            "effective_trader",
             "base_token",
             "quote_token",
             "base_token_amount",
```
louise-poole, line 84: True - in this case why did we have 'trader' there in the first place I wonder? Up to you!

---

🤖 The encoder sends `receiver: router_address` and `validate` forces the quote's trader to equal it, so those 20 bytes always carry an address `_decodeData` already has as `address(this)` under delegatecall. The diff grew the payload by 20 bytes where it could have stayed flat.

Set `quote.trader = address(this)` in `_decodeData` and drop the field from the Rust list. This changes the encoded format and breaks the standalone executor tests that trade as ALICE, so it needs a redeploy and is your call, not mine.

crates/tycho-execution/CLAUDE.md
```diff
@@ -237,7 +237,11 @@ group is PLE-encoded (`[len: u16][data]...`); Ekubo uses concatenation instead (
 **Parallel encoding**: `encode_swap_groups` (`strategy_encoders.rs`) and `TychoRouterEncoder::encode_solutions`
 spawn threads via `map_on_threads` (`evm/utils.rs`) **only when an encoder in the batch blocks on a quote** —
 `SwapEncoder::blocks_on_quote()` defaults to `false` and is `true` only for the RFQ encoders (Bebop, Hashflow,
-Liquorice, Metric). Otherwise encoding runs serially on the calling thread. Input order is preserved either way.
+Liquorice, Metric). Otherwise encoding runs serially on the calling thread. Input order is preserved
+either way. Hashflow quotes stay independent under this parallelism because the encoder sends a fresh random
+effectiveTrader with every quote request: Hashflow scopes its strictly increasing quote nonces per effective
+trader, so quotes with unique addresses never invalidate each other. The cost is a cold nonce storage slot on
+Hashflow's router (~15k extra gas per swap).
```
louise-poole: Important gas adjustment

---

🤖 `get_amount_out` (`hashflow/state.rs:127`) returns a flat `134_000` gas for every Hashflow swap, and this PR adds a cold nonce SSTORE that the estimate doesn't include. The solver picks routes on that number, so Hashflow now looks ~12% cheaper than it is and wins hops it shouldn't.

Bump the constant by the cold-slot cost and say in the comment that the fresh effectiveTrader is where it comes from, so the next person doesn't re-round it back down.

crates/tycho-execution/contracts/test/protocols/Hashflow.t.sol
```diff
@@ -17,6 +17,7 @@ contract HashflowUtils is Test {
             quote.pool, // pool (20 bytes)
             quote.externalAccount, // externalAccount (20 bytes)
             quote.trader, // trader (20 bytes)
+            quote.effectiveTrader, // effectiveTrader (20 bytes)
```
louise-poole, line 20: Did you manage to test this with a real Hashflow quote?

---

🤖 `rfqtQuote` still sets `effectiveTrader` and `trader` to ALICE, and `protocol_integration_tests.rs:1252–1257` keeps them equal too. The whole point of the change is that they now differ in production, and no fork test or calldata fixture covers that shape, so a swap against the real router is only checked in the old configuration.

Get one market-maker-signed quote with a distinct effectiveTrader and use it for the fork test, or say in the PR whether you ran the new shape against a real Hashflow quote.

crates/tycho-execution/src/encoding/evm/swap_encoder/hashflow.rs
louise-poole: Docs consistency improvement

---

🤖 Both places say ~15k extra gas per swap while the PR body says ~17k. Writing a fresh nonce slot is a zero-to-nonzero cold SSTORE at 22 100 gas against 5 000 to bump an existing one, so the delta is ~17.1k. That number is the entire justification for minting an address per hop, and a low figure makes the trade read as cheaper than it is.

Put ~17k in both places. Dropping the figure and saying "one cold nonce storage slot" also works, and won't rot when Hashflow's router or the EVM's cold-access pricing changes.

crates/tycho-execution/src/encoding/evm/testing_utils.rs
```diff
@@ -28,6 +33,10 @@ pub struct MockRFQState {
     /// How long `request_signed_quote` waits before it answers, like a network round trip.
     #[serde(default)]
     pub delay: Duration,
+    /// The `(token_in, sender)` of every quote request, in arrival order. Share one log across
+    /// states to observe the request order and quote senders of a route.
```
louise-poole: Small doc improvement.

---

🤖 Cut the second sentence, which advertises a cross-state sharing use no test has. The recorded address is the effective trader, not a sender — name it that.

louise-poole (commented):
Approach makes sense - a fresh effectiveTrader per quote is the right way to keep the nonce sequences apart. Two things I'd like to see: `sender` kept true to what it represents rather than carrying the random address, and the gas estimate bumped now that every swap writes a cold nonce slot. The rest is up to you.

crates/tycho-execution/src/encoding/evm/swap_encoder/hashflow.rs
Troshchk: It is used only for the nonce check on the hashflow side

crates/tycho-execution/contracts/test/protocols/Hashflow.t.sol
Troshchk: Yes, added this after your review: testSwapEffectiveTraderNotCaller (1cb50b113) and testHashflowIntegration (804e29fe1)

crates/tycho-execution/contracts/src/executors/HashflowExecutor.sol
Troshchk: We do not expect to change it often, so I would keep it as is. Also, adding more consts hurts readability in my opinion

crates/tycho-execution/src/encoding/evm/swap_encoder/hashflow.rs
Troshchk: I decided to not remove it, it has influence on a test mainly and we do not gain anything from this change

crates/tycho-simulation/src/rfq/protocols/hashflow/models.rs
Troshchk: It tests that a quote echoing a different effectiveTrader than requested is rejected, there are same tests for all the other fields too

crates/tycho-simulation/src/rfq/protocols/hashflow/models.rs
Troshchk: I have rephrased it

crates/tycho-simulation/src/rfq/protocols/hashflow/client.rs
```diff
@@ -949,9 +1013,73 @@ mod tests {
...
+            .request_binding_quote(&params)
+            .await
+            .unwrap_err();
+
+        assert!(format!("{err:?}").contains("Effective trader mismatch"));
+    }
+
+    #[tokio::test]
+    async fn test_request_binding_quote_field_mapping() {
+        // The wire request carries the receiver as Hashflow's trader and a fresh random
```
tamaralipows, line 1038: Is it just me?  I am starting to hate the word "wire". I guess another Anthropic watermark. Nothing to do here lol just venting.

tamaralipows (approved):
Thank you! Looks good from my side.

crates/tycho-simulation/src/rfq/protocols/hashflow/client.rs
Troshchk: Same same 😅

louise-poole (approved):

PR #1423, opened 2026-09-08
fix(indexer): base revert attr miss label on component existence

brunoguerios:
## Problem

`[PROD-BASE] Revert Attribute Miss` fired 4 times between 2026-09-04 and 2026-09-08 with
`component_state_found=false`, always `uniswap_v3`, on both Base indexers in the same minute:

```
Attributes with no prior state in buffer or DB during revert; reverting them as deletions
components=[("0xd3845774…", 2)]
```

No component was lost. All 4 were noise.

## Root cause

The label came from `get_protocol_states`, which returns state rows of flushed blocks only. A pool
created 97 blocks earlier still has its creation in the reorg buffer and no DB rows. When the
Uniswap fee controller set `protocol_fees/token0|1` for the first time and Substreams reverted that
block as flashblock partials, the attribute miss was real but the label called the component absent.

## Fix

For the residual miss ids the extractor now checks component existence: a `protocol_components`
entry anywhere in reorg buffer history, then a `protocol_component` row in the DB. The label is
renamed to `component_known`; `false` now means no creation exists anywhere. Revert behaviour is
unchanged.

The summary warning now lists the misses per label: `attr_misses=[(id, n)]` for known
components and `component_misses=[(id, n)]` for unknown ones. The `components=` and `total=`
fields are gone, so the grep line above no longer matches after release.

## Notes for reviewers

Both lookups run only on the rare miss path. The protocol cache is not used for the check: it
keeps creations that a revert undid, so a second revert would label a reverted ghost as known.
The `component_state_found` series stops increasing on release, so every current alert rule
goes quiet until the rule rename lands.

## Follow-ups

helm-configuration: switch all 19 `Revert Attribute Miss` rules to `component_known`: 10 under
`helmwave/prod/values/tycho/`, 2 under `helmwave/prod/values/tycho-fynd/`, 7 under
`helmwave/dev/values/tycho/`. The tycho-fynd and dev files also embed `component_state_found`
in the alert `description` annotation.

### DIFF
--- a/crates/tycho-indexer/src/extractor/protocol_extractor.rs
+++ b/crates/tycho-indexer/src/extractor/protocol_extractor.rs
@@ -231,6 +231,37 @@ where
         Ok(())
     }
 
+    /// Returns the ids among `ids` with no protocol component entry in the reorg buffer history
+    /// and no component row in the DB.
+    async fn missing_components(
+        &self,
+        reorg_buffer: &ReorgBuffer<BlockUpdateWithCursor<BlockChanges>>,
+        ids: Vec<&ComponentId>,
+    ) -> Result<HashSet<ComponentId>, ExtractionError> {
+        let (_, absent_from_buffer) = reorg_buffer.lookup_components(&ids);
+        if absent_from_buffer.is_empty() {
+            return Ok(HashSet::new());
+        }
+        let in_db: HashSet<ComponentId> = self
+            .gateway
+            .inner
+            .get_protocol_components(
+                &absent_from_buffer
+                    .iter()
+                    .map(String::as_str)
+                    .collect::<Vec<&str>>(),
+            )
+            .await
+            .map_err(ExtractionError::Storage)?
+            .into_iter()
+            .map(|component| component.id)
+            .collect();
+        Ok(absent_from_buffer
+            .into_iter()
+            .filter(|id| !in_db.contains(id))
+            .collect())
+    }
+
     async fn is_first_message(&self) -> bool {
         !self
             .inner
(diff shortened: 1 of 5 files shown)


crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -231,6 +231,38 @@ where
...
 
+    /// Returns the ids among `ids` with no protocol component entry in the reorg buffer history
+    /// and no component row in the DB.
+    ///
+    /// The protocol cache is not consulted: it keeps creations that a revert undid, so it would
+    /// label a reverted ghost as known.
+    async fn find_unknown_components(
+        &self,
+        reorg_buffer: &ReorgBuffer<BlockUpdateWithCursor<BlockChanges>>,
+        ids: &[&ComponentId],
```
zizou0x: From the call site, I think it would be better to make this `IntoIterator`, it would avoid having to collect into vec and then `iter()` inside `missing_components()` (we should change it there too)

crates/tycho-indexer/src/extractor/protocol_extractor.rs
```diff
@@ -231,6 +231,38 @@ where
         Ok(())
     }
 
+    /// Returns the ids among `ids` with no protocol component entry in the reorg buffer history
```
zizou0x: By "component entry" you mean a component creation?

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: Claude suggested to go with a `HashSet` as it was already being used by `missing_components`. 
Let me know if you'd rather keep it as `IntoIterator` instead.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: No it's indeed an entry here. It checks if the component id exists either in the buffer or the DB.

crates/tycho-indexer/src/extractor/protocol_extractor.rs
zizou0x: I'd make find_unknown_components take IntoIterator and missing_components take Hashset, I think it makes this easier to use and still efficient

crates/tycho-indexer/src/extractor/protocol_extractor.rs
zizou0x: What is a protocol component entry in the reorg buffer? Isn't it equivalent to a component creation?

zizou0x (approved):

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: Yes, you're right 👍 
`protocol_components` entry within `reorg_buffer` is indeed equivalent to a component creation as other updates are registered in other attributes (e.g. updates to `state_updates`).
I'll rewrite to make it more clear

crates/tycho-indexer/src/extractor/protocol_extractor.rs
brunoguerios: `find_unknown_components` now expects an `IntoIterator` and `missing_components` now expects a `HashSet` ✅

PR #1424, opened 2026-09-08
fix(storage): retry write batches on serialization failure

brunoguerios:
## Problem

`bsc-tycho-indexer` (0.376.0) exited with code 1 at 02:16:36 UTC on 2026-09-04. The API was down
3 min 18 s, the auth proxy served ~950 HTTP 502s, and `bsc-fynd` and `bsc-fynd-spot` restarted. The
`uniswap_v4` write path had been stalled since 02:12:54 while the stream looked healthy:

    ERROR db_write: DBTransactionFailed error=PostgresError(Unexpected("could not serialize
    access due to concurrent update")) block_range=[119836433, 119836463] extractor_id="uniswap_v4"

## Root cause

The write executor runs each block batch at REPEATABLE READ and re-inserts every token of every
touched component with `INSERT ... ON CONFLICT DO NOTHING`. When another transaction updates one of
those token rows and commits after the batch snapshot, Postgres raises SQLSTATE 40001 on that
insert. The nightly `analyze-tokens` job is that writer: it updated ~93k BSC token rows in 94
commits during the window, as it does every night. The executor's retry loop matched only
`deadlock detected`, so the 40001 propagated. The error surfaced one commit batch later (512
blocks), and 0.376.0 has no extractor supervisor, so the process exited.

## Fix

The executor retries a batch when the Postgres message inside `StorageError::Unexpected` is a
transaction conflict: `deadlock detected` (40P01) or `could not serialize access …`, the prefix
Postgres uses for every serialization failure (40001). The check is one helper in
`tycho-storage`; no public type changes. Retries stay at three attempts with the existing 1 s and
2 s delays. The re-run gets a fresh snapshot and the token insert becomes a no-op.

A serial DB test holds the token row lock on a dedicated connection, forces a real 40001, and
checks that the block id advanced twice, so it fails when the batch does not run twice. A second
test allows one attempt and checks that the conflict reaches the caller.

## Notes for reviewers

The isolation level is unchanged. Classification is message-based on purpose: a new
`StorageError` variant adds a case to a public enum of a published crate. Postgres messages are
English unless `lc_messages` is changed, which the deadlock check already relied on.
Serialization failures wait through the deadlock backoff; an immediate retry would work but is not
worth a second predicate for a few conflicts per night.

## Follow-ups

- Stop re-inserting known tokens every batch (only the unknown set needs the insert).
- `analyze-tokens`: skip the write for tokens whose fields did not change. Its own write path
  (`update_tokens`) does not retry on conflicts.

### DIFF
--- a/crates/tycho-storage/src/postgres/mod.rs
+++ b/crates/tycho-storage/src/postgres/mod.rs
@@ -339,8 +339,22 @@ impl<T, O> WithOrdinal<T, O> {
 #[derive(Debug)]
 struct PostgresError(StorageError);
 
+/// Returns the Postgres message if the error aborted the transaction because it conflicted with a
+/// concurrent one (SQLSTATE class 40): a serialization failure or a deadlock.
+///
+/// Diesel maps 40001 to `SerializationFailure` but 40P01 to `Unknown`, hence the message check.
+fn transaction_conflict(err: &diesel::result::Error) -> Option<String> {
+    let diesel::result::Error::DatabaseError(kind, info) = err else { return None };
+    let is_conflict = matches!(kind, diesel::result::DatabaseErrorKind::SerializationFailure) ||
+        info.message() == "deadlock detected";
+    is_conflict.then(|| info.message().to_owned())
+}
+
 impl From<diesel::result::Error> for PostgresError {
     fn from(value: diesel::result::Error) -> Self {
+        if let Some(reason) = transaction_conflict(&value) {
+            return PostgresError(StorageError::TransactionConflict(reason));
+        }
         PostgresError(StorageError::Unexpected(format!("DieselError: {value}")))
     }
 }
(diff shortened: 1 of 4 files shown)


crates/tycho-storage/src/postgres/mod.rs
```diff
@@ -357,6 +357,14 @@ impl From<StorageError> for PostgresError {
...
 }
 
+/// True if `message` is the Postgres error of a transaction that aborted because it conflicted
+/// with a concurrent one: a serialization failure (SQLSTATE 40001) or a deadlock (40P01). Every
+/// 40001 message starts with "could not serialize access"; the check is a `contains` because one
+/// conversion path prefixes the message with `DieselError: `.
+fn is_transaction_conflict(message: &str) -> bool {
+    message.contains("deadlock detected") || message.contains("could not serialize access")
+}
+
```
brunoguerios, line 367: Wanted to highlight this check here. The original implementation added an error to `StorageError` called `TransactionConflict`, which I believe would be the ideal solution. Problem is that it is a breaking change as that error is external and exhaustive. This was the solution found to avoid the external breaking change. Let me know if I should revert back to the original fix instead.

zizou0x (approved):
LGTM

PR #1427, opened 2026-09-09
feat: add etherfi substreams and fix its simulation

zach030:
Completes the EtherFi integration: the `ethereum-etherfi` Substreams package, and a rewrite of the
`EtherfiState` simulation against the contracts live since the escrow migration. Migrated from
propeller-heads/tycho-protocol-sdk#346; `EtherfiExecutor` and `EtherfiSwapEncoder` are already
merged.

## The contracts had moved

LiquidityPool, eETH and EtherFiRedemptionManager were upgraded together at block 25533308, before
this package's `start_block` 25940000, and the implementations the earlier code cited were three
and four upgrades behind. `ethAmountLockedForWithdrawal` is gone and its slot is a storage gap, so
redemption liquidity is `totalValueInLp`. `eETH.mintShares` and `burnShares` now draw on two
buckets of a separate `EtherFiRateLimiter`, which bound deposits and redemptions.

## Indexing

Two components: `eETH` (`0x35fA…8ac2`, deposit and redeem) and `weETH` (`0xCd5f…b7ee`, wrap and
unwrap). Every tracked slot names its contract and position, and each value packed into a word is
one attribute; every slot is checked against the contract's own getter at `start_block`. Balances
are derived from the tracked slots on every block that moves one, so rebases and donations are
counted; they are the protocol's own accounting, hence `skip_balance_check`.

The contracts predate any block worth indexing from, so the manifest carries the slot values at
`start_block` in `params`, and `scripts/compute_initial_state.sh <block>` regenerates them.

The manifest also records the implementation behind each of the four EIP-1967 proxies. The package
watches their implementation slot and pauses both components with `PausingReason::Substreams` on
the transaction that leaves any of them on another implementation, judged by the address the last
write in that transaction leaves there.

## Simulation

`EtherfiState` is a `Venue` enum: each component quotes exactly its own two directions and any
other pair is an error. Redemption follows `_calcRedemption`, deposits are bounded by the mint
bucket and by the `uint128` `totalValueInLp` is stored in, and `apply_block` tracks the execution
block's timestamp so the buckets refill between the protocol's own writes. Every reported limit
quotes at its own size and returns exactly its own output.

The package reports the eETH component's ETH side as `0x00…00` (`Chain::native_token`), not the
router's `0xEeee…EEeE` sentinel, which the indexer cannot resolve as a token.

## Testing

| suite | result |
|---|---|
| `cargo test -p tycho-simulation --lib etherfi` | 42/42 |
| `cargo test -p ethereum-etherfi` | 38/38 |
| `forge test --match-path test/protocols/Etherfi.t.sol` | 9/9, mainnet fork at 25940000 |
| `protocol-testing range --package ethereum-etherfi` | 1/1 over 25940000→25940200 |

Every formula is checked wei-exact against the verified implementations, and every tracked slot
against its getter at `start_block` and at head. Seven mutations that used to leave the suite green
now fail it, among them rounding `sharesForWithdrawalAmount` down, dropping the burn bucket from
the redemption limit, and routing every balance slot to one store key.

`test-protocols` does not currently run on any PR: the workflow triggers on `pull_request_target`,
where `github.sha` is the base branch tip, so its change detection always finds nothing.

## Known gaps

**Execution is skipped in the harness.** The executor's runtime bytecode needs code at four
constructor arguments the pinned fork does not have. #1455 reworks how those fixtures are produced
and names the stub-planting this branch previously carried as one of the two changes it supersedes,
so EtherFi joins the fixtures once that lands. Until then the harness covers indexing and quoting,
and the Foundry tests cover execution against the live contracts.

**`estimate_gas_usage` prices etherfi legs with neither the output transfer nor the approval.**
Its tables are keyed on the protocol name, but what they encode is per leg: three of EtherFi's four
directions send their output to the router, and the two that debit eETH from it pay a
`forceApprove`, so a plain entry in either table would be wrong for the redemption. The estimate is
informational and never reaches the calldata, so this is left for the rework that reads the
classification from `getTransferData`, where each executor already states it per direction.


crates/tycho-simulation/src/evm/protocol/etherfi/state.rs
```diff
@@ -121,7 +124,15 @@ impl EtherfiState {
             return Ok(U256::ZERO);
         }
         let numerator = amount * self.total_shares;
-        Ok(numerator + total_pooled_ether - U256::ONE / total_pooled_ether)
+        // Ceiling division, matching LiquidityPool.sharesForWithdrawalAmount on chain.
+        Ok((numerator + total_pooled_ether - U256::ONE) / total_pooled_ether)
```
die-herdplatte: There is https://docs.rs/ruint/latest/ruint/struct.Uint.html#method.div_ceil

protocols/substreams/ethereum-etherfi/src/modules.rs
```diff
@@ -0,0 +1,476 @@
...
+            .entry(tx.index)
+            .or_insert_with(|| TransactionChangesBuilder::new(tx));
+
+        let filtered_storage_changes: Vec<StorageChange> = change
+            .storage_changes
+            .iter()
+            .filter(|storage_change| {
+                storage_change.address == LIQUIDITY_POOL_ADDRESS.to_vec() ||
+                    storage_change.address == EETH_ADDRESS.to_vec() ||
+                    storage_change.address == REDEMPTION_MANAGER_ADDRESS.to_vec()
```
die-herdplatte: <sub>🤖 Claude Code ──────────────────────────</sub>

`get_changed_attributes` matches a storage change **by slot key only** — it never looks at `change.address` (storage.rs, `if change.key == storage_location.slot`). This filter hands it changes from three different contracts at once, so any of them writing one of the six tracked slot numbers is decoded as the EtherFi attribute of that number.

That is not hypothetical for slot `0xca` (`totalShares`): the LiquidityPool uses it too, and today it holds an address.

```
$ cast storage 0x308861A430be4cce5502d0A12724771Fc6DaF216 0xca
0x0000000000000000000000008b71140ad2e5d1e7018d2a7f8a288bd3cd38916f
```

So the first admin transaction that rewrites that LiquidityPool address slot emits `totalShares = 794362…` (~7.9e47) for the eETH component, and — via the same bug at line 393, where `WEETH_POOL_TRACKED_SLOTS` also contains `0xca` and the filter also contains the LiquidityPool — for the weETH component as well. Every `shares_for_amount` / `amount_for_share` afterwards is off by ~24 orders of magnitude until the next real `eETH.totalShares` write repairs it, and the same applies in reverse for `0xcf`/`0xdc` on eETH and the RedemptionManager.

Every other package that uses this helper filters to exactly one address first (see `ethereum-rocketpool/src/modules/2_map_protocol_changes.rs:153`, `.filter(|call| call.address == ROCKET_VAULT_ADDRESS)`). Pairing each tracked slot with its contract — or calling `get_changed_attributes` once per address with only that address's slots — closes it.

<sub>────────────────────────────────────────────────────</sub>

protocols/substreams/ethereum-etherfi/src/modules.rs
```diff
@@ -0,0 +1,476 @@
...
+                                .as_bytes()
+                                .to_vec() &&
+                            components_store
+                                .get_last(format!("0x{}", hex::encode(EETH_ADDRESS)))
+                                .is_some()
+                        {
+                            builder.add_entity_change(&EntityChanges {
+                                component_id: format!("0x{}", hex::encode(EETH_ADDRESS)),
+                                attributes: vec![Attribute {
+                                    name: "liquidityPoolNativeBalance".into(),
```
die-herdplatte: <sub>🤖 Claude Code ──────────────────────────</sub>

This update is emitted as `liquidityPoolNativeBalance`, which is the name the snapshot decoder reads (`crates/tycho-simulation/src/evm/protocol/etherfi/decoder.rs:62`) — but `EtherfiState::delta_transition` looks the *delta* up under a different key:

```rust
// crates/tycho-simulation/src/evm/protocol/etherfi/state.rs:451
.get("liquidity_pool_native_balance")
```

Nothing in the tree emits that snake_case name, so the lookup never hits. `liquidity_pool_native_balance` is therefore frozen at whatever the initial snapshot carried, for the whole lifetime of the component — every subsequent block's real balance is discarded.

Consequences, all on the eETH component:

- `get_limits(eETH, ETH)` and `get_amount_out(eETH, ETH)` size redemption against a stale liquid balance. As the pool drains, quotes are offered that revert on chain; as it refills past the floor, the direction keeps reporting a zero limit and stays skipped.
- `liquid_eth_amount = liquidity_pool_native_balance - eth_amount_locked_for_withdrawl` (state.rs:269) is an alloy `U256` subtraction, which **panics** on underflow. `ethAmountLockedForWithdrawl` *is* updated by `delta_transition`, so once the frozen balance falls below a growing locked amount, the panic is reachable from an ordinary `get_amount_out` call rather than a `SimulationError`.

The fix belongs on the simulation side — rename the key in `delta_transition` to `liquidityPoolNativeBalance` so it matches both this module and the decoder.

<sub>────────────────────────────────────────────────────</sub>

protocols/substreams/ethereum-etherfi/src/modules.rs
```diff
@@ -0,0 +1,476 @@
...
+            ord: 0,
+            tx: Some(tx.into()),
+            token: ETH_ADDRESS.to_vec(),
+            delta: initial_state
+                .liquidity_pool_native_balance()?
+                .to_signed_bytes_be(),
+            component_id: EETH_ADDRESS
+                .to_hex()
+                .as_bytes()
+                .to_vec(),
```
die-herdplatte: <sub>🤖 Claude Code ──────────────────────────</sub>

The seed is applied **in addition to** the deltas this module derives from the same block. `store_components` is written from `map_protocol_components` at `start_block` and stores run before their downstream maps within a block, so the `get_last` guards below already pass at `start_block` and `emit_balance_deltas_for_eeth`/`_for_weeth` emit that block's deltas too.

`compute_initial_state.sh` reads the snapshot at `--block $BLOCK_NUMBER`, i.e. the state *after* `start_block` executed. So any LiquidityPool deposit/redemption or weETH wrap/unwrap inside `start_block` is counted twice, and because `store_component_balances` is an `add` store the offset is permanent — the component balance (and the `liquidityPoolNativeBalance` attribute derived from it at line 337) stays wrong for the life of the integration.

Block 25940000 happens to contain no EtherFi activity, which is why the current params look right:

```
$ cast balance 0x3088…f216 --block 25939999 -> 1051493289032982041238
$ cast balance 0x3088…f216 --block 25940000 -> 1051493289032982041238
```

So the next person who reruns the script for a different start block gets a silently corrupt seed. Either read the snapshot at `start_block - 1`, or suppress the derived deltas at `start_block`.

<sub>────────────────────────────────────────────────────</sub>

crates/tycho-simulation/src/evm/protocol/etherfi/state.rs
```diff
@@ -350,14 +389,22 @@ impl ProtocolSim for EtherfiState {
...
-            return Ok((u256_to_biguint(U256::MAX), u256_to_biguint(U256::MAX)));
+            // Wrapping is uncapped by the protocol, but no more eETH can be wrapped than exists.
+            let max_eeth_amount = self.total_value_in_lp + self.total_value_out_of_lp;
+            let max_weeth_amount = self.shares_for_amount(max_eeth_amount)?;
+            return Ok((u256_to_biguint(max_eeth_amount), u256_to_biguint(max_weeth_amount)));
         }
 
         if sell_token.as_ref() == ETH_ADDRESS && buy_token.as_ref() == EETH_ADDRESS {
-            return Ok((u256_to_biguint(U256::MAX), u256_to_biguint(U256::MAX)));
+            let max_eth_amount = Self::uncapped_limit();
```
die-herdplatte: <sub>🤖 Claude Code ──────────────────────────</sub>

`uncapped_limit()` is `u128::MAX - 1` (~3.4e38 wei), chosen to be arithmetically safe for *this crate's* `U256` math — but `LiquidityPool.totalValueInLp` is a `uint128` on chain, and `deposit()` adds the msg.value to it under Solidity 0.8 checked arithmetic:

```solidity
// deposit() -> _deposit(msg.sender, msg.value, 0)
totalValueInLp += uint128(_amountInLp);
```

A deposit of `u128::MAX - 1` therefore reverts on overflow for **any** non-zero current `totalValueInLp` (it is ~1.05e21 today), as does anything within ~1e21 wei of the limit.

So the reported ETH -> eETH max_in is not a loose bound, it is a number that can never settle, and `get_amount_out` will happily quote it. `skip_execution: true` keeps the harness quiet, but a caller that sizes from the limit gets a revert rather than a partial fill.

`U256::from(u128::MAX) - self.total_value_in_lp` is a bound the venue can honour. `totalValueOutOfLp` does not belong in it: `_amountOutOfLp` is zero on this path, so that accumulator does not move, and the two are independent `uint128`s whose sum is only ever read through `getTotalPooledEther()` as a widened `uint256`. The other guards on the path add nothing — `amount > type(uint128).max` is subsumed once `totalValueInLp > 0`, and `_checkTotalValueInLp()` compares `totalValueInLp` against `address(this).balance`, which a deposit raises equally.

<sub>────────────────────────────────────────────────────</sub>

crates/tycho-simulation/src/evm/protocol/etherfi/state.rs
```diff
@@ -350,14 +389,22 @@ impl ProtocolSim for EtherfiState {
...
+            let max_eeth_amount = self.amount_for_share(self.shares_for_amount(max_eth_amount)?)?;
+            return Ok((u256_to_biguint(max_eth_amount), u256_to_biguint(max_eeth_amount)));
         }
 
-        Err(SimulationError::FatalError("unsupported swap".to_string()))
+        // A component only quotes the directions its venue supports (eETH cannot be minted from
+        // weETH's side, and staking is one-way). Report a zero limit rather than an error so
+        // callers skip the direction instead of treating the component as broken.
+        Ok((BigUint::ZERO, BigUint::ZERO))
```
die-herdplatte: <sub>🤖 Claude Code ──────────────────────────</sub>

Minor, and partly a judgement call: this fallback is now reached by *any* unrecognised pair, not just the three genuinely-unsupported EtherFi directions. Combined with #1431 (which skips zero-limit directions instead of failing the component), a token-address mismatch — exactly the class of bug the `ETH_ADDRESS` change in this PR fixes — no longer surfaces anywhere: the component indexes, decodes, and reports every direction as untradable, and the harness passes it.

Matching the four supported pairs explicitly and keeping the `FatalError` for anything else would preserve the "callers skip the direction" behaviour where it is intended while still making a mis-wired component loud.

<sub>────────────────────────────────────────────────────</sub>

protocols/substreams/ethereum-etherfi/src/storage.rs
```diff
@@ -0,0 +1,151 @@
...
+    number_of_bytes: 16,
+    signed: false,
+};
+
+const ETH_AMOUNT_LOCKED_FOR_WITHDRAWL_SLOT: StorageLocation = StorageLocation {
+    name: "ethAmountLockedForWithdrawl",
+    slot: hex!("00000000000000000000000000000000000000000000000000000000000000dc"),
+    offset: 1,
+    number_of_bytes: 16,
+    signed: false,
```
die-herdplatte: <sub>🤖 Claude Code ──────────────────────────</sub>

`ethAmountLockedForWithdrawal` does not exist in the deployed LiquidityPool, and the redemption liquidity it feeds is computed the way a superseded implementation computed it.

The implementations cited in this file's comments are not the ones behind the proxies. At head *and* at this package's own `start_block` (25940000), the EIP-1967 slots resolve to:

| Proxy | Live implementation | Cited here |
|---|---|---|
| LiquidityPool `0x3088…f216` | `0x17a16747d03006c9754548ac0d0aff48783a4a45` | `0xA5C1dd…` |
| eETH `0x35fA…8ac2` | `0xd1901dd36cbf4a81386d0162df2707f7ddb60527` | `0xcb3d91…` |
| RedemptionManager `0xDadE…7Ae0` | `0x5d53b303d62a7861f88650045b8d5deb59dfb3dc` | `0xe3f384…` |

In the live LiquidityPool the variable is gone — `cast call … 'ethAmountLockedForWithdrawal()'` reverts. Its slot is now a deprecated gap, and `initializeOnUpgradeV2` zeroed it:

```solidity
// Legacy `ethAmountLockedForWithdrawal` was a uint128 packed at bit offset 8
// (byte 1) of __gap_3[0]; read it out directly from that slot.
```

So this `StorageLocation` tracks `__gap_3[0]` under a name describing a variable that no longer exists. It reads zero today, but a later upgrade reusing the gap would feed an unrelated value into the redemption path as though it were locked ETH.

The consumer is wrong in the same way. `EtherfiState` computes

```rust
let liquid_eth_amount = liquidity_pool_native_balance - eth_amount_locked_for_withdrawl;
```

which is verbatim the cited implementation's

```solidity
return address(liquidityPool).balance - liquidityPool.ethAmountLockedForWithdrawal();
```

The live contract instead returns the accounting variable, and says why:

```solidity
function getInstantLiquidityAmount(address token) public view returns (uint256) {
    if(token == ETH_ADDRESS) {
        // Post-escrow-migration, locked ETH has physically left LP into the holder
        // contracts, so LP.balance already excludes it. No further subtraction needed.
        return liquidityPool.totalValueInLp();
```

`totalValueInLp` is already tracked here and already updated correctly, so the fix removes work rather than adding it: drop this slot and the `liquidityPoolNativeBalance` attribute, and use `total_value_in_lp` as the liquidity term in both `get_amount_out` and `get_limits`. That also resolves the attribute-name mismatch reported separately — there is no attribute left to freeze, and no subtraction left to underflow.

Nothing diverges numerically today because all three quantities coincide: 1051493289032982041238 wei at block 25940000 and 22470618011836097573240 wei at head, for `LP.balance`, `totalValueInLp()` and `getInstantLiquidityAmount(ETH)` alike. They are not the same quantity, though — `totalValueInLp` is accounting state, `LP.balance` is whatever anyone has sent the contract.

<sub>────────────────────────────────────────────────────</sub>

protocols/substreams/ethereum-etherfi/src/modules.rs
```diff
@@ -0,0 +1,476 @@
...
+                tx: Some(tx_trace.into()),
+                token: EETH_ADDRESS.to_vec(),
+                delta: wrap.e_eth_amount.to_signed_bytes_be(),
+                component_id: WEETH_ADDRESS
+                    .to_hex()
+                    .as_bytes()
+                    .to_vec(),
+            });
+        }
+        if let Some(wrap) = WrapWithPermit::match_and_decode(call) {
```
die-herdplatte: These should all be else-if's, right?

protocols/substreams/ethereum-etherfi/src/modules.rs
```diff
@@ -0,0 +1,476 @@
...
+    deltas.sort_unstable_by_key(|a| a.ord);
+    Ok(BlockBalanceDeltas { balance_deltas: deltas })
+}
+
+fn emit_balance_deltas_for_weeth(tx_trace: &TransactionTrace) -> Vec<BalanceDelta> {
+    // tokens: eETH and weETH, but we only track eeth balance here for accurate pool tvl
+    // decode function calls:
+    //  weETH.wrap, weETH.wrapWithPermit: eeth+
+    //  weETH.unwrap: eeth-
+    let mut deltas = vec![];
```
die-herdplatte: <sub>🤖 Claude Code ──────────────────────────</sub>

The weETH component's eETH balance is reconstructed by decoding three function calls, so it tracks only what those three calls move — and eETH's balance moves in two ways they do not cover.

**Anything that isn't `wrap`/`wrapWithPermit`/`unwrap`.** A plain `eETH.transfer(weETH, x)` donation never appears, and neither does any entrypoint added in a later weETH upgrade. `Transfer` events are the token's own accounting and cover both by construction, which is why `tycho_substreams::balances::extract_balance_deltas_from_tx` exists; eight packages use it, `ethereum-sky` among them — the package this one's `build.rs` is copied from.

The other half of this file already prefers the robust source. `emit_balance_deltas_for_eeth` decodes nothing: it reads `call.balance_changes` for the LiquidityPool, i.e. the EVM's own native-balance accounting, which catches every movement whether or not an ABI describes it. The two components are using opposite idioms with nothing in the code explaining why.

**Rebase, which neither calls nor events can see.** eETH emits `Transfer` on mint, burn and transfer only:

```solidity
emit Transfer(address(0), _user, amount);   // mint
emit Transfer(_user, address(0), amount);   // burn
emit Transfer(_sender, _recipient, _amount);
```

A rebase raises `balanceOf(weETH)` through `totalPooledEther` in the LiquidityPool and emits nothing on eETH at all. So the accumulated balance is `seed + Σwrap − Σunwrap`, which omits every rebase since the snapshot and drifts below the real balance at roughly the staking yield rate, permanently and silently. Switching to `Transfer` events fixes the first half and not this one.

Both go away if the balance is derived instead of accumulated. It is exactly `amountForShare(weETH.totalSupply())` — share math over values this package already indexes, with no delta path and nothing to seed. That is the same conclusion as the redemption-liquidity finding: the protocol's accounting variables are the truth here, and the delta path is reconstructing them less reliably than reading them.

Bounded in impact — balances feed TVL and routing weight, not quotes, which come from the tracked slots — so this is a slow TVL under-report rather than a mispricing. It is also invisible to the harness, since `skip_balance_check: true` is what suppresses the comparison that would catch it.

<sub>────────────────────────────────────────────────────</sub>

protocols/testing/scripts/update_runtime_bytecode.sh
```diff
@@ -156,6 +157,25 @@ process.stdout.write([e.contract].concat((e.args || []).map(String)).join(' '));
...
+plant_stub_code() {
+    for arg in "$@"; do
+        [[ "$arg" =~ ^0x[0-9a-fA-F]{40}$ ]] || continue
+        if [[ "$(cast code "$arg" --rpc-url "$LOCAL_RPC")" != "0x" ]]; then
+            continue
+        fi
+        echo "  No code at $arg on the fork; planting a stub."
+        cast rpc anvil_setCode "$arg" "0x00" --rpc-url "$LOCAL_RPC" >/dev/null
+    done
+}
```
die-herdplatte: <sub>🤖 Claude Code ──────────────────────────</sub>

Two problems: this silently absorbs a class of config error, and it exists to work around a constraint the script imposes on itself.

**It cannot tell a late deployment from a wrong address.** The guard is "this arg has no code on the fork", which is true both for EtherFi's redemption manager (deployed at 22090400, after the pinned block) and for an address that is simply wrong in `executor_deployments.json` — a typo, a testnet address, a copied-from-the-wrong-chain address. The second case is a config bug that `EtherfiExecutor__NotAContract()` was there to catch, and it now reports `No code at 0x…; planting a stub` and proceeds to bake the bad address into an immutable. Given that this same PR ships an executor whose five constructor args are all hardcoded addresses, that is the failure mode worth keeping loud. If the stub survives, it should be an explicit allowlist of address-plus-reason and an error for anything else.

**The constraint it works around is self-inflicted.** The obvious fix — move `FORK_BLOCK` past 22090400 — is blocked only because the deploy addresses depend on the fork block, via a deployer nobody chose deliberately. `export-runtime-bytecode.js:30` hardcodes

```js
new ethers.Wallet('0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80', provider)
```

which is anvil/hardhat account #0, `0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266`. Its key is published in every Foundry tutorial, so anyone can transact from it on mainnet, and they continuously do: nonce 789 at block 21000000, 8086 at head, balance swept to 0. anvil inherits that nonce rather than resetting it (verified: a fork at 21000000 reports 789), and the export deploys with nonce-based CREATE, so the runtime bytecode of every fixture baking `address(this)` — `TychoRouterV3` via EIP712, `UniswapV4Executor` — is a function of a nonce that third parties control. Pinning the block is what makes that reproducible at all.

Two consequences, one of them already latent here: `FORK_BLOCK` cannot move without regenerating every address-baking fixture, and because all fixtures share one anvil instance sequentially, each one's deploy address depends on **its position in `EXECUTOR_FIXTURES`**. Appending `Etherfi` last is what kept this diff to one fixture; inserting it anywhere else would have silently rewritten unrelated ones, and `--check` would have reported drift with no contract having changed.

Deploying from a key with no mainnet history (nonce 0 at every block), or `anvil_setNonce` before each deploy, or CREATE2 with a per-fixture salt, removes both. `FORK_BLOCK` is then free to move past 22090400 and this function is not needed. The fork itself should stay — `SkyExecutor`'s constructor reads `dai()`, `usds()`, `gem()` and `to18ConversionFactor()` from its args and bakes the results into immutables, so its fixture genuinely depends on live mainnet state and no stub can stand in for it.

<sub>────────────────────────────────────────────────────</sub>

die-herdplatte (changes requested):

die-herdplatte:
<sub>🤖 Claude Code ──────────────────────────</sub>

**Findings outside the diff**

These concern files this PR does not change, so GitHub has no diff position to anchor a comment to.

### `crates/tycho-simulation/src/evm/protocol/etherfi/state.rs:343` — `get_limits(weETH → eETH)` under-reports the sell cap by the share rate

```rust
let max_weeth_amount = self.shares_for_amount(self.total_shares)?;
let max_eeth_amount = self.total_shares;
```

`shares_for_amount` converts an eETH *amount* to a share count, but it is applied here to `total_shares`, which is already a share count. The result is `total_shares² / totalPooledEther` — the share count divided by the share rate.

The returned pair stays self-consistent (`amount_for_share(max_weeth_amount) == total_shares`, so a caller selling the stated maximum does receive the stated output), which is why nothing fails. It is simply smaller than it should be: at head the rate is ~1.10, so the cap sits ~9% under what unwrapping allows.

`WeETH.unwrap` is `liquidityPool.amountForShare(_weETHAmount)` with no cap of its own, so the bound is just how much weETH can exist — `total_shares` — paying out `amountForShare` of that, i.e. `totalPooledEther`. That is the mirror of the `eETH → weETH` arm a few lines below, which already reports `(totalPooledEther, total_shares)`:

```rust
let max_weeth_amount = self.total_shares;
let max_eeth_amount = self.total_value_in_lp + self.total_value_out_of_lp;
```

<sub>────────────────────────────────────────────────────</sub>

crates/tycho-simulation/src/evm/protocol/etherfi/state.rs
zach030: 🤖 Done — `U256::div_ceil` in both bucket conversions.

protocols/substreams/ethereum-etherfi/src/modules.rs
zach030: 🤖 Fixed in 2da5d84c3. Every tracked slot names its contract, and lookup matches on (contract, position).

Test `the_same_position_on_another_contract_is_not_tracked` covers exactly the collision you found on the LiquidityPool's slot 202.

protocols/substreams/ethereum-etherfi/src/modules.rs
zach030: 🤖 Fixed. The attribute and the whole native-balance path are gone with the thread below: redemption liquidity is `total_value_in_lp`.

To stop the class recurring rather than the instance, `delta_transition` now refuses any name the component does not carry (be8e13b41). A package-side rename fails the harness instead of freezing a value at its snapshot. The two names the stream decoder injects into every delta are exempt.

protocols/substreams/ethereum-etherfi/src/modules.rs
zach030: 🤖 Fixed in 2da5d84c3. The relative-delta path is gone.

Balances are derived from the tracked slots on every block that moves one, and `start_block` emits the snapshot only. Nothing accumulates, so there is nothing to double-count.

crates/tycho-simulation/src/evm/protocol/etherfi/state.rs
zach030: 🤖 Fixed in 2da5d84c3, and the real bound turned out tighter than `uint128`.

The same upgrade moved eETH minting behind an `EtherFiRateLimiter` bucket: 40,000 ETH capacity, 22.2 ETH/s refill. `eETH.mintShares` draws the minted balance from it, so the limit is `min(mint bucket, u128::MAX − totalValueInLp)`. That is a size the venue settles, and the harness now runs the eETH component with `skip_execution: false` — deposits execute at 40, 400 and 4,000 ETH.

crates/tycho-simulation/src/evm/protocol/etherfi/state.rs
zach030: 🤖 Agreed, done. `EtherfiState` is a `Venue` enum, so each component matches its own two pairs and anything else is a `FatalError` in `get_limits`, `get_amount_out` and `spot_price` alike.

A test walks every ordered pair of the tokens either component could hold and requires the unsupported ones to be refused by all three.

protocols/substreams/ethereum-etherfi/src/storage.rs
zach030: 🤖 Confirmed and fixed in 2da5d84c3. You were right about more than the one variable: all three proxies were upgraded together at block 25533308, two months before this package was written. The `0xdc` slot and `liquidityPoolNativeBalance` are gone, and redemption liquidity is `total_value_in_lp`. The same upgrade added the eETH mint and burn rate-limit buckets, now tracked as well.

Two guards so the class does not come back:

- `compute_initial_state.sh` verifies every tracked slot against the contract's own getter, and each proxy's EIP-1967 implementation against the one the slots were read from. An upgrade, or a slot that has become a gap, stops the snapshot instead of producing plausible values (be8e13b41).
- The executor's Foundry tests forked at 24332199, six months before the upgrade, so the redemption test exercised contracts that are no longer deployed. They fork at 25940000 now and raise the pool above its reserve floor first, so redemption actually runs against the live implementation (fbada9df8).

protocols/substreams/ethereum-etherfi/src/modules.rs
zach030: 🤖 That code went with the ABI decoding path — the weETH balance is derived from the tracked slots now.

protocols/substreams/ethereum-etherfi/src/modules.rs
zach030: 🤖 Fixed in 2da5d84c3, derived rather than accumulated, as you suggested.

One difference: it reads `eETH.shares(weETH)` rather than `weETH.totalSupply()`. The two differ by donated shares. `shares × totalPooledEther / totalShares` is `balanceOf(weETH)` by definition, checked to the wei against the chain at 25940000, and it follows a rebase with no delta at all.

protocols/testing/scripts/update_runtime_bytecode.sh
zach030: 🤖 Stub planting takes an allowlist of address plus reason now, and any other codeless argument stops the run (dac956970). The redemption manager is the one entry, with its deployment block.

The deployer-key point is right and I left it alone: it reaches every address-baking fixture, so it belongs in its own PR rather than buried in this one.

zach030:
🤖 Fixed in 2da5d84c3: `get_limits(weETH → eETH)` reports `(shares[weETH], amountForShare(shares[weETH]))`, the wrapper's actual eETH balance.

Since a self-consistent limit was what hid this, there is now a test that walks every ordered pair of the tokens either component could hold and requires each reported limit to quote at its own size and return exactly its own output. An under- or over-reported limit fails whichever direction it is on, and a direction added later is covered without a new case.

crates/tycho-simulation/src/evm/protocol/etherfi/state.rs
```diff
@@ -456,211 +763,1139 @@ impl ProtocolSim for EtherfiState {
...
+        }
+        let low_watermark = state
+            .low_watermark(&pool)
+            .expect("watermark");
+
+        let (max_in, _) = state
+            .get_limits(Bytes::from(EETH_ADDRESS), Bytes::from(ETH_ADDRESS))
+            .expect("limits");
+
+        assert_eq!(max_in, u256_to_biguint(state.total_value_in_lp - low_watermark));
```
die-herdplatte: <sub>🤖 Claude Code ──────────────────────────</sub>

The expected value is recomputed with the function under test, and that is hiding a slip in the test's own setup.

`max_in` comes from `get_limits`, which derives it as `total_value_in_lp - low_watermark(pool)`. The assertion then derives it the same way, from the same `low_watermark`. So the two sides cannot disagree: a wrong watermark — wrong bps field, wrong packing, `total_value_out_of_lp` omitted from the TVL — produces a wrong `max_in` and a matching wrong expectation, and the test still passes. What survives is the subtraction, which is not the part that can be wrong.

The setup says "settle on a fixed point", and the assertion is what stops anyone noticing that it does not. Printing the values at this line:

```
max_in  = 9999999977930797520
ten_eth = 10000000000000000000
```

22069202480 wei short of the 10 ETH the comment describes. Each iteration raises `total_value_in_lp`, which raises `getTotalPooledEther()`, which raises the floor again; five rounds land near the fixed point but not on it, and the final `low_watermark` is recomputed from a `total_value_in_lp` one step ahead of the one that produced it.

Either loop until it actually converges —

```rust
while state.total_value_in_lp - state.low_watermark(&pool)? != ten_eth {
    state.total_value_in_lp = state.low_watermark(&pool)? + ten_eth;
}
```

— and then assert the literal `ten_eth`, or pick `total_value_in_lp` and `total_value_out_of_lp` so that 1% of their sum is exact and write both the floor and the expected `max_in` as literals. Either way the expectation comes from the test rather than from `low_watermark`, and the 10 ETH in the comment becomes the thing being checked.

Low severity: the property the test is named for does hold, and every other test in the file uses literals or chain-derived values. Flagging it because it is the one assertion here that passes for a wrong implementation.

<sub>────────────────────────────────────────────────────</sub>

crates/tycho-simulation/src/evm/protocol/etherfi/state.rs
```diff
@@ -413,85 +647,111 @@ impl ProtocolSim for EtherfiState {
...
-                limit,
-                exit_fee_split_to_treasury_in_bps: exit_fee_split,
-                exit_fee_in_bps: exit_fee,
-                low_watermark_in_bps_of_tvl: low_watermark,
-            });
+        }
+        // A name this component does not carry is a package the simulation has drifted from:
+        // a rename, a typo, or an attribute meant for the other component. Applying the rest
+        // would leave the value it carries frozen at whatever the snapshot held.
+        if let Some(name) = attributes.keys().find(|name| {
```
die-herdplatte: Same as for Lido: https://github.com/propeller-heads/tycho/pull/929#discussion_r4023734943

crates/tycho-simulation/src/evm/protocol/etherfi/state.rs
```diff
@@ -413,85 +647,111 @@ impl ProtocolSim for EtherfiState {
...
+                    bucket.last_refill = value;
+                }
+                if let Some(value) = read_u64(names.refill_rate)? {
+                    bucket.refill_rate = value;
+                }
+                Ok(())
+            };
+
+        // Applied to a copy so a malformed attribute leaves `self` as it was.
+        let mut next = self.clone();
```
die-herdplatte: Same as for Lido: https://github.com/propeller-heads/tycho/pull/929#discussion_r4023735126

protocols/testing/scripts/update_runtime_bytecode.sh
die-herdplatte, line 1: Do we still need to make a change to this file here or will it be resolved by another PR?

die-herdplatte (changes requested):

crates/tycho-simulation/src/evm/protocol/etherfi/state.rs
zach030: 🤖 Fixed in fd8b31231, taking your second option: the two pool values are chosen so that 1% of their sum is exact, and the expected limit is the literal 10 ETH. No loop, so nothing to converge.

I checked your claim rather than taking it: with the old assertion restored, dropping `total_value_out_of_lp` from the watermark still passes. Under the new one that fails, and so does reading the exit-fee field instead of the watermark field.

You also suggested writing the floor as a literal. I only wrote the expected limit, so the 1,000 ETH floor is in a comment rather than in an assertion. Say the word and I will pin it too.

crates/tycho-simulation/src/evm/protocol/etherfi/state.rs
zach030: 🤖 Removed in 96118c097, `INJECTED_ATTRS` with it, same as on the Lido side.

One thing this package needed that Lido did not: my first attempt decoded all the attributes unconditionally, so the pool component started width-checking the wrapper's names and vice versa — the same forward-compatibility problem one layer down. Each component now decodes only the names it carries, pinned by `delta_transition_ignores_the_other_components_names`.

crates/tycho-simulation/src/evm/protocol/etherfi/state.rs
zach030: 🤖 Done in 96118c097, in your shape: every attribute is decoded before the first assignment, so nothing fallible runs once the state starts changing, and the copy is gone.

The comment gives your reason rather than restating the mechanism — the pending-block path applies a delta, logs whatever it returns and quotes from the result either way.

Still pinned: moving any assignment above a later `?` fails `delta_transition_rejects_an_attribute_wider_than_its_field`, which asserts the state is untouched.

protocols/testing/scripts/update_runtime_bytecode.sh
zach030: 🤖 No — removed in 1bb11d2fc. `update_runtime_bytecode.sh` is byte-identical to the branch point again, the fixture and its `execution.rs` entry are gone, and both components go back to `skip_execution: true`.

#1455 names this change as one of the two it supersedes, so it belongs there rather than here. EtherFi joins the fixtures once it lands.

Your point about the allowlist stands, and it was worse than one missing entry: enumerating every constructor argument without code at `FORK_BLOCK` turns up six addresses across three reasons — Uniswap V4's PoolManager (21688329), `0x0000000aa232…` (22971782), FermiSwap's (24950688), the redemption manager (22090400), and `0xEeee…EEeE`, which is the native-ETH sentinel and will never be a contract. My allowlist had one of them, which is why CI failed. The guard does not just conflate a late deployment with a wrong address; it also swallows an address that is not a contract by design.

The decoder registration in `state_registry.rs` stays, since that is what lets the harness quote these components at all.

protocols/testing/src/execution.rs
```diff
@@ -31,7 +31,6 @@ const EKUBO_V3_ROBINHOOD_BYTECODE_JSON: &str =
 const FLUIDV1_BYTECODE_JSON: &str = include_str!("../fixtures/FluidV1.runtime.json");
 const LIQUIDITYPARTY_BYTECODE_JSON: &str = include_str!("../fixtures/LiquidityParty.runtime.json");
 const SKY_BYTECODE_JSON: &str = include_str!("../fixtures/Sky.runtime.json");
-const ETHERFI_BYTECODE_JSON: &str = include_str!("../fixtures/Etherfi.runtime.json");
```
die-herdplatte: Do we still want to include it here even though we haven't changed `update_runtime_bytecode.sh`?
So we don't forget to set `skip_execution: false` later

protocols/substreams/ethereum-etherfi/src/upgrades.rs
```diff
@@ -0,0 +1,287 @@
...
+    fn a_slot_value_that_is_not_an_address_is_an_error() {
+        let mut change = rate_limiter_upgrade_to(OTHER);
+        change.new_value = vec![1u8; 32];
+        let block = block_with(vec![change], false);
+        assert!(detect_upgrades(&block, &initial_state()).is_err());
+    }
+}
+
+#[cfg(test)]
+mod review_regressions {
```
die-herdplatte: <sub>🤖 Claude Code ──────────────────────────</sub>

`review_regressions` names the process that produced these tests, not what they establish. A reader opening this file a year from now has no review to consult, and the name tells them nothing about the behaviour being pinned. Same module at `src/modules.rs:685`.

The tests themselves are the valuable part of this round — the nested-call cases are exactly what the sibling `tests` module was missing. Fold them in under names that say what holds (`nested_calls_use_the_last_executed_write`, and it already reads that way), and the two modules stop duplicating imports the sibling already has.

Worth noting they also format differently from every other test module here: no blank lines between items, so `#[test]` sits flush against the preceding `use`.

<sub>────────────────────────────────────────────────────</sub>

protocols/substreams/ethereum-etherfi/src/utils.rs
```diff
@@ -0,0 +1,46 @@
...
+
+pub fn bytes_from_hex(value: &str) -> Result<Vec<u8>> {
+    let value = value
+        .strip_prefix("0x")
+        .unwrap_or(value);
+    hex::decode(value).map_err(|e| anyhow!("Failed to decode hex value: {e}"))
+}
+
+/// Successful writes sorted by execution ordinal, including parent writes after child calls.
+pub fn ordered_storage_changes(tx: &TransactionTrace) -> Vec<&StorageChange> {
```
die-herdplatte: <sub>🤖 Claude Code ──────────────────────────</sub>

This collects and sorts *every* storage change in a transaction, and it runs twice per transaction on every block — once from `handle_state_updates`, once from `detect_upgrades` — to find at most a handful of keys. On a busy mainnet block that is tens of thousands of pointers allocated and sorted, nearly all of them discarded by the very next `tracked_slot` / key comparison.

Filtering before sorting gives the same answer for a fraction of the work: the relative order of the retained changes is unchanged, and the sorted subset is all either caller ever looks at. The callers already know what they want — a tracked slot, or the EIP-1967 position on a tracked proxy — so a predicate parameter, or returning the iterator and letting each caller filter then sort, both work.

To be clear about what must not change: the sort itself is load-bearing. The engine's `baseStore.set` appends to `kvOps` and `ApplyOps` replays them in call order, never in ordinal order — so the last *call* wins, not the highest ordinal, and `getFirst` scans `deltas` in the same call order. Sorting is what makes those two facts line up with execution order. Only the breadth of what gets sorted is the issue.

<sub>────────────────────────────────────────────────────</sub>

crates/tycho-simulation/src/evm/protocol/etherfi/state.rs
```diff
@@ -366,85 +676,135 @@ impl ProtocolSim for EtherfiState {
...
+                mint_bucket: read_bucket(&MINT_BUCKET)?,
+                burn_bucket: read_bucket(&BURN_BUCKET)?,
+                exit_fee_split_to_treasury_bps: read_u16(EXIT_FEE_SPLIT_TO_TREASURY_BPS_ATTR)?,
+                exit_fee_bps: read_u16(EXIT_FEE_BPS_ATTR)?,
+                low_watermark_bps: read_u16(LOW_WATERMARK_BPS_ATTR)?,
+            },
+            Venue::Wrapper(_) => VenueFields::Wrapper { weeth_shares: read(WEETH_SHARES_ATTR)? },
+        };
+
+        if let Some(value) = total_value_out_of_lp {
```
die-herdplatte, line 743: <sub>🤖 Claude Code ──────────────────────────</sub>

The comment above promises "Decode all fields before mutation so invalid deltas leave a consistent quote state", but the venue arm at line 786 returns `DecodeError` *after* `total_value_out_of_lp`, `total_value_in_lp` and `total_shares` have already been assigned to `self`. On that path the state is left half-updated, which is the outcome dropping the working copy was meant to rule out.

It is unreachable today — `venue_fields` is built from `self.venue` a few lines up, so the pair can never mismatch. But an arm exists because someone expects it to be reachable one day, and this one is positioned so that it cannot keep the guarantee its own comment makes.

Moving the three assignments below the `match`, or folding them into the matched arms, costs nothing and makes the atomicity structural rather than incidental.

<sub>────────────────────────────────────────────────────</sub>

die-herdplatte (changes requested):

protocols/testing/src/execution.rs
zach030: 🤖 The PR touches none of `execution.rs`, the fixture or `update_runtime_bytecode.sh`; the hunk this thread anchors to is from an earlier revision. The flag stays `skip_execution: true`, and its comment now names #1455 and the two steps to take once it lands, so it is not forgotten (5e40dd2aa).

protocols/substreams/ethereum-etherfi/src/upgrades.rs
zach030: 🤖 Folded into `tests` in both files under the names you suggested, with the sibling's imports and spacing (2e92f4c3e).

protocols/substreams/ethereum-etherfi/src/utils.rs
zach030: 🤖 It takes the caller's predicate now and sorts only the kept writes; each of the three callers passes the slot test it was about to apply. The sort stays, for the reason you gave, and the doc says so. Removing it fails three tests (2e92f4c3e).

crates/tycho-simulation/src/evm/protocol/etherfi/state.rs
zach030: 🤖 Comment rewritten to what the code guarantees: every fallible read precedes the first assignment, and the mismatch arm is unreachable because `venue_fields` is decoded against `self.venue`. The code is unchanged (27adc945d).

die-herdplatte (approved):

PR #1428, opened 2026-09-09
feat(simulation): fail closed on stale Titan price levels

tvinagre:
## Summary
- Parse Titan's frame `timestamp` and accept a frame only when it is younger than `stale_after`, not more than one slot in the future, not older than the newest accepted frame (equal allowed), and its block neither regresses nor jumps beyond what elapsed time explains. The block frontier lives in `ServingState::Serving`, so it exists only while something is served and one absurd block cannot wedge the stream; after every component turned stale the next frame is judged as a first frame, so the block number on updates may decrease across that gap (documented).
- Track freshness per component: a component no accepted frame carried for `stale_after` (24 s) is emitted in `removed_pairs`; the next accepted frame carrying it re-adds it. Omission from a frame no longer removes anything, since Titan frames are best-effort (7.7% partial in a 15 min capture).
- `PriceLevelStreamState` carries a monotonic, never-serialized one-slot quote guard and refuses to quote past it. `without_quote_guard()` turns it off for consumers that quote late by design; `eq` ignores the guard.
- The PropAMMRouter whitelist read is bounded (15 s), retried with backoff, refreshed every 10 min, and moves venues between families through a removal followed by a re-add. Nothing is served until the read succeeds. `build()` returns `Result<_, PriceLevelStreamBuildError>`: no node URL, an unparsable one, or a `stale_after` outside `(0, 1 h]` fail at build time instead of opening a stream that never serves. New builder knobs: `stale_after`, `whitelist_refresh_interval`, `fallback_router_rpc_url`, `without_quote_guard`.
- Only parsed frames count as transport liveness, measured while waiting on the socket, so a consumer that pauses between polls does not cause a reconnect; the idle timeout default drops from 30 s to 10 s.
- `price_level_stream_*` metrics through the `metrics` facade: label values are enums in `telemetry.rs`, per-venue series are pre-initialised to zero, no wire value becomes a label except the address of an auto-detected venue, and `price_level_stream_frame_age_seconds` records the signed age of every accepted frame (negative under a forward clock skew). Auto-detection registers at most 64 venues per process.
- The stream loop reads its clock from tokio, so the stale sweep agrees with the timer also under `tokio::time::pause()`.
- Module, builder, crate and solver docs rewritten for the new contract.

Behaviour change for in-repo consumers: the `price_printer` example panics and the integration-test processor returns an error when neither `RPC_URL` nor `fallback_router_rpc_url` is set. That is the intended fail-closed behaviour.

Pre-existing, unrelated to this branch: `evm::protocol::vm::state::tests::test_engine_block_advance_invalidates_cached_limits` (shared-fixture ordering, file identical to `main`) and `tycho-execution` `permit2::tests::test_permit` (RPC revert) fail locally; 26 rustdoc `-D warnings` errors outside `price_level_stream/` block a crate-wide strict doc build.

## Test plan
- [x] `cargo test -p tycho-simulation --lib price_level_stream`: 119 passed, 0 failed, 1 ignored, stable over repeated runs
- [x] Independent whole-branch review after the maintainer round: 0 critical findings; the 3 important ones (tokio vs std clock in the loop, skew-blind frame age, undocumented block regression after a full expiry) are fixed or documented in the last commit
- [x] `cargo +nightly-2026-06-28 fmt --all --check` and `cargo +nightly-2026-06-28 clippy -p tycho-simulation -p tycho-integration-test --all-targets --all-features -- -D warnings` clean
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc -p tycho-simulation --no-deps --all-features` warning-free inside `price_level_stream/`; `cargo check --workspace --all-targets --all-features` and `cargo check --workspace --no-default-features --locked` pass
- [x] Workspace nextest (no DB) before the review round: 2,301 passed, 2 pre-existing failures listed above
- [ ] Fynd canary on `ethereum-fynd-spot` after the release, watching `price_level_stream_serving_state`, `price_level_stream_served_components`, `price_level_stream_frames_rejected_total`, `price_level_stream_frame_age_seconds`, `price_level_stream_stale_removals_total`, `propamm_fallback_quotes_total`
- [ ] Follow-ups: Fynd consumer-contract test; #1466 (integration-test `SampledEmitter::handle` on removal-only updates)


tvinagre (commented):
**22 comments** — 0 Blocker · 3 Important · 9 Suggestion · 8 Nit · 2 Context.

### Important
- **DESIGN `#1`** `build()` returns a stream that never serves anything when no node URL is set — `stream.rs:305`
- **DESIGN `#2`** The block frontier is kept valid by call order, not by the type — `tracker.rs:332`
- **IMPL `#1`** Idle watchdog counts the time the consumer spends not polling — `titan.rs:147`

### Suggestions
- **DESIGN `#3`** `on_frame` clones the config per venue and the component per pair on every frame — `tracker.rs:351`
- **DESIGN `#4`** `RouterVenuesRead` carries failures that the tracker only drops — `fallback_router.rs:63`
- **DESIGN `#5`** Three builder fields encode one choice: where the whitelist comes from — `stream.rs:56`
- **DESIGN `#6`** `SnapshotTracker::new` takes seven positional parameters, two of them bare `bool`s — `tracker.rs:185`
- **DESIGN `#7`** Metric label values are string literals spread over three modules — `telemetry.rs:35`
- **IMPL `#2`** Frames older than `QUOTE_TTL` are accepted, served, and counted as healthy, but every state they emit refuses to quote — `tracker.rs:343`
- **IMPL `#3`** Removal-only updates break the held-snapshot logic of the in-repo consumer — `price_level_stream_processor.rs:158`
- **IMPL `#4`** The quote guard can lapse before the integration test simulates the held snapshot — `state.rs:180`
- **IMPL `#5`** A malformed node URL is retried forever as if it were a transient failure — `fallback_router.rs:105`

### Nits
- **DESIGN `#8`** The 12 s slot time is written four times — `tracker.rs:29`
- **DESIGN `#9`** `max_backoff` now also caps the whitelist read retries — `stream.rs:336`
- **DESIGN `#10`** `router_venues_reader` takes three adjacent `Duration`s — `fallback_router.rs:80`
- **DESIGN `#11`** The whitelist arm uses a helper where the deadline arm uses a guard — `stream.rs:373`
- **DESIGN `#12`** `build()` writes the `AwaitingWhitelist` gauge that `SnapshotTracker::new` writes next — `stream.rs:325`
- **DESIGN `#13`** `on_stale_deadline` takes a `Now` but reads only `now.instant` — `tracker.rs:476`
- **DESIGN `#14`** `quotable_until` is a `pub` field and also has a `pub` setter — `state.rs:60`
- **IMPL `#6`** `now.instant + stale_after` panics for very large durations — `tracker.rs:343`

### Context
- **BIZ `#2`** A missing or failing whitelist node now means no pAMM is served, where before the stream fell back to the direct family — `stream.rs:305`
- **BIZ `#5`** `tycho-integration-test` holds price level states before it quotes them, and the branch does not touch it — `state.rs:89`

crates/tycho-simulation/src/price_level_stream/stream.rs
```diff
@@ -255,504 +302,123 @@ impl PriceLevelStreamBuilder {
...
+        };
+        if fallback_router && rpc_url.is_none() {
+            tracing::error!(
+                "No node URL to read the PropAMMRouter whitelist from: set RPC_URL, call \
+                 fallback_router_rpc_url, or opt out with without_fallback_router. No pAMM will \
+                 be served"
             );
-            HashSet::new()
+            telemetry::source_state(SourceState::AwaitingWhitelist);
         }
```
tvinagre: This decision is correct, as it prevents users from getting state in a way they didn't sign-up to. If the user requests protocols with the fallback and can't get the fallback, they should be warned and not get the state. Worth a disclosure on the docs.

---

**[BIZ `#2`]**

**What changed**: Before this branch, a missing `RPC_URL` or a failed whitelist read logged a warning and served every venue under `pricelevelstream:`. Now the tracker starts in `Source::AwaitingWhitelist` and `on_frame` returns `None` for every frame until a read succeeds. Without a node URL the stream logs one ERROR and never serves anything. No commit subject states this change; "bound, retry, and refresh the PropAMMRouter whitelist read" reads as a hardening of the read only.

**Where**: `crates/tycho-simulation/src/price_level_stream/stream.rs:305–326`, `crates/tycho-simulation/src/price_level_stream/tracker.rs:237–246, 338–341`

- `crates/tycho-simulation/examples/price_printer/main.rs:173–177` builds the stream with no `fallback_router_rpc_url` and no `without_fallback_router`. Without `RPC_URL` in the environment, the example now shows no pAMM component.
- `docs/for-solvers/request-for-quote-protocols.md:262` still tells integrators that the builder "reads the whitelist once" and "without that variable it warns and keeps every venue on the direct path". The branch does not touch this file.
- A node URL that answers but never returns a decodable whitelist (wrong chain, no code at the router address) retries forever. The only signals are one WARN per failed read, `price_level_stream_whitelist_reads_total{outcome="error"}`, and `price_level_stream_source_state` staying at 0. The stream itself stays silent, which looks the same as a quiet Titan to a consumer.
- After one successful read, a failed refresh keeps the last known whitelist with no age limit (`crates/tycho-simulation/src/price_level_stream/tracker.rs:229–235`). Only the first read is fail-closed.

crates/tycho-simulation/src/price_level_stream/state.rs
```diff
@@ -60,7 +76,24 @@ impl PriceLevelStreamState {
...
+    /// Errors once `now` has reached `quotable_until`.
+    fn ensure_quotable(&self, now: Instant) -> Result<(), SimulationError> {
+        match self.quotable_until {
+            Some(until) if now >= until => Err(SimulationError::RecoverableError(
+                "price levels expired: the frame that carried them is older than one block"
+                    .to_string(),
+            )),
+            Some(_) | None => Ok(()),
+        }
     }
```
tvinagre, line 104: Please drop `quotable_until` from `PriceLevelStreamState::eq`: two states with the same ladders from two frames should compare equal, so consumers that skip unchanged states through `ProtocolSim::eq` keep working. On the integration-test timing: if the hold-back exceeds 12 s, Titan is lagging and the quote should fail, so that part is fine as is.

---

**[BIZ `#5`]**

**What changed**: `SampledEmitter` holds the chosen update until the stream moves past its block, then `main.rs` waits for that block to land on-chain, and only then calls `get_limits` and `get_amount_out` on the held state. Those calls now return `RecoverableError` once the state's frame is 12 s old, and the integration test records each one as a `get_limits` or `get_amount_out` failure for the venue. The emitter also now receives update shapes it did not see before: removal-only updates from the deadline timer, and block numbers that can go backwards after everything expired.

**Where**: `crates/tycho-simulation/src/price_level_stream/state.rs:89–97, 181, 208, 263`, `crates/tycho-simulation/src/price_level_stream/tracker.rs:520–524, 544–548`

- The time between Titan building the held frame and the quote calls at `crates/tycho-integration-test/src/main.rs:1633, 1700`. It is the hold-back in `crates/tycho-integration-test/src/stream_processor/price_level_stream_processor.rs:158–196`, plus `await_target_block` polling at `crates/tycho-integration-test/src/main.rs:1137–1146`, plus the sequential processing of the sampled components. I did not measure it; if it exceeds 12 s, the failure metrics rise with no change in Titan's quality.
- `removal_update` stamps a removal-only update with `newest_block`. In `SampledEmitter::handle` an update with the same block as the held one replaces it (`crates/tycho-integration-test/src/stream_processor/price_level_stream_processor.rs:189–192`), so a stale removal swaps the held states for an empty `states` map.
- `refresh_source` resets `newest_block` to 0 whenever nothing is served, so the next accepted frame can carry a lower block than earlier updates. The old `build` docs promised that "updates never move backwards in block number"; the emitter's `block > snapshot.block_number_or_timestamp` check (`crates/tycho-integration-test/src/stream_processor/price_level_stream_processor.rs:169`) relies on that order.
- `eq` now includes `quotable_until` (`crates/tycho-simulation/src/price_level_stream/state.rs:292–306`), so two states with the same ladders from two frames do not compare equal. Any consumer that skips unchanged states through `ProtocolSim::eq` now sees a change on every frame.

crates/tycho-simulation/src/price_level_stream/stream.rs
tvinagre: **[DESIGN `#1` · Important]**

**Description**: `build()` resolves `rpc_url` before it creates the stream, so it already knows the whitelist can never be read. It logs one error and still returns a stream that connects to Titan, parses every frame, and never yields an `Update`. Both in-repo consumers (`crates/tycho-simulation/examples/price_printer/main.rs:173–177` and `crates/tycho-integration-test/src/stream_processor/price_level_stream_processor.rs:70–74`) set no URL and wait on `stream.next()`, so without `RPC_URL` they hang with no signal besides the log line. The project's Rust conventions say to fail fast and to prefer returning results.

**Suggestion**: Return the failure from `build()`, for example `pub fn build(self) -> Result<impl Stream<Item = Update> + Send, BuildError>` with a `MissingFallbackRouterRpcUrl` variant. Pros: the caller sees the misconfiguration at startup; no WebSocket is opened for a stream that cannot serve; the `AwaitingWhitelist` gauge write in `build()` goes away. Cons: the signature of a public function changes; the two consumers need a `?` or an `expect`.

**Scope**: Every consumer of `PriceLevelStreamBuilder` that relied on the old warn-and-use-the-direct-path behavior.

crates/tycho-simulation/src/price_level_stream/tracker.rs
```diff
@@ -0,0 +1,1750 @@
...
+        self.rejecting = false;
+        telemetry::frame_accepted();
+        self.newest_timestamp = frame.timestamp;
+        self.newest_block = frame.block_number;
+        self.last_accepted = Some(now.instant);
+
+        if let Source::AwaitingWhitelist = self.source {
+            self.refresh_source();
+            return None;
+        }
```
tvinagre, line 345: **[DESIGN `#2` · Important]**

**Root cause**: The rule "a block frontier exists only while something is served" is spread over `newest_block`, `last_accepted`, `Source`, and `served` (`crates/tycho-simulation/src/price_level_stream/tracker.rs:120–129, 166–176`). No type holds the rule, so `refresh_source` enforces it as a side effect, next to its two other jobs (recompute `Source`, write the per-venue gauges).

- **Architecture**: both removal paths must build their `Update` before they call `refresh_source` (`crates/tycho-simulation/src/price_level_stream/tracker.rs:273–277, 504–509`), and two comments have to explain the order. Three "does not wedge the tracker" tests and the commit "reset the block frontier whenever nothing is served" show this rule already broke once.
- **Behavior**: while the whitelist is awaited, `on_frame` sets `newest_block` and `last_accepted` and then calls `refresh_source` only to reset both (`crates/tycho-simulation/src/price_level_stream/tracker.rs:334–341`).
- **Complexity**: to support that one call, `refresh_source` computes an `awaiting_whitelist` flag, builds a `served_state` it then may not assign, and carries a doc paragraph on why the whitelist wait must survive it (`crates/tycho-simulation/src/price_level_stream/tracker.rs:526–553`).

**Proposed fix**: Move the frontier into the state that owns it: `Source::Serving { deadline, newest_block, last_accepted }`. `accept` reads the frontier from `self.source`, `on_frame` returns early while the whitelist is awaited and builds `Serving` after the loop only when `served` is not empty, and the removal paths read `newest_block` from `Serving` before they change state. A frontier without a served component can then not be represented, so the ordering comments, the reset code, and the `awaiting_whitelist` special case all go away, and `refresh_source` shrinks to the gauge update. `newest_timestamp` stays a separate field because it is never reset. A smaller first step with no type change: in `on_frame`, set `newest_timestamp`, return `None` if the source is `AwaitingWhitelist`, and only then set `newest_block` and `last_accepted`; this removes the special case in `refresh_source` (about 8 lines) but keeps the ordering rule in the removal paths.

crates/tycho-simulation/src/price_level_stream/tracker.rs
```diff
@@ -0,0 +1,1750 @@
...
+                self.served.insert(
+                    id_string,
+                    Served {
+                        component,
+                        venue: config.protocol.clone(),
+                        address: config.address.clone(),
+                        deadline,
+                    },
+                );
+            }
```
tvinagre, line 403: **[DESIGN `#3` · Suggestion]**

**Description**: `resolve_config` returns an owned `PriceLevelStreamConfig` clone for every venue of every frame (`crates/tycho-simulation/src/price_level_stream/tracker.rs:407–410`). For a pair that is already served, `on_frame` clones the whole `ProtocolComponent` out of `served` and inserts a new `Served` with cloned `venue` and `address`, although only `deadline` changes. The code this diff replaces borrowed the config and moved the component. The clones exist because `resolve_config` and `merge_pairs` borrow all of `self`.

**Suggestion**: Make `merge_pairs` a free function over `&self.tokens`, let `resolve_config` only ensure the registry entry, then borrow `&self.registry[&pamm]` in `on_frame`. For a served pair, do `if let Some(served) = self.served.get_mut(&id) { served.deadline = deadline }` and build and insert a `Served` only for a new pair.

crates/tycho-simulation/src/price_level_stream/fallback_router.rs
```diff
@@ -44,12 +52,80 @@ pub enum FetchVenuesError {
...
+        after: Duration,
+    },
+}
+
+/// The outcome of one whitelist read, as delivered to the tracker.
+#[derive(Debug)]
+pub(super) enum RouterVenuesRead {
+    Ok(HashSet<Bytes>),
+    Failed(FetchVenuesError),
+}
```
tvinagre: **[DESIGN `#4` · Suggestion]**

**Description**: `RouterVenuesRead` is `Result<HashSet<Bytes>, FetchVenuesError>` under a new name. The reader already logs, counts, and retries every failure, and its doc tells consumers not to log failures again. `on_router_venues` still receives the `Failed` variant, logs it at debug, and returns `None` (`crates/tycho-simulation/src/price_level_stream/tracker.rs:228–235`). So the tracker depends on `fallback_router` types for a case in which it does nothing.

**Suggestion**: Let `router_venues_reader` yield only successful reads (`Stream<Item = HashSet<Bytes>>`) and change the tracker method to `on_router_venues(&mut self, venues: HashSet<Bytes>)`. Pros: the enum, the `Failed` arm, and the tracker's import of `fallback_router` go away. Cons: the two reader tests that match on `Failed` must assert on the fetch call count or on the `WHITELIST_READS` counter instead.

crates/tycho-simulation/src/price_level_stream/stream.rs
```diff
@@ -54,6 +56,16 @@ pub struct PriceLevelStreamBuilder {
...
+    /// See [`stale_after`](Self::stale_after).
+    stale_after: Duration,
+    /// See [`whitelist_refresh_interval`](Self::whitelist_refresh_interval).
+    whitelist_refresh_interval: Duration,
+    /// See [`fallback_router_rpc_url`](Self::fallback_router_rpc_url).
+    fallback_router_rpc_url: Option<String>,
+    /// Whether `RPC_URL` from the environment (or `.env`) is consulted when no explicit
+    /// fallback router URL is set. Only tests turn this off, to pin down the no-URL path
+    /// without touching process-global state.
+    env_rpc_url: bool,
```
tvinagre: **[DESIGN `#5` · Suggestion]**

**Description**: `fallback_router: bool`, `fallback_router_rpc_url: Option<String>`, and `env_rpc_url: bool` together describe one of three cases (no whitelist, explicit URL, URL from the environment). `build()` needs a four-arm match to decode them, one arm exists only to log an ignored URL, and `env_rpc_url` is a production field that only a `#[cfg(test)]` method changes. The project's Rust conventions ask for enums instead of boolean flags.

**Suggestion**: Store one `enum WhitelistSource { Disabled, Url(String), Env }` that `without_fallback_router` and `fallback_router_rpc_url` set. Move the environment lookup into a small function that takes the lookup as a parameter, so the no-URL test passes `|| None` and the `env_rpc_url` field and `without_env_rpc_url` go away. This also removes test-only state from the production struct: today the field, its `Default` entry, a destructure entry, and a match arm exist only so that `missing_node_url_serves_nothing_and_reports_awaiting_whitelist` can skip the `RPC_URL` lookup (`crates/tycho-simulation/src/price_level_stream/stream.rs:247–251`).

crates/tycho-simulation/src/price_level_stream/tracker.rs
```diff
@@ -0,0 +1,1750 @@
...
+impl SnapshotTracker {
+    pub(super) fn new(
+        registry: HashMap<Bytes, PriceLevelStreamConfig>,
+        denied: HashSet<Bytes>,
+        tokens: HashMap<Bytes, Token>,
+        auto_detect: bool,
+        auto_detected_gas_cost: BigUint,
+        stale_after: Duration,
+        fallback_router: bool,
+    ) -> Self {
```
tvinagre: **[DESIGN `#6` · Suggestion]**

**Description**: This diff adds `stale_after` and `fallback_router` to a list that already ended in `auto_detect, auto_detected_gas_cost`. The ten call sites in the tests read `false, BigUint::from(..), DEFAULT_STALE_AFTER, false`, and nothing stops a caller from swapping the two `bool`s. `fallback_router` is used only to pick the first `Source`, so the `bool` hides a state choice.

**Suggestion**: Pass one `TrackerSettings` struct with named fields (the builder already holds all of them), or at least replace `fallback_router: bool` with an enum such as `Whitelist::Awaited` / `Whitelist::NotUsed`. The test helpers can then use struct update syntax instead of repeating seven arguments.

crates/tycho-simulation/src/price_level_stream/telemetry.rs
```diff
@@ -0,0 +1,165 @@
...
+    gauge!(SOURCE_STATE).set(state as u8 as f64);
+}
+
+pub(super) fn reconnect(reason: &'static str) {
+    counter!(RECONNECTS, "reason" => reason).increment(1);
+}
+
+pub(super) fn whitelist_read(outcome: &'static str) {
+    counter!(WHITELIST_READS, "outcome" => outcome).increment(1);
+}
```
tvinagre: **[DESIGN `#7` · Suggestion]**

**Description**: The `telemetry` module doc promises "fixed enumerations" as label values, but only `source_state` takes an enum. `frame_rejected`, `reconnect`, and `whitelist_read` take `&'static str`. The `reason` values of `FRAMES_REJECTED` come from `Rejection::as_str` in `crates/tycho-simulation/src/price_level_stream/tracker.rs:96–106` and from the literal `"parse_error"` in `crates/tycho-simulation/src/price_level_stream/titan.rs:191`. The reconnect reasons are seven literals in `crates/tycho-simulation/src/price_level_stream/titan.rs:155–235` (`"idle_timeout"` twice), and the read outcomes are two literals in `crates/tycho-simulation/src/price_level_stream/fallback_router.rs:100, 107`. No single place lists the label values a dashboard can rely on, and a typo compiles.

**Suggestion**: Define `RejectReason`, `ReconnectReason`, and `ReadOutcome` enums in `telemetry.rs` with one `as_str` each, and make the helpers take them. `Rejection` in the tracker can convert into `RejectReason`, or the tracker can use `RejectReason` directly with a `ParseError` variant for `titan.rs`.

crates/tycho-simulation/src/price_level_stream/tracker.rs
```diff
@@ -0,0 +1,1750 @@
...
+
+pub(super) const NANOS_PER_SECOND: u64 = 1_000_000_000;
+
+/// Furthest a frame's `timestamp` may lie ahead of the local wall clock: one slot. Titan's
+/// frames were observed at least 120 ms behind the local clock; anything ahead of it by more
+/// than clock skew is implausible.
+const MAX_FUTURE_SKEW_NANOS: u64 = 12 * NANOS_PER_SECOND;
+
+/// Post-merge Ethereum slot duration in seconds.
+const SECONDS_PER_SLOT: u64 = 12;
```
tvinagre: **[DESIGN `#8` · Nit]**

`DEFAULT_STALE_AFTER` (24), `MAX_FUTURE_SKEW_NANOS` (12 × nanos), `SECONDS_PER_SLOT` (12), and `QUOTE_TTL` in `crates/tycho-simulation/src/price_level_stream/state.rs:25` each encode the block time on their own. Define one `SLOT: Duration`, derive the other three from it, and use "slot" in every doc instead of mixing "slot" and "block time".

crates/tycho-simulation/src/price_level_stream/stream.rs
```diff
@@ -255,504 +302,123 @@ impl PriceLevelStreamBuilder {
...
-                        tracing::debug!(%pamm, "Skipping unregistered pAMM");
-                        continue;
-                    }
-                    if self.denied.contains(&pamm) {
-                        tracing::debug!(%pamm, "Skipping denied pAMM");
-                        continue;
+            stale_after,
+            fallback_router,
+        );
+        let max_backoff = connection.max_backoff;
```
tvinagre: **[DESIGN `#9` · Nit]**

The builder's `max_backoff` knob is documented as the Titan reconnect cap (`crates/tycho-simulation/src/price_level_stream/stream.rs:143–148`), and `fallback_router.rs` imports `backoff` from `titan.rs`, which this diff widens to `pub(super)` (`crates/tycho-simulation/src/price_level_stream/titan.rs:58`). Either document the shared knob or give the reader its own cap, and move `backoff` out of the module that says it holds only Titan specifics.

crates/tycho-simulation/src/price_level_stream/fallback_router.rs
```diff
@@ -44,12 +52,80 @@ pub enum FetchVenuesError {
...
+/// `read_timeout`; failures (including timeouts) are retried with exponential backoff
+/// (`2^attempt` seconds, capped at `max_backoff`); successes are repeated every
+/// `refresh_interval`. Never ends. Logs one WARN per failure; consumers of the outcomes should
+/// not log them again.
+pub(super) fn router_venues_reader<F, Fut>(
+    fetch: F,
+    read_timeout: Duration,
+    max_backoff: Duration,
+    refresh_interval: Duration,
+) -> impl Stream<Item = RouterVenuesRead> + Send
```
tvinagre: **[DESIGN `#10` · Nit]**

`read_timeout`, `max_backoff`, and `refresh_interval` are easy to swap at the call site, as in the tests' `(1 s, 5 ms, 20 ms)`. Pass a small settings struct with named fields.

crates/tycho-simulation/src/price_level_stream/stream.rs
```diff
@@ -255,504 +302,123 @@ impl PriceLevelStreamBuilder {
...
+/// all, so the `select!` arm simply never fires.
+async fn next_whitelist_read(reader: &mut Option<WhitelistReader>) -> RouterVenuesRead {
+    let Some(reader) = reader else {
+        return std::future::pending().await;
+    };
+    match reader.next().await {
+        Some(read) => read,
+        None => std::future::pending().await,
+    }
 }
```
tvinagre: **[DESIGN `#11` · Nit]**

Build the reader as a plain stream, with `futures::stream::pending()` boxed when there is no URL. The arm then becomes `Some(read) = whitelist.next()`, like the frames arm, and `Option<WhitelistReader>` and `next_whitelist_read` go away.

crates/tycho-simulation/src/price_level_stream/stream.rs
```diff
@@ -255,504 +302,123 @@ impl PriceLevelStreamBuilder {
...
+                .flatten(),
+        };
+        if fallback_router && rpc_url.is_none() {
+            tracing::error!(
+                "No node URL to read the PropAMMRouter whitelist from: set RPC_URL, call \
+                 fallback_router_rpc_url, or opt out with without_fallback_router. No pAMM will \
+                 be served"
             );
-            HashSet::new()
+            telemetry::source_state(SourceState::AwaitingWhitelist);
```
tvinagre: **[DESIGN `#12` · Nit]**

`SnapshotTracker::new` already emits `AwaitingWhitelist` when `fallback_router` is true (`crates/tycho-simulation/src/price_level_stream/tracker.rs:194–195`). Drop the call in `build()` so only the tracker writes the source state.

crates/tycho-simulation/src/price_level_stream/tracker.rs
```diff
@@ -0,0 +1,1750 @@
...
+                entry.0 = quotes;
+            } else {
+                entry.1 = quotes;
+            }
+        }
+        merged
+    }
+
+    /// Removes every served component whose deadline has passed, as one [`Update`].
+    pub(super) fn on_stale_deadline(&mut self, now: Now) -> Option<Update> {
```
tvinagre: **[DESIGN `#13` · Nit]**

Take `now: Instant`, so the signature shows that expiry never depends on the wall clock.

crates/tycho-simulation/src/price_level_stream/state.rs
```diff
@@ -42,6 +53,11 @@ pub struct PriceLevelStreamState {
     pub quotes_0_to_1: Vec<PriceLevelStreamQuote>,
     pub quotes_1_to_0: Vec<PriceLevelStreamQuote>,
     pub gas_cost: BigUint,
+    /// Monotonic instant from which every query on this state is refused. A live, in-process
+    /// property: never serialized, and `None` (never expires) after deserialization, so
+    /// recordings replay as they always did.
+    #[serde(skip)]
+    pub quotable_until: Option<Instant>,
```
tvinagre: **[DESIGN `#14` · Nit]**

Any holder of a state can clear the guard by assigning `None`. Make the field private with a getter, and keep `with_quotable_until` as the only way to set it.

crates/tycho-simulation/src/price_level_stream/titan.rs
```diff
@@ -129,39 +140,55 @@ pub(super) fn messages(
...
+                            .read_idle_timeout
+                            .saturating_sub(last_parsed.elapsed());
+                        if remaining.is_zero() {
+                            warn!(
+                                idle_secs = settings.read_idle_timeout.as_secs(),
+                                "No parsed Titan frame within idle timeout; reconnecting"
+                            );
+                            telemetry::reconnect("idle_timeout");
+                            break;
+                        }
```
tvinagre: **[IMPL `#1` · Important]**

**Category**: bug

**Description**: `last_parsed` is set before `yield message`, and the generator stays suspended at that `yield` until the consumer polls again. If the consumer does not poll for `read_idle_timeout` (10 s by default), the next loop pass finds `remaining.is_zero()`, logs "No parsed Titan frame within idle timeout", counts `reconnect("idle_timeout")`, drops a healthy socket, and backs off 2 s. Before the diff the timeout started fresh on every `ws_stream.next()`, so a slow consumer never caused a reconnect. I checked this by reading `messages` and the `yield update` in `crates/tycho-simulation/src/price_level_stream/stream.rs:365–367`; I did not run it.

**Plain English**: An integrator whose service pauses for ten seconds or more (a long simulation batch, a full downstream queue) comes back to find the price feed has disconnected itself and reconnected. The log and the reconnect metric blame a silent upstream, although the upstream was sending data the whole time. The operator then investigates a network problem that does not exist. The feed should only reconnect when the upstream really sent nothing while the service was listening.

**Suggestion**: Reset the watchdog when the generator resumes: `yield message; last_parsed = Instant::now();`. The idle window then measures only time spent waiting on the socket.

crates/tycho-simulation/src/price_level_stream/tracker.rs
```diff
@@ -0,0 +1,1750 @@
...
+
+        if let Source::AwaitingWhitelist = self.source {
+            self.refresh_source();
+            return None;
+        }
+
+        let deadline = now.instant +
+            self.stale_after
+                .saturating_sub(frame_age);
+        let quotable_until = now.instant + QUOTE_TTL.saturating_sub(frame_age);
```
tvinagre: Not asking to reject these frames: the 12 s rule is flaky on chains with variable block time, and we don't know how far into the block we are, so a timestamp comparison is not a reliable rejection criterion. Please add a frame-age histogram at acceptance instead, so we can measure Titan's lag and act on the data. Ignore the last sentence of the suggestion about not emitting `new_pairs`.

---

**[IMPL `#2` · Suggestion]**

**Category**: bug

**Description**: `accept` lets a frame through when its age is below `stale_after` (24 s), but `quotable_until` is `now + QUOTE_TTL.saturating_sub(frame_age)`, so a frame aged 12–24 s emits states whose guard has already lapsed. Such a frame still increments `frame_accepted`, re-adds components in `new_pairs`, pushes their deadline forward, and keeps `source_state` at `Serving`. A local wall clock that runs 12–24 s ahead of Titan's (or a delivery delay of that size) puts every frame in this band: all metrics look healthy and nothing logs, yet every `get_amount_out`, `spot_price` and `get_limits` returns "price levels expired". The test `frame_older_than_one_block_yields_an_unquotable_state` shows the state behavior is intended; the missing signal is the problem.

**Plain English**: An operator's dashboard shows the feed connected, frames accepted, and components served. Meanwhile every price request against those components is refused as expired, and routing silently stops using these venues. Nothing in the logs or metrics points at the cause, which is a clock offset or a slow link. The operator should get a clear signal when accepted data arrives too old to be quoted.

**Suggestion**: Export the frame age at acceptance (a gauge or histogram in seconds) and count or WARN once per streak when `frame_age >= QUOTE_TTL`. Consider not emitting `new_pairs` for a component whose first state is already unquotable.

crates/tycho-simulation/src/price_level_stream/tracker.rs
```diff
@@ -0,0 +1,1750 @@
...
+            Source::Serving { deadline } => Some(deadline),
+            Source::AwaitingWhitelist | Source::Unserved => None,
+        }
+    }
+
+    fn removal_update(&self, removed: HashMap<String, ProtocolComponent>) -> Update {
+        Update::new(self.newest_block, HashMap::new(), HashMap::new())
+            .set_is_partial(true)
+            .set_removed_pairs(removed)
+    }
```
tvinagre: Out of scope for this PR since it only affects `tycho-integration-test`, but that consumer lives in this repo and its metrics are what we watch, so please open a follow-up ticket for it.

---

**[IMPL `#3` · Suggestion]**

**Category**: bug

**Description**: `removal_update` (`crates/tycho-simulation/src/price_level_stream/tracker.rs:520–524`) now emits updates with empty `states` stamped with `newest_block`, from both `on_stale_deadline` and `on_router_venues`. `SampledEmitter::handle` treats every update as a frame: a removal-only update whose block equals the held snapshot's block replaces it (`*snapshot = update`), so if it is the last update of that block the emitted sample has no states. `handle` also calls `mark_protocol_ready` on every update, so an expiry caused by Titan silence flips the venues back to Ready. Before the diff a removal almost always rode on a full frame. I read `handle` and `emit`; the consumer file is not changed by this branch.

**Plain English**: The integration test samples one snapshot per block interval and simulates it. When a component expires near the end of a sampled block, the sample that gets simulated is empty and that interval checks nothing. When the upstream goes quiet and components expire, the health gauge briefly reports the venues as ready again at exactly the moment they went stale. A removal should update the component cache without replacing the held snapshot or signalling liveness.

**Suggestion**: In `handle`, skip the snapshot replacement, the `chosen` selection, and the ready marking when `update.states` is empty. Add a consumer test that feeds a removal-only update for the held block.

**Scope**: The consumer file is outside the diff, but the behavior that breaks it (state-less removal updates carrying the current block) is introduced here.

crates/tycho-simulation/src/price_level_stream/state.rs
```diff
@@ -144,6 +178,7 @@ impl ProtocolSim for PriceLevelStreamState {
     }
 
     fn spot_price(&self, base: &Token, quote: &Token) -> Result<f64, SimulationError> {
+        self.ensure_quotable(Instant::now())?;
```
tvinagre, line 188: We should have a way to bypass quote_guard. `without_quote_guard` might be a good implementation for slow quoters. Make sure it's well documented.

---

**[IMPL `#4` · Suggestion]**

**Category**: bug

**Description**: Every emitted state now refuses `spot_price`, `get_amount_out` and `get_limits` once its frame is 12 s old. `tycho-integration-test` holds a snapshot until the stream moves past its block, sends it through a bounded channel, then waits up to 10 × 500 ms in `await_target_block` (`crates/tycho-integration-test/src/main.rs:1137–1146`) before it calls `get_limits` and `get_amount_out` (`crates/tycho-integration-test/src/main.rs:1633, 1700`). Any queueing behind a slow protocol update, a lagging RPC, or a frame that targets head + 2 (which `BLOCK_JUMP_SLACK` says Titan does) pushes the simulation past the guard. The result is `get_limits_failure` / `get_amount_out_failure` for every pAMM component of that sample. I checked the code path; I could not measure the real latency between frame build time and simulation.

**Plain English**: The integration test picks a price snapshot, waits for its block to land on chain, and then asks the snapshot for prices. With the new expiry, a wait of more than a few seconds makes every one of those price requests fail as expired. The dashboards then show simulation failures for these venues although the quotes were fine when they were made. The test should either simulate before the quotes expire or not count an expired quote as a venue failure.

**Suggestion**: Measure the delay in the integration test (frame timestamp to first `get_limits`). If it can exceed the guard, have the processor clear the guard on states it holds back for on-chain validation (a `without_quote_guard` on `PriceLevelStreamState`), or record expiry under its own metric.

**Scope**: Affects `tycho-integration-test` only; live routing consumers quote immediately and are not affected.

crates/tycho-simulation/src/price_level_stream/fallback_router.rs
```diff
@@ -44,12 +52,80 @@ pub enum FetchVenuesError {
...
+                    let delay = backoff(attempt, max_backoff);
+                    tracing::warn!(
+                        error = %error,
+                        attempt,
+                        retry_secs = delay.as_secs_f64(),
+                        "PropAMMRouter whitelist read failed; retrying"
+                    );
+                    yield RouterVenuesRead::Failed(error);
+                    sleep(delay).await;
+                }
```
tvinagre, line 120: **[IMPL `#5` · Suggestion]**

**Category**: bug

**Description**: `router_venues_reader` handles every `FetchVenuesError` the same way: WARN "retrying", back off, call again. `FetchVenuesError::InvalidUrl` (`crates/tycho-simulation/src/price_level_stream/fallback_router.rs:137`) can never succeed, so a typo in `fallback_router_rpc_url` or `RPC_URL` leaves the stream in `AwaitingWhitelist` forever while it logs a retry every 32 s. The missing-URL case gets one clear ERROR in `build()` (`crates/tycho-simulation/src/price_level_stream/stream.rs:319–326`); the unparsable-URL case gets a message that says the opposite of what is true.

**Plain English**: An operator who mistypes the node address starts the service and sees no pAMM prices at all. The log says the whitelist read failed and will be retried, so the operator waits for it to recover. It never does, because the address can never work. The service should say once, clearly, that the configured address is invalid and nothing will be served until it is fixed.

**Suggestion**: Parse the URL once in `build()` and treat a parse failure like the missing-URL case (one ERROR, no reader). Alternatively log `InvalidUrl` at ERROR in the reader without the word "retrying".

crates/tycho-simulation/src/price_level_stream/tracker.rs
```diff
@@ -0,0 +1,1750 @@
...
+        self.last_accepted = Some(now.instant);
+
+        if let Source::AwaitingWhitelist = self.source {
+            self.refresh_source();
+            return None;
+        }
+
+        let deadline = now.instant +
+            self.stale_after
+                .saturating_sub(frame_age);
```
tvinagre: **[IMPL `#6` · Nit]**

`Instant + Duration` panics on overflow, so `stale_after(Duration::MAX)` (a natural way to try to disable expiry) panics on the first accepted frame, and `stale_after(Duration::ZERO)` rejects every frame as `too_old`. Clamp or validate the value in the builder setter, or use `checked_add`.

tvinagre:
## Review follow-up

I pushed four commits on top of `1294dfcd6` that resolve the naming, documentation, test and small-simplification findings from my review, so the review comments above are only what still needs a decision from you.

- `39bf79660` refactor: renames — `SnapshotTracker` → `FreshnessTracker`, `Served` → `ServedComponent` (`venue_name`, `venue_address`, `stale_at`), `Source` → `ServingState` (metric `price_level_stream_serving_state`), telemetry helpers → `record_*` with label `venue`, `..._unregistered_pamm_entries_total`, `check_frame`, `logged_unregistered`, `newest_timestamp_nanos`, `Now::monotonic`, `whitelist_reader` / `WhitelistRead` / `on_whitelist_read`, `telemetry::recorded`.
- `e7ec05578` docs: doc-comments, inline comments and log strings rewritten in one vocabulary (a component "turns stale", only a state "expires", "slot", `newest_block`/`last_accepted` instead of "frontier", "implausible block"); every metric constant documented with labels and values, listed under `# Observability`; solver docs page and crate `CLAUDE.md` updated to the current whitelist and best-effort-frame behaviour.
- `56a69b3be` tests: pins the removal update's block number, family change of one of two venues in both directions, gauges falling back to zero, a successful whitelist read serving `propammfallback:`, `connect_failed` / `closed` / `parse_error` / `whitelist_reads_total{outcome="ok"}` metric assertions; telemetry test looks metrics up by literal name; split multi-run tests; shared `frame_then_repeat`, `record_async`, `fermiswap` helpers; deleted `set_router_venues_for_test` and the tautological default tests; `fast_builder` `stale_after` widened to 1 s. `build()` gained a private `build_with_whitelist` seam for the reader test.
- `87cad39a0` refactor: both removal paths use `HashMap::extract_if`; the two idle-timeout exits in `titan.rs` are one; the read timeout is an `unwrap_or`.

**Still yours to address** — the inline review comments (Design, Implementation, Business Logic). Some of them now show as outdated because the renames moved lines; the findings still apply. Three I left alone because they depend on your call on those comments:
- DESIGN `#2` vs. deriving `ServingState::Serving`'s deadline from `served` — the two pull in opposite directions; pick one.
- DESIGN `#4` (`WhitelistRead` → `Result`, or the reader yielding successes only) — I kept the enum and the tracker's DEBUG log on a failed read until you decide.
- The `serving_state` field / `refresh_serving_state` shape follows DESIGN `#2` whichever way you take it.

Tests: `cargo test -p tycho-simulation price_level_stream` → 110 passed; clippy clean for the module.

crates/tycho-simulation/src/price_level_stream/stream.rs
tvinagre: Disclosed in `docs/for-solvers/request-for-quote-protocols.md` (pAMM Price Level Stream section) and in the `price_level_stream` module docs: `build()` now returns an error without a node URL or with one that does not parse, and with a reachable node that does not answer the stream serves nothing until the first read succeeds. 499e7cc3b.

crates/tycho-simulation/src/price_level_stream/state.rs
tvinagre: Dropped `quotable_until` from `eq` in 499e7cc3b; the field is private now and read through `quotable_until()`. One caveat is documented on `eq`: a consumer that skips a state because it equals the one it holds keeps quoting on the older guard. No consumer does that today (Fynd only delegates `eq` inside a wrapper, and `tycho-common` documents `ProtocolSim::eq` as a test aid), so equality follows the ladders. The integration-test timing is left as is, per your note.

crates/tycho-simulation/src/price_level_stream/stream.rs
tvinagre: Done in 499e7cc3b: `build()` returns `Result<impl Stream<Item = Update> + Send, PriceLevelStreamBuildError>` with `MissingFallbackRouterRpcUrl`, `InvalidFallbackRouterRpcUrl { url, reason }` (IMPL #5) and `StaleAfterOutOfRange { given }` (IMPL #6). No socket is opened and no gauge is written on the error path. `price_printer` uses `expect`; the integration-test processor propagates it with `into_diagnostic()?`. Tests: `missing_node_url_fails_to_build`, `invalid_node_url_fails_to_build`, `stale_after_outside_its_range_fails_to_build`.

crates/tycho-simulation/src/price_level_stream/tracker.rs
tvinagre: Took the type change in 499e7cc3b: `ServingState::Serving { deadline, frontier: Frontier { block, accepted_at } }`. `check_frame` reads the frontier from the state (`let ServingState::Serving { .. } = self.serving_state else { return Ok(frame_age) }`), `on_frame` returns before touching anything while the whitelist is awaited and calls `settle(Frontier { block, accepted_at })` after the loop, which yields `Unserved` when nothing is served. Both removal paths take their block from the `Serving` frontier before `settle`, so the ordering comments, the reset code and the `awaiting_whitelist` special case are gone; `refresh_serving_state` became `settle` (state plus gauges) and `set_serving_state`.

On the question in your follow-up: I kept `deadline` stored in `Serving` next to the frontier, recomputed at every `settle`, so `stale_deadline()` stays O(1) and there is exactly one place that derives the state from `served`.

crates/tycho-simulation/src/price_level_stream/tracker.rs
tvinagre: Done in 499e7cc3b: `merge_pairs` is a free function over `&tokens`; `resolve_config` became `admit(&mut self, pamm)`, which only inserts an auto-detected config, and `on_frame` then borrows the config through `self.registry.get(&pamm)`. A served pair is refreshed with `get_mut(..).stale_at = stale_at`; a `ServedComponent` and the component clone for `new_pairs` are built only for a new pair.

crates/tycho-simulation/src/price_level_stream/fallback_router.rs
tvinagre: Done in 499e7cc3b: `whitelist_reader` yields `HashSet<Bytes>` only, `WhitelistRead` is gone, and the tracker's `on_whitelist_read(venues: HashSet<Bytes>)` no longer imports `fallback_router`. The reader tests assert on the `whitelist_reads_total` counter and the fetch call count (`reader_retries_failures_and_refreshes_after_success`, `reader_fails_a_read_that_never_resolves`, `failed_reads_are_counted_and_never_yielded`). The two tracker tests for failed reads were removed since a failure can no longer reach the tracker; `whitelist_read_while_unserved_applies_to_the_next_frame` covers the remaining tracker path.

crates/tycho-simulation/src/price_level_stream/stream.rs
tvinagre: Done in 499e7cc3b: `WhitelistSource { Disabled, Url(String), Env }`, set by `without_fallback_router` and `fallback_router_rpc_url` (the later call wins, documented on both). The environment lookup is a parameter of the private `whitelist_reader(&self, env_rpc_url: impl FnOnce() -> Option<String>)` seam that `build()` calls with `rpc_url_from_env`; `env_rpc_url` and `without_env_rpc_url` are gone and the no-URL test passes `|| None`.

crates/tycho-simulation/src/price_level_stream/tracker.rs
tvinagre: Done in 499e7cc3b: `FreshnessTracker::new(TrackerSettings { .. })` with named fields, and `fallback_router: bool` replaced by `Whitelist::Awaited | Whitelist::NotUsed`. The test helpers use struct update syntax over a `settings(configs)` base.

crates/tycho-simulation/src/price_level_stream/telemetry.rs
tvinagre: Done in 499e7cc3b: `RejectReason`, `ReconnectReason` and `ReadOutcome` live in `telemetry.rs` with one `as_str` each, and the helpers take them. The tracker uses `RejectReason` directly (`ParseError` is only produced by `titan.rs`). The variant docs carry the meaning each label value had in the constants' docs, and `label_values_are_distinct` guards the strings.

crates/tycho-simulation/src/price_level_stream/tracker.rs
tvinagre: Done in 499e7cc3b: one `SLOT` in `mod.rs`; `QUOTE_TTL = SLOT`, `DEFAULT_STALE_AFTER = 2 * SLOT`, `MAX_FUTURE_SKEW_NANOS = SLOT.as_secs() * NANOS_PER_SECOND`, and `SECONDS_PER_SLOT` is gone.

crates/tycho-simulation/src/price_level_stream/stream.rs
tvinagre: Done in 499e7cc3b: `backoff` moved to `mod.rs` (with its test) and is shared; the `max_backoff` knob and `ConnectionSettings::max_backoff` document that it also caps the whitelist read retry. No second cap: nothing needs the two to differ yet.

crates/tycho-simulation/src/price_level_stream/fallback_router.rs
tvinagre: Done in 499e7cc3b: `WhitelistReaderSettings { read_timeout, max_backoff, refresh_interval }`.

crates/tycho-simulation/src/price_level_stream/stream.rs
tvinagre: Done in 499e7cc3b: the reader is a plain `WhitelistReader` stream (`tokio_stream::pending()` when the router is off), the arm is `Some(venues) = whitelist.next()`, and `Option<WhitelistReader>` and `next_whitelist_read` are gone.

crates/tycho-simulation/src/price_level_stream/stream.rs
tvinagre: Gone in 499e7cc3b: the error path now returns from `build()` before a tracker exists, so only `FreshnessTracker::new` writes the serving state.

crates/tycho-simulation/src/price_level_stream/tracker.rs
tvinagre: Done in 499e7cc3b: `on_stale_deadline(&mut self, now: Instant)`.

crates/tycho-simulation/src/price_level_stream/state.rs
tvinagre: Done in 499e7cc3b: the field is private, `with_quotable_until` is the only setter, and `quotable_until()` is the getter.

crates/tycho-simulation/src/price_level_stream/titan.rs
tvinagre: Confirmed and fixed in 916bf7849: `last_parsed = Instant::now()` now runs after the `yield`, i.e. when the consumer resumes the generator, so the window only measures time spent waiting on the socket. `slow_consumer_does_not_trigger_the_idle_timeout` pauses for three idle windows while frames keep flowing and checks one connection and zero `idle_timeout` reconnects; it fails on the previous order.

crates/tycho-simulation/src/price_level_stream/tracker.rs
tvinagre: Done in 499e7cc3b as asked: `price_level_stream_frame_age_seconds`, a histogram recorded on every accepted frame in `on_frame`, documented in `telemetry.rs` and the module docs. No rejection and no `new_pairs` change. `accepted_frame_records_its_age` checks the value and that rejected frames record nothing.

crates/tycho-simulation/src/price_level_stream/tracker.rs
tvinagre: Follow-up ticket: #1466, covering both the snapshot replacement and the ready marking on removal-only updates, with the suggested fix and test.

crates/tycho-simulation/src/price_level_stream/state.rs
tvinagre: Added `PriceLevelStreamBuilder::without_quote_guard()` in 499e7cc3b: the tracker then emits states with no `quotable_until`. The method doc spells out when to use it (a consumer that quotes more than one slot after arrival by design), that the `stale_after` removal still applies and becomes the only bound on a quoted ladder's age, and that it must never be used on a live router; the module docs and the solver docs page mention it. The integration test does not use it, per your note on BIZ #5. Tests: `quote_guard_off_emits_states_that_never_expire` (tracker) and `without_quote_guard_emits_states_that_never_expire` (stream).

crates/tycho-simulation/src/price_level_stream/fallback_router.rs
tvinagre: Done in 499e7cc3b: `build()` parses the URL once and returns `InvalidFallbackRouterRpcUrl { url, reason }` without opening a reader or a socket; `invalid_node_url_fails_to_build` covers it. The reader keeps its retry loop for transient failures.

crates/tycho-simulation/src/price_level_stream/tracker.rs
tvinagre: Validated at build time in 499e7cc3b: `stale_after` must be non-zero and at most `MAX_STALE_AFTER` (1 h), otherwise `build()` returns `StaleAfterOutOfRange { given }`; documented on the setter. `stale_after_outside_its_range_fails_to_build` covers zero, max plus one second and `Duration::MAX`.

crates/tycho-simulation/src/price_level_stream/tracker.rs
tvinagre: Also for that ticket (the token in this session cannot edit or comment on issues, so noting it here): the emitter's `next_emission_block = block + block_interval` and `take_if(|snapshot| block > snapshot.block_number_or_timestamp)` assume block numbers never decrease. Since this branch the block number only decreases after every served component turned stale, when the next accepted frame is judged as a first frame (how the stream recovers from an implausible block). After such a frame `next_emission_block` can sit far ahead of the chain and the emitter samples nothing for the rest of the process; the fix should re-anchor it when an update's block is below the last seen one. Documented on `build()`, in the module docs and on the solver page.

crates/tycho-simulation/src/price_level_stream/stream.rs
```diff
@@ -194,42 +253,142 @@ impl PriceLevelStreamBuilder {
...
+    /// otherwise.
+    pub fn stale_after(mut self, duration: Duration) -> Self {
+        self.stale_after = duration;
+        self
+    }
+
+    /// Overrides how often the PropAMMRouter whitelist is re-read (default: 10 minutes). A
+    /// venue whose family changes is removed at once and re-added under the new family by the
+    /// next frame carrying it.
+    pub fn whitelist_refresh_interval(mut self, interval: Duration) -> Self {
```
tvinagre, line 280: **ponytail — yagni:** `whitelist_refresh_interval` is a new public knob with no callers or tests in the repository. Keep the 10-minute behavior, delete the builder field/setter, and pass `DEFAULT_WHITELIST_REFRESH_INTERVAL` directly to `WhitelistReaderSettings` (about 10 lines gone).

crates/tycho-simulation/src/price_level_stream/telemetry.rs
```diff
@@ -0,0 +1,383 @@
...
+    OutOfOrder,
+    /// The block is below the newest accepted block.
+    BlockRegression,
+    /// The block is above the newest accepted block by more than one block per elapsed slot
+    /// plus 2.
+    BlockJump,
+}
+
+impl RejectReason {
+    pub(super) fn as_str(self) -> &'static str {
```
tvinagre, line 63: **ponytail — native:** `strum_macros` is already a `tycho-simulation` dependency. Derive `IntoStaticStr` with `#[strum(serialize_all = "snake_case")]` for these three label enums instead of maintaining handwritten mapping tables. With the mapping generated from variant names, `label_values_are_distinct` can go too (roughly 45 lines total).

crates/tycho-simulation/src/price_level_stream/titan.rs
```diff
@@ -129,39 +133,59 @@ pub(super) fn messages(
...
-                        {
-                            Ok(Some(message)) => message,
+                        // The idle timeout counts from the last parsed frame, so control frames
+                        // and unparsable text cannot keep a socket that sends no frames alive.
+                        let remaining = settings
+                            .read_idle_timeout
+                            .saturating_sub(last_parsed.elapsed());
+                        // `None` once the idle timeout has elapsed, either before this read or
+                        // while waiting on it.
+                        let next = if remaining.is_zero() {
```
tvinagre, line 145: **ponytail — native:** This manually implements an idle deadline with `std::Instant`, `elapsed`, `saturating_sub`, and nested `Option`s. Hold a Tokio deadline and call `timeout_at(deadline, ws_stream.next())`, then reset the deadline after the `yield`; that preserves the slow-consumer behavior with a materially shorter loop.

crates/tycho-simulation/src/price_level_stream/test_support.rs
```diff
@@ -0,0 +1,173 @@
...
+/// socket closes.
+pub(super) fn frame_then_repeat(
+    first: Message,
+    repeat: Message,
+    interval: Duration,
+) -> impl Fn(usize, FakeConnection) -> BoxFuture<'static, ()> + Send + Sync + 'static {
+    move |_, mut socket| {
+        let first = first.clone();
+        let repeat = repeat.clone();
+        Box::pin(async move {
```
tvinagre, line 147: **ponytail — shrink:** This helper and `stream.rs::fresh_frame_every` only send until the socket errors. Both nested loops can be `while socket.send(...).await.is_ok() { tokio::time::sleep(interval).await; }` (about 15 lines gone, same behavior).

crates/tycho-simulation/src/price_level_stream/test_support.rs
```diff
@@ -0,0 +1,173 @@
...
+
+use super::config::PriceLevelStreamConfig;
+
+/// The FermiSwapper router, one of the default venues.
+pub(super) const PAMM: &str = "0x5979458912f80b96d30d4220af8e2e4925a33320";
+pub(super) const WBTC: &str = "0x2260fac5e5542a773aa44fbcfedf7c193bc2c599";
+pub(super) const USDC: &str = "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48";
+pub(super) const WETH: &str = "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2";
+
+pub(super) fn token(address: &str, symbol: &str, decimals: u32) -> Token {
```
tvinagre, line 35: **ponytail — shrink:** This shared `token` helper and the `WBTC`/`USDC`/`WETH` constants exactly duplicate the three constructors in `state.rs:L336-370`. Reuse these there and delete the duplicated constructors/imports (about 24 lines).

crates/tycho-simulation/src/price_level_stream/titan.rs
```diff
@@ -333,14 +381,146 @@ mod tests {
         }
     }
 
+    /// A frame captured verbatim from the live stream (2026-09-05). It carries `timestamp`,
+    /// `slot`, and an undocumented per-venue `maker` field the parser must ignore.
+    const CAPTURED_MESSAGE_2026_09_05: &str =
```
tvinagre, line 386: **ponytail — delete:** The existing captured-live-frame test already covers a real frame with `timestamp`; only unknown `maker`-field tolerance is new here. Add `maker` to the small documented sample and delete this second test plus `pamm_price_levels_1788624558231060482.json` (about 20 lines and 47,114 bytes).

tvinagre (commented):
Ponytail review: complexity-only suggestions that preserve the branch's stated behavior and safety checks. Estimated net reduction: about 151 lines, plus one 47,114-byte duplicate fixture.

crates/tycho-simulation/src/price_level_stream/stream.rs
```diff
@@ -898,134 +605,528 @@ mod tests {
...
+        assert_removal_only(&removal, 1);
+        // Every later replay is too old to be accepted: nothing comes back.
+        assert!(next_within(&mut stream, Duration::from_millis(500))
             .await
-            .is_empty());
+            .is_none());
+    }
+
+    #[tokio::test]
+    async fn ping_only_traffic_removes_within_stale_after() {
```
tvinagre, line 857: **ponytail — shrink:** `ping_only_traffic_removes_within_stale_after` and `malformed_text_removes_within_stale_after` duplicate all setup and assertions; only the repeated `Message` differs. Loop over the two messages in one async test and preserve both cases (about 14 lines gone).

crates/tycho-simulation/src/price_level_stream/titan.rs
```diff
@@ -333,14 +381,146 @@ mod tests {
...
+    /// Polls `stream` until it yields a frame, failing if none arrives within two seconds.
+    async fn next_frame(stream: &mut (impl Stream<Item = TitanPriceLevelMessage> + Unpin)) {
+        tokio::time::timeout(Duration::from_secs(2), stream.next())
+            .await
+            .expect("a frame within two seconds")
+            .expect("the stream never ends");
+    }
+
+    #[test]
+    fn ping_only_traffic_does_not_count_as_liveness() {
```
tvinagre, line 429: **ponytail — shrink:** This and `malformed_text_does_not_count_as_liveness` duplicate the fake server, stream, reconnect, and metric setup. Parameterize one test over `(message, expected_parse_errors)` (about 12 lines gone).

crates/tycho-simulation/src/price_level_stream/mod.rs
```diff
@@ -36,8 +99,43 @@
...
+mod test_support;
 mod titan;
+mod tracker;
+
+/// The post-merge Ethereum slot. Every freshness window of the stream is a multiple of it.
+const SLOT: Duration = Duration::from_secs(12);
+
+/// The delay before the retry after `attempt` consecutive failures: `2^attempt` seconds, capped
+/// at `max_backoff`. Shared by the Titan reconnect and the whitelist read retry.
+fn backoff(attempt: u32, max_backoff: Duration) -> Duration {
```
tvinagre, line 119: **ponytail — stdlib:** Collapse the temporary plus `map(...).unwrap_or(...)` into `2u64.checked_pow(attempt).map_or(Duration::MAX, Duration::from_secs).min(max_backoff)` (3 lines gone, same overflow behavior).

PR #1429, opened 2026-09-09
fix(token-analyzer): compute transfer fee without intermediate underflow

brunoguerios:
## Problem

The nightly `analyze-tokens` job (02:00 UTC) logs this error:

```
WARN Token quality detection failed error="Failed to calculate transfer fee: overflow"
```

In the 2026-09-09 prod run, every failed token hit this line: Base 13 of 13, BSC 257 of 257. The
36 h before that had 1,580 hits across chains.

When the analyzer returns an error, the job skips the token and leaves its quality unchanged. The
token stays in the 6–10 retry window and is retried every night, forever.

## Root cause

`calculate_fee` subtracts the receiver's new balance from the amount sent, then adds the old
balance. The subtraction underflows when the receiver already holds more than the fee. Three token
types hit this:

- a fee-on-transfer token, when settlement already holds dust
- a rounding token that credits 1 wei less than sent
- a bonus token that credits more than sent

Two smaller defects made it worse. The arithmetic helpers reported `"overflow"` for every failure,
so the message misled. And `handle_response` returned the error as `Err`, so the job read a token
defect as an RPC fault.

## Fix

`calculate_fee` now adds first and subtracts second: `(before + sent) - after`. That cannot
underflow. A receiver that gets at least `sent` pays no fee. A zero-amount transfer pays no fee.
Inputs that worked before give the same result. The four `error_*` helpers are gone.

`handle_response` now returns `Bad` instead of `Err`. The warn line now names the token address.

`TraceCallDetector` calls the same function and gets the fix.

## Notes for reviewers

Fee tokens now get a tax value and quality 50. Rounding and bonus tokens report 0 bps and stay
`Bad`, so they fall 1 per night to the floor at 5. The analyzer already treats every token with an
inexact balance move that way.

Check after the first 02:00 UTC run post-deploy: `failed=0` on the Retry pass line for
`base-analyze-tokens` and `bsc-analyze-tokens`.

## Follow-ups

- Two `checked_add` guards after the fee call still return hard errors. `calculate_fee` computes
  the same sums first, so they are unreachable. Left alone.
- BSC demotes ~65,900 tokens per night. That is the revert-cohort drain from 63334ed3f, not this
  bug.

### DIFF
--- a/crates/tycho-ethereum/src/services/token_analyzer/common.rs
+++ b/crates/tycho-ethereum/src/services/token_analyzer/common.rs
@@ -1,3 +1,5 @@
+use std::cmp;
+
 use alloy::{
     primitives::{keccak256, Address, U256},
     rpc::types::{BlockNumberOrTag, TransactionInput, TransactionRequest},
@@ -14,8 +16,9 @@ pub(crate) fn arbitrary_recipient() -> Address {
 
 /// Computes the transfer fee in basis points (0–10_000) from observed balance deltas.
 ///
-/// Returns the higher of the inbound and outbound fee rates. Returns zero if neither transfer
-/// shows a fee. Errors only on arithmetic overflow, which is not expected in practice.
+/// Returns the higher of the inbound and outbound fee rates. A transfer that credits the
+/// receiver with at least the amount sent has no fee. Errors only if a balance plus the amount
+/// sent overflows U256.
 pub(crate) fn calculate_fee(
     amount: U256,
     middle_amount: U256,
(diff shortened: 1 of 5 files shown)


crates/tycho-ethereum/src/services/token_analyzer/common.rs
```diff
@@ -12,67 +12,51 @@ pub(crate) fn arbitrary_recipient() -> Address {
...
+    /// `shortfall * 10_000` overflows U256.
+    fn fee_bps(&self) -> Result<U256, String> {
+        let expected_after = self
+            .balance_before
+            .checked_add(self.sent)
+            .ok_or_else(|| format!("balance {} + {} overflows", self.balance_before, self.sent))?;
+        if self.sent.is_zero() || self.balance_after >= expected_after {
+            return Ok(U256::ZERO);
+        }
+        let shortfall = expected_after - self.balance_after;
```
zizou0x, line 39: Is it safe that we do it unchecked here?

crates/tycho-ethereum/src/services/token_analyzer/ethcall_detector.rs
```diff
@@ -245,10 +257,7 @@ impl EthCallDetector {
             ));
         }
 
-        let computed_recipient_after = r
-            .recipientBefore
-            .checked_add(middle_amount)
-            .ok_or("recipient balance overflow when checking transfer out")?;
+        let computed_recipient_after = r.recipientBefore + middle_amount;
```
zizou0x, line 261: Same question, is it safe to remove the check here?

crates/tycho-ethereum/src/services/token_analyzer/ethcall_detector.rs
```diff
@@ -206,20 +206,32 @@ impl EthCallDetector {
...
+                return Ok((
+                    TokenQuality::bad(format!("Failed to calculate transfer fee: {e}")),
+                    Some(gas_per_transfer),
+                    None,
+                ))
+            }
+        };
+
+        // calculate_fee_bps already rejected both sums if they overflow.
+        let computed_balance_after_in = r.balanceBeforeIn + amount;
```
zizou0x, line 234: Same question here

crates/tycho-ethereum/src/services/token_analyzer/common.rs
brunoguerios: Yes. The `if` two lines up returns unless `balance_after < expected_after`, so this subtraction cannot underflow. Added a comment in 5147287f4.

crates/tycho-ethereum/src/services/token_analyzer/ethcall_detector.rs
brunoguerios: Yes. `calculate_fee_bps` runs `checked_add` on this exact pair (`balanceBeforeIn + amount`, the inbound transfer) and the `match` above returns `Bad` on overflow. When this line runs, the same sum has already succeeded. Reworded the comment in 5147287f4.

crates/tycho-ethereum/src/services/token_analyzer/ethcall_detector.rs
brunoguerios: Yes, same reasoning for the outbound transfer. `recipientBefore + middle_amount` is the outbound `checked_add` in `calculate_fee_bps`, and it only returns `Ok` when both sums pass. Added a comment in 5147287f4.

zizou0x (approved):
LGTM

PR #1433, opened 2026-09-10
fix: set fee calculator receiver at deployment time

louise-poole:
`FeeCalculator` takes the router fee receiver as a constructor argument instead of
  defaulting to `msg.sender`, and starts with positive slippage capture enabled.

  ## Why

  Deployment goes through the CREATE2 factory `0x4e59b448…`, so `msg.sender` in the
  constructor was the factory, and it became the router fee receiver until someone called
  `setRouterFeeReceiver`. The factory only performs `create2` and can never call
  `withdraw`, so every fee credited to it in that window is stuck. This happened on
  ethereum, base, arbitrum, bsc, polygon and robinhood; the largest amount is 386.75 USDG
  + 0.003 ETH on robinhood, where the receiver was corrected six days after deployment.

  All eight deployed calculators have positive slippage capture switched on, so a fresh
  deployment now matches that rather than passing the surplus on until it is toggled.

### DIFF
--- a/crates/tycho-execution/CLAUDE.md
+++ b/crates/tycho-execution/CLAUDE.md
@@ -100,6 +100,11 @@ Three fee layers, deducted from swap output:
 3. **Router fee on client fee** (stored): `_routerFeeOnClientFeeBps` -- Tycho's cut of the client fee (deducted from the
    client's portion, not from the user).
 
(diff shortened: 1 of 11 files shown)


tamaralipows (approved):
Thank you!

PR #1434, opened 2026-09-10
feat(execution): encode fallback swaps for the TychoFallbackRouter

Troshchk:


crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
```diff
@@ -0,0 +1,410 @@
...
+                data.extend_from_slice(bytes_to_address(hook)?.as_slice());
+                data.extend_from_slice(hook_data.as_ref());
+            }
+            FallbackProtocol::Curve { pool, pool_type, i, j } => {
+                data.extend_from_slice(bytes_to_address(pool)?.as_slice());
+                data.extend_from_slice(&[*pool_type, *i, *j]);
+            }
+            FallbackProtocol::FluidV1 { dex, zero2one } => {
+                data.extend_from_slice(bytes_to_address(dex)?.as_slice());
+                data.push(u8::from(*zero2one));
```
tamaralipows, line 134: I was wondering if we can't just reuse the existing encoders for these encodings?

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
```diff
@@ -0,0 +1,410 @@
...
+            FallbackProtocol::UniswapV4 { fee, tick_spacing, hook, hook_data } => {
+                if *fee >= 1 << 24 {
+                    return Err(EncodingError::InvalidInput(format!(
+                        "Uniswap V4 fallback fee {fee} does not fit uint24"
+                    )));
+                }
+                if *tick_spacing < -(1 << 23) || *tick_spacing >= 1 << 23 {
+                    return Err(EncodingError::InvalidInput(format!(
+                        "Uniswap V4 fallback tick spacing {tick_spacing} does not fit int24"
+                    )));
```
tamaralipows, line 121: Also one thing to double check: I think this is invalid for angstrom? Or do we want to support an angstrom hook as fallback? I would think it's perhaps better to revert if it's an angstrom hook, fyi @louise-poole

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
```diff
@@ -0,0 +1,410 @@
...
+        match self {
+            FallbackProtocol::UniswapV2 { .. } => 0,
+            FallbackProtocol::UniswapV3 { .. } => 1,
+            FallbackProtocol::UniswapV4 { .. } => 2,
+            FallbackProtocol::Curve { .. } => 3,
+            FallbackProtocol::FluidV1 { .. } => 4,
+        }
+    }
+
+    /// Packs the venue byte of `TychoFallbackRouter.Venue` followed by the venue data its
```
tamaralipows: Pls update with latest main, it's no longer called Venue :)

crates/tycho-execution/src/encoding/evm/constants.rs
```diff
@@ -141,12 +141,26 @@ pub const PRICE_LEVEL_STREAM_KEY: &str = "pricelevelstream";
...
+/// Protocol system prefix for pAMM components executed through `TychoFallbackRouter`, which
+/// retries a failing pAMM on the fallback venue named in the swap's `user_data`. Venue suffixes
+/// follow `PRICE_LEVEL_STREAM_PREFIX`. Replaces `PROPAMM_FALLBACK_PREFIX` (Titan's PropAMMRouter,
+/// deprecated): any pAMM qualifies, and the solver picks the fallback venue per swap instead of
+/// the router owning one Uniswap V3 mapping.
+pub const FALLBACK_PREFIX: &str = "fallback:";
+
+/// The executor-config key serving the whole fallback protocol family, mirroring
+/// `PRICE_LEVEL_STREAM_KEY`.
+pub const FALLBACK_KEY: &str = "fallback";
```
tamaralipows, line 162: Do we need two different consts for this

crates/tycho-execution/src/encoding/evm/gas_estimator.rs
```diff
@@ -88,6 +90,10 @@ pub fn needs_approval(protocol_system: &str) -> bool {
 /// The estimate prices the venue path because that is the path a fresh quote takes.
 pub const PROPAMM_FALLBACK_OVERHEAD_GAS: u64 = 70_000;
 
+/// This prices the pAMM path because that is the path a fresh
+/// quote takes. A swap that actually falls back pays for the fallback venue on top.
+pub const FALLBACK_ROUTER_OVERHEAD_GAS: u64 = 70_000;
```
tamaralipows: It's not really clear from the docstring how this is different from `PROPAMM_FALLBACK_OVERHEAD_GAS`?

tamaralipows (commented):
Thanks! Just a few questions

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
Troshchk: Hm, this is what I was a bit confused yesterday, coz I still saw some venues on you branch
Will rename 👍

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
Troshchk: Not really, the existing encoders read the static attributes of the protocol components, where here the protocol component is pamm and the info related to the fallback protocol is coming from the user_data

crates/tycho-execution/src/encoding/evm/constants.rs
Troshchk: These are different string and it follows structure of the propammfallback and  pricelevelstream
I think it is ok to leave them, lmk if you thinkdifferently

crates/tycho-execution/src/encoding/evm/gas_estimator.rs
Troshchk: Added number from a real measurement

crates/tycho-execution/src/encoding/evm/constants.rs
tamaralipows: can't we just append the : to the `FALLBACK_KEY` in-code?

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
```diff
@@ -0,0 +1,517 @@
...
+            })
+    }
+
+    /// Rejects a Uniswap V4 fallback whose hook is the chain's Angstrom hook.
+    fn reject_angstrom_hook(&self, protocol: &FallbackProtocol) -> Result<(), EncodingError> {
+        if let FallbackProtocol::UniswapV4 { hook, .. } = protocol {
+            if hook == &self.angstrom_hook_address {
+                return Err(EncodingError::InvalidInput(
+                    "Angstrom pools are unsupported as a fallback protocol: they are locked every \
+                     block and need a live unlock attestation the fallback path cannot supply"
```
tamaralipows, line 183: I think we can just leave it as "Angstrom pools are unsupported as a fallback protocol"

crates/tycho-execution/config/protocol_specific_addresses.json
```diff
@@ -3,6 +3,9 @@
     "uniswap_v4": {
       "angstrom_hook_address": "0x0000000aa232009084Bd71A5797d089AA4Edfad4"
     },
+    "fallback": {
+      "angstrom_hook_address": "0x0000000aa232009084Bd71A5797d089AA4Edfad4"
+    },
```
tamaralipows, line 8: did we need a new attribute for that? It wasn't possible just to get from uniswap_v4? I guess that would disrupt our flow too much right...

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
```diff
@@ -0,0 +1,517 @@
...
+
+    /// The ordinal of the matching `TychoFallbackRouter.FallbackProtocol` variant — the wire
+    /// format's protocol byte.
+    fn protocol_byte(&self) -> u8 {
+        match self {
+            FallbackProtocol::UniswapV2 { .. } => 0,
+            FallbackProtocol::UniswapV3 { .. } => 1,
+            FallbackProtocol::UniswapV4 { .. } => 2,
+            FallbackProtocol::Curve { .. } => 3,
+            FallbackProtocol::FluidV1 { .. } => 4,
```
tamaralipows, line 92: Something that I haven't double checked on the contract side: Is it possible to check here already to make sure that the protocol we are choosing actually exists in the chain we are on? For example if we are on Base, we shouldn't be using Curve or FluidV1, right? This info is in `executor_addresses.json`

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
```diff
@@ -0,0 +1,517 @@
...
+}
+
+impl FallbackProtocol {
+    fn from_swap_user_data(user_data: &Option<Bytes>) -> Result<Self, EncodingError> {
+        match user_data.as_ref() {
+            Some(bytes) if !bytes.is_empty() => serde_json::from_slice(bytes).map_err(|e| {
+                EncodingError::InvalidInput(format!(
+                    "Invalid fallback protocol user_data JSON: {e}"
+                ))
+            }),
```
tamaralipows, line 75: Another question regarding this. Will these protocols also catch the forks? For example if someone chooses sushiswap, we need to make sure it sends it to the UniswapV2 executor and doesn't reject the choice

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
```diff
@@ -0,0 +1,517 @@
...
+        Ok(data)
+    }
+}
+
+/// Encodes a swap that runs a pAMM through `TychoFallbackRouter` so a failing pAMM retries on
+/// the fallback protocol named in the swap's `user_data` instead of reverting the route.
+///
+/// The pAMM address comes from the component's `pamm_address` static attribute, which every
+/// fallback component carries. Swap data for `FallbackExecutor` is packed
+/// `token_in ++ token_out ++ pamm ++ protocol_byte ++ protocol_data` (see [`FallbackProtocol`]).
```
tamaralipows, line 148: No need to illustrate in the docstring how the calldata is packed

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
```diff
@@ -0,0 +1,517 @@
...
+/// The variants mirror `TychoFallbackRouter.FallbackProtocol`; the wire format is the protocol
+/// byte followed by the protocol data the contract decodes:
+///
+/// | Protocol     | JSON                                                       | Protocol data |
+/// |--------------|------------------------------------------------------------|---------------|
+/// | `uniswap_v2` | `{"protocol":"uniswap_v2","pair":"0x…","fee_bps":30}`         | `pair(20) ++ fee_bps(1)` |
+/// | `uniswap_v3` | `{"protocol":"uniswap_v3","pool":"0x…"}`                      | `pool(20)` |
+/// | `uniswap_v4` | `{"protocol":"uniswap_v4","fee":3000,"tick_spacing":60,"hook":"0x…","hook_data":"0x…"}` | `fee(3) ++ tick_spacing(3) ++ hook(20) ++ hook_data` |
+/// | `curve`      | `{"protocol":"curve","pool":"0x…","pool_type":1,"i":0,"j":1}` | `pool(20) ++ pool_type(1) ++ i(1) ++ j(1)` |
+/// | `fluid_v1`   | `{"protocol":"fluid_v1","dex":"0x…","zero2one":true}`         | `dex(20) ++ zero2one(1)` |
```
tamaralipows, line 34: I think it's fine to keep the json if you want but the protocol data is unnecessary. In general this table is kindof hard to read, do we need it?

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
```diff
@@ -0,0 +1,517 @@
...
+///
+/// Swap direction on Uniswap V2/V3/V4 comes from the sort order of the swap's tokens, so it is
+/// not part of the JSON. Fluid's `zero2one` is the dex's own token order, which cannot be
```
tamaralipows, line 37: I don't think it's necessary to go into so much detail re: swap direction and why certain fields are included and others not

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
```diff
@@ -0,0 +1,517 @@
...
+/// derived, and Curve's `i`/`j` are the pool's coin indices.
+#[derive(Clone, Debug, Deserialize)]
+#[serde(tag = "protocol", rename_all = "snake_case")]
```
tamaralipows, line 40: should we name the tag `fallback_protocol` for clarity instead of just `protocol`?

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
```diff
@@ -0,0 +1,517 @@
...
+    fn new(
+        executor_address: Bytes,
+        chain: Chain,
+        config: Option<HashMap<String, String>>,
+    ) -> Result<Self, EncodingError> {
+        if chain != Chain::Ethereum {
+            return Err(EncodingError::FatalError(
+                "Fallback swaps are only supported on Ethereum".to_string(),
+            ));
+        }
```
tamaralipows, line 202: Learned today in the call that this is false and that we will also soon be on Base I think

tamaralipows (commented):
Thank you! My biggest concern now is that forks are properly mapped to the proper executor - please double check this point in particular

crates/tycho-execution/config/protocol_specific_addresses.json
Troshchk: Yes, exactly
It is related to this part:
```  
protocol_specific_config
      .get(protocol)  
      .cloned()
```
Every protocol only sees specific config for itself, but not for other protocols. Changing that would touch the shared `create_encoder()`, so I decided to not do it this way.

I have also added test `test_fallback_angstrom_hook_matches_uniswap_v4` to make sure these two values do not diverge

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
Troshchk: Will be addressed in a separate PR

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
Troshchk: I think we can add it as soon as we go multichain, maybe we can add it to `protocol_specific_addresses.json`? I think it also gives more granular control -- if protocol on some chain exists, but we don't want to allow using it on that chain as a fallback we will be able to do it. We can enforce it inside `encode_swap`.
Alternatively, we can do what you suggest, but I am not sure how do we handle hooks in this case? And we will also need to solve the naming discrepancy, like `curve` and `vm:curve`
In any case, I think we should do it when we go multichain
What do you think?

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
Troshchk: Good catch, fixed

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
Troshchk: Removed

crates/tycho-execution/contracts/src/executors/PropAMMFallbackExecutor.sol
```diff
@@ -10,8 +10,10 @@ error PropAMMFallbackExecutor__InvalidDataLength();
 /// @title PropAMMFallbackExecutor
 /// @notice Swaps against a pAMM through Titan's PropAMMRouter, which falls back to a single-hop
 /// Uniswap V3 pool when the venue reverts.
-/// @custom:deprecated Use `FallbackExecutor`, which picks the fallback venue itself, from five
-/// kinds, for any pAMM. Encode no new routes against this one; it is removed in a follow-up PR.
+/// @custom:deprecated Use `FallbackExecutor`: it works for any pAMM and lets the caller name the
+/// fallback protocol (Uniswap V2/V3/V4, Curve, or Fluid V1) per swap, instead of the router
+/// owning one hardcoded Uniswap V3 mapping. Encode no new routes against this one; it is removed
+/// in a follow-up PR.
```
louise-poole: What is meant by "Encode no new routes against this one; it is removed in a follow-up PR."?

crates/tycho-execution/contracts/test/protocols/Fallback.t.sol
```diff
@@ -115,10 +115,14 @@ contract EmptyReservePair {
...
-        int256, /* amountSpecified */
-        uint160, /* sqrtPriceLimitX96 */
+        address,
+        /* recipient */
+        bool,
+        /* zeroForOne */
+        int256,
+        /* amountSpecified */
+        uint160,
+        /* sqrtPriceLimitX96 */
```
louise-poole: Small formatting q: why put the comment on new lines? I found the same-line comments easier to read here.

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
```diff
@@ -0,0 +1,557 @@
...
+        Ok(data)
+    }
+}
+
+/// Encodes a swap that runs a pAMM through `TychoFallbackRouter` so a failing pAMM retries on
+/// the fallback protocol named in the swap's `user_data` instead of reverting the route.
+///
+/// The pAMM address comes from the component's `pamm_address` static attribute, which every
+/// fallback component carries.
```
louise-poole: This makes it sound like the fallback component (like the Uniswap v2 component we fall back to) carries `pamm_address` as a static attribute. This should probably be reworded.

louise-poole: Or did I misunderstand something here?

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
```diff
@@ -0,0 +1,557 @@
...
+impl FallbackSwapEncoder {
+    fn pamm_address(swap: &Swap) -> Result<Bytes, EncodingError> {
+        let component = swap.component();
+        component
+            .static_attributes
+            .get(PAMM_ADDRESS_ATTRIBUTE)
+            .cloned()
+            .ok_or_else(|| {
+                EncodingError::FatalError(format!(
+                    "Fallback component {} is missing the {PAMM_ADDRESS_ATTRIBUTE} static \
```
louise-poole: (related to my doc comment above)

crates/tycho-execution/src/encoding/evm/constants.rs
```diff
@@ -141,12 +151,26 @@ pub const PRICE_LEVEL_STREAM_KEY: &str = "pricelevelstream";
...
 /// `PRICE_LEVEL_STREAM_PREFIX`; only whitelisted venues may use it. Calldata matches the direct
 /// path, so both prefixes share `PropAMMSwapEncoder` and differ only in the executor.
+///
+/// Deprecated with `PropAMMFallbackExecutor`: encode no new routes against this family, use
+/// [`FALLBACK_PREFIX`]. Removed in a follow-up PR together with the executor.
 pub const PROPAMM_FALLBACK_PREFIX: &str = "propammfallback:";
 
 /// The executor-config key serving the whole PropAMMRouter protocol family, mirroring
 /// `PRICE_LEVEL_STREAM_KEY`.
 pub const PROPAMM_FALLBACK_KEY: &str = "propammfallback";
```
louise-poole, line 161: Can we also comment that this is deprecated too?

louise-poole (commented):
Looking good! My main concern is getting the error message clear/correct. Otherwise I'm happy with everything else!

crates/tycho-execution/contracts/src/executors/PropAMMFallbackExecutor.sol
Troshchk: I have removed it -- I added info about removing initially, but it is not needed

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
Troshchk: You are right, I have reworded it here and in the error

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
Troshchk: Reworded

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
```diff
@@ -137,8 +137,9 @@ impl FallbackProtocol {
 /// Encodes a swap that runs a pAMM through `TychoFallbackRouter` so a failing pAMM retries on
 /// the fallback protocol named in the swap's `user_data` instead of reverting the route.
 ///
-/// The pAMM address comes from the component's `pamm_address` static attribute, which every
-/// fallback component carries.
+/// The pAMM address comes from the `pamm_address` static attribute of the swapped component — the
+/// pAMM component itself (`protocol_system` = `fallback:{protocol}`), not the fallback protocol's
+/// pool.
```
louise-poole: I don't think we need to clarify 'not the fallback protocol's pool'. The current wording feels much clearer to start, but then bringing in mention of the fallback pool adds confusion again. This is a simple helper method so I think we can keep the doc string simple:
```suggestion
/// The pAMM address comes from the `pamm_address` static attribute of the component.
```

louise-poole (approved):
One tiny comment, otherwise approved!

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
```diff
@@ -36,13 +36,26 @@ const MAX_UNISWAP_V2_FEE_BPS: u8 = 30;
...
+/// unchanged; the fork name only widens what the encoder accepts. Keep these aliases in sync with
+/// those match arms.
+///
+/// The Slipstream family (`aerodrome_slipstreams`, `velodrome_slipstreams`, `up_v3`, `ramses_v3`)
+/// would run the same V3 fallback path — the wire format is just the pool address — but those
+/// venues are Base/L2-only, so their aliases are deferred until Base is supported.
 #[derive(Clone, Debug, Deserialize)]
 #[serde(tag = "protocol", rename_all = "snake_case")]
 enum FallbackProtocol {
+    #[serde(alias = "sushiswap_v2", alias = "pancakeswap_v2", alias = "quickswap_v2")]
```
tamaralipows: pancakeswap_v2 actually uses a different executor because its a different fee. Please see the `executor_addresses.json`:
```
  "ethereum": {
    "uniswap_v2": "0x0017c84f2B3414514B67Bfc9a63830c8E0E690d0",
    "sushiswap_v2": "0x0017c84f2B3414514B67Bfc9a63830c8E0E690d0",
    "pancakeswap_v2": "0xFb89Ca30E5Ff1863237dD24aAF2189bD1cA6CC0D",
```

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
```diff
@@ -36,13 +36,26 @@ const MAX_UNISWAP_V2_FEE_BPS: u8 = 30;
...
     UniswapV2 {
         pair: Bytes,
         fee_bps: u8,
     },
+    #[serde(alias = "pancakeswap_v3", alias = "sushiswap_v3", alias = "robinswap_v3")]
```
tamaralipows: What about `ramses_v3` on polygon and robinhood, and `up_v3` on robinhood? Since I see you already covering other chains here. Please see the `executor_addresses.json`

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
tamaralipows: We are already going multichain, no? Since we would like to deply on base. Or do you mean to add this in the slipstreams PR? I'm not sure I understood the naming discrepancy - from `vm:curve` to `curve` would already be a problem now no? And I am sure we are doing this stripping in several places already. And not sure how `protocol_specific_addresses.json` would help - what do you mean by that?

tamaralipows (approved):
Thanks. I'm a bit confused about your comment but if you prefer to do in a separate PR that's fine. Pls double check if im right about the missing v3 fork aliases

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
Troshchk: Right now we have this gate:
`if chain != Chain::Ethereum {`
So I think we can add check whether the protocol exists on chain when it is actually possible to fallback on other chains. 

By `protocol_specific_addresses.json` I meant to add a per chain allowlist of fallback protocols in the fallback section.
My worry with `executor_addresses.json` is that the fallback doesn't reuse executors -- TychoFallbackRouter implements them inline and we can choose a protocol to be fallback, when no executor exists for it(having executor is not necessary for protocol to be a fallback). This is why I suggest to control which protocol can be fallback via `protocol_specific_addresses.json` and to not rely on the `executor_addresses.json`. Lmk if my assumption is incorrect here

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
Troshchk: Since this PR is ethereum only, I removed quickswap_v2/sushiswap_v3/robinswap_v3, I think they should be coming together with the support of the relevant chains

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
louise-poole: Agreed! We'd need to update the contract anyway to start supporting new chains so can update this encoder at the same time too when that happens.

louise-poole: I think we can have hardcoded fallback protocol lists per chain. And actually should expose it publicly so that Fynd can import it and know what protocols it can fallback to with this version of the encoder.

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
louise-poole: Agreed 👍

PR #1438, opened 2026-09-12
feat: integrate BaiBai CurveBook v3 on Base

mteam88:
Add WETH/USDC indexing, native exact-input simulation and swap execution for BaiBai CurveBook v3 on Base. The indexer tracks curve storage, fill cursors, expiry, custody holdings, claim reservations and taker fee overrides. Simulation preserves integer rounding, consumes gross curve liquidity and debits net payouts from custody. Quotes require no RPC calls.

Fees follow the deployed contract: the executing taker's pair override, then their taker-wide override, then zero; explicit zero overrides are preserved. `DecoderContext::caller` selects a custom execution caller and defaults to the Tycho Base router. The executor uses normal router slippage protection. Bid limits stop before decreasing proceeds and account for available custody.

All three dependencies are upgradeable proxies. Upgrades after the validated block 51,191,196 pause the component and native simulation until the integration is reviewed, updated and reindexed. Historical upgrades through that block are covered by replay. The runtime-fixture script/CI changes and shared retention fix are excluded, per review; historical testing uses the [maintainer-owned retention fix](https://github.com/propeller-heads/tycho/pull/1438#discussion_r4013660533) as a local validation patch.

Validation:

- Fresh authenticated replay of Base blocks 50,895,895–51,191,196 with the retention patch applied only in the isolated validation checkout: full harness passed; all six fork executions exactly matched simulation in both directions.
- Independent comparison matched all 32 indexed storage words and both custody balances. Complete fee-event history matched the indexed overrides and the `fees_indexed` snapshot marker. This pinned history has zero overrides; nonzero fees are covered by native and fork tests.
- 1,181 tests passed with the Rust CI unit command, plus eleven BaiBai native tests, seven Substreams tests, the Rust calldata generator and six Base fork executor tests. WASM build, workspace/Substreams Clippy, Rustdoc, no-default-features and changed-file formatting passed.
- Previous independent deployed-bytecode comparison: 108 quote fixtures match.
- Slither analyzed 187 contracts with 101 detectors: no BaiBai findings; one `reentrancy-balance` finding in `TychoFallbackRouter.executePropAMM`, whose source is unchanged from upstream main.

Includes protocol registration, executor deployment configuration and a runtime fixture. A production executor deployment and indexing registration are still required; no production executor address is assigned here.


protocols/substreams/base-baibai/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+                continue;
+            };
+            balance_deltas.push(BalanceDelta {
+                ord: log.ordinal,
+                tx: Some(view.receipt.transaction.into()),
+                token: log.address.clone(),
+                delta: value.to_signed_bytes_be(),
+                component_id: config.id().into_bytes(),
+            });
+        }
```
zach030, line 83: For WETH, you'd also need to handle the `Deposit` and `Withdrawal` events.

crates/tycho-execution/contracts/src/executors/BaibaiExecutor.sol
```diff
@@ -0,0 +1,96 @@
...
+    /// @inheritdoc IExecutor
+    function fundsExpectedAddress(bytes calldata)
+        external
+        view
+        returns (address)
+    {
+        return msg.sender;
+    }
+
+    /// @inheritdoc IExecutor
```
zach030, line 47: Please fix the Slither errors in the executor contract.

protocols/testing/scripts/update_runtime_bytecode.sh
```diff
@@ -97,6 +98,11 @@ EXECUTOR_FIXTURES=(
     "Sky|ethereum|sky"
 )
 
+if [[ -z "${BASE_RPC_URL:-}" ]]; then
+    echo "Error: BASE_RPC_URL must be set for the BaiBai fixture." >&2
+    exit 1
+fi
+
```
zach030, line 105: `BASE_RPC_URL` should be optional here.

protocols/substreams/base-baibai/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+        .calls
+        .iter()
+        .filter(|call| !call.state_reverted)
+    {
+        for write in &call.storage_changes {
+            if let Some(index) = slots
+                .iter()
+                .position(|(address, slot)| {
+                    write.address == address.as_slice() && write.key == slot.as_slice()
+                })
```
zach030, line 173: Would be more efficient to filter by address here.
```
if write.address != curve_book && write.address != custodian {
    continue;
}
```

crates/tycho-execution/contracts/src/executors/BaibaiExecutor.sol
```diff
@@ -0,0 +1,96 @@
...
+        uint256 minAmountOut,
+        address to
+    ) external returns (uint256);
+}
+
+/// @notice Executes exact-input swaps through one BaiBai entrypoint.
+/// @dev Data is base (20 bytes) followed by sellBase (one byte, 0 or 1).
+contract BaibaiExecutor is IExecutor {
+    error InvalidData();
+    error NonzeroTakerFee();
```
zach030, line 26: Can you update these errors to match the repo convention, i.e. the `XxxExecutor__Yyy` format?

crates/tycho-execution/contracts/src/executors/BaibaiExecutor.sol
```diff
@@ -0,0 +1,96 @@
...
+    function swap(uint256 amountIn, bytes calldata data, address receiver)
+        external
+        payable
+    {
+        (address base, address tokenIn,) = _decode(data);
+        // The native simulator quotes the zero-fee path. Under delegatecall,
+        // address(this) is the router, which is also the entrypoint's fee identity.
+        if (entrypoint.takerFeeBps(address(this), base) != 0) {
+            revert NonzeroTakerFee();
+        }
```
zach030, line 57: Is the zero fee here the default, or do you need to whitelist the router for it? 
IMO this is a bit strict, if the fee logic in the entrypoint changes or we deploy a new router, it'd keep reverting.

crates/tycho-execution/contracts/test/protocols/Baibai.t.sol
```diff
@@ -0,0 +1,215 @@
...
+
+    function setUp() public override {
+        super.setUp();
+        fixture = vm.readFile(
+            "../../tycho-simulation/tests/assets/baibai-quotes.json"
+        );
+        _curve(0);
+        deal(BASE, CUSTODIAN, 1e24);
+        deal(QUOTE, CUSTODIAN, 1e24);
+    }
```
zach030, line 42: Thanks for the thorough tests! One thing though: in our design, `tycho-execution` and `tycho-simulation` don't share files. 

Simulation writes the calldata into `crates/tycho-execution/contracts/test/assets/calldata.txt`, and execution reads it from there.

For the executor tests, can we use foundry's vm cheatcodes to mock the state instead? Like forking at a specific block or using `vm.warp` to adjust the timestamp to cover the cases you listed. The main goal is just to verify the executor's swap logic is correct.

crates/tycho-execution/contracts/foundry.toml
```diff
@@ -10,7 +10,7 @@ via_ir = true
 # (protocols/testing/fixtures) are reproducible across Foundry versions.
 solc = '0.8.33'
 bytecode_hash = 'none'
-fs_permissions = [{ access = "read", path = "./test/assets" }, { access = "write", path = "./test" }]
+fs_permissions = [{ access = "read", path = "../../tycho-simulation/tests/assets" }, { access = "read", path = "./test/assets" }, { access = "write", path = "./test" }]
```
zach030, line 13: Same comment as the executor test above.

crates/tycho-indexer/src/cli.rs
```diff
@@ -224,6 +224,10 @@ pub struct RunSpkgArgs {
     #[clap(long)]
     stop_block: Option<String>,
 
+    /// Keep historical state from this UTC timestamp. Defaults to the current time.
```
zach030, line 227: Thanks for catching this! We'll open a separate PR for the fix and link it below.

protocols/substreams/base-baibai/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+        .into_iter()
+        .map(|(i, write)| attribute(&format!("word_{i}"), write.new_value.clone()))
+        .collect();
+    attrs.sort_by(|a, b| a.name.cmp(&b.name));
+    attrs
+}
+
+fn attribute(name: &str, value: Vec<u8>) -> Attribute {
+    Attribute { name: name.into(), value, change: ChangeType::Update.into() }
+}
```
zach030, line 192: If it's the first creation, the `ChangeType` should be `Creation`.

protocols/substreams/base-baibai/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+            // start_block is the entrypoint's deployment block, before CurveBook v3
+            // and custody storage exist. Subsequent writes replace these zero words.
+            let mut attributes: Vec<_> = (0..32)
+                .map(|i| attribute(&format!("word_{i}"), vec![0; 32]))
+                .collect();
+            attributes.push(attribute("balance_owner", config.custodian.to_vec()));
+            builder.add_entity_change(&EntityChanges { component_id: component.id, attributes });
+        }
+    }
+    if block.number >= config.start_block {
```
zach030, line 123: The check is redundant now, since the substreams `initialBlock` is already equal to `start_block`.

zach030 (commented):
Thanks for the PR! Left some comments, please have a look.

A few changes aren't related to your protocol, they look like tycho issues. We'll fix those in a separate PR, thanks for the catch! We'll keep you posted.

mteam88:
Updated in `dcd189964` to price BaiBai fees for the executing router and track fee changes through indexed state, while retaining normal router slippage protection. Also addressed WETH balance events, Creation attributes, address filtering, executor conventions/Slither, optional Base RPC and independent fork tests. BaiBai setup is local to its test class, avoiding conflicts with upstream executor registrations. Fresh replay passed all six execution comparisons; independent storage, custody and fee-state checks passed. Leaving the separate retention-fix conversation open for the upstream follow-up.

crates/tycho-execution/contracts/src/executors/BaibaiExecutor.sol
mteam88: Zero is the current default; the router does not need an allowlist for it. Simulation now uses the execution router as taker and follows its indexed fee settings, including explicit-zero overrides. The executor uses the normal router minimum-output check, so fee changes are handled through pricing and ordinary slippage protection.

crates/tycho-indexer/src/cli.rs
mteam88: Removed the shared retention CLI/runner changes from this PR so they can be handled in your separate fix. BaiBai-specific caller context remains. The successful historical replay used the retention patch locally; the PR description now makes that validation dependency explicit.

protocols/testing/src/execution.rs
```diff
@@ -65,7 +66,10 @@ static EXECUTOR_MAPPING: LazyLock<HashMap<&'static str, &'static str>> = LazyLoc
 /// [`EXECUTOR_MAPPING`], which holds the executor used on every other chain.
 static CHAIN_SPECIFIC_EXECUTORS: LazyLock<HashMap<(Chain, &'static str), &'static str>> =
     LazyLock::new(|| {
-        HashMap::from([((Chain::Robinhood, "ekubo_v3"), EKUBO_V3_ROBINHOOD_BYTECODE_JSON)])
+        HashMap::from([
+            ((Chain::Robinhood, "ekubo_v3"), EKUBO_V3_ROBINHOOD_BYTECODE_JSON),
+            ((Chain::Base, "baibai"), BAIBAI_BYTECODE_JSON),
```
zach030: No need to add baibai to `CHAIN_SPECIFIC_EXECUTORS`, it's only deployed on Base, so just adding it to `EXECUTOR_MAPPING` should be enough.

protocols/substreams/base-baibai/src/config.rs
```diff
@@ -0,0 +1,136 @@
...
+            [config.entrypoint, config.curve_book, config.custodian, config.base, config.quote];
+        ensure!(
+            addresses
+                .iter()
+                .all(|address| !address.is_zero()),
+            "zero BaiBai address"
+        );
+        for (i, address) in addresses.iter().enumerate() {
+            ensure!(!addresses[..i].contains(address), "duplicate BaiBai address");
+        }
```
zach030: This feels too defensive, the config is maintained by us anyway.

protocols/substreams/base-baibai/src/lib.rs
```diff
@@ -0,0 +1,506 @@
...
+        .map(|(i, write)| attribute(&format!("word_{i}"), write.new_value.clone()))
+        .collect();
+    attrs.sort_by(|a, b| a.name.cmp(&b.name));
+    attrs
+}
+
+/// Fee events are emitted by the proxy, including during delegatecall. Only the final
+/// value per transaction matters; clearing an override retains an explicit unconfigured value.
+fn fee_attributes(config: &Config, tx: &eth::v2::TransactionTrace) -> Result<Vec<Attribute>> {
+    use alloy_primitives::{keccak256, Address, U256};
```
zach030: I'd suggest move this `use` to the top of the file.

protocols/substreams/base-baibai/src/lib.rs
```diff
@@ -0,0 +1,506 @@
...
+    if !tx
+        .calls
+        .iter()
+        .any(|call| !call.state_reverted && call.address == config.entrypoint.as_slice())
+    {
+        return Ok(vec![]);
+    }
+    let set = keccak256("TakerFeeSet(address,address,uint16)");
+    let clear = keccak256("TakerFeeCleared(address,address)");
```
zach030: Can we just hardcode the computed value? 
e.g. `pub const SET_TOPIC: [u8; 32] = hex!("xxxxx");`

protocols/substreams/base-baibai/src/lib.rs
```diff
@@ -0,0 +1,506 @@
...
+            "invalid BaiBai fee event topics"
+        );
+        let base = Address::from_slice(&log.topics[2][12..]);
+        if base != config.base && !base.is_zero() {
+            continue;
+        }
+        let configured = topic == set.as_slice();
+        anyhow::ensure!(!configured || log.data.len() == 32, "invalid BaiBai fee event data");
+        let bps = if configured { U256::from_be_slice(&log.data) } else { U256::ZERO };
+        anyhow::ensure!(bps <= U256::from(1000), "invalid BaiBai fee");
```
zach030: Can we remove this check in the substreams? If bps is ever updated to over 1000, the whole protocol crashes. Better to move this check to the simulation side.

protocols/substreams/base-baibai/src/lib.rs
```diff
@@ -0,0 +1,506 @@
...
+        .filter(|(log, _)| log.address == config.entrypoint.as_slice())
+    {
+        let Some(topic) = log.topics.first() else { continue };
+        if topic != set.as_slice() && topic != clear.as_slice() {
+            continue;
+        }
+        anyhow::ensure!(
+            log.topics.len() == 3 && log.topics[1].len() == 32 && log.topics[2].len() == 32,
+            "invalid BaiBai fee event topics"
+        );
```
zach030: This validation is redundant imo. `log.topics.len` is tied to the contract, and if the entrypoint contract gets upgraded, our whole integration would break anyway. As for `log.topics[1].len()`, it's always 32, so the extra check isn't needed either.

protocols/substreams/base-baibai/src/lib.rs
```diff
@@ -0,0 +1,506 @@
...
+        let base = Address::from_slice(&log.topics[2][12..]);
+        if base != config.base && !base.is_zero() {
+            continue;
+        }
+        let configured = topic == set.as_slice();
+        anyhow::ensure!(!configured || log.data.len() == 32, "invalid BaiBai fee event data");
```
zach030: Same, this is also redundant.

protocols/testing/scripts/update_runtime_bytecode.sh
zach030, line 1: Can you revert this change? We have a new approach to handle this, and we'll add baibai once it's merged.

protocols/substreams/base-baibai/src/lib.rs
```diff
@@ -0,0 +1,506 @@
...
+        anyhow::ensure!(bps <= U256::from(1000), "invalid BaiBai fee");
+        let name = format!(
+            "{}_fee_{:x}",
+            if base.is_zero() { "router" } else { "pair" },
+            Address::from_slice(&log.topics[1][12..])
+        );
+        let value = vec![u8::from(configured), (bps.to::<u16>() >> 8) as u8, bps.to::<u16>() as u8];
+        let entry = latest
+            .entry(name)
+            .or_insert((0, vec![]));
```
zach030, line 249: Can you add more comments here on the fee mechanism? 
From what I understand, there are three fees: 
- `default_fee_bps` (the default)
- `pair_fee` (per-pair)
- `router_fee` (per-taker). 

Also, `router_fee` isn't a accurate name, can we rename it to `taker_fee`?

protocols/substreams/base-baibai/src/config.rs
```diff
@@ -0,0 +1,136 @@
...
+        Ok(config)
+    }
+
+    pub fn default_fee_slot(&self) -> B256 {
+        B256::from(namespace("baibai.storage.Entrypoint") + U256::from(1))
+    }
```
zach030: Can you double-check the storage layout? I checked the storage values and slot `ns+1` is always zero.

protocols/substreams/base-baibai/src/lib.rs
```diff
@@ -0,0 +1,506 @@
...
+        for write in &call.storage_changes {
+            if write.address == config.entrypoint.as_slice() &&
+                write.key == slot.as_slice() &&
+                default
+                    .map_or(true, |last: &eth::v2::StorageChange| write.ordinal > last.ordinal)
+            {
+                default = Some(write);
+            }
+        }
+    }
```
zach030: I'd suggest moving this into the `storage_attributes` function, so we don't have to iterate over all the `storage_changes` again.

crates/tycho-simulation/src/evm/protocol/baibai/decoder.rs
```diff
@@ -0,0 +1,168 @@
...
+        let caller = context
+            .caller
+            .as_ref()
+            .filter(|caller| caller.len() == 20)
+            .ok_or_else(|| {
+                InvalidSnapshotError::ValueError(
+                    "BaiBai requires DecoderContext::caller with the execution router address"
+                        .into(),
+                )
+            })?;
```
zach030: `caller` is optional, so we shouldn't error out here. I'd fall back to the default Tycho router on Base instead. (We'd need a chain check in the decode.)

.github/workflows/ci-foundry.yaml
zach030, line 1: Pls revert this. We'll handle it together with a separate PR for the runtime bytecode.

crates/tycho-execution/contracts/test/protocols/Baibai.t.sol
zach030, line 1: Can you add more test cases in `crates/tycho-execution/tests/protocol_integration_tests.rs`? The `Baibai.t.sol` tests will read the test cases from `calldata.txt` and run them here.

zach030 (commented):
Thanks for the update! Did another round of review, pls have a look!

One question: the current substreams hardcodes the `entrypoint` and the `curve_book` contract slot it depends on. But if the contract gets upgraded, would the dependent slots and the pricing logic change? I noticed the entrypoint was [updated](https://basescan.org/tx/0xd94ac5ecaab6f3605b268d0aa678590caf02aee3dfd1251b91e0f49c1286a203) a few days ago. Also, is the custodian upgradeable too? If the custodian changes, we'd need to update `relative_balances` to query the new custodian's balance.

Based on our past integration experience, if a protocol can be upgraded, it's best to just pause all the components. tycho-substreams already provides a `change_component_pause_state` method to pause components.

mteam88:
Addressed the latest review in 1f37dc7c7. The entrypoint, CurveBook and custodian are all upgradeable; upgrades after the validated history now pause the pool, and native simulation honors that pause. The cited entrypoint upgrade is already included in the pinned replay. Removed the unused default-fee slot, renamed the taker-wide attribute, added the Base-router fallback and Rust-to-Foundry calldata cases, and reverted the runtime-script/CI changes as requested.

Fresh full replay passed with six exact execution matches, plus independent storage/balance/fee-history comparison. Rust CI: 1,181 tests passed. The retention follow-up remains open; its patch was used only in the isolated validation checkout. One Slither warning remains in unchanged upstream `TychoFallbackRouter`; no BaiBai findings.

protocols/substreams/base-baibai/src/config.rs
mteam88: Confirmed: the deployed entrypoint stores only the taker-fee mapping in that namespace; ns+1 is not a default-fee field. Removed the slot read and use the deployed fallback of zero. This also removes the second storage scan. Updates to contract behavior are covered by the new dependency-upgrade pause.

protocols/substreams/base-baibai/src/lib.rs
mteam88: Renamed to taker_fee_<taker> and documented precedence: the taker's pair override, then the taker-wide override (base zero), then zero. A configured zero remains an override; clearing an entry falls through to the next level.

protocols/substreams/base-baibai/src/lib.rs
mteam88: An omitted caller now defaults to the Tycho Base router, with an explicit Base chain check. Snapshot tests cover the fallback selecting that router's fee, explicit caller selection, other-chain rejection and malformed caller addresses.

crates/tycho-execution/contracts/test/protocols/Baibai.t.sol
mteam88: Added both-direction cases to protocol_integration_tests.rs. Baibai.t.sol now reads their generated calldata.txt entries and executes them on the Base fork, both with zero fees and with a router fee distinct from the end user's fee. All six BaiBai fork tests pass.

mteam88:
@zach030 rebased onto current main and resolved the conflicts. Updated the generated BaiBai test calldata for the changed upstream executor setup; all six fork tests, native/encoder/Substreams tests and scoped Clippy pass. Could you take another look?

protocols/substreams/base-baibai/src/config.rs
```diff
@@ -0,0 +1,120 @@
...
+use anyhow::Result;
+use substreams_ethereum::pb::eth::v2::{Block, TransactionTrace};
+
+#[derive(serde::Deserialize)]
+pub struct Config {
+    pub entrypoint: Address,
+    pub curve_book: Address,
+    pub custodian: Address,
+    pub base: Address,
+    pub quote: Address,
```
zach030: Do you only support one pair now? Would it expand in the future?

crates/tycho-simulation/src/evm/protocol/baibai/decoder.rs
```diff
@@ -0,0 +1,176 @@
...
+        }
+        if snapshot
+            .state
+            .attributes
+            .contains_key("paused")
+        {
+            return Err(InvalidSnapshotError::ValueError(
+                "BaiBai is paused after an upgrade".into(),
+            ));
+        }
```
zach030: We'll build the pause-removal logic into tycho-client, so we can remove this here for now.

crates/tycho-simulation/src/evm/protocol/baibai/decoder.rs
```diff
@@ -0,0 +1,176 @@
...
+            paused: false,
+        };
+        apply_words(&mut state.words, &snapshot.state.attributes, true)?;
+        if !snapshot
+            .state
+            .attributes
+            .contains_key("fees_indexed")
+        {
+            return Err(InvalidSnapshotError::MissingAttribute("fees_indexed".into()));
+        }
```
zach030: What's `fees_indexed` for? It's created by default in the substreams and always set to 1.

protocols/testing/Cargo.toml
```diff
@@ -4,6 +4,7 @@ version.workspace = true
 edition.workspace = true
 
 [dependencies]
+chrono           = { workspace = true }
```
zach030: This should be removed.

zach030 (commented):
Thanks! Looks good! Just left a few comments.

mteam88:
Heads-up: the BaiBai CurveBook was upgraded on Base at block **51696183** ([tx 0xc9d0e733da27ab34e42e499e704b881b39fda83a4cbe1e4545d2bb41e666ef5b](https://basescan.org/tx/0xc9d0e733da27ab34e42e499e704b881b39fda83a4cbe1e4545d2bb41e666ef5b)). The new implementation, [0xAd6beCedB2c4b898CA9d4dE248Dc7207fe4d1C80](https://basescan.org/address/0xAd6beCedB2c4b898CA9d4dE248Dc7207fe4d1C80#code), is source-verified.

**What changed:** the book now emits an explicit pair-creation event:

```solidity
event PairListed(address indexed base, address indexed quote);
```

A base's first successful SHAPE emits it. The upgrade itself backfilled it once for WETH/USDC through a one-shot `announcePairs` reinitializer. Nothing else changed: `CurveUpdated`, `Fill`, quote/swap math, and existing storage are the same. The diff is the event, a single emit in `_applyShape`, and the reinitializer ([spire-labs/mm-platform#89](https://github.com/spire-labs/mm-platform/pull/89)).

**Effect on this PR:** 51696183 is past `VALIDATED_BLOCK` (51,191,196), so as written the adapter pauses every BaiBai component from that block on. My next push will:

- raise `VALIDATED_BLOCK` to 51,696,183, after reviewing the new implementation's wiring and storage layout
- keep pair discovery on the first `CurveUpdated`. WETH's history starts before the new event existed, so `PairListed` alone can't cover the replay from `initialBlock`. Consumers who want a single creation-event filter can use `PairListed` from 51696183 onward.
- rerun the replay and the fork/native/Substreams tests across the upgrade block

PR #1446, opened 2026-09-15
feat(execution): quote the pAMM against its fallback before swapping

tamaralipows:
## What

The fallback router quotes the pAMM and the caller's fallback protocol before it swaps. When the fallback quotes more output, the fallback runs directly and the pAMM is never called. Otherwise the pAMM runs first and the fallback still catches its failure, as before.

## How it works

- The pAMM quote is the pAMM standard's own quote call.
- Uniswap V2 is priced from the pair's reserves. Uniswap V3 asks Eden Network's static quoter, a view contract that walks the pool's ticks off storage without swapping. Curve asks the pool for its output. Fluid asks the dex to price a swap paid to the dead address, which reverts with the amount before it moves any token.
- Uniswap V4 has no quote function. The router runs the real swap in a self-call and reverts with the receiver's balance diff, so the swap rolls back and only the amount survives.
- A pAMM that cannot quote does not get its swap attempted, so the fallback quote is skipped as well and the fallback runs straight away.
- Every fallback quote runs under try/catch and counts as zero when it fails. Malformed protocol data never reverts the swap on its own. Equal quotes keep the pAMM.
- The fallback event carries a reason: the fallback quoted higher, or the pAMM won the quote and then reverted or delivered nothing.

## Gas

A leg now takes one of four routes. Before this PR there were two, because nothing was quoted: the pAMM was always attempted first.

All numbers are gas of the `TychoFallbackRouter.swap` call alone, from cold state on a mainnet fork, with the mock pAMM from the tests. A real pAMM's quote and swap cost more than the mock's, but the added quote cost is the same. The failing pAMM reverts plainly; one that burns its gas costs more on top.

**1. The pAMM quotes higher and fills.** The leg now also pays for the fallback quote.

| pAMM fills | Before | After | Quote cost |
|---|---|---|---|
| Uniswap V2 | 71,222 | 82,361 | +11,139 |
| Uniswap V3 | 71,222 | 102,181 | +30,959 |
| Uniswap V4 | 76,866 | 183,441 | +106,575 |
| Curve | 72,307 | 111,387 | +39,080 |
| Fluid | 81,344 | 124,956 | +43,612 |

**2. The fallback quotes higher and fills.** This route is new. Before this PR the same leg went to the pAMM and filled there, at the worse price, which is why the Before column is a pAMM fill and the gas is not comparable. What the user gains is the price difference between the two venues.

| Fallback fills on price | Before (pAMM filled) | After (fallback fills) |
|---|---|---|
| Uniswap V2 | 71,222 | 99,537 |
| Uniswap V3 | 71,222 | 124,455 |
| Uniswap V4 | 76,866 | 212,451 |
| Curve | 72,307 | 216,952 |
| Fluid | 81,344 | 207,214 |

**3. The pAMM quotes higher, then reverts on-chain.** Both venues are quoted, the pAMM is attempted and rolled back, and the fallback fills. This is the most expensive route, and the quote makes it more expensive than before.

| pAMM wins, then fails | Before | After |
|---|---|---|
| Uniswap V2 | 131,676 | 139,513 |
| Uniswap V3 | 141,244 | 164,431 |
| Uniswap V4 | 144,140 | 257,038 |
| Curve | 240,778 | 266,755 |
| Fluid | 210,052 | 251,781 |

**4. The pAMM cannot quote.** Its swap is not attempted and the fallback quote is skipped too, so the fallback runs straight away. This route is about 36k cheaper than before.

| pAMM cannot quote | Before | After |
|---|---|---|
| Uniswap V2 | 138,350 | 102,377 |
| Uniswap V3 | 147,918 | 111,474 |
| Uniswap V4 | 150,055 | 117,354 |
| Curve | 246,715 | 208,347 |
| Fluid | 215,956 | 177,574 |

For reference, the three ways to quote the same 10k USDC swap on the USDC/WETH 0.05% pool, all returning the same amount: static quoter 28,149, in-contract simulation 108,568, Uniswap QuoterV2 114,567.

## Configuration

The router takes a third constructor argument, the Uniswap V3 static quoter. The deploy script reads it from the protocol-specific address config under `fallback_router.uniswap_v3_static_quoter`; Ethereum is set to Eden's deployment. The swap data the encoder emits for `FallbackExecutor` is byte-for-byte the same as before: `[tokenIn][tokenOut][pamm][protocol byte][protocol data]`. The `FallbackSwap` event gained a `reason` field, so anything decoding that event needs the new ABI.


crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
```diff
@@ -168,23 +220,125 @@ contract TychoFallbackRouter is ReentrancyGuardTransient {
...
+        }
+        if (protocol == FallbackProtocol.UniswapV3) {
+            return _quoteUniswapV3(swap_, protocolData);
+        }
+        if (protocol == FallbackProtocol.Curve) {
+            return _quoteCurve(swap_, protocolData);
+        }
+        if (protocol == FallbackProtocol.FluidV1) {
+            return _quoteFluidV1(swap_, protocolData);
+        }
```
Troshchk: Why V4 is not here?

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
```diff
@@ -168,23 +220,125 @@ contract TychoFallbackRouter is ReentrancyGuardTransient {
...
+    }
+
+    /// @notice Runs the fallback, then reverts `TychoFallbackRouter__SimulatedAmountOut` with
+    /// the amount it delivered so the swap rolls back. External only so `quoteFallback` can
+    /// try/catch it.
+    function simulateFallback(Swap calldata swap_, bytes calldata fallbackSwap)
+        external
+    {
+        _requireSelf();
```
Troshchk, line 268: Is this than needed in quoteFallbak()?
`_requireSelf();` is also called there

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
```diff
@@ -168,23 +220,125 @@ contract TychoFallbackRouter is ReentrancyGuardTransient {
...
+        returns (uint256 amountOut)
+    {
+        _requireSelf();
+        (FallbackProtocol protocol, bytes calldata protocolData) =
+            _decodeFallback(fallbackSwap);
+
+        if (protocol == FallbackProtocol.UniswapV2) {
+            (IUniswapV2Pair pair, uint256 feeBps) =
+                _decodeUniswapV2(protocolData);
+            return _uniswapV2AmountOut(swap_, pair, feeBps);
```
Troshchk: I wonder if it is possible to name it the same way as the other protocols: _quote ?

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
Troshchk: Ah, got it from the docs later -- v4 value comes from `simulateFallback`, but why don't we match it like other protocols:
`if (protocol == FallbackProtocol.UniswapV4) {`

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
```diff
@@ -168,23 +220,125 @@ contract TychoFallbackRouter is ReentrancyGuardTransient {
...
+
+    /// @notice Runs the fallback, then reverts `TychoFallbackRouter__SimulatedAmountOut` with
+    /// the amount it delivered so the swap rolls back. External only so `quoteFallback` can
+    /// try/catch it.
+    function simulateFallback(Swap calldata swap_, bytes calldata fallbackSwap)
+        external
+    {
+        _requireSelf();
+        uint256 balanceBefore = IERC20(swap_.tokenOut).balanceOf(swap_.receiver);
+        _executeFallback(swap_, fallbackSwap);
```
Troshchk: Isn't this function used only for uniswapv4?
Cannot we use `_swapUniswapV4` instead of  `_executeFallback`?

Troshchk (commented):
Thank you!
Leaving some qs

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
```diff
@@ -194,48 +348,76 @@ contract TychoFallbackRouter is ReentrancyGuardTransient {
             _swapUniswapV4(swap_, protocolData);
         } else if (protocol == FallbackProtocol.Curve) {
             _swapCurve(swap_, protocolData);
-        } else if (protocol == FallbackProtocol.FluidV1) {
-            _swapFluidV1(swap_, protocolData);
         } else {
+            _swapFluidV1(swap_, protocolData);
+        }
+    }
```
louise-poole, line 371: Very good suggestion - please implement?

---

🤖 `_executeFallback` used to name `FallbackProtocol.FluidV1` explicitly and end with `revert TychoFallbackRouter__UnknownProtocol`. It now ends with a bare `else` that calls `_swapFluidV1`. Add a value to `FallbackProtocol` and forget a branch, and that protocol's legs run against Fluid with the caller's pool bytes instead of reverting. The only guard left is `_decodeFallback`, which bounds-checks the byte against `type(FallbackProtocol).max` and so passes the new value straight through.

Restore the explicit `else if (protocol == FallbackProtocol.FluidV1)` branch and keep the `UnknownProtocol` revert as the final `else`.

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
```diff
@@ -168,23 +220,125 @@ contract TychoFallbackRouter is ReentrancyGuardTransient {
...
+        address pool = _decodeUniswapV3(data);
+        bool zeroForOne = swap_.tokenIn < swap_.tokenOut;
+        (int256 amount0, int256 amount1) = uniswapV3StaticQuoter.quote(
+            pool,
+            zeroForOne,
+            int256(swap_.amountIn),
+            zeroForOne
+                ? TickMath.MIN_SQRT_PRICE + 1
+                : TickMath.MAX_SQRT_PRICE - 1
+        );
```
louise-poole: Minor suggestion

---

🤖 `_quoteUniswapV3` and `_swapUniswapV3` each compute `zeroForOne` from the token sort order and each pick `MIN_SQRT_PRICE + 1` / `MAX_SQRT_PRICE - 1`. Change the limit in one and the quote stops describing the swap that runs, with nothing to catch it — the router then compares the pAMM against a price it will not fill at.

Extract `_uniswapV3Params(Swap calldata swap_) returns (bool zeroForOne, uint160 sqrtPriceLimit)` and call it from both.

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
```diff
@@ -93,60 +108,97 @@ contract TychoFallbackRouter is ReentrancyGuardTransient {
...
         if (
             address(poolManager_) == address(0) || fluidLiquidity_ == address(0)
+                || address(uniswapV3StaticQuoter_) == address(0)
         ) {
             revert TychoFallbackRouter__AddressZero();
         }
         poolManager = poolManager_;
         fluidLiquidity = fluidLiquidity_;
+        uniswapV3StaticQuoter = uniswapV3StaticQuoter_;
     }
```
louise-poole, line 141: This leads me to ask: would we deploy a different fallback executor per chain depending on the protocols we can fall back to? Or do we want one giant generic one that should work on all chains?

---

🤖 `uniswapV3StaticQuoter` is rejected at zero, so the router cannot be deployed on a chain where Eden has no static quoter — even for a deployment that would only ever encode Uniswap V2, Curve or Fluid fallbacks. `ethereum` is the only chain carrying the key today.

Accept a zero quoter and let `_quoteUniswapV3` revert into the zero-quote path that already exists, which degrades a V3 fallback to "never preferred over the pAMM" rather than blocking deployment. Take that over widening the deploy script's error message.

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
```diff
@@ -93,60 +108,97 @@ contract TychoFallbackRouter is ReentrancyGuardTransient {
...
 
+        FallbackReason reason = FallbackReason.FallbackQuotedHigher;
+        if (fallbackAmountOut <= pammAmountOut) {
+            try this.executePropAMM(swap_, pamm) {
+                return;
+            } catch {}
+            reason = FallbackReason.PropAMMFailed;
+        }
+
         FallbackProtocol protocol = _executeFallback(swap_, fallbackSwap);
```
louise-poole, line 191: Good suggestion - but I don't think it's vital if it adds too much complexity/overhead.

---

🤖 When the fallback quotes higher, `_executeFallback` runs in the outer frame with no try/catch and the pAMM is never called at all. Plenty of pools answer a price question while refusing to trade: a killed Curve pool still serves `get_dy`, a paused Fluid dex still prices the dead-address estimate. Such a pool now reverts the whole route, where before this change the pAMM always got the first attempt and would have filled it.

A trade sent through this router is meant to survive one venue failing, which is the reason it exists. With the new pricing step, whichever venue prices higher becomes the only one tried, and a pool can quote a price it will not honour. A user whose route lands on such a pool sees the whole transaction revert and pays the gas, even though the other venue was ready to fill. The venue that wins the price comparison should get the same second-chance handling the other one already gets.

Run the fallback-first path through a self-call under try/catch as well, and attempt the pAMM when it reverts. Two attempts total either way, just in the other order.

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
```diff
@@ -93,60 +108,97 @@ contract TychoFallbackRouter is ReentrancyGuardTransient {
...
+            fallbackAmountOut = amountOut;
         } catch {}
 
+        FallbackReason reason = FallbackReason.FallbackQuotedHigher;
+        if (fallbackAmountOut <= pammAmountOut) {
+            try this.executePropAMM(swap_, pamm) {
+                return;
+            } catch {}
+            reason = FallbackReason.PropAMMFailed;
+        }
```
louise-poole: Okay suggestion - I do think this might be taking it too far though and exploding the gas usage.

---

🤖 `executePropAMM` still passes `minAmountOut = 0` and only checks that the receiver's balance moved at all. A pAMM that quotes above the fallback wins the comparison, delivers one wei, and the fallback it displaced is never run. The comparison picks the venue but nothing holds the winner to the number it won with, so the quote is advisory in exactly the case where it decided something.

The router now asks both venues what they would pay and sends the trade to whichever promises more. Nothing checks the winner keeps its promise: it can promise a large amount, win, then hand over dust, and the venue that lost is never tried. The user falls back on the route-wide minimum they set, which can sit far below what the losing venue would have paid. A venue that wins on price should have to deliver at least what it beat, or fall through.

Pass `fallbackAmountOut` into `executePropAMM` and revert when the receiver's balance diff lands below it. The existing try/catch then falls through to the fallback exactly as it already does for `TychoFallbackRouter__NoOutput`, and the transfer to the pAMM rolls back with it.

louise-poole, line 189: Good optimisation improvement

---

🤖 With `pammAmountOut == 0` the fallback runs unless it also quotes zero, so the quote's only remaining job is deciding whether to attempt a pAMM that has just said it cannot serve the pair. For Uniswap V4 that decision costs a full simulated swap — the PR's own table puts the stale-pAMM V4 path at 226,511 gas against 149,691 before.

Go straight to `_executeFallback` when `pammAmountOut == 0` and skip the quote. This changes behavior in one case — a pAMM and a fallback that both quote zero no longer attempt the pAMM swap — so confirm that case is acceptable before taking it.

crates/tycho-execution/contracts/interfaces/ICurvePool.sol
```diff
@@ -0,0 +1,48 @@
...
+}
+
+/// @notice The Curve pool type byte the encoder emits.
+library CurvePoolType {
+    /// @notice Stable (1) and stable_ng (10) pools take the `int128` index signatures; crypto
+    /// and llamma pools take `uint256`.
+    function isStable(uint8 poolType) internal pure returns (bool) {
+        return poolType == 1 || poolType == 10;
+    }
+}
```
louise-poole: Good minor suggestion

---

🤖 A file-level `function isCurveStablePool(uint8 poolType) pure returns (bool)` serves the same three call sites with one fewer name to import.

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
```diff
@@ -168,23 +220,125 @@ contract TychoFallbackRouter is ReentrancyGuardTransient {
         }
     }
 
+    /// @notice Quotes the fallback protocol. External only so `swap` can try/catch it: a
+    /// protocol that cannot quote, or malformed protocol data, is a zero quote.
```
louise-poole: Small doc improvement

---

🤖 The NatSpec says "a protocol that cannot quote, or malformed protocol data, is a zero quote", but `quoteFallback` reverts in both cases — `testQuoteFallbackRevertsWithTheCause` asserts exactly that. The zero comes from `swap`'s `try/catch`. Anyone reading the function on its own expects a return value it never produces.

Say what it does: it reverts with the protocol's own error, or the decoder's, when it cannot quote. The zero-quote consequence belongs on `swap`, which already states it.

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
```diff
@@ -168,23 +220,125 @@ contract TychoFallbackRouter is ReentrancyGuardTransient {
...
+        external
+    {
+        _requireSelf();
+        uint256 balanceBefore = IERC20(swap_.tokenOut).balanceOf(swap_.receiver);
+        _executeFallback(swap_, fallbackSwap);
+        revert TychoFallbackRouter__SimulatedAmountOut(IERC20(swap_.tokenOut)
+                    .balanceOf(swap_.receiver) - balanceBefore);
+    }
+
+    /// @dev Same arguments as `_swapUniswapV3`, so the quote is the fill.
```
louise-poole: Small doc improvement

---

🤖 `_quoteUniswapV3`'s `@dev` reads "Same arguments as `_swapUniswapV3`, so the quote is the fill". The arguments do match, but the quote comes from Eden's reimplementation of the tick walk while the fill comes from the pool, so the two diverge on anything the quoter models differently.

Rewrite as "Same arguments as `_swapUniswapV3`, priced by the static quoter's own tick walk rather than by the pool."

crates/tycho-execution/CLAUDE.md
```diff
@@ -167,14 +167,30 @@ TychoRouterV3 --TransferType.Transfer--> TychoFallbackRouter --> pAMM     (rever
...
+`uniswapV3StaticQuoter` immutable, `IUniswapV3StaticQuoter`), Curve asks `get_dy`, Fluid asks the dex to price a
+swap paid to `0xdEaD`, which reverts `FluidDexSwapResult(amountOut)` before moving any token. Uniswap V4 has no quote
+function, so `simulateFallback` runs the real swap and reverts `TychoFallbackRouter__SimulatedAmountOut` with the
+receiver's balance diff, rolling it back. The pAMM quote is a low-level call whose return data counts only when it
+is at least 32 bytes; the fallback quote runs in the self-only external `quoteFallback` under try/catch. A quote that
+reverts, returns nothing decodable or gets malformed protocol data counts as zero. Two zero quotes or equal quotes keep
+the pAMM-first order. The fallback quote costs gas on every leg,
+including the ones the pAMM fills: about 110k for a Uniswap V4 simulation, about 40k for Curve or Fluid, about 30k for
+Uniswap V3, about 12k for Uniswap V2 (router call, cold state, mainnet fork). A stale pAMM is skipped without its
+swap being attempted, which saves that failed attempt.
```
louise-poole: Small doc improvement

---

🤖 "Two zero quotes or equal quotes keep the pAMM-first order" and "A stale pAMM is skipped without its swap being attempted" cannot both hold. A stale pAMM whose fallback also fails to quote does get attempted, which `testUnquotableFallbackKeepsPropAMMFirst` covers. The trailing "which saves that failed attempt" reads as a note on the previous implementation rather than a statement of current behavior.

Scope the claim — "a pAMM that cannot quote is skipped whenever the fallback quotes anything" — and drop the comparison to what the router used to do.

louise-poole (commented):
The one I'd want fixed before merge is the bare `else` in `_executeFallback` — an unmatched protocol byte swapping against Fluid isn't a failure mode I want sitting there. The rest is minor: two questions on the quote path, and doc and naming nits.

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
```diff
@@ -168,23 +220,125 @@ contract TychoFallbackRouter is ReentrancyGuardTransient {
...
+        try IFluidV1Dex(dex)
+            .swapIn(zero2one, swap_.amountIn, 0, address(0xdEaD)) {
+            return 0;
+        } catch (bytes memory reason) {
+            return _decodeAmountOut(reason, FluidDexSwapResult.selector);
+        }
+    }
+
+    /// @dev The `uint256` a `selector(uint256)` revert carries, or zero for any other revert.
+    function _decodeAmountOut(bytes memory reason, bytes4 selector)
```
louise-poole: Nice minor suggestion

---

🤖 It decodes `selector(uint256)` revert data, not a return value. Rename it `_amountOutFromRevert` and the `@dev` line stops being load-bearing.

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
```diff
@@ -93,60 +108,97 @@ contract TychoFallbackRouter is ReentrancyGuardTransient {
...
+            quoted && quote.length >= 32 ? abi.decode(quote, (uint256)) : 0;
+
+        uint256 fallbackAmountOut = 0;
+        try this.quoteFallback(swap_, fallbackSwap) returns (
+            uint256 amountOut
+        ) {
+            fallbackAmountOut = amountOut;
         } catch {}
 
+        FallbackReason reason = FallbackReason.FallbackQuotedHigher;
```
louise-poole, line 171: Good small suggestion

---

🤖 `swap` uses `reason` for a `FallbackReason`; `quoteFallback` and `_quoteFluidV1` use it for raw revert bytes. Rename the revert-bytes locals to `revertData`.

crates/tycho-execution/contracts/test/protocols/Fallback.t.sol
```diff
@@ -205,13 +244,17 @@ contract TychoFallbackRouterTest is TychoFallbackRouterTestBase {
     uint256 constant CURVE_USDC_OUT = 999_895_324;
     uint256 constant CURVE_CRYPTO_USDC_OUT = 2_766_051_040;
 
+    /// A MockPropAMM price of 5 WETH per 10 000 USDC: above every pool here, so a pAMM at this
+    /// price wins the quote.
+    uint256 constant PAMM_ABOVE_MARKET = 5e26;
```
louise-poole: Good sml suggestion

---

🤖 The constant is a MockPropAMM price. `PAMM_PRICE_ABOVE_MARKET` reads correctly at the `setPrice` call sites.

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
tamaralipows: I assume this will be on your PR?

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
tamaralipows: Done in d815937. Both `_executeFallback` and `quoteFallback` now name every protocol and revert `TychoFallbackRouter__UnknownProtocol` in the final `else`.

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
tamaralipows: Done. `_uniswapV3Params` returns the direction and price limit, called from both `_quoteUniswapV3` and `_swapUniswapV3`.

crates/tycho-execution/contracts/interfaces/ICurvePool.sol
tamaralipows: Done. `isCurveStablePool` is now a file-level function in `ICurvePool.sol`.

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
tamaralipows: Done. The NatSpec now says it reverts with the protocol's own error, or the decoder's.

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
tamaralipows: Done. It now reads "priced by the static quoter's own tick walk rather than by the pool".

crates/tycho-execution/CLAUDE.md
tamaralipows: Done, in `CLAUDE.md` and in the `FallbackReason` enum comment, which carried the same claim: a pAMM that cannot quote is skipped only when the fallback quotes something.

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
tamaralipows: Done. Renamed `_amountOutFromRevert`, and its `reason` argument to `revertData`.

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
tamaralipows: Done. Both revert-bytes locals are `revertData` now.

crates/tycho-execution/contracts/test/protocols/Fallback.t.sol
tamaralipows: Done. Renamed `PAMM_PRICE_ABOVE_MARKET`.

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
tamaralipows: Good catch — added. `quoteFallback` now has an explicit `if (protocol == FallbackProtocol.UniswapV4)` branch, and the final `else` reverts `TychoFallbackRouter__UnknownProtocol` instead of falling into the simulation.

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
tamaralipows: Yes, it is needed. `simulateUniswapV4` is its own external entry point, so without `_requireSelf()` anyone could call it directly. `quoteFallback` reaches it through `this.`, so `msg.sender` is the router. `testQuoteEntryPointsRejectExternalCaller` covers both.

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
tamaralipows: Done — renamed `_quoteUniswapV2`.

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
tamaralipows: Done. It is Uniswap V4 only, so it is now `simulateUniswapV4(Swap calldata swap_, bytes calldata protocolData)` and calls `_swapUniswapV4` directly. That also drops the second decode of the protocol byte.

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
tamaralipows: Done in 465f846. `swap` goes straight to `_executeFallback` when the pAMM quotes zero. The both-zero case is acceptable: a pAMM whose `quote` reverts cannot fill either. `testUnquotablePropAMMWithUnquotableFallback` pins it with a pAMM that reverts on `quote` but pays on `swap`.

Gas on the stale-pAMM path: V4 365,632 to 278,873, Fluid 282,816 to 252,005, Curve 369,637 to 352,072, V3 306,612 to 290,746, V2 279,820 to 276,611.

tamaralipows: Leaving `minAmountOut = 0` for now. The router-level minimum stays the binding check.

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
tamaralipows: Out of scope here — leaving the zero check as it is and handling the per-chain question in the follow-up PR.

crates/tycho-execution/contracts/src/fallback/TychoFallbackRouter.sol
tamaralipows: Not taking this one. No other executor retries a regular AMM — we do not expect one to revert. The fallback exists for pAMM failure, so the second attempt belongs on the pAMM side only.

louise-poole (approved):
LGTM! Thank youuu!

PR #1451, opened 2026-09-15
feat(indexer): implement DeltaWindow retention, reverts and resolution

brunoguerios:
## Problem

The `DeltaWindow` skeleton (#1303) has `todo!()` bodies. Nothing can retain, resolve, or
report commit status yet.

## Fix

Implements `insert` with monotonic watermarks and the revert guard, `eviction_bound`,
`floor`, `resolve`, and the watermark-based `commit_status`. Adds `ReorgBuffer::oldest_block`
and exposes `BlockNumberOrTimestamp::greater_than` to the crate. `fold_and_evict` stays
`todo!()` for the next PR.

## Notes for reviewers

No production wiring; the file keeps `#![allow(dead_code)]`. The revert guard errors when a
purge would remove a block at or below `min(finalized, db_committed)`. That can only happen
after the extractor has already rejected the same revert, since `db_committed <= finalized`,
so the guard is an assertion rather than a recovery path. `resolve` reports versions below
the floor as `BelowFloor`; today's buffer serves the front block for those. The only consumer
is the future state service, so nothing observable changes here. `commit_status` no longer
reports the oldest buffered block as `Committed`; the facade PR wires it in.

## Follow-ups

ENG-6298 fold on eviction, stacked on this branch.


brunoguerios:
Superseded by #1456, which reviews the whole DeltaWindow stack against main as one branch. Retention, the revert guard and version resolution are in it unchanged. Closing to keep review in one place; the branch goes away with it.

PR #1452, opened 2026-09-15
feat(indexer): fold evicted DeltaWindow blocks into a sink

brunoguerios:
## Problem

Blocks can be retained but never leave the window: `fold_and_evict` is `todo!()`.

## Fix

Implements `fold_and_evict`: no-op until `min_fold_batch` blocks sit at or below the eviction
bound, then folds them in ascending order into the `FoldSink` and evicts exactly the folded
prefix. A fold error leaves the failing block and everything above it buffered and returns.
Adds `DiscardSink`, the only implementation until the entity cache exists. Records
`delta_window_fold_duration_ms{extractor}` per fold and warns above 5 ms.

## Notes for reviewers

`ReorgBuffer::drain_blocks_until` is untouched; the extractor's commit path shares it. The
metric name carries `_ms` to match `database_commit_duration_ms`; ENG-6310's table lists it
as `delta_window_fold_duration`, so Observability should adopt this name.

## Follow-ups

ENG-6300 wires the window into `PendingDeltas`. ENG-6291 replaces `DiscardSink` with the
entity cache.


brunoguerios:
Superseded by #1456. Fold-on-eviction, the fold batch size and the fold duration metric are in it unchanged. Closing to keep review in one place; the branch goes away with it.

PR #1453, opened 2026-09-15
feat(indexer): back PendingDeltas with DeltaWindow

brunoguerios:
## Problem

`PendingDeltas` drops each block the moment the database commits it. A request for a
just-committed version cannot be served from memory, and when `PendingDeltas::insert` fails
the pump exits and leaves the buffer frozen for that pod.

## Root cause

The drain policy lived inline in `PendingDeltas::insert` as `drain_blocks_until(db_committed + 1)`,
and the pump propagated any insert error with `?`.

## Fix

Each extractor's buffer is now a `DeltaWindow`. Blocks stay until they are finalized,
committed, and deeper than `--delta-window-depth` (env `DELTA_WINDOW_DEPTH`, default 128), then
fold into a discard sink. `--delta-window-fold-batch` (default 1) groups folds. The
`PendingDeltasBuffer` trait keeps its five signatures; the `rpc.rs` mock tests are untouched.
On an insert error the pump resets that extractor's window, as it already does on
`ExtractorRestarted`, and keeps serving the others.

## Notes for reviewers

Behaviour changes, all intended:
- `pending_deltas_buffer_size` grows. Prod today: Ethereum 7.3 MiB, Base 12.5 MiB per pod.
- `get_block_commit_status` no longer reports the oldest buffered block as `Committed`.
- `get_new_components` reads only blocks above `db_committed`, since window and database rows
  now overlap.
- `search_block` can find committed blocks, so more versions resolve from memory.
- A bad revert resets one window instead of ending the pump.

`ReorgBuffer::get_commit_status` and `PendingDeltas::new` are now reached only from tests and
carry `#[allow(dead_code)]`. The former keeps the rstest case that pins the buffer-bound
behaviour the window diverges from.

Rollback without a new image: `--delta-window-depth 1`, which reproduces the old drain.
On chains with a large `--database-insert-batch-size` the `db_committed` term binds and the
depth has no effect until it exceeds that batch.

## Follow-ups

ENG-6309 keeps ownership of these flags on `GlobalArgs`. ENG-6384 sets per-chain values.
The remaining `#[allow(dead_code)]` items go when ENG-6293 consumes them.


brunoguerios:
Superseded by #1456. PendingDeltas now sits on a DeltaWindow per extractor there, plus the depth and fold-batch CLI flags. Closing to keep review in one place; the branch goes away with it.

PR #1454, opened 2026-09-15
feat(indexer): add DeltaWindow patch capture

brunoguerios:
## Problem

The state service (ENG-6293) builds a response as a cache entry plus the window's changes
for the requested keys up to a version. The window applies those changes in place today and
has no way to hand them out as data.

## Fix

Adds `DeltaWindow::capture_patch(components, accounts, upto)` returning a `WindowPatch`: per
component, the state deltas and balances from each block that touched it; per account, the
account deltas and token balances. Ascending by block, filtered exactly as
`update_native_state` and `update_vm_state` filter today.

## Notes for reviewers

No consumer yet, so the items carry `#[allow(dead_code)]` until ENG-6293. Cloning is per
matched change, not per block. `upto` below the floor is the caller's responsibility, via
`resolve`.

ENG-6299 lists a third acceptance criterion: a race test where a patch is captured, a fold
runs, then the entry is read. That sequence spans capture, lock release, and a cache read, so
it cannot be exercised from the window alone — the cache and the lock discipline around it are
ENG-6293. The criterion moves there. The two criteria this PR can meet are covered:
capture returns exactly the changes between floor and V for both entity families, and it takes
no lock of its own.

## Follow-ups

ENG-6293 consumes it and may reshape `WindowPatch` to what the read path needs.


brunoguerios:
Superseded by #1456: the same branch, rebased onto main and reviewed there instead of on the stack. This PR's diff is stale after the rebase. Closing to keep review in one place.

PR #1456, opened 2026-09-16
feat(indexer): Delta Window (aggregated)

brunoguerios:
## What

`DeltaWindow` replaces the RPC-side `ReorgBuffer` inside `PendingDeltas`. It is the retention
layer the entity cache (ENG-6289) folds from: the cache needs committed blocks to stay in
memory long enough to be handed to it, and handed over exactly once. Nothing folds yet —
`DiscardSink` stands in until ENG-6291.

## How it works

One window per extractor. A block stays until it is at or below
`min(finalized, db_committed, tip - depth)`, then folds into a `FoldSink` and is evicted. Fold
and eviction are a single operation, so no block's deltas can be lost between the window and
the store behind the sink. Depth and fold batching come from `--delta-window-depth` (env
`DELTA_WINDOW_DEPTH`, default 128) and `--delta-window-fold-batch` (default 1).

`resolve`, `floor` and `capture_patch` are for the state service (ENG-6293) and have no
consumer yet.

## Behaviour changes

- `get_block_commit_status` no longer reports the oldest buffered block as `Committed`. That
  block is not in the database yet, so the old answer sent the read somewhere that could not
  serve it. This value decides whether a read hits the database alone or merges window deltas
  on top — review it first.
- An insert error resets that extractor's window instead of ending the pump and freezing the
  buffer for the pod.
- `pending_deltas_buffer_size` grows. Prod today: Ethereum 7.3 MiB, Base 12.5 MiB per pod.
- `get_new_components` reads only blocks above `db_committed`; window and database rows now
  overlap by up to `depth` blocks.
- `search_block` can find committed blocks, so more version lookups resolve from memory.

The `PendingDeltasBuffer` trait keeps its five signatures and the `rpc.rs` mock tests are
untouched.

## Notes for reviewers

Unwired items carry `#[allow(dead_code)]` until ENG-6293. `ReorgBuffer::get_commit_status` and
`PendingDeltas::new` are now test-only and carry the same; the former keeps the rstest case
that pins the buffer-bound behaviour the window deliberately diverges from.

Rollback without a new image: `--delta-window-depth 1` reproduces the old drain. On chains with
a large `--database-insert-batch-size` the `db_committed` term binds and depth has no effect
until it exceeds that batch.

## Follow-ups

ENG-6291 replaces `DiscardSink` with the entity cache. ENG-6293 consumes `capture_patch` and
takes ENG-6299's race criterion, which spans capture, lock release and a cache read and cannot
be exercised from the window alone. ENG-6384 sets per-chain depth. Observability should adopt
`delta_window_fold_duration_ms`; ENG-6310's table lists it without the unit suffix.


crates/tycho-indexer/src/extractor/reorg_buffer.rs
```diff
@@ -301,6 +308,26 @@ where
...
+    }
+
+    /// The buffered message for block `number`, or `None` if that block is not buffered.
+    /// Buffered blocks are contiguous, so this is an index lookup.
+    pub fn block_at(&self, number: u64) -> Option<&B> {
+        let first = self.oldest()?.block().number;
+        let index = usize::try_from(number.checked_sub(first)?).ok()?;
+        self.block_messages
+            .get(index)
+            .filter(|b| b.block().number == number)
```
zizou0x: Nice safeguard but we should be raising a warning here, it's not expected is it? It'd mean the buffer isn't contiguous anymore

crates/tycho-indexer/src/services/state/window.rs
```diff
@@ -0,0 +1,1063 @@
...
+/// Window of block deltas for one extractor.
+///
+/// Owns an RPC-side [`ReorgBuffer`] and decides retention: a block stays until it is finalized,
+/// committed, and deeper than `config.depth`. Not internally synchronized; the owner holds it
+/// behind a `Mutex`. Window and database overlap by up to `depth` blocks; readers that merge both
+/// must start window reads at `db_committed + 1` (see [`DeltaWindow::uncommitted_blocks`]).
+#[derive(DeepSizeOf)]
+pub(crate) struct DeltaWindow {
+    extractor: String,
+    buffer: ReorgBuffer<Arc<BlockAggregatedChanges>>,
```
zizou0x, line 154: We should add a TODO to rename this struct, it's not Reorg only. But I wouldn't do it in this PR because it's already big

crates/tycho-indexer/src/services/state/window.rs
```diff
@@ -0,0 +1,1063 @@
...
+        let db_committed = self.db_committed?;
+        Some(
+            finalized
+                .min(db_committed)
+                .min(tip.saturating_sub(self.config.depth)),
+        )
+    }
+}
+
+#[allow(dead_code)] // consumed by the state service, ENG-6293
```
zizou0x, line 385: Let's make sure we merge these with the above. Please add a "TODO" to remember.

zizou0x (approved):

crates/tycho-indexer/src/services/state/window.rs
brunoguerios: By merge you mean to get them together in the same `impl` block?
Yeah, I just separated them so it's easier to spot everything that is dead for now and requires future work.
Will add a TODO so we don't forget.

crates/tycho-indexer/src/extractor/reorg_buffer.rs
brunoguerios: Yeah, it's indeed not expected. Claude took a bit further and said it should actually be an error and we should propagate it so it's considered fatal. I'll log an error for now and we can decide on refactoring into a Result later. Let me know your thoughts.

PR #1460, opened 2026-09-18
feat: integrate Everlong FLAMM on Base

alex-alra-arteaga:
### What this adds

- Adds Everlong FLAMM on Base (chain 8453) as a native Tycho liquidity source: `protocol_system = flamm`, `protocol_type_name = flamm_pool`, two components per pool (the swap venue and the lever-up venue).
- Adds the substreams package `protocols/substreams/base-flamm` (five modules, `initialBlock 51154966`, version 0.1.0 with a `CHANGELOG.md`, the 1.96.0 toolchain pin and `tycho-substreams 0.8.1` like `base-lunarbase`). It streams every storage word a FLAMM quote reads as component attributes: the pool's own contracts, the Morpho Blue market and position it borrows against, the interest-rate model and four Chainlink feeds.
- Adds the `ProtocolSim` `crates/tycho-simulation/src/evm/protocol/flamm::FlammPoolState` (22 modules, one per contract or library, each function cited to its Solidity line). It reproduces the pool's fills from the attributes to the wei at the execution block, with no VM and no DCI.
- Registers the protocol where a native integration is registered: `protocols/testing/src/state_registry.rs` (`flamm` -> `FlammPoolState` with `flamm_filter`), `tycho-integration-test`'s Base defaults, `filters::flamm_filter` (re-exported next to `ekubo_v3_extension_filter`) with `flamm` in `EXCHANGES_REQUIRING_FILTER`, and the `flamm` row of `docs/for-solvers/supported-protocols.md`.
- Adds the test data the simulation and package tests replay (59 gzipped Solidity-generated parity fixtures, three chain snapshots, the real Base blocks of the end-to-end replay, and the scripts that recorded them) and amends the unit-test job's nextest filter so the `evm::protocol::flamm` tests run in CI (they read committed fixtures and need neither RPC nor anvil).
- `FLAMMExecutor.sol`, a `ProtocolWillDebit` executor that calls `FLAMM.swap` (venue 0) or `FLAMM.leverUp` (venue 1) with the pool as receiver and reverts unless the pool consumed the whole input.
- The `flamm` swap encoder, registered in `SwapEncoderRegistry` for Base, and `flamm` in `PROTOCOLS_NEEDING_APPROVAL`.
- `Executor::FLAMM` in the router security model.
- The protocol test harness's executor fixture (`FLAMM.runtime.json`, the `EXECUTOR_MAPPING` row, the `update_runtime_bytecode.sh` entry) and `skip_execution: false` for the swap component of `test_activation_and_first_swap`.
- The deployment declaration: `base.flamm` in `config/executor_deployments.json` and `flamm` in the Base list of `deploy-executors.js`. No production executor address is added until Tycho deploys and whitelists it (`executor_addresses.json` and `docs/for-solvers/execution/contract-addresses.md` are untouched).

The design document this PR implements (state sourcing, seeds, the read-set census, the VM measurements) is available on request; the parts the code depends on are in the package README, the schema comments in `words.rs` / `keys.rs`, and the fixtures.

### Why a native integration

FLAMM is a single-pool, financed-leg AMM (cbBTC / USDC). Every quote reads state the pool does not own (Morpho Blue, its interest-rate model, three Chainlink feeds and the L2 sequencer feed), and every sell borrows USDC on Morpho inside the fill. We built and measured the VM route before proposing this one; it does not work for four independent reasons, the first of which PropellerHeads named:

1. DCI cannot run on Base today (`tycho-indexer` `protocol_extractor.rs`: DCI's block-layer validation needs parent-hash linkage, which partial blocks break; `tycho-client` disables DCI queries on Base), and the public Base RPCs reject `debug_storageRangeAt`.
2. Tycho's token model breaks FLAMM sells: a sell borrows USDC on Morpho, which moves real USDC to the pool through the real token path, while the `TokenProxy` balance the VM overwrote does not move, so `MorphoBlueAccount` reverts its exact-delta check (`ExactDelta()`), and `EVMPoolState::get_balance_overwrites` keeps one holder per token.
3. The Morpho and Chainlink contracts predate FLAMM by years and would need synthetic bytecode and hand-picked storage in the package.
4. A Chainlink aggregator rotation or a FLAMM beacon upgrade would need an operator rewind of the extractor.

The native route has none of these. The external storage FLAMM reads is small and fully enumerated: five Morpho Blue words, one IRM word, and the words behind four Chainlink proxies, all either written on chain within minutes or announced by a storage change the package tracks. The math is a port of the Go simulator of the same pool (the KyberSwap integration), checked here against the same Solidity-generated fixtures, against the deployed pool's own `previewSwap` / `previewLever` answers and against every swap it has settled.

### The protocol and its contracts

Solidity source of truth: `EverlongLabs/blockend` @ `80abd43`; deployment record `script/flamm/c104/deployments/c104.8453.json`. Runtime codehashes below are `eth_getCode` at the creation block 51154990 and equal the record; the decoder pins the ones marked "pinned" and refuses a component whose static attributes name other code.

| Contract | Address | Runtime codehash |
|---|---|---|
| Pool `FLAMM` (beacon proxy; created 51154990, activated 51298416) | `0xc0fdCB1799cCc2CEBaA1fe247157b0dF33D57572` | proxy `0x5045564a…dbb63` |
| Implementation (beacon target) | `0xaAD580BeAa2cbd8Ab5F3956a5c56EDa1D5ee7184` | `0x2eb0fb32b59b1cca33e7bd6cad0a214bc6a293dc219d6ff9a5db1c58d0c10cb7` (pinned) |
| `FLAMMFactory` (beacon) | `0x1BfcE014774D0DD7e04bC595D46Fa09F7dCCF45f` | `0x96428fbb…d4c1` |
| `EverlongHook` (swap hook: curve, fee law) | `0x65CBD227cBC61248ae77a5fC813A29C54C092134` | `0x63ca81587b713df89dc9a657dae2cb5a70910cbafbe23b9ebbbdedc1bfdc3e1d` (pinned; also the factory's allowlisted invariant codehash) |
| `EverlongLeverageHook` (leverage venue) | `0xE0A98d8e60035832B8BaD7f7af7B9B0b3A7308F3` | `0xc6b46f4287cafa36c360234956b22dc635eb725dd17134a668939e731b819756` (pinned) |
| `LeverageSpreadHook` (keeper-posted spread) | `0x04988aF54ec88D2de77b191025EAef2fe488f93b` | `0x79d826c02ae2d2de5f8fdebfc1207731a04a96c0733e7af7c1f030b69d65a425` (pinned) |
| `MMRouter` (financing router) | `0x19A9b39E6710AAD109C829294b0841F0851c6bB4` | `0x6ca3c38096320c757612b113e543c37862f48bce5a2f19de2375357bc31ddcc8` (pinned) |
| `MorphoBlueAccount` (venue 0 account) | `0x6760E3b032eE2d670Cb684d9076b8f48cb066c48` | `0x7d5828b262882bb77ce568da078e381979429cc8a40d624e90e15a1895eee847` (pinned) |
| `PriceFeed` | `0xbED275459578c87A63F2F50A0B077c720e838816` | `0xe6054506…547f` |
| Morpho Blue v1.0.0 | `0xBBBBBbbBBb9cC5e90e3b3Af64bdAF62C37EEFFCb` | address pinned |
| `AdaptiveCurveIrm` v1.0.0 | `0x46415998764C29aB2a25CbeA6254146D50D22687` | `0x9978b522abfe0f3b8279800375d833b9d9660ae4f6321a2efb1f1f98850a0cbe` (pinned, address pinned) |
| `MorphoChainlinkOracleV2` (venue 0 oracle) | `0x663BECd10daE6C4A3Dcd89F1d76c1174199639B9` | immutables `SCALE_FACTOR`, `BASE_FEED_1` as static attributes |
| Chainlink cbBTC/USD proxy / OCR2 aggregator | `0x07DA0E54543a844a80ABE69c8A12F22B3aA59f9D` / `0x51cE3091cF646587e02cAd83b580992f8723e718` | |
| Chainlink USDC/USD proxy / OCR2 aggregator | `0x7e860098F58bBFC8648a4311B374B1D669A2bc6B` / `0x68bE4c50235205EDE361ac8244b1Ee221CDDa5E2` | |
| Chainlink BTC/USD proxy / SVR `DualAggregator` (Morpho's oracle feed) | `0x64c911996D3c6aC71f9b455B1E8E7266BcbD848F` / `0xe5eC87A39445b8d5B751b116802A53C5AE7e9DF1` | |
| L2 sequencer uptime proxy / feed | `0xBCF85224fc0756B9Fa45aA7892530B47e10b6433` / `0x606c6ECbD272E2174F6710b5974f23FE9899602E` | |

Tokens: cbBTC `0xcbB7C0000aB88B473b1f5aFd9ef808440eed33Bf` (8 decimals, the pool asset), USDC `0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913` (6 decimals, loan asset 0). Morpho market `0x9103c3b4e834476c9a62ea009ba2c884ee42e94e6e314a26f04d312434191836` (USDC / cbBTC, LLTV 0.86).

The pool's history on Base so far: created 51154990, activated 51298416, first settled swap 51302916 (tx `0x46c3cd72…901fa`, 15000 sats -> 11301759 USDC), five more swaps (51343234, 51347390, 51420672, 51420867, 51430828), six deposits, four withdrawals, keeper recenters, and leverage unpaused at 51433699 (the curator Safe calling `setLevPaused(false)`); the keeper has not re-posted a spread since, so every lever-up still reverts `SpreadUnavailable` on chain (see "Known limitations").

### State sourcing

Everything a quote reads, and where the package gets it (the full word-by-word schema, with byte offsets, the Go field each word feeds and the Solidity line, is reproduced in `words.rs` of the simulation and `keys.rs` / `feeds.rs` of the package, and the package README has the summary table):

| State | Words | Source on chain | Emitted as |
|---|---|---|---|
| FLAMM-owned storage: pool namespace (`FLAMMStore.S`, ERC-7201, 32 words + `loans[0]` + ERC-20 `_totalSupply`), `EverlongHook` (17), `LeverageSpreadHook` (1), the pool's `MMRouter` record (22), `MorphoBlueAccount` (7), `PriceFeed` token config (4), `FLAMMFactory` (4) | 87 in the schema; the forwarded key sets are supersets (the whole pool namespace, up to 8 loans / 16 venues / 4 order words) | contracts created in blocks 51154978-51154990, all inside the indexed range (`initialBlock 51154966`) | raw 32-byte words `<role>:<slot>` (`pool:` / `hook:` / `spread:` / `router:` / `account:` / `pricefeed:` / `factory:`), exactly the value `eth_getStorageAt` returns |
| Morpho Blue `market[id]` (3 words), `position[id][account]` (2), `AdaptiveCurveIrm.rateAtTarget[id]` (1) | 6 | pre-existing contracts; keys derived from the market id and the account address | raw words `mm:0:market:{0,1,2}`, `mm:0:position:{0,1}`, `irm:0:rate_at_target`, seeded at 51154965 |
| Chainlink cbBTC/USD, USDC/USD (behind `PriceFeed`), BTC/USD (behind the Morpho oracle), the sequencer uptime feed | the proxies' rotation word (slot 2) and access controller (slot 5); the aggregators' read-access pair and the words `latestRoundData` reads (`HotVars`, `s_transmissions`, `s_feedState`, `s_cutoffTime`) | the aggregators' storage, seeded at 51154965 | decoded `feed:<f>:{aggregator, phase, access_controller, check_enabled, access_list, kind, round, answer, started_at, updated_at}` per role `asset` / `loan0` / `seq`, and for `mo0` `round`, `secondary_round`, `cutoff` and a `feed:mo0:tx:<round>` ring of the 21 rounds the `DualAggregator`'s secondary-path reveal can answer with |
| Immutables (`LOAN_SCALE`, the hook's genesis hashes, `SEQUENCER_FEED` / `SEQUENCER_GRACE`, `UPGRADE_DELAY`, the oracle's `SCALE_FACTOR` / `BASE_FEED_1`, the `DualAggregator`'s secondary proxy and sync depth, the IRM codehash) | 13 | bytecode | 38 static attributes per component, from the `PoolCreated` log and `createPool` calldata, the creation transaction's storage writes, the package's deployments registry (codehashes) and the manifest `immutables` |

Two decisions to flag:

- **Rotations are storage changes, not events.** Base's `EACAggregatorProxy` (v0.6) emits nothing on `confirmAggregator` (the three historical rotations at 44229243 / 50530334 / 50837554 rewrote slot 2 with zero proxy logs), so the package tracks the proxies' slot 2 as a storage change and derives every feed attribute from the aggregators' storage words (`feeds::feed_state`), the words they write before emitting their events; the events are decoded only in tests as a cross-check of the layouts. A rotation to an aggregator the manifest does not list leaves the feed with `aggregator`, `phase` and `access_controller` only, and the decoder refuses to quote until a package update lists and seeds it (fail closed, no rewind).
- **Seeds are manifest params.** 48 `address:slot:value` words at `initialBlock - 1 = 51154965` (the Morpho market and position words, the IRM rate, each proxy's slots 2 and 5, the guarded aggregators' `checkEnabled` / `s_accessList[proxy]`, the OCR2 `HotVars` and latest transmission, the uptime feed's `s_feedState`, the `DualAggregator`'s `HotVars`, cutoff and 21-round ring), plus the per-pool `immutables` and a `deployments` registry of nine `role:codehash` pairs. Every value is a chain fact with a fixture and a unit test that fails if the manifest drifts (`testdata/seeds_51154965.json` cross-checks each seed with its view: `aggregator()`, `phaseId()`, `accessController()`, `checkEnabled()`, `hasAccess(proxy, "")`, `latestRoundData()`, `getRoundData(r)`, `Morpho.market(id)`, `rateAtTarget(id)`). This is the one place the state does not come from the stream, and the first open question below.

A FLAMM-owned word that is absent was never written and reads as zero, except the words the pinned code writes non-zero at construction (the hook's tuning rows, slots 4 to 6, and its reservation price; the pool's dials, fee bounds, loan band and share supply), which the decoder requires; an absent Morpho / IRM / feed word is unknown and the decoder refuses to quote.

Balances (TVL) are the pool's tradable inventory as the contracts count it: pool asset `= physicalPoolAsset + Σ min(position.collateral, managedCollateral)`, loan asset `= liquid + recognized Morpho supply` at the stored market totals. They are not a quote input; the range test sets `skip_balance_check` (the pool's ERC-20 balances are not its inventory, a sell borrows on Morpho).

### Component model

Two components per pool with the same attribute set and tokens `[pool asset, loan asset 0]`:

| Component | id | Quotes |
|---|---|---|
| swap venue | the pool address, `0xc0fdcb1799ccc2cebaa1fe247157b0df33d57572` | `FLAMM.swap`, both directions |
| lever-up venue | `pool || 0x00000000 || uint64(1)`, `0xc0fdcb1799ccc2cebaa1fe247157b0df33d57572000000000000000000000001` | `FLAMM.leverUp`, pool asset in only |

`component_kind` (0 / 1) is a static attribute and must agree with the id's shape. Lever-down is not a component: it pulls only its net pay leg (about half the input by construction, `FLAMMLeverLib.sol:135-140`), which the router would strand. The components' `contracts` list is empty: the indexer resolves that list against accounts the stream created and a native integration creates none, so the pool's contracts are address-valued static attributes instead. No reserved attribute name is used.

`map_components` emits a pool only if its invariant hook's codehash is allowlisted (`hook_codehashes`), the manifest carries its immutables, every one of its contracts was created with registered code inside the range, and the manifest tracks its external words; a pool that fails any of these is refused with the reason logged. The five modules (`store_deployments` -> `store_words` -> `map_components` -> `store_pools` -> `map_protocol_changes`) are pure functions of block + params, which is what makes them safe under Base partial blocks.

### Simulation

`crates/tycho-simulation/src/evm/protocol/flamm/`: 22 source modules (11.6k lines), one per contract or library (`almcurve`, `fee`, `hook`, `morpho`, `irm`, `account`, `router`, `gate`, `levcurve`, `levhook`, `pricefeed`, `state`, `swap`, `lever`, `context`, `deps`, `error`, `math`, `words`, `feeds`, `decoder`, `sim`), every function cited to its Solidity file and line, every refusal the revert the chain would raise (`FlammError`, one variant per deployed custom error with the selector table), Solidity's operation order and rounding kept throughout (the port does not reuse `safe_math` because the revert class is part of exactness).

The `ProtocolSim` contract as implemented (`sim.rs`):

- **`get_amount_out(amount_in, token_in, token_out)`** runs the pool's whole transaction on a copy of the composed state at the execution clock: the priced context, the plan with its funding passes, the hook fill, the validation and the settlement through the Router and Morpho, exactly as `FLAMM.swap` / `FLAMM.leverUp` would at `block.timestamp == clock`. It returns the amount and the post-state (the pool after that transaction, including the Morpho position and market it touched). Only a size the pool fills in full is a quote: a size the pool would clip is refused as a partial fill naming the used amount (`FLAMMExecutor` reverts on any partial fill). A zero input is the empty trade with the venue's base gas, and so is a buy's dust (see `get_limits`).
- **`get_limits(sell, buy)`** returns the largest size the venue fills in full in that direction and its output, found by doubling from one base unit and bisecting on full fills (no RPC). The contract, exactly: every size the venue fills in full lies in `[1, limit]`; a size above `limit` is refused (a partial fill or the pool's own revert); inside `[1, limit]` every size answers, a fill or, for the dust the pool refuses, the empty trade. A sell of the pool asset fills from one base unit at every recorded block. A buy is dust below a few thousand USDC base units (a payout that rounds to zero sats is `FillInvalid`; a payout of one to six sats whose quantization leaves the band around the checked cross is `PriceBand`); the refused sizes are not an interval, and at every recorded block the largest refused buy is below a hundredth of a percent of the limit, so the sizes a consumer derives from the limit (the protocol test harness quotes 0.1 %, 1 % and 10 % of it) fill in full. `(0, 0)` for a direction the venue does not trade (the lever-up venue's loan asset in) and for a state that fills no size. With this, `[0, limit]` is total and `query_pool_swap`'s generic search runs on both venues.
- **`spot_price(base, quote)`** follows the trait: the `quote` that buys one `base` at zero size, gross of the fee of the direction that buys `base` (`add_fee_markup`'s form). The swap venue prices both orderings at the hook's exact `spot()` (the reservation curve at the stored coordinate), each with its own direction's fee (they differ by the direction skew). The lever-up venue has no closed-form spot (the frozen `CollRebalancerMath` curve) and one rate, the loan asset it pays per pool asset at negligible size, read off a fill of 64 times the smallest size it fills in full (halving back if that does not fill); it answers both orderings from that rate: `spot_price(USDC, cbBTC)` is the reciprocal gross of the spread (the trait's ordering), `spot_price(cbBTC, USDC)` the rate itself (the ordering `query_pool_swap` and the harness read for the one direction the venue trades).
- **`fee()`** is what a fill pays right now: for the swap venue the sell-direction bounded `previewFeeWad` (the hook's fee law reads the direction and the book, never the size), for the lever-up venue the live spread as `FLAMMLeverLib._spread` clamps it (`[2500 ppm, min(swapPriceBandWad / 1e12, 100000 ppm)]`); `1.0` when nothing can fill (no live spread, a fee the floor refuses, an unreadable state).
- **`delta_transition`** applies the streamed updates and deletions to the attribute map and re-decodes. A value of the wrong width is a `TransitionError`; an attribute set that no longer decodes (a feed's rounds deleted by a rotation, a pin no longer met) leaves the component alive and refusing until a later delta restores it. The decoder's `block_number` / `block_timestamp` are honoured.
- **`apply_block(BlockContext)`** moves the execution clock to the block's timestamp and re-evaluates the Morpho oracle's SVR reveal there; Morpho's accrual, the IRM's adaptation, the feeds' heartbeats, the sequencer grace and the spread's age are then evaluated at that clock by every quote. It returns whether a quote can observe the move: a deadline crossed (a feed's heartbeat, the sequencer grace, the spread's `maxSpreadAge`, the oracle's reveal cutoff, a scheduled change's `executableAt`, the IRM grace, the rate ceiling's verdict flipping) or a pool positioned in its venue, whose debt and supply accrue every second, so for the live pool, which holds a position, the flag is true on every new block timestamp and the state is re-emitted every block; an unpositioned pool is quiet between deadlines. A repeated timestamp (the flashblocks of one block) is a no-op.
- **Gas** is the Go port's measured defaults (largest receipt gas of the adapter's fills on Base forks plus 25 %, rounded up to 10,000): sell 1,360,000 plus 10,300 per cap-solve and 210,000 per funding pass after the first, buy 1,510,000, lever-up 3,670,000. `flamm` is in `PROTOCOLS_NEEDING_APPROVAL` (the pool pulls the input with `transferFrom`).
- **Refusals** are typed: a state that cannot be decoded or has drifted from its pinned identity (`DecodeError::{Missing, Malformed, Pin, Drift, Unsupported}`), a venue whose IRM or oracle cannot be read, a quarantined venue, a scheduled implementation or hook-set change that is executable, and the pool's own reverts, all surface as `SimulationError`s, never as a guessed amount.
- **`flamm_filter`** (exported from `filters.rs`, registered in `protocols/testing`'s registry and `tycho-integration-test`'s Base defaults; `flamm` is in `EXCHANGES_REQUIRING_FILTER` so a stream without it warns) skips at snapshot time a component the decoder refuses instead of failing the stream.
- `clone_box` / `eq` / `as_any` and typetag serde (a JSON round trip reproduces the state, the leverage port included).

### Execution

Why the executor refuses partial fills: `TychoRouterV3` strands any input a `ProtocolWillDebit` executor does not consume (`_transferFromUser` pulls the nominal amount, the Dispatcher only resets the allowance, `_finalizeBalances` checks transient deltas and nothing sweeps; reproduced on a fork: 400 of 1000 stranded). FLAMM can clip a sell (gate room, notional cap, funding ceiling), and a lever-down pulls only its net pay leg, about half the input by construction (`FLAMMLeverLib.sol:135-140`). So the simulator quotes only sizes the pool fills in full, the executor asserts it, and lever-down is not a venue until FLAMM ships an exact-pay entrypoint (a beacon upgrade through `FLAMMFactory.scheduleImplementation`).

**The executor.** `crates/tycho-execution/contracts/src/executors/FLAMMExecutor.sol` (BUSL-1.1, Everlong copyright), verified against `EverlongLabs/blockend` @ `80abd43`:

- Swap data is `pool (20) || tokenIn (20) || tokenOut (20) || venue (1)`, 61 bytes exactly; any other length reverts `FLAMMExecutor__InvalidDataLength`, a venue byte above 1 `FLAMMExecutor__UnknownVenue` (also from `getTransferData`, which decodes the same way).
- `getTransferData` is pure: `ProtocolWillDebit`, receiver = pool, `outputToRouter = false` (the pool pulls `amountInUsed` from `msg.sender` with `transferFrom`, `FLAMMStore.sol:324-328`, and pays the receiver directly). `fundsExpectedAddress` = `msg.sender`. Transfer type, receiver and `outputToRouter` are hardcoded, not encodable.
- The factory is an immutable (zero rejected, `FLAMMExecutor__ZeroFactory`); `factory.isPool(pool)` is checked before any call (`FLAMMFactory.sol:59`; true for every beacon proxy the factory created, provenance rather than endorsement, since creation is permissionless) and a foreign address reverts `FLAMMExecutor__UnknownPool`.
- Venue 0: `pool.swap(tokenIn, tokenOut, amountIn, 0, receiver, block.timestamp)`, either direction (`FLAMMSwapLib.sol:163-172, 204-205`).
- Venue 1: requires `tokenIn == pool.asset()` and `tokenOut == pool.loanAsset()`, then `pool.leverUp(amountIn, 0, receiver, block.timestamp)` (`FLAMMLeverLib.sol:47-60, 101`). The reversed pair reverts `FLAMMExecutor__LeverDownUnsupported`, any other pair `FLAMMExecutor__InvalidLeverPair(tokenIn, tokenOut)`.
- Both venues revert `FLAMMExecutor__PartialFill(amountInUsed, amountIn)` unless the pool reports `amountInUsed == amountIn`.
- Security checklist: no ERC20 transfers, no approvals, no storage writes, no delegatecalls, no assembly, no callbacks, no caller-supplied calldata forwarded (the two selectors are fixed); `minAmountOut` is 0 (the router enforces the user's minimum). The `slither-disable-next-line locked-ether` annotation follows `LunarBaseExecutor`.

`CLAUDE.md` of the crate lists FLAMM among the direct-to-receiver executors.

**The encoder.** `src/encoding/evm/swap_encoder/flamm.rs`, registered as `flamm` in `SwapEncoderRegistry`, Base only (`new` rejects any other chain). A 20-byte component id is the swap venue (venue byte 0); a 32-byte id `pool || 0x00000000 || uint64(1)` is the lever-up venue (venue byte 1); any other length or discriminator is a `FatalError`. No protocol-specific constant address is needed (the factory is the executor's constructor argument). `flamm` is added to `PROTOCOLS_NEEDING_APPROVAL` (`gas_estimator.rs`): a `ProtocolWillDebit` executor is approved by the router before the pull, and that list is what `needs_approval` classifies by.

**The security model.** `model/src/model/executors.rs`: `Executor::FLAMM`, `VARIANTS` 13 -> 14, the LunarBase-style branch: `ProtocolWillDebit`, receiver = `data[0:20]` requested as `SENDER_CONTROLLED`, token in / out from the data, `outputToRouter = false`, `fundsExpectedAddress` = the router, callbacks unimplemented. The pool comes from calldata and creation is permissionless (the creator chooses the hooks the pool calls into), so the model treats the pool as caller-controlled code; the executor's own partial-fill revert is satisfied trivially by a sender-controlled pool, which is the case the model exercises. A pool that is not sender-controlled is `Ignore`d ("would require simulating real pool"), as the other executors with calldata-selected pools do.

`cargo run --release` in `model/` on this head: 97.3M simulations in 18 s, 0 suspicious, 0 warnings; FLAMM 682,616 successes, 0 suspicious. The one "all simulations reverted" note (`transfer_allowances_during_swap = true`) is pre-existing on `main` (baseline run on a pristine clone: 13 executors, 0 suspicious, the same note).

**The protocol test harness.** `protocols/testing`: `fixtures/FLAMM.runtime.json`, the `flamm` row of `EXECUTOR_MAPPING` (`src/execution.rs`), and `FLAMM|base|flamm` in `scripts/update_runtime_bytecode.sh`. The fixture was produced the way the script produces each executor fixture (deploy `FLAMMExecutor(0x1BfcE014…F45f)` on a local anvil, capture the runtime code, minify); for an executor whose only immutable is the constructor argument the result is independent of the fork. Re-checked on this head: `forge create` of `FLAMMExecutor(0x1BfcE014…F45f)` on a plain anvil and `cast code` of the result equal the committed fixture byte for byte, and the same procedure reproduces the committed `FluidV1.runtime.json`. The script itself needs an Ethereum mainnet `RPC_URL` for the router fixture, so it was not run here; the CI job `runtime-bytecode-check` runs `update_runtime_bytecode.sh --check` on an upstream PR.

The hosted harness will fork Base at the stop block 51302920 and run the quoted sizes (0.1 %, 1 % and 10 % of `get_limits` in each direction: 158168 sats and 180858003 USDC units are the largest full fills there) through the executor, the first time the executor is exercised by that harness (see "Not verified").

**Tests.** `contracts/test/protocols/FLAMM.t.sol`, on a Base fork at 51302915 (the parent of the pool's first settled swap), 23 tests, all green:

- `FLAMMExecutorTest` (17): decode and its refusals (length, venue), `getTransferData` / `fundsExpectedAddress`, the constructor's zero-factory revert and the immutable, the sell replay (15000 sats -> exactly 11301759 USDC, the chain's own fill in the next block), a buy of 10 USDC (`previewSwap`-exact sats), a lever-up of 10000 sats with the venue opened by pranks (the curator's `setLevPaused(false)`, the keeper's `setSpread(17500)`) exact against `previewLever`, lever-down and foreign-pair refusals, an unknown pool, partial-fill reverts on both venues via `vm.mockCall`, and the encoder calldata round trip through `loadCallDataFromFile`.
- `TychoRouterForFLAMMTest` (6, `TychoRouterTestSetup` with `getChain() = "base"`): the Rust-generated `singleSwap` calldata (sell, exactly 11301759), a buy and a lever-up through the router, a sequential buy-then-sell (the router-funded second leg), a partial fill bubbling through the router with nothing stranded, and `testExecutorAddressMatchesTestConfig`. Like `LunarBase.t.sol`, the executor is deployed in the test's own `setUp` after the shared one rather than in `deployExecutors` (the Base executors are not in the mainnet set); it is placed with `deployCodeTo` at `0x94771550282853f6E0124c302F7dE1Cf50aa45CA`, so no deployment the shared setup adds or removes moves it. The three places that carry the address (the test constant, `base.flamm` in `config/test_executor_addresses.json`, which the Rust integration test encodes into, and the `test_single_encoding_strategy_flamm` line of `test/assets/calldata.txt`) are named in the test and checked against each other, and the test also checks the code at the address and its factory. The activation timelock is served in the past the way the shared setup does it, so the fork's Chainlink rounds stay fresh for the pool's price feed.

Rust: the 6 encoder unit tests (the three layouts, which they write to `calldata.txt`, and the three rejections: unknown discriminator, other id lengths, a non-Base chain) and `tests/protocol_integration_tests.rs::test_single_encoding_strategy_flamm` (the router calldata). `calldata.txt` gains exactly those 4 lines.

Gas on the fork at 51302915, as the tests log it:

| Path | Gas |
|---|---|
| `FLAMMExecutor.swap`, sell 15000 sats, direct call | 734,190 |
| `FLAMMExecutor.swap`, buy 10 USDC, direct call | 735,169 |
| `FLAMMExecutor.swap`, lever-up 10000 sats, direct call | 1,766,344 |
| `TychoRouterV3.singleSwap`, sell 15000 sats | 1,071,782 |
| `TychoRouterV3.singleSwap`, lever-up 10000 sats | 1,872,147 |

The cost is the pool's: a sell borrows on Morpho Blue inside the fill, a lever-up prices on the leverage hook. The simulation's gas constants (sell 1,360,000 plus the cap-solve and funding-pass slopes, buy 1,510,000, lever-up 3,670,000) are the Go port's measured defaults over the adapter's fills on Base forks plus 25 %, above every figure here.

Run with `BASE_RPC_URL` set, from `crates/tycho-execution/contracts/`: `forge test --match-path test/protocols/FLAMM.t.sol`.

### Range test configuration

`protocols/substreams/base-flamm/integration_test.tycho.yaml` has two range tests, both starting at `initialBlock` (the contracts are created in the range, so no `initialized_accounts`):

- `test_pool_creation` (51154966-51155010): the creator's deployments and the pool's creation; both components expected with their 38 static attributes. The pool is born paused, so `skip_simulation` and `skip_execution` are set on both components with the reason in the yaml.
- `test_activation_and_first_swap` (51154966-51302920): creation, the curator's activation (51298416) and the first settled swap (51302916). The swap component simulates and executes (`skip_simulation: false`, `skip_execution: false`). The lever-up component keeps both skips (leverage is paused on chain at the stop block), with the reason in the yaml.

`skip_balance_check: true` is set at the top level with the reason (the components' balances are the pool's inventory across Morpho, not the pool contract's ERC-20 balances), and the package's `e2e_yaml_skip_flags_are_truthful` test asserts that every skip is justified by the indexed pause bits and that execution is skipped exactly where simulation is.

The second test streams ~148k Base blocks, which is far above the "ideally below 1,000 blocks" guidance of the testing docs: the pool sat paused for ~143k blocks between its creation and its first quotable state, and a native integration cannot shorten the range with `initialized_accounts` (the components' creation must be witnessed). Open question 5 below asks how you would like this handled.

### Testing

All numbers are from runs on this branch's head, on the CI-pinned toolchain. Every replay is row by row, to the wei and by revert class, with no sampling and no tolerance; 0 mismatches everywhere.

**Solidity-generated parity fixtures** (`testdata/`, 59 gzipped files, 5.9 MB, byte-identical copies of the Go port's Foundry-generated fixtures at c104 `80abd43`, each with its generator, stored and uncompressed sha256 in `testdata/README.md`; `tests/fixtures.rs` pins every stored digest and `every_fixture_is_pinned_and_present` fails if a file is added, dropped or changed):

- Swap hook: `alm_curve_grid` 12,224 rows (swap 11,049, reserves 423, price 352, y 328, support 72); `fee_fill_grid` 6,581 fills, 600 `lnWad` and 300 volatility rows; `hook_fill_grid` 9,971 contexts (4,195 filled, 368 reverted, 26 spots); `hook_live_swap`, the real 15000-sat sell (fee `17499999999999999`, 11301759 USDC, committed book equal to chain storage); edges `alm_curve` 3,308 / `fee` 3,200 / `hook_fill` 7,724 (1,258 filled) / `hook_fill_loan_scale` 2,020 / `hook_live_tx` 4.
- Financing: `mm_muldiv_edges` 3,775; `mm_irm_grid` 1,345; `gate_math` 400 books; `gate_int_edges` 20; `mm_live_views` 11 snapshots of every router / account / Morpho / pool view; `mm_real_sell` (the block-51302916 swap settled from the in-block pre-state); `mm_live_settle` 8 steps; `mm_multi_venue` 42 steps; edges `mm_morpho` 566 / `mm_irm` 661 / `mm_account` 459 / `mm_router` 1,701 / `gate` 515 / `mm_settle` 186 / `mm_market` 1,183; `router_settlement_edges` one loan 3,745 ops / two loans 5,610; stateful sequences `router_a` / `router_b` 520 steps each (15,080 view rows each), `router_liquidation` 20, `swap_settlement_sequence_a-d` 600 steps each and `_e` 60 (the carried ledger compared before and after every step), `mm_accrual_grid` 900.
- Leverage: `lev_curve_tape_v1` 22,465 rows over 16 blocks (manifest and per-block digests re-verified); `lev_curve_fork_fixture` 6,100 leverage + 6,100 deleverage + 2,680 anchor rows and the 16 frozen parameters; `lev_hook_local_fixture` 609 rows (411 fills / 198 reverts); `lev_hook_fork_fixture` 402 (221 / 162); `lev_hook_band_fixture` 910; `edges/lev_curve_edges` 76,994 of 77,002 compared (the 8 private-helper rows the Go test skips too).
- Pool core over the composed state (`FLAMMSwapLib`, `FLAMMLeverLib`, `PriceFeed`): `core_e2e_grid` 34,052 / 25,070 / 23,501 checks at 51302915 / 51313000 / 51324800; `core_e2e_seq` 200 / 199 / 199; `core_edge_grid` 13,372 / 12,876 / 12,863; `core_edge_seq` 3,838 / 3,901 / 3,777; every preview word, every executed return and event word, every post-state field, and the sensitivity runs (grace, heartbeat, peg band, max price, spread, max age, pause bits, features each break rows).

**The deployed pool's own answers** (`tests/protocol_sim.rs`, component snapshots and `eth_call` grids recorded from `mainnet.base.org` at three blocks):

- The decoded snapshot at 51302915 equals the Solidity-generated `core_e2e` state at the same block field for field, oracle at the clock included.
- 622 `previewSwap` rows over 51302915 / 51313000 / 51409000: 322 full fills quoted to the wei with the venue's gas, 300 reverts (85 of them at or below the limit, quoted as the empty trade; 215 above it, refused with the recorded class); `get_limits` equals the chain's own bisection edge and its output in both directions at all three blocks (sells 179237 / 140550 / 452225 sats, buys 165217404 / 185398907 / 390516692 USDC units; edge + 1 refused); `swap_fee_wad` equals every recorded `feeWad`; the hook's `spot()` equals the recorded spot word.
- The lever-up venue on the Solidity-generated `armed` scenario (the chain has never offered a fill-able state): 8,855 rows, 632 lever-up fills quoted to the wei, 1,874 lever-up refusals, every lever-down row refused; its limit and spot. On the `capped` scenario: 991 full fills, 67 chain-clipped rows refused as partial fills naming the used amount, 673 typed refusals above the limit, 2,095 dust rows below it quoted as the empty trade; limit + 1 is a partial fill.
- The real swap from the parent snapshot: decode 51302915, `apply_block(51302916)`, 15000 sats -> 11301759 USDC, gas 1,360,000; the quote's post-state equals the delta-applied chain state to the wei (physical `0x32d4a`, the hook book, managed collateral `0x666d`, the Morpho borrow totals and position, `rateAtTarget 0x57fee5c5`) after netting a third party's 90,140,614-unit withdrawal in the same block.
- `apply_block` is deterministic and idempotent (a repeated timestamp is a no-op; a heartbeat boundary is inclusive; a stale feed refuses `StalePrice` and `fee` is 1.0); the `DualAggregator` reveal follows the clock (rounds 3583 / 3582 / 3581; an absent visited ring round refuses); rotation deletions leave a live, refusing state that recovers on restore; a malformed width is a `TransitionError`; a pending implementation refuses at `executableAt` and an executed upgrade refuses on the pin; every missing attribute, pin, kind, drift, venue-count, token and id-shape case fails closed at snapshot time; `flamm_filter` admits exactly what the decoder accepts.

**The substreams' output, end to end** (`base-flamm/src/e2e_tests.rs` and `flamm/tests/e2e.rs`, the local stand-in for the hosted range harness): `testdata/e2e_blocks.json.gz` (204 KB) holds real Base blocks assembled from `eth_getStorageAt` / `eth_getTransactionReceipt` / `eth_getCode` calls against `mainnet.base.org` (no `debug_*`): the creator's deployments 51154978-51154988, the creation 51154990, both stop blocks of the range test, the activation, every settled swap, two deposits, a withdrawal, a keeper recenter, three recent Chainlink-round blocks and the leverage unpause 51433699, with the net diff of the blocks between them as synthetic catch-up blocks. An `Indexer` drives the five module cores in manifest order with substreams store semantics and folds the changes as `tycho-storage` holds them and `tycho-client` delivers them (`ProtocolComponentStateDelta::merge`); the fold is `snapshots/e2e_stream.json.gz`, which the package test asserts equals its own output and which the simulation test replays as `TychoStreamDecoder::decode` would (a snapshot at the creation header, a `delta_transition` per block, `apply_block` to the next block's clock):

- the words store equals `eth_getStorageAt` for every tracked word (359 to 360 of them) at every pinned block; the fold at 51302915 equals the schema snapshot (147 attributes and balances) with only `feed:<f>:kind` and the forwarded-range words extra; the deployments store holds exactly the 9 registered contracts; the creation block emits both components with the ids, tokens, 38 static attributes and creation tx of `integration_test.tycho.yaml` (parsed by the test), born paused and `levPaused`, 143 snapshot rows; every deletion names a held row and every first write is a `Creation`; the pause bits justify every skip flag in the yaml;
- 42 stream blocks; both components decode at every one; at the 11 pinned blocks, 2,142 `previewSwap` rows at the block's own clock (1,194 full fills quoted with the recorded `amountOut` and `feeWad`, 948 reverts refused with the same class, 0 clipped), `get_limits` equal to the chain's edges in both directions at the 10 unpaused blocks (e.g. 158168 sats / 180858003 USDC units at 51302920, 7678229 / 6103810726 at 51433218), the hook's spot equal, 83 `previewLever` rows refused (`LevPaused` through 51433218, `SpreadUnavailable` / `NothingToFill` / `PriceBand` at 51433699);
- all 6 settled swaps reproduced from the parent state at the swap block's clock to the wei (15000 -> 11301759, 20000000 -> 25653, 40000 -> 30259159, 13508629 -> 17192, 508629 -> 646, 6550 -> 4870774) with the receipt's `feeWad`, full fills where the calldata gives the request, and the quote's post-state equal to the block's diff for the pool words, the hook, the feeds, the Router legs, the position, the IRM rate and the whole Morpho market once the other transactions' Supply / Withdraw / Borrow / Repay events of the same block are added;
- the harness's step at both stop blocks (limits, spot, 0.1 % / 1 % / 10 % quotes at stop + 1): nothing tradable at 51155010 (`skip_simulation: true` is required there), both directions at 51302920, the lever venue's limits zero everywhere.

The recorders of every RPC-derived fixture (`schema.py`, `rpc.py`, `reads.py`, `snapshot.py`, `keys.py`, `universe.py`, `grids.py`, `package.py`, `fetch.py`, `morpho_events.py`, `pack.py`, `regen.sh`) are committed under `testdata/gen/` with a README and per-script sha256; `package.py` reproduces the three snapshot fixtures byte for byte and `pack.py` reproduces `e2e_grids.json.gz` and `e2e_blocks.json.gz` (a full re-fetch takes about 80 minutes under `mainnet.base.org`'s rate limit; `regen.sh` re-packs and re-pins without re-fetching).

**Package unit tests** (`cargo test -p base-flamm`, 51 tests): the factory ABI against the real creation log (topic, selector), the manifest params against the fixtures (48 seeds, 9 codehashes against `eth_getCode` at 51154990 and the deployment record, 13 immutables against their getters), every seed against its view, the real `createPool` input and `PoolCreated` decode, `store_deployments` on real code, the creation-block replay (both components' static attributes equal to the schema snapshot; refusal on an unlisted hook, missing immutables, an unregistered router), the creation snapshot, the first swap's 12 changed words as updates on both components (inventory after `= physical before + 15000 sats`), the store filter, one real round of each aggregator decoded from its events against the aggregator's storage after the block (OCR2 rounds 14785 / 26, the sequencer's status round at 47851108 and refresh at 47894326, the `DualAggregator`'s round 3583 with the ring eviction of 3562 and a `SecondaryRoundIdUpdated`), synthetic proxy rotations to listed, unlisted and unseeded aggregators and back, and multi-block histories replayed against the indexer's storage rule (`verify_tests.rs`).

**Not run here:**

- The hosted range harness (`protocols/testing`, the two tests above, with execution enabled) needs a StreamingFast token, a Postgres and `tycho-indexer`, none of which this environment had, so the executor has not yet been exercised by that harness (its behaviour on the same fork block and sizes is covered by `FLAMM.t.sol`). The local replay above pins what the package emits, so a hosted-run difference is localised to the Firehose stream versus the recorded storage diffs. The runbook (prerequisites, the `cargo run -- range --package base-flamm --chain base` and docker routes, env vars, the expected output per test and what a failure at each step means) is in `protocols/substreams/base-flamm/README.md`, "The range test". The yaml's expected components are the values the package emits in the local replay, not harness output.
- The lever-up venue has never been fill-able on chain (paused through 51433698, `SpreadUnavailable` since), so its `ProtocolSim` path is proven on the Solidity-generated `armed` scenario, not on chain answers.
- Same-block visibility of upstream store updates inside the pool's creation block (`store_deployments` -> `store_words` -> `map_components`) is relied upon per substreams semantics and exercised only with in-memory stores.
- `update_runtime_bytecode.sh --check` end to end (needs an Ethereum mainnet `RPC_URL` for the router fixture; the FLAMM fixture itself is reproduced above).

### CI

Every CI job's command was run locally on the head commit with the workflow's pinned toolchain (`nightly-2026-06-28`, `.github/workflows/ci-rust.yaml`); the Rust and substreams lines are from the indexing and simulation commits, the execution lines from the executor commits:

- `cargo +nightly-2026-06-28 fmt --all --check`: clean.
- `cargo +nightly-2026-06-28 clippy -p tycho-simulation -p protocol-testing -p tycho-integration-test --all-targets --all-features -- -D warnings`: clean (the stable toolchain agrees).
- `RUSTDOCFLAGS='-D rustdoc::broken_intra_doc_links' cargo doc -p tycho-simulation -p protocol-testing -p tycho-integration-test --no-deps --all-features --locked`: exit 0; the 33 rustdoc warnings it prints are all pre-existing on `main` (curve, uniswap_v4, cowamm, balancer_v3, vm and titan files), none in `flamm` and none in a file this PR touches.
- `cargo test -p tycho-simulation --all-features flamm`: 170 passed, 0 failed (dev profile); `cargo test -p tycho-simulation --all-features --lib stream`: 119 passed.
- The unit job's nextest filter, amended in this PR from `test(evm)` to `(test(evm) and not test(flamm))`: `cargo nextest run -p tycho-simulation --all-features --locked -E '…'` 445 tests run, 445 passed (168 of them flamm; 0 before the amendment); the same filter on `-p tycho-integration-test -p protocol-testing`: 48 passed. If you would rather keep the filter as it is, the amendment is one line and the flamm tests still run locally.
- `protocols/substreams`: `cargo +nightly-2026-06-28 fmt --package base-flamm -- --check` clean; `cargo +nightly-2026-06-28 clippy --package base-flamm --all-targets -- -D warnings` clean; `cargo test --package base-flamm` 51 passed; `cargo build --package base-flamm --target wasm32-unknown-unknown` OK (dev and release); `substreams pack base-flamm.yaml` builds the package and `substreams info` on it lists the 5 modules at initial block 51154966.
- Every commit of the series builds on its own (`git rebase --exec` with `cargo check -p tycho-simulation --all-targets`, plus `protocol-testing` / `tycho-integration-test` at the registry commit and `base-flamm` for wasm and native targets at the package commit).
- `cargo +nightly-2026-06-28 clippy -p tycho-execution -p tycho-router-model -p protocol-testing --all-targets --all-features -- -D warnings`: clean (stable agrees).
- `RUSTDOCFLAGS='-D rustdoc::broken_intra_doc_links' cargo doc -p tycho-execution -p tycho-router-model -p protocol-testing --no-deps --all-features --locked`: exit 0.
- `cargo test -p tycho-execution --lib flamm`: 6 passed; `cargo test -p tycho-execution --test protocol_integration_tests test_single_encoding_strategy_flamm`: 1 passed; `cargo test -p tycho-router-model`: 4 + 2 doctests passed. (The four pre-existing permit2 / approvals lib tests need an Ethereum `RPC_URL`; unrelated.)
- `forge fmt --check src/executors/FLAMMExecutor.sol test/protocols/FLAMM.t.sol`: clean (the project-wide check reports `LunarBaseExecutor.sol` and `PropAMMFallback.t.sol`, both pre-existing on `main` with forge 1.7.1). `forge build`: OK, no lint lines for the FLAMM files. `BASE_RPC_URL=https://mainnet.base.org forge test --match-path test/protocols/FLAMM.t.sol`: 23 passed, 0 failed.
- `node --check scripts/deploy-executors.js`: OK, and every protocol in the Base list resolves to an `executor_deployments.json` entry.
- `protocols/substreams`: `cargo test --package base-flamm` 51 passed after the yaml flip; `fmt` and `clippy --all-targets -- -D warnings` on the pinned nightly clean.

The series is rebased onto `main` at 0.408.0 (`d9a64a29f`); the indexing commits rebased without conflicts, the execution commits met upstream's fallback encoder (#1434) in the `use` list of `swap_encoder_registry.rs` and the tail of `calldata.txt`, both adjacent insertions kept in full, and every command above was re-run on the rebased head. On the fork, the same series on 0.406.0 passed every workflow that needs no secrets: Lint (fmt + clippy), Unit tests (no external deps), Rustdoc, Check (no default features), DB tests, Cargo machete, Substreams Lint, Substreams Test, Slither static analysis (the `ci-foundry.yaml` job, on this executor), Validate PR title. The Foundry test jobs passed `forge fmt --check` and then failed on the fork for lack of RPC secrets (`vm.createSelectFork: invalid rpc url`), EVM tests (requires RPC) likewise, and Cargo Audit reported RUSTSEC-2026-0285 on `rustls 0.23.31`. That advisory is patched at `rustls >= 0.23.45`, which `main` carries since 0.406.2 (#1443) and this branch's root `Cargo.lock` inherits unchanged (its diff against `main` is the three `tar` crates below).

**Dependencies:** `tycho-simulation` gains the dev-dependencies `flate2`, `sha2` (both already in the lock) and `tar` (new: `tar`, `filetime`, `xattr`); `src/evm/protocol/flamm/testdata` is added to the package `exclude`. `base-flamm` gains the dev-dependency `flate2` (`flate2`, `miniz_oxide`, `crc32fast`, `adler2`, `simd-adler32`, `zlib-rs` in `protocols/substreams/Cargo.lock`).

### Known limitations

- **The lever-up venue is not quotable on chain today.** Leverage was paused from the creation through 51433698; since the unpause at 51433699 every `previewLever(true, .)` reverts `SpreadUnavailable`: the `LeverageSpreadHook`'s slot 0 still holds the constructor's 17500 ppm post (`lastSetTs 1789099323`, `maxSpreadAge 3600`) and no `SpreadSet` has been emitted since. The simulation reproduces exactly that (the venue refuses, `get_limits` is `(0, 0)`, `fee` is 1.0), and a keeper post would make it quotable without a code change; until then the venue's fills are proven on the Solidity-generated `armed` scenario only.
- **Lever-down is excluded** (see "Component model"), neither simulated nor executed; it waits for an exact-pay entrypoint on the FLAMM side.
- **Balances** value the managed Morpho supply at the stored market totals rather than accrued to `block.timestamp`, and take every venue as readable; they are not a quote input and the range test skips the balance check.
- **Time dependence.** Quotes are functions of the execution clock, which `apply_block` sets from `BlockContext.timestamp`. For the live pool, which holds a position, `apply_block` reports a change on every new block timestamp, so both components are re-emitted every 2-second Base block (never a wrong answer; the re-emit optimisation is defeated for a positioned pool). A Chainlink round landing between the quote and the fill can move the ±8 % band check (the Go port measured 567 cbBTC/USD rounds: median 9.7 bps, p99 52.1 bps); the SVR `DualAggregator` behind Morpho's oracle can change its answer with time alone, which affects only the Router's 2 % band and Morpho's health check, and the port evaluates its reveal at the clock. The scheduled-change guard refuses at `clock >= executableAt` (exact at the execution clock; the Go port refuses 1800 s ahead because its clock is not the execution block).
- **Dust and the cost of a refused quote.** A buy of a few thousand USDC units is dust the pool refuses; `get_amount_out` quotes it as the empty trade so that `[0, limit]` is total, and to classify a refused size it computes the venue's limit (a doubling-and-bisection scan of about 70 settlements, milliseconds in a debug build), which a solver probing beyond `get_limits` pays once per refused call.
- **The lever venue's spot is read off a fill** (64 times the smallest full fill), with a residual of the payout's rounding plus the curve's move over that size, bounded in the test (the one-sat rate within 1.5e-3, the max-size rate within 3 %, and the rate at most `feed × (1 − spread)`).
- **Seeds, immutables and codehashes are manifest params.** A further pool needs a package update carrying its `immutables` entry and, unless shared with the live pool, its external words, aggregators and addresses; a package update that changes `aggregators`, `addresses` or `words` must re-index the protocol from `initialBlock` with cleared state, not resume from the cursor (rows written by the previous package are not the new package's function of the words). A rotation to an unlisted aggregator, or a beacon upgrade to unmodelled code, fails closed until a package update.
- **Snapshot versus delta refusals.** A snapshot with a missing required attribute is refused at decode time (`flamm_filter` skips the component), while a delta that leaves the attributes undecodable keeps the component alive and refusing until a later delta restores it; a component created inside a feed's rotation window would therefore be skipped at snapshot until a restart.
- **Statics not pinned:** `hook_genesis_params_hash` (keeper-tunable in the live hook; the port pins the strategy hash), `factory_upgrade_delay`, and the price feed / factory codehashes (pinned by the package's deployments registry at creation, not by the decoder).
- The three deliberate departures from the Go port, all unreachable from tracked chain state and exercised by no fixture: `irm.rs` uses checked `I256` arithmetic where the Go uses unbounded integers; `morpho.rs` accrues with a checked `totalSupplyAssets - feeAmount`; `venue_health` returns `Panic(0x12)` on a zero pin (validated non-zero at `registerPool`).
- **Partial fills revert** by design; the simulation quotes only sizes the pool fills in full, so a partial fill at execution means the state moved between the quote and the fill (a clip by the gate room, the notional cap or the funding ceiling), and the router's revert leaves nothing stranded.
- **The pool is taken from calldata** and checked against the permissionless factory's `isPool`; the security model therefore treats it as sender-controlled. The alternative, one executor per pool with the pool as an immutable, is open question 4.
- **Not deployed.** The executor is declared for deployment (`executor_deployments.json`, `deploy-executors.js`) but not deployed; `executor_addresses.json` and the solver docs' contract addresses wait for the deployment and whitelisting.

### Open questions for PropellerHeads

1. **Seeds.** Are manifest params acceptable for the 48 start-block words, the immutables and the deployments registry (each documented with the `eth_call` / `eth_getStorageAt` that produced it and asserted by unit tests), or would you rather see an `initialized_accounts`-style mechanism (which needs `debug_storageRangeAt`) or a longer `initialBlock` (the sequencer feed's last update can be years back)?
2. **`apply_block` semantics on partial blocks.** We rely on `BlockContext.timestamp` being the expected execution block's timestamp and treat a repeated timestamp as the flashblocks of one block. Is that the intended contract for native sims on Base? And is a positioned pool re-emitting on every block (the `changed` flag is honest, see above) acceptable for the extractor and for clients, or would you prefer the flag to compare quotes at a few sizes instead of the accrual inputs?
3. **Attribute volume and change types.** About 151 dynamic attributes per component, both components carrying the same set, updated in roughly 40-60 % of Base blocks (Morpho totals and Chainlink rounds). The package emits the first write of a tracked word the indexer holds no row for as a `Creation` and later writes as `Update`s, because `tycho-indexer`'s revert path counts an `Update` without a prior row as an attribute miss; is that the change-type convention you want?
4. **Executor pool-from-calldata.** The executor takes the pool from calldata and checks `factory.isPool`; creation is permissionless, so the security model treats the pool as sender-controlled. Acceptable with that entry, or would you prefer the pool pinned as an immutable (one executor per pool, no `isPool` call)?
5. **Harness range.** `test_activation_and_first_swap` streams ~148k Base blocks against the docs' "ideally below 1,000 blocks" (the pool sat paused for ~143k of them, and a native integration cannot skip its components' creation). Is that acceptable in your CI, or would you prefer a shorter range plus the local replay as evidence? A third test up to a recent block (e.g. 51433700, after the unpause) would exercise the richer state once the harness is available; likewise, is the Base fork at 51302920 with the three sizes per direction sufficient execution evidence, or do you want a fork test at a recent block as well (six swaps, deposits, withdrawals and keeper recenters since)?
6. **Hosting.** Would PropellerHeads run the `base-flamm` extractor on the hosted Base stack once the package passes the range test, and is there a preferred process for the package updates that seeds require (a rotation to a new aggregator, a further pool)?
7. **DCI on Base.** Is there a timeline for DCI on flashblocks chains? It does not change this design (the token-model and pre-existing-contract problems remain), but it would change how a future FLAMM pool's external words could be seeded.
8. **Deployment and whitelisting.** `npx hardhat run scripts/deploy-executors.js --network base` now deploys `FLAMMExecutor(factory)` through the CREATE2 factory (re-runnable; it skips an existing deployment). Who runs it, who whitelists the address on the router, and who updates `executor_addresses.json` and the solver docs?
9. **Lever-down.** When FLAMM ships an exact-pay lever-down entrypoint, would you take it as a third component and a venue byte 2 on this executor, or as a separate executor?

### After merge

- The hosted range test (`cargo run -- range --package base-flamm --chain base`) and the dev-cluster indexing, simulation and execution run are on your side; the README's "The range test" section says what each step should print.
- The package release is the usual manual one: tag `base-flamm-0.1.0` and dispatch `release-substreams-package` with `base-flamm` as the package and manifest.
- The executor's deployment (`deploy-executors.js`, Base list) and whitelisting are on your side, then `executor_addresses.json` and `docs/for-solvers/execution/contract-addresses.md`.

### Commits

```
2d2ef5f69 feat: port the FLAMM curve, fee law and swap hook to tycho-simulation
1b6ca3547 feat: port the FLAMM financing, gate and leverage math
161ae7a16 feat: add the flamm pool core, ProtocolSim, decoder and registry
093d11e46 test: chain parity fixtures and end-to-end replay for flamm
e22254a54 feat: add the base-flamm substreams package
e276519d7 docs: list flamm under supported protocols
0bdf27c4a feat: export flamm_filter and require it in the stream builder
9323e6dfc feat: add FLAMMExecutor
120ed36fb feat: add the flamm swap encoder
a5a9b17dc test: FLAMM executor and router integration tests
46bcc39a5 feat: model FLAMMExecutor in the security model
41a63ad62 test: wire the flamm executor fixture and enable execution checks
21353b17f chore: list flamm among the Base executors to deploy
```

177 files changed, 43,150 insertions, 20 deletions against `main` (0.408.0, `d9a64a29f`): `protocols/substreams/base-flamm` (+ workspace `Cargo.toml` / `Cargo.lock`), `crates/tycho-simulation/src/evm/protocol/flamm` (+ `evm/protocol/mod.rs`, `evm/protocol/filters.rs`, `evm/stream.rs`, the crate's `Cargo.toml`, the root `Cargo.lock`), `crates/tycho-execution` (`contracts/src/executors/FLAMMExecutor.sol`, `contracts/test/protocols/FLAMM.t.sol`, `contracts/test/assets/calldata.txt`, `contracts/scripts/deploy-executors.js`, `src/encoding/evm/swap_encoder/{flamm.rs, mod.rs, swap_encoder_registry.rs}`, `src/encoding/evm/gas_estimator.rs`, `tests/protocol_integration_tests.rs`, `model/src/model/executors.rs`, `config/{executor_deployments.json, test_executor_addresses.json}`, `CLAUDE.md`), `protocols/testing` (`src/state_registry.rs`, `src/execution.rs`, `fixtures/FLAMM.runtime.json`, `scripts/update_runtime_bytecode.sh`), `crates/tycho-integration-test/src/stream_processor/protocol_stream_processor.rs`, `.github/workflows/ci-rust.yaml`, `docs/for-solvers/supported-protocols.md`. The commit SHAs above are the fork's; the same series was reviewed internally as alex-alra-arteaga/tycho#1 (indexing and simulation) and #2 (execution, stacked).


alex-alra-arteaga:
## Design document

The design this PR implements, as shared with PropellerHeads when we discussed whether FLAMM needed DCI on Base.

---

# Everlong FLAMM on Base: a native Tycho integration without DCI

Design proposal for PropellerHeads review. Everything below rests on the deployed contracts at Base block ~51.3M, the Tycho monorepo at `0.406.0`, and the measurements listed in the appendix. Where a claim came from a run, the run is named.

## 1. Summary

FLAMM is a single-pool, financed-leg AMM on Base (cbBTC/USDC). Every quote reads state from contracts FLAMM does not own — Morpho Blue, its interest-rate model, three Chainlink feeds and the L2 sequencer feed — and every sell borrows USDC from Morpho inside the fill. That is why a VM integration needs DCI, and DCI cannot run on Base today.

We propose a **native integration**: the quote is computed in Rust from state that substreams can track without tracing. It works because the set of external storage FLAMM reads is small (47 logical slots across 11 contracts), fully enumerated, and either written frequently on chain or announced by events. Nothing in it needs `debug_storageRangeAt`.

We already have a wei-exact port of the whole quoting path in Go (built for the KyberSwap integration), validated against the deployed pool on tens of thousands of previews and thousands of executed fills. The Rust `ProtocolSim` is a port of that, checked against the same Solidity-generated fixtures.

What we ask PropellerHeads to review: the state-sourcing plan (section 4), the four seed values (section 4.4), and the open questions (section 9).

## 2. The protocol in one page

| | |
|---|---|
| Pool | `0xc0fdCB1799cCc2CEBaA1fe247157b0dF33D57572`, beacon proxy; beacon = `FLAMMFactory 0x1BfcE014774D0DD7e04bC595D46Fa09F7dCCF45f`; implementation `0xaAD580BeAa2cbd8Ab5F3956a5c56EDa1D5ee7184`; created block 51154990, activated 51298416 |
| Pair | cbBTC `0xcbB7C0000aB88B473b1f5aFd9ef808440eed33Bf` (8 dp, the pool asset) / USDC `0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913` (6 dp, the loan asset) |
| Swap venue | `swap(tokenIn, tokenOut, amountIn, minOut, to, deadline) → (amountInUsed, amountOut)`, exact-in, both directions, partial fills possible. Price from `EverlongHook` (curve `AlmCurve`, fee law `EverlongStrategy`), gated by `FLAMMGateLib` (loan-to-value room) and a ±8% band around the `PriceFeed` cross |
| Leverage venue | `leverUp(cbBTC in)` / `leverDown(USDC in)`, all-or-nothing, priced by `EverlongLeverageHook` on the frozen `CollRebalancerMath` curve with a keeper-posted spread. Paused on chain today |
| Financing | A sell pays USDC from the pool's liquid balance, then from its Morpho supply, then **borrows on Morpho Blue** against cbBTC posted as collateral, through `MMRouter` and `MorphoBlueAccount`. A buy repays, lends the excess and reclaims collateral |
| Oracles | `PriceFeed` combines Chainlink cbBTC/USD and USDC/USD behind the L2 sequencer uptime feed (reverts on stale rounds or a down sequencer). Morpho's market oracle is `MorphoChainlinkOracleV2` over Chainlink BTC/USD, used for the Router's 2% band check and Morpho's own health check |

Source: `EverlongLabs/blockend` @ `80abd43` (`src/core/flamm/*`, `src/hooks/everlong/*`, `src/core/mm/*`, `src/core/PriceFeed.sol`). Runtime codehashes of all 21 deployed contracts match the deployment record `script/flamm/c104/deployments/c104.8453.json`.

## 3. Why not VM

We built and measured the VM route before proposing native. Four independent problems, the first being the one you named:

1. **DCI cannot run on Base.** `tycho-indexer/src/extractor/protocol_extractor.rs` (DCI revert handling): "DCI cannot run on partial-block chains (its block-layer validation needs parent-hash linkage, which ephemeral partial hashes break) … Revisit when DCI supports a flashblocks chain (e.g. Base)." `tycho-client` also disables DCI queries on Base (commit `7cbc552`). Both public Base RPCs reject `debug_storageRangeAt` (`-32601`). A reth `debug_storageRangeAt` would remove the RPC obstacle but not the partial-block one.

2. **Tycho's token model breaks FLAMM sells.** In VM simulation, tokens become `TokenProxy` with fake balances for the component. A FLAMM sell borrows USDC on Morpho, which transfers real USDC to the pool through the real token path; the pool's fake balance does not move, and `MorphoBlueAccount` reverts on its exact-delta check (`ExactDelta()`, `MorphoBlueAccount.sol:139-145`). Reproduced on a fork with the exact `vm/assets/TokenProxy.bin` runtime. Making the pool, Morpho and the account all custom holders fixes single quotes, but `EVMPoolState::get_balance_overwrites` keeps only one holder per token (`HashMap::extend` at `state.rs:866,884`, reproduced), and chained quotes reset balances between calls and revert `CustodyMismatch` on the second fill at today's state. Both need changes to `tycho-simulation` core.

3. **Pre-existing contracts would need seeding.** Even with DCI, the Morpho/Chainlink accounts pre-date FLAMM by years. Without DCI a VM package has to embed their bytecode and hand-picked storage as synthetic creations — no precedent in the repo, and it writes partial chain-global storage that other integrations may later list.

4. **Rotations need operator rewinds.** A Chainlink aggregator rotation or a FLAMM beacon upgrade would need the extractor rewound by hand (`ProtocolExtractor::new` resumes from `extraction_state`; no reset command exists).

The native route has none of these: no token simulation, no bytecode, no rewinds, and it fits the existing `ProtocolSim` + substreams pattern that LunarBase and Aerodrome already use on Base.

## 4. State sourcing

Everything a quote reads, and where it comes from. The read set was measured with `vm.startStateDiffRecording` over every simulated path (previews, executed swaps in both directions, chained sequences, leverage) at four Base blocks; executed fills are a strict superset of previews and the union is what is listed here.

### 4.1 FLAMM-owned contracts — full storage from creation

All were created in blocks 51154966–51154999, so from `initialBlock = 51154966` substreams sees every write. Emitted as component attributes (one attribute per storage word, decoded by the pinned storage layout; the layout is tied to the implementation codehash so a beacon upgrade fails closed until the package is updated).

| Contract | Slots read on a quote | Notes |
|---|---|---|
| Pool `0xc0fd…7572` (`FLAMMStore`, ERC-7201 namespace) | 23 | physical/posted pool asset, `loans[0]` config and liquid, dials, limits, features/pause bits, hook set, lever pause/spread word, pending ceremonies |
| `EverlongHook 0x65CB…2134` | 17 | kappa, x, reserves/idle, anchor, reservation price, rv/ema, tuning row, support |
| `MMRouter 0x19A9…6bB4` | 22 | venue list, orders, caps, pins, managed collateral/supply shares, global pause |
| `MorphoBlueAccount 0x6760…6c48` | 7 | market params, IRM grace state |
| `PriceFeed 0xbED2…8816` | 4 | token feed config, heartbeats |
| `FLAMMFactory` | 2 | `_implementation` (beacon), `isPool[pool]` |
| `LeverageSpreadHook 0x0498…f93b` | 1 | spread post + timestamp (leverage only) |

Libraries (`AlmCurve`, `FLAMMSwapLib`, `FLAMMGateLib`, `MMRouterLib`, `FLAMMLeverLib`, `CollRebalancerMath`) are code-only and ported; nothing to index.

### 4.2 Morpho Blue and the IRM — five Morpho slots plus one IRM word, absolute values

FLAMM reads exactly:

- `market[id]` +0 `totalSupplyAssets|totalSupplyShares`, +1 `totalBorrowAssets|totalBorrowShares`, +2 `lastUpdate|fee` (`keccak(id·3)` = `0xb37d8d77…6a6a/6b/6c`);
- `position[id][MorphoBlueAccount]` +0 `supplyShares`, +1 `borrowShares|collateral` (`0x1ca893d1…fce9/fcea`);
- `AdaptiveCurveIrm.rateAtTarget[id]` (`0xe6f1c64c…ee12`).

Market id `0x9103c3b4e834476c9a62ea009ba2c884ee42e94e6e314a26f04d312434191836` (USDC/cbBTC, LLTV 0.86). `idToMarketParams`, `isAuthorized`, `nonce` and the fee recipient are never read on any FLAMM path.

The package subscribes to storage changes of the two addresses and keeps only those six keys. The values are absolute, so no event arithmetic. The market slots change in 23.3% of Base blocks and the IRM slot on every accrual of the market, so a fresh index is complete within minutes of the start block and long before the pool's activation at 51298416; the position slots are only ever written by FLAMM itself. The decoder refuses to quote until all six attributes have been seen (fail closed rather than zero-as-unknown).

### 4.3 Chainlink and the sequencer — events, not slots

`PriceFeed.cross` reads `latestRoundData()` on each `EACAggregatorProxy`, which forwards to the current aggregator. Track:

| Feed | Proxy | Aggregator today | Source |
|---|---|---|---|
| cbBTC/USD | `0x07DA0E54…9f9D` | OCR2 `0x51cE3091…e718` | aggregator `NewTransmission` / `AnswerUpdated(answer, roundId, updatedAt)` |
| USDC/USD | `0x7e860098…c6B` | OCR2 `0x68bE4C50…a5E2` | same |
| BTC/USD (Morpho oracle) | `0x64c91199…848F` | SVR `DualAggregator 0xe5eC87A3…9Df1` | same events; see 7.2 for the reveal caveat |
| Sequencer uptime | `0xBCF85224…6433` | `OptimismSequencerUptimeFeed 0x606c6ECb…602e` | `AnswerUpdated` / `RoundUpdated` (status, startedAt) |

Rotation: `AggregatorConfirmed(previous, latest)` on the proxy switches the tracked aggregator; the component refuses to quote until the new aggregator's first update, at most one heartbeat (3600 s cbBTC, 90000 s USDC). No rewind, no package release.

Heartbeats and the sequencer grace period (3600 s) come from `PriceFeed` storage (4.1).

### 4.4 Seeds — four constants, all verifiable by one `eth_call` at block 51154965

State that changes rarely and may have no write inside the indexed range:

1. the current aggregator behind each of the four proxies (phase/aggregator word) — rotations are months apart;
2. the sequencer feed's current status and `startedAt` — changes only on an outage;
3. the last round of each aggregator before the start block — needed so the first quote is possible before the next round;
4. Morpho market totals and `rateAtTarget` at the start block — not strictly needed (they are rewritten within minutes) but they make the first attested state deterministic.

We propose to carry these as substreams manifest params, each documented with the `eth_call` that produced it, and to have the integration test assert them against the chain at the start block. This is the one place the design departs from "everything comes from the stream"; we would like your view on whether params or an `initialized_accounts`-style mechanism is preferred (question 9.1).

### 4.5 What is deliberately not tracked

- Token balances and allowances: a native sim does not move tokens. FLAMM's custody checks are exact-delta checks that hold on chain by construction.
- `FLAMMLens`, the deployed `FLAMMTychoSwapAdapter`, `EverlongCore`: not on the quote path.
- Morpho positions of other users: FLAMM reads only market totals and its own position.

## 5. Substreams package `base-flamm`

`protocols/substreams/base-flamm/`, `initialBlock 51154966`, native implementation type. One map module plus stores:

- **`map_components`**: `FLAMMFactory.PoolCreated` → two components on the pool: swap (id = pool address) and lever-up (id = `pool(20) ‖ 0x00000000 ‖ uint64(1)`, BopAMM's discriminator layout). Tokens `[cbBTC, USDC]` from the `createPool` calldata. Only pools whose invariant-hook codehash matches the registry are emitted (the factory is permissionless).
- **`map_protocol_changes`**: for each block, the changed storage words of the FLAMM-owned contracts (filtered to the pool's own hook/router/account addresses), the six Morpho/IRM words, and the Chainlink/sequencer events, all as entity attributes on both components; plus `balance_changes` for TVL (5.1).
- Stores: the pool→hooks/router/account mapping so the filter knows which addresses belong to which pool; the current aggregator per proxy.

The module is a pure function of block + params, so it is safe under Base partial blocks.

### 5.1 TVL

Component balances are the pool's tradable inventory, derived from tracked words: cbBTC = physical + collateral posted in Morpho (`position.collateral`), USDC = liquid + supplied + borrow capacity against the posted collateral at the Router's caps. Today that is roughly 0.0023 BTC; it is stated here because it matters for filters, not because it is large.

## 6. Simulation (`crates/tycho-simulation/src/evm/protocol/flamm/`)

A `ProtocolSim` whose state struct holds the words from section 4. Methods:

- `get_amount_out`: runs the full fill — plan (room, notional cap, up to four funding passes against the Morpho-derived funding ceiling), hook fill (curve + fee law), validation (band, floors/caps), settlement (Morpho share math, IRM rate, collateral posting) — and returns the amount plus the post-state, exactly as the pool would settle at the execution block. Partial fills are refused (see 7.1).
- `get_limits`: the largest fully-consumed size in each direction, found by bisection on the port (no RPC).
- `spot_price`, `fee`: from the hook state.
- `delta_transition`: applies attribute updates.
- `apply_block(BlockContext)`: takes the execution block timestamp for Morpho accrual, IRM adaptation and feed/sequencer staleness. This is what makes the quote exact at the execution block rather than at the last pool update.

The math is a port of the Go simulator (12 modules; `almcurve`, `fee`, `hook`, `gate`, `morpho`, `irm`, `account`, `router`, `swap`, `levcurve`, `levhook`, `lever`), each function tied to its Solidity line. The Go port's parity tests are driven by Solidity-generated fixtures (Foundry generators against the deployed bytecode, sha256-pinned) and by executed fills on Base forks; the Rust port reuses the same fixtures.

## 7. Execution

`FLAMMExecutor` (`crates/tycho-execution/contracts/src/executors/`): `ProtocolWillDebit`, receiver = pool, calldata `pool ‖ tokenIn ‖ tokenOut ‖ venue`, no transfers/approvals/storage, and **it reverts unless `amountInUsed == amountIn`**. Encoder registered as `flamm`; security-model entry since the pool address comes from calldata (permissionless factory).

### 7.1 Why no partial fills and no lever-down

`TychoRouterV3` strands input a `ProtocolWillDebit` executor does not consume: `_transferFromUser` pulls the nominal amount, the Dispatcher only resets the allowance, `_finalizeBalances` checks transient deltas, and there is no sweep (`TransferManager.sol:152-160`, `Dispatcher.sol:162-165`, `Vault.sol:346-376`; reproduced on a fork: 400 of 1000 stranded). FLAMM can clip a sell (room, notional cap, funding ceiling) and a lever-down pulls only its net pay leg — about 50% of the input by construction (`FLAMMLeverLib.sol:135-140`). So: the simulator only quotes sizes that fill in full, the executor asserts it, and lever-down waits for an exact-pay entrypoint on the FLAMM side (a beacon upgrade through `FLAMMFactory.scheduleImplementation`).

### 7.2 Time-dependent state

Quotes are exact at the execution block. Three things move between the indexed block and execution, all handled fail-closed:

- Morpho interest and IRM adaptation: computed to the execution timestamp from tracked words (exact, given `apply_block`).
- Chainlink rounds: a new round between quote and fill can move the ±8% band check and the leverage gates. The Go simulator carries a configurable band margin measured from 567 real cbBTC/USD rounds (median 9.7 bps, p99 52.1); the Rust port carries the same knob.
- The SVR `DualAggregator` behind Morpho's BTC/USD oracle can change its answer with time alone (secondary-round reveal). It affects only the Router's 2% band and Morpho's health check, never the price; the port refuses when the reveal window could flip either.

## 8. Evidence we already have

From the Go port of the same math (KyberSwap integration, both PRs open for internal review):

- ~34.3k `previewSwap` and ~46k `previewLever` comparisons against the deployed pool: identical to the wei, reverts matched by class.
- 6,140 fuzzed pool states (curator setters, keeper actions, Morpho activity by others, time warps) with ~105k grid fills: 0 mismatches.
- 2,397 fills executed on Base forks through an executor harness with the simulator advanced only by its own post-state: every amount exact.
- The pool's only settled swap (block 51302916, tx `0x46c3cd72…01fa`, 15,000 sats → 11,301,759 USDC) reproduced from tracked state.
- A 93-agent adversarial audit of that port, all confirmed findings fixed.

The Rust port inherits the fixtures and the differential harness.

## 9. Open questions for PropellerHeads

1. **Seeds.** Are manifest params acceptable for the four start-block constants in 4.4, or would you rather see `initialized_accounts` (which needs `debug_storageRangeAt`) or a longer `initialBlock` (the sequencer feed's last update may be years back)?
2. **Attribute volume.** ~76 FLAMM words + 6 Morpho/IRM words + ~10 feed words per component, updated in roughly 40–60% of Base blocks (Morpho totals and Chainlink rounds). Any concern for the extractor or for clients?
3. **`apply_block` semantics.** We rely on `BlockContext.timestamp` being the expected execution block. Is that the intended contract for native sims on Base with partial blocks?
4. **Execution.** Is a native executor whose pool address comes from calldata acceptable with a security-model entry, given the factory is permissionless, or would you prefer the pool pinned as an immutable (one executor per pool)?
5. **Test harness.** The range test covering creation → activation → first swap spans ~148k Base blocks (the pool sat paused for ~143k of them). Is that acceptable in your CI, or would you prefer a shorter range plus fork evidence?
6. **Hosting.** Would PropellerHeads run the `base-flamm` extractor on the hosted Base stack once the package passes the range test?

## Appendix: measurements referenced

- Read-set census: Foundry `startStateDiffRecording` over 15 paths × 4 blocks (51302915, 51313000, 51322600, 51322869); 34 non-harness accounts, 76 FLAMM-owned slots, 47 logical external slots.
- VM token-model spike: 48 Foundry tests with the exact `vm/assets/TokenProxy.bin` runtime at 51302915; default overwrites → `ExactDelta` on every borrowing sell; three-holder variant passes single and chained quotes; `HashMap::extend` collapse reproduced in Rust.
- TychoRouter partial-consumption: 5 Foundry tests against the current router sources.
- DCI on Base: `protocol_extractor.rs` comment, `CHANGELOG` 0.78.1 / commit `7cbc552`, `-32601` from `mainnet.base.org` and `base-rpc.publicnode.com`.
- Chainlink round sizes: 567 cbBTC/USD rounds over blocks 51066959–51369359.

PR #1461, opened 2026-09-18
feat(client)!: close the websocket when its connection handle drops

zizou0x:
## What

`DeltasClient::connect` returns a `ConnectionHandle` instead of a bare `JoinHandle`. Dropping the handle shuts the websocket task down the same way `close()` does; awaiting it yields the task result as before.

## Why

`TychoStreamBuilder::build` connects the websocket, then starts the `BlockSynchronizer`. When `block_sync.run()` fails — a stalled extractor that delivers no first message, or no known extractors — the `?` dropped the `JoinHandle`, which detaches the task. The monitor task that calls `close()` is only spawned after that point, and the indexer's 5 s heartbeat pings keep the client's 60 s idle timeout from firing, so the connection stayed open until process exit. Each failed build held one indexer websocket.

On 2026-09-17 this filled the robinhood `fynd-backend` connection budget during a substreams stall: hindsight rebuilds its solver in a loop after a feed death, and every failed rebuild cost a slot.

With this change the `?` drops the handle and the socket closes. `stream.rs` is unchanged: the monitor task's `select!` polls the handle through its `Future` impl, and dropping the losing branch closes the socket before the explicit `close()` call.

`connect` also stops the task when it returns `NotConnected`. That check reads connection state the task clears on every retry, so it can be false while the task is alive and reconnecting; returning the error without stopping the task left a connection nothing could close. It aborts rather than signalling the command channel, because a task sleeping between reconnection attempts does not read that channel until it has connected once more.

`ConnectionHandle` is `#[must_use]`: discarding it closes the connection, which should not happen silently. This does not catch `let _ = client.connect().await;`, which suppresses the lint.

## Not changed

- Dropping the handle while the task sleeps between reconnection attempts takes effect once the task's loop next reads its command channel, same as `close()` today. The error path above uses `abort` precisely to avoid that delay.
- `tycho-simulation` compiles unchanged; it never held the handle.

## Testing

Unit:

- `test_dropping_connection_handle_closes_socket`: the server observes the close after the handle drops, the client reports not connected, and `subscribe` fails fast.
- `test_connection_handle_keeps_socket_open_while_held`: negative control.
- `test_connection_handle_abort_cancels_the_task`: a task that never reads its command channel is still stopped.

End-to-end against `tycho-dev`, in one long-lived process, counting that process's established connections to the indexer across every A record of the host:

| scenario | `main` | this branch |
|---|---|---|
| 3 builds that fail after connecting | 3 sockets held | 0 |
| healthy `uniswap_v2` stream, real blocks | 2 sockets, 2 messages | 2 sockets, 2 messages |
| after dropping the consumer | 0 | 0 |

The failing builds register an extractor the server does not know, so `build` connects and then fails in `BlockSynchronizer::run`. The healthy-stream row is the check that the connection is not closed too early; it matches `main` exactly.

Local CI without DB or RPC: fmt clean on the pinned nightly, clippy clean, docs and no-default-features builds pass. tycho-client: 155 passed. Workspace: the only failures are `RPC_URL`-dependent tests in tycho-execution, tycho-simulation and protocol-testing, which pass in CI.

## Breaking

`DeltasClient::connect` return type. Direct users of `WsDeltasClient` that called `.abort()` on the returned `JoinHandle` should drop the handle instead.


crates/tycho-client/src/deltas.rs
```diff
@@ -164,12 +167,59 @@ pub trait DeltasClient {
...
+impl ConnectionHandle {
+    /// Stops the websocket task immediately, for a connection no caller will ever own.
+    ///
+    /// Unlike dropping the handle, this does not wait for the task to reach the point where it
+    /// reads its command channel: a task sleeping between reconnection attempts would otherwise
+    /// reconnect once more before it stops.
+    fn abort(self) {
+        self.task.abort();
+    }
+}
```
louise-poole, line 214: Good suggestion from Claudio here.

---

🤖 `abort` is private, so dropping the handle is all a caller gets. A drop only queues a command, and a task asleep between reconnects won't read it until it has reconnected once more. `connect` needs the fast version for its own error path, and anyone moving off `JoinHandle::abort` has no replacement.

Make `abort` public, and say in the doc that it skips the reconnect delay.

crates/tycho-client/src/deltas.rs
```diff
@@ -164,12 +167,59 @@ pub trait DeltasClient {
...
+    fn drop(&mut self) {
+        // A closed channel means the task already ended. A full one cannot happen: this is the
+        // only message sent per shutdown into a channel sized for message buffering.
+        let _ = self.cmd_tx.try_send(());
+        // Mark dead and notify so ensure_connection() callers return NotConnected instead of
+        // waiting for a connection that will never come.
+        self.dead.store(true, Ordering::SeqCst);
+        self.conn_notify.notify_waiters();
+    }
+}
```
louise-poole, line 234: Important bug, please have a look?

---

🤖 `Drop` sets the shared `dead` flag and nothing ever clears it, `connect()` included. Call `connect()` again on the same client and you get `Ok` and a real socket, but `ensure_connection` sees `dead` and fails every `subscribe` with `NotConnected`. The error path is the same story: `abort(self)` runs `Drop`, so a transient `NotConnected` retires the client permanently. `close()` already did this before, but a handle drop and a failed `connect()` are the paths people actually retry after.

Open a stream, shut it down, then open another on the same client. The second one reports success, but nothing you subscribe to ever delivers. No log line or return value tells you the client is finished, so you just have to build a new one.

Clear `dead` at the top of `connect()`, right after the `AlreadyConnected` check. For the error path you could also move the `is_connected()` check above the handle construction and call `jh.abort()` directly, so no handle is built for a connection that never existed.

crates/tycho-client/src/deltas.rs
```diff
@@ -1678,6 +1769,88 @@ mod tests {
...
+            Duration::from_millis(10),
+            client.subscribe(
+                ExtractorIdentity::new(Chain::Ethereum, "vm:ambient"),
+                SubscriptionOptions::new(),
+            ),
+        )
+        .await
+        .expect("subscribe on a closed client must fail fast");
+        assert!(subscription_res.is_err());
+    }
```
louise-poole, line 1868: Add a test for the earlier scenario we flagged.

---

🤖 The drop test stops at "subscribe fails". Nobody calls `connect()` a second time, which is where the permanent `dead` flag shows up: `Ok` comes back and every `subscribe` is broken (same flag as the comment above).

Add a second `connect()` and assert whichever outcome you want — a working connection, or a clear error.

crates/tycho-client/src/feed/synchronizer.rs
```diff
@@ -1349,7 +1349,7 @@ mod test {
                 .await
         }
 
-        async fn connect(&self) -> Result<JoinHandle<Result<(), DeltasError>>, DeltasError> {
+        async fn connect(&self) -> Result<crate::deltas::ConnectionHandle, DeltasError> {
```
louise-poole: To appease my OCD:

---

🤖 Add `ConnectionHandle` to the existing `crate::{deltas::MockDeltasClient, ...}` import rather than spelling the path out inline.

crates/tycho-client/src/deltas.rs
```diff
@@ -1590,7 +1681,7 @@ mod tests {
         )
         .unwrap();
 
-        let jh: JoinHandle<Result<(), DeltasError>> = client
+        let jh: ConnectionHandle = client
```
louise-poole: Another OCD nit pick:

---

🤖 `ws_jh` and `jh` hold a `ConnectionHandle` now, so the names still point at the old type. `ws_conn` and `conn` read better. Not a hill I'd die on.

crates/tycho-client/CLAUDE.md
```diff
@@ -6,7 +6,8 @@ Consumer library implementing the snapshot + deltas pattern for real-time protoc
 
 ```
 rpc.rs              HTTP snapshot client — fetches protocol state at a block height
-deltas.rs           WebSocket client — streams real-time state deltas
+deltas.rs           WebSocket client — streams real-time state deltas. connect() returns a
+                    ConnectionHandle; the socket lives exactly as long as that handle (drop = close)
```
louise-poole: Small doc fix

---

🤖 The socket can end while the handle is still alive: `close()` on any clone, exhausted reconnects, or the server closing. It can also outlive the drop, because `Drop` only queues a command the task reads after its current reconnect sleep. A handle dropped mid-reconnect lets the task open one more connection first, which is the case `abort`'s own doc points out.

Say what actually happens: dropping the handle shuts the connection down, a task mid-reconnect stops after its next attempt, and `close()` or exhausted reconnects end it too. Same fix in the CLAUDE.md module map and in the `must_use` message, which still says "as soon as this handle is dropped".

crates/tycho-client/src/deltas.rs
```diff
@@ -164,12 +167,59 @@ pub trait DeltasClient {
...
+    dead: Arc<AtomicBool>,
+    conn_notify: Arc<Notify>,
+}
+
+impl ConnectionHandle {
+    /// Stops the websocket task immediately, for a connection no caller will ever own.
+    ///
+    /// Unlike dropping the handle, this does not wait for the task to reach the point where it
+    /// reads its command channel: a task sleeping between reconnection attempts would otherwise
+    /// reconnect once more before it stops.
```
louise-poole: Small doc fix pls

---

🤖 "For a connection no caller will ever own" and the reconnect rationale describe the call site, which the crate's docstring convention rules out. Say what it does: stops the task without waiting for it to read its command channel.

louise-poole (commented):
Nice fix, and the end-to-end socket counts are convincing. One thing I'd like resolved before merge: the dead flag is never cleared, so a dropped handle or a failed connect retires the client for good. The rest is small stuff.

---

**stream.rs shuts the client down twice now** — `crates/tycho-client/src/stream.rs:421–433`

Small error handling improvement/fix

---

🤖 The `select!` drops `ws_jh` whichever branch wins, and that already signals shutdown and marks the client dead. The `ws_client.close().await` after it adds nothing. It can also lose a race: if the task has dropped its command receiver by then, `close()` errors and the `warn!` logs a failure that isn't one.

Delete the `close()` call and its warn branch. If you'd rather keep it, put an explicit `drop(ws_jh)` before it so the ordering is deliberate.

---

**The `dead` field doc no longer matches what sets it** — `crates/tycho-client/src/deltas.rs:243–244`

Small doc fix pls.

---

🤖 "If set the client has exhausted its reconnection attempts" is now one of three reasons — `close()` and a handle drop set it too. Reword as "Set once the client is permanently closed: reconnects exhausted, `close()`, or the connection handle dropped".

crates/tycho-client/src/deltas.rs
```diff
@@ -164,12 +167,59 @@ pub trait DeltasClient {
...
+
+impl Drop for ConnectionHandle {
+    fn drop(&mut self) {
+        // A closed channel means the task already ended. A full one cannot happen: this is the
+        // only message sent per shutdown into a channel sized for message buffering.
+        let _ = self.cmd_tx.try_send(());
+        // Mark dead and notify so ensure_connection() callers return NotConnected instead of
+        // waiting for a connection that will never come.
+        self.dead.store(true, Ordering::SeqCst);
+        self.conn_notify.notify_waiters();
```
louise-poole, line 232: Small doc fix pls.

---

🤖 This is the third thing that sets `dead`, but the field's doc at line 243 still says "If set the client has exhausted its reconnection attempts". Reword it to "Set once the client is permanently closed: reconnects exhausted, `close()`, or the connection handle dropped".

louise-poole (commented):
Small follow up:

louise-poole (approved):
LGTM! Thank you!!! Please just test locally before merging?

PR #1462, opened 2026-09-18
feat(execution): Fallback router multichain support

louise-poole:
One `TychoFallbackRouter` build now deploys to any chain, and Aerodrome V1 is fallback protocol byte 5.

  ## Contract

  - The three per-chain singletons (Uniswap V4 PoolManager, Fluid liquidity layer, Uniswap V3 static quoter) stay constructor immutables but may each be `address(0)`.
    - A zero PoolManager or liquidity layer makes that protocol byte revert `TychoFallbackRouter__ProtocolUnavailable`. The check lives in `_decodeFallback`, so the protocol quotes as zero and reverts by name on execution.
    - A zero quoter quotes Uniswap V3 by simulation through the generic `simulateFallback`, the same path Uniswap V4 uses. This costs a real swap's gas per quote, so deploying Eden's quoter on the chain is the cheaper option.
  - `AerodromeV1` enum variant, protocol data `[pool: 20]`. Quote and fill both come from the pool's own `getAmountOut`, since Solidly pools price their fee and stable curve themselves and do not fit byte 0's constant-fee V2 formula. Direction is derived from token sort order like V2.
  - `IAerodromeV1Pool` moves to `contracts/interfaces/`, following the Curve and Fluid ABIs.

  ## Rust encoder

  - Builds on any chain. The Angstrom hook config is optional and only used to reject that hook when present.
  - New `aerodrome_v1` protocol with a `pool` field.
  - Fork names resolve through `UNISWAP_V2_FORKS`, `UNISWAP_V3_FORKS` and the four Slipstreams deployments instead of hand-written serde aliases.

  ## Deployment

  `deploy-fallback-router.js` reads the three singletons as optional and prints which protocols a chain's deployment disables. The `FallbackExecutor` still goes through `deploy-executors.js` once a `fallback` entry with the router address is added to `executor_deployments.json`. Nothing is deployed yet.


crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
```diff
@@ -18,23 +25,145 @@ const PAMM_ADDRESS_ATTRIBUTE: &str = "pamm_address";
...
 const MAX_UNISWAP_V2_FEE_BPS: u8 = 30;
 
-/// The fallback protocol that fills a pAMM swap when the pAMM fails. The solver picks it and its
-/// pool, JSON-encoded into `Swap::user_data` (e.g.
-/// `{"fallback_protocol":"uniswap_v3","pool":"0x…"}`); a swap without one is rejected. Each
-/// protocol's JSON fields are the variant fields below. The variants mirror
-/// `TychoFallbackRouter.FallbackProtocol`.
+/// Slipstream deployments and their forks. The registry encodes them through
+/// `SlipstreamsSwapEncoder` because their executor data differs from Uniswap V3's, but the pool
+/// itself keeps V3's `swap` ABI and callback, which is all `TychoFallbackRouter` uses.
```
tamaralipows: Really? Why do we have a separate `SlipstreamsSwapEncoder` then? Is that just because the executor is different?

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
```diff
@@ -18,23 +25,145 @@ const PAMM_ADDRESS_ATTRIBUTE: &str = "pamm_address";
...
+impl FallbackProtocol {
+    /// Every protocol, in protocol-byte order.
+    pub const ALL: [FallbackProtocol; 6] = [
+        FallbackProtocol::UniswapV2,
+        FallbackProtocol::UniswapV3,
+        FallbackProtocol::UniswapV4,
+        FallbackProtocol::Curve,
+        FallbackProtocol::FluidV1,
+        FallbackProtocol::AerodromeV1,
+    ];
```
tamaralipows: If we add a protocol now we need to change this in 5+ places. Any more efficient way to organize this?

crates/tycho-execution/src/encoding/evm/swap_encoder/fallback.rs
```diff
@@ -18,23 +25,145 @@ const PAMM_ADDRESS_ATTRIBUTE: &str = "pamm_address";
...
+    /// singleton exactly when it has the protocol's executor. Every other protocol is addressed
+    /// per swap and needs nothing from the deployment.
+    pub fn supported_on(self, chain: Chain) -> bool {
+        let executor = match self {
+            FallbackProtocol::UniswapV4 => "uniswap_v4",
+            FallbackProtocol::FluidV1 => "fluid_v1",
+            FallbackProtocol::UniswapV2 |
+            FallbackProtocol::UniswapV3 |
+            FallbackProtocol::Curve |
+            FallbackProtocol::AerodromeV1 => return true,
```
tamaralipows: Not sure I understand what's happening here, why is aerodrome perpetually true but only deployed on base?

tamaralipows (approved):
Thank you! Logic looks good, please in particular double check if we can arrange this enum better so we don't need to update in soo many places when we add a new protocol

PR #1463, opened 2026-09-18
feat: take the SignedExclusiveSwap address as a constructor arg

tamaralipows:
## What

The Ekubo V3 executor no longer hardcodes the SignedExclusiveSwap extension address. The address is now a constructor argument, so each chain gets its own.

## How it works

The address is stored in an immutable and read when the executor classifies a hop. A hop whose pool config extension equals it takes the signed path. A zero address reverts at deployment.

## Configuration

The per-chain address lives in the executor deployment config. Ethereum uses the deployed SignedExclusiveSwap. Robinhood Chain has no deployment yet, so it takes the Ethereum address as a placeholder.

## Tests

`EkuboV3ExecutorTest` passes (6 tests, mainnet fork). The Ekubo Rust encoder tests pass. The runtime bytecode fixtures for both Ekubo V3 executors are regenerated.


crates/tycho-execution/config/executor_deployments.json
```diff
@@ -224,7 +227,11 @@
   },
   "robinhood": {
     "uniswap_v2": { "contract": "UniswapV2Executor", "args": [30] },
-    "ekubo_v3": { "contract": "EkuboV3RobinhoodExecutor", "args": [] },
+    "ekubo_v3": {
+      "contract": "EkuboV3RobinhoodExecutor",
+      "comment": "SignedExclusiveSwap is not deployed on Robinhood Chain yet; the Ethereum address is a placeholder until it is.",
+      "args": ["0x55b703eED01b35641963da2FB2E14885993605A3"]
```
louise-poole, line 233: Would it not be fine for the zero address to placehold? What happens if we deploy Ekubo to chain we don't want to have exclusive liquidity on? Seems too strict to require a SignedExclusiveSwap address just for that no?

louise-poole (approved):
Approved to not block, but please have a look at the comment first 🙏

crates/tycho-execution/config/executor_deployments.json
```diff
@@ -58,7 +58,10 @@
       "contract": "RocketpoolExecutor",
       "args": ["0xCE15294273CFb9D9b628F4D61636623decDF4fdC"]
     },
-    "ekubo_v3": { "contract": "EkuboV3Executor", "args": [] },
+    "ekubo_v3": {
+      "contract": "EkuboV3Executor",
+      "args": ["0x55b703eED01b35641963da2FB2E14885993605A3"]
```
Troshchk, line 63: does this address change with the redeploy?

Troshchk (commented):
Thank you!
Just a small q for my understanding

crates/tycho-execution/config/executor_deployments.json
tamaralipows: No. `0x55b7…05A3` is Ekubo's SignedExclusiveSwap contract on Ethereum and stays as is. The Ethereum executor is not redeployed either: the deployed one hardcodes the same address, so it behaves the same as this version. Only the Robinhood executor was redeployed, with the Robinhood extension address.

crates/tycho-execution/config/executor_deployments.json
tamaralipows: Fair point. The revert is there because `_swapHop` classifies a hop with `extension == signedExclusiveSwap`, and pools without an extension have `extension == 0`, so a zero immutable would send every plain pool down the signed path. Making zero mean "no signed path" needs an extra non-zero guard in that check. The placeholder is gone now (Robinhood has its own deployment) and the Robinhood executor is already deployed, so I would leave it for when a chain without a SignedExclusiveSwap deployment comes up.

PR #1465, opened 2026-09-18
feat(indexer): implement the EntityCache entity store

brunoguerios:
## Problem

The delta windows fold every evicted block into `DiscardSink`. Nothing keeps the folded state, so `/contract_state` and `/protocol_state` still read the database for every request.

## Fix

Adds the `EntityCache` entity store behind `FoldSink::fold`. The branch builds on the skeleton from #1306 and includes those commits, so this diff against `main` is the complete store.

- `Tagged<T>` pairs a value with the timestamp of the block that wrote it, the same unit as the database's `valid_from`. A write applies when it is not older than the tag, so replays and writes from a second extractor converge.
- `CachedAccount` tags every slot, balance and code. Code writes recompute `code_hash`. Snapshot and creation constructors build the same `Account` as `into_account_without_tx`.
- `CachedComponentState` keeps one timestamp per entry, since one extractor writes each protocol system in order. Older blocks are replays and are skipped.
- `EntityCache::fold` applies one block under the write lock: components first (new → state deltas → balances → deleted), then accounts (creation creates, deletion removes, unknown updates are skipped).
- Memory: `report` publishes `entity_cache_entries{family}` every 60 s; `measure` walks the cache every 10 reports and publishes `entity_cache_size_bytes{family}` and `entity_cache_measure_duration_ms`. No write path counts bytes.

## Notes for reviewers

- Read the commits after the skeleton in order: `Tagged` → account constructors → account folds → component state → cache shell → component folds → account folds → golden test → reporter → module map. Each carries its own tests.
- Component states are keyed by protocol system then id; the system is `block.extractor`, the same string the RPC uses to pick a window.
- ENG-6303 asked for a running byte total reconciled by a walk. This PR publishes the walk alone; the ticket's acceptance criteria are updated to match.
- Still `#![allow(dead_code)]`: construction and the reporter task come with the startup load (ENG-6305).
- Skeleton commits are shared with #1306. Merging this PR merges the skeleton; #1306 can then be closed.

## Follow-ups

- ENG-6305 constructs the cache, spawns `run_reporter`, replaces `DiscardSink`, and reads `entity_cache_measure_duration_ms` on dev. If a walk takes seconds, chunk it over a collected key list.
- ENG-6304 snapshot read fills `AccountTags` from `valid_from`.
- ENG-6293 reads through `EntityCache::read` and the tag getters.


crates/tycho-indexer/src/services/state/cache.rs
```diff
@@ -0,0 +1,1333 @@
...
+    pub(crate) fn from_creation(delta: &AccountDelta, tag: WriteTag) -> Self {
+        let account = delta.clone().into_account_without_tx();
+        let tags = AccountWriteTags::uniform(&account, tag);
+        Self::from_snapshot(account, tags)
+    }
+
+    /// Applies one delta; every changed value gets `tag`, values with a newer tag stay. A deleted
+    /// slot becomes the zero value, as in [`Account::apply_delta`]. A delta that carries code
+    /// replaces the code and its hash together.
+    pub(crate) fn apply_delta(&mut self, delta: &AccountDelta, tag: WriteTag) {
```
brunoguerios: Used to be named `fold` - Claude suggested to rename to remove any confusion with fold from Window, which refers to a batched operation.
Fine if you think the previous name was better.

crates/tycho-indexer/src/services/state/cache.rs
```diff
@@ -0,0 +1,1333 @@
...
+///
+/// The timestamp is the unit of the database's `valid_from`. The number orders blocks that share
+/// a timestamp — consecutive blocks do on fast chains — the way the transaction index does in the
+/// database. A snapshot row loads with number 0, so a folded block at the same timestamp still
+/// applies.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub(crate) struct WriteTag {
+    pub(crate) block_ts: NaiveDateTime,
+    pub(crate) block_number: u64,
+}
```
brunoguerios: Not sure this is overcomplicating things, but I thought Claude's suggestion to have it as a struct naming the attributes and also having both `ts` and `blockNumber` together could be beneficial.

crates/tycho-indexer/src/services/state/cache.rs
```diff
@@ -0,0 +1,1333 @@
...
+
+    /// Moves the entry to `tag` and returns `true`, unless `tag` is older: the single writer
+    /// folds in order, so an older block is a replay and is skipped.
+    fn advance_to(&mut self, tag: WriteTag) -> bool {
+        if tag < self.updated_at {
+            return false;
+        }
+        self.updated_at = tag;
+        true
+    }
```
brunoguerios: Used to be `accepts`, but I agreed with Claude that since it mutates the tag, it should indicate that action in the name itself.

crates/tycho-indexer/src/services/state/cache.rs
zizou0x: Is it needed? I remember thinking this through during planing and timestamp alone seemed to be enough.

Right now it seems the block_number isn't used at all, if that's the case I'd keep it like on the skeleton and use timestamp only

crates/tycho-indexer/src/services/state/cache.rs
```diff
@@ -0,0 +1,1333 @@
...
+
+    pub(crate) fn written_at(&self) -> WriteTag {
+        self.written_at
+    }
+
+    /// Writes `value` at `tag` unless this already holds a newer value. Equal tags apply: delta
+    /// values are absolute, so re-applying a block is harmless.
+    pub(crate) fn write(&mut self, value: T, tag: WriteTag) {
+        if tag < self.written_at {
```
zizou0x: Any particular reason why you decided to re-apply the block instead of ignoring it? I think it should be the same value no? Maybe we should add a check for when it's == that the values are the same and warn if they arent.

crates/tycho-indexer/src/services/state/cache.rs
```diff
@@ -0,0 +1,1333 @@
...
+
+impl From<&Block> for WriteTag {
+    fn from(block: &Block) -> Self {
+        Self { block_ts: block.ts, block_number: block.number }
+    }
+}
+
+/// A cached value together with the tag of the write that set it.
+#[derive(Debug, Clone, PartialEq)]
+pub(crate) struct Tagged<T> {
```
zizou0x: Not 100% convinced about this naming, I guess it could just be something like `Timestamped<T>`? Tag feels too broad here, wdyt?

crates/tycho-indexer/src/services/state/cache.rs
```diff
@@ -0,0 +1,1333 @@
...
+            newest = newest.max(slot.written_at());
+        }
+        for balance in self.token_balances.values() {
+            newest = newest.max(balance.written_at());
+        }
+        newest
+    }
+
+    /// Builds the [`Account`] the cached values describe.
+    pub(crate) fn materialize(&self, chain: Chain, address: &Address) -> Account {
```
zizou0x: Any reason why we don't cache Chain and Address? Like add it to CachedAccount so we don't need this function and we can just `impl From`

crates/tycho-indexer/src/services/state/cache.rs
```diff
@@ -0,0 +1,1333 @@
...
+    fn advance_to(&mut self, tag: WriteTag) -> bool {
+        if tag < self.updated_at {
+            return false;
+        }
+        self.updated_at = tag;
+        true
+    }
+
+    /// Builds the [`ProtocolComponentState`] the cached values describe.
+    pub(crate) fn materialize(&self, component_id: &str) -> ProtocolComponentState {
```
zizou0x: Same comment, here, why not caching the id?

crates/tycho-indexer/src/services/state/cache.rs
```diff
@@ -0,0 +1,1333 @@
...
+impl FoldSink for EntityCache {
+    /// Applies the whole block under the write lock. Not folded: `new_tokens`, `component_tvl`,
+    /// `dci_update` — those stay database-served.
+    ///
+    /// No check can fail today. Any future check goes before the first mutation, so a replay of
+    /// the same block converges.
+    fn fold(&self, block: &BlockAggregatedChanges) -> Result<(), StorageError> {
+        let mut state = self.write_lock();
+        state.fold_components(block);
+        state.fold_accounts(block);
```
zizou0x, line 617: What's the motivation for having two separate functions?

crates/tycho-indexer/src/services/state/cache.rs
```diff
@@ -0,0 +1,1333 @@
...
+
+    /// Write handle for the startup load. Nothing reads or folds while it exists.
+    pub(crate) fn loader(&self) -> CacheLoader<'_> {
+        CacheLoader(self.write_lock())
+    }
+}
+
+/// Inserts entries without a tag check, under one write lock held for as long as the handle
+/// lives. It exists only for the startup load, which runs before the extractors start.
+pub(crate) struct CacheLoader<'a>(RwLockWriteGuard<'a, CacheState>);
```
zizou0x: Why is that needed?

crates/tycho-indexer/src/services/state/cache.rs
brunoguerios: This was modified based on a feedback from Glados - here's more context on why it's needed:

Because a timestamp alone cannot both accept replays and reject older blocks once two blocks share a timestamp.

The comparison on a bare timestamp makes those two blocks indistinguishable. Several extractors write the same account, and they sit at different points of the chain, so block 4 can be folded after block 5. With one timestamp, block 4's write passes the `>= timestamp` check and overwrites what block 5 wrote — a value regresses.

The block number is the tiebreaker. `Derived Ord` compares `block_ts` first, then `block_number` (so the field order in the struct is load-bearing):

- Different timestamps → the timestamp decides, as before.
- Same timestamp → the higher block wins, the way the transaction index orders within one version in the database.
- Identical tag → still applies; that is the replay case, and it is harmless.

crates/tycho-indexer/src/services/state/cache.rs
brunoguerios: Claude's main argument here was memory allocation - it believed extracting Chain and Address from within the cache object would save significant memory, but I challenged it now and it actually not that significant 🤷‍♂️ 
I think the convenience of keeping it `impl From` is enough to move those into the object. I'll refactor it here.

crates/tycho-indexer/src/services/state/cache.rs
brunoguerios: Same argument as above - I'll apply the same refactor ✅

crates/tycho-indexer/src/services/state/cache.rs
brunoguerios: No - it's a leftover from previous implementation that had only timestamp for comparison. In that case, it was following the "not older" comparison. Now that we also track blockNumber, this problem no longer exist and we should be able to compare with `<=` so we skip when it's the same `timestamp` AND `blockNumber`.

crates/tycho-indexer/src/services/state/cache.rs
brunoguerios: For readability.
To me it's much easier to follow when they are separated like this. Trying to avoid having 200 line functions 😅. 
Would you rather keep them together?
Let me know and I'll refactor here 😉

crates/tycho-indexer/src/services/state/cache.rs
brunoguerios: Claude suggested this to better separate the startup load flow.

Main benefit is that this loader holds the write lock once, then writes everything and releases at the end - previous implementation with `insert_loaded_component` and `insert_lodade_account` would hold/release on every write.

Main difference from these startup load writes and `fold` is that they bypass the tag comparison.

When I challenged Claude on why keeping the loader, it actually doubled down and said this initial load should be built and provided as input in the construction, because these inserts bypassing the tag comparison should not be available post-construction.

I get that this would be the ideal scenario, but I'd say we should be fine as is for now. During the Startup Load story we can re-evaluate if a refactor in that direction is useful or not.

Let me know if you think the single write lock/release is enough reason to keep this or if you'd rather follow a different approach.

crates/tycho-indexer/src/services/state/cache.rs
```diff
@@ -533,18 +539,27 @@ impl CacheState {
...
+                .account_deltas
+                .contains_key(address)
+            {
+                continue;
+            }
             match self.accounts.get_mut(address) {
-                Some(entry) => entry.apply_balances(balances, tag),
+                Some(entry) => entry.apply_block(None, Some(balances), tag),
                 None => trace!(%address, "Balances for an unknown account skipped"),
             }
```
zizou0x: Wouldn't it be cleaner to collect changes into an iterator first and then iterate and apply them all at once?

crates/tycho-indexer/src/services/state/cache.rs
```diff
@@ -330,53 +330,57 @@ impl CachedComponentState {
...
+    /// Applies one block's changes, unless the entry already holds that block or a newer one. One
+    /// tag covers the whole entry, so the guard is here rather than per value. Attribute updates
+    /// apply before deletions, like [`ProtocolComponentState::apply_state_delta`].
+    pub(crate) fn apply_block(
+        &mut self,
+        delta: Option<&ProtocolComponentStateDelta>,
+        balances: Option<&HashMap<Bytes, ComponentBalance>>,
+        tag: WriteTag,
+    ) {
+        if tag <= self.updated_at {
```
zizou0x: Is this expected to happen? The behaviour is correct but probably we should add a log here to be able to monitor?

crates/tycho-indexer/src/services/state/cache.rs
```diff
@@ -0,0 +1,1415 @@
...
+        for (address, delta) in &block.account_deltas {
+            if delta.change_type() == ChangeType::Deletion {
+                if self
+                    .accounts
+                    .get(address)
+                    .is_some_and(|entry| entry.newest_write() < at)
+                {
+                    self.accounts.remove(address);
+                }
+                continue;
```
zizou0x: I think this is "newest_write" solution is flaky. Account deletions aren't expected AFAIK, so I would just log a warning and ignore (keep it in the cache, not a big risk).

Then we can get rid of the "newest_write" function

crates/tycho-indexer/src/services/state/cache.rs
```diff
@@ -0,0 +1,1415 @@
...
+                            at,
+                        )
+                    });
+            }
+        }
+        let Some(system_components) = self
+            .components
+            .get_mut(&block.extractor)
+        else {
+            trace!(system = %block.extractor, "Changes for a system with no cached components skipped");
```
zizou0x: I'd put all these "unknown component" at least at debug level, I don't think this is expected to happen at all, maybe it should even be a warn for now while we develop and we can adjust later if it's spammy.

crates/tycho-indexer/src/services/state/cache.rs
zizou0x: I would completely remove this for now, and see what we can do when doing the startup load implementation. I am almost sure there is a cleaner solution that will appear. We should most likely follow the same pattern we have for DCI or TokenCache: we give a db connection to an initialization function and the EntityCache initializes itself.

brunoguerios: Ok - I'll remove it 👍

crates/tycho-indexer/src/services/state/cache.rs
```diff
@@ -0,0 +1,1415 @@
...
+    pub(crate) fn written_at(&self) -> WriteTimestamp {
+        self.written_at
+    }
+
+    /// Writes `value` at `at` unless this already holds a value from that block or a newer one.
+    /// An equal timestamp is the same block folded again, so there is nothing to apply.
+    pub(crate) fn write(&mut self, value: T, at: WriteTimestamp) {
+        if at <= self.written_at {
+            return;
+        }
```
zizou0x, line 127: How hard would it be to make T: PartialEq and log a warn if they aren't eq? 

Basically I think we would like to know whenever there was a write that was ignored but the value was actually not the same. I think it would be valuable for debugging.

crates/tycho-indexer/src/services/state/cache.rs
zizou0x: We have a function in the extractor that append a microsecond to the timestamp to make sure the ordering stays correct https://github.com/propeller-heads/tycho/blob/74839dad6079954cfa8a4029c73350d90faed557/crates/tycho-indexer/src/extractor/protocol_extractor.rs#L1048-L1069. That's also reflected in the db:

image


I am not against having the block here too, just need to make sure all edge cases are covered. I think we have two options: 

1. Rely on the extractor appending microseconds, we only keep timestamp and we don't expect timestamp conflicts to happen. We need to confirm that the microsecond appended timestamps are always available to whoever insert data in the cache.
2. Keep block numbers but make it clean. The initialization to 0 isn't clean, it should be None if we don't know, 0 is a real actual block number so it's confusing. Also we should polish the struct interface to gate the creation for each path: on snapshot we take a path that only sets timestamps and no block and on update we need both, the field themselves would have to stay private to not allow creating struct without using these two creation function. We can add read functions for the readers.

crates/tycho-indexer/src/services/state/cache.rs
brunoguerios: Done. `zip_changes` pairs every touched key with its delta and its balances, and both folds walk that once. fold_components got the same shape.

crates/tycho-indexer/src/services/state/cache.rs
brunoguerios: No, it's not expected in the happy path. With the window starting every replay above the DB cursor, the guard only fires on a replay overlap or an ordering fault. Added a warn ✅

crates/tycho-indexer/src/services/state/cache.rs
brunoguerios: Removed the deletion and added a warn in case we reach this arm.

crates/tycho-indexer/src/services/state/cache.rs
brunoguerios: Raised them to `warn` ✅

crates/tycho-indexer/src/services/state/cache.rs
brunoguerios: I tried to simply `warn` in place if `at` == `self.written_at`, but at that point we have no meaningful information other than the timestamp.
The solution I found was to return a `WriteOutcome` and warn within the `apply_block` implementation.
Added a bit of overhead, but I'd say it's worth the trade-off 🤷‍♂️

crates/tycho-indexer/src/services/state/cache.rs
```diff
@@ -509,97 +538,84 @@ impl CacheState {
...
-                    trace!(system = %block.extractor, %id, "State delta for an unknown component skipped")
+                Some(entry) => entry.apply_block(delta, balances, at),
+                None if block
+                    .new_protocol_components
+                    .contains_key(id) =>
+                {
+                    system_components.insert(
+                        id.clone(),
+                        CachedComponentState::from_creation(id, delta, balances, at),
+                    );
```
zizou0x, line 555: Wouldn't it be cleaner to just apply the new protocol component first then apply the state on top, so we can get rid of this edge case

crates/tycho-indexer/src/services/state/cache.rs
```diff
@@ -509,97 +538,84 @@ impl CacheState {
...
-            if block.state_deltas.contains_key(id) {
-                continue;
-            }
-            match system_components.get_mut(id) {
-                Some(entry) => entry.apply_block(None, Some(balances), at),
-                None => {
-                    trace!(system = %block.extractor, %id, "Balances for an unknown component skipped")
-                }
+        for id in block.new_protocol_components.keys() {
+            if !system_components.contains_key(id) {
```
zizou0x, line 566: Also an unexpected case, let's add a warning if we're trying to add a component that is already cache

crates/tycho-indexer/src/services/state/cache.rs
```diff
@@ -51,26 +54,44 @@ use super::window::FoldSink;
...
 impl From<&Block> for WriteTimestamp {
     fn from(block: &Block) -> Self {
-        Self { block_ts: block.ts, block_number: block.number }
+        Self::new(block.ts, block.number)
     }
 }
 
+/// What a write did to a [`Timestamped`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum WriteOutcome {
```
zizou0x, line 83: Can't it be private? Same for the functions that uses it

zizou0x (approved):
Just some minor comments

crates/tycho-indexer/src/services/state/cache.rs
brunoguerios: Problem here is that we now have a guard that skips updates on components in the exact same block, so creating them first and later updating their state doesn't work.
The previous guard that allowed updates on the same block (expecting them to be a no-op) would work and allow us to remove these edge cases.
I'd say allowing them to re-apply looks cleaner, but I'd say it's fine as is.
Wdyt?

brunoguerios: Found a way to collect everything into an iterator and then applying at once - done in the last commit ✅

PR #1467, opened 2026-09-19
feat: add Arc chain and native USDC model

tvinagre:
## Summary

- Add Arc mainnet as a built-in chain with chain ID `5042` in the Rust and Python wire models.
- Model Arc native USDC and routable USDC as two representations of one shared balance.
- Keep Arc out of native-wrapper simulation and wrapper-only encoding paths.
- Seed both Arc token representations with idempotent native-price records.


crates/tycho-common/src/models/mod.rs
```diff
@@ -95,10 +96,108 @@ pub enum Chain {
...
+    /// A contract converts between separate native and wrapped-token balances.
+    Wrapper,
+    /// Native and routable representations expose the same economic balance.
+    SharedBalance,
+    /// The chain has no distinct routable representation or wrapper contract.
+    NativeOnly,
+}
+
+/// Descriptive alias for [`NativeAssetKind`] at relationship-oriented call sites.
+pub type NativeAssetRelationship = NativeAssetKind;
```
brunoguerios: `NativeAssetKind`, this alias, `kind()` and `relationship()` have no callers outside tests. The alias is a second name for one concept, and `relationship()` duplicates `kind()`. Callers that need the discriminant can `match` or `matches!` on `NativeAsset`. Suggest deleting all four.

crates/tycho-common/src/models/mod.rs
```diff
@@ -95,10 +96,108 @@ pub enum Chain {
...
+    pub fn representations(&self) -> impl Iterator<Item = &Token> {
+        let native = self.native_token();
+        std::iter::once(native).chain(
+            self.routable_token()
+                .filter(move |routable| native.address != routable.address),
+        )
+    }
+
+    /// Converts native precision into routable precision, truncating non-representable dust.
+    pub fn native_to_routable_amount(&self, amount: &BigUint) -> Option<BigUint> {
```
brunoguerios: `native_to_routable_amount`, `routable_to_native_amount` and `convert_token_precision` have no callers. Suggest deleting them until a consumer exists, or name the follow-up PR that uses them in the description.

crates/tycho-common/src/models/mod.rs
```diff
@@ -95,10 +96,108 @@ pub enum Chain {
...
+    pub fn wrapper(&self) -> Option<&Token> {
+        match self {
+            NativeAsset::Wrapper { native: _, wrapper } => Some(wrapper),
+            NativeAsset::SharedBalance { native: _, routable: _ } |
+            NativeAsset::NativeOnly { native: _ } => None,
+        }
+    }
+
+    /// Returns each distinct token representation of this economic asset.
+    pub fn representations(&self) -> impl Iterator<Item = &Token> {
```
brunoguerios, line 141: The address-dedupe filter is dead: the `Custom` arm normalises equal addresses to `NativeOnly` at construction and no built-in arm has equal addresses. This can be `once(native).chain(self.routable_token())`, and `test_native_asset_representations_deduplicate_matching_addresses` can go, since it builds an invariant-violating value by hand.

brunoguerios (commented):
I checked the model against the Arc docs (stablecoin-native model page): zero-address 18-decimal native USDC, 6-decimal ERC-20 at `0x3600…`, no wrapper, chain ID 5042. The model matches. The seeded price of `10^decimals` for both tokens is right: TVL is `balance_atomic / price`, so `1e6` for the 6-decimal token means one whole routable USDC equals one whole native USDC. Curve encoding and native-wrapper injection refuse Arc, which is correct.

No blockers. Inline comments cover the tycho-common API surface and two test nits. The points below touch files outside the diff.

**Docs**
- `docs/for-solvers/execution/encoding/README.md` (Native Tokens section) and `docs/for-solvers/execution/migration-guide-v2-to-v3.md` say the stream injects a `native_wrapper` component on every chain. After this PR that is false for Arc. Say: on every chain with a wrapper contract. Add one sentence for Arc: native USDC and ERC-20 USDC share one balance, there is no wrapper, route to the ERC-20 at `0x3600…` directly.
- `docs/for-solvers/self-hosted-evm-chain.md`: equal `native` and `wrapped_native` addresses now mean no wrapper and no `native_wrapper` component. The example config uses the zero address for WETH, which disables the wrapper. State the rule in the field table and give the example a real WETH address.

**PR description**
- Say that `wrapped_native_token()` and `try_wrapped_native_token()` now return `Option<Token>`, that Starknet returns `None` instead of a zero-address WETH, and that `NativeWrapperState::new` returns `Option<Self>`. Downstream solvers depend on tycho-common and need this in the release notes.
- Mention the stream change: the native-wrapper component is now injected on the first successful update instead of the first stream item, so a leading decode error no longer suppresses it. Good fix, worth a line.

**Suggestions outside the diff**
- `tycho-client` `default_timing`: Arc falls into the default arm and gets `(1, 3, 50)`. Robinhood, also sub-second, has an explicit `(1, 5, 100)`. Add an Arc arm with the same values.
- `native_token()` runs inside a per-balance loop in `tycho-storage` `contract.rs` and per entrypoint in the DCI entrypoint generator. See the inline comment on `try_native_token` for the cost.
- Quickstart: Arc has no default sell token, so it panics without `--sell-token`. Since the default buy token on Arc is now the ERC-20 USDC, the natural default sell token is EURC or USYC. Add an arm when the addresses are confirmed, or note the flag in the Readme.

**Custom chains**
Config schema is unchanged and existing `chains.yaml` files keep working. A custom chain with equal native and wrapped addresses becomes `NativeOnly`: `wrapped_native_token()` returns `None`, injection stays skipped, `ensure_chain` writes the same rows. The one runtime change is that `CurveSwapEncoder::new` now fails for such a chain, reachable only through manual `register_encoder`. Fine, and the error names the chain.

**For the first Arc protocol integration**
The Arc docs say each USDC move emits two `Transfer` logs: one from a system emitter at 18 decimals and one from the ERC-20 at 6 decimals. A substream that derives balances from `Transfer` logs must filter on emitter, and `0x3600…` balances must come from `balanceOf`, not `.balance`, or dust below 1e-6 USDC leaks in.

crates/tycho-common/src/models/mod.rs
```diff
@@ -419,67 +542,134 @@ impl Chain {
     /// Like [`Chain::native_token`] but returns [`ChainConfigError::UnknownChain`] when a custom
     /// chain has no registered config.
     pub fn try_native_token(&self) -> Result<Token, ChainConfigError> {
```
brunoguerios, line 489: This now builds the full `NativeAsset` (two `Token`s plus a hex parse of the wrapper address) and clones the native one. `native_token()` runs on hot paths: inside the per-balance loop in `tycho-storage` `contract.rs` and per entrypoint in the DCI entrypoint generator. Either keep a direct native-only match here, or hoist the call out of those loops in a follow-up.

crates/tycho-common/src/models/mod.rs
```diff
@@ -419,67 +542,134 @@ impl Chain {
...
+            },
+            Chain::Robinhood => NativeAsset::Wrapper {
+                native: native_eth(Chain::Robinhood),
+                // aeWETH is the Arbitrum bridge wrapper for native ETH and exposes WETH9.
+                wrapper: wrapped_native_eth(
+                    Chain::Robinhood,
+                    "0x0Bd7D308f8E1639FAb988df18A8011f41EAcAD73",
+                ),
+            },
+            Chain::Arc => NativeAsset::SharedBalance {
```
brunoguerios, line 601: Worth a short WHY comment with the Arc docs link here or on the two helpers: USDC is the gas token, the ERC-20 at `0x3600…` exposes the same balance at 6 decimals, no wrapper is deployed or supported. Readers of the `SharedBalance` arm otherwise have to go find the source.

crates/tycho-common/src/models/mod.rs
```diff
@@ -419,67 +542,134 @@ impl Chain {
...
+        Ok(self
+            .try_native_asset()?
+            .wrapper()
+            .cloned())
+    }
+
+    /// Returns the routable representation of the native asset, if one exists. Panics if a custom
+    /// chain has no registered config; use [`Chain::try_routable_native_token`] for a non-panicking
+    /// variant.
+    pub fn routable_native_token(&self) -> Option<Token> {
```
brunoguerios: `routable_native_token()` and `try_routable_native_token()` have one caller, the quickstart, which can use `chain.native_asset().routable_token()`. Dropping them keeps the `Chain` surface at `native_token`, `wrapped_native_token`, `native_asset`. Not a strong opinion.

crates/tycho-common/src/models/mod.rs
```diff
@@ -872,6 +1094,125 @@ mod tests {
         assert_eq!(Chain::Robinhood.block_time_secs(), 1);
     }
 
+    #[test]
+    fn test_arc_chain_identity_and_dto_round_trip() {
+        assert_eq!(Chain::Arc.id(), 5042);
+        assert_eq!(Chain::Arc.to_string(), "arc");
+        assert_eq!("arc".parse::<Chain>().unwrap(), Chain::Arc);
+        assert!("5042002".parse::<Chain>().is_err());
```
brunoguerios: `FromStr` parses chain names, never IDs, so this asserts nothing Arc-specific. Suggest removing.

crates/tycho-storage/src/postgres/mod.rs
```diff
@@ -1544,6 +1547,56 @@ mod tests_ensure_chain {
...
+            .select((schema::account::address, schema::token::decimals, schema::token_price::price))
+            .order(schema::account::address.asc())
+            .load(&mut conn)
+            .await
+            .unwrap();
+
+        let mut routable_address = vec![0u8; 20];
+        routable_address[0] = 0x36;
+        assert_eq!(rows, vec![(vec![0u8; 20], 18, 1e18), (routable_address, 6, 1e6)]);
+        for (_, decimals, atomic_price) in rows {
```
brunoguerios: This loop restates the exact-row assertion above it. Keep one.

brunoguerios (approved):

crates/tycho-common/src/models/mod.rs
tvinagre: Addressed in `46237c2e8`: removed `NativeAssetKind`, its alias, and the duplicate `kind()` / `relationship()` accessors; tests now match on `NativeAsset` directly.

crates/tycho-common/src/models/mod.rs
tvinagre: Addressed in `46237c2e8`: removed the unused precision-conversion APIs and their dedicated test.

crates/tycho-common/src/models/mod.rs
tvinagre: Addressed in `46237c2e8`: `representations()` now directly chains native and routable representations, and the invariant-violating dedupe test is gone.

crates/tycho-common/src/models/mod.rs
tvinagre: Addressed in `46237c2e8`: `try_native_token()` again matches built-in/custom chains directly, avoiding construction of the full `NativeAsset` on hot paths.

crates/tycho-common/src/models/mod.rs
tvinagre: Addressed in `46237c2e8`: added a concise Arc-specific WHY comment with the authoritative stablecoin-native model link.

crates/tycho-common/src/models/mod.rs
tvinagre: Addressed in `46237c2e8`: removed both routable-native convenience methods; quickstart now reads the routable token from `native_asset()`.

crates/tycho-common/src/models/mod.rs
tvinagre: Addressed in `46237c2e8`: removed the unrelated numeric `FromStr` assertion.

crates/tycho-storage/src/postgres/mod.rs
tvinagre: Addressed in `46237c2e8`: removed the redundant per-row loop and kept the stronger exact-row assertion.

